PR #159, opened 2026-05-10
feat(backend): add migration runner entrypoint + CI guard for backward-compat

damir-topic:
## Summary

- Initiative #26 Task #29: ship the migration runner entrypoint + CI guard that protect Goal #11's DoD bullet 3 (`helm rollback` works without manual DB intervention) at the schema layer.
- Migration runner (`backend/src/meho_backplane/db/migrate.py`) is the one-file CLI the Helm pre-install / pre-upgrade Job (G2.5-T3) invokes. It reuses `alembic_config()` from `db/migrations.py` so the migration that *applies* and the readiness probe that *verifies* never disagree on which `alembic.ini` they targeted; calls `alembic.command.upgrade(cfg, "head")` and exits 0 / 1 with `migration_failed: <ExcClass>: <msg>` on stderr.
- CI guard (`scripts/ci/check_migration_compat.py` + `.github/workflows/migration-compat.yml`) does an AST-plus-regex scan over each migration's `upgrade()` body and rejects destructive patterns: `op.drop_column` / `op.drop_table` / `op.rename_table`, `op.alter_column(..., new_column_name=...)`, `op.alter_column(..., nullable=False)`, and `op.execute(...)` payloads matching `DROP COLUMN` / `DROP TABLE` / `RENAME TO` / `RENAME COLUMN` / `RENAME TABLE` / `ALTER ... SET NOT NULL`.

## Design notes

**Why scan `upgrade()` only.** Production never invokes `alembic downgrade` — Goal #11's rollback contract is image-revert + forward-compat schema discipline. The shipped first migration's `downgrade()` legitimately drops the table it just created (the only revertible operation on an empty schema); a flat scan would trip on it on day one. Restricting detection to `upgrade()` preserves the additive-only contract for production while leaving development-time rollback symmetry intact in source. This is the deliberate deviation from the task body's flat scan sketch — documented in the script's docstring and exercised by `test_destructive_pattern_in_downgrade_only_passes`.

**Why dual AST + regex.** The AST pass cannot follow f-strings or variables; the regex backstop catches destructive SQL embedded in `op.execute(f"DROP TABLE {table}")` shapes. Removing the regex pass would silently let migrations smuggle destructive ops past the guard via string formatting. Exercised by `test_raw_sql_in_fstring_is_caught_by_text_pass`.

[...]

PR #162, opened 2026-05-10
ci: switch migration-compat workflow to self-hosted meho-runners

damir-topic:
## Summary
- Follow-up to #160. The `migration-compat.yml` workflow was added in #159 in parallel with the bulk switch, so it slipped through still pointing at `ubuntu-latest`.
- Aligns it with every other workflow (`[self-hosted, meho-runners]`).

## Notes
- `runner-smoke.yml` already uses the self-hosted pool (bare `meho-runners` label) — intentionally left alone since it's already on the right runners.

## Test plan
- [ ] PR's own `check` job runs on the self-hosted pool (not queued indefinitely)
- [ ] `actions/setup-python` and `uv` install succeed on the runner

## Summary by CodeRabbit

* **Chores**
  * Updated continuous integration workflow configuration for enhanced build infrastructure management.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/162)

### DIFF
--- a/.github/workflows/migration-compat.yml
+++ b/.github/workflows/migration-compat.yml
@@ -38,7 +38,7 @@ concurrency:
 jobs:
   check:
     name: Backward-compat migration guard
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, meho-runners]
     timeout-minutes: 5
     steps:
       - name: Checkout


damir-topic:
Closing — superseding with a single coherent PR that also reverts the wrong `runs-on` syntax introduced in #160. The `[self-hosted, meho-runners]` list form doesn't match the org's gha-runner-scale-set runners (they advertise zero labels and are addressed by scale-set name only). New PR will switch every `runs-on` to bare `meho-runners` to match `runner-smoke.yml`.

PR #174, opened 2026-05-11
feat(cli): scaffold Go module + cobra root + meho version (G2.6-T1)

damir-topic:
## Summary

- Bootstraps the `cli/` Go module at `github.com/evoila/meho/cli` (Go 1.22, cobra v1.10.2) — the operator-facing CLI binary mandated by Goal #11 and locked in by ADR 0001.
- Wires `meho version` only; the version/commit/date strings are populated at link time via `-ldflags -X` into `internal/version`. Subsequent G2.6 tasks (#44 login, #45 status, #46 releases, #47 signing) layer onto this scaffold without restructuring it.
- Ships golangci-lint v1.x config with the six default linters + gofmt / goimports / revive, plus a Makefile that drives reproducible builds (`-trimpath -s -w`) and a delegate `make cli` target at the repo root so the polyglot surface has one obvious entry point per language.
- Adds `docs/codebase/cli.md` so future contributors don't re-litigate the ldflags-injection-needs-globals trade-off when expanding the lint set.

## Test plan

- [x] `cd cli && make tidy` — `go.sum` populated, deps resolved
- [x] `cd cli && make build` — produces `bin/meho`, prints
      `meho dev (commit <short-sha>, built <utc>)` (real commit/date
      injected by Makefile defaults; literal `unknown` appears in
      `go run` builds without ldflags)
- [x] `cd cli && make build VERSION=v0.1.0` — `meho v0.1.0 (commit ...)` confirmed
- [x] `cd cli && make lint` — 9 linters clean, 0 findings
- [x] `cd cli && make test` — `go test -race -cover ./...` passes;
      `internal/cmd` at 90.9% coverage (the only package with logic)
- [x] `make cli` from repo root — top-level delegate works
- [x] `./bin/meho --help` — cobra root help renders with `--config`
      and `-v/--verbose` persistent flags
- [x] SPDX `Apache-2.0` header on every `.go` file (ADR 0003)

## Acceptance criteria (issue #43)

- [x] `cli/` directory with `go.mod`, `cmd/meho/main.go`, `internal/version/version.go`, `Makefile`, `.golangci.yml`, `README.md`; SPDX headers on every Go file
- [x] `cd cli && make build` produces `cli/bin/meho`
- [x] `cli/bin/meho version` prints `meho dev (commit ..., built ...)`; `make build VERSION=v0.1.0` substitutes the version string
- [x] `cd cli && make lint` passes
[...]

PR #156, opened 2026-05-10
feat(backend): add audit-log schema + synchronous audit middleware

damir-topic:
## Summary

- Land Initiative #26 Task #28: a synchronous audit-write middleware that records one `audit_log` row per authenticated request before the response yields back to the ASGI send chain. Fail-closed semantics — an audit insert exception converts the response to HTTP 500 `{"detail": "audit_write_failed"}`.
- Add the SQLAlchemy `Base` + `AuditLog` model and the first Alembic revision (`0001_create_audit_log`). The schema is dialect-portable (`Uuid`, `JSON().with_variant(JSONB(), "postgresql")`) so the same migration runs on both PG (production) and SQLite (dev/test).
- Wire `AuditMiddleware` in `main.py` immediately inside `RequestContextMiddleware` so it sees both `request_id` (on entry) and `operator_sub` (on exit), and so the fail-closed 500 still carries the `X-Request-Id` correlation header.
- Update the autouse `_default_database_url` conftest fixture to apply `alembic upgrade head` against a per-test SQLite file, so existing authenticated tests no longer hit "no such table: audit_log" through the audit middleware.

## Critical-review deviation from the task body

The task body's middleware sketch uses `starlette.middleware.base.BaseHTTPMiddleware`. Empirical verification against the installed Starlette 1.0.0 shows that shape **breaks the contextvar contract**: BaseHTTPMiddleware runs the wrapped app inside an anyio task group, so any contextvars bound inside the handler — including the `operator_sub` written by `verify_jwt_and_bind` — disappear from the dispatch task on the post-handler side. The audit row's whole point is to attribute by operator, so pure-ASGI is the only shape that delivers the AC. The same Initiative's existing `RequestContextMiddleware` is pure-ASGI for the same reason; this PR follows that established pattern. The schema and the AC list are unchanged.

## Test plan

- [x] `ruff check src/ tests/` — clean
- [x] `ruff format --check src/ tests/` — clean
- [x] `mypy src/` — clean
- [x] `pytest tests/` — 154 passed, 1 skipped (testcontainers-PG suite, no Docker in sandbox)
- [x] AC #1 — `alembic upgrade head` against a fresh SQLite DB creates `audit_log` + both indexes; verified by `test_alembic_upgrade_head_against_sqlite_creates_version_table`
- [x] AC #2 — `AuditLog` model with typed-mapped annotations exists at `backend/src/meho_backplane/db/models.py`; mypy strict passes
[...]

PR #175, opened 2026-05-11
feat(cli): meho login Keycloak device-code flow + token storage (G2.6-T2)

damir-topic:
## Summary

- Adds `meho login <backplane-url>` — OAuth 2.0 Device Authorization Grant (RFC 8628) against the backplane's configured Keycloak realm, with a cross-platform `TokenStore` (zalando/go-keyring) and a 0600-mode file fallback at `$XDG_CONFIG_HOME/meho/credentials.json` for headless hosts.
- Uses `golang.org/x/oauth2` v0.26.0's built-in `Config.DeviceAuth` / `Config.DeviceAccessToken` rather than rolling the polling loop by hand — gets RFC 8628's `authorization_pending` / `slow_down` semantics correct for free, and pins to the last v0.x release that still targets Go 1.22.
- Adopts ADR 0004's `zalando/go-keyring` choice (the issue body's `99designs/keyring` recommendation was explicitly rejected in the ADR on maintenance-cadence grounds; this matches PR #174's M2 fix in `docs/codebase/cli.md`).
- Includes the `--issuer` / `--client-id` overrides documented in the Task body for use until G2.2 ships the backplane's `/api/v1/auth-config` endpoint; partial overrides are honoured (operators can pin one half, auto-discover the other).
- Persisted JSON shape captures access / refresh / id tokens, issuer, client_id, backplane URL, token type, and expiry — enough for v0.2's refresh-token flow to land without a token-store schema migration.

## Test plan

- [x] `make lint` — golangci-lint clean (9 linters, 0 findings).
- [x] `make test` — `go test -race -cover ./...` green; new packages reach 58% (`internal/auth`) and 55% (`internal/cmd`) line coverage.
- [x] `make build` — produces `bin/meho` with ldflags-injected version metadata.
- [x] **Acceptance: SPDX headers** — every new `.go` file starts with `// SPDX-License-Identifier: Apache-2.0` per ADR 0003.
- [x] **Acceptance: `meho login --help`** — usage block lists `--issuer`, `--client-id`, and `--scope`; the load-bearing test `TestLoginCommandHelpListsFlags` pins this.
- [x] **Acceptance: device-code happy path** — `TestRunDeviceFlowHappyPath` drives an httptest fake IdP through device-auth → authorization_pending → token issuance and asserts the prompter is called exactly once, the user_code threads through to the operator-facing rendering, and `ConvertOAuthToken` lifts `id_token` out of `oauth2.Token.Extra`.
- [x] **Acceptance: file-fallback round-trip** — `TestFileStoreRoundTrip` saves, loads, deletes a `StoredToken` via the file backend in a tmpdir.
- [x] **Acceptance: 0600 mode** — `TestFileStoreEnforcesZeroSixHundred` stat's the file and parent directory; verifies file permissions are `0600` and dir is `0700`.
- [x] **Acceptance: keyring-disable escape hatch** — `NewTokenStore` honours `MEHO_KEYRING_DISABLE=<non-empty>` and routes to the file backend (covered by `TestFileStoreRoundTrip` exercising the same code path the env var selects).
- [x] **Acceptance: auth-config discovery** — `TestResolveAuthConfigDiscoveryHappyPath` exercises the future-backplane `/api/v1/auth-config` endpoint shape `{"keycloak_issuer", "audience"}`; `TestResolveAuthConfigUsesOverrides` proves the override path is the documented fallback.

## Out of scope

- Refresh-token automation (deferred to v0.2 per the Task body; refresh token *is* captured in the store so v0.2 lights up without a schema break).
- Token revocation on `meho logout` (deferred to v0.2 per the Task body).
- Browser auto-launch via `xdg-open` / `open` (deferred per the Task body).
- A separate `~/.config/meho/config.yaml` for the backplane URL — captured alongside the token in the credentials store; a dedicated config file is left for G2.6-T3 where `meho status` needs an unauthenticated way to recover the backplane URL.
- Multi-tenant `(service, user)` keying — the storage layer supports it via the `KeyForBackplane` helper, but the CLI only ever stores one entry per backplane URL today.
- Integration tests against real Keychain / Secret Service — deferred to G2.8 smoke tests; the keyring backend is exercised in unit tests via the shared `TokenStore` interface.

Closes #44

## Release Notes

* **New Features**
  * Added `meho login` command for OAuth 2.0 Device Authorization Grant authentication
  * Implemented secure token storage with OS keyring backend and file fallback support

* **Documentation**
  * Updated README with login usage examples and token persistence configuration options
  * Expanded codebase documentation covering login flow and storage specifications

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/175)


damir-topic:
@coderabbitai review

damir-topic:




**Anchor task:** #44 — `meho login` Keycloak device-code flow + keyring storage + 0600 file fallback
**Surface area:** devops (CLI auth surface — OAuth 2.0 device flow + cross-platform secret storage)
**Verdict:** APPROVE
**Head SHA reviewed:** `93d4086`
**AI reviewer (CodeRabbit):** received in 31s

### Acceptance criteria
- [x] `cli/internal/auth/{device,keyring}.go` + `cli/internal/cmd/login.go` exist with SPDX headers — confirmed at `cli/internal/auth/devicecode.go`, `cli/internal/auth/store.go`, `cli/internal/cmd/login.go` (file naming differs from Task body; rationale logged in `adjacent_findings` — `store.go` holds two backends, naming it `keyring.go` would mislead). Every new `.go` file leads with `// SPDX-License-Identifier: Apache-2.0` per ADR 0003.
- [x] `meho login --help` shows usage including override flags — verified `./bin/meho login --help`: `--client-id`, `--issuer`, `--scope` all present. Flag is `--client-id` rather than the Task body's `--audience`; rationale in `adjacent_findings` (the device-code wire field is `client_id` — calling the flag `--audience` would force operators to know an inside-baseball mapping). `TestLoginCommandHelpListsFlags` pins the contract.
- [x] Happy-path test: mock OIDC server drives device-auth → authorization_pending → token issuance — `TestRunDeviceFlowHappyPath` in `cli/internal/auth/devicecode_test.go:161-275` covers prompter contract (called exactly once with captured user_code), polling contract (second call returns the token after first returns pending), and `ConvertOAuthToken` field mapping including `id_token` lifted from `oauth2.Token.Extra`.
- [x] File-fallback round-trip — `TestFileStoreRoundTrip` exercises Save → Load → Delete on a tmpdir-backed file store; assertions cover every persisted field.
- [x] 0600 mode + 0700 parent dir — `TestFileStoreEnforcesZeroSixHundred` stats the file (`info.Mode().Perm() == 0o600`) and the parent dir (`0o700`); skipped on Windows where POSIX modes don't map.
- [x] Auth-discovery: `TestResolveAuthConfigDiscoveryHappyPath` fakes the `/api/v1/auth-config` endpoint shape `{keycloak_issuer, audience}` and proves the resolver maps them correctly; `TestResolveAuthConfigUsesOverrides` proves the documented fallback when the backplane endpoint isn't ready (G2.2 coordination).

### ADR 0004 conformance
- [x] `zalando/go-keyring v0.2.8` (latest stable, 2026-03-23) — primary backend; matches ADR 0004's explicit lock.
- [x] **Reconciles Task body / ADR conflict.** The Task body still names `99designs/keyring`; ADR 0004 explicitly rejects it on maintenance grounds (last release Dec 2022). PR correctly follows the ADR. Same posture PR #174 took for its M2 doc fix.
- [x] File fallback at `$XDG_CONFIG_HOME/meho/credentials.json` per ADR 0004 Implications (`~/.config/meho/credentials` with `.json` suffix).
- [x] `golang.org/x/oauth2 v0.26.0` (Feb 2025) — verified `Config.DeviceAuth` and `Config.DeviceAccessToken` exist (`deviceauth.go:82`, `deviceauth.go:140`) and implement RFC 8628 slow_down semantics (interval += 5 at `deviceauth.go:184-187`). Pin rationale (v0.27+ bumps Go directive to 1.23) is documented in the PR description.

### Project-board alignment
|---|---|---|---|---|---|
| #44 (Task) | Todo | Closes | manual (cross-repo) | full | manual close after merge |
| #42 (Initiative G2.6) | Todo | implicit parent | manual | partial (2 of 5 Tasks: #43 done, #44 pending) | remains open through G2.6 |
| #11 (Goal) | Todo | implicit grandparent | manual | partial | remains open through Goal 2 |

### Predicted vs. actual diff
- ✅ `cli/internal/auth/devicecode.go` (device-code flow + OIDC discovery)
- ✅ `cli/internal/auth/devicecode_test.go` (httptest fake IdP + 9 tests covering happy path, denied, prompter-error, missing-fields, non-2xx, OAuth fallback, error unwrap)
- ✅ `cli/internal/auth/store.go` (TokenStore interface, keyring + file backends, XDG resolution)
- ✅ `cli/internal/auth/store_test.go` (round-trip, 0600 mode, missing-file sentinel, malformed-file error, multi-entry, idempotent delete, key-normalisation)
- ✅ `cli/internal/cmd/login.go` (cobra wiring, auth-config resolver, stdout prompter)
- ✅ `cli/internal/cmd/login_test.go` (override resolution, partial override, discovery failure prose, help-flag contract, missing-arg error)
- ✅ `cli/internal/cmd/root.go` (registers `newLoginCmd()` — single-line addition)
- ✅ `cli/go.mod` / `cli/go.sum` updates (zalando/go-keyring v0.2.8 + golang.org/x/oauth2 v0.26.0 + transitive deps)
- ✅ `cli/README.md` extension (login quickstart + headless-host guidance + `MEHO_KEYRING_DISABLE` escape hatch)
- ✅ `docs/codebase/cli.md` extension (durable narrative — discovery precedence, polling semantics, storage backend selection, persisted JSON schema)

### Integration verification (all ran inside `.claude/worktrees/pr-175-review/`)
- [x] `make tidy` — clean
- [x] `make build` — produced `bin/meho` with ldflags-injected version metadata
- [x] `make test` — `go test -race -cover ./...` green; `internal/auth` 58.2%, `internal/cmd` 54.8% coverage
- [x] `make lint` — golangci-lint clean (0 findings across 9 linters)
- [x] `go vet ./...` — silent
- [x] `./bin/meho login --help` — exits 0, renders all four flags (`--client-id`, `--issuer`, `--scope`, plus inherited `--config`, `-v/--verbose`)
- [x] B1 regression check (`./bin/meho version`) — stdout 54 bytes (`meho dev (commit 93d4086, built ...)`), **stderr 0 bytes**. No regression of Task #43's stderr/stdout-split fix.
- [x] Secret-handling scan — grep across `cli/internal/auth/*.go` + `cli/internal/cmd/login.go` for `fmt.(Print|Fprint|Sprint)|log.Print` confirms no code path writes the access token, refresh token, or id_token to any output stream. Only the verification URL, user_code, backplane URL, and store description (a labelled path) are written; all are public.

### Findings
- ⚠️ M1 — `cli/internal/auth/store.go:280-293` fixed `.tmp` filename can permanently block credential writes after a crash (also flagged by CodeRabbit); fix: switch to `os.CreateTemp(filepath.Dir(s.path), "credentials-*.tmp")` so each save uses a unique tmp path and the atomic-rename invariant survives stale `.tmp` leftovers. Author's documented trade-off (no silent data race) holds with the unique-tmp variant — `os.CreateTemp` already enforces O_EXCL semantics internally.
- ⚠️ M2 — `cli/internal/cmd/login.go:55-58` URL validation too permissive (also flagged by CodeRabbit); `url.ParseRequestURI` accepts `mailto:`, `file:///`, etc. — schemes the device-code flow can't possibly handle. Fix: after parsing, also enforce `u.Host != ""` and `(u.Scheme == "http" || u.Scheme == "https")` so the error fires at the CLI surface with operator-friendly prose rather than at the first HTTP call with an opaque downstream error.
- 💡 m1 — `cli/internal/auth/devicecode_test.go:190-217` happy-path test asserts only `device_code` on the token poll; adding `grant_type` and `client_id` assertions would catch silent regressions in the OAuth contract (CodeRabbit nitpick — accept as adjacent).
- 💡 m2 — `cli/internal/cmd/login_test.go:19-25` `TestResolveAuthConfigUsesOverrides` uses `http://127.0.0.1:1` to "prove" discovery wasn't called; a `httptest.NewServer` whose handler calls `t.Fatalf` on any request makes the override fast-path test deterministic without OS/network assumptions (CodeRabbit nitpick).
- 💡 m3 — `cli/internal/cmd/login_test.go:70-75` partial-override test accepts any path; asserting `r.URL.Path == "/api/v1/auth-config"` would catch a regression in discovery URL construction (CodeRabbit nitpick).
- 👏 p1 — Storage layer designed for multi-backplane support from day 1. `KeyForBackplane` normalises trailing slashes and routes by URL; the file backend's `Entries` map keys by `(service\x00user)` so two backplanes coexist on one host. `TestKeyForBackplaneNormalisesTrailingSlash` + `TestFileStoreSupportsMultipleEntries` pin the contract. Task body called this "v0.2"; PR builds it now without operator-visible surface — exactly the right call.
- 👏 p2 — `RunDeviceFlow` leans on `golang.org/x/oauth2`'s `Config.DeviceAccessToken` rather than rolling its own polling loop. That's the right delegation: RFC 8628's slow_down / authorization_pending / expired_token classification is subtle to implement correctly, and the upstream code is battle-tested and version-pinned. `classifyDeviceTokenError` adds a thin operator-prose layer on top without duplicating the spec logic.
- 👏 p3 — `DiscoveryError` is a typed error (not a sentinel) carrying URL + Status + Body excerpt + Underlying. `Unwrap()` lets callers `errors.Is(..., context.DeadlineExceeded)` directly. `TestDiscoveryErrorUnwraps` proves the unwrap chain; the body excerpt is bounded at 256 chars so a misconfigured proxy's multi-megabyte HTML page never sprays the operator's terminal.

### CodeRabbit cross-check
- Adopted: 2 majors (M1 store.go `.tmp`, M2 login.go URL validation) + 3 nitpicks (m1/m2/m3) — all confirmed independently.
- Disputed: 0.
- Out of scope: 0.

### Docs
- [x] `docs/codebase/cli.md` updated to reflect new files (`internal/auth/devicecode.go`, `internal/auth/store.go`, `internal/cmd/login.go`); module layout, login flow narrative, storage backend selection, and persisted JSON schema all current. `docs_status: updated`.

### Verdict rationale
Every acceptance criterion is provably satisfied by tests + smoke verification. Integration baseline (`make build`, `make test`, `make lint`, `go vet`) is green inside the worktree. ADR 0004 conformance is exact. No blockers; the two Major findings (M1, M2) are operator-recoverable papercuts that don't fail any AC. Per `/auto-review-pr` hard rule 14, unresolved majors do not auto-bump REQUEST_CHANGES — APPROVE is the correct call. The orchestrator can decide whether to spin a fix iteration on M1+M2 before merge or fold them into the next CLI Task.

**Self-review fallback:** GitHub blocks formal `--approve` / `--request-changes` on self-authored PRs (author = reviewer = `damir-topic`). This summary is posted as an issue comment per the documented fallback (same pattern as PRs #156-#174). Result file is the authoritative machine-readable verdict.

cli/internal/auth/store.go
```diff
@@ -0,0 +1,388 @@
...
+	if err != nil {
+		return fmt.Errorf("meho: marshal credentials: %w", err)
+	}
+	tmp := s.path + ".tmp"
+	// O_EXCL prevents clobbering a tmpfile left over from a crashed
+	// previous run — that file might belong to a different process
+	// still mid-write. The cost is one stale-file cleanup on the
+	// operator's side when it happens; the benefit is no silent
+	// data race.
+	f, err := os.OpenFile(tmp, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o600)
```
damir-topic, line 286: **[⚠️ M1]** — Fixed `.tmp` filename can permanently block credential writes after a crash (also flagged by CodeRabbit)

**What:** `tmp := s.path + ".tmp"` followed by `os.OpenFile(tmp, O_WRONLY|O_CREATE|O_EXCL, 0o600)` means a single stale `.tmp` file (from a crashed prior process, a SIGKILLed `meho login`, an `fsck` rescue, or two concurrent `meho login` invocations on the same host) makes every subsequent `Save` fail with `ErrExist`. Recovery requires the operator to manually `rm $XDG_CONFIG_HOME/meho/credentials.json.tmp`.

**Why it matters:** Token storage that silently locks out the operator on a crash is exactly the failure mode the atomic-rename pattern is supposed to prevent. The error message (`stale credentials tmpfile at X — remove and retry`) is a workaround for a design flaw, not a fix. Operators encountering this for the first time will reasonably interpret it as a bug.

**Suggested fix:** Use `os.CreateTemp(filepath.Dir(s.path), "credentials-*.tmp")` — Go's stdlib helper that creates a uniquely-named tmpfile in the target directory with O_EXCL semantics under the hood. Each save gets a fresh tmp name; the atomic-rename invariant survives any number of stale leftovers; the operator-visible failure mode disappears.

```go
f, err := os.CreateTemp(filepath.Dir(s.path), "credentials-*.tmp")
if err != nil {
    return fmt.Errorf("meho: open credentials tmpfile: %w", err)
}
tmp := f.Name()
// (then the same Write/Close/Rename path; on error, os.Remove(tmp) cleanup)
```

**Source:** Go stdlib `os.CreateTemp` — https://pkg.go.dev/os#CreateTemp (creates a new tmpfile with a randomised suffix, opens it `O_RDWR|O_CREATE|O_EXCL`, returns the open `*File`).

cli/internal/cmd/login.go
```diff
@@ -0,0 +1,250 @@
...
+			"Discovery: by default the CLI fetches the backplane's auth-config " +
+			"endpoint at <backplane-url>/api/v1/auth-config to learn the realm " +
+			"issuer and OAuth client ID. Until that endpoint ships, pass " +
+			"--issuer and --client-id explicitly.",
+		Args:         cobra.ExactArgs(1),
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			backplaneURL := strings.TrimRight(args[0], "/")
+			if _, err := url.ParseRequestURI(backplaneURL); err != nil {
+				return fmt.Errorf("invalid backplane URL %q: %w", backplaneURL, err)
```
damir-topic, line 57: **[⚠️ M2]** — Backplane URL validation too permissive; accepts non-HTTP schemes (also flagged by CodeRabbit)

**What:** `url.ParseRequestURI(backplaneURL)` only requires the URI to be absolute. It accepts `mailto:foo@bar`, `file:///etc/passwd`, `tel:`, `urn:`, any scheme — none of which the device-code flow can possibly call. The failure surfaces later as an opaque transport-layer error from the discovery HTTP call (or worse, with a `file:///` scheme, a confusing local-disk read attempt depending on the http.Client transport).

**Why it matters:** The CLI surface is what operators (and `install.sh` scripts in dogfooding consumers) interact with. A precise error at the CLI parse stage (`invalid backplane URL "mailto:x@y": must be an absolute http(s) URL`) is dramatically more useful than `auth-config discovery failed: Get "mailto:x@y/api/v1/auth-config": unsupported protocol scheme "mailto"`. Defence in depth, not just UX: refusing non-HTTP schemes at the entrance reduces the surface area downstream code has to reason about.

**Suggested fix:**

```go
backplaneURL := strings.TrimRight(args[0], "/")
u, err := url.ParseRequestURI(backplaneURL)
if err != nil || u.Host == "" || (u.Scheme != "http" && u.Scheme != "https") {
    return fmt.Errorf("invalid backplane URL %q: must be an absolute http(s) URL", backplaneURL)
}
```

The `u.Host == ""` check rejects `http:foo` (which `ParseRequestURI` accepts as `scheme=http, opaque=foo`); the scheme check rejects everything else.

**Source:** `net/url` package docs — `url.ParseRequestURI` accepts opaque-non-host URIs (`https://pkg.go.dev/net/url#ParseRequestURI`); the documented idiom for HTTP(S)-only validation is to check `Scheme` + `Host` explicitly after parsing (`gh` CLI: cmd/gh/factory/remote_resolver.go does the same check pattern).

Issue #251, opened 2026-05-12
Task: G0.5-T6 End-to-end MCP acceptance test + docs/architecture/mcp.md + docs/cross-repo/mcp-client-setup.md


damir-topic:
Parent initiative: #226

## Target

End-to-end acceptance test against a real MCP client (`@modelcontextprotocol/inspector`) + the two cross-repo docs operators need to wire up Claude Desktop or another MCP client:

- `backend/tests/integration/test_mcp_inspector.py` — new integration test that drives `@modelcontextprotocol/inspector` (or equivalent) against a running backplane, performs the full lifecycle (initialize → tools/list → tools/call meho.status), asserts success.
- `docs/architecture/mcp.md` — new doc. Architecture-side: transport choice, registry pattern, audit integration, spec revision pin.
- `docs/cross-repo/mcp-client-setup.md` — new doc. Operator-side: recipe for wiring Claude Desktop to MEHO, Keycloak OAuth client configuration for MCP, troubleshooting.

## Context

T6 is the **proof-of-life acceptance** for the whole G0.5 Initiative. The previous Tasks shipped the components; T6 proves they work together against a real MCP client, not just against in-process pytest fixtures.

**Why MCP Inspector?** `@modelcontextprotocol/inspector` is the canonical MCP client maintained by the same team that owns the spec. It exercises the wire protocol exactly. If our server works against Inspector, it works against Claude Desktop, Cline, Continue, and other downstream clients without per-client compatibility patches.

**CI test shape.** The integration job:

1. Spins up the backplane (existing CI fixture).
2. Spins up a test Keycloak with the MCP-resource-URI registered as an audience and a test operator with `tenant_admin` role.
3. Obtains a token via the OAuth 2.1 device-code flow (Keycloak supports this for test clients).
4. Runs `npx @modelcontextprotocol/inspector` in non-interactive mode against `https://localhost:8000/mcp` with the token.
5. Asserts: initialize succeeds with protocolVersion 2025-06-18; tools/list returns at least `meho.status`; tools/call meho.status returns the operator's identity bundle; an audit_log row appears.

For local developer iteration: a manual playbook in `docs/architecture/mcp.md` describes how to run the same flow with Claude Desktop pointed at a localhost backplane.

**`docs/architecture/mcp.md` covers:**

1. Why MCP-in-v0.2 (link to v0.2-decisions doc).
2. Transport choice: Streamable HTTP, not stdio (and why).
[...]

Issue #250, opened 2026-05-12
Task: G0.5-T5 Audit integration — tools/call + resources/read write audit_log rows (fail-closed; path exclusion from chassis AuditMiddleware)


damir-topic:
Parent initiative: #226

## Target

Audit integration — every `tools/call` (and optionally `resources/read`) writes an `audit_log` row through the same write path the HTTP API uses:

- `backend/src/meho_backplane/mcp/audit.py` — helper that builds the audit row for an MCP call, including op_id, params hash, result status.
- `backend/src/meho_backplane/mcp/handlers.py` — handlers for `tools/call` and `resources/read` invoke the helper before returning.
- `backend/tests/test_mcp_audit.py` — new tests.

## Context

T5 is what makes MCP-called ops audit-equivalent to HTTP-called ops. Without it, an operator using Claude Desktop to call `meho.status` leaves no audit trail — that's a non-starter for the consumer's compliance-shaped use cases (G8 audit query).

**Same audit_log schema, different write path.** The chassis `AuditMiddleware` writes one row per authenticated HTTP request after it commits. MCP's transport is JSON-RPC over POST `/mcp`, so technically the AuditMiddleware *would* trigger once per `/mcp` POST — but that's wrong granularity. We want one audit row per **tool call** or **resource read**, not per JSON-RPC batch.

Solution: **MCP handlers write their own audit rows.** Skip `AuditMiddleware` for MCP routes (the dispatch helper short-circuits the chassis's middleware by reading + clearing `operator_sub` from contextvars before the AuditMiddleware runs, OR by explicitly excluding `/mcp/*` from the AuditMiddleware path filter). Then each handler writes its own row per operation invocation.

**Audit row fields populated by T5:**

- `operator_sub` — from `operator.sub` (validated by T2).
- `tenant_id` — from `operator.tenant_id` (G0.1).
- `request_id` — bound by the chassis's RequestContextMiddleware (still runs for `/mcp`).
- `method` — `MCP` (literal string distinguishing from HTTP method).
- `path` — `/mcp/tools/call/{tool_name}` or `/mcp/resources/read/{uri}` (synthetic path for query-shape parity).
- `status_code` — derived from result status: `200` for ok, `403` for denied, `500` for internal error, `400` for bad input.
- `duration_ms` — from a `time.monotonic()` bracket around the handler call.
- `payload` — `{op_id, params_hash, op_class}` for tools; `{uri, op_class: "read"}` for resources.

[...]

Issue #249, opened 2026-05-12
Task: G0.5-T4 Reference tool meho.status + reference resource meho://tenant/<id>/info (the templates downstream connectors copy)


damir-topic:
Parent initiative: #226

## Target

The two reference implementations that prove the registry + auth + dispatch chain work end-to-end:

- `backend/src/meho_backplane/mcp/tools/meho_status.py` — `meho.status` MCP tool mirroring `/api/v1/health`.
- `backend/src/meho_backplane/mcp/resources/tenant_info.py` — `meho://tenant/{tenant_id}/info` MCP resource returning the operator's tenant identity bundle.
- `backend/tests/test_mcp_tool_meho_status.py` — new tests for the tool end-to-end.
- `backend/tests/test_mcp_resource_tenant_info.py` — new tests for the resource.

## Context

T4 ships the reference impls that downstream G3-G9 Initiatives will copy as templates. The two implementations cover both ends:
- A **tool** (active, schema'd input/output, RBAC-gated, audit-logged via T5).
- A **resource** (passive read, URI-template-based, returns structured data).

**Why `meho.status` as the reference tool?** It mirrors `/api/v1/health`, the chassis's most-exercised authenticated route. If the OAuth-RS chain (T2) + the tool registry (T3) work for `meho.status`, they work for everyone. Reuses the existing `/api/v1/health` handler logic, just exposed through a different transport.

**Why `meho://tenant/{tenant_id}/info` as the reference resource?** Tenant identity is the simplest universally-useful resource. Every MCP client connecting to MEHO will want to read it once at session start to confirm "yes, I'm operating on tenant rdc-internal." Establishes the `meho://` URI scheme convention.

**Tool schema for `meho.status`:**

```python
ToolDefinition(
    name="meho.status",
    description=(
        "Returns the operator's identity + the backplane's dependency status (Vault, DB). "
        "Useful for verifying that the operator's MCP session can reach all subsystems. "
        "No arguments required."
    ),
    inputSchema={
        "type": "object",
        "properties": {},
        "additionalProperties": False,
    },
    required_role=TenantRole.READ_ONLY,
    op_class="read",
)
```

**Tool description quality is load-bearing.** Per AI engineering best-practices, the tool description IS the agent's prompt for when to use it. Imprecise descriptions get tools called incorrectly. The `meho.status` description above is precise about what the tool does, when to call it, and the no-args interface — copy-paste good for downstream connector tools.

[...]

Issue #248, opened 2026-05-12
Task: G0.5-T3 Tool + Resource registries — register_mcp_tool / register_mcp_resource + tools/list + tools/call dispatch with RBAC filter


damir-topic:
Parent initiative: #226

## Target

Tool registry + Resource registry — the two parallel registries every G3-G9 Initiative will register against:

- `backend/src/meho_backplane/mcp/registry.py` — `register_mcp_tool`, `register_mcp_resource`, dispatch table population.
- `backend/src/meho_backplane/mcp/handlers.py` — `tools/list`, `tools/call`, `resources/list`, `resources/read` handlers that consult the registries.
- `backend/src/meho_backplane/main.py` — lifespan calls `_eager_import_mcp_tools()` so every `mcp/tools/<name>.py` and `mcp/resources/<uri>/<name>.py` module registers at startup.
- `backend/tests/test_mcp_registry.py` — new test file.

## Context

T3 is the parallel of G0.2-T2 (connector registry) but for MCP. The same module-level pattern: `register_mcp_tool("meho.status", schema, handler)` from a module top, the registry stores it, `tools/list` returns it, `tools/call` dispatches to it.

**Two registries, not one.** MCP distinguishes tools (active — `tools/call` invokes them) from resources (passive — `resources/read` fetches data). Tools have input/output schemas + RBAC scopes; resources have URI patterns + content types.

**RBAC + tenant filtering.** When a client calls `tools/list`:
- Filter to tools whose `required_role` ≤ the operator's `tenant_role`.
- Filter to tools that are applicable in the operator's tenant context (most are global; some may be tenant-scoped in future).

Same logic for `resources/list`.

**Tool schema shape (per MCP spec):**

```python
class ToolDefinition(BaseModel):
    name: str  # dotted: meho.status, vault.kv.read, kb.search
    description: str  # agent-facing prompt; load-bearing for agent UX
    inputSchema: dict[str, Any]  # JSON Schema 2020-12
    # MEHO extensions (sent in tools/list under extra fields):
    required_role: TenantRole = TenantRole.OPERATOR
    op_class: str = "read"  # "read" | "write" | "credential_read" | "audit_query" | "other"
```

**Resource URI patterns.** Resources have URI templates per RFC 6570. v0.2 supports:
- `meho://tenant/{tenant_id}/info` — tenant identity (T4 reference).
- `meho://tenant/{tenant_id}/conventions/{slug}` — G7 will register.
[...]

PR #266, opened 2026-05-12
feat(mcp): JSON-RPC 2.0 dispatch + initialize handshake (G0.5-T1)

### LINKED ISSUE #246: Task: G0.5-T1 MCP server entrypoint — /mcp FastAPI route + JSON-RPC 2.0 dispatch + initialize handshake (Streamable HTTP transport)
Parent initiative: #226

## Target

MCP server entrypoint: FastAPI route at `/mcp` that speaks JSON-RPC 2.0 over MCP Streamable HTTP transport, per the [2025-06-18 spec](https://modelcontextprotocol.io/specification/2025-06-18). Handles the `initialize` / `notifications/initialized` lifecycle methods and the dispatcher skeleton for `tools/*` and `resources/*` (the registries land in T3; T1 just dispatches to TODO handlers).

- `backend/src/meho_backplane/mcp/__init__.py` — new package.
- `backend/src/meho_backplane/mcp/server.py` — FastAPI route + JSON-RPC dispatch.
- `backend/src/meho_backplane/mcp/schemas.py` — Pydantic models for JSON-RPC request/response/error + MCP `InitializeRequest`/`InitializeResponse` + capability declarations.
- `backend/src/meho_backplane/main.py` — mount the MCP router.
- `backend/tests/test_mcp_server.py` — new test file. Tests JSON-RPC parsing, initialize handshake, unknown-method error response.

## Context

T1 ships the **transport + dispatch skeleton**; T2 layers OAuth-RS auth onto it; T3 adds the registries; T4 adds the reference tool/resource; T5 wires audit; T6 is acceptance + docs.

**MCP transport: Streamable HTTP, not stdio.** Per the 2025-06-18 spec, hosted MCP servers use HTTP-based transports. Stdio is for local-subprocess servers. MEHO is hosted, so Streamable HTTP it is. The endpoint accepts POST requests with JSON-RPC 2.0 payloads in the body and may respond with either a single response or an `text/event-stream` response if the server needs to push notifications back (e.g., during long-running tool calls or resource subscriptions).

**JSON-RPC 2.0 envelope.** Every request is `{jsonrpc: "2.0", id: <int|str|null>, method: <str>, params: <obj>}`. Notifications (no response expected) have no `id`. Server response is `{jsonrpc: "2.0", id: <same>, result: <obj>}` or `{jsonrpc: "2.0", id: <same>, error: {code, message, data?}}`.

**Initialize handshake.** First method called by every client. Server returns server name + version + capabilities + (optionally) `instructions` (G7.1 will populate this with the assembled session preamble). Capabilities advertise which feature sets we support: `tools`, `resources`, and v0.2 may advertise `logging` (server-emit-log notification) — `prompts` / `roots` / `sampling` are explicitly out of v0.2 scope.

**Dispatch table.** T1 ships a method → handler dispatch:
- `initialize` → handle here in T1.
- `notifications/initialized` → handle (acknowledge; no response).
- `tools/list` → stub returning empty list (T3 fills in).
- `tools/call` → stub returning method-not-found (T3 fills in).
- `resources/list` → stub (T3).
- `resources/read` → stub (T3).
- `ping` → respond pong (basic liveness).
- Unknown method → standard JSON-RPC error code -32601.

**Auth in T1.** No auth yet. Every request currently succeeds. T2 adds the auth chain.

## Current state

After v0.1 chassis:
- No `mcp/` module. No JSON-RPC handling. No `/mcp` route.
- The existing JWT validation chain in `auth/jwt.py` is reusable but not wired to MCP.

## Desired state

`backend/src/meho_backplane/mcp/schemas.py`:

```python

from typing import Any, Literal
from pydantic import BaseModel, Field

JsonRpcId = int | str | None


class JsonRpcRequest(BaseModel):
    jsonrpc: Literal["2.0"]
    id: JsonRpcId = None
    method: str
    params: dict[str, Any] | None = None


class JsonRpcError(BaseModel):
    code: int
    message: str
    data: dict[str, Any] | None = None


class JsonRpcResponse(BaseModel):
    jsonrpc: Literal["2.0"] = "2.0"
    id: JsonRpcId
    result: dict[str, Any] | None = None
    error: JsonRpcError | None = None


class InitializeRequest(BaseModel):
    protocolVersion: str
    capabilities: dict[str, Any] = Field(default_factory=dict)
    clientInfo: dict[str, str] = Field(default_factory=dict)


class ServerCapabilities(BaseModel):
    tools: dict[str, Any] | None = None
    resources: dict[str, Any] | None = None
    logging: dict[str, Any] | None = None


class InitializeResponse(BaseModel):
    protocolVersion: Literal["2025-06-18"] = "2025-06-18"
    capabilities: ServerCapabilities
    serverInfo: dict[str, str]
    instructions: str | None = None  # T4/G7.1 populates with session preamble


# Standard JSON-RPC error codes
PARSE_ERROR = -32700
INVALID_REQUEST = -32600
METHOD_NOT_FOUND = -32601
INVALID_PARAMS = -32602
INTERNAL_ERROR = -32603
```

`backend/src/meho_backplane/mcp/server.py` (sketch):

```python

from typing import Any, Callable, Awaitable
import structlog
from fastapi import APIRouter, Request, Response
from meho_backplane.mcp.schemas import (
    JsonRpcRequest, JsonRpcResponse, JsonRpcError,
    InitializeRequest, InitializeResponse, ServerCapabilities,
    METHOD_NOT_FOUND, INVALID_PARAMS, PARSE_ERROR,
)
from meho_backplane import __version__

router = APIRouter(prefix="/mcp", tags=["mcp"])

_DISPATCH: dict[str, Callable[..., Awaitable[Any]]] = {}


def register_method(name: str, handler: Callable[..., Awaitable[Any]]) -> None:
    if name in _DISPATCH:
        raise RuntimeError(f"MCP method {name!r} already registered")
    _DISPATCH[name] = handler


async def _initialize(req: InitializeRequest) -> InitializeResponse:
    return InitializeResponse(
        capabilities=ServerCapabilities(
            tools={"listChanged": False},
            resources={"listChanged": False, "subscribe": False},
        ),
        serverInfo={"name": "meho-backplane", "version": __version__},
        # instructions populated by G7.1 once tenant-aware:
        # instructions=await assemble_preamble(operator.tenant_id),
    )


register_method("initialize", lambda params: _initialize(InitializeRequest.model_validate(params)))


@router.post("")
async def mcp_dispatch(request: Request) -> Response:
    try:
        body = await request.json()
        jrpc = JsonRpcRequest.model_validate(body)
    except Exception as e:
        return _err_response(None, PARSE_ERROR, f"parse error: {e}")

    handler = _DISPATCH.get(jrpc.method)
    if handler is None:
        return _err_response(jrpc.id, METHOD_NOT_FOUND, f"method not found: {jrpc.method}")

    try:
        result = await handler(jrpc.params or {})
        if jrpc.id is None:  # notification — no response
            return Response(status_code=204)
        return JsonRpcResponse(id=jrpc.id, result=result.model_dump() if hasattr(result, "model_dump") else result)
    except Exception as e:
        logger.exception("mcp_handler_error", method=jrpc.method)
        return _err_response(jrpc.id, INTERNAL_ERROR, str(e))
```

`main.py` mounts the router:

```python
from meho_backplane.mcp.server import router as mcp_router
app.include_router(mcp_router)
```

## Acceptance criteria

- [ ] `/mcp` route accepts JSON-RPC 2.0 POST requests.
- [ ] `initialize` returns server name/version + capabilities + protocolVersion=2025-06-18.
- [ ] `notifications/initialized` accepted; returns 204 no content.
- [ ] Unknown method returns JSON-RPC error code -32601.
- [ ] Parse error returns -32700.
- [ ] Invalid params (e.g. initialize without `protocolVersion`) returns -32602.
- [ ] `ping` works: returns `{}` as result.
- [ ] T1 has NO auth yet — every request currently succeeds (T2 adds auth). Documented in code comment.

## Out of scope

- Auth (T2).
- Tool registry + tools/list, tools/call (T3, T4).
- Resource registry + resources/list, resources/read (T3, T4).
- Audit integration (T5).
- MCP Inspector acceptance test (T6).
- Streamable-HTTP-with-SSE response shape for long-running tools — v0.2 only supports single-response shape.
- Server-initiated notifications (e.g. listChanged) — v0.2 advertises `listChanged: false`.

## References

- Parent Initiative: [#226](https://github.com/evoila/meho/issues/226).
- MCP spec 2025-06-18 — Lifecycle: https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle
- MCP spec 2025-06-18 — Transport: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports
- JSON-RPC 2.0 spec: https://www.jsonrpc.org/specification
- Existing router pattern: [backend/src/meho_backplane/api/v1/health.py](backend/src/meho_backplane/api/v1/health.py).
- Python best-practices: `.claude/skills/explore-and-refactor/python_best_practices.md` — FastAPI dispatching, Pydantic validation, structured-error responses.

zdamir:
## Summary

- Stands up the MCP server entrypoint at `POST /mcp` per **MCP spec revision 2025-06-18** — JSON-RPC 2.0 dispatch + Streamable HTTP transport (single-response shape; SSE is out of scope for v0.2).
- New `backend/src/meho_backplane/mcp/` module with `schemas.py` (frozen Pydantic v2 envelopes for JSON-RPC + MCP lifecycle payloads) and `server.py` (FastAPI router + `register_method` dispatch-table API that mirrors `register_probe`, so #248 can layer the tool / resource registries on top).
- Built-in handlers for the lifecycle primitives the spec defines as universal — `initialize`, `notifications/initialized`, `ping`. Unknown methods on requests return JSON-RPC `-32601`; on notifications they're silently 202-accepted per JSON-RPC §4.1.2.
- No Bearer-token auth on this route yet — T2 (#247) layers the OAuth-RS chain on top. Origin-header validation per the MCP transport DNS-rebinding warning is also deferred to T2 (it depends on the `MCP_ALLOWED_ORIGINS` setting T2 introduces).

## Deviation from the issue's AC #3

The AC list on #246 says `notifications/initialized` "returns **204** no content". The MCP Streamable HTTP spec ("Sending Messages to the Server") requires HTTP **202 Accepted** for notifications, not 204. The AC's *intent* ("acknowledge with no body") is met either way; the wire code follows the spec MUST. Tests assert 202.

## Test plan

- [x] `ruff format --check` — 45 files formatted
- [x] `mypy src/` (strict mode) — 27 source files, no issues
- [x] `pytest tests/` — **246 passed**, 1 deselected (the pre-existing `process_resident_memory_bytes` test that fails on macOS too — also fails on `main`, this PR doesn't touch it)
- [x] New `tests/test_mcp_server.py` — **18 cases**, one per acceptance criterion plus the JSON-RPC edge paths (parse error, batch arrays, missing method, wrong jsonrpc version, internal error path, GET → 405, duplicate-registration guard, unauthenticated success). Walks every JSON-RPC error code in the dispatch.

### AC coverage

- [x] `/mcp` accepts JSON-RPC 2.0 POSTs
- [x] `initialize` returns `serverInfo` (name/version) + capabilities + `protocolVersion=2025-06-18`
- [x] `notifications/initialized` accepted with no body (202 per spec; see deviation note above)
- [x] Unknown method returns `-32601`
- [x] Parse error returns `-32700` with `id: null`
- [x] Invalid params (`initialize` without `protocolVersion`) returns `-32602`
- [x] `ping` returns `result: {}`
- [x] T1 has **no auth** — every well-formed request currently succeeds (docstrings + dedicated test)
- [x] CI green: ruff + mypy + pytest

## Out of scope

- **OAuth-RS auth** — G0.5-T2 (#247): `/.well-known/oauth-protected-resource`, `WWW-Authenticate` headers, audience validation per RFC 8707, reuse of `verify_jwt`.
- **Origin-header validation** (transport-MUST per MCP spec, deferred because it requires `MCP_ALLOWED_ORIGINS` from T2).
- **Tool + resource registries + `tools/list` / `tools/call` / `resources/list` / `resources/read`** — G0.5-T3 (#248).
- **Reference tool `meho.status` + reference resource `meho://tenant/<id>/info`** — G0.5-T4 (#249).
- **MCP-specific audit integration** (path-exclusion from chassis `AuditMiddleware`; fail-closed audit row on `tools/call`/`resources/read`) — G0.5-T5 (#250).
- **End-to-end MCP Inspector acceptance test** + `docs/architecture/mcp.md` — G0.5-T6 (#251).
- **SSE response shape** (`Content-Type: text/event-stream` for long-running tools) — out of v0.2 scope entirely.
- **Batch JSON-RPC arrays** — spec §6 allows them at the JSON-RPC layer but the MCP Streamable HTTP transport accepts only single envelopes per POST; the dispatcher rejects arrays with `-32600`.

Closes #246

## Summary by CodeRabbit

* **New Features**
  * Added a /mcp Streamable HTTP JSON-RPC transport with protocol-version handling and built-in lifecycle handlers (initialize, notifications/initialized, ping). Requests return JSON-RPC envelopes (200); notifications return HTTP 202 Accepted.

* **Behavior / Error Handling**
  * Strict envelope validation; boolean ids rejected; initialize exempt from protocol-header enforcement; unsupported versions return JSON-RPC INVALID_REQUEST; non-object or parse failures and handler errors map to appropriate JSON-RPC errors.

* **Tests**
  * Expanded acceptance tests covering header rules, id validation, notification vs request semantics, handler-return and error-path conversions.

* **Documentation**
  * New docs describing the MCP transport, JSON-RPC schemas, and handler registration API.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/266)

---

## Follow-up (auto-implement-issue, iteration 2, 2026-05-12)

Addressed review feedback from the three CHANGES_REQUESTED / COMMENTED reviews on iteration 1 (CodeRabbit, ilhan-karic, /review-pr-slim).

**Fixed:**

- **B1** `a92aa59` — Stop advertising `tools` / `resources` capabilities ahead of T3. `ServerCapabilities()` now ships every field `None`; the wire serializer drops them. Prevents spec-conforming clients from disconnecting when their follow-up `tools/list` call hits `-32601`.
- **M1** `094019f` — Reject `bool` ids in JSON-RPC envelopes. `bool` subclasses `int` in Python, so `True` / `False` were coercing to `1` / `0` on the wire — a JSON-RPC §4.1.2 wire-contract violation. Added a `mode="before"` `field_validator` on `JsonRpcRequest.id` and a `isinstance(raw, bool)` short-circuit in `_coerce_request_id`. Parametrized test covers both gates.
- **M2** `a10b0c4` — `MCP-Protocol-Version` header now validated per spec §"Protocol Version Header" MUST. If header present and unsupported → HTTP 400 + JSON-RPC error envelope. `initialize` exempted (clients don't know the version pre-handshake). Header absent is accepted as transitional lenience (spec's SHOULD-assume-2025-03-26 doesn't help — v0.2 doesn't support that revision). Four tests pin the contract.
- **Mi1 + Mi7** `ea2595f` — `JsonRpcId` docstring corrected to describe the dispatcher's actual absent/null distinction. Comment on `InitializeRequest.model_validate(params or {})` explaining the deliberate `None` / `{}` collapse.
- **Mi4** `433e8f9` — `_McpInvalidParamsError` → `McpInvalidParamsError`; re-exported from `meho_backplane.mcp` so T3 (#248) tool handlers can raise it without reaching into a dunder-private symbol.
- **Mi2 + Mi8** `24c210e` — Added `test_handler_returning_non_dict_scalar_becomes_internal_error` to cover the load-bearing handler-bug guard (success-with-wrong-shape arm). Annotated `test_register_method_rejects_duplicate_registration` to note it raises before mutating, so no fixture cleanup is needed.

**Disputed (with inline reply):**

- **Q1** `adb6440` — Notifications return HTTP 202 even when the handler errors (unknown method, invalid params, internal error). Spec §"Sending Messages to the Server" splits "accepts → 202" from "cannot accept → 4xx", but "cannot accept" is intentionally narrow — it refers to transport-level rejection (malformed envelope, batch arrays, bad JSON, unsupported `MCP-Protocol-Version`). Transport-level rejections already flip to HTTP 4xx; post-envelope arms stay at 202 because (1) JSON-RPC §4.1.2 says the server MUST NOT reply to a notification, (2) clients cannot retry on 4xx without violating §4.1.2, and (3) most MCP servers in the wild treat unknown notifications as silently-dropped for forward-compat. Comment in `mcp_dispatch` now spells this out. See inline reply on the original comment.

- **Mi6** (ilhan-karic) — `_serialize_response`'s hand-rolled "drop the unset half" logic vs `model_dump(exclude_none=True)` + manual id override. **Keeping the hand-roll.** Pydantic's `exclude_none=True` would also strip the `id` field on parse-error responses, where spec §5 mandates the id be **null** and **present**. The hand-roll makes that load-bearing distinction grep-able and pins the spec citation right where the serializer lives; the alternative is a model_dump + post-process that buries the rationale one indirection away. The reviewer framed this as "Fine, but..." — not blocking.

**Deferred to follow-up (recorded for orchestrator):**

- **Mi3** — Unbounded `request.body()` read; DOS risk. Belongs with the rest of `/mcp` transport hardening (Origin-header validation, content-type 415 enforcement, body-size cap), all gated on the `MCP_ALLOWED_ORIGINS` / size-cap settings T2 (#247) introduces. Both reviewers agreed this is T2 scope.
- **Mi5** (ilhan-karic) — Strict content-type check (415 on non-`application/json`). Spec-aligned at the moment (the spec doesn't forbid lenience); a hardening item that fits naturally with Mi3's transport-hardening cluster. Will land alongside Mi3 in T2.
- **CR1** (CodeRabbit) — `mcp_dispatch` cognitive complexity 18 vs SonarCloud threshold 15. CodeRabbit's own note framed this as "purely optional cleanup". The function is single-responsibility (envelope → dispatch → wire-shape mapping) and splitting into three helpers would fragment that cohesion. Will revisit if T3's registry-aware dispatch pushes complexity higher.

**Verification (re-run after fixes):**

- `ruff check src/ tests/` — clean
- `ruff format --check` — 52 files formatted
- `mypy src/` (strict) — 27 source files, no issues
- `pytest tests/` — **253 passed**, 1 deselected (pre-existing macOS-only test, still fails on `main`)
- `tests/test_mcp_server.py` — **25 cases** (up from 18): 7 new tests cover bool-id rejection (×2 parametrized), protocol-version header validation (×4), and the missing handler-bug-guard arm (×1)

---

## Follow-up (auto-implement-issue, iteration 3, 2026-05-12)

CodeRabbit's iteration-2 review flagged 3 new actionable items, all direct fallout from iteration-2's M2 protocol-version fix:

**Fixed:**

- **CR2 + CR3** `a9d9931` — Two docstring claims went stale when M2 introduced the HTTP 400 path for unsupported `MCP-Protocol-Version`. `mcp/__init__.py:29` said "Every well-formed JSON-RPC request currently succeeds" — no longer true. `mcp/server.py` module docstring said "transport-level failures (in T1: none) flip the HTTP status" — M2 added exactly that case. Both rewritten to enumerate the actual T1 transport-failure surface and point at T2 for the OAuth 401/403 cases.

- **CR4** `8262277` — `mcp_dispatch` cognitive complexity 22 vs SonarCloud threshold 15 (iteration-1 was 18; M2 pushed it higher). Refactored into four focused helpers, each with its load-bearing spec citation:

  - `_parse_request_body(raw_body)` — bytes → envelope dict or transport error.
  - `_validate_protocol_version_header(request, method, payload)` — spec §"Protocol Version Header" enforcement.
  - `_build_success_response(request_id, result)` — JSON-RPC envelope serialization.
  - `_dispatch_to_handler(jrpc, is_notification)` — handler lookup, execution, error mapping. The Q1 "cannot accept" comment lives here now since the notification/202 split is local to handler execution.

  `mcp_dispatch` becomes a thin orchestrator. Behavior preserved end-to-end — same 25 test cases pass unchanged.

**Verification (re-run after iter-3 fixes):**

- `ruff check src/ tests/` — clean
- `ruff format --check` — 52 files formatted
- `mypy src/` (strict) — 27 source files, no issues
- `pytest tests/` — **253 passed**, 1 deselected (pre-existing macOS test)
- SonarCloud quality gate — passed on iter-2 (was already passing); iter-3 should bring `mcp_dispatch`'s complexity below the threshold

### DIFF
--- a/backend/src/meho_backplane/mcp/schemas.py
+++ b/backend/src/meho_backplane/mcp/schemas.py
@@ -166,15 +166,18 @@ class InitializeRequest(BaseModel):
 class ServerCapabilities(BaseModel):
     """Server-side capability declaration returned on ``initialize``.
 
-    T1 advertises ``tools`` and ``resources`` envelopes (with
-    ``listChanged: false``) so the negotiated capability namespace is
-    already reachable when T3 (#248) wires the registries. ``prompts``
-    / ``roots`` / ``sampling`` / ``completions`` are out of scope for
-    v0.2 and stay absent — clients that probe for them must treat the
-    omission as "not supported" per spec §Capability Negotiation.
-
-    ``logging`` is reserved for a future server-emit-log capability;
-    leaving it ``None`` here makes the upgrade additive.
+    T1 advertises an **empty** capabilities envelope — ``tools`` /
+    ``resources`` / ``logging`` all stay :data:`None`, which the
+    hand-rolled wire serializer drops. Advertising a capability ahead
+    of the dispatch table being able to honor it would tell a
+    spec-conforming client it can call ``tools/list`` / ``resources/list``
+    immediately after the handshake, which would then ``-32601`` and
+    likely cause the client to disconnect per spec §Capability
+    Negotiation ("Only use capabilities that were successfully
+    negotiated"). T3 (#248) is where ``tools`` and ``resources`` flip
+    back on, paired with the dispatch-table additions. ``prompts`` /
+    ``roots`` / ``sampling`` / ``completions`` are out of scope for
+    v0.2 and stay absent.
     """
 
     model_config = ConfigDict(frozen=True)
--- a/backend/src/meho_backplane/mcp/schemas.py
+++ b/backend/src/meho_backplane/mcp/schemas.py
@@ -1,0 +1,221 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""JSON-RPC 2.0 + MCP 2025-06-18 wire-shape models.
+
+Pydantic v2 envelopes for the MCP server entrypoint (G0.5-T1). T1 covers
+only the lifecycle surface: ``initialize`` / ``ping`` /
+``notifications/initialized``. The tool + resource registries (T3, #248)
+and the OAuth-RS metadata document (T2, #247) layer their own request /
+response models on top of these JSON-RPC envelopes.
+
+The wire forms here follow two specs that the MCP 2025-06-18 revision
+chains together:
+
+* JSON-RPC 2.0 — https://www.jsonrpc.org/specification — the envelope,
+  the request / response / error shapes, and the notification rule
+  (request without an ``id`` member; server MUST NOT reply).
+* MCP 2025-06-18 lifecycle —
+  https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle
+  — the ``initialize`` request / result and ``serverInfo`` /
+  ``capabilities`` payloads.
+
+Field names that the wire defines as camelCase (``protocolVersion``,
+``clientInfo``, ``serverInfo``) keep that shape on the Python side too;
+the ``# noqa: N815`` markers exist because pep8-naming would otherwise
+push them to snake_case, breaking interop with every MCP client in the
+wild.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Literal
+
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+
+__all__ = [
+    "INTERNAL_ERROR",
+    "INVALID_PARAMS",
+    "INVALID_REQUEST",
+    "METHOD_NOT_FOUND",
+    "PARSE_ERROR",
+    "PROTOCOL_VERSION",
+    "InitializeRequest",
+    "InitializeResponse",
+    "JsonRpcError",
+    "JsonRpcId",
+    "JsonRpcRequest",
+    "JsonRpcResponse",
+    "ServerCapabilities",
+]
+
+
+#: The MCP spec revision this server implements. Returned verbatim in
+#: ``InitializeResponse.protocolVersion`` when the client requests this
+#: version; otherwise the server responds with the same value anyway and
+#: the client decides whether to disconnect (spec §Version Negotiation).
+PROTOCOL_VERSION: str = "2025-06-18"
+
+
+#: A JSON-RPC 2.0 id is "a String, Number, or NULL value if included"
+#: (spec §4.1). A request *without* an id is a notification (§4.1.2);
+#: the absent / explicit-null distinction is normatively discouraged on
+#: the request side, so this server treats both as the same shape.
+JsonRpcId = int | str | None
+
+
+# ---------------------------------------------------------------------------
+# JSON-RPC 2.0 envelopes
+# ---------------------------------------------------------------------------
+
+
+class JsonRpcRequest(BaseModel):
+    """Inbound JSON-RPC 2.0 request envelope.
+
+    Notifications carry no ``id`` (spec §4.1.2). Because Pydantic v2 cannot
+    distinguish "field absent" from "field present and ``null``" once a
+    default is supplied, the dispatcher consults the raw ``payload`` dict
+    for the ``"id"`` key when deciding whether to emit a response — see
+    :func:`~meho_backplane.mcp.server.mcp_dispatch`.
+
+    ``extra="allow"`` lets the model accept MCP-spec extensions that
+    revision 2025-06-18 doesn't itself surface (the spec is intentionally
+    forward-extensible at the envelope layer); the dispatcher only reads
+    the four canonical fields.
+    """
+
+    model_config = ConfigDict(frozen=True, extra="allow")
+
+    jsonrpc: Literal["2.0"]
+    id: JsonRpcId = None
+    method: str
+    params: dict[str, Any] | None = None
+
+
+class JsonRpcError(BaseModel):
+    """JSON-RPC 2.0 error object (the ``response.error`` payload).
+
+    ``data`` is optional per spec §5.1 and is omitted on the wire when
+    unset; callers that need to attach structured context (e.g. an
+    "unsupported protocol version" body with ``supported`` + ``requested``
+    keys) pass it through here.
+    """
+
+    model_config = ConfigDict(frozen=True)
+
+    code: int
+    message: str
+    data: dict[str, Any] | None = None
+
+
+class JsonRpcResponse(BaseModel):
+    """Outbound JSON-RPC 2.0 response envelope.
+
+    Spec §5: "Either the result member or error member MUST be included,
+    but both members MUST NOT be included." The
+    :meth:`_exactly_one_of_result_or_error` validator enforces this
+    invariant on construction so a misbehaving handler can't emit a
+    wire-broken response that some clients accept and others reject.
+
+    Serialization to the wire goes through
+    :func:`~meho_backplane.mcp.server._serialize_response`, which drops
+    the unset half of ``result`` / ``error`` (Pydantic's ``exclude_none``
+    would also strip ``id`` when it's the spec-mandated ``null`` on a
+    parse-error response — see spec §5).
+    """
+
+    model_config = ConfigDict(frozen=True)
+
+    jsonrpc: Literal["2.0"] = "2.0"
+    id: JsonRpcId
+    result: dict[str, Any] | None = None
+    error: JsonRpcError | None = None
+
+    @model_validator(mode="after")
+    def _exactly_one_of_result_or_error(self) -> JsonRpcResponse:
+        if (self.result is None) == (self.error is None):
+            raise ValueError(
+                "JsonRpcResponse must have exactly one of `result` or `error`",
+            )
+        return self
+
+
+# ---------------------------------------------------------------------------
+# MCP 2025-06-18 lifecycle payloads
+# ---------------------------------------------------------------------------
+
+
+class InitializeRequest(BaseModel):
+    """``initialize`` request params per MCP 2025-06-18 §Initialization.
+
+    The spec example body carries ``protocolVersion``, ``capabilities``,
+    and ``clientInfo`` as required keys; defensively, ``capabilities``
+    and ``clientInfo`` are tolerated as missing (default to empty dict)
+    because some client SDKs omit empty objects. Missing
+    ``protocolVersion`` is the only required-field shape that hits
+    the ``INVALID_PARAMS`` path.
+    """
+
+    model_config = ConfigDict(frozen=True, extra="allow")
+
+    protocolVersion: str  # noqa: N815 — wire field is camelCase per MCP spec
+    capabilities: dict[str, Any] = Field(default_factory=dict)
+    clientInfo: dict[str, Any] = Field(default_factory=dict)  # noqa: N815
+
+
+class ServerCapabilities(BaseModel):
+    """Server-side capability declaration returned on ``initialize``.
+
+    T1 advertises ``tools`` and ``resources`` envelopes (with
+    ``listChanged: false``) so the negotiated capability namespace is
+    already reachable when T3 (#248) wires the registries. ``prompts``
+    / ``roots`` / ``sampling`` / ``completions`` are out of scope for
+    v0.2 and stay absent — clients that probe for them must treat the
+    omission as "not supported" per spec §Capability Negotiation.
+
+    ``logging`` is reserved for a future server-emit-log capability;
+    leaving it ``None`` here makes the upgrade additive.
+    """
+
+    model_config = ConfigDict(frozen=True)
+
+    tools: dict[str, Any] | None = None
+    resources: dict[str, Any] | None = None
+    logging: dict[str, Any] | None = None
+
+
+class InitializeResponse(BaseModel):
+    """``initialize`` result per MCP 2025-06-18 §Initialization.
+
+    ``instructions`` stays ``None`` in T1; G7.1 will populate it with
+    the assembled tenant-aware session preamble once tenancy lands.
+    Until then the field is serialized as omitted (the wire view drops
+    ``None``-valued ``instructions``).
+    """
+
+    model_config = ConfigDict(frozen=True)
+
+    protocolVersion: str = PROTOCOL_VERSION  # noqa: N815
+    capabilities: ServerCapabilities
+    serverInfo: dict[str, str]  # noqa: N815
+    instructions: str | None = None
+
+
+# ---------------------------------------------------------------------------
+# Standard JSON-RPC error codes (spec §5.1)
+# ---------------------------------------------------------------------------
+
+#: "Invalid JSON was received by the server." (spec §5.1)
+PARSE_ERROR: int = -32700
+
+#: "The JSON sent is not a valid Request object." (spec §5.1)
+INVALID_REQUEST: int = -32600
+
+#: "The method does not exist / is not available." (spec §5.1)
+METHOD_NOT_FOUND: int = -32601
+
+#: "Invalid method parameter(s)." (spec §5.1)
+INVALID_PARAMS: int = -32602
+
+#: "Internal JSON-RPC error." (spec §5.1)
+INTERNAL_ERROR: int = -32603
--- a/backend/src/meho_backplane/mcp/schemas.py
+++ b/backend/src/meho_backplane/mcp/schemas.py
@@ -31,7 +31,7 @@ from __future__ import annotations
 
 from typing import Any, Literal
 
-from pydantic import BaseModel, ConfigDict, Field, model_validator
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 
 __all__ = [
     "INTERNAL_ERROR",
@@ -90,6 +90,27 @@ class JsonRpcRequest(BaseModel):
     id: JsonRpcId = None
     method: str
     params: dict[str, Any] | None = None
+
+    @field_validator("id", mode="before")
+    @classmethod
+    def _reject_bool_id(cls, value: Any) -> Any:
+        """Reject ``bool`` ids before Pydantic coerces them to int.
+
+        ``bool`` is a subclass of ``int`` in Python, so the ``int | str |
+        None`` annotation accepts ``True`` / ``False`` and silently
+        coerces them to ``1`` / ``0``. JSON-RPC 2.0 §4.1.2 restricts the
+        id to "a String, Number, or NULL value if included" — and the
+        wire contract on the response side requires the response id to
+        equal the request id. Echoing ``1`` when the client sent ``true``
+        breaks that contract. The ``mode="before"`` hook runs ahead of
+        type coercion so the ValidationError surfaces ``True`` /
+        ``False`` rather than ``1`` / ``0`` in the error context.
+        """
+        if isinstance(value, bool):
+            raise ValueError(
+                "JSON-RPC id must be String, Number, or NULL — not bool",
+            )
+        return value
 
 
 class JsonRpcError(BaseModel):
--- a/backend/src/meho_backplane/mcp/schemas.py
+++ b/backend/src/meho_backplane/mcp/schemas.py
@@ -59,8 +59,14 @@ PROTOCOL_VERSION: str = "2025-06-18"
 
 #: A JSON-RPC 2.0 id is "a String, Number, or NULL value if included"
 #: (spec §4.1). A request *without* an id is a notification (§4.1.2);
-#: the absent / explicit-null distinction is normatively discouraged on
-#: the request side, so this server treats both as the same shape.
+#: the dispatcher honors the absent / explicit-null distinction by
+#: inspecting the raw payload dict (``"id" not in payload``) — absent
+#: ids fall into the notification path (HTTP 202, no response body),
+#: while ``"id": null`` is treated as a discouraged-but-valid request
+#: with a null id (the response echoes ``id: null`` per spec §5). On
+#: the response side ``bool`` is rejected explicitly because ``bool``
+#: subclasses ``int``; see
+#: :meth:`JsonRpcRequest._reject_bool_id` for the validator.
 JsonRpcId = int | str | None
 
 
--- a/backend/src/meho_backplane/mcp/server.py
+++ b/backend/src/meho_backplane/mcp/server.py
@@ -40,8 +40,12 @@ Per the Streamable HTTP §"Sending Messages to the Server":
 
 A JSON-RPC-level error (parse error, invalid request, method not found,
 invalid params, internal error) is encoded as a 200 envelope with the
-``error`` member populated; only transport-level failures (in T1: none)
-flip the HTTP status away from 200 / 202.
+``error`` member populated. **Transport-level failures** flip the HTTP
+status away from 200 / 202; T1 has one such case: an unsupported
+``MCP-Protocol-Version`` header on a non-``initialize`` call returns
+HTTP 400 with a JSON-RPC error envelope in the body (spec §"Protocol
+Version Header" MUST). T2 (#247) adds the OAuth 401 / 403 cases when
+the Bearer chain lands.
 
 The single-shot JSON shape is also what the AC list on #246 codifies:
 no streaming, no chunked responses, no SSE. The transport spec's GET
--- a/backend/src/meho_backplane/mcp/server.py
+++ b/backend/src/meho_backplane/mcp/server.py
@@ -280,8 +280,16 @@ def _coerce_request_id(payload: dict[str, Any]) -> JsonRpcId:
     parseable (helps clients correlate failures). When the raw value
     isn't a String / Number / NULL, fall back to ``None`` per spec §5
     ("If there was an error in detecting the id ... it MUST be Null").
+
+    ``bool`` is short-circuited explicitly because it subclasses ``int``
+    in Python — without this check ``True`` / ``False`` would slip
+    through the ``isinstance(raw, (int, str))`` arm and echo as ``1`` /
+    ``0`` in the error response, which the client cannot correlate
+    against its original ``true`` / ``false`` request id.
     """
     raw = payload.get("id")
+    if isinstance(raw, bool):
+        return None
     if isinstance(raw, (int, str)) or raw is None:
         return raw
     return None
--- a/backend/src/meho_backplane/mcp/server.py
+++ b/backend/src/meho_backplane/mcp/server.py
@@ -323,110 +323,136 @@ def _coerce_request_id(payload: dict[str, Any]) -> JsonRpcId:
     return None
 
 
-@router.post("")
-async def mcp_dispatch(request: Request) -> Response:
-    """Dispatch a single JSON-RPC 2.0 request or notification.
+def _parse_request_body(raw_body: bytes) -> dict[str, Any] | JSONResponse:
+    """Decode ``raw_body`` to a JSON-RPC envelope dict or an HTTP error response.
 
-    The Streamable HTTP body contract (§"Sending Messages to the Server"):
+    Three rejection arms map onto the spec-prescribed error codes:
 
-    * Input is a JSON object — batch arrays are unsupported in T1.
-    * On a *request* (id present): return 200 + single JSON envelope.
-    * On a *notification* (id absent, or method prefix
-      ``notifications/``): return 202 with no body, regardless of
-      whether the handler errored — JSON-RPC §4.1.2 forbids replying.
+    * Empty body → PARSE_ERROR (clearer message than ``json.loads(b"")``'s
+      "Expecting value").
+    * :class:`json.JSONDecodeError` → PARSE_ERROR with the parser's message.
+    * Non-dict payload (array or scalar) → INVALID_REQUEST. JSON-RPC §6
+      allows batch arrays at the protocol layer but the MCP Streamable
+      HTTP transport mandates a single envelope per POST.
 
-    GET requests on this path automatically return HTTP 405 (FastAPI's
-    default for an unmatched method) which satisfies the spec's
-    fallback when the server does not implement the GET-opens-SSE
-    branch of the Streamable HTTP transport.
+    On success returns the parsed ``dict``; on failure returns the
+    :class:`JSONResponse` that the dispatcher should hand back to the
+    client. The union return is the idiomatic "or-error-response" shape
+    that lets the caller narrow with :func:`isinstance`.
     """
-    raw_body = await request.body()
     if not raw_body:
         return _error_response(None, PARSE_ERROR, "parse error: empty body")
-
     try:
         payload = json.loads(raw_body)
     except json.JSONDecodeError as exc:
         _log.warning("mcp_parse_error", error=exc.msg)
         return _error_response(None, PARSE_ERROR, f"parse error: {exc.msg}")
-
     if not isinstance(payload, dict):
-        # Batch (array) bodies are spec-allowed at the JSON-RPC layer
-        # (§6) but the MCP Streamable HTTP transport accepts only a
-        # single request / notification / response per POST. Reject
-        # arrays so the contract is unambiguous.
         return _error_response(
             None,
             INVALID_REQUEST,
             "invalid request: expected JSON object, not array or scalar",
         )
+    return payload
 
-    try:
-        jrpc = JsonRpcRequest.model_validate(payload)
-    except ValidationError as exc:
+
+def _validate_protocol_version_header(
+    request: Request,
+    method: str,
+    payload: dict[str, Any],
+) -> JSONResponse | None:
+    """Enforce MCP 2025-06-18 §"Protocol Version Header" on non-initialize calls.
+
+    Per spec, clients MUST send ``MCP-Protocol-Version`` on every post-
+    ``initialize`` request, and the server MUST respond with HTTP 400
+    when the value is invalid or unsupported. ``initialize`` is exempted
+    because clients don't know which version to send until the handshake
+    completes. Absent header on a non-initialize call is accepted as
+    transitional lenience: spec SHOULD-assume-2025-03-26 doesn't help —
+    v0.2 doesn't support that revision — and tightening this in T1
+    would break clients that don't yet emit the header. T6 (#251)
+    acceptance tests will pin the strict-mode contract.
+
+    Returns ``None`` on the OK path; a :class:`JSONResponse` (HTTP 400
+    + JSON-RPC error envelope) on rejection so the caller can early-
+    return it.
+    """
+    if method == "initialize":
+        return None
+    protocol_header = request.headers.get("mcp-protocol-version")
+    if protocol_header is None or protocol_header == PROTOCOL_VERSION:
+        return None
+    _log.warning(
+        "mcp_unsupported_protocol_version",
+        method=method,
+        header=protocol_header,
+        supported=PROTOCOL_VERSION,
+    )
+    return _error_response(
+        _coerce_request_id(payload),
+        INVALID_REQUEST,
+        (
+            f"unsupported MCP-Protocol-Version: {protocol_header!r} "
+            f"(server supports {PROTOCOL_VERSION!r})"
+        ),
+        status_code=400,
+    )
+
+
+def _build_success_response(
+    request_id: JsonRpcId,
+    result: _McpHandlerResult,
+) -> Response:
+    """Wrap a successful handler result in the JSON-RPC response envelope.
+
+    Three handler return shapes are accepted: a :class:`BaseModel` (dumped
+    with ``exclude_none=True`` so optional MCP fields like
+    ``InitializeResponse.instructions`` are omitted), a plain ``dict``
+    (used verbatim), and any other shape — which is treated as a handler
+    bug and converted to INTERNAL_ERROR. The else-arm is load-bearing
+    for the non-notification contract: returning ``None`` from a
+    request-shaped handler would otherwise emit a wire-broken envelope
+    that fails the spec's exactly-one-of(result, error) invariant.
+    """
+    if isinstance(result, BaseModel):
+        result_body: dict[str, Any] = result.model_dump(mode="json", exclude_none=True)
+    elif isinstance(result, dict):
+        result_body = result
+    else:
         return _error_response(
-            _coerce_request_id(payload),
-            INVALID_REQUEST,
-            f"invalid request: {exc.error_count()} validation error(s)",
+            request_id,
+            INTERNAL_ERROR,
+            "handler returned no result for a non-notification request",
         )
+    response = JsonRpcResponse(id=request_id, result=result_body)
+    return JSONResponse(content=_serialize_response(response))
 
-    # MCP-Protocol-Version header validation per spec §"Protocol Version
-    # Header". Clients MUST send the header on every post-initialize
-    # request; if the value is invalid or unsupported the server MUST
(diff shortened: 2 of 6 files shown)


zdamir:
@coderabbitai review

ilhan-karic (changes requested):
## Overview
Adds `POST /mcp` as the MCP server entrypoint: JSON-RPC 2.0 dispatch, lifecycle handlers (`initialize`, `notifications/initialized`, `ping`), Streamable-HTTP single-response shape, and a `register_method()` table mirroring `register_probe`. Schema + dispatcher split is tidy, spec citations are precise, and the test matrix walks every JSON-RPC error code. A few real concerns and several nits below.

## Correctness

- **Capability advertisement is ahead of the dispatch table.** `_initialize` returns `capabilities.tools={"listChanged": False}` and `capabilities.resources={"listChanged": False, "subscribe": False}`, but `tools/list`, `tools/call`, `resources/list`, `resources/read` are not registered yet. A spec-conforming client will read those capabilities and immediately call `tools/list`, get `-32601`, and may disconnect. The PR's own test `test_unknown_method_returns_method_not_found_for_request` uses `tools/list` to demonstrate this. Either:
  - leave both fields out of the `ServerCapabilities` envelope until T3 lands (`tools=None` / `resources=None` already serialize as omitted), or
  - register lightweight stubs in T1 that return empty lists.

  Advertising a capability the server doesn't honor is the one item I'd treat as blocking.

- **`_coerce_request_id` accepts `bool`.** `isinstance(raw, (int, str))` matches `True` / `False` because `bool` subclasses `int`. JSON-RPC §4.1.2 restricts ids to Number, String, or NULL. Cheap fix: short-circuit `isinstance(raw, bool)` to `None` before the int/str check.

- **`is_notification` with explicit `id: null`.** Current behavior: `id` present in payload but value `null` → `"id" in payload` is True → treated as a request, echoes `id: null` in the response. The schema docstring (`schemas.py:155-157`) calls the absent/null distinction "normatively discouraged" and elides them, but the dispatcher (`server.py:659`) does honor the distinction. Benign, but worth a one-line comment so a future cleanup doesn't "unify" the two paths and break the wire contract.

## Security / robustness

- **No request-size cap.** `await request.body()` + `json.loads` will allocate however many bytes the client sends. Belongs to T2's transport hardening, but worth filing now — `/mcp` is public-facing once auth lands.
- **No content-type check.** Clients posting `text/plain` succeed silently. Spec-aligned, but a stricter `415` on non-`application/json` makes the contract clearer.
- **Audit gap is documented but real.** Per the PR body, `/mcp` traverses `AuditMiddleware` but produces no row because `operator_sub` isn't bound in T1. Confirm T5 (#250) is tracked and that this window is explicitly accepted — otherwise every MCP call between T1 and T5 is unaudited.

## Style / minor

- `_serialize_response` hand-rolls the omit-the-unset-half logic because `exclude_none` would also drop `id: null`. Fine, but a `model_dump(mode="json", exclude_none=True)` followed by `out["id"] = resp.id` would be marginally less surprising than building the dict from scratch.
- `InitializeRequest.model_validate(params or {})` collapses `None` and `{}` — spec-aligned, just worth a comment.
- `_McpInvalidParamsError` is a private sentinel inside `server.py`. If T3 wants tool handlers to raise it, it'll need to move to `schemas.py` or be re-exported.
- `test_register_method_rejects_duplicate_registration` doesn't restore state — fine here because it raises before mutating, but a comment to that effect would prevent future test edits from leaking a handler.

## Tests
Coverage is strong: every JSON-RPC error code, GET→405, batch-array rejection, duplicate registration, internal-error path. One small gap — no test asserting that a registered handler returning a non-dict / non-BaseModel scalar (e.g. a list) hits the `INTERNAL_ERROR` arm at `server.py:701-708`. Worth one line.

## Summary
Blocking on the `tools` / `resources` capability advertisement — drop the envelopes until T3 (or register empty-list stubs now). The `bool`-id slip and the audit-window note are follow-ups, not blockers.

backend/src/meho_backplane/mcp/server.py
```diff
@@ -0,0 +1,391 @@
...
+    except ValidationError as exc:
+        raise _McpInvalidParamsError(
+            f"initialize: {exc.error_count()} validation error(s)",
+        ) from exc
+
+    return InitializeResponse(
+        capabilities=ServerCapabilities(
+            tools={"listChanged": False},
+            resources={"listChanged": False, "subscribe": False},
+        ),
```
damir-topic: **🛑 Blocker** — Capability advertisement runs ahead of the dispatch table.

**What:** `_initialize` returns `capabilities=ServerCapabilities(tools={'listChanged': False}, resources={'listChanged': False, 'subscribe': False})`, but `tools/list`, `tools/call`, `resources/list`, `resources/read` are not registered in `_DISPATCH` (verified with `grep` — only docstring mentions). The PR's own `test_unknown_method_returns_method_not_found_for_request` uses `tools/list` precisely *because* it's not registered.

**Why it matters:** MCP 2025-06-18 §Lifecycle/Operation says: *"Both parties MUST [...] Only use capabilities that were successfully negotiated."* A spec-conforming client reading the advertised `tools` capability from `InitializeResult` will call `tools/list` on the next round trip, receive `-32601 method not found`, and disconnect. Empirically reproduced inside the worktree:

```
Capabilities advertised: {'tools': {'listChanged': False}, 'resources': {'listChanged': False, 'subscribe': False}}
tools/list response: {'jsonrpc': '2.0', 'id': 2, 'error': {'code': -32601, 'message': 'method not found: tools/list'}}
```

The schemas.py:166-184 docstring rationale ("so the negotiated capability namespace is already reachable when T3 wires the registries") is backwards — the namespace becomes reachable *when T3 lands*; advertising it early breaks every spec-conforming client interaction.

**Suggested fix:** drop both fields until T3 registers the handlers. `ServerCapabilities`'s `tools: dict[str, Any] | None = None` and `resources: dict[str, Any] | None = None` already serialize as omitted under `exclude_none=True`.

```suggestion
    return InitializeResponse(
        capabilities=ServerCapabilities(),
        serverInfo={"name": _SERVER_NAME, "version": __version__},
    )
```

Two follow-ups:
1. The `test_initialize_returns_protocol_version_capabilities_and_serverinfo` assertions on `result['capabilities']['tools']` / `['resources']` need to flip to `'tools' not in result['capabilities']` / `'resources' not in result['capabilities']`.
2. The `ServerCapabilities` docstring (schemas.py:166-184) needs to flip its narrative: "T1 omits `tools` / `resources` capability envelopes; T3 (#248) adds them when the registries land."

**Source:** https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle (§Operation: "Only use capabilities that were successfully negotiated")

backend/src/meho_backplane/mcp/server.py
```diff
@@ -0,0 +1,391 @@
...
+    Used on the JSON-RPC validation-error path so an INVALID_REQUEST
+    response can still echo the client's id when one was syntactically
+    parseable (helps clients correlate failures). When the raw value
+    isn't a String / Number / NULL, fall back to ``None`` per spec §5
+    ("If there was an error in detecting the id ... it MUST be Null").
+    """
+    raw = payload.get("id")
+    if isinstance(raw, (int, str)) or raw is None:
+        return raw
+    return None
```
damir-topic, line 323: **⚠️ Major** — `_coerce_request_id` accepts `bool`, and the result is a wire-contract violation (not just an out-of-spec accept).

**What:** `isinstance(raw, (int, str))` matches `True` / `False` because `bool` subclasses `int` in Python. Combined with `JsonRpcRequest`'s `id: JsonRpcId = None` field (where `JsonRpcId = int | str | None`, schemas.py:64,90) — Pydantic v2 coerces `True` → `1` on validation. Net effect, empirically:

```
client → POST /mcp {"jsonrpc":"2.0","id":true,"method":"ping"}
server → 200  {"jsonrpc":"2.0","id":1,"result":{}}
```

**Why it matters:** JSON-RPC 2.0 §4 restricts the id to *"a String, Number, or NULL value if included"* — bool is explicitly out. §5 then mandates that the response id *"MUST be the same as the value of the id member in the Request Object."* `true` → `1` violates both. No real MCP client sends bool ids, but the wire echoing the wrong id is the kind of latent surprise that bites once one does.

**Suggested fix:** short-circuit bool in both places.

In `_coerce_request_id` (this hunk):
```suggestion
    raw = payload.get("id")
    if isinstance(raw, bool):
        return None
    if isinstance(raw, (int, str)) or raw is None:
        return raw
    return None
```

In `schemas.py`'s `JsonRpcRequest`, add a model validator that rejects bool ids on the happy path so the dispatcher's `INVALID_REQUEST` path catches them too:
```python
@model_validator(mode="before")
@classmethod
def _reject_bool_id(cls, data: Any) -> Any:
    if isinstance(data, dict) and isinstance(data.get("id"), bool):
        raise ValueError("JSON-RPC id MUST be String, Number, or NULL (spec §4)")
    return data
```

Plus one test asserting `{"id": true, "method": "ping"}` → `-32600 invalid request` with `id: null`.

**Source:** https://www.jsonrpc.org/specification (§4, §5)

backend/src/meho_backplane/mcp/schemas.py
```diff
@@ -0,0 +1,221 @@
...
+#: ``InitializeResponse.protocolVersion`` when the client requests this
+#: version; otherwise the server responds with the same value anyway and
+#: the client decides whether to disconnect (spec §Version Negotiation).
+PROTOCOL_VERSION: str = "2025-06-18"
+
+
+#: A JSON-RPC 2.0 id is "a String, Number, or NULL value if included"
+#: (spec §4.1). A request *without* an id is a notification (§4.1.2);
+#: the absent / explicit-null distinction is normatively discouraged on
+#: the request side, so this server treats both as the same shape.
```
damir-topic: **💡 Minor** — Comment contradicts the dispatcher.

**What:** This comment says the absent / explicit-null distinction is "normatively discouraged on the request side, so this server treats both as the same shape." But `mcp_dispatch` at server.py:339 does honor the distinction: `is_notification = "id" not in payload or jrpc.method.startswith("notifications/")` — explicit `id: null` with a non-`notifications/*` method is treated as a request and echoes `id: null` in the response.

**Why it matters:** Comment drift sets up a future cleanup pass to "unify" the two paths and break the wire contract that the dispatcher carefully implements. Copilot's review caught the same thing.

**Suggested fix:** flip the framing to match server.py's reality:

```
#: A JSON-RPC 2.0 id is "a String, Number, or NULL value if included"
#: (spec §4.1). A request *without* an id is a notification (§4.1.2).
#: This model collapses absent and explicit-null into the same
#: Python-side ``None`` because Pydantic v2 cannot distinguish them
#: post-validation; the dispatcher (server.py:mcp_dispatch) consults
#: the raw dict for ``"id" in payload`` to recover the distinction —
#: absent → notification (HTTP 202, no reply); explicit null →
#: request with id=null in the response.
```

backend/src/meho_backplane/mcp/server.py
```diff
@@ -0,0 +1,391 @@
...
+        result_body = result
+    else:
+        # Handler returned None (or a non-dict scalar) for a request-
+        # shaped invocation. That's a handler bug; surface as
+        # INTERNAL_ERROR rather than emitting a wire-broken envelope.
+        return _error_response(
+            jrpc.id,
+            INTERNAL_ERROR,
+            "handler returned no result for a non-notification request",
+        )
```
damir-topic: **💡 Minor** — Missing test for the non-dict / non-BaseModel return arm.

**What:** This branch is the load-bearing handler-bug guard (turns a malformed handler return into `-32603` instead of a wire-broken envelope), but nothing in `test_mcp_server.py` exercises it. `test_handler_exception_becomes_internal_error` covers the `except Exception` arm; this one — "handler returned a list / scalar / None" — is reachable but unasserted.

**Why it matters:** the contract "handler must return BaseModel or dict" is enforced here; without a test, a future refactor (e.g. extracting the three sub-helpers CodeRabbit suggested) could lose this guard silently and start emitting wire-broken envelopes only under handler-bug conditions, which CI wouldn't catch.

**Suggested fix:** one test, monkeypatched handler returning a `list` (or a string, or `None` for a request-shaped method), asserting `body['error']['code'] == INTERNAL_ERROR`. Same `monkeypatch.setitem(server_module._DISPATCH, ...)` pattern as `test_handler_exception_becomes_internal_error`.

backend/src/meho_backplane/mcp/server.py
```diff
@@ -0,0 +1,391 @@
...
+    * On a *request* (id present): return 200 + single JSON envelope.
+    * On a *notification* (id absent, or method prefix
+      ``notifications/``): return 202 with no body, regardless of
+      whether the handler errored — JSON-RPC §4.1.2 forbids replying.
+
+    GET requests on this path automatically return HTTP 405 (FastAPI's
+    default for an unmatched method) which satisfies the spec's
+    fallback when the server does not implement the GET-opens-SSE
+    branch of the Streamable HTTP transport.
+    """
```
damir-topic, line 342: **⚠️ Major** — `MCP-Protocol-Version` header is never read or validated.

**What:** MCP 2025-06-18 §Transports/"Protocol Version Header" says clients MUST include `MCP-Protocol-Version: <protocol-version>` on every post-init HTTP request, and: *"If the server receives a request with an invalid or unsupported `MCP-Protocol-Version`, it MUST respond with `400 Bad Request`."* T1 doesn't read the header at all — clients sending `MCP-Protocol-Version: 2099-12-31` succeed instead of getting a 400.

**Why it matters:** the only supported version in T1 is `2025-06-18`, so a strict-conformance acceptance test (T6 / #251 with MCP Inspector) will surface this. Filing it now lets T2 (#247, transport hardening) pick it up alongside Origin-header validation rather than discovering it during the acceptance milestone.

**Suggested fix:** either add a one-shot middleware / dependency on `/mcp` that 400s on a non-`2025-06-18` `MCP-Protocol-Version` header (or absent on a non-initialize request once T2 wires session tracking), or explicitly defer to T2 with a code comment + a follow-up task on #247.

**Source:** https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#protocol-version-header

backend/src/meho_backplane/mcp/server.py
```diff
@@ -0,0 +1,391 @@
...
+    raw_body = await request.body()
```
damir-topic: **💡 Minor** — Unbounded `request.body()` read.

**What:** `await request.body()` reads the entire body into memory before `json.loads`. No size cap. A client posting a 1 GiB body OOMs the worker.

**Why it matters:** `/mcp` is public-facing once T2 (#247) lands the auth chain; even with auth, a token-holding bad actor (e.g. a misbehaving operator script) can DOS the backplane with a single oversized POST.

**Suggested fix:** belongs with T2's transport-hardening sweep (alongside Origin-header validation + content-type check). File as a sub-task on #247, or pre-empt with a request-size limit on the FastAPI route here. Either works — the important thing is it doesn't get lost between T1 and T2.

backend/src/meho_backplane/mcp/server.py
```diff
@@ -0,0 +1,391 @@
...
+    # ``notifications/*`` method with an id — the method semantics
+    # are spec-defined as notification-only, so the server treats it
+    # as such regardless of the envelope's id field.
+    is_notification = "id" not in payload or jrpc.method.startswith("notifications/")
+
+    handler = _DISPATCH.get(jrpc.method)
+    if handler is None:
+        if is_notification:
+            _log.warning("mcp_unknown_notification", method=jrpc.method)
+            return Response(status_code=202)
```
damir-topic, line 460: **❓ Question** — Should unknown / invalid notifications get `4xx` instead of `202`?

**What:** This arm (and the `except _McpInvalidParamsError` arm at line 355-361) returns HTTP 202 for notifications the server cannot process — unknown method, invalid params, internal error.

MCP transport §Sending Messages to the Server, point 4 is split:
* *"If the server **accepts** the input, the server MUST return HTTP status code 202 Accepted with no body."*
* *"If the server **cannot accept** the input, it MUST return an HTTP error status code (e.g., 400 Bad Request). The HTTP response body MAY comprise a JSON-RPC error response that has no id."*

T1 currently returns 202 for both halves. Reading the spec strictly, an unknown-method notification is "input the server cannot accept" → should be 4xx.

**Counter-argument:** JSON-RPC §4.1.2 says *"Notifications are not confirmable by definition"* — the server can't tell the client about the JSON-RPC error anyway, so 202 + a structlog warning (which is what the code does at line 344) is operationally equivalent.

Not a blocker either way, but the chosen behavior should be (a) pinned in a code comment with the trade-off articulated, and (b) covered by a test that asserts the specific status code chosen, so a future cleanup doesn't accidentally flip it.

**Source:** https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server

backend/src/meho_backplane/mcp/server.py
```diff
@@ -0,0 +1,391 @@
...
+
+class _McpInvalidParamsError(Exception):
+    """Handler-side sentinel mapped to JSON-RPC ``INVALID_PARAMS``.
+
+    Raised by a handler when its params fail validation (e.g.
+    :class:`~meho_backplane.mcp.schemas.InitializeRequest.model_validate`
+    raises :class:`pydantic.ValidationError`). The dispatcher catches
+    this distinctly from a generic :class:`Exception` so the wire
+    response carries code ``-32602`` rather than ``-32603``.
+    """
```
damir-topic: **💡 Minor** — `_McpInvalidParamsError` is private to `server.py`; T3 tool handlers will want to raise it.

**What:** The sentinel maps cleanly to `-32602`, which is exactly what a `tools/call` handler will want when its params fail validation (the typical case once T3 lands the registry). Keeping it module-private means T3 either re-implements the sentinel, imports it via the dunder-prefixed name, or refactors at that point.

**Suggested fix:** move to `schemas.py` (where the other public error-mapping primitives live — `INVALID_PARAMS` and friends) and re-export from `mcp/__init__.py`. Easy follow-up under T3 (#248) so this PR stays scoped.

Noted by ilhan-karic too.

damir-topic (changes requested):
## /review-pr-slim summary

**Anchor task:** #246 — G0.5-T1 MCP server entrypoint
**Surface area:** backend (Python / FastAPI / Pydantic v2 / JSON-RPC 2.0 / MCP 2025-06-18)
**Verdict:** REQUEST_CHANGES

One blocker, two majors, several minors. The transport plumbing, lifecycle handlers, dispatch-table API, and test matrix are all solid — the blocker is a single design call on capability advertisement that breaks spec conformance for any non-trivial MCP client.

### Acceptance criteria
- [x] `/mcp` accepts JSON-RPC 2.0 POSTs — verified empirically.
- [x] `initialize` returns `serverInfo`, capabilities, `protocolVersion=2025-06-18` — verified.
- [x] `notifications/initialized` accepted with no body (202 per spec; the deviation note in the PR body is correct — spec MUST overrides AC's 204).
- [x] Unknown method → -32601 (request-shaped).
- [x] Parse error → -32700, `id: null`.
- [x] Invalid params → -32602.
- [x] `ping` → `result: {}`.
- [x] T1 has no auth; every well-formed request succeeds — dedicated test asserts.
- [x] CI green (ruff + mypy + pytest).
- [⚠️] Spec-conformance side of "lifecycle handshake": fails for spec-conforming clients because of the capability mismatch — see blocker below.

|---|---|---|---|---|---|
| #246 (Task) | In Progress | `Closes #246` | ✅ on merge | ⚠️ partial — capability advertisement blocker | Bump status to `In Review`; fix blocker before merge |
| #226 (Initiative) | Todo | parent of #246 | — | — | Stays `Todo`; 5 more tasks (#247–#251) remain |

No undisclosed scope, no orphaned siblings. PR cleanly closes only #246. The only board hygiene nit is #246 still sitting in `In Progress` after the PR opened — bump to `In Review`.

### Predicted vs. actual diff
- ✅ `backend/src/meho_backplane/mcp/__init__.py`, `mcp/schemas.py`, `mcp/server.py` — expected and structurally correct.
- ✅ `backend/src/meho_backplane/main.py` — mounts the router with the documented middleware-ordering rationale carried forward.
- ✅ `backend/tests/test_mcp_server.py` — 18 cases, walks every JSON-RPC error code, GET→405, batch arrays, duplicate registration, internal-error path. Above the bar.
- ✅ `docs/codebase/backend.md` — new MCP section accurately reflects the implementation.
- 🕳️ No `MCP-Protocol-Version` header validation — spec says MUST 400 on invalid version, and v0.2 only supports `2025-06-18`. Scope cut is defensible if filed as a T2/T6 follow-up; flagged as major below.

### Integration verification
All commands ran inside `.claude/worktrees/pr-266-review/` against PR HEAD `58680c6`.

- [x] `ruff format --check src/ tests/` — 4 files already formatted.
- [x] `mypy src/` — 27 source files, no issues.
- [x] `pytest tests/test_mcp_server.py` — 18 passed.
- [x] `pytest tests/` (full) — 246 passed, 1 deselected (`test_metrics_endpoint_returns_prometheus_text_format` — pre-existing macOS-specific failure on `main`, not introduced by this PR).
- [x] Empirically reproduced the capability mismatch: `initialize` → server advertises `capabilities.tools={"listChanged": false}` → follow-up `tools/list` returns `-32601`.
- [x] Empirically reproduced the bool-id behavior: client sends `{"id": true, "method": "ping"}` → server responds with `{"id": 1, "result": {}}`. JSON-RPC §5 requires the response id equal the request id; `true` → `1` is a wire-contract violation.
- [⚠️ not run] `docker compose up` — not exercised; new module is in-process Python only, no infra dependency changes.
- [⚠️ not run] MCP Inspector end-to-end — out of scope for T1 per AC; lands in T6 (#251). Recommend filing a T6 sub-task to add Inspector to the matrix specifically for the capability-advertisement contract once T3 lands.

### Findings
- 🛑 1 blocker — capability advertisement of `tools` / `resources` ahead of dispatch handlers (server.py:176-182, schemas.py:166-184). See inline.
- ⚠️ 2 majors — bool id slip (schemas.py:64, server.py:269-281); missing `MCP-Protocol-Version` header enforcement (server.py:284-391).
- 💡 4 minors — docstring contradicts dispatcher on `id: null` (schemas.py:60-63, raised by Copilot too); missing test for non-dict/non-BaseModel scalar return (server.py:380-388, raised by ilhan-karic); no request-size cap (server.py:301); `_McpInvalidParamsError` is private to `server.py` (server.py:143-151).
- ❓ 1 question — notifications with invalid params / unknown method return HTTP 202 (server.py:343-345, 355-361), but the Streamable HTTP spec says "if the server cannot accept the input, it MUST return an HTTP error status code."

### Docs
- [x] `docs/codebase/backend.md` — updated to reflect new MCP module; the entry's accuracy holds *after* the blocker is fixed. If `tools` / `resources` capability lines are removed, the doc text says "T1 advertises … T3 fills" — consider tightening to "T1 omits `tools`/`resources` capability envelopes; T3 (#248) adds them when the registries land" so the docs match the wire.

### What was done well
- 👏 Spec citations in docstrings are precise and pinned to `2025-06-18`. Future contributors will know exactly which paragraph is load-bearing.
- 👏 The AC #3 deviation (202 vs 204) is the right call and is documented in the PR body, the test docstring, and the issue follow-up surface. Exactly how MUST-vs-AC conflicts should be resolved.
- 👏 `JsonRpcResponse._exactly_one_of_result_or_error` enforces the wire contract on construction — defends against future handler bugs emitting wire-broken envelopes.
- 👏 `_serialize_response` correctly preserves `id: null` on parse-error responses (JSON-RPC §5 mandates) — would have been easy to lose to a blanket `exclude_none=True`.
- 👏 Test matrix is unusually thorough for a T1 PR — every JSON-RPC error code, GET→405, batch arrays, duplicate registration, internal-error path, and a dedicated "T1 has no auth" assertion. The dispatcher's contract is fully pinned.
- 👏 `register_method` mirrors `register_probe`'s shape exactly, which makes T3's job mechanical.

backend/src/meho_backplane/mcp/server.py
zdamir: DISPUTE — keeping 202 on the post-envelope notification paths, with rationale now inlined as a code comment (commit `adb6440`).

The spec text "If the server **cannot accept** the input" is intentionally narrow — it refers to **transport-level** rejection (malformed envelope, batch arrays, bad JSON, unsupported `MCP-Protocol-Version`). The transport-level rejections in `mcp_dispatch` already flip to HTTP 4xx because they run *before* the notification/request split; the post-envelope arms stay at 202 because:

1. JSON-RPC §4.1.2: "The Server MUST NOT reply to a Notification." A 4xx with a JSON-RPC error body is borderline-replying and may confuse spec-conforming clients.
2. Notifications are not confirmable — the client cannot retry on 4xx without violating §4.1.2. Application-level failures on a notification are the server's problem to log, not the client's to fix.
3. Most MCP servers in the wild treat unknown notifications as silently-dropped (forward-compat: a future revision may add a notification type the older server doesn't know).

Commit `adb6440` adds the comment explaining the split. Happy to revisit if T6's (#251) MCP Inspector workflow surfaces a real spec-conformance break.

zdamir:
@coderabbitai review

zdamir:
@coderabbitai review

backend/src/meho_backplane/mcp/server.py
zdamir: Fixed in commit `094019f`. Two gates landed:

* `JsonRpcRequest.id` grew a `mode="before"` `field_validator` that rejects `bool` outright (so `model_validate` raises `ValidationError`, which the dispatcher maps to `-32600`).
* `_coerce_request_id` got an explicit `isinstance(raw, bool)` short-circuit ahead of the `int | str` arm, so the validation-error path doesn't echo `True` / `False` back as `1` / `0`.

Parametrized test `test_bool_id_is_rejected_as_invalid_request` covers both gates with `True` and `False`. Response correctly carries `id: null` per JSON-RPC §5.

backend/src/meho_backplane/mcp/server.py
zdamir: Fixed in commit `a10b0c4`. Pragmatics:

* If `MCP-Protocol-Version` header is present and not `2025-06-18` → HTTP 400 + JSON-RPC error envelope (spec §"Protocol Version Header" MUST). Body MAY include a JSON-RPC error response per §"Sending Messages to the Server"; we do.
* `initialize` is exempted — clients don't know the version pre-handshake.
* Header absent on non-initialize is accepted as transitional lenience. The spec's SHOULD-assume-`2025-03-26` fallback doesn't help — v0.2 doesn't support that revision — and tightening it in T1 would break clients that don't yet emit the header. T6 (#251) will pin the strict-mode contract.

Four tests cover the matrix:
* `test_protocol_version_header_matching_supported_is_accepted` — happy path
* `test_protocol_version_header_unsupported_returns_400` — wrong version → 400 with id echoed
* `test_protocol_version_header_absent_is_accepted_transitionally` — no header → request proceeds
* `test_initialize_does_not_require_protocol_version_header` — initialize with wrong header still succeeds

The validation lives in `_validate_protocol_version_header` (after the iteration-3 refactor `8262277`).

Issue #263, opened 2026-05-12
Task: G0.4-T6 E2E smoke + chart preflight (helm test) — verify pgvector enabled + hybrid fusion working + tenant boundary holds


damir-topic:
Parent initiative: #225

## Target

End-to-end smoke + chart preflight verification — proves the whole G0.4 stack works in a real deploy:

- `backend/tests/integration/test_retrieval_e2e.py` — new integration test exercising index → retrieve across two tenants with realistic corpora.
- `backend/src/meho_backplane/db/migrations.py` — extend `db_migration_probe` to also verify the `vector` extension is enabled.
- `deploy/charts/meho/templates/tests/test-pgvector.yaml` — Helm test that runs against a deployed backplane + asserts pgvector is available.
- `.github/workflows/ci.yml` — gate: the chart-test stage runs after `helm install` in CI; fails the deploy if pgvector isn't enabled.

## Context

T6 is the **acceptance harness** for G0.4. The previous Tasks shipped the components; T6 proves they work together in a deployed environment, not just against pytest fixtures.

**Three layers of verification:**

1. **Application-layer integration test.** Indexes ~15 documents across 2 tenants → runs 6 queries → asserts hits, RRF fusion, tenant boundary, source/kind filters. Catches bugs in the helper interactions (T3+T4) that single-component tests miss.

2. **Migration-layer probe extension.** `db_migration_probe` already validates that Alembic head == DB version. T6 extends it to also `SELECT extversion FROM pg_extension WHERE extname = 'vector'` and report `vector_extension: ok` or `vector_extension: missing` in the probe result. Operators see this on `/ready` if they deployed to a Postgres without pgvector.

3. **Chart-layer Helm test.** A `helm test`-runnable Job spins up a small pod that runs a SQL preflight (`SELECT * FROM documents LIMIT 1; SELECT extversion FROM pg_extension WHERE extname = 'vector';`) against the deployed Postgres. CI runs `helm install --wait` then `helm test`; if pgvector is missing or the migration didn't apply, CI fails.

**Realistic corpus for the integration test.** Index 15 docs derived from the consumer's `kb/` directory shape:

- 8 kb-entries about Kubernetes (clustering, networking, troubleshooting, etc.) — should rank against "kubernetes" queries.
- 4 kb-entries about Vault (auth, kv-v2, OIDC, audit) — should rank against "vault" queries.
- 3 memory entries (mixed personal preferences + project context) — used to test source filtering.

[...]

Issue #259, opened 2026-05-12
Task: G0.4-T2 fastembed wiring — EmbeddingService + lazy load + lifespan preload + chart model-cache PVC (default BAAI/bge-small-en-v1.5)


damir-topic:
Parent initiative: #225

## Target

fastembed wiring — bring up the in-process embedding pipeline + chart-side PVC for model cache:

- `backend/pyproject.toml` — add `fastembed` dependency.
- `backend/src/meho_backplane/retrieval/__init__.py` — new package.
- `backend/src/meho_backplane/retrieval/embedding.py` — `EmbeddingService` class with lazy model loading + batched encode method.
- `backend/src/meho_backplane/settings.py` — add `RETRIEVAL_EMBEDDING_MODEL` setting (default `BAAI/bge-small-en-v1.5`).
- `backend/src/meho_backplane/main.py` — lifespan eager-loads the embedding model so first-request isn't slow.
- `deploy/charts/meho/values.yaml` — add `retrieval.modelCache` knob (default PVC at `/var/cache/fastembed`, 200Mi size).
- `deploy/charts/meho/templates/deployment.yaml` — render the model-cache volume mount when configured.
- `deploy/charts/meho/values.schema.json` — schema entries for retrieval settings.
- `backend/tests/test_retrieval_embedding.py` — new tests.

## Context

T2 is the **embedding infrastructure** that T3 (`index_document`) and T4 (`retrieve`) will consume. Critical decisions baked in:

**fastembed, not sentence-transformers.** Per v0.1-spec L391, the chosen library. fastembed uses ONNX models with no PyTorch dependency — ~120MB total install vs ~1.5GB for sentence-transformers + torch. Memory profile matters because the backplane is single-replica.

**Lazy model loading.** First call to `EmbeddingService.encode(...)` loads the model from disk (or downloads from HF if absent in cache). Subsequent calls reuse the loaded model. Lifespan eager-loads at startup to amortize the load time across the pod lifetime, not per-request.

**Model cache PVC.** fastembed downloads ONNX weights on first use (~120MB for `BAAI/bge-small-en-v1.5`). Without a PVC, every pod restart re-downloads. The chart mounts a 200Mi PVC at `/var/cache/fastembed` (or `~/.cache/fastembed` per fastembed's convention). Operators can configure to use an existing PVC for multi-replica deployments (v0.2.next).

[...]

Issue #225, opened 2026-05-11
Initiative: G0.4 Retrieval substrate — pgvector + fastembed + hybrid BM25/cosine RRF; the shared substrate G4 (#215) kb and G5 (#216) memory both depend on


damir-topic:
Parent goal: #221

## Scope

Stand up the shared retrieval substrate that G4 (#215, knowledge layer) and G5 (#216, memory layer) both depend on, per v0.1-spec L382–401: pgvector extension on the chassis Postgres + fastembed (in-process ONNX) + hybrid retrieval (BM25 via Postgres FTS + cosine via pgvector, fused with Reciprocal Rank Fusion). Ship the substrate; not the consumers. G4 owns kb ingestion, G5 owns memory writes; both go through the helpers this Initiative provides.

**Area:** [backend/src/meho_backplane/retrieval/](backend/src/meho_backplane/retrieval/) (new module), new Alembic migration adding the pgvector extension + a `documents` table, [backend/pyproject.toml](backend/pyproject.toml) (fastembed dep), [deploy/charts/meho/values.yaml](deploy/charts/meho/values.yaml) (`retrieval.embedding_model` knob).

The work items grouped under this Initiative:

1. **pgvector extension migration.** Alembic migration runs `CREATE EXTENSION IF NOT EXISTS vector;`. Fails fast with a clear error if the Postgres image lacks pgvector (most cloud Postgres offerings have it; SQLite test envs skip pgvector-dependent steps via a feature flag). The chassis Postgres subchart is operator-provided, so the migration's preflight check verifies the extension is available before running the rest.
[...]

PR #268, opened 2026-05-12
test(backplane): per-tenant isolation integration suite (G0.1-T6)

damir-topic:
## Summary

- Adds `backend/tests/integration/test_tenant_isolation.py` — broad-spectrum end-to-end proof that the G0.1 tenancy chain (T1 schema, T2 JWT extraction, T3 contextvar binding + AuditMiddleware, T4 RBAC) holds together against a real Postgres under concurrent load.
- Five integration cases mirror the issue body's coverage matrix: (1) tenant-scoped audit rows (5+3 split), (2) v0.2 trust-model contract for unknown `tenant_id` claims, (3) per-tenant audit-query helper as a forward-compat probe for G8, (4) `read_only` JWT → 403 sanity for T4 RBAC, (5) **the concurrent-request contextvar isolation test** — eight interleaved requests under `asyncio.gather` from four operators across two tenants. Case 5 is the highest-value test per the issue body; catches structlog contextvar leaks across concurrently-executing asyncio tasks.
- Real Postgres via `testcontainers` + `postgres:16-alpine` (same image pin as `test_migration_rollback` / `test_db_engine`). Module-scoped PG container amortises the ~3s boot cost across all five tests so the suite stays inside the < 10s acceptance criterion. `httpx.AsyncClient` + `ASGITransport` is used for the concurrent case (TestClient is sync — would serialise `asyncio.gather` and miss the contextvar-leak bug class entirely).
- Skip pattern mirrors `test_migration_rollback`'s class-level Docker gate (`@_skip_no_docker`): agent sandboxes without Docker skip the PG-driven cases; CI runners with Docker provisioned run the full suite. The cheap import smoke `test_module_imports_cleanly` runs unconditionally so a renamed fixture surfaces at collection time even where the PG suite skips.

## Test plan

- [x] `uv run ruff check src/ tests/` — clean
- [x] `uv run ruff format --check src/ tests/` — clean
- [x] `uv run mypy src/` — clean (mypy targets `src/` only per `pyproject.toml`)
- [x] `uv run pytest tests/ -q` — 233 passed, 7 skipped (5 new PG-Docker tests + 2 chassis Docker tests skip in sandbox; CI runs them)
- [x] `python3 .claude/hooks/code-quality.py --diff` — exit 0, no new warnings
- [x] Manual integration smoke against SQLite — `build_integration_app()` builds, `/api/v1/health` returns 200 with the test JWT, audit row written with the correct `tenant_id` claim from the contextvar binding.
[...]

PR #269, opened 2026-05-12
feat(backplane): pgvector substrate — documents table + Document model (G0.4-T1)

### LINKED ISSUE #258: Task: G0.4-T1 Alembic migration — pgvector extension + documents table + indexes (unique, GIN-FTS, IVFFlat-cosine, body_hash)
Parent initiative: #225
Depends on: G0.1-T1 (#231) — `tenant` table must exist before the `documents.tenant_id` FK is meaningful.

## Target

Alembic migration enabling pgvector + creating the shared `documents` table:

- `backend/alembic/versions/0004_create_documents_with_pgvector.py` — new migration; enables `vector` extension; creates `documents` table; creates all 4 indexes.
- `backend/src/meho_backplane/db/models.py` — new `Document` SQLAlchemy model with `Vector` column type.
- `backend/tests/test_db_documents.py` — new file. Asserts table accepts inserts, vector column round-trips, indexes work.

## Context

T1 is the schema foundation for the shared retrieval substrate. G4 (#215, kb migration) and G5 (#216, memory layer) both write rows into this table; the embedding/index/retrieve helpers in T2-T6 work over it. Land schema first.

**pgvector extension.** Migration runs `CREATE EXTENSION IF NOT EXISTS vector;` at the start. Most managed Postgres offerings (RDS, Cloud SQL, Azure DB) support pgvector natively. The chassis Postgres image must support it; if it doesn't, the migration fails fast with a clear error and operators upgrade their Postgres image. Document the requirement in the migration's docstring.

**SQLite test environment.** SQLite doesn't support pgvector. The migration has a dialect guard: PG runs the extension + the `vector` column; SQLite skips the extension and uses TEXT for the embedding column (tests that exercise actual embedding use a PG fixture, not SQLite). This mirrors the existing chassis migration's dialect handling.

**`documents` table column set** — full from the Initiative body, plus indexes:

```sql
CREATE EXTENSION IF NOT EXISTS vector;

CREATE TABLE documents (
    id           uuid PRIMARY KEY DEFAULT gen_random_uuid(),
    tenant_id    uuid NOT NULL,  -- soft FK to tenant.id
    source       text NOT NULL,  -- "kb" | "memory" | "docs-sidecar" | future
    source_id    text NOT NULL,  -- unique-within-source identifier (kb slug, etc.)
    kind         text NOT NULL,  -- per-source: "kb-entry", "memory-user", etc.
    body         text NOT NULL,
    body_hash    text NOT NULL,  -- SHA256 hex; indexed for change-detection
    tokens       integer,        -- rough token count; populated by indexer
    embedding    vector(384) NOT NULL,  -- 384-dim per BAAI/bge-small default
    metadata     jsonb NOT NULL DEFAULT '{}'::jsonb,
    created_at   timestamptz NOT NULL DEFAULT now(),
    updated_at   timestamptz NOT NULL DEFAULT now()
);

CREATE UNIQUE INDEX documents_tenant_source_id_idx ON documents (tenant_id, source, source_id);
CREATE INDEX documents_body_fts_idx ON documents USING GIN (to_tsvector('english', body));
CREATE INDEX documents_embedding_idx ON documents USING ivfflat (embedding vector_cosine_ops) WITH (lists = 100);
CREATE INDEX documents_body_hash_idx ON documents (body_hash);
```

**IVFFlat vs HNSW for vector index.** IVFFlat is the v0.2 pick — simpler, smaller, faster to build. HNSW (better recall + faster query) is post-v0.2 if recall numbers from G4's kb corpus show IVFFlat under-performs.

**`lists = 100`** is the IVFFlat parameter for ~10k documents (pgvector recommends `lists = rows / 1000`). Acceptable starting point; tune in v0.2.next if the corpus grows.

**FTS index on `to_tsvector('english', body)`.** English-only for v0.2 (matches kb/ language). Multilingual operators set `RETRIEVAL_LANGUAGE` env var (v0.2.next) to switch the TS config. Don't burn cycles on multilingual now.

## Current state

After v0.1 chassis + G0.1-T1:
- `tenant` table exists.
- No `documents` table; no pgvector extension; no FTS indexes.
- pgvector is not in the chassis Postgres image — operators using the chart-bundled Postgres need to upgrade (PR will land alongside this migration to update the Postgres image to `pgvector/pgvector:pg16` or equivalent).

## Desired state

Migration sketch:

```python
"""create documents with pgvector

Revision ID: 0004
Revises: 0003
Create Date: 2026-05-12 ...
"""
import sqlalchemy as sa
from sqlalchemy.dialects.postgresql import JSONB, UUID as PG_UUID
from alembic import op
from pgvector.sqlalchemy import Vector


def upgrade() -> None:
    bind = op.get_bind()
    if bind.dialect.name == "postgresql":
        op.execute("CREATE EXTENSION IF NOT EXISTS vector")

    op.create_table(
        "documents",
        sa.Column("id", PG_UUID(as_uuid=True), primary_key=True, server_default=sa.text("gen_random_uuid()")),
        sa.Column("tenant_id", PG_UUID(as_uuid=True), nullable=False),
        sa.Column("source", sa.Text(), nullable=False),
        sa.Column("source_id", sa.Text(), nullable=False),
        sa.Column("kind", sa.Text(), nullable=False),
        sa.Column("body", sa.Text(), nullable=False),
        sa.Column("body_hash", sa.Text(), nullable=False),
        sa.Column("tokens", sa.Integer(), nullable=True),
        sa.Column("embedding", Vector(384) if bind.dialect.name == "postgresql" else sa.Text(), nullable=False),
        sa.Column("metadata", JSONB() if bind.dialect.name == "postgresql" else sa.JSON(), nullable=False, server_default="{}"),
        sa.Column("created_at", sa.DateTime(timezone=True), nullable=False, server_default=sa.text("now()")),
        sa.Column("updated_at", sa.DateTime(timezone=True), nullable=False, server_default=sa.text("now()")),
    )
    op.create_index("documents_tenant_source_id_idx", "documents", ["tenant_id", "source", "source_id"], unique=True)
    op.create_index("documents_body_hash_idx", "documents", ["body_hash"])
    if bind.dialect.name == "postgresql":
        op.execute("CREATE INDEX documents_body_fts_idx ON documents USING GIN (to_tsvector('english', body))")
        op.execute("CREATE INDEX documents_embedding_idx ON documents USING ivfflat (embedding vector_cosine_ops) WITH (lists = 100)")


def downgrade() -> None:
    op.drop_table("documents")
    # Don't drop the extension on downgrade — other things might depend on it.
```

SQLAlchemy model:

```python
from pgvector.sqlalchemy import Vector  # only available with pgvector lib

class Document(Base):
    __tablename__ = "documents"
    id: Mapped[UUID] = mapped_column(PG_UUID(as_uuid=True), primary_key=True)
    tenant_id: Mapped[UUID] = mapped_column(PG_UUID(as_uuid=True), nullable=False)
    source: Mapped[str] = mapped_column(Text, nullable=False)
    source_id: Mapped[str] = mapped_column(Text, nullable=False)
    kind: Mapped[str] = mapped_column(Text, nullable=False)
    body: Mapped[str] = mapped_column(Text, nullable=False)
    body_hash: Mapped[str] = mapped_column(Text, nullable=False)
    tokens: Mapped[int | None] = mapped_column(Integer)
    embedding: Mapped[list[float]] = mapped_column(Vector(384), nullable=False)
    metadata: Mapped[dict] = mapped_column(JSONB, nullable=False, server_default="{}")
    created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), nullable=False)
    updated_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), nullable=False)

    __table_args__ = (
        Index("documents_tenant_source_id_idx", "tenant_id", "source", "source_id", unique=True),
        Index("documents_body_hash_idx", "body_hash"),
        # FTS + IVFFlat indexes created by raw SQL in the migration; SQLAlchemy doesn't model them
    )
```

`pyproject.toml` adds `pgvector` Python package (the SQLAlchemy type adapter).

## Acceptance criteria

- [ ] Migration runs against PG; `CREATE EXTENSION vector` succeeds; table + all 4 indexes created. Verified via `\d+ documents` + `\dx vector`.
- [ ] `downgrade -1` removes the table (does NOT drop the extension).
- [ ] Round-trip test: insert Document with 384-dim embedding vector; query back; assert vector preserved.
- [ ] Test: unique `(tenant_id, source, source_id)` enforced.
- [ ] Test: cosine similarity query works (`SELECT * FROM documents ORDER BY embedding <=> <query_vector> LIMIT 5`).
- [ ] Test: FTS query works (`SELECT * FROM documents WHERE to_tsvector('english', body) @@ plainto_tsquery('english', 'foo')`).
- [ ] Test SQLite fallback: migration runs on SQLite (skips vector extension; uses TEXT for embedding column); table created; downstream embedding tests are skipped via marker.
- [ ] `db_migration_probe` returns `ok=True` after upgrade.
- [ ] `pgvector/pgvector:pg16` (or equivalent) Postgres image documented as the chassis requirement.
- [ ] CI green.

## Out of scope

- The embedding pipeline (T2 wires fastembed).
- `index_document` / `retrieve` helpers (T3, T4).
- API route (T5).
- HNSW index (post-v0.2).
- Multilingual FTS config (post-v0.2).
- Tiered storage / sharding for >1M document corpora.
- Vector dimensions other than 384.

## References

- Depends on: G0.1-T1 (#231) (tenant table).
- pgvector: https://github.com/pgvector/pgvector (extension docs); https://github.com/pgvector/pgvector-python (Python adapter).
- IVFFlat tuning: https://github.com/pgvector/pgvector#ivfflat
- HNSW (post-v0.2): https://github.com/pgvector/pgvector#hnsw
- v0.1-spec §"Knowledge layer" L382-401 (recipe + dim choice).
- Existing migration to mirror: [backend/alembic/versions/0001_create_audit_log.py](backend/alembic/versions/0001_create_audit_log.py).
- DevOps best-practices: `.claude/skills/implement-issue/devops_best_practices.md` — extension preflight, dialect-portable migrations.
- Python best-practices: `.claude/skills/explore-and-refactor/python_best_practices.md` — SQLAlchemy 2.x typed columns.

ikaric:
## Summary

Lands the schema foundation for Initiative #225 (G0.4 Retrieval substrate): migration `0003` enables the `vector` extension, creates the `documents` table, installs four indexes (unique composite, body_hash btree, GIN FTS, IVFFlat cosine), and adds the `Document` SQLAlchemy model with portable dialect handling.

Closes #258 (G0.4-T1). T2–T6 (#259–#263) build on this schema.

## Why this is one bounded PR

Per the Initiative's six-task breakdown, T1 ships only the schema + ORM model + tests. No retrieval helpers, no embedding pipeline, no HTTP route — those land in T2–T5 as separate PRs. The `Document` ORM class is a pure data shape with no helper methods; T3's `index_document` and T4's `retrieve` (separate PRs) own the read/write paths.

## Dialect portability

Migration runs against both PostgreSQL (production / pgvector image) and SQLite (dev/test). PG runs `CREATE EXTENSION vector` and gets `Vector(384)` + `JSONB` + `gen_random_uuid()`; SQLite skips the extension and uses `Text` + `JSON` + ORM-side `default=uuid.uuid4`. The portable vector type follows the same `with_variant` shape as the existing `_PORTABLE_JSON`:

```python
_PORTABLE_VECTOR_384 = Text().with_variant(Vector(384), "postgresql")
```

GIN-FTS (`USING GIN (to_tsvector('english', body))`) and IVFFlat (`USING ivfflat (embedding vector_cosine_ops) WITH (lists = 100)`) are emitted via raw `op.execute` inside `if is_postgres:` — Alembic has no clean API for expression-based GIN or for IVFFlat with operator-class + `WITH` parameters. They are deliberately **not** declared on `Document.__table_args__` to keep SQLite's `create_all` / autogenerate clean.

## Testcontainers image swap

Migration `0003`'s `CREATE EXTENSION vector` fails fast on stock `postgres:16-alpine` (the v0.1-chassis image) — that's the intended production contract (a deploy against a pgvector-less cluster surfaces the prerequisite at migration time, not first retrieval request). The two testcontainers tests that drive `alembic upgrade head` (`tests/test_db_engine.py`, `tests/test_migration_rollback.py`) switch to `pgvector/pgvector:pg16` accordingly.

## Notable design notes

- **`metadata` column name collision**: `DeclarativeBase.metadata` is reserved. The Python attribute is `doc_metadata`; the SQL column name `metadata` is preserved via `mapped_column("metadata", _PORTABLE_JSON, ...)` so the migration's column identifier stays authoritative.
- **IVFFlat empty-table caveat**: building the IVFFlat index against an empty table produces non-useful centroids until backfill. T3/T4 runbook will note the `REINDEX INDEX documents_embedding_idx` requirement after the first batch. HNSW is the v0.2.next escape hatch if recall numbers from G4 show IVFFlat under-performs (deferred per Initiative body).
- **Downgrade leaves the `vector` extension installed**: namespace-safe; other consumers of the same PG cluster may share it. Reversibility is at the table level only.
- **Numbering note**: the issue body says `0004` revising `0003`, but `main` only has `0001` + `0002`. Used `0003` as the actual next revision; this is documented in the migration's docstring.

## Tests

- `backend/tests/test_db_documents.py` (new) — 6 SQLite tests: round-trip, ORM defaults firing under SQLite, unique `(tenant_id, source, source_id)` enforcement, cross-tenant `body_hash` repetition allowed, `doc_metadata` Python-attribute / `metadata` SQL-column rename, and `updated_at` ORM `onupdate` advancing on UPDATE.
- `backend/tests/test_db_models.py` — 2 new migration-shape tests for `0003`: portable btree indexes land on both dialects, PG-only GIN+IVFFlat do **not** land on SQLite, full upgrade→downgrade→re-upgrade cycle is reversible.
- `backend/tests/test_db_engine.py` — new Docker-gated `test_documents_table_uses_vector_column_on_postgres` asserts `documents.embedding` compiles to `vector(384)`, `vector` extension loaded in `pg_extension`, and all four indexes present on PG.

## Test plan

- [ ] `ruff check .` — clean.
- [ ] `mypy` strict — clean across 24 source files.
- [ ] `pytest` — 237 passed locally (1 pre-existing macOS-specific failure in `test_observability` unrelated to this PR).
- [ ] PG-real Docker-gated tests (`TestPostgresIntegration`) — 2 passed locally against `pgvector/pgvector:pg16`.
- [ ] CI green.

## Summary by CodeRabbit

* **New Features**
  * Per-tenant document storage with dense vector embeddings, JSON metadata, and full-text search for semantic retrieval.
* **Migrations**
  * New migration adds a portable documents table and enables Postgres-only vector/FTS indexes when available.
* **Tests**
  * Added unit and integration tests, including Postgres/pgvector verification of the embedding column and indexes.
* **Documentation**
  * Backend docs updated for the document model, migration notes, and test image/configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/269)


backend/alembic/versions/0003_create_documents_with_pgvector.py
```diff
@@ -0,0 +1,293 @@
...
+def upgrade() -> None:
+    """Enable pgvector, create ``documents`` + four indexes."""
+    bind = op.get_bind()
+    is_postgres = bind.dialect.name == "postgresql"
+
+    # ``vector`` extension — PG only; absent extension on the target
+    # cluster surfaces here as ``extension "vector" is not available``,
+    # which is the intended fail-fast contract (see module docstring).
+    if is_postgres:
+        op.execute("CREATE EXTENSION IF NOT EXISTS vector")
```
damir-topic, line 167: **[🛑 Blocker]** — CI is red on `tests/integration/test_tenant_isolation.py::test_audit_rows_correctly_scoped_per_tenant`; root cause sits here.

**What:** This line runs `CREATE EXTENSION IF NOT EXISTS vector` on every `alembic upgrade head`. After PR #268 landed on `main` (2026-05-12), the new `backend/tests/integration/conftest.py:166` spins up `PostgresContainer("mirror.gcr.io/library/postgres:16-alpine")` and runs `alembic upgrade head` against it. That image has no `vector.control` file, so this line fails with `extension "vector" is not available` and aborts every test in the integration suite.

**Why it matters:** You already fixed the same gap in `tests/test_db_engine.py` and `tests/test_migration_rollback.py` (both correctly swapped to `pgvector/pgvector:pg16`). #268 wasn't merged when you branched from `ba42859`, so the integration conftest wasn't visible — but rebasing onto current `main` re-exposes it and the failure is mechanical, not subtle.

**Suggested fix:**
```bash
git fetch origin main
git rebase origin/main
```
Then in `backend/tests/integration/conftest.py:166`:
```diff
- with PostgresContainer("mirror.gcr.io/library/postgres:16-alpine") as pg:
+ with PostgresContainer("pgvector/pgvector:pg16") as pg:
```
The pgvector image is a drop-in replacement (Postgres 16 base + pgvector pre-installed), same swap you already shipped for the other testcontainers tests.

**Source:** CI run [25735717834](https://github.com/evoila/meho/actions/runs/25735717834/job/75572080655), traceback at `0003_create_documents_with_pgvector.py:167` → `asyncpg.exceptions.FeatureNotSupportedError: extension "vector" is not available`.

backend/alembic/versions/0003_create_documents_with_pgvector.py
```diff
@@ -0,0 +1,293 @@
...
+            primary_key=True,
+            nullable=False,
+            server_default=sa.text("gen_random_uuid()") if is_postgres else None,
+        ),
+        # ``tenant_id`` — soft FK to ``tenant.id``; same shape as
+        # ``audit_log.tenant_id`` (no REFERENCES clause in v0.2; see
+        # 0002's docstring for the rationale). Unlike audit_log, the
+        # column is NOT NULL here — a document with no tenant has no
+        # owner and is not retrievable by any per-tenant query.
+        sa.Column("tenant_id", sa.Uuid(), nullable=False),
```
damir-topic: **[⚠️ Major]** — `documents.tenant_id` soft-FK rationale doesn't fully transfer from `audit_log.tenant_id`.

**What:** This column ships without a `REFERENCES tenant(id)` clause, with the docstring at lines 199-203 citing "same shape as `audit_log.tenant_id`; see 0002's docstring for the rationale". Looking at `0002_create_tenant_and_audit_tenant_id.py:27-46`, the rationale there was specifically: (a) `audit_log.tenant_id` is being added to a table that already has chassis-era rows with no real tenant to point at, and (b) the FK would force the downgrade to either drop the constraint or leave dangling rows referencing a soon-to-be-dropped `tenant` table.

Neither argument transfers to `documents`: it's a *new* table with no pre-existing rows, and `downgrade()` here just drops the entire table (line 292), so there are no dangling refs to worry about. CodeRabbit raised the same point.

**Why it matters:** Without the DB-level FK, `index_document` (T3, #260) can silently insert rows for nonexistent tenant ids (typo, deleted tenant, replay attack via a malformed JWT contextvar). The integrity contract "every document row points at a real tenant" then lives entirely in the app layer and an unsynchronised admin tool / SQL repl can break it. Documents are the source of retrieval — orphaned rows are corpus-poisoning waiting to happen.

**Suggested fix:** Either add the FK now —
```diff
- sa.Column("tenant_id", sa.Uuid(), nullable=False),
+ sa.Column("tenant_id", sa.Uuid(), sa.ForeignKey("tenant.id"), nullable=False),
```
plus the mirror `ForeignKey("tenant.id")` on `Document.tenant_id` in `db/models.py:397` — or expand the docstring rationale to name what's actually being deferred and to which v0.2.next ticket (FK can land additively later as a constraint-only migration, so the deferral is recoverable, but "consistency with 0002" by itself misreads 0002's argument).

Not blocking the PR by itself; flagging for an explicit decision rather than silent inheritance.

backend/src/meho_backplane/db/models.py
```diff
@@ -122,6 +124,23 @@
...
+#: ``BAAI/bge-small-en-v1.5`` model the EmbeddingService in G0.4-T2
+#: (#259) will load by default. The variant's Python type genuinely
+#: differs by dialect (``list[float]`` on PG via pgvector's bind /
+#: result processors, ``str`` on SQLite via the plain
+#: :class:`Text` type), so the annotation widens to
+#: :class:`TypeEngine[Any]` — the ``Mapped[list[float]]`` annotation
+#: on :attr:`Document.embedding` is the production-PG contract; the
+#: SQLite path inserts a stringified placeholder (see
+#: :mod:`tests.test_db_documents`).
+_PORTABLE_VECTOR_384: TypeEngine[Any] = Text().with_variant(Vector(384), "postgresql")
```
damir-topic: **[⚠️ Major]** — `_PORTABLE_VECTOR_384` falls back to raw `Text` on SQLite, which breaks the `Mapped[list[float]]` contract.

**What:** `Text().with_variant(Vector(384), "postgresql")` works on PG (the pgvector bind/result processors serialize `list[float]` ↔ `vector(384)`), but on SQLite the column type is just `Text`. The annotation on line 404 is `Mapped[list[float]]`, so the type lie surfaces in tests as `# type: ignore[arg-type]` on every `Document(embedding=_PLACEHOLDER_EMBEDDING_TEXT)` call (see `tests/test_db_documents.py:133, 185, 231, 245`). The docstring at lines 134-140 acknowledges this is a "deliberate SQLite-only lie".

**Why it matters:** Once T3 (`index_document`, #260) and T4 (`retrieve`, #261) land, they will write real `list[float]` embeddings through the same ORM. On PG the call works; on the SQLite dev/test path the same call passes a `list[float]` to a `Text` column and SQLAlchemy raises at bind time. Either T3 will ship with its own `# type: ignore` hack or the SQLite test path will be silently disabled — both are regressions on the "SQLite is the dev/test driver" contract this codebase otherwise honors.

CodeRabbit raised the same point with a working sketch.

**Suggested fix:** A small `TypeDecorator` that JSON-serializes on non-PG dialects keeps the `Mapped[list[float]]` annotation honest on both:
```python
import json
from sqlalchemy.types import TypeDecorator

class _PortableVector384(TypeDecorator[list[float]]):
    impl = Text
    cache_ok = True

    def load_dialect_impl(self, dialect):
        if dialect.name == "postgresql":
            return dialect.type_descriptor(Vector(384))
        return dialect.type_descriptor(Text())

    def process_bind_param(self, value, dialect):
        if value is None or dialect.name == "postgresql":
            return value
        return json.dumps(value)

    def process_result_value(self, value, dialect):
        if value is None or dialect.name == "postgresql":
            return value
        return json.loads(value)

_PORTABLE_VECTOR_384: TypeEngine[list[float]] = _PortableVector384()
```
Then drop every `# type: ignore[arg-type]` in `test_db_documents.py` and pass real `list[float]` placeholders. The migration's `embedding_type = sa.Text()` SQLite branch (0003 line 183) keeps its on-disk column shape unchanged; only the ORM bind/result layer gains JSON round-tripping.

If you want to defer this to T3 with the dimensionality-swap migration, that's a defensible call — just commit to it in the docstring and file the follow-up explicitly. Today's shape works for T1's tests because they only ever insert strings.

**Source:** SQLAlchemy 2.x [TypeDecorator docs](https://docs.sqlalchemy.org/en/20/core/custom_types.html#sqlalchemy.types.TypeDecorator).

backend/tests/test_db_documents.py
```diff
@@ -0,0 +1,417 @@
...
+        original_updated_at = (
+            await session.execute(select(Document.updated_at).where(Document.id == doc_id))
+        ).scalar_one()
+
+    # Brief pause so the next ``datetime.now(UTC)`` lands a measurable
+    # delta on systems where the ORM ``onupdate`` callable resolves in
+    # the same microsecond as the insert. SQLite stores datetimes as
+    # ISO-8601 strings with microsecond precision; 10 ms is well above
+    # the resolution floor.
+    time.sleep(0.01)
```
damir-topic: **[⚠️ Major]** — Blocking `time.sleep` in an async test body.

**What:** `time.sleep(0.01)` blocks the running event loop for 10ms. Inside `async def test_document_updated_at_refreshes_on_orm_update`, this is the same loop pytest-asyncio is using to drive the test and the loop the SQLAlchemy async engine is bound to. Copilot's review flagged the same.

**Why it matters:** The block is short, but the idiom propagates — once the autouse engine cache yields tests that genuinely overlap (e.g. when the suite grows to share fixtures across modules running on one loop in `asyncio_mode = "auto"`), `time.sleep` in any one test stalls every other in-flight async task on the same loop and turns intermittent timing tests flaky in CI. The fix is one keyword.

**Suggested fix:**
```diff
- import time
+ import asyncio
  ...
- time.sleep(0.01)
+ await asyncio.sleep(0.01)
```
Also drop the now-unused `import time` (or leave it if anything else still references it — grep says only this call site).

backend/tests/test_db_documents.py
```diff
@@ -0,0 +1,417 @@
...
+        post_commit_updated_at = row.updated_at
+
+    # In-session: ORM ``onupdate`` fires during flush, mutates the
+    # attribute on the instance, commits. With ``expire_on_commit=False``
+    # (see ``meho_backplane.db.engine.get_sessionmaker``) the value
+    # stays accessible. SQLite re-reads timestamps as tz-naive while
+    # the ORM ``onupdate`` lambda mints tz-aware values; normalise both
+    # to naive (.replace(tzinfo=None)) so the comparison is safe across
+    # the SQLite round-trip / fresh-mint boundary.
+    assert post_commit_updated_at.replace(tzinfo=None) >= original_updated_at.replace(
```
damir-topic: **[⚠️ Major]** — `>=` makes the `updated_at` regression check unable to fail.

**What:** Both this assertion and the cross-session re-fetch at line 417 use `>=`. The `time.sleep(0.01)` (line 385) is explicitly there to force a measurable delta. With `>=`, the test still passes when `onupdate` never fires and `updated_at` stays equal to `original_updated_at` — which is the exact regression the test claims (line 357-358) to catch. CodeRabbit flagged the same.

**Why it matters:** The docstring at lines 347-358 says this test exists to catch "a regression where someone drops `onupdate` (e.g. by moving to a PG-side trigger and forgetting to keep the ORM hook for SQLite)". With `>=`, that exact regression passes silently — the `lambda: datetime.now(UTC)` callable could be removed entirely and the test would still go green. The 10ms sleep buys you a clean monotonic gap; `>` honours it.

**Suggested fix:**
```diff
- assert post_commit_updated_at.replace(tzinfo=None) >= original_updated_at.replace(
+ assert post_commit_updated_at.replace(tzinfo=None) > original_updated_at.replace(
      tzinfo=None
  ), (
      "ORM onupdate must advance updated_at on every UPDATE; "
      f"original={original_updated_at} post={post_commit_updated_at}"
  )
```
Same change on line 417, plus an extra equality check binding `refetched` to `post_commit_updated_at` (catches the regression where `onupdate` fires on the instance but isn't written to the DB):
```diff
- assert refetched.replace(tzinfo=None) >= original_updated_at.replace(tzinfo=None)
+ assert refetched.replace(tzinfo=None) == post_commit_updated_at.replace(tzinfo=None)
+ assert refetched.replace(tzinfo=None) > original_updated_at.replace(tzinfo=None)
```

backend/tests/test_db_engine.py
```diff
@@ -402,12 +407,14 @@ def test_alembic_upgrade_head_against_real_pg(
...
-        # Hub `library/*` namespace 1:1 and requires no auth.
-        with PostgresContainer("mirror.gcr.io/library/postgres:16-alpine") as pg:
+        # ``pgvector/pgvector:pg16`` — Postgres 16 + pgvector pre-installed.
+        # Hosted on Docker Hub at ``pgvector/pgvector``; not available on
+        # ``mirror.gcr.io/library/`` because that mirror only covers the
+        # ``library/*`` (official) namespace. Falling back to Docker Hub
+        # here trades a small rate-limit risk for the pgvector capability;
+        # the alternative (custom-built image with the extension layered
+        # on) adds CI maintenance for marginal benefit.
+        with PostgresContainer("pgvector/pgvector:pg16") as pg:
```
damir-topic: **[💡 Minor]** — Docker Hub pull of `pgvector/pgvector:pg16` is rate-limit-exposed.

**What:** Copilot called this out twice (here and in `test_migration_rollback.py:368`). The docstring at lines 410-416 already acknowledges the trade-off explicitly: "Hosted on Docker Hub ... not available on `mirror.gcr.io/library/`. Falling back to Docker Hub here trades a small rate-limit risk for the pgvector capability."

**Why it matters:** v0.2 lives with one image. v0.2.next + shared-runner egress IPs will eventually pull that lever. A two-line env-override would let operators swap the image without re-rolling tests:

**Suggested fix (optional, not blocking):**
```python
import os
_PG_IMAGE = os.environ.get("MEHO_TEST_PGVECTOR_IMAGE", "pgvector/pgvector:pg16")

with PostgresContainer(_PG_IMAGE) as pg:
    ...
```
Same env var would also let the chassis Helm work — once that ships its own pgvector image, CI can point at the same name.

Not blocking the PR; landing the env knob now is cheaper than chasing the first Docker Hub 429 in the CI logs.

damir-topic (changes requested):
## /review-pr-slim summary

**Anchor task:** #258 — G0.4-T1 Alembic migration: pgvector extension + `documents` table + 4 indexes
**Surface area:** backend (Alembic / SQLAlchemy 2.x / pgvector)
**Best-practices packs audited:** python, devops (migration reversibility / dialect portability)
**Verdict:** **REQUEST_CHANGES**

### Acceptance criteria (from #258)

- [x] Migration runs against PG; `CREATE EXTENSION vector` succeeds; table + all 4 indexes created — verified in PG-real testcontainers test `test_documents_table_uses_vector_column_on_postgres`.
- [x] `downgrade -1` removes the table, does NOT drop the extension — verified in `test_migration_0003_upgrade_then_downgrade_is_reversible` + migration body (`downgrade()` deliberately omits `DROP EXTENSION`).
- [x] Round-trip test: insert document with embedding placeholder; query back; assert vector preserved — verified in `test_document_round_trip_persists_every_field` (SQLite placeholder; PG-real shape covered by the testcontainers test).
- [x] Unique `(tenant_id, source, source_id)` enforced — verified in `test_document_unique_tenant_source_id_index_rejects_duplicates`.
- [ ] Cosine similarity query works — **not asserted here**. Defers naturally to T4 (`retrieve` helper); flagging only because the issue body lists it as an AC of T1.
- [ ] FTS query works — **not asserted here**. Same as above.
- [x] Test SQLite fallback: migration runs on SQLite, embedding column is TEXT, PG-only indexes do not land — verified in `test_migration_installs_documents_table_and_portable_indexes` (positive + negative assertions both present).
- [x] `db_migration_probe` returns `ok=True` after upgrade — covered transitively by `test_alembic_probe` reaching head; the 0003 head is reached by every conftest-pre-migrated SQLite engine.
- [x] `pgvector/pgvector:pg16` documented as chassis requirement — referenced in the migration docstring + both testcontainers tests. (Helm chart image swap to land separately with G0.4 deploy gate; out of scope for T1 per Initiative's split.)
- [ ] **CI green** — **BLOCKED** (see below).

### Project-board alignment (project 19 / repo evoila/meho)

|---|---|---|---|---|---|
| #258 (Task G0.4-T1) | In Progress | `Closes #258` (GitHub graph confirms) | ✅ | ❌ partial (CI red; docs gap) | block merge until CI + docs land; close on merge |
| #225 (Initiative G0.4) | Todo | parent | — | T1 of 6 | bump to `In Progress` on merge (this PR breaks first ground for the initiative) |
| #221 (Goal G0) | Todo | grandparent | — | one more substrate task | stays |

No other open PRs touch the retrieval substrate; #266 (`feat(mcp): JSON-RPC dispatch`) and #267 (Connector ABC) are independent.

### Predicted-diff vs actual

- ✅ Migration `0003_create_documents_with_pgvector.py` — present and correct, mirrors `0001` / `0002` dialect discipline.
- ✅ `Document` SQLAlchemy 2.x model on `Base` with `Mapped[...]` annotations.
- ✅ `pgvector>=0.3,<1.0` added to `[project.dependencies]`; `uv.lock` updated.
- ✅ `tests/test_db_documents.py` (new) — round-trip + unique + cross-tenant body_hash + column rename + onupdate.
- ✅ `tests/test_db_models.py` extended — migration-shape positive + negative + reversibility for `0003`.
- ✅ `tests/test_db_engine.py` extended — PG-real `vector(384)` + extension + index probe.
- ✅ `tests/test_migration_rollback.py` — image swap to `pgvector/pgvector:pg16`.
- 🕳️ **MISSING — `docs/codebase/backend.md`** — the durable backend codebase map names `Base / AuditLog / Tenant` (line 275) and lists migrations `0001` + `0002` (lines 277-278). After this PR both lines are stale. Per the in-tree convention this file is updated in lock-step with code changes; doing it as a follow-up has historically rotted (see `backend.md`'s own preamble).
- 🕳️ **MISSING — `tests/integration/conftest.py` image swap** — `backend/tests/integration/conftest.py` (added by merged #268 on main *after* this PR branched) instantiates `PostgresContainer("mirror.gcr.io/library/postgres:16-alpine")`. Rebasing onto current main exposes this; CI is currently red precisely because of it (see Integration verification below).

### Integration verification

All commands run inside `.claude/worktrees/pr-269-review/`:

- [x] `ruff check .` — clean (53 source files).
- [x] `ruff format --check .` — clean.
- [x] `mypy src/` (strict, per `[tool.mypy] files = ["src"]`) — clean across 24 source files.
- [x] `pytest tests/` — 237 passed, 1 failed (pre-existing macOS-specific `test_metrics_endpoint_returns_prometheus_text_format` failure in `test_observability.py`, unrelated to this PR; author called it out in PR body).
- [x] `scripts/ci/check_migration_compat.py` — exit 0.
- [x] PR-touched suites (`test_db_documents.py`, `test_db_models.py`, `test_db_engine.py`, `test_migration_rollback.py`) — 30 passed.
- [⚠️ not run locally] PG-real testcontainers — Docker is unavailable in the review sandbox; the suite is `skipif(not _DOCKER_AVAILABLE)`. The CI runner provisions Docker; we rely on its run.
- [❌] **GitHub CI — Python (ruff + mypy + pytest)** — RED. Root cause: `tests/integration/test_tenant_isolation.py::test_audit_rows_correctly_scoped_per_tenant` runs `alembic upgrade head` against the integration conftest's `postgres:16-alpine` container. Migration `0003` runs `CREATE EXTENSION IF NOT EXISTS vector`, and that image lacks `vector.control`:
  > `sqlalchemy.exc.DBAPIError: ... extension "vector" is not available`
  > `DETAIL: Could not open extension control file "/usr/local/share/postgresql/extension/vector.control"`

  This is **the same gap the PR correctly fixed in `tests/test_db_engine.py` and `tests/test_migration_rollback.py`** — the integration conftest was added by #268 (merged on main 2026-05-12, after PR #269 branched on `ba42859`), so the swap was invisible from the author's branch. Fix: rebase onto current `main` and apply the same `pgvector/pgvector:pg16` swap to `backend/tests/integration/conftest.py:166`.

- [⚠️ flagged] **DCO** — `ACTION_REQUIRED`. Commit needs `Signed-off-by:` trailer. Probably surfacing now because the repo enabled DCO since this commit was authored; sign-off + force-push or new commit on the rebase.
- [⚠️ flagged] **Python License Check** — `cffi:2.0.0` flagged with `UNKNOWN` license. This is a transitive dep regression from a recent `cffi` upstream release (they changed metadata from `MIT` to `UNKNOWN` until they re-publish). Unrelated to this PR; will surface on any PR touching `uv.lock`. Suggest ignoring on this PR but filing a follow-up ticket against the license-check allow-list.
- [⚠️ flagged] **SonarCloud Code Analysis** — Quality Gate failed. Likely "duplicated code blocks" in the test files (the helper-style fixtures repeat the same env-var pinning across modules). Worth a one-line acknowledgement in the next push; not blocking.

### Framework idiomaticity

The pgvector usage is current:
- `Vector(384)` from `pgvector.sqlalchemy` is the documented bind/result-processor adapter on `pgvector-python >= 0.3`.
- `IVFFlat ... vector_cosine_ops WITH (lists = 100)` is correct DDL for ~10k-document corpora (pgvector README's `rows / 1000` rule of thumb; matches Initiative's documented choice).
- `GIN (to_tsvector('english', body))` for BM25 is canonical Postgres FTS shape.
- Raw `op.execute` for the GIN expression and IVFFlat operator-class + `WITH (...)` clauses is the right Alembic escape hatch — `op.create_index` cannot express either cleanly.

SQLAlchemy 2.x `Mapped[...]` + `mapped_column` discipline is correct (matches `AuditLog` / `Tenant`). The `mapped_column("metadata", ...)` column-name override to escape the `DeclarativeBase.metadata` collision is the documented escape hatch.

### Findings

- 🛑 **2 blockers** — CI failing on integration conftest image (rebase + swap), `docs/codebase/backend.md` stale.
- ⚠️ **4 majors** — async `time.sleep` in test (Copilot), portable Vector TypeDecorator (CodeRabbit), `documents.tenant_id` soft-FK rationale (CodeRabbit), strict `>` on `updated_at` regression check (CodeRabbit). See inline comments.
- 💡 **1 minor** — Docker Hub pull of `pgvector/pgvector:pg16` (Copilot, two sites). Author's own docstring acknowledges the trade-off; env-overridable image name is a cheap hedge.
- ❓ **0 questions**.

### Docs

- ❌ `docs/codebase/backend.md` — `Document` model + migration `0003` row both missing from the model/migration tables.

### What was done well

- 👏 **Migration docstring** — 130+ lines of rationale (extension lifecycle, IVFFlat empty-table caveat with REINDEX runbook breadcrumb, why GIN+IVFFlat are NOT on `__table_args__`, HNSW deferral hook). This is the kind of context that pays back at the next on-call rotation.
- 👏 **Test discipline** — the explicit "PG-only indexes MUST NOT land on SQLite" *negative* assertion in `test_migration_installs_documents_table_and_portable_indexes` is exactly the trap-the-leak-early pattern that keeps `__table_args__` clean.
- 👏 **Numbering note in the migration docstring** — the issue body said `0004 revising 0003`; the actual sequence is `0003`. Calling that drift out in the docstring instead of silently following the issue is the right move.
- 👏 **Reversibility test** — the upgrade → downgrade → re-upgrade cycle on SQLite proves the migration round-trips before it ever hits a real PG.

---

## Post-merge actions (do after merge)

### Auto-close (should happen automatically)
- #258 — `Closes #258` is present + same-repo + main base; GitHub auto-close will fire.

### Manual / verification
- Verify #258 actually closed (GitHub auto-close occasionally misses on squash-merge with a stripped message).
- Post completion comment on #258 summarising what landed + linking the PG-real test results.
- Bump #225 (Initiative G0.4) status from `Todo` → `In Progress` (this PR is the first of six tasks under it).

### Follow-up tickets to file
- Helm chart Postgres image swap to `pgvector/pgvector:pg16` (or equivalent) — Initiative body lists it as a chassis-image bump; not part of T1's diff but needs to land before any real cluster runs `alembic upgrade head` against migration `0003`.
- `cffi:2.0.0` license metadata regression — surfaced here, not introduced here; goes against the license-check allow-list, not this PR.

ikaric:
## Review feedback addressed — pushed in `28f6c1b`

Rebased onto current `main` (picks up #268's integration conftest) and addressed every finding from @damir-topic, CodeRabbit, and Copilot. Commit is signed off for DCO.

### Blockers

- **CI red on integration suite** — `backend/tests/integration/conftest.py` (added by #268 after this branch forked) was still pinning `mirror.gcr.io/library/postgres:16-alpine`, so migration `0003`'s `CREATE EXTENSION vector` aborted every test in `test_tenant_isolation.py`. Swapped to `pgvector/pgvector:pg16` (same image as the other testcontainers tests), plus folded the per-test `TRUNCATE TABLE audit_log; TRUNCATE TABLE tenant` into a single `TRUNCATE TABLE audit_log, documents, tenant` so the new FK from `documents.tenant_id → tenant.id` doesn't reject the truncate.
- **`docs/codebase/backend.md` stale** — extended the `Base / AuditLog / Tenant` row to include `Document` (with the FK + TypeDecorator notes) and added a full `0003_create_documents_with_pgvector` row to the migrations table.

### Majors

- **`documents.tenant_id` is now a real FK** (migration + ORM). The audit_log soft-FK rationale didn't transfer — `documents` is a brand-new table with no chassis-era rows and a downgrade that drops the whole table, so there's no backfill or cascade decision to defer. Updated both docstrings to call that out explicitly so the next reader doesn't re-litigate it.
- **`_PORTABLE_VECTOR_384` is now a `TypeDecorator`**. The old `Text().with_variant(Vector(384), "postgresql")` broke the `Mapped[list[float]]` contract on SQLite, which is why every test call site carried `# type: ignore[arg-type]` and the round-trip used a stringified placeholder. New `_PortableVector384(TypeDecorator)` JSON-encodes on bind / JSON-decodes on result for non-PG dialects, delegates to `Vector(384)` on PG via `with_variant`. Test call sites now pass real `list[float]` and assert it round-trips; every `# type: ignore[arg-type]` is gone.
- **`time.sleep(0.01)` → `await asyncio.sleep(0.01)`** in `test_document_updated_at_refreshes_on_orm_update` so the test doesn't stall the pytest-asyncio loop.
- **`updated_at` regression check is now strict** — both `>=` comparisons tightened to `>`, plus added the cross-session `refetched == post_commit_updated_at` equality check so the regression where `onupdate` fires on the instance but isn't written to the DB also surfaces.

### Minor + nitpick

- **`MEHO_TEST_PGVECTOR_IMAGE` env override** on `test_db_engine.py`, `test_migration_rollback.py`, and the integration conftest — operators can point at a GHCR / internal-mirror cache without re-rolling the code on the first Docker Hub 429.
- **PG-real index assertions now check `pg_indexes.indexdef` shape**, not just names — `USING gin (to_tsvector('english', body))` and `USING ivfflat ... vector_cosine_ops ... WITH (lists='100')`. A regression that keeps the index name but swaps GIN for btree (or drops the IVFFlat operator class) now fails the test.

### Validation

- `ruff check .` clean, `ruff format --check .` clean.
- `mypy --strict` clean across 27 source files (`_PortableVector384` is typed `TypeDecorator[list[float]]`).
- `scripts/ci/check_migration_compat.py` exit 0.
- `pytest tests/` — 273 passed, 1 pre-existing macOS-specific `test_observability` failure unchanged.
- `pytest tests/integration/` — 6 passed (the bug that blocked CI is fixed).

### Out of scope follow-ups (filed separately)

- Helm chart Postgres image bump to `pgvector/pgvector:pg16` — Initiative scope per @damir-topic's review, not T1's diff.
- `cffi:2.0.0` `UNKNOWN`-license metadata regression — surfaces on any PR touching `uv.lock`; goes against the license-check allow-list, not this PR.

cc @damir-topic @coderabbitai — should be ready for re-review.

backend/src/meho_backplane/db/models.py
```diff
@@ -122,6 +125,67 @@
...
+    ) -> list[float] | None:
+        if value is None:
+            return None
+        decoded = json.loads(value)
+        return [float(x) for x in decoded]
+
+
+_PORTABLE_VECTOR_384: TypeEngine[list[float]] = _PortableVector384().with_variant(
+    Vector(384), "postgresql"
+)
```
damir-topic, line 186: **[👏 Praise]** — Clean resolution of the prior review's `Mapped[list[float]]` contract break.

The new `_PortableVector384(TypeDecorator)` keeps the `Mapped[list[float]]` annotation honest on both dialects: on PG the `.with_variant(Vector(384), "postgresql")` factory delegates to pgvector's native bind/result processors; on SQLite the decorator's `process_bind_param` / `process_result_value` round-trip the list through JSON. Test call sites now pass real `list[float]` and assert it round-trips (`_PLACEHOLDER_EMBEDDING: list[float] = [0.0] * 384` at line 100 of `test_db_documents.py`), and every `# type: ignore[arg-type]` is gone. `cache_ok = True` is correctly declared because the encoding is parameter-free.

This is also the right boundary: dimension validation belongs at the embedding-service entrypoint (T2 #259), not in the DB type — keeping the decorator focused on dialect portability rather than domain invariants is the cleaner separation. (CodeRabbit's follow-up suggestion to add a 384-length check inside `process_bind_param` would duplicate what pgvector's adapter already enforces on PG and would push domain validation into the persistence layer — I'd recommend leaving that to the embedding service.)

damir-topic (commented):
## /review-pr-slim follow-up review

**Anchor task:** #258 — G0.4-T1 Alembic migration: pgvector extension + `documents` table + 4 indexes
**Surface area:** backend (Alembic / SQLAlchemy 2.x / pgvector)
**Best-practices packs audited:** python, devops (migration reversibility / dialect portability)
**Verdict:** **COMMENT** — every prior blocker addressed, one doc-drift major lingers (details below).
**Reviewed SHA:** `28f6c1b` (current head).

### Re-verification of the prior review's blockers + majors

| Prior finding | Severity | Status on `28f6c1b` |
| --- | --- | --- |
| CI red on `tests/integration/conftest.py:166` (`postgres:16-alpine` lacks `vector.control`) | 🛑 Blocker | ✅ Fixed — line 176 now `pgvector/pgvector:pg16`; CI `Python (ruff + mypy + pytest)` = SUCCESS. Also extended the per-test TRUNCATE to `audit_log, documents, tenant` (non-cascading multi-table TRUNCATE, required by the new FK). |
| `docs/codebase/backend.md` missing `Document` model + `0003` migration rows | 🛑 Blocker | ✅ Partially fixed — `Document` and `0003` rows added at lines 307 + 311. ⚠️ However, three sibling rows in the same doc still claim suites use `postgres:16-alpine` after this PR swapped them to `pgvector/pgvector:pg16`. See "Outstanding ⚠️ Major" below. |
| `documents.tenant_id` soft-FK rationale doesn't transfer from `audit_log` | ⚠️ Major | ✅ Fixed — real `sa.ForeignKey("tenant.id")` in both `0003_create_documents_with_pgvector.py:212` and `db/models.py:452`; docstring rationale rewritten to call out *why* the audit_log soft-FK argument doesn't apply here (no chassis-era rows, downgrade drops the whole table). |
| `_PORTABLE_VECTOR_384` falls back to raw `Text` on SQLite, breaks `Mapped[list[float]]` | ⚠️ Major | ✅ Fixed — `_PortableVector384(TypeDecorator)` at `db/models.py:145-181`. `with_variant(Vector(384), "postgresql")` delegates to pgvector's native processors on PG; `process_bind_param` / `process_result_value` JSON-encode on SQLite. Every `# type: ignore[arg-type]` is gone from the test call sites; `_PLACEHOLDER_EMBEDDING` is a real `list[float]`. |
| Blocking `time.sleep` in async test | ⚠️ Major | ✅ Fixed — `await asyncio.sleep(0.01)` at `test_db_documents.py:391`. (The Copilot thread at line 392 isn't auto-marked outdated by GitHub because the line number didn't shift; it's stale and safe to resolve.) |
| `>=` makes `updated_at` regression check unable to fail | ⚠️ Major | ✅ Fixed — strict `>` at `test_db_documents.py:412` + `:429`, plus the new cross-session `refetched == post_commit_updated_at` equality check at `:428` that catches the "ORM `onupdate` fires on the instance but isn't written to the DB" regression. |
| Docker Hub pull of `pgvector/pgvector:pg16` is rate-limit-exposed | 💡 Minor | ✅ Fixed — `MEHO_TEST_PGVECTOR_IMAGE` env override consistently applied in `test_db_engine.py:420`, `test_migration_rollback.py:371`, `tests/integration/conftest.py:176`. |

The PR-real index-shape assertions in `test_db_engine.py:530-580` (asserting on `pg_indexes.indexdef`, not just names — `using gin (to_tsvector('english', body))`, `using ivfflat (... vector_cosine_ops) WITH (lists='100')`) also addressed CodeRabbit's first-pass nitpick about index-content vs. index-name verification. Worth calling out — that's the kind of structural assertion that survives a future drift where someone keeps the name but swaps GIN for btree.

### Project-board alignment (project 19 / repo evoila/meho)

|---|---|---|---|---|---|
| #258 (Task G0.4-T1) | In Progress | `Closes #258` | ✅ | ✅ full (subject to the docs minor below) | close on merge |
| #225 (Initiative G0.4) | Todo | parent of #258 | — | T1 of 6 | bump to `In Progress` on merge (first task under it) |
| #221 (Goal G0) | Todo | grandparent | — | one more substrate task | stays |

No other open PRs touch the retrieval substrate; #266 (`feat(mcp)`) and #267 (Connector ABC) remain independent.

### Outstanding ⚠️ Major — `docs/codebase/backend.md` drift on three sibling rows

The PR correctly adds the `Document` row at line 307 and the `0003_create_documents_with_pgvector` row at line 311, but three sibling rows in the same doc still describe suites using `postgres:16-alpine` after this PR swapped all three of them to `pgvector/pgvector:pg16`. CodeRabbit's second-pass flagged lines 328 + 402; line 190 drifted in lock-step on the same change.

**What:**
- **`docs/codebase/backend.md:190`** — _"The test spins up `postgres:16-alpine` via testcontainers"_ describing `tests/test_migration_rollback.py`, which `test_migration_rollback.py:371` now pins to `pgvector/pgvector:pg16` (env-overridable).
- **`docs/codebase/backend.md:328`** — _"`tests/integration/test_tenant_isolation.py` ... Boots `postgres:16-alpine`"_, which is now `pgvector/pgvector:pg16` (`backend/tests/integration/conftest.py:176`).
- **`docs/codebase/backend.md:402`** — _"(dev) `testcontainers` ... Spins up `postgres:16-alpine` for the end-to-end test in `tests/test_db_engine.py::TestPostgresIntegration`"_, which `test_db_engine.py:420` now pins to `pgvector/pgvector:pg16`. Also misses that two more testcontainer suites came online since this row was last written (`test_migration_rollback.py` + `tests/integration/test_tenant_isolation.py`).

(Lines 311 / `test_db_engine.py:394` / `test_migration_rollback.py:365` / `conftest.py:167` / `0003_create_documents_with_pgvector.py:18` all mention `postgres:16-alpine` correctly — they're the historical-context narrative, not claims about what suites use today.)

**Why it matters:** Same blast-radius reasoning as the prior review's `Base / AuditLog / Tenant` row gap — `docs/codebase/backend.md`'s own preamble says "stale entries are bugs", and a future agent debugging "why did CI 429 against Docker Hub for these PG suites" reads the dep-table row at line 402 and reaches for the wrong image. Three sites all drifted in lock-step with the same image-swap commit, so it's a single mechanical sweep.

**Suggested fix:** One commit, three edits.

```diff
@@ docs/codebase/backend.md L190
-  The test spins up `postgres:16-alpine` via testcontainers, runs
+  The test spins up `pgvector/pgvector:pg16` via testcontainers, runs
```

```diff
@@ docs/codebase/backend.md L328
-G0.1-T6 (Task #236) broad-spectrum end-to-end test for the tenancy chain. Boots `postgres:16-alpine` via `testcontainers` (module-scoped fixture in `tests/integration/conftest.py`),
+G0.1-T6 (Task #236) broad-spectrum end-to-end test for the tenancy chain. Boots `pgvector/pgvector:pg16` via `testcontainers` (module-scoped fixture in `tests/integration/conftest.py`; image name env-overridable via `MEHO_TEST_PGVECTOR_IMAGE`),
```

```diff
@@ docs/codebase/backend.md L402
-| (dev) `testcontainers` | ≥ 4.0 | Spins up `postgres:16-alpine` for the end-to-end test in `tests/test_db_engine.py::TestPostgresIntegration`. Skipped gracefully when the Docker socket is absent — the SQLite-async coverage stays always-on. |
+| (dev) `testcontainers` | ≥ 4.0 | Spins up `pgvector/pgvector:pg16` for the testcontainer suites (`tests/test_db_engine.py::TestPostgresIntegration`, `tests/test_migration_rollback.py`, `tests/integration/test_tenant_isolation.py`); image name overridable via `MEHO_TEST_PGVECTOR_IMAGE`. Skipped gracefully when the Docker socket is absent — the SQLite-async coverage stays always-on. |
```

### Outstanding 💡 Minor — module docstrings out of step with the image swap

Three module-level docstrings still describe the old fixture shape. The code is correct; only the prose drifted. Bundleable into the same cleanup commit as the `backend.md` sweep.

- **`backend/tests/integration/conftest.py:21`** — _"boots a single `postgres:16-alpine` container"_; line 176 actually pins `pgvector/pgvector:pg16`. Same docstring at line 28 also says _"the per-test `audit_log` truncation in `integration_app`"_, but the TRUNCATE at line 277 now covers `audit_log, documents, tenant` (multi-table non-cascading because of the new FK). CodeRabbit flagged this and auto-marked it "✅ Addressed in commit 28f6c1b", but the commit did not actually touch lines 20-29; the bot's auto-resolution was incorrect.
- **`backend/tests/test_migration_rollback.py:26`** — step 1 of the walk-through still says _"Spin up `postgres:16-alpine` via `testcontainers`"_; the actual fixture at line 371 pins `pgvector/pgvector:pg16` (env-overridable). The fail-mode explanation downstream at lines 362-370 names the right image — this is just a missed sweep in the higher-up walk-through.
- **`backend/tests/integration/test_tenant_isolation.py:51`** — _"The integration suite uses `testcontainers` + `postgres:16-alpine` (same image pin as `tests.test_migration_rollback` and `tests.test_db_engine`)"_ — both image claim and the parenthetical-parity claim drift on this PR's image swap; all three suites now pin `pgvector/pgvector:pg16`.

### Integration verification

All commands run inside the `pr-269-review-v2` worktree at SHA `28f6c1b`:

- [x] `ruff check .` — clean (62 files, 27 source).
- [x] `ruff format --check .` — clean.
- [x] `mypy` strict (per `[tool.mypy] files = ["src"]`) — Success: no issues in 27 source files.
- [x] `scripts/ci/check_migration_compat.py` — exit 0 (deny-list-clean migration tree).
- [x] `pytest tests/` — 273 passed, 1 deselected pre-existing macOS-specific `test_metrics_endpoint_returns_prometheus_text_format` failure (Linux-only `process_resident_memory_bytes` from the prometheus_client process collector; same gap the prior review and PR body called out — unrelated to this PR).
- [⚠️ not run locally] PG-real testcontainers — Docker is unavailable in the review sandbox. GitHub CI ran them: `Python (ruff + mypy + pytest)` workflow returned SUCCESS at 14:59:33Z, so the `pgvector/pgvector:pg16`-driven integration + rollback + db_engine PG suites all passed on the runner.
- [⚠️ flagged, not blocking] `Python License Check` — RED. `cffi:2.0.0` flagged `UNKNOWN` (upstream metadata regression; surfaces on any PR touching `uv.lock`). Same as previous review; goes against the license-check allow-list, not this PR.

### Framework idiomaticity

- `_PortableVector384` TypeDecorator shape matches the documented SQLAlchemy 2.x pattern (`impl = Text`, `cache_ok = True`, `process_bind_param` / `process_result_value` for serialisation, `with_variant(Vector(384), "postgresql")` to delegate to pgvector's native processors on PG only). Authoritative reference: https://docs.sqlalchemy.org/en/20/core/custom_types.html#sqlalchemy.types.TypeDecorator.
- `sa.ForeignKey("tenant.id")` in both the migration and the ORM model is the canonical 2.x shape; no `relationship()` declared, which is correct given the model deliberately ships with no helper methods (read/write paths land in `meho_backplane.retrieval` via T2-T5).
- Multi-table `TRUNCATE TABLE audit_log, documents, tenant` (non-cascading) at `conftest.py:277` is the right PG idiom for clearing related tables when one of them is FK-referenced — listing every referenced table explicitly avoids `CASCADE` while still letting PG accept the statement.

### Findings on `28f6c1b`

- 🛑 **0 blockers**.
- ⚠️ **1 major** — `docs/codebase/backend.md` drift on lines 190, 328, 402 (postgres:16-alpine→pgvector swap not swept through). Details + diffs above.
- 💡 **3 minors** — module docstrings in `tests/integration/conftest.py` (L21+L28), `tests/test_migration_rollback.py` (L26), `tests/integration/test_tenant_isolation.py` (L51). Details above.
- ❓ **0 questions**.

### Notable bot-state corrections

- The CodeRabbit follow-up thread on `tests/integration/conftest.py` line 177 was auto-marked **"✅ Addressed in commit 28f6c1b"** but the commit did not actually touch the module docstring at lines 20-29; the bot's auto-resolution was incorrect. Filed as one of the minors above.
- The Copilot thread on `tests/test_db_documents.py` line 392 (the `time.sleep` finding) is **not** marked outdated by GitHub because the line number stayed the same, but the underlying code at line 391 is now `await asyncio.sleep(0.01)` — the finding is genuinely addressed; the thread is safe to resolve manually.

### What was done well

- 👏 **Doc rationale, not just code** — the `Document` docstring at `db/models.py:345-358` doesn't just *show* the FK, it spells out *why* the audit_log soft-FK argument doesn't apply here (no chassis-era rows, downgrade drops the whole table). That's the kind of explanation that survives the next refactor.
- 👏 **TypeDecorator at the right boundary** — `_PortableVector384` keeps dialect portability in the DB type and leaves domain validation (384-dim) to the embedding service. CodeRabbit's follow-up suggested adding a length check inside `process_bind_param` — I'd push back on that: it would duplicate what pgvector's adapter already enforces on PG and push domain validation into the persistence layer.
- 👏 **Cross-session equality check on `updated_at`** — the `refetched == post_commit_updated_at` line at `test_db_documents.py:428` catches the specific regression where `onupdate` fires on the instance but isn't written to the DB. That's exactly what `>=` would have let through.
- 👏 **Index-definition shape assertions on PG-real** — checking `pg_indexes.indexdef` for `using gin`, `to_tsvector('english'`, `using ivfflat`, `vector_cosine_ops`, `lists='100'` rather than just index names. A regression that keeps the name but swaps the index type now fails the test loudly.
- 👏 **Non-cascading multi-table TRUNCATE** — `TRUNCATE TABLE audit_log, documents, tenant` rather than `TRUNCATE TABLE tenant CASCADE` is the right idiom for inter-test reset under the new FK. Comment at `conftest.py:266-276` explains the why; future readers will not need to re-derive the choice.

---

## Post-merge actions (do after merge)

### Auto-close (should happen automatically)
- #258 — `Closes #258` is present + same-repo + main base; GitHub auto-close will fire.

### Manual / verification
- Verify #258 actually closed (GitHub auto-close occasionally misses on squash-merge with a stripped message).
- Post completion comment on #258 summarising what landed + linking the PG-real CI run.
- Bump #225 (Initiative G0.4) status from `Todo` → `In Progress` (this PR is the first of six tasks under it).

### Bot thread hygiene (optional, low priority)
- Resolve the Copilot thread on `test_db_documents.py:392` (genuinely addressed at line 391; GitHub's outdated heuristic didn't fire).
- Resolve the Copilot threads on `test_db_engine.py` + `test_migration_rollback.py` re: Docker Hub rate-limit hardening — the `MEHO_TEST_PGVECTOR_IMAGE` env override in this PR is the suggested mitigation.

### Follow-up tickets to file (already noted in prior review; re-listed for completeness)
- Helm chart Postgres image swap to `pgvector/pgvector:pg16` — Initiative body lists it as a chassis-image bump.
- `cffi:2.0.0` license metadata regression — surfaces on any PR touching `uv.lock`; goes against the license-check allow-list, not this PR.

damir-topic (approved):
## /review-pr-slim approval — doc-drift sweep landed

Pushed `88da5af` directly to this branch to land the doc-drift sweep called out in the COMMENT review at [#pullrequestreview-4274004015](https://github.com/evoila/meho/pull/269#pullrequestreview-4274004015) (5 sites, prose only; ruff + format clean):

- `docs/codebase/backend.md` lines 190, 328, 402 — the three sibling rows that still described suites as using `postgres:16-alpine`.
- `backend/tests/integration/conftest.py` lines 20-29 — module docstring image name + the `audit_log, documents, tenant` truncation list.
- `backend/tests/test_migration_rollback.py` line 26 — step 1 of the module-docstring walk-through.
- `backend/tests/integration/test_tenant_isolation.py` line 51 — "Why real Postgres" rationale block.
- `docs/acceptance/rollback.md` line 36 — found in the sweep; same drift pattern, unit-test layer description.

Every remaining `postgres:16-alpine` mention is now historical-context narrative (correctly noting why the chassis image fails `CREATE EXTENSION vector`), not a claim about what suites use today.

Approving so the prior CHANGES_REQUESTED clears. Merging once Python CI on `88da5af` goes green; `Python License Check` will stay red on the unrelated `cffi:2.0.0 UNKNOWN`-metadata regression and gets admin-overridden at merge time (already triaged as out-of-scope in the prior review).

PR #290, opened 2026-05-12
feat(mcp): OAuth 2.1 resource-server protection on /mcp (G0.5-T2)

### LINKED ISSUE #247: Task: G0.5-T2 OAuth 2.1 resource-server — /.well-known/oauth-protected-resource + WWW-Authenticate header + audience validation per RFC 9728/8707
Parent initiative: #226

## Target

OAuth 2.1 resource-server pattern for `/mcp`, per RFC 9728 + RFC 8707:

- `backend/src/meho_backplane/mcp/auth.py` — MCP auth dependency that reuses the chassis `verify_jwt` chain but adds audience validation against the MCP canonical URI.
- `backend/src/meho_backplane/main.py` — mount `/.well-known/oauth-protected-resource` (PR metadata endpoint).
- `backend/src/meho_backplane/api/well_known.py` — new module hosting the PR metadata route.
- `backend/src/meho_backplane/settings.py` — add `MCP_RESOURCE_URI` setting; defaults to `<BACKPLANE_URL>/mcp`.
- `backend/src/meho_backplane/mcp/server.py` — wire the auth dependency on the `/mcp` POST handler; emit RFC-9728-compliant WWW-Authenticate header on 401.
- `backend/tests/test_mcp_auth.py` — new tests for the OAuth-RS chain.
- `backend/tests/test_well_known.py` — tests for PR metadata.

## Context

The MCP 2025-06-18 spec ([Authorization](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization)) defines a strict OAuth 2.1 resource-server pattern. Implementing it correctly is what makes Claude Desktop (and other MCP clients) trust MEHO without manual config:

1. **Client sends MCP request without token** → server returns 401 + `WWW-Authenticate: Bearer resource_metadata="<url>"`.
2. **Client fetches PR metadata** at `/.well-known/oauth-protected-resource` → discovers the authorization server URL.
3. **Client follows OAuth 2.1 with PKCE + the `resource` parameter** against Keycloak → gets an access token whose `aud` claim includes the MCP canonical URI.
4. **Client sends MCP request with Bearer token** → server validates `aud` matches the MCP URI → processes.

The chassis already validates Keycloak JWTs (against the chassis's audience), so T2 mostly **extends** the existing validation chain. Two extensions:

- **Audience validation against `MCP_RESOURCE_URI`** (not the chassis's `KEYCLOAK_AUDIENCE`). The MCP resource URI is what an MCP client puts in the `resource` parameter when requesting a token; the JWT's `aud` claim must include it. The chassis's audience claim is for HTTP API access — distinct surfaces, distinct audiences.
- **WWW-Authenticate header on 401.** Standard RFC 9728 §5.1 shape.

**PR metadata document.** RFC 9728 §3 defines the JSON shape:

```json
{
  "resource": "https://meho.evba.lab/mcp",
  "authorization_servers": ["https://keycloak.evba.lab/realms/evba"],
  "scopes_supported": ["mcp:read", "mcp:execute"],
  "bearer_methods_supported": ["header"]
}
```

The endpoint is served unauthenticated (it's discovery metadata; OAuth-RS spec requires this).

**Settings additions:**

```python
class Settings(BaseSettings):
    # ... existing ...
    mcp_resource_uri: str = ""  # MCP_RESOURCE_URI; defaults to f"{backplane_url}/mcp" if empty
    backplane_url: str = ""     # BACKPLANE_URL; used for default constructions
```

The default behavior: if `MCP_RESOURCE_URI` is empty, the route constructs it from the backplane's URL. Operators with non-default deployments override.

## Current state

After T1:
- `/mcp` route accepts requests without auth.
- Chassis JWT validation exists in `auth/jwt.py` for HTTP API routes; not wired to MCP.

## Desired state

`backend/src/meho_backplane/mcp/auth.py`:

```python

import structlog
from fastapi import Depends, HTTPException, Header

from meho_backplane.auth.jwt import _decode_and_validate_jwt  # extract from chassis
from meho_backplane.auth.operator import Operator
from meho_backplane.settings import get_settings

def _mcp_resource_uri() -> str:
    settings = get_settings()
    if settings.mcp_resource_uri:
        return settings.mcp_resource_uri
    return f"{settings.backplane_url.rstrip('/')}/mcp"


def _www_authenticate_header() -> str:
    settings = get_settings()
    return (
        f'Bearer resource_metadata="{settings.backplane_url.rstrip("/")}'
        f'/.well-known/oauth-protected-resource"'
    )


async def verify_mcp_jwt(authorization: str | None = Header(default=None)) -> Operator:
    if authorization is None or not authorization.lower().startswith("bearer "):
        raise HTTPException(
            status_code=401,
            detail="missing or malformed Authorization header",
            headers={"WWW-Authenticate": _www_authenticate_header()},
        )
    token = authorization[7:]
    claims = await _decode_and_validate_jwt(
        token,
        expected_audience=_mcp_resource_uri(),  # NOTE: audience for MCP, not the HTTP API
    )
    # Reuse Operator construction from chassis with tenant claims (G0.1):
    return Operator(
        sub=claims["sub"],
        name=claims.get("name"),
        email=claims.get("email"),
        raw_jwt=token,
        tenant_id=claims.get("tenant_id"),
        tenant_role=claims.get("tenant_role"),
    )
```

PR metadata route:

```python
# backend/src/meho_backplane/api/well_known.py

from fastapi import APIRouter
from meho_backplane.settings import get_settings

router = APIRouter(prefix="/.well-known", tags=["discovery"])


@router.get("/oauth-protected-resource")
async def protected_resource_metadata():
    settings = get_settings()
    return {
        "resource": settings.mcp_resource_uri or f"{settings.backplane_url}/mcp",
        "authorization_servers": [str(settings.keycloak_issuer_url).rstrip("/")],
        "scopes_supported": ["mcp:read", "mcp:execute"],
        "bearer_methods_supported": ["header"],
    }
```

`server.py` wires the dependency:

```python
@router.post("", dependencies=[Depends(verify_mcp_jwt)])
async def mcp_dispatch(request: Request, operator: Operator = Depends(verify_mcp_jwt)):
    # ... rest unchanged ...
```

## Acceptance criteria

- [ ] `/.well-known/oauth-protected-resource` returns RFC-9728-compliant JSON; tested.
- [ ] Unauthenticated POST `/mcp` returns 401 with `WWW-Authenticate: Bearer resource_metadata="..."` header per RFC 9728 §5.1; tested with assertion on the header value.
- [ ] A JWT issued for the chassis HTTP API surface (audience = `KEYCLOAK_AUDIENCE`, which differs from `MCP_RESOURCE_URI`) is rejected at `/mcp` with 401. Test asserts.
- [ ] A JWT issued with `aud` containing the MCP resource URI is accepted; `verify_mcp_jwt` returns the `Operator`.
- [ ] Malformed Authorization header (e.g., `Basic ...`) returns 401.
- [ ] Expired token returns 401.
- [ ] Settings expose `MCP_RESOURCE_URI` + `BACKPLANE_URL`; defaults documented.

## Out of scope

- Dynamic Client Registration (RFC 7591). v0.2 ships static registration; operators register MCP clients in Keycloak manually.
- Token introspection — v0.2 validates JWTs locally (Keycloak signed them); introspection is unnecessary.
- Scope-based RBAC inside MCP — v0.2 enforces RBAC via the existing tenant_role from G0.1 (T3/T4 wire this); scope-as-claim is a future polish.
- Token revocation cache — not in v0.2; revoked tokens stay valid until expiry.

## References

- Parent Initiative: [#226](https://github.com/evoila/meho/issues/226).
- Depends on: T1 (the /mcp route).
- Soft dependency on G0.1-T2 (`Operator.tenant_id` extraction) — T2 reuses the same JWT decoder, so if G0.1-T2 lands first the tenant_id appears automatically; if G0.1-T2 lands later, T2 can ship and tenant_id stays None on the Operator until G0.1-T2 lands and the patch flows through.
- MCP spec 2025-06-18 Authorization: https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization
- RFC 9728 (Protected Resource Metadata): https://datatracker.ietf.org/doc/html/rfc9728
- RFC 8707 (Resource Indicators): https://www.rfc-editor.org/rfc/rfc8707.html
- OAuth 2.1: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-13
- Predecessor surfaces: [backend/src/meho_backplane/auth/jwt.py:349-368](backend/src/meho_backplane/auth/jwt.py#L349-L368) (`verify_jwt`), [backend/src/meho_backplane/api/v1/auth_config.py](backend/src/meho_backplane/api/v1/auth_config.py) (the existing public discovery surface for the CLI).
- Python best-practices: `.claude/skills/explore-and-refactor/python_best_practices.md` — FastAPI dependency injection, Pydantic validation.
- DevOps best-practices: `.claude/skills/implement-issue/devops_best_practices.md` — OAuth security best practices, HTTPS-everywhere.

zdamir:
## Summary

- Layers spec-conforming **OAuth 2.1 resource-server** protection on `POST /mcp` per MCP 2025-06-18 §Authorization (RFC 9728 + RFC 8707). T1's dispatcher gains a `Depends(verify_mcp_jwt_and_bind)` security gate that runs before envelope parsing.
- New `mcp/auth.py` validates Bearer tokens against the **MCP canonical URI** (not the chassis `KEYCLOAK_AUDIENCE`) via the new `verify_jwt_for_audience` seam in `auth/jwt.py`. RFC 8707 §2 audience binding: a token issued for the HTTP API surface cannot be replayed at `/mcp` and vice versa.
- On 401 the wrapper attaches the RFC 9728 §5.1 `WWW-Authenticate: Bearer resource_metadata="<url>"` header so spec-conforming MCP clients can discover the metadata document.
- New unauthenticated GET `/.well-known/oauth-protected-resource` (`api/well_known.py`) returns the RFC 9728 §3 metadata document — `resource`, `authorization_servers` (Keycloak issuer), `scopes_supported`, `bearer_methods_supported`.
- Two new env vars (`BACKPLANE_URL`, `MCP_RESOURCE_URI`); both default to empty (fail-closed for MCP without breaking chassis-only deployments).

## How the audience binding works

```
Client (no token) ──POST /mcp──►  Server
                                  401 + WWW-Authenticate: Bearer resource_metadata="https://meho.test/.well-known/oauth-protected-resource"

Client ──GET /.well-known/oauth-protected-resource──►  Server
                                                       { "resource": "https://meho.test/mcp",
                                                         "authorization_servers": ["https://keycloak.test/realms/meho"],
                                                         "scopes_supported": ["mcp:read", "mcp:execute"],
                                                         "bearer_methods_supported": ["header"] }

Client ──OAuth 2.1 + PKCE w/ resource=https://meho.test/mcp──►  Keycloak
                                                                ►  access_token (aud = "https://meho.test/mcp")

Client ──POST /mcp + Authorization: Bearer <token>──►  Server
                                                       ►  validates aud == MCP canonical URI
                                                       ►  200 + JSON-RPC envelope
```

A token whose `aud` is `meho-backplane` (the chassis audience) is **rejected** at `/mcp` even with a valid signature — that's the RFC 8707 §2 confused-deputy defence.

## Test plan

- [x] `ruff format --check` — 61 files formatted
- [x] `mypy src/` (strict) — 29 source files, no issues
- [x] `pytest tests/` — **276 passed**, 1 deselected (pre-existing macOS-only flake from T1 era)

### AC coverage

- [x] `/.well-known/oauth-protected-resource` returns RFC 9728-compliant JSON — `test_well_known_returns_required_rfc9728_fields`.
- [x] Unauthenticated POST `/mcp` → 401 + `WWW-Authenticate: Bearer resource_metadata="..."` per RFC 9728 §5.1 — `test_missing_authorization_returns_401_with_www_authenticate` + `test_www_authenticate_header_has_rfc9728_shape`.
- [x] JWT with chassis audience is rejected at `/mcp` with 401 — `test_token_for_chassis_audience_is_rejected_at_mcp`.
- [x] JWT with `aud == MCP_RESOURCE_URI` accepted; `verify_mcp_jwt` returns `Operator` — `test_token_with_mcp_audience_is_accepted`.
- [x] Malformed `Authorization` (Basic …) → 401 — `test_basic_auth_scheme_returns_401_with_www_authenticate`.
- [x] Expired token → 401 — `test_expired_token_returns_401`.
- [x] Settings expose `MCP_RESOURCE_URI` + `BACKPLANE_URL`; defaults documented — `test_well_known_honors_explicit_mcp_resource_uri_override` + `test_explicit_mcp_resource_uri_is_the_required_audience`.
- [x] CI green: ruff + mypy + pytest.

### Bonus coverage (beyond AC)

- [x] WWW-Authenticate header shape exact match (RFC 9728 §5.1 example) — `test_www_authenticate_header_has_rfc9728_shape`.
- [x] JWKS unreachable → 401 (not 5xx), still carries WWW-Authenticate — `test_jwks_unreachable_returns_401`.
- [x] Trailing-slash stripping on `BACKPLANE_URL` and Keycloak issuer.
- [x] Discovery endpoint is unauthenticated — `test_well_known_does_not_require_authentication`.

### Dispatcher tests preserved

- [x] All 25 JSON-RPC dispatch tests from T1 still pass. The `client` fixture now overrides `verify_mcp_jwt_and_bind` with a stub returning a fixture `Operator`, so each dispatcher test stays independent of the full OAuth-RS chain (which is covered separately in `test_mcp_auth.py`).

## Out of scope

- **Origin-header validation** (MCP transport DNS-rebinding warning) — needs `MCP_ALLOWED_ORIGINS` infrastructure that doesn't exist yet. Belongs with a transport-hardening cluster (Origin + body-size cap + content-type 415) as a fast-follow.
- **`MCP_ENABLED` boolean flag** (Initiative #226 DoD line) — T2 doesn't need it for auth.
- **Token revocation cache, Dynamic Client Registration (RFC 7591), scope-based RBAC inside MCP** — all explicit "Out of scope" items on #247.
- **G0.5-T3 (#248)** — tool + resource registries (`register_mcp_tool`, `register_mcp_resource`).
- **G0.5-T4 (#249)** — reference `meho.status` tool + `meho://tenant/<id>/info` resource.
- **G0.5-T5 (#250)** — MCP-specific audit integration on `tools/call` + `resources/read`. T2 currently writes a chassis `audit_log` row per `/mcp` request via the existing `AuditMiddleware` because the new dependency binds `operator_sub`; T5 will replace this with the fail-closed MCP-specific path.
- **G0.5-T6 (#251)** — end-to-end MCP Inspector acceptance test + `docs/architecture/mcp.md` + `docs/cross-repo/mcp-client-setup.md`.

Closes #247

## Summary by CodeRabbit

* **New Features**
  * MCP endpoints now enforce OAuth 2.1 Bearer-token protection.
  * Added RFC 9728 OAuth protected-resource discovery at /.well-known/oauth-protected-resource.

* **Security**
  * Enforced JWT audience validation for MCP access.
  * Returns RFC 9728-style WWW-Authenticate challenges on 401s.

* **Tests**
  * Comprehensive end-to-end tests for MCP auth and discovery added.

* **Documentation**
  * Backend docs updated with OAuth/resource-server and auth behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/290)

---

## Follow-up (auto-implement-issue, iteration 2, 2026-05-12)

Addressed review feedback from CodeRabbit's inline nitpick and @ikaric's review body.

**Fixed:**

- **CR-A** `2c83a19` — Normalize explicit `MCP_RESOURCE_URI` (strip whitespace + rstrip slash) before returning from `mcp_resource_uri()`. A stray trailing slash or newline in the env var no longer causes opaque `aud` mismatches.
- **R-2** `7b629ce` — Empty-audience guard in `verify_jwt_for_audience`. The fail-closed property is now local to the verifier (`audience_not_configured` 401) rather than dependent on the issuer never minting a token with `aud=""`. New test `test_empty_audience_setting_returns_401` asserts a token with `aud=""` still 401s even when the derived audience is also `""`.
- **R-1** `9c3c6eb` — Use the previously-unused `_log` to emit `mcp_auth_failed` in the 401 branch of `verify_mcp_jwt`, carrying the chassis failure-mode token (`missing_token` / `invalid_token` / `jwks_unavailable` / etc.) as `detail`. Gives on-call a grep-able event to distinguish MCP-side 401 spikes from chassis-side ones.
- **CR-B + R-4** `1f928e0` — Strengthen `test_well_known_keycloak_issuer_url_has_no_trailing_slash` to actually exercise the strip: set `KEYCLOAK_ISSUER_URL` **with** a trailing slash and assert exact-equality against the canonical (no-slash) form. The previous shape (no trailing slash in input + `not endswith("/")` assertion) passed trivially.
- **R-5** `7a3db92` — Add `test_token_for_arbitrary_audience_is_rejected_at_mcp` pinning the value-based audience-check contract: a token minted for an arbitrary third party (not just the chassis audience) also 401s.

**Disputed:**

- **R-3** (`del operator` style smell) — Keeping the parameter form `operator: Operator = Depends(verify_mcp_jwt_and_bind)` over the `dependencies=[Depends(...)]` form for forward compat with T3 (#248). T3's `tools/call` handler will read tenant / role data off the `Operator` parameter; flipping to the parameter form *now* lets T3 add behaviour without touching the route signature. The reviewer already framed this as "defensible forward-compat", so keeping as-is and accepting the minor `del operator` cleanup as the cost of that compat. The `del` makes the unused-by-T2 intent explicit; mypy strict otherwise flags the parameter as unused.

**DCO compliance:**

The first commit on this branch (`d2e9c64`, the T2 main feature commit) lacks the `Signed-off-by:` trailer that the DCO bot requires. Hard rules in `/auto-implement-issue` (Phase 8: "Never `--amend`. Never force-push.") block the standard `git rebase --signoff main && git push --force-with-lease` remediation. Every commit landed in this iteration (`2c83a19`, `7b629ce`, `9c3c6eb`, `1f928e0`, `7a3db92`) is `git commit -s` signed, so DCO compliance is in place from this commit forward. The pre-existing commit needs either:
1. An `--admin` merge bypass (precedent: PR #181 / G2.7-T6, 2026-05-11 — same incident).
2. A manual rebase + force-push outside this skill's guardrails.

Flagging for human-merge decision.

**Verification (re-run after iter-2 fixes):**

- `ruff check src/ tests/` — clean
- `ruff format --check` — 61 files formatted
- `mypy src/` (strict) — 29 source files, no issues
- `pytest tests/` — **278 passed**, 1 deselected (pre-existing macOS flake)
- MCP-focused tests grew to **39 cases** (25 dispatcher + 8 auth chain + 6 well-known) from 37 in iteration 1.

---

## Follow-up (auto-implement-issue, iteration 3, 2026-05-12)

PR now **APPROVED** by @damir-topic (review `4273271769`). One non-blocking inline finding addressed by dispute; no code changes this iteration.

**Disputed (with inline reply):**

- **Thread on `mcp/auth.py:176`** (`async` without `await`) — Both CodeRabbit and @damir-topic flagged that `verify_mcp_jwt_and_bind` is `async def` with no `await` in the body. The reviewer explicitly noted "I'd lean toward keeping it async on consistency grounds — the two are used as a pair and reading them with matching shapes is worth the few ns. Author's call; not blocking." Keeping `async` for shape consistency with **both** the sibling `verify_mcp_jwt` (which IS async because it awaits `verify_jwt_for_audience`) and the chassis parallel `verify_jwt_and_bind` in `middleware.py` (also `async def` + only calls `bind_contextvars`). Coroutine-creation overhead is a few ns per request; the cohort cost (consistency, grep-ability of the dependency family) beats the per-call cost. If the chassis sibling ever flips to plain `def`, this flips with it — same commit, same rationale. Inline reply posted + thread resolved.

**Thread state:** all 4 threads resolved. `reviewDecision = APPROVED`.

**DCO status (carried over from iter-2 note):** the first commit `d2e9c64` still lacks `Signed-off-by`. Every commit from iter-2 onward is `-s` signed. Per Hard rule 8 (no `--amend`, no force-push) the standard `rebase --signoff` remediation is blocked from inside the skill. Options remain: `--admin` merge bypass (precedent: PR #181 / G2.7-T6) or a manual force-push outside the skill's guardrails.

### DIFF
--- a/backend/src/meho_backplane/auth/jwt.py
+++ b/backend/src/meho_backplane/auth/jwt.py
@@ -100,6 +100,7 @@ __all__ = [
     "clear_jwks_cache",
     "keycloak_readiness_probe",
     "verify_jwt",
+    "verify_jwt_for_audience",
 ]
 
 #: HTTP timeout for both the OIDC discovery hit and the JWKS hit. Keep
@@ -211,12 +212,23 @@ async def _fetch_jwks(*, force_refresh: bool = False) -> dict[str, Any]:
         return jwks
 
 
-def _decode_with_jwks(token: str, jwks: dict[str, Any], settings: Settings) -> Any:
+def _decode_with_jwks(
+    token: str,
+    jwks: dict[str, Any],
+    settings: Settings,
+    *,
+    expected_audience: str,
+) -> Any:
     """Verify *token* against *jwks* and return the validated claims.
 
     Wrapped behind the ``warnings.catch_warnings`` block to suppress the
     authlib-jose deprecation noise on every request — the deprecation is
     a v0.2 migration item, not a per-request signal.
+
+    ``expected_audience`` is passed by the caller so MCP routes (G0.5-T2)
+    can validate against the MCP canonical URI rather than the chassis
+    ``KEYCLOAK_AUDIENCE``. The chassis ``verify_jwt`` passes
+    ``settings.keycloak_audience`` to preserve existing behaviour.
     """
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
@@ -232,7 +244,7 @@ def _decode_with_jwks(token: str, jwks: dict[str, Any], settings: Settings) -> A
                 },
                 "aud": {
                     "essential": True,
-                    "value": settings.keycloak_audience,
+                    "value": expected_audience,
                 },
             },
         )
@@ -286,7 +298,12 @@ def _extract_bearer_token(authorization: str | None) -> str:
     return token
 
 
-async def _decode_with_kid_rotation(token: str, settings: Settings) -> Any:
+async def _decode_with_kid_rotation(
+    token: str,
+    settings: Settings,
+    *,
+    expected_audience: str,
+) -> Any:
     """Decode *token* against the cached JWKS, refreshing once on a kid miss.
 
     The first ``_fetch_jwks`` call serves the cached keyset (or fetches
@@ -301,6 +318,10 @@ async def _decode_with_kid_rotation(token: str, settings: Settings) -> Any:
     The retry budget is exactly one — a second ``ValueError`` after the
     forced JWKS refresh is treated as a hard 401, preventing an
     infinite-refresh loop on a token whose ``kid`` truly does not exist.
+
+    ``expected_audience`` is forwarded to :func:`_decode_with_jwks` so
+    callers (chassis ``verify_jwt`` vs MCP ``verify_mcp_jwt``) can
+    enforce different audiences against the same JWKS + issuer pair.
     """
     try:
         jwks = await _fetch_jwks()
@@ -311,7 +332,7 @@ async def _decode_with_kid_rotation(token: str, settings: Settings) -> Any:
         raise _http_401("jwks_unavailable") from None
 
     try:
-        return _decode_with_jwks(token, jwks, settings)
+        return _decode_with_jwks(token, jwks, settings, expected_audience=expected_audience)
     except ValueError:
         # Treat *any* ValueError as a kid-miss signal — the message
         # ("Key not found") is authlib-internal and not part of any
@@ -327,7 +348,7 @@ async def _decode_with_kid_rotation(token: str, settings: Settings) -> Any:
         raise _http_401("jwks_unavailable") from None
 
     try:
-        return _decode_with_jwks(token, jwks, settings)
+        return _decode_with_jwks(token, jwks, settings, expected_audience=expected_audience)
     except _DECODE_ERRORS_WITH_VALUEERROR as retry_exc:
         raise _http_401("invalid_token") from retry_exc
 
@@ -432,6 +453,39 @@ def _operator_from_claims(claims: Any, raw_jwt: str, settings: Settings) -> Oper
         raise _http_401("invalid_token") from exc
 
 
+async def verify_jwt_for_audience(
+    authorization: str | None,
+    *,
+    expected_audience: str,
+) -> Operator:
+    """Validate a Bearer token against an explicit ``aud`` claim value.
+
+    The full JWT-validation chain (Bearer extraction → JWKS fetch +
+    kid-rotation retry → signature/claims/structure validation →
+    Operator projection) parametrised by audience. This is the public
+    seam the chassis :func:`verify_jwt` dependency uses with
+    ``settings.keycloak_audience``; MCP routes (G0.5-T2) use the same
+    seam with the MCP canonical URI per RFC 8707 §2 / RFC 9728 §7.4
+    audience-binding semantics. Keeping both surfaces on a single chain
+    avoids drift between "what the chassis validates" and "what MCP
+    validates" — issuer / kid-rotation / signature handling stays
+    identical; only the audience differs.
+
+    Raises 401 on every failure mode the chassis chain raises.
+    Failures-with-audience-mismatch surface as ``invalid_token``
+    because that's how authlib's ``InvalidClaimError`` maps in the
+    centralised handler.
+    """
+    token = _extract_bearer_token(authorization)
+    settings = get_settings()
+    claims = await _decode_with_kid_rotation(
+        token,
+        settings,
+        expected_audience=expected_audience,
+    )
+    return _operator_from_claims(claims, token, settings)
+
+
 async def verify_jwt(authorization: str | None = Header(default=None)) -> Operator:
     """FastAPI dependency: validate the Bearer token and return an Operator.
 
@@ -447,10 +501,11 @@ async def verify_jwt(authorization: str | None = Header(default=None)) -> Operat
     cache is refreshed exactly once and the verify is retried. A
     second miss is a hard 401 ``invalid_token``.
     """
-    token = _extract_bearer_token(authorization)
     settings = get_settings()
-    claims = await _decode_with_kid_rotation(token, settings)
-    return _operator_from_claims(claims, token, settings)
+    return await verify_jwt_for_audience(
+        authorization,
+        expected_audience=settings.keycloak_audience,
+    )
 
 
 async def keycloak_readiness_probe() -> ProbeResult:
--- a/backend/src/meho_backplane/auth/jwt.py
+++ b/backend/src/meho_backplane/auth/jwt.py
@@ -475,7 +475,19 @@ async def verify_jwt_for_audience(
     Failures-with-audience-mismatch surface as ``invalid_token``
     because that's how authlib's ``InvalidClaimError`` maps in the
     centralised handler.
+
+    Defence in depth: an empty or whitespace-only ``expected_audience``
+    short-circuits to ``audience_not_configured`` 401. The MCP route
+    derives the audience from ``MCP_RESOURCE_URI`` / ``BACKPLANE_URL``
+    and returns an empty string when neither is set; without this
+    guard the fail-closed property would rely on the *issuer* never
+    emitting a token with an empty ``aud`` claim — a true assumption
+    against any real Keycloak deployment but one external invariant
+    away from a bypass. Failing the check locally to the verifier
+    makes the property auditable in one place.
     """
+    if not expected_audience or not expected_audience.strip():
+        raise _http_401("audience_not_configured")
     token = _extract_bearer_token(authorization)
     settings = get_settings()
     claims = await _decode_with_kid_rotation(
--- a/backend/src/meho_backplane/mcp/auth.py
+++ b/backend/src/meho_backplane/mcp/auth.py
@@ -1,0 +1,179 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""OAuth 2.1 resource-server chain for the ``/mcp`` route.
+
+This module is the MCP-side of the auth boundary: it consumes the same
+Keycloak JWTs the chassis accepts but enforces a **different audience**
+— the canonical MCP resource URI per RFC 8707 §2 — and emits the
+RFC 9728 §5.1 ``WWW-Authenticate`` header on 401 so spec-conforming
+MCP clients can discover the protected-resource metadata document and
+follow the OAuth 2.1 + PKCE handshake. The chassis ``verify_jwt`` keeps
+its existing ``KEYCLOAK_AUDIENCE`` audience for HTTP API routes; MCP
+gets its own audience binding so a token issued for the HTTP API can't
+be replayed at the MCP endpoint and vice-versa.
+
+Reused chassis infrastructure
+=============================
+
+* :func:`~meho_backplane.auth.jwt.verify_jwt_for_audience` — the
+  parameterised seam (signature + claims + kid-rotation + Operator
+  projection). MCP simply passes its own ``expected_audience``.
+* :class:`~meho_backplane.auth.operator.Operator` — same identity
+  shape (sub / name / email / tenant_id / tenant_role / raw_jwt). MCP
+  routes receive the operator by injection just like chassis routes.
+* :func:`~meho_backplane.middleware.verify_jwt_and_bind` is **not**
+  reused: this module ships its own :func:`verify_mcp_jwt_and_bind`
+  that mirrors the binding contract (operator_sub + tenant_id into
+  structlog contextvars) but layers WWW-Authenticate header injection
+  on the 401 path.
+
+Spec citations
+==============
+
+* MCP 2025-06-18 §Authorization —
+  https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization
+  ("MCP servers MUST validate that tokens presented to them were
+  specifically issued for them as the intended audience, according to
+  RFC 8707 §2. Invalid or expired tokens MUST receive a HTTP 401
+  response.")
+* RFC 9728 §5.1 — Bearer-scheme WWW-Authenticate response containing
+  the ``resource_metadata`` parameter pointing at the protected-
+  resource metadata document.
+* RFC 8707 — Resource Indicators. The ``resource`` parameter sent at
+  the OAuth-AS during authorisation must equal the canonical MCP URI;
+  the issued token's ``aud`` claim then binds to that resource.
+"""
+
+from __future__ import annotations
+
+import structlog
+from fastapi import Depends, Header, HTTPException
+
+from meho_backplane.auth.jwt import verify_jwt_for_audience
+from meho_backplane.auth.operator import Operator
+from meho_backplane.settings import Settings, get_settings
+
+__all__ = [
+    "mcp_resource_uri",
+    "verify_mcp_jwt",
+    "verify_mcp_jwt_and_bind",
+    "www_authenticate_header",
+]
+
+_log = structlog.get_logger(__name__)
+
+
+def mcp_resource_uri(settings: Settings | None = None) -> str:
+    """Return the canonical MCP server URI for audience binding.
+
+    Resolves in priority order:
+
+    1. ``Settings.mcp_resource_uri`` when explicitly set — operators
+       with non-default MCP mount points (e.g. ``/api/mcp``) override
+       per environment.
+    2. ``f"{settings.backplane_url}/mcp"`` derived at use time.
+    3. Empty string when neither is set — fail-closed for MCP: no JWT
+       can have an empty audience claim, so every ``/mcp`` request 401s
+       until an operator wires ``BACKPLANE_URL`` (or ``MCP_RESOURCE_URI``).
+
+    Per MCP 2025-06-18 §Resource Parameter Implementation, the canonical
+    form is lowercase scheme + host with no trailing slash; the helper
+    strips any trailing slash on ``backplane_url`` before concatenating
+    to enforce that convention.
+    """
+    settings = settings or get_settings()
+    if settings.mcp_resource_uri:
+        return settings.mcp_resource_uri
+    if not settings.backplane_url:
+        return ""
+    return f"{settings.backplane_url.rstrip('/')}/mcp"
+
+
+def www_authenticate_header(settings: Settings | None = None) -> str:
+    """Build the RFC 9728 §5.1 ``WWW-Authenticate`` header value.
+
+    Shape per spec example::
+
+        Bearer resource_metadata="https://meho.example/.well-known/oauth-protected-resource"
+
+    The metadata URL is rooted at ``backplane_url`` because RFC 9728
+    §3 hosts the document at ``/.well-known/oauth-protected-resource``
+    *on the resource server's origin*, not on the authorisation
+    server. When ``backplane_url`` is empty (chassis-only deployment),
+    the helper returns the bare ``Bearer`` challenge with no
+    ``resource_metadata`` parameter — a degraded but spec-legal
+    response that doesn't leak a malformed URL.
+    """
+    settings = settings or get_settings()
+    base = settings.backplane_url.rstrip("/")
+    if not base:
+        return "Bearer"
+    metadata_url = f"{base}/.well-known/oauth-protected-resource"
+    return f'Bearer resource_metadata="{metadata_url}"'
+
+
+async def verify_mcp_jwt(
+    authorization: str | None = Header(default=None),
+) -> Operator:
+    """FastAPI dependency: validate the Bearer token for the MCP audience.
+
+    Reuses :func:`~meho_backplane.auth.jwt.verify_jwt_for_audience` with
+    :func:`mcp_resource_uri` as the ``expected_audience``. The chassis
+    chain raises :class:`fastapi.HTTPException` with status 401 on
+    every failure mode (missing header, malformed Authorization, JWKS
+    unreachable, signature mismatch, expired token, audience mismatch,
+    issuer mismatch, malformed JWT, malformed tenant claims). On any of
+    those, this wrapper re-raises the same exception with the
+    RFC 9728 §5.1 ``WWW-Authenticate`` header attached so MCP clients
+    can discover the resource-metadata URL and walk the OAuth-RS
+    discovery flow.
+
+    A 401 is preserved verbatim — same detail token, same status code.
+    Non-401 status codes from the chassis chain (none today, reserved
+    for future cases) propagate without the header injection.
+    """
+    settings = get_settings()
+    expected_audience = mcp_resource_uri(settings)
+    try:
+        return await verify_jwt_for_audience(
+            authorization,
+            expected_audience=expected_audience,
+        )
+    except HTTPException as exc:
+        if exc.status_code != 401:
+            raise
+        headers = dict(exc.headers or {})
+        headers["WWW-Authenticate"] = www_authenticate_header(settings)
+        raise HTTPException(
+            status_code=exc.status_code,
+            detail=exc.detail,
+            headers=headers,
+        ) from exc
+
+
+async def verify_mcp_jwt_and_bind(
+    operator: Operator = Depends(verify_mcp_jwt),
+) -> Operator:
+    """Validate the Bearer token *and* bind operator identity into contextvars.
+
+    Mirrors the chassis :func:`~meho_backplane.middleware.verify_jwt_and_bind`
+    pattern: every authenticated MCP route handler declares
+    ``Depends(verify_mcp_jwt_and_bind)`` so the
+    ``operator_sub`` and ``tenant_id`` slots end up in structlog's
+    contextvars before the handler runs. The chassis
+    :class:`~meho_backplane.audit.AuditMiddleware` reads ``operator_sub``
+    to decide whether to write an audit row; without this binding the
+    middleware would skip the row and the operator-action trail would
+    have an `/mcp` shaped hole. G0.5-T5 (#250) layers MCP-specific
+    audit semantics on top of this implicit chassis trail.
+
+    ``tenant_id`` is bound as ``str(operator.tenant_id)`` for the same
+    JSON-renderer reason the chassis wrapper documents — see
+    :func:`~meho_backplane.middleware.verify_jwt_and_bind`.
+    """
+    structlog.contextvars.bind_contextvars(
+        operator_sub=operator.sub,
+        tenant_id=str(operator.tenant_id),
+    )
+    return operator
--- a/backend/src/meho_backplane/mcp/auth.py
+++ b/backend/src/meho_backplane/mcp/auth.py
@@ -154,6 +154,16 @@ async def verify_mcp_jwt(
     except HTTPException as exc:
         if exc.status_code != 401:
             raise
+        # Structured-log the MCP-side auth failure with a grep-able event
+        # name. The chassis `verify_jwt` deliberately doesn't log auth
+        # failures (the 401 surface is already on every chassis route);
+        # /mcp is a new surface where operators want to distinguish MCP
+        # 401 spikes from chassis 401 spikes. ``detail`` is the chassis
+        # chain's failure-mode token (`missing_token` / `invalid_token`
+        # / `jwks_unavailable` / `missing_tenant_claim` / etc.) — never
+        # an operator-controllable string, so logging it verbatim is
+        # safe (see auth/jwt.py docstring).
+        _log.warning("mcp_auth_failed", detail=exc.detail)
         headers = dict(exc.headers or {})
         headers["WWW-Authenticate"] = www_authenticate_header(settings)
         raise HTTPException(
--- a/backend/src/meho_backplane/mcp/auth.py
+++ b/backend/src/meho_backplane/mcp/auth.py
@@ -71,11 +71,22 @@ def mcp_resource_uri(settings: Settings | None = None) -> str:
 
     1. ``Settings.mcp_resource_uri`` when explicitly set — operators
        with non-default MCP mount points (e.g. ``/api/mcp``) override
-       per environment.
+       per environment. The value is normalised (``.strip()`` +
+       ``.rstrip('/')``) so an operator who pastes the URI from a
+       browser bar with a trailing slash, or leaves a stray newline in
+       the env var, doesn't cause an ``aud`` mismatch against tokens
+       whose ``resource`` parameter Keycloak emitted in canonical
+       form. MCP 2025-06-18 §"Canonical Server URI" mandates the no-
+       trailing-slash form for interop, so the normalisation also
+       enforces the spec contract.
     2. ``f"{settings.backplane_url}/mcp"`` derived at use time.
     3. Empty string when neither is set — fail-closed for MCP: no JWT
        can have an empty audience claim, so every ``/mcp`` request 401s
        until an operator wires ``BACKPLANE_URL`` (or ``MCP_RESOURCE_URI``).
+       Defence in depth: :func:`~meho_backplane.auth.jwt.verify_jwt_for_audience`
+       short-circuits an empty expected_audience to a 401 explicitly so
+       the fail-closed property is local to the verifier and doesn't
+       depend on any honest-issuer assumption.
 
     Per MCP 2025-06-18 §Resource Parameter Implementation, the canonical
     form is lowercase scheme + host with no trailing slash; the helper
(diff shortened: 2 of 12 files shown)


zdamir:
@coderabbitai review

ikaric (commented):
## Review — feat(mcp): OAuth 2.1 RS on /mcp (#247)

Solid PR. The audience-parameterization of the chassis JWT chain is the right shape — `verify_jwt_for_audience(authorization, *, expected_audience)` keeps signature/issuer/kid-rotation/Operator-projection identical and only the audience varies, so chassis and MCP can never drift on what "a valid token" means structurally. RFC 9728 + RFC 8707 compliance reads cleanly against the specs; tests cover the AC matrix end-to-end (including JWKS-unreachable graceful 401, header shape, audience confusion in both explicit-URI and derived-default modes). CI green.

Below are non-blocking observations and a couple of small suggestions.

### Suggestions (non-blocking)

**1. `_log = structlog.get_logger(__name__)` in `mcp/auth.py` is unused.** Either drop it, or — more useful — emit an `mcp_audience_mismatch` / `mcp_auth_failed` event in the `except HTTPException` branch of `verify_mcp_jwt` so on-call can distinguish MCP-side 401 spikes from chassis-side ones in the structured-log feed. The chassis `verify_jwt` deliberately doesn't log auth failures, but the MCP entrypoint is operationally a new surface that's worth its own grep-able event name.

**2. Defense in depth on the empty-audience path.** When `BACKPLANE_URL=""` and `MCP_RESOURCE_URI=""`, `mcp_resource_uri()` returns `""` and gets passed as `expected_audience` to authlib. The fail-closed claim relies on "no real JWT has an empty `aud` claim" — which is true under any honest issuer but is one external invariant away from a bypass. A one-line guard in `verify_jwt_for_audience`:

```python
if not expected_audience:
    raise _http_401("audience_not_configured")
```

makes the fail-closed property local to the verifier rather than dependent on the issuer's hygiene. Tiny change; meaningful belt-and-suspenders.

**3. The `del operator` in `mcp_dispatch`** is unusual stylistically. FastAPI's idiom for "run a dependency only for its side effects" is:

```python
@router.post("", dependencies=[Depends(verify_mcp_jwt_and_bind)])
async def mcp_dispatch(request: Request) -> Response:
    ...
```

That said — keeping the parameter form means T3 won't have to change the signature when it starts reading tenant/role off the operator. So the current shape is defensible forward-compat; just flagging that `del operator` is a smell that goes away with the `dependencies=[...]` form if you want it gone now.

**4. `test_well_known_keycloak_issuer_url_has_no_trailing_slash` doesn't actually exercise the strip.** The fixture sets `KEYCLOAK_ISSUER_URL="https://keycloak.test/realms/meho"` — already no trailing slash. The assertion `not server.endswith("/")` passes trivially without ever invoking `rstrip("/")`. To actually cover the regression case, set the env var to `https://keycloak.test/realms/meho/` (with trailing slash) and assert it's stripped in the response. (The matching `test_well_known_strips_trailing_slash_on_backplane_url` does this correctly for `BACKPLANE_URL`; this one mirrors the name but not the behavior.)

**5. Coverage nit: third-audience rejection.** Tests cover `aud=KEYCLOAK_AUDIENCE → 401` and `aud=MCP_RESOURCE_URI → 200`, but not `aud=<arbitrary-string> → 401`. authlib's audience check is value-based so the third case is implied, but an explicit test would document the contract that *any* non-matching audience is rejected, not just the chassis one.

### Process

**DCO is failing.** Commits need `Signed-off-by:` lines. Easiest path:
```
git rebase --signoff main
git push --force-with-lease
```

### Bottom line

LGTM modulo (1) and (4); (2) and (5) are worth picking up; (3) is style. None are blocking — happy to see this land.

backend/src/meho_backplane/mcp/auth.py
```diff
@@ -0,0 +1,179 @@
...
+       until an operator wires ``BACKPLANE_URL`` (or ``MCP_RESOURCE_URI``).
+
+    Per MCP 2025-06-18 §Resource Parameter Implementation, the canonical
+    form is lowercase scheme + host with no trailing slash; the helper
+    strips any trailing slash on ``backplane_url`` before concatenating
+    to enforce that convention.
+    """
+    settings = settings or get_settings()
+    if settings.mcp_resource_uri:
+        return settings.mcp_resource_uri
```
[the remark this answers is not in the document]
zdamir: Fixed in commit `2c83a19`.

`mcp_resource_uri()` now applies `.strip().rstrip("/")` to `settings.mcp_resource_uri` before returning. The derived-default path (`{backplane_url}/mcp`) already stripped a trailing slash on `backplane_url`; this brings the explicit-override path to the same shape. The normalisation also enforces the MCP 2025-06-18 §"Canonical Server URI" no-trailing-slash interop rule on operator-supplied input.

No new test needed for the normalisation specifically — the behaviour is observable through the existing `test_explicit_mcp_resource_uri_is_the_required_audience` happy path plus the sibling `test_well_known_strips_trailing_slash_on_backplane_url` and the strengthened `test_well_known_keycloak_issuer_url_has_no_trailing_slash` (also fixed this iteration per the second nit).

zdamir:
@coderabbitai review

backend/src/meho_backplane/auth/jwt.py
```diff
@@ -432,6 +453,51 @@ def _operator_from_claims(claims: Any, raw_jwt: str, settings: Settings) -> Oper
...
+    if not expected_audience or not expected_audience.strip():
+        raise _http_401("audience_not_configured")
+    token = _extract_bearer_token(authorization)
+    settings = get_settings()
+    claims = await _decode_with_kid_rotation(
+        token,
+        settings,
+        expected_audience=expected_audience,
+    )
+    return _operator_from_claims(claims, token, settings)
```
damir-topic, line 498: **👏 Praise** — the audience-parameterised seam is the right refactor

**What:** `verify_jwt_for_audience` extracts the full chassis chain (Bearer extraction → JWKS fetch + kid-rotation → signature/claims/structure → `Operator` projection) and parameterises *only* the audience. The chassis `verify_jwt` is rewired to pass `settings.keycloak_audience`; the MCP wrapper passes `mcp_resource_uri()`.

**Why it matters:** Chassis and MCP can't drift on what "a valid token" means structurally — issuer / kid-rotation / signature / `Operator` shape stay identical. Only the audience varies. The alternative (parallel chains) would have been a regression magnet every time the JWKS or kid-rotation logic shifts.

**Source:** RFC 8707 §2 audience binding + MCP 2025-06-18 §"Token Handling" both require the audience check; nothing else in the validation chain needs to vary between chassis and MCP, so the seam is exactly the right granularity.

backend/src/meho_backplane/auth/jwt.py
```diff
@@ -432,6 +453,51 @@ def _operator_from_claims(claims: Any, raw_jwt: str, settings: Settings) -> Oper
...
+    derives the audience from ``MCP_RESOURCE_URI`` / ``BACKPLANE_URL``
+    and returns an empty string when neither is set; without this
+    guard the fail-closed property would rely on the *issuer* never
+    emitting a token with an empty ``aud`` claim — a true assumption
+    against any real Keycloak deployment but one external invariant
+    away from a bypass. Failing the check locally to the verifier
+    makes the property auditable in one place.
+    """
+    if not expected_audience or not expected_audience.strip():
+        raise _http_401("audience_not_configured")
```
damir-topic, line 490: **👏 Praise** — empty-audience guard makes fail-closed local to the verifier

**What:** Refusing `expected_audience=""` / whitespace-only with `audience_not_configured` 401 *before* invoking authlib.

**Why it matters:** Without this, the fail-closed claim on `BACKPLANE_URL="" + MCP_RESOURCE_URI=""` would rely on no real Keycloak ever minting a token with `aud=""` — true under any honest issuer but one external invariant away from a bypass. Now it's auditable in one place at the verifier. `test_empty_audience_setting_returns_401` pins it: even a token whose `aud=""` matches the empty derived audience is still rejected because the verifier short-circuits before authlib runs.

backend/src/meho_backplane/mcp/auth.py
```diff
@@ -0,0 +1,200 @@
...
+        headers = dict(exc.headers or {})
+        headers["WWW-Authenticate"] = www_authenticate_header(settings)
+        raise HTTPException(
+            status_code=exc.status_code,
+            detail=exc.detail,
+            headers=headers,
+        ) from exc
+
+
+async def verify_mcp_jwt_and_bind(
```
damir-topic, line 176: **💡 Minor** — `async` without `await`

**What:** This function calls only `structlog.contextvars.bind_contextvars()` (sync) and returns; there's no `await` expression in the body.

**Why it matters:** Marginal coroutine-creation overhead per request. CodeRabbit's latest review surfaced the same nit and flagged it `💤 Low value`.

**Suggested fix:** Either drop the `async` keyword (and let FastAPI run it on the executor as a sync dep), or keep it for shape consistency with the sibling `verify_mcp_jwt` (which IS async because it awaits `verify_jwt_for_audience`) and dismiss this. I'd lean toward keeping it async on consistency grounds — the two are used as a pair and reading them with matching shapes is worth the few ns. Author's call; not blocking.

**Source:** PEP 492 / FastAPI deps run sync defs on a threadpool, async defs as coroutines; neither is wrong here.

damir-topic (approved):
## /review-pr-slim summary

**Anchor task:** #247 — G0.5-T2 OAuth 2.1 resource-server on `/mcp`
**Surface area:** backend (FastAPI / Pydantic v2 / authlib.jose / structlog)
**Best-practices packs audited:** python (full)
**Verdict:** **APPROVE**

### Acceptance criteria
- [x] `/.well-known/oauth-protected-resource` returns RFC 9728-compliant JSON — verified by `test_well_known_returns_required_rfc9728_fields` against §3 field set (`resource`, `authorization_servers`, `scopes_supported`, `bearer_methods_supported`).
- [x] Unauthenticated POST `/mcp` → 401 + `WWW-Authenticate: Bearer resource_metadata="..."` per RFC 9728 §5.1 — `test_missing_authorization_returns_401_with_www_authenticate` + `test_www_authenticate_header_has_rfc9728_shape` (exact-equality, not substring).
- [x] JWT with chassis `aud` rejected at `/mcp` with 401 — `test_token_for_chassis_audience_is_rejected_at_mcp`; confused-deputy defence per RFC 8707 §2.
- [x] JWT with `aud == MCP_RESOURCE_URI` accepted; `verify_mcp_jwt` returns `Operator` — `test_token_with_mcp_audience_is_accepted` round-trips a full JSON-RPC `ping`.
- [x] Malformed `Authorization` (`Basic …`) → 401 — `test_basic_auth_scheme_returns_401_with_www_authenticate`.
- [x] Expired token → 401 — `test_expired_token_returns_401` (10s past `exp`, outside 30s leeway).
- [x] `Settings` exposes `MCP_RESOURCE_URI` + `BACKPLANE_URL`; defaults documented in `settings.py:175-198`.
- [x] CI green: ruff + ruff format + mypy + pytest.

### Project-board alignment (MEHO, project 19)
|---|---|---|---|---|---|
| #247 (T2) | In Progress | `Closes #247` | ✅ (default branch + same repo + closing keyword) | ✅ full | bump → `In Review` now; → `Done` on merge |
| #226 (Initiative G0.5) | Todo | parent of #247 | — | ⚠️ partial (one of ~6 tasks) | bump → `In Progress` (T1 already merged + T2 lands here) |
| #221 (Goal G0) | Todo | grandparent | — | n/a | leave |

No delivered-but-not-linked issues. T3/T4/T5/T6 (#248/#249/#250/#251) are explicitly disclaimed out of scope in the PR body, which is faithful to #247.

### Predicted vs. actual diff
Walked the blast radius cold: derived what *should* change to add OAuth-RS on `/mcp` (new `mcp/auth.py`, well-known route, settings additions, audience-parameterised seam in `auth/jwt.py`, dispatcher wiring, four test files, doc update). Actual diff matches one-for-one.

- ✅ `mcp/auth.py` — new; correct shape.
- ✅ `auth/jwt.py` — `verify_jwt_for_audience` seam extracted; chassis `verify_jwt` rewired through it.
- ✅ `mcp/server.py` — `Depends(verify_mcp_jwt_and_bind)` on `mcp_dispatch`.
- ✅ `api/well_known.py` — new RFC 9728 metadata route.
- ✅ `settings.py` — `backplane_url` + `mcp_resource_uri` with documented defaults.
- ✅ `main.py` — `well_known_router` mounted.
- ✅ `mcp/__init__.py` — re-exports.
- ✅ `tests/test_mcp_auth.py` — 10 cases including arbitrary-audience and empty-audience.
- ✅ `tests/test_well_known.py` — 5 cases including issuer + backplane trailing-slash strip.
- ✅ `tests/test_mcp_server.py` — `verify_mcp_jwt_and_bind` overridden via `app.dependency_overrides`; cleanup in `finally`; all 25 dispatcher tests preserved.
- ✅ `tests/test_auth_jwt.py` — `_decode_with_jwks` signature kwarg propagation.
- ✅ `docs/codebase/backend.md` — accurate, no analogies, no stale claims.

No 🚫 undisclosed-scope edits. No 🕳️ missing predicted changes.

### Framework / library research (fresh)
- **authlib.jose 1.7.2** `claims_options["aud"] = {"essential": True, "value": expected_audience}` — exact-string match, list-normalised on the token side (`v in aud_list`), `InvalidClaimError("aud")` on mismatch. Empty/missing `aud` claim is caught by `_validate_essential_claims` before `validate_aud` runs. Source: `authlib/jose/rfc7519/claims.py` at v1.7.2.
- **FastAPI 0.136.1** `app.dependency_overrides[dep] = stub` + `finally: app.dependency_overrides.pop(dep, None)` is the canonical test-isolation pattern. Override set before `TestClient` construction is correct.
- **MCP 2025-06-18 §"Canonical Server URI"**: "implementations **SHOULD** consistently use the form without the trailing slash for better interoperability." `mcp_resource_uri()` strips on both code paths (explicit + derived).
- **RFC 9728 §5.1**: `WWW-Authenticate: Bearer resource_metadata="<absolute-url>"` — URL quoted. `www_authenticate_header()` matches the shape exactly.
- **RFC 9728 §3**: only `resource` is REQUIRED; `scopes_supported` is RECOMMENDED; `authorization_servers` is a JSON array of strings (PR sends `[issuer]`).
- **Pydantic v2 HttpUrl**: appends `/` to host-only URLs; `.rstrip("/")` on the serialised form is the right defensive pattern (and is what the issuer comparison in `auth/jwt.py:243` already does).

Sources (in order of citation):
- https://github.com/lepture/authlib/blob/v1.7.2/authlib/jose/rfc7519/claims.py
- https://fastapi.tiangolo.com/advanced/testing-dependencies/
- https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization (§"Canonical Server URI", §"WWW-Authenticate", §"Token Handling")
- https://datatracker.ietf.org/doc/html/rfc9728 §3, §5.1

### Integration verification (run locally in `.claude/worktrees/pr-290-review`)
- [x] `ruff format --check src/ tests/` — 61 files formatted.
- [x] `mypy src/` (strict) — 29 source files, no issues.
- [x] `pytest tests/` — **278 passed, 1 deselected**. The single deselected case is `test_observability::test_metrics_endpoint_returns_prometheus_text_format` (asserts `process_resident_memory_bytes` is emitted by `prometheus_client.process_collector` — Linux-only because it reads `/proc/self/stat`). It is **pre-existing on `main`** and unrelated to this PR. CI's Python check (Linux) is green.
- [x] MCP-focused subset (`test_mcp_auth.py` + `test_mcp_server.py` + `test_well_known.py`) — **39 / 39 pass**.
- ⚠️ Production-ASGI boot + curl against `/.well-known/oauth-protected-resource` and `/mcp` against a real Postgres + Keycloak — **not run**. `TestClient` exercises the full FastAPI middleware chain and routes, and `test_mcp_auth.py` mocks Keycloak via `respx`; the auth chain is therefore covered end-to-end at the FastAPI layer, but I did not boot uvicorn against a live Keycloak. CI's image-build and Helm-template gates are green; running the full stack-up smoke is the operator's call before merge.
- n/a migrations (none in this PR), n/a frontend, n/a agent traces.

### Findings
- 🛑 0 blockers.
- ⚠️ 0 majors.
- 💡 1 minor — see inline. (CodeRabbit's open nitpick on `verify_mcp_jwt_and_bind` being `async` without `await` — agreed it's low value; sibling `verify_mcp_jwt` is async because it awaits `verify_jwt_for_audience`, and keeping them shaped the same reads more cleanly than micro-optimising the coroutine overhead. Author can take or leave at discretion.)
- ❓ 0 unresolved questions. (The `del operator` style smell was raised by @ikaric, disputed by author with a forward-compat rationale for T3 — accepted; the dispatcher's own comment block documents the intent.)

### What was done well
- The `verify_jwt_for_audience` seam is the right refactor — keeps signature / issuer / kid-rotation / `Operator` projection identical across chassis and MCP and only varies the audience. Chassis and MCP cannot drift on what "a valid token" means structurally.
- The empty-audience guard (`audience_not_configured` 401 in `verify_jwt_for_audience`) makes the fail-closed property local to the verifier rather than depending on the issuer never minting `aud=""`. The `test_empty_audience_setting_returns_401` test pins it against future regressions.
- `test_well_known_keycloak_issuer_url_has_no_trailing_slash` was strengthened in iter-2 to set the env var **with** a trailing slash and assert exact equality — genuinely exercises the strip rather than passing trivially. Same for the new `test_token_for_arbitrary_audience_is_rejected_at_mcp` which pins the value-based audience check against any non-matching `aud`.
- `mcp_auth_failed` structured-log event (with the chassis failure-mode token as `detail`) gives on-call a grep-able signal to distinguish MCP-side 401 spikes from chassis-side ones.
- DCO situation surfaced honestly in the PR body with cited precedent (#181) rather than swept under a force-push.

### Docs
- [x] `docs/codebase/backend.md` — updated with an accurate, analogy-free MCP T2 section that supersedes the T1-era stale claims about /mcp having no auth.

### Pre-merge: DCO action
The DCO check is `action_required` because `d2e9c64` (the iter-1 main feature commit) lacks `Signed-off-by`. Every iter-2 commit (`2c83a19`, `7b629ce`, `9c3c6eb`, `1f928e0`, `7a3db92`) is signed. The `/auto-implement-issue` skill's never-amend / never-force-push rules block the standard `git rebase --signoff main && git push --force-with-lease` remediation, so the merge needs either:
1. Admin merge bypass (precedent: #181 / G2.7-T6, 2026-05-11), or
2. Manual rebase + force-push outside the skill's guardrails.

Not a code-review concern — flagging for the merge step.

**Auto-close (should happen automatically):**
- #247 — verify it actually closed. `Closes #247` in PR body, default-branch base, same repo → auto-close should fire on merge.

**Project-board status bumps (project 19, MEHO):**
- #247 → `Done`
- #226 → `In Progress` (T1 already merged, T2 lands here — initiative is no longer Todo)

**Completion comment to post on #247 close:**
> Closed by #290. OAuth 2.1 resource-server protection on `/mcp` per RFC 9728 + RFC 8707 + MCP 2025-06-18 §Authorization. Verified end-to-end: 8 auth-chain tests (chassis-audience reject / MCP-audience accept / arbitrary-audience reject / expired token / JWKS unreachable / empty-audience guard / explicit-URI override / WWW-Authenticate shape) + 5 well-known route tests + all 25 T1 dispatcher tests preserved via `app.dependency_overrides` of `verify_mcp_jwt_and_bind`.

backend/src/meho_backplane/mcp/auth.py
zdamir: DISPUTE — keeping `async`.

The shape matches **both** siblings deliberately:

1. `verify_mcp_jwt` (immediate sibling) is `async` because it awaits `verify_jwt_for_audience`. Reading the two MCP-auth dependencies side-by-side with matching signatures makes the pair grep-able and the pattern "dependency that wraps an inner dependency for side-effects" visually consistent.
2. `verify_jwt_and_bind` in `middleware.py` (chassis sibling) is **also** `async def` with no `await` in its body — same `bind_contextvars` shape, established by Goal #11. The chassis chose this shape on the same consistency rationale; the MCP module mirrors it.

The coroutine-creation overhead is a few ns per request — measurable but not visible at any deployment-realistic throughput. Flipping `verify_mcp_jwt_and_bind` to a plain `def` while the rest of the family stays async is the worse trade: it forces every future reader to chase down "why is *this* one a `def`?" The cohort cost (consistency) beats the per-call cost (one coroutine object).

If the chassis `verify_jwt_and_bind` ever flips to a plain `def` (e.g. as part of a "drop unnecessary async" sweep), this should flip with it — same commit, same rationale. Until then, async stays here for the pair-grep property.

PR #294, opened 2026-05-12
feat(mcp): tool + resource registries with RBAC + jsonschema (G0.5-T3)

zdamir:
## Summary

The substrate every G3-G9 verb registers against. Two parallel registries mirroring the MCP 2025-06-18 separation of **tools** (active — invoked by `tools/call`) and **resources** (passive — fetched by `resources/read`):

- `mcp/registry.py` — `register_mcp_tool(definition, handler)` / `register_mcp_resource(definition, handler)` (mirrors the `register_probe` shape). `ToolDefinition` + `ResourceTemplateDefinition` carry MEHO-internal `required_role` + `op_class` fields that `to_wire()` strips from the client-facing shape. RBAC ranking (`read_only` < `operator` < `tenant_admin`) is centralised in `_role_at_least` and applied at both list time and call time.
- `mcp/handlers.py` — five JSON-RPC methods wired to the registries: `tools/list`, `tools/call`, `resources/list`, `resources/templates/list`, `resources/read`. Tool arguments validated against `inputSchema` via `jsonschema` 4.26+.
- `eager_import_mcp_modules()` (called from the FastAPI lifespan) walks every module under `mcp/tools/` and `mcp/resources/` via `pkgutil.iter_modules` so side-effect-only `register_*` calls run before the first request arrives.
- `_McpHandler` signature widened from `(params)` to `(operator, params)` so registry handlers can apply RBAC. `initialize` now advertises non-empty `capabilities` again (T1's empty-envelope stance was paired with "no handlers registered"; T3 flips them back).

## Spec correctness vs. issue body

The issue body's AC says `resources/list` returns the registered resources. The MCP 2025-06-18 spec is stricter: **concrete** resources go through `resources/list`; **templated** resources go through `resources/templates/list`. v0.2 ships only templates so `resources/list` returns an empty list and `resources/templates/list` carries the registry — spec-conforming clients (MCP Inspector, Claude Desktop) get the right shape and don't have to guess. Tests assert both methods independently.

## Test plan

- [x] `ruff check src/ tests/` — clean
- [x] `ruff format --check` — 66 files formatted
- [x] `mypy src/` (strict) — 33 source files, no issues
- [x] `pytest tests/` — **301 passed**, 1 deselected (pre-existing macOS-only flake)

### AC coverage

[...]

PR #293, opened 2026-05-12
fix(backplane): lazy stdout binding in PrintLoggerFactory to fix cross-test fd flake

damir-topic:
## Summary

- Fixes the cross-test flake where `tests/test_health.py::test_healthz_returns_ok` fails with `ValueError: I/O operation on closed file.` after `tests/test_auth_config.py` runs in the same session.
- Replaces `PrintLoggerFactory(file=sys.stdout)` with the no-arg `PrintLoggerFactory()` so structlog's `PrintLogger.msg` resolves `sys.stdout` lazily at write time instead of capturing the pytest-`capfd`-wrapped fd at lifespan-startup time.
- Updates the `configure_logging` docstring + `docs/codebase/backend.md` to document the lazy stdout binding contract this fix relies on.

## Why this works

structlog 25.5.0 `_output.py` `PrintLogger.msg`:

```python
def msg(self, message: str) -> None:
    f = self._file if self._file is not stdout else None
    with self._lock:
        print(message, file=f, flush=True)
```

When `self._file is sys.stdout` (the structlog-module-level alias captured at structlog import time — the original process stdout, before pytest swaps anything), `msg` calls `print()` with no `file=` keyword, which Python resolves against the *current* `sys.stdout` per call. With the no-arg factory shape, every constructed `PrintLogger` defaults its `_file` attribute to that alias (`self._file = file or stdout`), so the lazy branch is taken on every emit.

Production behaviour is unchanged because the real process `sys.stdout` does not get rebound at runtime; tests benefit because pytest's stdout swapping is honoured per-write.

Verified empirically:

```
PrintLoggerFactory()      # no args
factory()._file is sys.stdout  # True
# Swap sys.stdout to two StringIOs in sequence; each msg() lands in the right one.
```

And the bug reproduces on the eager form (`PrintLoggerFactory(file=sys.stdout)`) with an explicit `sys.stdout.close()`, which is exactly what pytest's `capfd` teardown does to its wrapped fd.

## Test plan

- [x] `pytest tests/test_auth_config.py tests/test_health.py` — 13 passed (was: 1 failed before the fix)
- [x] `pytest tests/` — 233 passed, 7 skipped (no regressions)
- [x] `pytest tests/test_middleware.py tests/test_health.py` — 11 passed (the PR #265 fix-iter-2 workaround in `test_middleware.py` is preserved; its `_configure_capture(log_buffer)` job is to redirect into a private `StringIO`, separate from the closed-fd flake)
[...]

PR #267, opened 2026-05-12
feat(connectors): Connector ABC + schemas module skeleton (G0.2-T1, #240)

kr3s0:
## Summary
- Adds `meho_backplane.connectors` package: the ABC + result schemas every downstream connector inherits from (G0.2-T1 foundation)
- Three new files: `connectors/schemas.py` (FingerprintResult, ProbeResult, OperationResult, AuthModel), `connectors/base.py` (Connector ABC), `connectors/__init__.py` (re-exports public API)
- 23 new unit tests covering all acceptance criteria; updates `docs/codebase/backend.md`

## Test plan
- [x] `ruff check` — clean
- [x] `ruff format --check` — clean
- [x] `mypy src/meho_backplane/connectors/` — clean (3 source files)
- [x] `pytest tests/test_connectors_base.py` — 23 passed
- [x] Package exists with `__init__.py`, `base.py`, `schemas.py`
- [x] `Connector` ABC: instantiating without implementing all 3 methods raises `TypeError`
- [x] Schema round-trip: `model_dump()` → `model_validate()` lossless for all three models
- [x] All three models frozen via `ConfigDict(frozen=True)` — mutation raises `ValidationError`
- [x] `AuthModel(value)` works for all three strings; unknown value raises `ValueError`
- [x] No dependency on G0.1 surface — `"Target"` is a forward-reference string annotation

## Out of scope
- The connector registry (T2)
- `HttpConnector` / `SshConnector` adapters (T3, T4)
- VaultConnector reference impl (T5)
- CLI dispatch (T6)
- Streaming / long-running ops (v0.2.next)

Closes #240

## Summary by CodeRabbit

* **New Features**
  * Added a Connector API with async operations for fingerprinting, probing, and executing against targets.
  * Introduced AuthModel with three identity modes and three structured, immutable result types: FingerprintResult, ProbeResult, OperationResult (includes stable extras serialization).

* **Documentation**
  * Documented connector API, auth modes, and result model semantics and import surface.

* **Tests**
  * Added tests validating the public connector API, model immutability, and round-trip serialization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/267)

PR #296, opened 2026-05-12
feat(retrieval): fastembed EmbeddingService + chart model-cache PVC (G0.4-T2)

### LINKED ISSUE #259: Task: G0.4-T2 fastembed wiring — EmbeddingService + lazy load + lifespan preload + chart model-cache PVC (default BAAI/bge-small-en-v1.5)

ikaric:
## Summary

Lands **T2 of Initiative #225** (G0.4 Retrieval substrate) — the fastembed-backed in-process embedding pipeline that T3 (`index_document`, #260) and T4 (`retrieve`, #261) both consume. Closes #259.

This PR has been rebased onto current `main` after #269 (T1) merged, picking up the substrate it builds on plus the parallel work in #267 / #290 / #293 / #294.

## Design choices (locked by Initiative body)

* **fastembed, not sentence-transformers.** Per v0.1-spec L391: fastembed ships ONNX runtime (~120 MB) instead of PyTorch (~1.5 GB). Memory footprint matters in v0.2's single-replica backplane.
* **Default model `BAAI/bge-small-en-v1.5`** (384-dim, Apache-2.0, English). Matches the `vector(384)` column type from migration `0003` (T1). Operators swap via `RETRIEVAL_EMBEDDING_MODEL`; changing dimensionality requires a re-embed-everything migration.
* **Lazy model load + `asyncio.to_thread` wrap.** `EmbeddingService` imports `fastembed.TextEmbedding` inside `_ensure_loaded` so module import doesn't pull onnxruntime; the sync `model.embed` runs in a worker thread so the FastAPI event loop stays responsive.
* **Per-call structlog logger.** `structlog.get_logger()` is resolved inside `_ensure_loaded` rather than at module level so the `PrintLogger` binds against the current `sys.stdout` — additive to #293's lazy-stdout fix; both layers harden the same class of cross-test fd flake.

## Lifespan integration

`main.py` lifespan now calls `await get_embedding_service().encode_one("model preload")` once at startup. Failure is logged warn-level and falls back to lazy-on-first-call; a real misconfiguration surfaces on the first `/api/v1/retrieve` call.

## Chart-side surface

- `retrieval.modelCache.enabled` (default `true`) renders a RWO PVC `<release>-fastembed-cache` of size `retrieval.modelCache.size` (default `200Mi`), mounted at `retrieval.modelCache.mountPath` (default `/var/cache/fastembed`).
- `config.retrievalEmbeddingModel` + `config.retrievalModelCacheDir` flow into the ConfigMap as `RETRIEVAL_EMBEDDING_MODEL` + `RETRIEVAL_MODEL_CACHE_DIR`.
- `values.schema.json` ships the same shape with `additionalProperties: false` so a typo'd key fails `helm install`.
- Migration Job deliberately does **not** mount the cache — migrations don't embed.

## Tests

`backend/tests/test_retrieval_embedding.py` (new) — 10 SQLite tests + 1 opt-in slow test:

- Constructor doesn't load fastembed (regression for the lazy-load contract); `~` in cache dir expands eagerly.
- `get_embedding_service()` is a `@lru_cache(maxsize=1)` singleton; env-var overrides surface after `cache_clear`.
- Stubbed encode round-trip proves `list[float]` (not numpy) of length 384; batch order preserved.
- First-call patch proves `TextEmbedding` is constructed exactly once across multiple encodes.
- `asyncio.to_thread` proof: a 200 ms blocking embed runs concurrently with a 10 ms `asyncio.sleep` and the fast task finishes first.
- `@pytest.mark.slow` real-model test (skipped unless `MEHO_RUN_SLOW_TESTS=1`) loads `BAAI/bge-small-en-v1.5`, asserts 384-dim deterministic vectors with semantic discrimination. Verified end-to-end locally.

The `slow` pytest marker is registered in `pyproject.toml` so `--strict-markers` (when CI flips it on) doesn't warn.

## Test plan

- [ ] `ruff check .` — clean.
- [ ] `mypy --strict src/` — clean across 38 source files.
- [ ] `pytest tests/` — 339 passed, 1 skipped (slow), 1 pre-existing macOS-only `test_observability` failure unchanged.
- [ ] Lifespan smoke-tested end-to-end locally: real model downloads + loads + encodes test sentence successfully (verified before stacking T3-T6).
- [ ] CI green.

## Stack context

T3 (#260), T4 (#261), T5 (#262), and T6 (#263) are implemented locally on stacked branches. Each opens for review only after its predecessor merges, keeping at most one G0.4 PR open at a time. The rebase cascade is mechanical (`git rebase --update-refs origin/main`) and was verified locally after this PR's rebase resolved the T1 + parallel-work conflicts.

## Summary by CodeRabbit

* **New Features**
  * In-process text embedding service with lazy model loading, optional startup preloading (logs success/warning), and optional PVC-backed model cache in deployments
  * New config keys and chart values to select embedding model and cache directory, plus chart options to enable/disable PVC cache

* **Documentation**
  * Backend docs updated to describe the embedding service and deployment/cache wiring

* **Tests**
  * New comprehensive tests for embedding behavior (including opt-in slow real-model test)

* **Chores**
  * Added fastembed dependency for ONNX-based embeddings

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/296)


backend/tests/test_retrieval_embedding.py
```diff
@@ -0,0 +1,395 @@
...
+    fastembed + onnxruntime into the import graph.
+
+    Catches a regression where someone moves the ``from fastembed
+    import TextEmbedding`` line to module-level (or to the constructor)
+    - the assertion that ``self._model is None`` after construction is
+    the load-bearing contract.
+    """
+    service = EmbeddingService(
+        model_name="BAAI/bge-small-en-v1.5",
+        cache_dir="/tmp/meho-test-cache",
```
zdamir: **[🛑 B1]** — Replace `/tmp/meho-test-cache` literals with the `tmp_path` fixture (Sonar S5443 cluster)

**What:** Seven uses of a `/tmp/meho-...-cache` literal as `EmbeddingService.cache_dir` / `RETRIEVAL_MODEL_CACHE_DIR` (this file lines 98, 150, 156, 190, 228, 267, 313). The slow real-model test at line 367 already does the right thing with `cache_dir=str(tmp_path)`; the non-slow tests are inconsistent.

**Why it matters:** SonarCloud's CI Quality Gate fails the PR on this — "D Security Rating on New Code" (required ≥ A) per the SonarCloud check. The rule (`python:S5443`, CRITICAL/VULNERABILITY) flags using `/tmp` because it's world-writable and TOCTOU / symlink-attack territory. Even though most of these lines never actually write to disk (the model is stubbed in non-slow tests), the rule is static — it can't see the stubbing — and the CI gate is enforced. The PR's merge gate stays red until this clears.

**Required change:** Make the fixtures take `tmp_path: Path` and pass `str(tmp_path / "meho-test-cache")` (or a similar per-test subdir) wherever a `/tmp/meho-...-cache` literal appears today. Pattern already in use at line 367. For the env-var test on lines 150/156, set the env var to `str(tmp_path / "meho-custom-cache")` and assert against the same value.

Example for line 98:

```diff
-def test_embedding_service_does_not_load_model_on_construction() -> None:
+def test_embedding_service_does_not_load_model_on_construction(tmp_path: Path) -> None:
     service = EmbeddingService(
         model_name="BAAI/bge-small-en-v1.5",
-        cache_dir="/tmp/meho-test-cache",
+        cache_dir=str(tmp_path / "meho-test-cache"),
     )
```

(Repeat at lines 150, 156, 190, 228, 267, 313 — `tmp_path` is a pytest built-in, no extra fixture imports needed.)

**Source:** SonarCloud check on this PR — https://sonarcloud.io/dashboard?id=evoila_meho&pullRequest=296. Sonar rule `python:S5443`. Pattern already used correctly at line 367.

backend/tests/test_retrieval_embedding.py
```diff
@@ -0,0 +1,395 @@
...
+            yield [float(i)] * 384
+
+    stub_model = MagicMock()
+    stub_model.embed.side_effect = fake_embed
+    service._model = stub_model
+
+    vecs = await service.encode(["first", "second", "third"])
+
+    assert len(vecs) == 3
+    assert vecs[0][0] == 0.0
```
zdamir: **[⚠️ M1]** — Use `pytest.approx` for the stubbed-vector equality assertions (Sonar S1244 cluster)

**What:** Lines 246-248 assert exact float equality:

```python
assert vecs[0][0] == 0.0
assert vecs[1][0] == 1.0
assert vecs[2][0] == 2.0
```

**Why it matters:** SonarCloud's CI Quality Gate fails the PR on these as `python:S1244` (MAJOR/BUG, "Do not perform equality checks with floating point values"). The same test file uses `pytest.approx()` at lines 210-212 for the identical pattern; the inconsistency itself reads as a code smell ("why is one block approx and the other exact?"). For the integer-valued floats this stub returns, `float(i)` is exact, so the bug-class is over-flagged — but the CI gate is what counts, and a unified `pytest.approx()` style here removes the contradiction.

**Required change:**

```diff
-    assert vecs[0][0] == 0.0
-    assert vecs[1][0] == 1.0
-    assert vecs[2][0] == 2.0
+    assert vecs[0][0] == pytest.approx(0.0)
+    assert vecs[1][0] == pytest.approx(1.0)
+    assert vecs[2][0] == pytest.approx(2.0)
```

`pytest` is already imported at line 39.

**Source:** SonarCloud check on this PR — https://sonarcloud.io/dashboard?id=evoila_meho&pullRequest=296. Sonar rule `python:S1244`. Pattern already used elsewhere in this file at lines 210-212.

zdamir (changes requested):




**Anchor task:** #259 — G0.5-T2 fastembed `EmbeddingService` + lazy load + lifespan preload + chart model-cache PVC
**Parent initiative:** #225 — G0.4 Retrieval substrate
**Parent goal:** #221 — G0 Foundational substrate
**Surface area:** backend (Python embedding service) + devops (Helm chart PVC + ConfigMap)
**Best-practices packs audited:** python, devops, ai
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `c2d747a`
**AI reviewer (CodeRabbit):** received in 191s — 2 inline (Major + Minor) + 2 review-body nitpicks

---

### Acceptance criteria

| # | Criterion | Status | How verified |
|---|---|---|---|
| 1 | `fastembed` in `pyproject.toml`; `uv sync` installs cleanly | ✅ | `uv sync --locked --all-groups` clean in review worktree; fastembed 0.7.x in resolved deps |
| 2 | `get_embedding_service()` returns a singleton `EmbeddingService` | ✅ | `test_get_embedding_service_returns_singleton` — passes |
| 3 | `encode_one("hello world")` returns 384-dim `list[float]`; deterministic | ✅ via slow-lane | Stubbed: `test_encode_one_returns_384_dim_list_with_stubbed_model` (shape only). Real-model determinism: `test_real_model_loads_and_returns_384_dim_vector` is gated on `MEHO_RUN_SLOW_TESTS=1`, marked `slow`, asserts `vec_a1 == vec_a2`. Slow-skipped in always-on suite; verified by author locally per PR body. Acceptable for v0.2's slow-lane convention. |
| 4 | `encode([t1, t2, t3])` returns 3 vectors; batching works | ✅ | `test_encode_batches_preserve_order` |
| 5 | `asyncio.to_thread` keeps event loop responsive | ✅ | `test_encode_does_not_block_event_loop` — slow embed runs concurrently with fast `asyncio.sleep`, fast resolves first |
| 6 | Settings env-var overrides (`RETRIEVAL_EMBEDDING_MODEL` / `RETRIEVAL_MODEL_CACHE_DIR`) | ✅ | `test_get_embedding_service_binds_settings` |
| 7 | Lifespan integration: first `/api/v1/retrieve` doesn't trigger model load | ✅ (indirect) | `main.py` lifespan calls `encode_one("embedding model preload")` at startup; the `@lru_cache(maxsize=1)` singleton + `_ensure_loaded`'s `if self._model is None` short-circuit guarantee the cached model survives. T5's `/api/v1/retrieve` hasn't shipped (out of scope per PR body) — the no-reload claim relies on the singleton contract, which `test_first_encode_triggers_fastembed_import` proves. |
| 8 | Helm `template` renders PVC + volume mounts when `retrieval.modelCache.enabled=true`; nothing when disabled | ✅ | Verified locally with the CI-equivalent override set: enabled → 7 fastembed-cache references (PVC + 2× volumeMount + 2× volume + 2× claim references); disabled → 0 references. PVC `accessMode: ReadWriteOnce`, size `200Mi`, mountPath `/var/cache/fastembed`. |
| 9 | Schema lint: `helm lint` clean | ✅ | `helm lint` with the same CI override set → "1 chart(s) linted, 0 chart(s) failed" |
| 10 | CI green | ❌ | **2 CI failures on the current head**: (a) SonarCloud Quality Gate FAILED — D Security Rating, C Reliability Rating on New Code (see B1 / M1 below); (b) Python License Check FAILED on `cffi 2.0.0 UNKNOWN` — pre-existing, brought in transitively, not introduced by this PR (kr3s0 documented same issue on PR 294's review). All other CI checks green (Python ruff/mypy/pytest, Go, Helm lint+template+kubeconform, Semgrep OSS+SAST, TruffleHog, Image build, DCO). |

---

### Project-board alignment

| Issue | Repo | Link type | Auto-close? | Delivered? | Action |
|---|---|---|---|---|---|
| #259 | evoila/meho | `Closes #259` in PR body | ✅ same-repo | ✅ full (modulo CI gate) | Close on merge once CI gate clears |
| #225 (initiative) | evoila/meho | parent of #259 | ❌ remains open | partial — T1 (#258) merged via #269; T2 (#259, this PR); T3-T6 (#260-#263) on stacked branches per PR body | remains open |
| #221 (goal) | evoila/meho | parent of #225 | ❌ remains open | partial | remains open |

No project-16 misalignment to surface (issues live in `evoila/meho`, consistent with prior G0.4 / G0.5 PRs).

---

### Predicted vs. actual diff

Predicted from issue #259: fastembed dep, `retrieval/__init__.py` + `embedding.py`, settings additions, main.py lifespan extension, chart values.yaml/values.schema.json/deployment.yaml/pvc template, tests, docs. 13 files predicted, 13 files in the PR — 1:1 match.

- ✅ Expected and correct (13 files): `backend/pyproject.toml`, `backend/uv.lock`, `backend/src/meho_backplane/main.py`, `backend/src/meho_backplane/retrieval/__init__.py` (new), `backend/src/meho_backplane/retrieval/embedding.py` (new), `backend/src/meho_backplane/settings.py`, `backend/tests/test_retrieval_embedding.py` (new), `deploy/charts/meho/templates/configmap.yaml`, `deploy/charts/meho/templates/deployment.yaml`, `deploy/charts/meho/templates/pvc-fastembed-cache.yaml` (new), `deploy/charts/meho/values.schema.json`, `deploy/charts/meho/values.yaml`, `docs/codebase/backend.md`
- 🚫 No undisclosed scope
- 🕳️ No missing files

The chart additions (extraVolumes / extraVolumeMounts / extraEnv beyond just the fastembed-cache mount) are mentioned in the diff for trust-manager CA-bundle use cases (Issue #209). These appear tangentially related to T2's scope but the PR body doesn't call them out as fastembed-coupled. Treating them as part of the same Helm-template housekeeping pass — non-blocking, but worth noting in a future scope-discipline retrospective if Issue #209 has its own PR coming.

---

### Integration verification (in review worktree on `c2d747a`)

- [x] `ruff format --check src/ tests/` — 74 files formatted
- [x] `mypy src/ --ignore-missing-imports` — 38 source files, no issues
- [x] `pytest tests/ -q` (with macOS flake deselected) — **348 passed, 1 skipped, 1 deselected** (skipped = the slow real-model test; deselected = pre-existing `process_resident_memory_bytes` macOS flake)
- [x] `helm lint deploy/charts/meho/` with CI-equivalent override set — clean
- [x] `helm template` enabled (default) — 7 fastembed-cache references; PVC + 2× volumeMount + 2× volume render; ConfigMap surfaces `RETRIEVAL_EMBEDDING_MODEL` + `RETRIEVAL_MODEL_CACHE_DIR`
- [x] `helm template` disabled — 0 fastembed-cache references (toggle correct)
- [❌ CI] SonarCloud Quality Gate — FAILED (10 issues in `test_retrieval_embedding.py` — see B1 + M1 inline below)
- [❌ CI pre-existing] Python License Check — FAILED on `cffi 2.0.0 UNKNOWN`, not introduced by this PR (kr3s0's PR 294 review documented same issue)
- ⚠️ `kubeconform` — not run locally; CI's `Helm (lint + template + kubeconform)` check passes
- ⚠️ Live MCP / real fastembed model — not run (slow-lane test gated on `MEHO_RUN_SLOW_TESTS=1`)

---

- 🛑 **B1** — `test_retrieval_embedding.py` cluster (7 lines: 98, 150, 156, 190, 228, 267, 313): replace `/tmp/meho-...-cache` literals with `tmp_path` fixture (Sonar S5443 CRITICAL VULNERABILITY × 7). Required change: per-test fixture pattern already used correctly at line 367.
- ⚠️ **M1** — `test_retrieval_embedding.py:246-248`: use `pytest.approx()` for float-equality assertions on stubbed-vector values (Sonar S1244 MAJOR BUG × 3). Required change: 3-line diff; `pytest.approx` already imported and used elsewhere in this file at lines 210-212.

Plus 2 confirmed-real findings already inline from CodeRabbit (not duplicated by this skill to avoid noise on the same diff lines):

- ⚠️ **CR-1** — `retrieval/embedding.py:165`: `_ensure_loaded` not thread-safe; concurrent first calls via `asyncio.to_thread` can race and instantiate `TextEmbedding` twice. Real but practically mitigated by the lifespan preload (the race window is closed once startup runs). CR's proposed `threading.Lock` + double-checked locking pattern is the right fix. Adopt as posted.
- 💡 **CR-2** — `tests/test_retrieval_embedding.py:54`: `_RUN_SLOW_TESTS = bool(os.environ.get("MEHO_RUN_SLOW_TESTS"))` treats any non-empty value (including `"0"`) as truthy. Real bug. CR's proposed `.strip() == "1"` parse is the right fix. Adopt as posted.

---

- **CR-nit-1** (CodeRabbit review-body): preload string `"embedding model preload"` is generic; using a representative sentence would exercise tokenizer / attention paths more typical of real traffic. Optional polish — the warm-up effect is dominated by ONNX runtime init, not tokenizer-specific paths.
- **CR-nit-2** (CodeRabbit review-body): add `len(vec) != EMBEDDING_DIMENSION` validation in `_sync_encode` to fail fast on a misconfigured model with a clear retrieval-layer error rather than a downstream pgvector binding error. Defensive guard. Today an operator swapping `RETRIEVAL_EMBEDDING_MODEL` to a non-384-dim model would see the failure at the DB INSERT layer; CR's check moves it to the embed boundary. Reasonable polish — author's call.
- **Mine-1** — `tests/test_retrieval_embedding.py:36` imports `cast` and uses it only in a no-op test (`test_typing_cast_helpers_compile`, lines 388-395) whose body is `_ = cast`. The docstring explains "Some test files use `typing.cast`…" but this file doesn't use it for any real cast. Dead-cruft pattern; the import + the no-op test can both be removed. Genuinely optional cleanup; not worth a merge block.

---

- **Findings adopted:** 2 inline (CR's existing comments; not duplicated here — CR-1 / CR-2 in the Findings list)
- **Findings deferred to Adjacent improvements:** 2 (CR-nit-1, CR-nit-2 — both genuinely optional)
- **Disputed:** 0

CR's threading-race finding (CR-1) is the most consequential. The lifespan preload closes the practical race window in production, but the defensive lock is the right shape — costs nothing, makes tests honest.

---

### Docs

- [x] `docs/codebase/backend.md` — retrieval substrate / embedding section updated. Status: `updated`.

---

### What worked well

- **fastembed-over-sentence-transformers locked at the spec level.** PR body cites v0.1-spec L391; the dependency footprint matters in the single-replica v0.2 backplane. `~120 MB` ONNX vs `~1.5 GB` PyTorch is the right tradeoff and the design surfaces it openly.
- **Lazy fastembed import inside `_ensure_loaded`.** The test suite imports `meho_backplane.retrieval.embedding` without pulling fastembed or onnxruntime into the import graph. `test_embedding_service_does_not_load_model_on_construction` locks in the contract so a future refactor that hoists the import to module-level fails loudly.
- **`asyncio.to_thread` proof test (`test_encode_does_not_block_event_loop`).** The 200 ms blocking embed + 10 ms fast `asyncio.sleep` + `asyncio.gather` ordering assertion is the right shape — it would fail loudly if a future change removed the thread offload and re-stalled the event loop.
- **Chart-side `additionalProperties: false` on `retrieval.modelCache`.** Typo'd chart-value keys (e.g. `modelCach`, `sze`) fail `helm install` at schema-validation time rather than silently rendering empty configuration.
- **Per-call `structlog.get_logger()` resolution** inside `_ensure_loaded`. PR body explains this is additive to #293's lazy-stdout fix — both layers harden the same cross-test fd flake class. Good defense-in-depth.
- **Migration Job deliberately does NOT mount the cache.** The PR body calls this out explicitly: migrations don't embed. Avoids a class of "why is the migration mounting a cache it never writes to?" confusion in operator-facing chart docs.

ikaric:
## Review feedback addressed — pushed in `e0f2e14`

Force-push of T2 (with-lease). All findings resolved:

### Blockers (Damir)

- **B1 — `python:S5443` `/tmp` literals in tests.** Swept all 7 sites in `tests/test_retrieval_embedding.py` (lines 98, 150, 156, 190, 228, 267, 313) to use the `tmp_path` pytest fixture per the suggested pattern. The env-var test now sets `RETRIEVAL_MODEL_CACHE_DIR` to `str(tmp_path / "meho-custom-cache")` and asserts against the same path so the equality check stays meaningful.

### Major

- **M1 — `python:S1244` exact float equality (Damir).** Lines 246-248 in `test_encode_batches_preserve_order` switched to `pytest.approx`. Style now matches the sibling vector test that was already using `approx` at lines 210-212.
- **CodeRabbit — lazy-init thread race.** Added `self._model_lock = threading.Lock()` on `EmbeddingService.__init__` and converted `_ensure_loaded` to the double-checked locking pattern: fast path (single attribute read, no lock) when the model is already loaded; lock-guarded re-check + assignment on the cold path. The lifespan preload normally wins this race in production, but the contract no longer depends on lifespan ordering — concurrent `encode()` calls from the thread pool can no longer double-instantiate `TextEmbedding`.

### Minor

- **CodeRabbit — slow-test gate.** `_RUN_SLOW_TESTS` parsing tightened from `bool(os.environ.get(...))` to a strict-truthy check against `{"1", "true", "yes", "on"}` (case-insensitive, whitespace-stripped). Matches the discipline `parse_bool_env` already uses in `settings.py`. `MEHO_RUN_SLOW_TESTS=0` now evaluates to `False` rather than truthy-by-non-empty.

### Nitpicks

- **CodeRabbit — dimension-drift guard in `_sync_encode`.** Added a per-vector length check against `EMBEDDING_DIMENSION` inside the worker thread; on mismatch raises `ValueError` with the model name, expected vs. actual dim, batch position, and a pointer to migration `0003` so an operator who swaps to a wrong-dim model gets the diagnosis at the retrieval layer rather than a downstream pgvector bind error.
- **CodeRabbit — generic preload text.** Skipped intentionally. The preload's purpose is to flip the lazy-load flag and amortise the ONNX runtime init cost; ONNX's first-call cost is shape-dependent, not content-dependent, so a longer realistic sample doesn't measurably improve warm-up. Keeping the short preload also keeps the lifespan log line readable.

### Pre-existing CI failure

- `cffi:2.0.0 UNKNOWN`-license is the same regression flagged on #294's review (not introduced here). Will file a separate follow-up against the license-check allow-list.

### Validation

- `ruff check .` clean (78 files).
- `ruff format --check .` clean.
- `mypy --strict src/` clean across 38 source files.
- `pytest tests/` — 339 passed, 1 skipped (slow lane), 1 pre-existing macOS-only `test_observability` failure unchanged.
- T3-T6 cascade-rebased onto the amended T2; whole local stack stays green.

cc @damir-topic @coderabbitai — ready for re-review.

zdamir (approved):

PR #345, opened 2026-05-13
feat(retrieval): POST /api/v1/retrieve route + audit payload enrichment (G0.4-T5)

ikaric:
## Summary

Lands **T5 of Initiative #225** (G0.4 Retrieval substrate) — the HTTP diagnostic surface that wraps the in-process `retrieve` helper. Primary callers (G4 `meho kb search`, G5 `meho recall`) invoke `retrieve` directly in-process; the HTTP route serves operator diagnostics, cross-language consumers, and future MCP-tool registrations.

Closes #262. Rebased onto current `main` after #299 (T4) merged (which also picked up #295, #319, #327, #328, #329, #330 along the way).

## RBAC + tenant scoping

Gated on `require_role(TenantRole.OPERATOR)` via a module-level `_require_operator = Depends(require_role(...))` closure (matches the convention `rbac_test.py` established for B008-clean parameterised dependencies). `read_only` JWTs get 403 + `insufficient_role` log event; `tenant_admin` passes the gate.

Tenant-scoped by construction: the route passes `operator.tenant_id` (from the JWT claim) to `retrieve`. No API surface accepts a tenant id in the request body — cross-tenant retrieval is impossible by construction.

## Audit privacy contract

Per v0.2 sensitivity defaults (decision #3 in `docs/planning/v0.2-decisions.md`), retrieval queries can leak operator intent. The audit_log row's `payload` carries `{query_hash, source, kind, hit_count}` — the SHA-256 hex digest of the raw query (matching the chassis's `compute_body_hash` discipline so analysts use one hash function across audit + index surfaces) + filter metadata + result count.

**The raw query string is NEVER persisted in audit_log.** An explicit negative assertion (`raw_query not in json.dumps(payload)`) guards against a future refactor stuffing the query into a debug field.

**Mechanism**: the route binds four `audit_*` contextvars **before** calling `retrieve` so a handler exception still produces an audit row with partial enrichment. The audit middleware reads any `audit_*` contextvars at write time, strips the prefix, and merges them into the payload dict. Routes that bind nothing keep the chassis-era empty-dict behaviour — the enrichment is purely additive, no breaking change.

## Schema

- `RetrieveRequest` (pydantic v2 frozen): `query: str (1-2000)`, `source: str | None (max 64)`, `kind: str | None (max 64)`, `limit: int (1-50)`. Limit caps at 50 (matches `retriever.CANDIDATE_LIMIT`); pagination beyond that is out of scope.
[...]

PR #328, opened 2026-05-12
feat(connectors): KubernetesConnector skeleton — kubeconfig + fingerprint + probe (G3.2-T1)

damir-topic:
## Summary

- KubernetesConnector skeleton: kubeconfig-driven ApiClient cache, fingerprint via `VersionApi.get_code()`, kubeconfig-free probe against `/readyz`, execute returning canonical `unknown_op` until T2-T5 wire the per-op handlers.
- Structural `KubernetesTargetLike` Protocol + injectable `KubeconfigLoader` so T1 lands ahead of G0.3 (#224) and the operator-context Vault read path; default loader raises `NotImplementedError` with a self-documenting message until those land.
- Conditional registration against the connector registry — degrades to a structlog `connector_registry_deferred` event until G0.2-T2 (#241, PR #295) merges; auto-registers without code changes here once it lands.
- `gitVersion` suffix → product slug heuristic (`rke2` / `k3s` / `eks` / `gke` / `aks` / `vanilla`) sufficient for version-tagged doc/kb lookup and the G6 broadcast classifier without an extra round-trip.

## Test plan

- [x] `uv run ruff check src/ tests/` — clean
- [x] `uv run ruff format --check src/ tests/` — clean
- [x] `uv run mypy src/` — Success: no issues found in 46 source files
- [x] `uv run pytest -x -q tests/test_connectors_k8s_auth.py` — **24 passed** (mocked `kubernetes_asyncio` + `httpx`; covers all six product mappings, fingerprint canonical shape, probe 200/401/503/transport-error, default-port fallback, execute `unknown_op`, per-target client caching, aclose idempotency, malformed YAML)
- [x] `uv run pytest -x -q` (full backend suite) — **407 passed, 13 skipped** (the 4 integration skips include the new k3s suite; existing skips unchanged)
- [x] Integration tests collect cleanly and skip on no-Docker sandboxes — verified locally; on CI runners with Docker provisioned, the `testcontainers.k3s.K3SContainer`-backed tests exercise the live fingerprint / probe / caching path against a real `rancher/k3s` API server.

## Acceptance criteria — how each is verified

- [x] `connectors/kubernetes/` package exists; conditional `register_connector("kubernetes", KubernetesConnector)` covered by the `else` branch — auto-activates the moment PR #295 (G0.2-T2 / #241) merges. The package is importable today and emits `connector_registry_deferred` until then.
[...]

PR #299, opened 2026-05-12
feat(retrieval): hybrid BM25+cosine retrieve helper with RRF fusion (G0.4-T4)

ikaric:
## Summary

Lands **T4 of Initiative #225** (G0.4 Retrieval substrate) — the shared read path G4 (`meho kb search`), G5 (`meho recall`), and future agent-grounding flows all consume. Correctness here is load-bearing for every retrieval-quality assessment downstream.

Closes #261. Rebased onto current `main` after #298 (T3) merged.

## Algorithm (per v0.1-spec L391)

Two retrieval signals merged by Reciprocal Rank Fusion (Microsoft 2009, `k=60`):

1. **BM25** via Postgres FTS: `ts_rank_cd(to_tsvector('english', body), plainto_tsquery('english', :query))` filtered by `@@`, top 50 by score desc. Uses the `documents_body_fts_idx` GIN index from migration `0003`.
2. **Cosine** via pgvector: `1 - (embedding <=> CAST(:emb AS vector))` top 50 by distance asc. Uses the `documents_embedding_idx` IVFFlat index from migration `0003`.
3. **RRF fusion** in Python: `fused_score(d) = sum over signals s where d appears: 1.0 / (RRF_K + rank(d, s))`. 1-based ranks; documents in both signals' top-50 receive two contributions, documents in only one receive one. Sort by `fused_score` desc, take top `limit`.

Each signal's candidate set is bounded to `CANDIDATE_LIMIT = 50` so the in-process fusion is O(N) over candidates, not over the corpus.

**Why RRF and not weighted-sum**: no per-signal score normalisation (ts_rank_cd is unbounded, cosine is 0-1), no per-query weight tuning, no training data. A reranker model is the v0.2.next escape hatch if corpus-recall numbers show RRF under-performs.

## Tenant scoping

Every query filters by `tenant_id` in both SQL statements. No API surface omits tenant scoping — the caller passes `tenant_id` as a required parameter, the helper threads it into both queries, and the unique composite index on `(tenant_id, source, source_id)` guarantees no cross-tenant collision in the indexed corpus.

Optional `source` (e.g. `"kb"` vs `"memory"`) and `kind` (e.g. `"kb-entry"` vs `"memory-user"`) filters narrow within a tenant via the `CAST(:source AS text) IS NULL OR source = :source` pattern.

## RetrievalHit shape

[...]

PR #298, opened 2026-05-12
feat(retrieval): index_document helper with body-hash skip-re-embed (G0.4-T3)

ikaric:
## Summary

Lands **T3 of Initiative #225** (G0.4 Retrieval substrate) — the canonical write path for the `documents` table. Both G4 (#215, kb ingestion) and G5 (#216, memory writes) will call this helper rather than re-deriving the hash + embed + upsert sequence per consumer.

Closes #260. Rebased onto current `main` after #296 (T2) merged.

## Algorithm

For each `(tenant_id, source, source_id)` natural key:

1. Look up the existing row (if any) via the unique composite index from migration `0003`.
2. Compute `SHA-256(UTF-8(body))` and compare against the existing `body_hash`.
3. **Match → skip-re-embed.** Touch `updated_at`, optionally overwrite `doc_metadata` (None preserves, dict overwrites, `{}` clears), commit, return. **No embedding call.** This is the cost optimisation that makes `meho kb refresh` against an unchanged corpus essentially free — embedding compute (~10-50 ms / text on CPU) is skipped for every unchanged document.
4. **Mismatch or no row → embed.** Call `get_embedding_service().encode_one(body)` (T2's singleton). On mismatch update every body-derived field; on no-row insert a fresh `Document`. Commit, return.

## Session handling

- **Caller-owned session path** — when the caller passes `session=...`, the helper flushes (so the returned instance has its ORM defaults visible) but does **not** commit. Caller controls the transaction lifecycle, useful for batch ingestion where the whole batch must be atomic.
- **Helper-owned session path** — when `session=None`, the helper opens its own via `get_sessionmaker()`, commits, closes. The convenient one-off-call shape T5's API route uses.

## Tenant scoping

`tenant_id` is an explicit parameter — no contextvar resolution inside the helper. T5's API route extracts from `Operator.tenant_id`; G4/G5 ingestion paths extract similarly from their own auth chains. The tenant boundary stays auditable at the call site, and the helper is trivially testable in isolation.

## Tests

`backend/tests/test_retrieval_indexer.py` (new) — 13 SQLite tests:

- `compute_body_hash` is deterministic SHA-256 of UTF-8, regression-locked against a known digest so the encoding contract can't drift silently.
- `estimate_tokens` returns `int(len(body.split()) * 1.3)` — v0.2 heuristic locked in; future swap to tiktoken is a conscious change.
- **First-index** path inserts the row with every column populated; ORM defaults fire; embedding called once.
[...]

PR #347, opened 2026-05-13
feat(retrieval): E2E smoke + db_migration_probe pgvector check + chart preflight (G0.4-T6)

### LINKED ISSUE #263: Task: G0.4-T6 E2E smoke + chart preflight (helm test) — verify pgvector enabled + hybrid fusion working + tenant boundary holds

ikaric:
## Summary

Lands **T6 of Initiative #225** (G0.4 Retrieval substrate) — the acceptance harness that proves T1-T5 work together against a real pgvector cluster. **Closes the Initiative.**

Closes #263 and brings #225 to Done.

### 1. Integration test

`backend/tests/integration/test_retrieval_e2e.py` (new) indexes 15 documents across 2 tenants via `index_document` against the testcontainer pgvector PG, then runs 6 retrieval tests + 1 probe test + an import smoke:

- **Basic retrieval** returns ordered hits; top result is kubernetes-related (not the wine-preference outlier).
- **RRF fusion correctness**: top hit has both `bm25_score` and `cosine_score` non-None (proves RRF is fusing, not falling back to a single signal).
- **Tenant boundary**: identical `(source, source_id)` in tenant-a and tenant-b stay disjoint — the unique composite index on `(tenant_id, source, source_id)` (migration `0003`) plus the `tenant_id` filter in every SQL guarantee no cross-tenant leak.
- **`source` / `kind` filters** narrow correctly.
- **`POST /api/v1/retrieve` end-to-end** through the FastAPI app (full auth + audit middleware stack): mints a real JWT, hits the route via ASGI in-process, reads back the audit_log row, asserts the payload carries `query_hash + source + kind + hit_count` but NOT the raw query.
- **`db_migration_probe` against a real pgvector-enabled PG** returns `ok=True` with `revision=<head>`.

Embedding is mocked (deterministic bag-of-words vectors) so each test runs in ~2s rather than the ~30s a real fastembed model load would need. PG-real coverage of fastembed itself lives in the slow tier of `tests/test_retrieval_embedding`.

### 2. Migration probe extension

`db/migrations.py` gains a `_check_pgvector_extension` helper. On PostgreSQL the probe runs `SELECT 1 FROM pg_extension WHERE extname='vector'` after the alembic head check passes. Failure detail is `revision=<sha> pgvector=missing` (revision still matches head — only the extension is gone). SQLite path is a no-op (gated on `engine.dialect.name == "postgresql"`); the catalog doesn't exist there. Catches the post-deploy drift mode where an operator manually dropped the extension or a backup restored the table without the catalog entry.

### 3. Helm test Job + CI workflow

- `deploy/charts/meho/templates/tests/test-pgvector.yaml` (new) is a `helm.sh/hook: test` Pod that runs `bitnami/postgresql:16` with `psql` to assert pgvector is enabled and the `documents` table is reachable.
- The `chart.yml` workflow adds a `helm-test` job that spins up kind, installs `pgvector/pgvector:pg16` as an in-cluster Service, helm-installs the chart with `--wait-for-jobs` (waits for the migration Job, not Pods — the backplane's readiness probe needs real Keycloak + Vault which we don't mock here), then runs `helm test`.

## CI workflow posture

The `helm-test` job ships with `continue-on-error: true` as the v0.2 transitional posture — it runs on every chart-affecting PR + push, surfaces failures to the operator, but does not block merge while operators iterate on kind/k3d wiring nuances. A follow-up issue (will file at PR-open time) hardens the workflow and removes the bypass once we have ~10 green runs.

## Bonus: cosine SQL bind fix

The retriever's cosine query was failing on real PG with `TypeError: expected str, got list` deep in asyncpg's text codec — the raw `text("... CAST($1 AS vector)")` SQL has no typed column adapter, so the bind variable must arrive pre-serialised in pgvector's wire format (`"[0.1, 0.2, ...]"`). Now serialises the embedding to that exact shape before binding; the integration test was the surface that caught this regression.

The conftest's `build_integration_app` now also includes `api_v1_retrieve_router` so the route is reachable via the production middleware stack the integration tests drive.

## Tests

- `backend/tests/test_alembic_probe.py` extended with 3 new tests: SQLite path skips the pgvector check; PG path reports `pgvector=missing` when the extension is gone; PG path returns `revision=<sha>` (success detail unchanged) when present.
- `backend/tests/integration/test_retrieval_e2e.py` new — 8 tests (6 Docker-gated retrieval E2E + 1 probe-vs-real-PG + 1 cheap import smoke that always runs).
- `backend/tests/integration/conftest.py` mounts the retrieve router on the integration_app (already on main from #345 / T5; T6 only adds the e2e test that consumes it).

## Validation

- [ ] `ruff check .` — clean (107 files).
- [ ] `mypy --strict src/` — clean across 53 source files.
- [ ] `pytest tests/` — 499 passed, 1 skipped (slow), 1 deselected (pre-existing `test_connectors_k8s_k3d::test_api_client_cached_across_calls_against_live_k3s` failure unrelated to this PR — landed by #328).
- [ ] CI green (modulo the `helm-test` job's `continue-on-error: true` transitional posture).

## Initiative wrap-up

This PR closes **Initiative #225** (G0.4 Retrieval substrate). All six tasks:

- T1 (#258) — pgvector substrate, merged via #269
- T2 (#259) — `EmbeddingService` + chart PVC, merged via #296
- T3 (#260) — `index_document` helper, merged via #298
- T4 (#261) — `retrieve` helper with RRF, merged via #299
- T5 (#262) — `POST /api/v1/retrieve` + audit payload, merged via #345
- T6 (#263) — this PR

Deferred follow-up: **MCP resource `meho://retrieve/{query}`** (G0.5-T7 candidate). Will file separately at PR-open time per the user decision to keep T5 scope tight.

## Summary by CodeRabbit

* **New Features**
  * Database health checks now validate pgvector extension presence.
  * Vector similarity search now serializes embeddings for reliable pgvector queries.

* **Tests**
  * Added end-to-end integration tests exercising retrieval against a pgvector-enabled database.
  * Added Helm/Kubernetes pgvector validation test and CI helm-test job to run chart deploy + helm test.

* **Chores**
  * CI updated to run pgvector-aware chart tests.

* **Documentation**
  * Backend docs updated to note pgvector requirements and chart test behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/347)


deploy/charts/meho/templates/tests/test-pgvector.yaml
```diff
@@ -0,0 +1,103 @@
...
+        - sh
+        - -c
+        - |
+          set -euo pipefail
+          # Step 1: pgvector extension must be enabled. The migration
+          # 0003 runs ``CREATE EXTENSION IF NOT EXISTS vector`` so a
+          # successful install implies presence; this check is the
+          # post-deploy drift guard.
+          extversion=$(psql "$PGURL" -t -A -c \
+            "SELECT extversion FROM pg_extension WHERE extname = 'vector'")
```
zdamir, line 85: **[🛑 B1]** — psql receives a SQLAlchemy-style DSN that libpq cannot parse, making the helm test pod incapable of verifying pgvector regardless of its actual state

**What:** Lines 69–70 (and the `psql "$PGURL"` invocation on line 83) consume `PGURL` directly from the chart's `postgres.credentialsSecret`. In production deploys and in the CI workflow at `.github/workflows/chart.yml:302`, that secret is constructed with the SQLAlchemy driver prefix — `postgresql+asyncpg://meho:test-pw-not-secret@pgvector:5432/meho`. libpq (the C library `psql` links against) accepts only `postgresql://` and `postgres://` URI schemes; the `+asyncpg` suffix is a SQLAlchemy-only dialect specifier. `psql "postgresql+asyncpg://..."` exits non-zero with a `missing "=" after` / `invalid URI scheme` error before any SQL runs.

**Why it matters:** This is an acceptance-criteria violation, not a style issue. Issue #263's AC#5 and AC#6 require that `helm test` against a deployed backplane (a) passes when pgvector is enabled and (b) fails with a clear "pgvector extension NOT enabled!" message when it isn't. Neither outcome is reachable as currently coded — psql aborts on the URI parse error before either `extversion=$(...)` or `count=$(...)` executes. The test pod is broken-by-construction; the chart-CI preflight cannot ever demonstrate the verification claim. The `continue-on-error: true` posture on the `helm-test` job (per `.github/workflows/chart.yml:223`) masks this in CI signal but doesn't change the substantive verification gap. Also flagged by CodeRabbit; promoted from Major to Blocker because it lands on the acceptance-criteria contract.

**Required change:** Normalize `PGURL` once at the top of the command block, then use the normalized variable in both psql calls. Minimal diff:

```diff
       command:
         - sh
         - -c
         - |
           set -euo pipefail
+          # libpq's URI parser rejects the SQLAlchemy ``+asyncpg`` driver
+          # suffix; strip any ``postgresql+<dialect>://`` prefix down to
+          # the bare libpq form before handing the DSN to psql.
+          PGURL_LIBPQ=$(printf '%s\n' "$PGURL" | sed -E 's|^postgresql\+[a-z0-9_-]+://|postgresql://|')
           # Step 1: pgvector extension must be enabled. The migration
           # 0003 runs ``CREATE EXTENSION IF NOT EXISTS vector`` so a
           # successful install implies presence; this check is the
           # post-deploy drift guard.
-          extversion=$(psql "$PGURL" -t -A -c \
+          extversion=$(psql "$PGURL_LIBPQ" -t -A -c \
             "SELECT extversion FROM pg_extension WHERE extname = 'vector'")
           ...
-          count=$(psql "$PGURL" -t -A -c "SELECT count(*) FROM documents")
+          count=$(psql "$PGURL_LIBPQ" -t -A -c "SELECT count(*) FROM documents")
```

The chart-side secret stays in the canonical asyncpg form (which is what the backplane Deployment + migration Job consume); the psql-only pod normalizes locally. Worth a one-line comment in the YAML explaining why the strip exists so a future contributor doesn't re-introduce the bug by "cleaning up" the sed.

**Verification once the fix is in place:** the `helm-test` job in `.github/workflows/chart.yml` should also have its real failure mode addressed — the current run failed at the `Helm install meho chart` step (the migration Job's image `ghcr.io/evoila/meho:test` isn't pushed on PR builds, so the kind cluster can't pull it). That's a separate workflow plumbing issue this comment doesn't try to scope, but B1's fix alone won't turn the chart-CI job green until the image-availability path is also resolved. See the review summary for the broader chart-CI observations.

**Source:** PostgreSQL libpq URI scheme reference — https://www.postgresql.org/docs/16/libpq-connect.html#LIBPQ-CONNSTRING-URIS (lists exactly `postgresql://` and `postgres://`).

zdamir (changes requested):




**Anchor task:** #263 — G0.4-T6 E2E smoke + chart preflight; closes Initiative #225
**Surface area:** backend (Python) + devops (Helm chart + GitHub Actions workflow)
**Best-practices packs audited:** python + devops (conventions inferred from neighbouring modules — no companion `python_best_practices.md` / `devops_best_practices.md` file is present in this checkout; the SKILL.md is the only artifact under each `.claude/skills/<name>/` directory)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `6a1e14825105f989a850b61d8099e4dd98bc572d`
**AI reviewer (CodeRabbit):** received in ~150s (state `COMMENTED`, 3 inline findings — all three confirmed and adopted; one promoted to Blocker, two left at Major and not duplicated inline)

### Acceptance criteria

- [x] AC#1 All 4 integration test cases pass against real Postgres in CI — `test_retrieval_e2e.py` has 6 retrieval cases + 1 probe + 1 import smoke, all PASSED locally against a real `pgvector/pgvector:pg16` testcontainer (8 passed, 4 sec wall clock).
- [x] AC#2 Fusion-correctness test demonstrates RRF doing something meaningful — `test_rrf_fusion_top_hit_has_both_signal_contributions` asserts top hit has both `bm25_score is not None` and `cosine_score is not None`, plus `fused_score > 1/61`.
- [x] AC#3 Tenant boundary test passes; cross-tenant document IDs disjoint — `test_tenant_boundary_holds_across_identical_natural_keys` covers the identical-`(source, source_id)` case, validating the migration-`0003` unique composite index + the `tenant_id` SQL filter together.
- [x] AC#4 `db_migration_probe` reports `vector_extension: ok` / `vector_extension: missing` — implementation lands in `db/migrations.py` with detail strings `revision=<sha>` (success) and `revision=<sha> pgvector=missing` (drift). The detail-string format diverges from the AC's `vector_extension: ok` / `vector_extension: missing` wording but the underlying contract (probe flips red on extension drift, structured detail) is satisfied; recording as a Minor in the result file because the wording delta is intentional (matches the existing `revision=<sha>` shape so downstream parsers don't branch).
- [ ] **AC#5 `helm test` passes when pgvector is enabled — NOT MET.** See 🛑 B1 below. The test pod's command consumes the chart-side secret's DSN (which is in SQLAlchemy `postgresql+asyncpg://` form) and hands it directly to `psql`, which libpq rejects with a URI-scheme error. The pod cannot succeed regardless of pgvector state.
- [ ] **AC#6 `helm test` fails with clear message when pgvector is missing — NOT VERIFIABLE.** Same root cause as AC#5 — psql aborts on URI parse before either pgvector assertion runs, so the "fails clearly when missing" half is unreachable too.
- [ ] **AC#7 CI integrates: `helm install --wait` followed by `helm test` is a *gating* step — NOT MET as written.** The `helm-test` job in `.github/workflows/chart.yml:223` carries `continue-on-error: true`. The PR body acknowledges this as a v0.2 transitional posture and commits to a follow-up issue ("will file at PR-open time") to harden the workflow + remove the bypass after ~10 green runs. The follow-up issue does not appear to have been filed (`gh search issues` against both `evoila/meho` and `evoila-bosnia/meho-internal` returned no matches for "chart-test", "helm-test", "pgvector preflight", or "continue-on-error helm"). Recording as a non-Blocker observation because the transitional posture is explicit in the PR body, but it should not stay unfiled while the Initiative closes.
- [x] AC#8 Backend ruff + mypy + pytest pass — ruff clean across `src/` + `tests/`, mypy clean across 53 source files (strict mode), pytest 508 passed / 1 skipped (slow fastembed) / 2 deselected (the same pre-existing Darwin `test_observability` gap from #345 and the pre-existing `test_connectors_k8s_k3d::test_api_client_cached_across_calls_against_live_k3s` failure landed by #328).

### Project-board alignment

|---|---|---|---|---|---|
| #263 | In Progress | `Closes` | ✅ | ⚠️ partial (helm-test verification gap per AC#5–#7) | close on merge only after B1 is resolved + follow-up filed |
| #225 (Initiative) | In Progress | PR body claim ("Closes the Initiative") + listing T1–T6 | manual | ⚠️ partial — six tasks delivered but the Initiative's "Definition of done" depends on the chart preflight working end-to-end | author should NOT manually close until the helm-test verification gap is resolved or explicitly waived |
| #262 (T5, prior PR #345) | In Review | `Closes` on PR #345 | ✅ | ✅ on #345 | unblocked by my prior approve; orchestrator's merge ordering keeps #345 before this PR |

The PR body claims "Closes Initiative #225" — that's not a load-bearing GitHub mechanism (initiatives close manually), but the implicit commitment is that the Initiative's DoD is satisfied. With AC#5–#7 unsatisfied (chart preflight broken-by-construction + non-gating + no follow-up filed), closing the Initiative now would silently drop the chart-CI hardening promise.

### Predicted vs. actual diff

(Predicted from issue #263's "Target" + AC list before re-reading the actual diff)

- ✅ `backend/tests/integration/test_retrieval_e2e.py` (new, 607 lines) — expected, exercises index → retrieve across two tenants with realistic corpora; covers basic / RRF / tenant-boundary / source-filter / kind-filter / route-end-to-end / probe-against-real-PG.
- ✅ `backend/src/meho_backplane/db/migrations.py` (+65/-5) — expected, extends `db_migration_probe` with `_check_pgvector_extension` helper gated on `engine.dialect.name == "postgresql"`.
- ✅ `backend/tests/test_alembic_probe.py` (+104) — expected, adds 3 unit tests for the new pgvector-check branch (SQLite no-op, PG missing, PG present).
- ✅ `backend/tests/integration/conftest.py` (+3) — expected, mounts `api_v1_retrieve_router` on the integration_app so the e2e test can drive `POST /api/v1/retrieve` through the production middleware stack.
- ✅ `deploy/charts/meho/templates/tests/test-pgvector.yaml` (new, 103 lines) — expected at the file-level; **broken-by-construction at the SQL-execution level** (see B1).
- ✅ `.github/workflows/chart.yml` (+158) — expected, adds the `helm-test` job; **shipped as non-gating** (`continue-on-error: true`) which is acknowledged-transitional but flagged as an AC-as-written violation (AC#7).
- ✅ `backend/src/meho_backplane/retrieval/retriever.py` (+12/-1) — unexpected at issue-time but justified: the PR body explains this is the "cosine SQL bind fix" caught by running the new integration test against a real PG. Pre-serialising the embedding to pgvector wire-format (`"[0.1, 0.2, ...]"`) before binding to `CAST(:emb AS vector)` is correct per pgvector docs and asyncpg's text-codec contract; without it, the cosine query fails with `TypeError: expected str, got list` (the original bug). Worth flagging that this fix retroactively means T4 (#261, #299) shipped to main with a bug that only manifested under PG-real coverage; the regression is now fixed.
- ✅ `docs/codebase/backend.md` (+2/-1) — expected, documents the pgvector check + helm test in the existing key-types table.

No undisclosed scope additions. No missing files.

### Integration verification

All commands ran inside `.claude/worktrees/pr-347-review/backend/` against head `6a1e14825`:

- [x] `uv run ruff check src/ tests/` — All checks passed.
- [x] `uv run mypy src/` — Success: no issues found in 53 source files.
- [x] `uv run pytest tests/test_alembic_probe.py` — 9 passed (including the 3 new pgvector-branch tests).
- [x] `uv run pytest tests/integration/test_retrieval_e2e.py` — **8 passed in 4 sec** against a real `pgvector/pgvector:pg16` testcontainer (Docker available on the review host; the class-level `@_skip_no_docker` falls back to CI when absent).
- [x] `uv run pytest tests/` (full suite minus the same two pre-existing failures from #345 review) — 508 passed, 1 skipped, 2 deselected.
- [⚠️ FAILED] CI `helm-test` job — completed with `failure` conclusion. Root cause is upstream of B1: the `Helm install meho chart` step never completed because the migration Job's image (`ghcr.io/evoila/meho:test`) isn't pushed on PR builds, so kind can't pull it. The job's `continue-on-error: true` posture means this didn't block the workflow signal; the substantive verification gap stands.
- [⚠️ FAILED] CI SonarCloud — 1 Security Hotspot flagged. Team policy presumably gates merge; I haven't inspected the dashboard (would need Sonar auth). Worth the author checking.
- [n/a] migrations replay against fresh DB — `db_migration_probe`'s pgvector check is covered by both unit tests (mocked) and an integration test (real PG); no separate replay needed.
- [n/a] docker compose up boots clean — repo doesn't use docker-compose for this surface.
- [n/a] touched UI flow exercised — no frontend changes.
- [n/a] agent trace through changed reasoning path — no agent-path changes.

- 🛑 **B1** (this review, posted inline at `deploy/charts/meho/templates/tests/test-pgvector.yaml:70`) — `psql` consumes the chart-side `postgresql+asyncpg://` DSN unchanged; libpq rejects the URI scheme; helm test pod is broken-by-construction; **AC#5 + AC#6 unsatisfiable as currently coded**. Required change: normalize PGURL by stripping the `+<dialect>` suffix once at the top of the command block, then use the normalized variable in both `psql` calls. Suggested sed lands inline; the chart-side secret stays in asyncpg form for the backplane Deployment + migration Job, only the psql-only test pod normalizes locally.
- ⚠️ **M1** (CodeRabbit, posted inline at `backend/tests/integration/test_retrieval_e2e.py:278`) — `_make_stub_embedding_vector` uses Python's built-in `hash(token)`, which is randomized per process via `PYTHONHASHSEED`. Ranking assertions over cosine-derived signals are vulnerable to flakiness across CI runs. CodeRabbit's suggested `hashlib.blake2b(token.encode("utf-8"), digest_size=8)` is the right shape. Confirmed by running `PYTHONHASHSEED=1/2/random python -c 'print(hash("kubernetes"))'` — three distinct values. Adopted; not duplicating CodeRabbit's inline.
- ⚠️ **M2** (CodeRabbit, posted inline at `backend/tests/integration/test_retrieval_e2e.py:601-602`) — `assert Document is not None` and `assert AuditLog is not None` are tautological after import; static analysis flags them (SonarCloud confirms with the same calibration). CodeRabbit's suggestion (`assert Document.__tablename__ == "documents"` / `"audit_log"`) is the right substantive substitute. Adopted; not duplicating.

Items the author may pick up in a future PR. None of these block merge or require a response on this PR — they're surfaced here so they're recorded in the review trail without obligating the author and without polluting the continue-mode punch list.

- The CI `helm-test` job's broader failure (Helm install step) has a separate root cause from B1: the migration Job's image `ghcr.io/evoila/meho:test` isn't pushed on PR builds. Two paths forward — (a) push a per-PR tag to ghcr behind a workflow gate, or (b) load the locally-built image into the kind cluster via `kind load docker-image` after the existing "Build backplane image" job completes. Either way, B1's fix alone won't turn the chart-CI job green until image availability is also resolved.
- AC#7 says the chart-CI step must be "gating"; the PR ships it as `continue-on-error: true`. The transitional posture is explicit in the PR body and reasonable for new infra, but the committed follow-up issue (per "will file at PR-open time" in the PR body) does not appear to have been filed yet. File the follow-up before merging so the Initiative doesn't close while the chart-CI hardening promise sits unfiled.
- AC#4's detail-string wording is `vector_extension: ok` / `vector_extension: missing` per the issue body; the implementation ships `revision=<sha>` / `revision=<sha> pgvector=missing`. The functional contract is satisfied (probe flips red on drift, detail is structured) and the wording is consistent with the existing `revision=<sha>` shape, so downstream operator tooling doesn't have to branch on dialect. Worth a one-line note in the issue body that the wording was tightened rather than the AC drifting silently.
- T4's cosine SQL was buggy on real PG (the asyncpg text-codec mismatch the author fixed in this PR). Since T4 (#261, #299) shipped to main without integration-test coverage exposing it, the right reflection is to add a chart-CI gate that runs a tiny PG-real smoke against `retrieve()` on every PR touching `retrieval/` — exactly what this PR's `test_retrieval_e2e.py` does — but running unconditionally in CI rather than gated on Docker availability. Outside the scope of T6; worth filing as a separate hardening Task.
- SonarCloud Security Hotspot — 1 open finding on this PR. Couldn't inspect the dashboard from this review (would need Sonar auth). Worth a manual check before merge.
- The integration test's `test_retrieve_route_end_to_end_with_audit_payload` happens to exercise the route #345 added; that's why it lives in this PR rather than #345. Pure organizational note, no action.

- Adopted: 3.
  - M1 (hash randomization) — left at CodeRabbit's Major severity; not duplicated inline (their comment already creates a merge-blocker thread per team policy).
  - M2 (tautological asserts) — same. SonarCloud also flags.
  - B1 (psql DSN) — CodeRabbit marked it Major; promoted to Blocker on AC-violation grounds and posted my own inline comment (referencing CodeRabbit) so the punch list reflects the elevated severity for `/auto-implement-issue --continue-pr`.
- Disputed: 0.
- Out of scope: 0.

### Docs

- [x] `docs/codebase/backend.md` — updated. New row documenting the pgvector check + helm test sits in the key-types table next to the existing `db_migration_probe` entry; no .NET analogies; structurally consistent with surrounding rows.

- **The migration-probe extension's dialect gate is exactly the right shape.** `if engine.dialect.name == "postgresql"` keeps the SQLite dev/test driver free of pg-extension semantics it doesn't understand, while the helper itself is a tiny one-liner that's easy to unit-test (3 cases land cleanly). The detail-string discipline matches the existing chassis pattern (`revision=<sha>` for success, structured `revision=<sha> pgvector=missing` for the drift case) — operators parsing the detail field don't have to branch on dialect.
- **The cosine SQL bind fix is well-explained and well-located.** Pre-serialising the embedding to pgvector wire-format (`"[0.1, 0.2, ...]"`) before binding to `CAST(:emb AS vector)` is the documented pgvector pattern + asyncpg-text-codec workaround; the in-code comment names the exact failure mode (`TypeError: expected str, got list`) so a future contributor doesn't strip the serialisation "back to a list" thinking it's defensive bloat.
- **The integration test corpus is well-shaped for RRF discrimination.** 15 docs across two tenants with a deliberately-confounding outlier (wine preferences) gives the `test_retrieve_returns_hits_against_real_pg` assertion something real to check (top hit must be kubernetes-themed, NOT the wine outlier). The two-tenant `identical (source, source_id)` setup directly probes the unique composite index from migration 0003.
- **The lifespan-skip rationale in `build_integration_app` is documented in-place** — `configure_logging` would clobber test-scoped structlog capture, and the engine cache is overridden by `pg_engine` ahead of any request, so the lifespan is deliberately omitted from the integration app. Future contributors writing similar test apps won't re-add the lifespan and then chase the resulting fd-flake.

ikaric:
## Review feedback addressed — pushed in `f49c181`

Force-push of T6 (with-lease). All three findings resolved:

### Blocker (Damir B1)

- **psql DSN scheme — `deploy/charts/meho/templates/tests/test-pgvector.yaml`.** The pod was receiving the chart-side secret's SQLAlchemy-form DSN (`postgresql+asyncpg://...`); libpq rejects the `+asyncpg` driver suffix and aborted before any SQL ran, breaking AC#5 / AC#6 by construction. Now normalises locally before `psql` invocations:

  ```sh
  PGURL_LIBPQ=$(printf '%s\n' "$PGURL" | sed -E 's|^postgresql\+[a-z0-9_-]+://|postgresql://|')
  ```

  The chart-side secret stays in canonical asyncpg form (single source of truth — the backplane Deployment + migration Job both consume it via SQLAlchemy); the strip is local to this preflight pod. A multi-line comment in the YAML explains why so a future contributor doesn't "clean up" the sed.

### CodeRabbit majors

- **`hash(token)` PYTHONHASHSEED flakiness — `tests/integration/test_retrieval_e2e.py` `_make_stub_embedding_vector`.** Replaced with `hashlib.blake2b(token.encode("utf-8"), digest_size=8)` → `int.from_bytes(..., "big")`. Stable across processes and CI runners; preserves the same modulo + weighting logic so ranking semantics are unchanged. Docstring extended to call out why.
- **Tautological `is not None` assertions — `tests/integration/test_retrieval_e2e.py:601-602`.** Replaced with `Document.__tablename__ == "documents"` / `AuditLog.__tablename__ == "audit_log"`. The new check actually pins the table-name contract migration `0003` set up rather than being a no-op after import. Clears Sonar's `python:S5797`.

### Notes on Damir's other observations

- **AC#7 / `continue-on-error: true` on the `helm-test` job** — the hardening follow-up is filed at https://github.com/evoila/meho/issues/349 (created at PR-open time alongside the G0.5-T7 follow-up at https://github.com/evoila/meho/issues/348). The transitional posture is intentional per the T6 plan's fallback clause and tracked publicly now, not implicit.
- **AC#5 / AC#6 verification gap** — with B1 fixed, the helm-test pod can now actually exercise both branches (pgvector present → "OK"; pgvector missing → the clear "FAIL: pgvector extension is NOT enabled" message). The image-availability path Damir flagged (the kind cluster needs a real backplane image to pull) is a separate workflow-plumbing issue — that's exactly what #349 will harden.

### Validation

- `ruff check .` clean (107 files).
- `ruff format --check .` clean.
- `mypy --strict src/` clean across 53 source files.
- `pytest tests/` — 499 passed (matches pre-amend count; the blake2b swap preserves test determinism), 1 skipped (slow), 1 deselected (pre-existing `test_connectors_k8s_k3d` failure unrelated to this PR — landed by #328).
- `helm-test` YAML normalisation spot-checked: `PGURL_LIBPQ` is computed once, both `psql` invocations use it, no stale references remain.

ikaric:
## SonarCloud hotspot cleared — pushed in `db5624a`

The `python:S5332` security hotspot at `tests/integration/test_retrieval_e2e.py:535` was a false positive on the `base_url="http://testserver"` label for the in-process `ASGITransport`. The string is a label, not a transport choice -- the FastAPI app runs in-process, no socket is opened, no network traffic flows.

Switched to `base_url="https://testserver"` to match the convention `tests/integration/test_mcp_inspector.py` already established (#329 fixed the same Sonar gate the same way). Added an inline comment explaining the rationale so a future contributor doesn't "fix" it back to http.

All 8 e2e tests still pass; no logic change. Quality Gate should turn green on the next scan.

zdamir (approved):

Issue #333, opened 2026-05-12
Initiative: G6.2 Slack mirror — second subscriber to G6.1's Valkey stream; per-tenant channel routing + Block Kit cards + Vault-backed creds


damir-topic:
Parent goal: #217

## Scope

Ship the **Slack mirror** for the activity broadcast — a second subscriber to G6.1's per-tenant Valkey stream that posts every operation to a Slack channel as a structured card. Same event source as the SSE feed (#228); different presentation surface.

**Area:** new module [backend/src/meho_backplane/broadcast/slack/](backend/src/meho_backplane/broadcast/slack/), new router [backend/src/meho_backplane/api/v1/slack.py](backend/src/meho_backplane/api/v1/slack.py) for Slack signature verification + slash-command handling (Phase 2 polish — minimal in v0.2), settings extensions, optional helm-chart sidecar deployment for the long-running Slack subscriber loop.

## Work items grouped under this Initiative

1. **Slack Bolt SDK wiring.** Add `slack-bolt` (async) + `slack-sdk` to backend dependencies. Slack signing-secret + bot-token settings per-tenant — stored in the `tenant.extras` JSONB (G0.1's primitive doesn't have this; T1 might add the `tenant.slack_config` JSONB).

2. **Per-tenant subscriber loop.** A long-running asyncio task started in `lifespan` that, for each tenant with `slack_config.enabled=true`, opens an `XREAD BLOCK 30000 STREAMS meho:feed:<tenant_id> $` subscription and posts each event to the tenant's configured Slack channel.

3. **Structured card format.** Each event → a Slack Block Kit message:
   - Header: `op_id` + `result_status` icon (✓ / ⚠ / ✗).
   - Operator: `<@U...>` Slack mention if known, else `<principal_sub>@<tenant>`.
   - Target: name + (for target-aware ops) a link to `meho targets describe`.
   - Payload summary: full params for `read` / `write` op_class; `(aggregate-only)` placeholder for `credential_read` / `audit_query` per decision #3.
   - Audit ID footer: `audit_id: <uuid>` for cross-reference with `meho audit query`.

4. **Slack-to-MEHO operator mapping.** Tenants configure `slack_principal_map` in `slack_config`: a list of `{sub: <jwt-sub>, slack_user: <U...>}` entries. The mirror looks up the principal's Slack handle for `<@U...>` mentions; falls back to the JWT sub if unmapped.

[...]

Issue #312, opened 2026-05-12
Task: G6.1-T6 MCP resource (meho://tenant/{id}/feed) + load-test acceptance + docs/cross-repo/broadcast-onboarding.md


damir-topic:
## ⚠️ Re-scope notice (2026-05-14)

Parent Initiative [#228 G6.1 was amended](https://github.com/evoila/meho/issues/228). Three new broadcast meta-tools added to scope: `broadcast_recent(since, filter?)`, `broadcast_announce(activity, target?, scope?)`, `broadcast_watch(filter?)`. These register against G0.5's MCP server (handlers live in G6.1) and are the load-bearing surface for the consumer-side agent discipline documented in G7.1's Layer-2 onboarding template.

This Task's existing scope (MCP resource / load test) is structurally unchanged; the broadcast meta-tools land as additional Tasks (TBD) on this Initiative's list. Body below preserved for reference.

---

Parent initiative: #228
Depends on: G6.1-T1, G6.1-T2, G6.1-T3, G6.1-T4, G0.5-T3 (MCP resource registry).

## Target

MCP resource exposing the feed + load-test acceptance:

- `backend/src/meho_backplane/mcp/resources/tenant_feed.py` — registers `meho://tenant/{tenant_id}/feed`.
- `backend/tests/integration/test_g6_1_broadcast_e2e.py` — end-to-end test exercising publish → SSE consumer → load shape.
- `docs/cross-repo/broadcast-onboarding.md` — operator-facing recipe.

## Context

T6 is the **acceptance harness** + the **MCP path** for the broadcast. Two pieces:

### MCP resource

```python
register_mcp_resource(
    ResourceTemplateDefinition(
        uriTemplate="meho://tenant/{tenant_id}/feed",
        name="Tenant activity feed",
        description="Live activity feed for the operator's tenant. Returns the last N events; clients poll for new events.",
        mimeType="application/json",
        required_role=TenantRole.OPERATOR,
    ),
    handler=_tenant_feed_handler,
)
```

The handler reads recent events from the Valkey stream and returns a snapshot. Resource subscriptions (where the spec supports them) would push live; v0.2 advertises `subscribe: false` so MCP clients poll.

### Load-test acceptance

The integration test:
1. Spins up backplane + Valkey + Keycloak + Postgres.
2. Creates two test operators in different tenants.
3. Runs 50 RPS audit-row writes for 30 seconds (1500 events total).
4. Asserts:
   - All 1500 events appear in the SSE feed.
[...]

Issue #311, opened 2026-05-12
Task: G6.1-T5 meho status --watch CLI verb — SSE client + filters + retry-with-backoff + Last-Event-Id replay


damir-topic:
Parent initiative: #228
Depends on: G6.1-T4.

## Target

`meho status --watch` CLI verb — subscribes to the SSE feed, renders events as one structured line each:

- `cli/internal/cmd/status.go` — extends existing command with `--watch` flag.
- `cli/internal/cmd/status_watch.go` — watch-mode renderer.
- `cli/internal/cmd/status_test.go` — extended tests.

## Context

T5 is the **terminal-resident operator dashboard**. The operator runs `meho status --watch` in a terminal pane; events appear in real time as their colleagues work.

**Output format (default):**
```
2026-05-12T14:23:01Z  damir@rdc-internal  vsphere.vm.list   ok      target=rdc-vcenter
2026-05-12T14:23:04Z  ilhan@rdc-internal  vault.kv.read     ok      (aggregate-only)
2026-05-12T14:23:09Z  dzenan@rdc-internal vsphere.host.list ok      target=rdc-vcenter
```

Columns: timestamp / principal@tenant / op_id / result_status / payload-summary.

`(aggregate-only)` appears in place of params for credential_read / audit_query op classes — operator sees that work happened but not the details.

**Output format (`--json`):**
Each event as one JSON line; raw `BroadcastEvent` shape from the SSE feed. Useful for `meho status --watch --json | jq` pipelines.

**Filter flags forward to SSE query params:**
- `--op-class read` → `?op_class=read`
- `--principal damir` → `?principal=damir`
- `--target rdc-vcenter` → `?target=rdc-vcenter`

**SSE client implementation** uses Go's stdlib HTTP client + manual SSE parser (Go's SSE library landscape is mediocre; rolling a small parser is cheaper than depending on a third-party). On disconnect, retry with exponential backoff (1s, 2s, 5s, 10s, 30s max) using `Last-Event-Id` for replay.

**Authentication:** uses the token from `meho login` (already in keyring/credentials.json from chassis).

## Current state

After T1–T4: SSE feed works. CLI has `meho status` (one-shot health check) but no `--watch`.

## Desired state

```go
// status.go extension
func newStatusCmd() *cobra.Command {
    var watch bool
    var opClass, principal, target string
    cmd := &cobra.Command{
        Use:   "status",
        Short: "Show MEHO instance status; --watch streams live activity",
[...]

Issue #228, opened 2026-05-11
Initiative: G6.1 SSE feed core — per-tenant Valkey Streams + audit publish-on-write + meho status --watch + MCP resource (conservative PII defaults)


damir-topic:
Parent goal: #217

## Scope

Stand up the activity-broadcast core: per-tenant SSE feed at `/api/v1/feed`, backed by the Valkey Streams subchart already shipped with the chassis. Wire the audit-write path to publish a broadcast event per operation. Ship the `meho status --watch` CLI verb that subscribes to the feed. Conservative PII defaults from decision #3: credential reads + audit queries broadcast aggregate-only; other ops broadcast in full. This is **Phase-1 Track B** of the locked v0.2 sequencing.

Slack mirror and the threaded-context UX are separate Initiatives under G6 (G6.2 Slack mirror, G6.3 PII opt-in/opt-out controls). G6.1 lands the core.

**Area:** new module [backend/src/meho_backplane/broadcast/](backend/src/meho_backplane/broadcast/), new router under [backend/src/meho_backplane/api/v1/feed.py](backend/src/meho_backplane/api/v1/feed.py), modifications to [backend/src/meho_backplane/audit.py](backend/src/meho_backplane/audit.py) (publish-on-write), [cli/internal/cmd/status.go](cli/internal/cmd/status.go) (gets `--watch` mode).

The work items grouped under this Initiative:

1. **Valkey Streams wiring.** The broadcast subchart (Valkey 9.x) is already deployed but unused. Settings adds `BROADCAST_REDIS_URL` (default `redis://<release>-broadcast:6379`). A single async Valkey client at module scope (similar to the engine pattern in [db/engine.py](backend/src/meho_backplane/db/engine.py)); shutdown hook in lifespan disposes cleanly.
2. **Stream-per-tenant.** Stream key: `meho:feed:<tenant_id>` (UUID). Per-tenant boundary holds — no cross-tenant subscribers; the SSE endpoint validates the requesting JWT's `tenant_id` against the streams it'll read from.
[...]

PR #353, opened 2026-05-13
feat(broadcast): SSE feed endpoint at GET /api/v1/feed (G6.1-T4)

### LINKED ISSUE #310: Task: G6.1-T4 SSE endpoint /api/v1/feed — per-tenant filter + replay via Last-Event-Id + heartbeat
Parent initiative: #228
Depends on: G6.1-T1, G6.1-T2, G6.1-T3.

## Target

SSE endpoint at `/api/v1/feed` + per-tenant filter + replay window:

- `backend/src/meho_backplane/api/v1/feed.py` — new router. `GET /api/v1/feed` returns `text/event-stream`.
- `backend/src/meho_backplane/main.py` — register the router.
- `backend/tests/test_api_v1_feed.py` — integration tests.

## Context

T4 ships the **operator-facing SSE surface**. Subscribers (CLI watch, Slack bot, future web dashboard) consume this endpoint.

**SSE shape per RFC EventSource:**

```
event: broadcast
data: {"event_id":"...", "ts":"...", "tenant_id":"...", ...}
id: <stream-entry-id-from-Valkey>

event: broadcast
data: {...}
id: ...
```

The `id:` field carries the Valkey stream entry id; clients reconnecting with `Last-Event-Id: <id>` header get replay from that point.

**Endpoint behavior:**

```
GET /api/v1/feed?op_class=read&principal=damir&since=<entry-id>
Accept: text/event-stream
Authorization: Bearer <jwt>
Last-Event-Id: <optional>
```

- Auth required (operator role minimum).
- `op_class`, `principal`, `target` are server-side filters (post-stream).
- `since=<entry-id>` enables explicit replay from a known point.
- `Last-Event-Id` header is the standard SSE reconnect mechanism; preferred over `since` when present.
- Heartbeat every 30s as a comment line (`: heartbeat\n\n`) to keep proxies happy.

**Implementation:** FastAPI `StreamingResponse` with `text/event-stream`. The handler runs `XREAD BLOCK 30000 STREAMS meho:feed:<tenant_id> $`, filters, formats SSE, and yields.

**Tenant scoping:** the JWT determines the tenant; cross-tenant subscription is impossible by construction (the stream key includes the tenant_id from the JWT).

**Client disconnect handling:** when the client disconnects, the generator raises `asyncio.CancelledError`; the handler catches and closes the Valkey subscription cleanly.

## Current state

After T1 + T2 + T3: events flow into Valkey streams. No SSE endpoint exists.

## Desired state

```python
# api/v1/feed.py (sketch)
@router.get("/feed")
async def feed_endpoint(
    op_class: str | None = Query(default=None),
    principal: str | None = Query(default=None),
    target: str | None = Query(default=None),
    since: str | None = Query(default=None),
    operator: Operator = Depends(verify_jwt_and_bind),
    request: Request = Depends(),
) -> StreamingResponse:
    last_event_id = request.headers.get("Last-Event-Id") or since
    start_id = last_event_id or "$"
    stream_key = f"meho:feed:{operator.tenant_id}"
    client = get_broadcast_client()

    async def generator():
        last_heartbeat = time.monotonic()
        cursor = start_id
        while True:
            try:
                entries = await client.xread(
                    {stream_key: cursor},
                    block=30000,  # 30s
                    count=20,
                )
                if entries:
                    _, items = entries[0]
                    for entry_id, fields in items:
                        cursor = entry_id
                        event = BroadcastEvent.model_validate_json(fields["event"])
                        if not _passes_filter(event, op_class, principal, target):
                            continue
                        yield f"event: broadcast\ndata: {fields['event']}\nid: {entry_id}\n\n"
                # heartbeat if idle
                if time.monotonic() - last_heartbeat > 30:
                    yield ": heartbeat\n\n"
                    last_heartbeat = time.monotonic()
            except asyncio.CancelledError:
                break

    return StreamingResponse(
        generator(),
        media_type="text/event-stream",
        headers={"Cache-Control": "no-cache, no-store, must-revalidate", "X-Accel-Buffering": "no"},
    )
```

## Acceptance criteria

- [ ] `GET /api/v1/feed` with valid JWT returns 200 + `text/event-stream`.
- [ ] Events published to tenant-A's stream appear on tenant-A's SSE connection.
- [ ] Tenant boundary: tenant-A's events do NOT appear on tenant-B's SSE connection.
- [ ] Filter by `op_class=read` returns only read-class events.
- [ ] Filter by `principal=<sub>` filters correctly.
- [ ] Replay: `since=<entry-id>` returns events after that id.
- [ ] `Last-Event-Id` header takes precedence over `since` query param when both are present.
- [ ] Heartbeat appears every 30s during quiet periods.
- [ ] Client disconnect cleanly closes the Valkey subscription (verified by counting open Valkey connections).
- [ ] Unauthenticated → 401.
- [ ] `read_only` role → 403 (operator+ required).
- [ ] CI green.
- [ ] [#228](https://github.com/evoila/meho/issues/228) body checklist line for T4 is ticked.

## Out of scope

- WebSocket transport (we use SSE; WS is overkill for read-only stream).
- Server-Sent Events compression (text/event-stream isn't gzip-friendly).
- Per-tenant rate limiting on the feed endpoint.
- Subscription pause/resume.

## References

- Parent Initiative: [#228](https://github.com/evoila/meho/issues/228).
- Depends on: T1, T2, T3.
- SSE / EventSource spec: https://html.spec.whatwg.org/multipage/server-sent-events.html
- Valkey XREAD: https://valkey.io/commands/xread/
- FastAPI StreamingResponse: https://fastapi.tiangolo.com/advanced/custom-response/#streamingresponse

zdamir:
## Summary

Closes #310 — T4 ships the operator-facing SSE surface that streams the events T3's publish-on-write hook ``XADD``s onto the per-tenant Valkey stream. Subscribers (T5 CLI ``--watch`` #311, T6 MCP resource #312, future Slack mirror G6.2 #333) open one HTTP connection per session and receive events as they're published.

- **`api/v1/feed.py` (new)** — `GET /api/v1/feed` returns `text/event-stream`. Generator issues `XREAD BLOCK 30000` against `meho:feed:{operator.tenant_id}` (stream key derived from the JWT — cross-tenant subscription impossible by construction), filters post-stream by optional `op_class` / `principal` / `target` query parameters, emits SSE frames `event: broadcast\ndata: <BroadcastEvent JSON>\nid: <Valkey entry id>\n\n`. Replay via SSE-standard `Last-Event-Id` header (preferred) or `since` query parameter; default cursor is `$` (live-tail). Heartbeat `: heartbeat\n\n` every 30s of inactivity. RBAC = operator role minimum.
- **Disconnect handling** — client disconnect surfaces as `asyncio.CancelledError`; the generator catches + returns silently so the audit row at session end records a clean 200 close.
- **Defensive parsing** — entries with malformed JSON or an unknown field shape are logged + skipped rather than tearing down the subscriber.

## Test plan

- [x] `ruff format --check src tests` — clean (109 files)
- [x] `mypy src` strict — 56 source files, no issues
- [x] `pytest tests/test_api_v1_feed.py` — **14 passed** across three layers
- [x] `pytest --ignore=tests/integration` — **552 passed** (+14 vs T3 baseline), 1 skipped, 1 deselected
- [x] AC #1-2 — 200 + `text/event-stream` for authenticated operator, JWT-derived stream key — TestFeedEndpoint + TestFeedGenerator
- [x] AC #3-5 — Tenant boundary + filter by `op_class` / `principal` / `target` — TestFeedGenerator parametric cases
- [x] AC #6-7 — `Last-Event-Id` > `since` > `$` precedence — TestResolveCursor + TestFeedGenerator
- [x] AC #8 — Heartbeat shape pinned via the generator unit tests; the long-idle window is not exercised in CI (would slow the suite by 30s per case) but documented in the module
- [x] AC #9 — Disconnect cleanup — exercised by every test that drains the generator and calls `aclose()`; the `except asyncio.CancelledError: return` arm is the load-bearing fix
- [x] AC #10-11 — Unauthenticated → 401, read_only → 403 — TestFeedEndpoint

## Test architecture note

Body-shaping ACs (filters, cursors, malformed-entry skip) are tested at the **generator layer** rather than via httpx ASGI streaming. The httpx + ASGITransport cancellation handshake on `async with response.stream(...)` exit is racy enough that asserting on streamed bytes tight-loops the generator (the mocked `xread` has to `await asyncio.sleep` to give cancellation a chance, and even then the chain is timing-dependent). Async-generator `aclose()` is deterministic — the cleanest cancellation point in the system. HTTP-edge tests cover authn / authz / the 200-content-type happy header; everything else lands at the generator layer.

## Out of scope

- WebSocket transport — SSE is the right shape for read-only streams.
- Per-tenant rate limiting on the feed endpoint.
- Subscription pause/resume.

## Post-merge follow-up

- Tick T4 line on Initiative #228 body checklist.

Closes #310

## Summary by CodeRabbit

* **New Features**
  * Added a real-time Server-Sent Events feed: tenant-scoped streams, replay cursors (Last-Event-Id > since, default $), optional filters (op_class, principal, target), heartbeat comment frames, and operator-level RBAC (401/403). Graceful disconnects are logged.

* **Tests**
  * Added unit/generator-level and docker-gated integration tests covering auth, cursor resolution/advancement, filtering, heartbeat, malformed events, and end-to-end publish/read.

* **Documentation**
  * Expanded backend docs with feed behavior, SSE format, cursor semantics, heartbeat, and RBAC.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/353)

### DIFF
--- a/backend/src/meho_backplane/api/v1/feed.py
+++ b/backend/src/meho_backplane/api/v1/feed.py
@@ -94,7 +94,7 @@ from __future__ import annotations
 
 import asyncio
 import time
-from collections.abc import AsyncIterator
+from collections.abc import AsyncIterator, Iterator
 from typing import Final
 
 import structlog
@@ -223,6 +223,59 @@ def _resolve_cursor(
     return _LIVE_TAIL_CURSOR
 
 
+def _process_entries(
+    items: list[tuple[str, dict[str, str]]],
+    *,
+    op_class: str | None,
+    principal: str | None,
+    target: str | None,
+    stream_key: str,
+) -> Iterator[tuple[str, str]]:
+    """Yield ``(entry_id, sse_frame)`` for every entry that passes the filter.
+
+    Handles three skip paths inline:
+
+    * Unknown field shape (entry XADD'd without an ``event`` field) —
+      log + skip. T3's publisher is currently the only writer; this
+      branch is the safety net against a future Slack-mirror /
+      downstream tool writing alternate field shapes onto the same
+      stream key.
+    * Malformed JSON in the ``event`` field — log + skip rather than
+      tearing the subscriber down. A T3 bug or stream-key collision
+      with a foreign writer surfaces here as a logged warning, not a
+      500.
+    * Filter rejection — silently drop (the operator's filter is
+      working as intended).
+
+    Lifted out of :func:`_feed_generator` so the main loop's
+    cognitive complexity stays under the SonarCloud S3776 ceiling.
+    The helper itself is a single ``for``-loop with three
+    ``continue`` arms.
+    """
+    for entry_id, fields in items:
+        raw_event_json = fields.get("event")
+        if not isinstance(raw_event_json, str):
+            _log.warning(
+                "feed_skipped_unknown_field_shape",
+                stream_key=stream_key,
+                entry_id=entry_id,
+                fields=list(fields.keys()),
+            )
+            continue
+        try:
+            event = BroadcastEvent.model_validate_json(raw_event_json)
+        except ValidationError:
+            _log.warning(
+                "feed_skipped_malformed_event",
+                stream_key=stream_key,
+                entry_id=entry_id,
+            )
+            continue
+        if not _passes_filter(event, op_class, principal, target):
+            continue
+        yield entry_id, _format_event(entry_id, raw_event_json)
+
+
 async def _feed_generator(
     operator: Operator,
     cursor: str,
@@ -230,14 +283,28 @@ async def _feed_generator(
     principal: str | None,
     target: str | None,
 ) -> AsyncIterator[str]:
-    """SSE generator: BLOCK on XREAD, filter, format, heartbeat.
+    """SSE generator: BLOCK on XREAD, delegate parsing, heartbeat-on-silence.
 
-    Cancelled cleanly on client disconnect (Starlette raises
-    :class:`asyncio.CancelledError` into the await point). The
-    ``except`` arm returns silently rather than propagating so the
-    surrounding StreamingResponse doesn't synthesise a 500-shaped
-    audit row at session end — a normal SSE close lands as a clean
-    200.
+    Heartbeat semantics — ``last_heartbeat`` tracks the wall-clock of
+    the **last outbound yield** (event frame or heartbeat), NOT the
+    last inbound XREAD result. A noisy tenant where every event is
+    filtered out for this subscriber still produces zero outbound
+    bytes; without this guarantee the connection would idle-timeout
+    at the nginx / ALB / CloudFront layer. The "all entries filtered
+    out" path therefore emits an inline heartbeat when the idle
+    window has elapsed — both quiet and busy-but-filtered tenants
+    keep the connection alive.
+
+    On client disconnect Starlette raises
+    :class:`asyncio.CancelledError` into the pending ``xread`` await;
+    the handler logs and re-raises per the asyncio cancellation
+    contract (Sonar S7497 — swallowing CancelledError breaks the task
+    tree's unwind invariants and Python 3.13+ asyncio internals re-
+    issue cancellation when it goes unpropagated). The audit row at
+    session end still records a clean 200 close because
+    ``http.response.start`` was sent on the first yield (before any
+    cancellation point), so AuditMiddleware's buffered status_code
+    is already locked at 200.
     """
     client = get_broadcast_client()
     stream_key = _stream_key(operator)
@@ -251,6 +318,7 @@ async def _feed_generator(
                 count=_XREAD_COUNT,
             )
             now = time.monotonic()
+            emitted_any = False
             if entries:
                 # redis-py returns ``[[stream_key, [(entry_id, fields), ...]], ...]``
                 # — one outer tuple per stream queried. We always
@@ -258,58 +326,32 @@ async def _feed_generator(
                 # feed), so ``entries[0][1]`` is the list of
                 # ``(entry_id, fields_dict)`` tuples.
                 _key, items = entries[0]
-                for entry_id, fields in items:
+                for entry_id, frame in _process_entries(
+                    items,
+                    op_class=op_class,
+                    principal=principal,
+                    target=target,
+                    stream_key=stream_key,
+                ):
                     cursor = entry_id
-                    raw_event_json = fields.get("event")
-                    if not isinstance(raw_event_json, str):
-                        # A producer that XADD'd a field other than
-                        # ``event`` would land here. T3's publisher
-                        # is the only writer; this branch is the
-                        # safety net against a future Slack-mirror /
-                        # downstream tool writing alternate field
-                        # shapes onto the same stream key.
-                        _log.warning(
-                            "feed_skipped_unknown_field_shape",
-                            stream_key=stream_key,
-                            entry_id=entry_id,
-                            fields=list(fields.keys()),
-                        )
-                        continue
-                    try:
-                        event = BroadcastEvent.model_validate_json(raw_event_json)
-                    except ValidationError:
-                        # A malformed event on the stream is a T3
-                        # bug or a stream-key collision with a
-                        # foreign writer. Log + skip rather than
-                        # tearing down every subscriber.
-                        _log.warning(
-                            "feed_skipped_malformed_event",
-                            stream_key=stream_key,
-                            entry_id=entry_id,
-                        )
-                        continue
-                    if not _passes_filter(event, op_class, principal, target):
-                        continue
-                    yield _format_event(entry_id, raw_event_json)
+                    yield frame
+                    emitted_any = True
+            if emitted_any:
                 last_heartbeat = now
-                continue
-            # XREAD returned no entries within the block window.
-            # Emit a heartbeat if the wall-clock idle window has
-            # elapsed (it always has, since BLOCK = HEARTBEAT_INTERVAL
-            # by construction — but the explicit check keeps the
-            # contract honest if either constant later changes).
-            if now - last_heartbeat >= _HEARTBEAT_INTERVAL_SECONDS:
+            elif now - last_heartbeat >= _HEARTBEAT_INTERVAL_SECONDS:
                 yield ": heartbeat\n\n"
                 last_heartbeat = now
     except asyncio.CancelledError:
-        # Client disconnect. Suppress propagation so the audit row at
-        # session end reflects a clean 200 close.
+        # Client disconnect. Log the structured event for operator
+        # triage, then re-raise so the task tree unwinds per asyncio's
+        # cancellation contract — Sonar S7497, Python 3.13+ asyncio
+        # internals (re-issue cancellation if it goes unpropagated).
         _log.info(
             "feed_subscriber_disconnected",
             stream_key=stream_key,
             operator_sub=operator.sub,
         )
-        return
+        raise
 
 
 @router.get("/feed")
--- a/backend/src/meho_backplane/api/v1/feed.py
+++ b/backend/src/meho_backplane/api/v1/feed.py
@@ -66,10 +66,23 @@ Heartbeat
 =========
 
 The generator emits ``: heartbeat\\n\\n`` (SSE comment line) every
-``_HEARTBEAT_INTERVAL_SECONDS`` of inactivity to keep intermediaries
-from idle-timing-out the connection. The heartbeat fires when the
-``XREAD BLOCK`` call returns no entries within the block window —
-the natural quiet-time signal Valkey gives the loop.
+``_HEARTBEAT_INTERVAL_SECONDS`` of **outbound silence** (no event
+frame yielded to the subscriber) to keep intermediaries from
+idle-timing-out the connection. Two scenarios trigger the heartbeat:
+
+1. **Valkey-quiet** — ``XREAD BLOCK`` returns no entries within the
+   block window. The natural quiet-time signal Valkey gives the loop.
+2. **Subscriber-filtered** — ``XREAD BLOCK`` returns entries but they
+   all fail the operator's ``op_class`` / ``principal`` / ``target``
+   filter, so the generator yields nothing outbound. Without this
+   second path a noisy tenant with a narrow-filtered subscriber
+   would emit zero outbound bytes (events filtered, no heartbeat
+   either) and the nginx / ALB / CloudFront ~60 s idle timeout would
+   drop the connection.
+
+``last_heartbeat`` tracks the wall-clock of the last outbound yield
+(event frame or heartbeat); inbound XREAD activity that doesn't
+produce an outbound frame does NOT reset it.
 
 Disconnect handling
 ===================
@@ -100,12 +113,13 @@ References
 from __future__ import annotations
 
 import asyncio
+import re
 import time
 from collections.abc import AsyncIterator, Iterator
 from typing import Final
 
 import structlog
-from fastapi import APIRouter, Depends, Query, Request
+from fastapi import APIRouter, Depends, HTTPException, Query, Request
 from fastapi.responses import StreamingResponse
 from pydantic import ValidationError
 
@@ -155,6 +169,50 @@ _XREAD_COUNT: Final[int] = 20
 #: a fresh connection without ``Last-Event-Id`` / ``since`` gets the
 #: live tail, not a backlog.
 _LIVE_TAIL_CURSOR: Final[str] = "$"
+
+#: Valkey stream entry id shape — ``<ms-timestamp>`` or
+#: ``<ms-timestamp>-<sequence>``. Accepts both forms because the
+#: bare-timestamp form is legal (Valkey auto-assigns sequence 0) even
+#: though XADD-emitted ids always include the sequence suffix.
+#: Used by :func:`_validate_cursor_or_400` at the route boundary to
+#: reject malformed ``Last-Event-Id`` / ``since`` values before they
+#: reach XREAD and trigger an SSE reconnect loop.
+_VALKEY_STREAM_ID_RE: Final[re.Pattern[str]] = re.compile(r"^\d+(?:-\d+)?$")
+
+
+def _validate_cursor_or_400(cursor: str) -> str:
+    """Validate the SSE replay cursor; raise HTTP 400 on bad input.
+
+    Cursor sources are operator-controlled (``Last-Event-Id`` header,
+    ``since`` query parameter). Without this gate, a malformed cursor
+    propagates verbatim into the ``XREAD`` call and Valkey rejects it
+    with a ``redis.ResponseError`` mid-stream — the SSE response was
+    already sent ``http.response.start``, so the failure surfaces as a
+    connection drop. The browser ``EventSource`` auto-reconnects per
+    the WHATWG spec with the SAME bad cursor and tightens into a
+    reconnect loop.
+
+    Returning HTTP 400 at the route boundary (before any streaming)
+    flips the SSE state machine to ``readyState=CLOSED`` (the spec
+    aborts auto-reconnect on 4xx-class responses), giving the client
+    a recoverable error rather than a hot-loop.
+
+    Accepts:
+
+    * ``"$"`` — Valkey's "live tail" anchor.
+    * ``"<int>"`` or ``"<int>-<int>"`` — Valkey stream id forms.
+
+    Rejects everything else with ``HTTPException(400)``. The detail
+    string deliberately doesn't echo the input — operator-controlled
+    cursors could carry log-injection payloads if reflected verbatim
+    into structured log shippers.
+    """
+    if cursor == _LIVE_TAIL_CURSOR or _VALKEY_STREAM_ID_RE.fullmatch(cursor):
+        return cursor
+    raise HTTPException(
+        status_code=400,
+        detail="invalid_cursor: expected Valkey stream id (e.g. '1715600000000-0') or '$'",
+    )
 
 
 def _stream_key(operator: Operator) -> str:
@@ -405,9 +463,11 @@ async def feed_endpoint(
         the stream into a single response (nginx default behaviour
         would otherwise stall every event behind its buffer flush).
     """
-    cursor = _resolve_cursor(
-        last_event_id_header=request.headers.get("Last-Event-Id"),
-        since=since,
+    cursor = _validate_cursor_or_400(
+        _resolve_cursor(
+            last_event_id_header=request.headers.get("Last-Event-Id"),
+            since=since,
+        )
     )
     generator = _feed_generator(
         operator=operator,
--- a/backend/src/meho_backplane/api/v1/feed.py
+++ b/backend/src/meho_backplane/api/v1/feed.py
@@ -76,10 +76,17 @@ Disconnect handling
 
 Client disconnect propagates as :class:`asyncio.CancelledError` into
 the generator (Starlette cancels the request task on
-``http.disconnect``). The ``except`` arm returns cleanly without
-re-raising so the audit row at session end records the normal exit
-rather than a synthetic 500 — and Valkey's BLOCKing XREAD releases
-the connection from the pool the next event-loop tick.
+``http.disconnect``). The ``except`` arm logs the structured
+``feed_subscriber_disconnected`` event for operator triage, then
+re-raises so the cancellation unwinds the task tree per asyncio's
+contract (Sonar S7497; Python 3.13+ asyncio re-issues cancellation
+if it goes unpropagated). The audit row at session end still records
+a clean 200 close because ``http.response.start`` was sent on the
+first yield — before any cancellation point — so
+:class:`~meho_backplane.audit.AuditMiddleware`'s buffered
+``status_code`` is already locked at 200 by the time the cancellation
+propagates. Valkey's BLOCKing XREAD releases the connection from the
+pool the next event-loop tick.
 
 References
 ----------
@@ -326,14 +333,32 @@ async def _feed_generator(
                 # feed), so ``entries[0][1]`` is the list of
                 # ``(entry_id, fields_dict)`` tuples.
                 _key, items = entries[0]
-                for entry_id, frame in _process_entries(
+                if items:
+                    # Advance the cursor past EVERY consumed entry,
+                    # not only the ones that survive the filter. The
+                    # M2 refactor moved skip-paths (filter mismatch,
+                    # malformed JSON, unknown field shape) inside
+                    # ``_process_entries``; the helper consumes those
+                    # entries without yielding, so a ``cursor =
+                    # entry_id`` placed inside the post-helper
+                    # ``for ... yield`` loop never advances past them.
+                    # Under explicit-cursor replay (``since=<id>`` or
+                    # ``Last-Event-Id``) a tenant where every entry is
+                    # filtered out would otherwise re-read the same
+                    # batch forever: XREAD with cursor=id_N returns
+                    # IDs > id_N, helper drops them, cursor stays at
+                    # id_N, next XREAD returns the same set. Set the
+                    # cursor here, BEFORE the yield loop, so the next
+                    # XREAD reads strictly past this batch regardless
+                    # of whether anything yielded out the other side.
+                    cursor = items[-1][0]
+                for _entry_id, frame in _process_entries(
                     items,
                     op_class=op_class,
                     principal=principal,
                     target=target,
                     stream_key=stream_key,
                 ):
-                    cursor = entry_id
                     yield frame
                     emitted_any = True
             if emitted_any:
--- a/backend/src/meho_backplane/api/v1/feed.py
+++ b/backend/src/meho_backplane/api/v1/feed.py
@@ -1,0 +1,359 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""``GET /api/v1/feed`` — Server-Sent Events feed for broadcast events (G6.1-T4).
+
+Per-tenant SSE endpoint that streams :class:`BroadcastEvent` records
+``XADD``\\ ed by T3's publish-on-write hook onto
+``meho:feed:{tenant_id}``. Subscribers — `meho status --watch` (T5,
+#311), the future Slack mirror (G6.2 #333), and any third-party
+operator dashboard — open one HTTP connection per session and
+receive events as soon as they're published.
+
+Transport
+=========
+
+Standard SSE per the WHATWG ``EventSource`` spec. Each event is
+emitted as::
+
+    event: broadcast
+    data: <model_dump_json of BroadcastEvent>
+    id: <Valkey stream entry id, e.g. 1715600000000-0>
+
+Clients reconnecting with ``Last-Event-Id: <id>`` receive replay from
+that point. The ``id:`` field carries the Valkey entry id verbatim so
+the reconnect handshake is server-stateless — the Valkey stream IS the
+session state.
+
+Filtering
+=========
+
+Three optional query parameters apply **after** the XREAD pull:
+
+* ``op_class`` — exact-match filter on
+  :attr:`BroadcastEvent.op_class` (e.g. ``read``, ``write``,
+  ``credential_read``).
+* ``principal`` — exact-match filter on
+  :attr:`BroadcastEvent.principal_sub` (the JWT ``sub`` claim, not the
+  human ``name``).
+* ``target`` — exact-match filter on
+  :attr:`BroadcastEvent.target_name`.
+
+All three default to "no filter" — an unfiltered feed yields every
+event the operator's tenant produces.
+
+Replay
+======
+
+Either of the two replay knobs may be set; ``Last-Event-Id`` header
+takes precedence over the ``since`` query parameter when both are
+present. The default cursor is ``$`` (Valkey's "from now" anchor), so
+a fresh connection yields only events ``XADD``\\ ed after the SSE
+handshake completes.
+
+Tenant scoping
+==============
+
+The stream key is ``meho:feed:{operator.tenant_id}`` derived from the
+validated JWT; the client cannot subscribe to a different tenant's
+stream by passing a tenant_id in the query string — there is no such
+parameter. RBAC requires ``operator`` role minimum (read_only → 403);
+the rationale matches the retrieval route's gate: read-only operators
+have lower-friction surfaces (``meho status``, kb search) that don't
+need real-time feed.
+
+Heartbeat
+=========
+
+The generator emits ``: heartbeat\\n\\n`` (SSE comment line) every
+``_HEARTBEAT_INTERVAL_SECONDS`` of inactivity to keep intermediaries
+from idle-timing-out the connection. The heartbeat fires when the
+``XREAD BLOCK`` call returns no entries within the block window —
+the natural quiet-time signal Valkey gives the loop.
+
+Disconnect handling
+===================
+
+Client disconnect propagates as :class:`asyncio.CancelledError` into
+the generator (Starlette cancels the request task on
+``http.disconnect``). The ``except`` arm returns cleanly without
+re-raising so the audit row at session end records the normal exit
+rather than a synthetic 500 — and Valkey's BLOCKing XREAD releases
+the connection from the pool the next event-loop tick.
+
+References
+----------
+
+* SSE / EventSource: https://html.spec.whatwg.org/multipage/server-sent-events.html
+* Valkey ``XREAD``: https://valkey.io/commands/xread/
+* FastAPI ``StreamingResponse``:
+  https://fastapi.tiangolo.com/advanced/custom-response/#streamingresponse
+"""
+
+from __future__ import annotations
+
+import asyncio
+import time
+from collections.abc import AsyncIterator
+from typing import Final
+
+import structlog
+from fastapi import APIRouter, Depends, Query, Request
+from fastapi.responses import StreamingResponse
+from pydantic import ValidationError
+
+from meho_backplane.auth.operator import Operator, TenantRole
+from meho_backplane.auth.rbac import require_role
+from meho_backplane.broadcast import BroadcastEvent, get_broadcast_client
+
+__all__ = ["router"]
+
+router = APIRouter(prefix="/api/v1", tags=["feed"])
+
+_log = structlog.get_logger(__name__)
+
+#: Operator-minimum gate, factored to module scope so B008 (the
+#: "no function call in default" lint rule) doesn't trip on
+#: ``Depends(require_role(...))`` inside the handler's argument list.
+#: Mirrors the in-repo pattern in :mod:`meho_backplane.api.v1.retrieve`.
+_REQUIRE_OPERATOR = Depends(require_role(TenantRole.OPERATOR))
+
+
+#: XREAD ``BLOCK`` window in milliseconds. The generator yields a
+#: heartbeat once per quiet window of this length; the value also
+#: bounds how long a client-disconnect-cancel takes to land in the
+#: generator (the BLOCK call has to time out or return entries before
+#: the surrounding loop can observe the cancelled context).
+#:
+#: 30 seconds matches the issue body's spec and the common SSE
+#: intermediary keep-alive window (nginx default proxy_read_timeout
+#: is 60s; AWS ALB idle timeout 60s).
+_XREAD_BLOCK_MS: Final[int] = 30_000
+
+#: Wall-clock idle interval after which the generator emits a
+#: heartbeat. Same value as the XREAD block window so a quiet period
+#: produces exactly one heartbeat per cycle — bypasses the "BLOCK 30s
+#: returned empty, did the connection drop?" ambiguity for both the
+#: client and any HTTP intermediaries.
+_HEARTBEAT_INTERVAL_SECONDS: Final[float] = 30.0
+
+#: Upper bound on entries pulled per XREAD call. Bounded to keep the
+#: generator's tick latency-aware: a burst of N events still trickles
+#: out to the client in chunks of up to this size rather than blocking
+#: the event-loop tick on a 1000-event ``model_validate_json`` sweep.
+_XREAD_COUNT: Final[int] = 20
+
+#: Initial cursor when the client doesn't provide one. Valkey's ``$``
+#: anchor means "only entries XADD'd after this XREAD call started" —
+#: a fresh connection without ``Last-Event-Id`` / ``since`` gets the
+#: live tail, not a backlog.
+_LIVE_TAIL_CURSOR: Final[str] = "$"
+
+
+def _stream_key(operator: Operator) -> str:
+    """Build the per-tenant stream key from the JWT-derived tenant id.
+
+    Centralised here so a future tenancy-isolation tightening (e.g. a
+    per-environment prefix) lands in one place. Mirrors the same
+    helper in :mod:`meho_backplane.broadcast.publisher`.
+    """
+    return f"meho:feed:{operator.tenant_id}"
+
+
+def _format_event(entry_id: str, raw_event_json: str) -> str:
+    """Format one Valkey entry as an SSE ``event: broadcast`` frame.
+
+    The ``data:`` field carries the JSON verbatim — no
+    pretty-printing or re-serialisation. Subscribers parse this back
+    into a :class:`BroadcastEvent` on their side, so a re-dump here
+    would only churn newlines (and SSE forbids embedded newlines in
+    ``data:`` fields unless explicitly split into multiple ``data:``
+    lines — easier to enforce by passing the canonical single-line
+    JSON straight through from
+    :meth:`BroadcastEvent.model_dump_json`).
+
+    The ``id:`` line is the Valkey stream entry id verbatim;
+    clients use it as the next ``Last-Event-Id`` value.
+    """
+    return f"event: broadcast\ndata: {raw_event_json}\nid: {entry_id}\n\n"
+
+
+def _passes_filter(
+    event: BroadcastEvent,
+    op_class: str | None,
+    principal: str | None,
+    target: str | None,
+) -> bool:
+    """Return ``True`` iff the event matches every non-None filter.
+
+    None means "no filter on this field". Exact-match semantics — no
+    substring or pattern matching today; G6.3 may revisit if operator
+    feedback flags the gap. ``target_name`` is nullable on
+    :class:`BroadcastEvent`; an event with ``target_name=None`` and
+    a non-None *target* filter never passes (the operator asked for
+    a specific target; an event with no target attribution doesn't
+    qualify).
+    """
+    if op_class is not None and event.op_class != op_class:
+        return False
+    if principal is not None and event.principal_sub != principal:
+        return False
+    return not (target is not None and event.target_name != target)
+
+
+def _resolve_cursor(
+    last_event_id_header: str | None,
+    since: str | None,
+) -> str:
+    """Pick the XREAD cursor per the ``Last-Event-Id`` > ``since`` > ``$`` order.
+
+    Per the SSE spec, ``Last-Event-Id`` is the canonical reconnect
+    mechanism — clients automatically resend it after a connection
+    drop. The ``since`` query parameter exists for callers that need
+    explicit replay control without relying on the SSE auto-reconnect
+    machinery (e.g. server-side bridges that consume the feed via
+    plain HTTP). When both are present the header wins because it
+    encodes "I'm continuing a session I already had"; the query
+    parameter is the explicit override.
+    """
+    if last_event_id_header:
+        return last_event_id_header
+    if since:
+        return since
+    return _LIVE_TAIL_CURSOR
+
+
+async def _feed_generator(
+    operator: Operator,
+    cursor: str,
+    op_class: str | None,
+    principal: str | None,
+    target: str | None,
+) -> AsyncIterator[str]:
+    """SSE generator: BLOCK on XREAD, filter, format, heartbeat.
+
+    Cancelled cleanly on client disconnect (Starlette raises
+    :class:`asyncio.CancelledError` into the await point). The
+    ``except`` arm returns silently rather than propagating so the
+    surrounding StreamingResponse doesn't synthesise a 500-shaped
+    audit row at session end — a normal SSE close lands as a clean
+    200.
+    """
+    client = get_broadcast_client()
+    stream_key = _stream_key(operator)
+    last_heartbeat = time.monotonic()
+
+    try:
+        while True:
+            entries = await client.xread(
+                {stream_key: cursor},
+                block=_XREAD_BLOCK_MS,
+                count=_XREAD_COUNT,
+            )
+            now = time.monotonic()
+            if entries:
+                # redis-py returns ``[[stream_key, [(entry_id, fields), ...]], ...]``
+                # — one outer tuple per stream queried. We always
+                # query exactly one stream (the operator's tenant
+                # feed), so ``entries[0][1]`` is the list of
+                # ``(entry_id, fields_dict)`` tuples.
+                _key, items = entries[0]
+                for entry_id, fields in items:
+                    cursor = entry_id
+                    raw_event_json = fields.get("event")
+                    if not isinstance(raw_event_json, str):
+                        # A producer that XADD'd a field other than
+                        # ``event`` would land here. T3's publisher
+                        # is the only writer; this branch is the
+                        # safety net against a future Slack-mirror /
+                        # downstream tool writing alternate field
+                        # shapes onto the same stream key.
+                        _log.warning(
+                            "feed_skipped_unknown_field_shape",
+                            stream_key=stream_key,
+                            entry_id=entry_id,
+                            fields=list(fields.keys()),
+                        )
+                        continue
+                    try:
+                        event = BroadcastEvent.model_validate_json(raw_event_json)
+                    except ValidationError:
+                        # A malformed event on the stream is a T3
+                        # bug or a stream-key collision with a
+                        # foreign writer. Log + skip rather than
+                        # tearing down every subscriber.
+                        _log.warning(
+                            "feed_skipped_malformed_event",
+                            stream_key=stream_key,
+                            entry_id=entry_id,
+                        )
+                        continue
+                    if not _passes_filter(event, op_class, principal, target):
+                        continue
+                    yield _format_event(entry_id, raw_event_json)
+                last_heartbeat = now
+                continue
+            # XREAD returned no entries within the block window.
+            # Emit a heartbeat if the wall-clock idle window has
+            # elapsed (it always has, since BLOCK = HEARTBEAT_INTERVAL
+            # by construction — but the explicit check keeps the
+            # contract honest if either constant later changes).
+            if now - last_heartbeat >= _HEARTBEAT_INTERVAL_SECONDS:
+                yield ": heartbeat\n\n"
+                last_heartbeat = now
+    except asyncio.CancelledError:
+        # Client disconnect. Suppress propagation so the audit row at
+        # session end reflects a clean 200 close.
+        _log.info(
+            "feed_subscriber_disconnected",
+            stream_key=stream_key,
+            operator_sub=operator.sub,
+        )
+        return
+
+
+@router.get("/feed")
+async def feed_endpoint(
+    request: Request,
+    op_class: str | None = Query(default=None, description="Filter by event op_class."),
+    principal: str | None = Query(default=None, description="Filter by principal_sub."),
+    target: str | None = Query(default=None, description="Filter by target_name."),
+    since: str | None = Query(
+        default=None,
+        description="Explicit replay cursor; superseded by Last-Event-Id when present.",
+    ),
+    operator: Operator = _REQUIRE_OPERATOR,
+) -> StreamingResponse:
+    """Stream broadcast events for the operator's tenant as Server-Sent Events.
+
+    Headers:
+        ``Last-Event-Id``: optional SSE reconnect cursor. Takes
+        precedence over the ``since`` query parameter when both are
+        present.
+
+    Returns:
+        :class:`StreamingResponse` with ``media_type=text/event-stream``.
+        ``Cache-Control: no-cache, no-store, must-revalidate`` and
+        ``X-Accel-Buffering: no`` keep intermediaries from buffering
+        the stream into a single response (nginx default behaviour
+        would otherwise stall every event behind its buffer flush).
+    """
+    cursor = _resolve_cursor(
+        last_event_id_header=request.headers.get("Last-Event-Id"),
+        since=since,
+    )
+    generator = _feed_generator(
+        operator=operator,
+        cursor=cursor,
+        op_class=op_class,
+        principal=principal,
+        target=target,
+    )
+    return StreamingResponse(
+        generator,
+        media_type="text/event-stream",
+        headers={
+            "Cache-Control": "no-cache, no-store, must-revalidate",
+            "X-Accel-Buffering": "no",
+        },
+    )
--- a/backend/tests/test_api_v1_feed.py
+++ b/backend/tests/test_api_v1_feed.py
@@ -256,18 +256,22 @@ async def _collect_n_frames(
 # ---------------------------------------------------------------------------
 
 
-def _mint_authenticated_token(
+async def _request_feed_authenticated(
     monkeypatch: pytest.MonkeyPatch,
     *,
     kid: str,
     sub: str,
     role: TenantRole,
-) -> tuple[str, object]:
-    """Bundle ``install_fake_vault`` + ``make_rsa_keypair`` + ``mint_token``.
+    extra_headers: dict[str, str] | None = None,
+    params: dict[str, str] | None = None,
+) -> httpx.Response:
+    """One-call HTTP-edge scaffold: vault + key + JWT + respx + client.
 
-    Returns ``(token, key)`` ready to wire into a ``respx.mock`` block.
-    Extracted so HTTP-edge tests can share the JWT-minting boilerplate
-    without tripping SonarCloud's duplication-on-new-code threshold.
+    Collapses the install_fake_vault / make_rsa_keypair / mint_token /
+    respx.mock / _make_client chain that every authenticated-edge test
+    needs into a single helper. Extracted so the three /api/v1/feed
+    edge cases share one body shape instead of three lookalike
+    scaffolds (SonarCloud duplication-on-new-code threshold).
     """
     from tests._vault_fakes import install_fake_vault
 
@@ -279,7 +283,12 @@ def _mint_authenticated_token(
         tenant_id=str(_TENANT_A),
         tenant_role=role.value,
     )
-    return token, key
+    headers = {"Authorization": f"Bearer {token}", **(extra_headers or {})}
+    app = _build_app()
+    with respx.mock as mock_router:
+        mock_discovery_and_jwks(mock_router, public_jwks(key))
+        async with _make_client(app) as client:
+            return await client.get("/api/v1/feed", headers=headers, params=params)
 
 
 class TestFeedEndpoint:
@@ -297,20 +306,12 @@ class TestFeedEndpoint:
         monkeypatch: pytest.MonkeyPatch,
     ) -> None:
         """``read_only`` operators can't subscribe; ``operator`` minimum required."""
-        token, key = _mint_authenticated_token(
+        response = await _request_feed_authenticated(
             monkeypatch,
             kid="kid-feed-read-only",
             sub="op-readonly",
             role=TenantRole.READ_ONLY,
         )
-        app = _build_app()
-        with respx.mock as mock_router:
-            mock_discovery_and_jwks(mock_router, public_jwks(key))
-            async with _make_client(app) as client:
-                response = await client.get(
-                    "/api/v1/feed",
-                    headers={"Authorization": f"Bearer {token}"},
-                )
         assert response.status_code == 403
         assert response.json() == {"detail": "insufficient_role"}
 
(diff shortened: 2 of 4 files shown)


zdamir:
@coderabbitai review

backend/src/meho_backplane/api/v1/feed.py
```diff
@@ -0,0 +1,359 @@
...
+                last_heartbeat = now
+                continue
+            # XREAD returned no entries within the block window.
+            # Emit a heartbeat if the wall-clock idle window has
+            # elapsed (it always has, since BLOCK = HEARTBEAT_INTERVAL
+            # by construction — but the explicit check keeps the
+            # contract honest if either constant later changes).
+            if now - last_heartbeat >= _HEARTBEAT_INTERVAL_SECONDS:
+                yield ": heartbeat\n\n"
+                last_heartbeat = now
```
zdamir, line 426: **[🛑 B1]** — Heartbeat clock should track outbound silence, not inbound XREAD activity (also flagged by CodeRabbit)

**What:** Lines 297-303 in `_feed_generator`:
```python
if entries:
    _key, items = entries[0]
    for entry_id, fields in items:
        ...
        if not _passes_filter(event, op_class, principal, target):
            continue
        yield _format_event(entry_id, raw_event_json)
    last_heartbeat = now            # ← resets even when zero frames yielded
    continue
```
`last_heartbeat` is reset whenever `xread` returns *any* entries, including entries that are entirely filtered out for this subscriber. In a high-traffic tenant where the subscriber's `op_class` / `principal` / `target` filter rejects every event, the connection emits **zero outbound bytes** but the heartbeat timer keeps resetting. After the configured idle window the wall-clock check `now - last_heartbeat >= _HEARTBEAT_INTERVAL_SECONDS` is never true, no heartbeat is emitted, and intermediaries (nginx default `proxy_read_timeout 60s`, AWS ALB idle timeout 60s) silently drop the connection.

**Why it matters:** Heartbeat exists specifically to prevent intermediary idle-timeouts. The contract from the module docstring — "emits `: heartbeat\n\n` every 30s of inactivity to keep intermediaries from idle-timing-out the connection" — is broken under the busy-but-all-filtered scenario, which is exactly the case where the heartbeat is load-bearing (the operator subscribed with narrow filters precisely because the tenant is noisy). T5 (`meho status --watch`) is the first downstream consumer that would hit this — its narrow filter for a single principal would chronically drop on a multi-operator tenant.

**Required change:** Track `last_heartbeat` against actual outbound yields, not inbound `xread` activity. Only reset when an event frame OR a heartbeat frame is emitted; otherwise let the idle window accumulate so the heartbeat path fires when the wall-clock check next evaluates. CodeRabbit's suggestion is the right shape:

```python
if entries:
    _key, items = entries[0]
    emitted_any = False
    for entry_id, fields in items:
        ...
        if not _passes_filter(event, op_class, principal, target):
            continue
        yield _format_event(entry_id, raw_event_json)
        emitted_any = True
    if emitted_any:
        last_heartbeat = now
    elif now - last_heartbeat >= _HEARTBEAT_INTERVAL_SECONDS:
        yield ": heartbeat\n\n"
        last_heartbeat = now
    continue
```

Add a regression test under `TestFeedGenerator` that pins the contract: `xread` returns N cycles of events that all fail the filter; the generator must emit at least one `: heartbeat\n\n` frame within `n` cycles. The existing `_xread_returning` helper builds the cycles; the new test asserts on `: heartbeat\n\n` appearing in the collected frames.

**Source:** SSE / EventSource spec §"Reconnection time" + Heartbeat conventions — https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation

backend/src/meho_backplane/api/v1/feed.py
```diff
@@ -0,0 +1,359 @@
...
+    except asyncio.CancelledError:
```
zdamir, line 427: **[⚠️ M1]** — `asyncio.CancelledError` must be re-raised after cleanup (Sonar S7497)

**What:** Lines 296-304 in `_feed_generator`:
```python
except asyncio.CancelledError:
    _log.info(
        "feed_subscriber_disconnected",
        stream_key=stream_key,
        operator_sub=operator.sub,
    )
    return
```
The handler catches `asyncio.CancelledError`, performs cleanup (a structlog info line), and `return`s — swallowing the cancellation. Sonar S7497 flags this as a Major reliability defect; Python's asyncio docs are explicit that `CancelledError` should be re-raised after cleanup so the task tree unwinds correctly.

**Why it matters:** Swallowing `CancelledError` breaks the asyncio cancellation contract: a parent task that issued `task.cancel()` expects `CancelledError` to propagate up. Starlette's `StreamingResponse` machinery and the outer ASGI middleware chain (AuditMiddleware, RequestContextMiddleware) rely on this propagation to terminate cleanly. Swallowing it means a child task can "ignore" cancellation, which the asyncio internals consider a programming error (Python 3.12 onward, asyncio.timeout will retry cancellation if not propagated correctly). On the SSE feed path the practical symptom is subtle: the audit row at session end may record an inconsistent status, or future asyncio versions may emit a `RuntimeError` for "Task was destroyed but it is pending".

Also contributes to the **C Reliability Rating on New Code** SonarCloud Quality Gate failure — fixing this single finding should bump the rating back to A (assuming no other Reliability-class issues remain).

**Required change:** Re-raise after the cleanup log:

```python
except asyncio.CancelledError:
    _log.info(
        "feed_subscriber_disconnected",
        stream_key=stream_key,
        operator_sub=operator.sub,
    )
    raise
```

The audit-row-at-session-end behavior the original docstring described still holds: `http.response.start` was sent on the first frame (or on first heartbeat tick if no events), so status_code=200 is already locked into the buffered ASGI messages by the time cancellation arrives. The AuditMiddleware records that 200 regardless of whether the generator exits via `return` or `raise CancelledError`. The cancellation propagates cleanly to the task tree above; nothing operator-visible changes.

**Source:** Python 3.13 docs — https://docs.python.org/3/library/asyncio-task.html#task-cancellation ("If you call `Task.cancel()`, then `asyncio.CancelledError` is raised in the task. The task should then re-raise the exception when it has finished cleanup work.") + Sonar rule https://rules.sonarsource.com/python/RSPEC-7497/

backend/src/meho_backplane/api/v1/feed.py
```diff
@@ -0,0 +1,359 @@
...
+    parameter is the explicit override.
+    """
+    if last_event_id_header:
+        return last_event_id_header
+    if since:
+        return since
+    return _LIVE_TAIL_CURSOR
+
+
+async def _feed_generator(
```
zdamir, line 344: **[⚠️ M2]** — `_feed_generator` cognitive complexity 21 > 15 (Sonar S3776)

**What:** `_feed_generator` (line 226+) accumulates cognitive complexity 21 against the 15 ceiling. Sonar's S3776 counts nested control flow with multipliers — the function has a `while True` loop, nested `if entries:` branch, inner `for entry_id, fields in items:` loop, three skip-and-continue branches (unknown field shape, malformed JSON, filter mismatch), a `yield`, a heartbeat check, and a `try/except CancelledError`. Each nesting level multiplies the count.

**Why it matters:** Cognitive complexity 21 is a maintainability hazard — every future change (M1 heartbeat-clock fix, B1 outbound-silence fix, a future feature like filtering by event type) compounds the nesting and increases the bug surface. Sonar S3776 is also the second of the two Reliability/Maintainability findings driving the Quality Gate failure; fixing it AND M1's re-raise should clear the gate.

**Required change:** Extract the per-`xread`-batch processing into a private helper so the main loop is a thin `while True` that delegates one cycle's worth of work. Sketch:

```python
def _process_entries(
    items: list[tuple[str, dict[str, str]]],
    op_class: str | None,
    principal: str | None,
    target: str | None,
    stream_key: str,
) -> Iterator[tuple[str, str]]:
    """Yield (entry_id, sse_frame) for every entry passing the filter.

    Logs + skips malformed JSON and unknown field shapes.
    """
    for entry_id, fields in items:
        raw_event_json = fields.get("event")
        if not isinstance(raw_event_json, str):
            _log.warning("feed_skipped_unknown_field_shape", ...)
            continue
        try:
            event = BroadcastEvent.model_validate_json(raw_event_json)
        except ValidationError:
            _log.warning("feed_skipped_malformed_event", ...)
            continue
        if not _passes_filter(event, op_class, principal, target):
            continue
        yield entry_id, _format_event(entry_id, raw_event_json)
```

Then the main loop collapses to:

```python
while True:
    entries = await client.xread({stream_key: cursor}, block=_XREAD_BLOCK_MS, count=_XREAD_COUNT)
    now = time.monotonic()
    emitted_any = False
    if entries:
        _key, items = entries[0]
        for entry_id, frame in _process_entries(items, op_class, principal, target, stream_key):
            cursor = entry_id
            yield frame
            emitted_any = True
    if emitted_any:
        last_heartbeat = now
    elif now - last_heartbeat >= _HEARTBEAT_INTERVAL_SECONDS:
        yield ": heartbeat\n\n"
        last_heartbeat = now
```

That brings the main function's complexity well under 15. The helper is itself low-complexity (one `for`, two `continue`s).

**Source:** SonarSource cognitive-complexity whitepaper — https://www.sonarsource.com/docs/CognitiveComplexity.pdf + rule https://rules.sonarsource.com/python/RSPEC-3776/

zdamir (commented):
> **Note:** Posting as `--comment` because GitHub blocks formal `REQUEST_CHANGES` on one's own PR. Formal verdict below is **REQUEST_CHANGES**; the orchestrator reads the result file, not this badge. 1 blocker + 2 majors carry the punch list for `/auto-implement-issue --continue-pr`.

**Anchor task:** #310 — G6.1-T4 SSE endpoint `/api/v1/feed` (per-tenant filter + replay via Last-Event-Id + heartbeat)
**Surface area:** backend (Python — FastAPI + asyncio + redis-py asyncio + SSE)
**Best-practices packs audited:** python (async-generator-cancellation, SSE heartbeat invariants, cognitive complexity)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `70c60582cd3980bfef4fa1916d68e7141df1993c`
**AI reviewer (CodeRabbit):** received in ~3 min; 1 actionable inline (M1 heartbeat-clock — adopted as B1 with a severity bump to blocker) + 1 nitpick (heartbeat-with-filter regression test — folded into B1's required change).

### Acceptance criteria
- [x] AC #1 — 200 + `text/event-stream` for authenticated operator — `TestFeedEndpoint` covers authn/authz happy path.
- [x] AC #2 — Per-tenant scoping — `test_basic_event_yield_with_tenant_scoping` asserts the stream key derives from JWT tenant_id.
- [x] AC #3 — Tenant boundary — by construction (stream key is `meho:feed:{operator.tenant_id}`, no API surface accepts override).
- [x] AC #4-5 — Filter by `op_class` / `principal` — covered.
- [x] AC #6 — Replay via `since` — `test_cursor_passes_through_to_xread`.
- [x] AC #7 — `Last-Event-Id` precedence — `TestResolveCursor` covers all three precedence cases.
- [⚠️] AC #8 — Heartbeat every 30s during quiet periods — the heartbeat shape is correct, but **B1** shows it can be silently disabled by the all-filtered-out scenario, breaking the contract. The fix lands the missing regression test the CR nitpick called out.
- [x] AC #9 — Client disconnect cleanly closes — exercised on every generator-level test via `gen.aclose()`. **Caveat:** see M1 — the current `return` after `CancelledError` swallows the cancellation; correct cleanup re-raises.
- [x] AC #10 — Unauthenticated → 401 — `test_unauthenticated_returns_401`.
- [x] AC #11 — `read_only` → 403 — `test_read_only_role_returns_403`.
- [⚠️] AC #12 — CI green — see "Integration verification" below: code-level CI gates green, but **3 lanes fail on Docker Hub rate-limit (429) infrastructure flakes** (Python pytest, Semgrep SAST, image build all failed pulling base images). **Not** caused by this PR's code; documented under integration verification.
- [ ] AC #13 — Initiative #228 body checklist for T4 ticked — post-merge action.

### Project-board alignment
|---|---|---|---|---|---|
| #310 (T4) | open | `Closes #310` | ✅ | ⚠️ partial (B1 contract gap) | close on merge after B1 fix |
| #228 (Initiative G6.1) | open | parent of #310 | — | ⚠️ partial (T4 of 6) | tick T4 checkbox post-merge |
| #311 (T5 CLI watch), #312 (T6 MCP resource) | open | downstream consumers | — | — | unblocked once T4 lands |

### Predicted vs. actual diff
- ✅ `backend/src/meho_backplane/main.py` — new `api_v1_feed_router` mount, comment matches the established style.
- ⚠️ `backend/src/meho_backplane/api/v1/feed.py` — schema-right but carries B1 (heartbeat clock contract) + M1 (CancelledError re-raise) + M2 (cognitive complexity 21 > 15).
- ✅ `backend/tests/test_api_v1_feed.py` — comprehensive three-layer coverage (HTTP edge + generator + real-Valkey integration); needs the B1-driven regression test for heartbeat-under-filtered-traffic.
- ✅ `docs/codebase/backend.md` — T4 paragraph at the right anchor, T3 paragraph's forward-reference updated.

### Integration verification
- [x] `ruff format --check src tests` — clean (109 files)
- [x] `mypy src` strict — 56 source files, no issues
- [x] `pytest tests/test_api_v1_feed.py` (mocked layer) — 13 passed
- [x] `pytest --ignore=tests/integration` — 552 passed (+14 vs T3 baseline), 1 skipped, 1 deselected (macOS-only flake)
- [⚠️ infrastructure-fail] **CI Python lane** — failed pulling `pgvector/pgvector:pg16` from Docker Hub (429 Too Many Requests rate limit). The test that errored (`test_mcp_inspector.py::test_full_mcp_lifecycle_succeeds`) is pre-existing G0.5-T6 code unrelated to this PR. Re-running the failed job after the rate-limit window or wiring Docker Hub auth on the runner should resolve.
- [⚠️ infrastructure-fail] **CI Semgrep lane** — failed pulling `semgrep/semgrep:1.160.0`. Same root cause.
- [⚠️ infrastructure-fail] **CI image build** — failed pulling `python` base image (429). Same root cause.
- [❌] **SonarCloud Quality Gate** — failed with 5.2% Duplication on New Code (threshold ≤ 3%) + C Reliability Rating on New Code. The C rating is driven by the two Sonar Major findings (M1 cancellation re-raise + M2 cognitive complexity); fixing both should bump back to A. The duplication is in the test file (the per-test JWT minting block is repeated across a few tests in the new file — not strictly identical to the chassis pattern but Sonar's clone-detection scores it). Folding the JWT setup into a shared fixture would close it.

- 🛑 **B1** — `feed.py:294-303` heartbeat clock resets on inbound `xread` activity, not outbound yields → all-filtered-out traffic silently disables the heartbeat, breaking the intermediary-keep-alive contract. Required change: track `last_heartbeat` against actual yields; emit the heartbeat inside the entries branch when zero filters matched AND the idle window elapsed. Add the CR-flagged regression test pinning the contract. (Also flagged by CodeRabbit.)
- ⚠️ **M1** — `feed.py:304` `asyncio.CancelledError` is swallowed (`return`) instead of re-raised after cleanup. Required change: replace `return` with `raise` after the log line. Resolves Sonar S7497 and contributes to the Reliability A rating.
- ⚠️ **M2** — `feed.py:226+` `_feed_generator` cognitive complexity 21 > 15. Required change: extract per-`xread`-batch processing into a `_process_entries` helper; the main loop collapses to a thin while loop. The B1 refactor and this one fold together nicely. Resolves Sonar S3776.

- Sonar S8410 (lines 318-321) wants `Annotated[T, Query(...)]` instead of `T = Query(default=...)`. The existing chassis routes (`retrieve.py`, `health.py`) don't use Annotated either — this is a codebase-wide style choice better made in a focused refactor PR, not unilaterally on one new route.
- Sonar S7483 (test:211) wants the `timeout` parameter on `_collect_n_frames` removed in favour of an inline `asyncio.timeout(LITERAL)` call. The function already uses `async with asyncio.timeout(timeout):` — the parameter exists to let tests pin different waits. Sonar's rule misfires here; defer.
- Sonar 5.2% Duplication finding — the per-test JWT-minting boilerplate repeats across `TestFeedEndpoint::test_read_only_role_returns_403` and similar test files (`test_tenant_isolation`, `test_api_v1_retrieve`). Folding the setup into a `tests/_oidc_jwt_helpers` factory or a shared fixture would close it. Worth doing as part of T5/T6 if the next broadcast tests grow the same boilerplate; not a blocker.

- Adopted: 1 (CR's `M1 — heartbeat clock` finding became B1 with severity bumped to blocker because the AC contract is broken, not just sub-optimal).
- Adopted: 1 (CR nitpick about heartbeat-with-filter regression test) — folded into B1's required change.
- Disputed: 0.
- Out of scope: 0.

### Docs
- [x] `docs/codebase/backend.md` — T4 paragraph at correct anchor; T3 forward-reference updated. Verified no .NET analogies.

- **Generator-layer testing for body-shaping ACs is the right call.** The PR docstring names the racy httpx + ASGITransport cancellation problem and tests `_feed_generator` directly via `gen.aclose()`. Future SSE tests in T5/T6 should follow the same pattern; deterministic async-generator cleanup is much cleaner than fighting ASGI stream-close timing.
- **Defensive parsing** — malformed-JSON `event` field and unknown-field-shape entries both log + skip rather than crashing the subscriber. The two warning event names are grep-able for operators chasing producer bugs.
- **`X-Accel-Buffering: no` + `Cache-Control: no-cache, no-store, must-revalidate`** — exactly the headers nginx and similar intermediaries need to not buffer the stream into a single response. Catching this up front saves the first operator-deployment-debugging session.

backend/src/meho_backplane/api/v1/feed.py
zdamir: Fixed in `9c7082c` — heartbeat clock now tracks outbound yields only. `emitted_any` flag set on actual frame emission; when entries arrive but all filter out AND idle window elapsed, an inline heartbeat is emitted. Regression test `test_heartbeat_emitted_when_all_entries_filter_out` pins the contract: two write events filtered out by op_class=read subscriber produces exactly one outbound heartbeat frame (time compressed via monkeypatched `_HEARTBEAT_INTERVAL_SECONDS=0.05`).

backend/src/meho_backplane/api/v1/feed.py
zdamir: Fixed in `9c7082c` — replaced `return` with `raise` after the log line. CancelledError now propagates per Python's task-cancellation contract. The audit row at session end still records 200 because http.response.start was sent on the first yield (before any cancellation point); the buffered status is already locked by the time cancellation arrives. Test helper `_collect_n_frames` was updated to catch `TimeoutError` from `asyncio.timeout` since the timeout is the test's termination mechanism when `n` is unreachable.

backend/src/meho_backplane/api/v1/feed.py
zdamir: Fixed in `9c7082c` — extracted `_process_entries` sync iterator helper that handles the three skip paths (unknown field shape, malformed JSON, filter rejection) and yields (entry_id, frame) tuples. The main `_feed_generator` now collapses to a thin `while True: xread → delegate → heartbeat-if-silent` loop. Cognitive complexity well under 15. The refactor also makes the B1 heartbeat-clock fix land naturally — `emitted_any` tracking has one obvious home.

zdamir:
@coderabbitai review

backend/src/meho_backplane/api/v1/feed.py
```diff
@@ -0,0 +1,401 @@
...
+                _key, items = entries[0]
+                for entry_id, frame in _process_entries(
+                    items,
+                    op_class=op_class,
+                    principal=principal,
+                    target=target,
+                    stream_key=stream_key,
+                ):
+                    cursor = entry_id
+                    yield frame
```
zdamir, line 420: **[🛑 B1]** — XREAD cursor doesn't advance past filtered/malformed entries (also flagged by CodeRabbit)

**What:** Lines 329-337 in `_feed_generator`:
```python
_key, items = entries[0]
for entry_id, frame in _process_entries(
    items, op_class=op_class, principal=principal,
    target=target, stream_key=stream_key,
):
    cursor = entry_id           # ← only updates on yielded frames
    yield frame
    emitted_any = True
```
`_process_entries` is a generator that yields only entries passing the filter (or successfully deserializing). Filtered-out and malformed entries are silently consumed inside the helper — the main loop never sees their entry IDs, so `cursor` never advances past them.

**Why it matters:** This is an **infinite re-read loop** on every replay path (`since=<id>` or `Last-Event-Id`) that lands in a tenant with active-but-filtered traffic:

1. Subscriber opens with `since=1715600000000-0` (or a Last-Event-Id reconnect cursor).
2. New entries `0001-0`, `0002-0`, `0003-0` are XADD'd, all with `op_class=write`.
3. Subscriber's filter is `op_class=read` — every entry is filtered out by `_process_entries`.
4. Main loop's `for ... cursor = entry_id` never executes (helper yielded nothing).
5. Cursor remains at `0000-0`. Next `XREAD {stream: 0000-0}` returns 0001-0, 0002-0, 0003-0 *again*. All filter out again. Cursor still 0000-0. **Tight loop.**
6. Heartbeats keep emitting on the wall-clock check (the iter-1 B1 fix), so the connection stays alive — but the subscriber receives nothing while burning CPU on every-cycle redundant XREAD.

This breaks AC #6 (`since=<entry-id>` returns events after that id) and AC #7 (`Last-Event-Id` precedence) under the busy-but-filtered tenant scenario. AC #6 says "returns events after that id"; in this defect the subscriber gets ZERO matching events after that id even when matching events are clearly there for other subscribers — they're just hiding behind the never-advancing cursor.

The live-tail (`$`) case is incidentally fine because Valkey re-evaluates `$` to "now" on each XREAD call, so older entries (even if filtered) aren't re-read. But the explicit-cursor cases — which are precisely the replay paths the task body calls out — break.

**Required change:** Always advance the cursor to the last entry ID in the batch, regardless of which entries the filter accepted. CodeRabbit's diff is the minimal fix:

```python
if entries:
    _key, items = entries[0]
    if items:
        # Always advance past consumed entries, even if filtered/skipped.
        cursor = items[-1][0]
    for entry_id, frame in _process_entries(
        items, op_class=op_class, principal=principal,
        target=target, stream_key=stream_key,
    ):
        yield frame
        emitted_any = True
```

Add a regression test under `TestFeedGenerator` that pins the contract:

```python
async def test_cursor_advances_past_filtered_entries(self, _feed_env: None) -> None:
    """Filtered entries advance the cursor so XREAD doesn't re-read forever."""
    # First batch: all entries filtered out. Second batch: a matching event.
    # Without the fix, the second batch's XREAD would see the same first-batch
    # IDs again because cursor never advanced.
    write_a = _make_event(op_class="write", op_id="vsphere.vm.create")
    write_b = _make_event(op_class="write", op_id="vsphere.vm.delete")
    read_c = _make_event(op_class="read", op_id="vsphere.vm.list")
    broadcast_client = get_broadcast_client()
    call_cursors: list[str] = []

    async def _xread_side_effect(streams: dict[str, str], **_k: object) -> object:
        cursor = next(iter(streams.values()))
        call_cursors.append(cursor)
        if len(call_cursors) == 1:
            return [("k", [
                ("1715600000001-0", {"event": write_a.model_dump_json()}),
                ("1715600000002-0", {"event": write_b.model_dump_json()}),
            ])]
        if len(call_cursors) == 2:
            return [("k", [("1715600000003-0", {"event": read_c.model_dump_json()})])]
        await asyncio.sleep(0.01)
        return None

    mock = AsyncMock(side_effect=_xread_side_effect)
    with patch.object(broadcast_client, "xread", new=mock):
        gen = _feed_generator(
            operator=_make_operator(), cursor="$",
            op_class="read", principal=None, target=None,
        )
        frames = await _collect_n_frames(gen, n=1)

    assert len(frames) == 1
    assert "id: 1715600000003-0" in frames[0]
    # Second XREAD call's cursor was advanced past the filtered batch.
    assert call_cursors[1] == "1715600000002-0"
```

**Source:** Valkey XREAD semantics — https://valkey.io/commands/xread/ (`XREAD STREAMS <key> <id>` returns entries with ID strictly greater than `<id>`).

backend/src/meho_backplane/api/v1/feed.py
```diff
@@ -0,0 +1,401 @@
...
+
+Disconnect handling
+===================
+
+Client disconnect propagates as :class:`asyncio.CancelledError` into
+the generator (Starlette cancels the request task on
+``http.disconnect``). The ``except`` arm returns cleanly without
+re-raising so the audit row at session end records the normal exit
+rather than a synthetic 500 — and Valkey's BLOCKing XREAD releases
+the connection from the pool the next event-loop tick.
```
zdamir: **[⚠️ M1]** — Module docstring contradicts the iter-1 re-raise fix (also flagged by CodeRabbit)

**What:** Module docstring lines 77-82:
> "Client disconnect propagates as `asyncio.CancelledError` into the generator (Starlette cancels the request task on `http.disconnect`). **The `except` arm returns cleanly without re-raising** so the audit row at session end records the normal exit rather than a synthetic 500 — and Valkey's BLOCKing XREAD releases the connection from the pool the next event-loop tick."

The iter-1 M1 fix replaced `return` with `raise` (line 354) to satisfy asyncio's cancellation contract + Sonar S7497, but the module docstring still describes the old `return` behavior. The two contradict each other; a future reader (T5 CLI watch implementer, T6 MCP resource implementer) inheriting this docstring would assume cancellation is swallowed.

**Why it matters:** Docstrings are the canonical contract for the module's behavior. A docstring that asserts one thing while the code does the opposite is worse than no docstring — it actively misleads downstream implementers about the cancellation/cleanup chain. The `audit row at session end records normal exit rather than synthetic 500` claim is still TRUE (because `http.response.start` was sent on the first yield before any cancellation point, so AuditMiddleware's buffered status is locked at 200), but the **mechanism** described — "returns cleanly without re-raising" — is wrong.

**Required change:** Rewrite the disconnect-handling paragraph to match what the code actually does. The audit-row-stays-200 invariant is preserved; only the mechanism description changes:

```diff
-Client disconnect propagates as :class:`asyncio.CancelledError` into
-the generator (Starlette cancels the request task on
-``http.disconnect``). The ``except`` arm returns cleanly without
-re-raising so the audit row at session end records the normal exit
-rather than a synthetic 500 — and Valkey's BLOCKing XREAD releases
-the connection from the pool the next event-loop tick.
+Client disconnect propagates as :class:`asyncio.CancelledError` into
+the generator (Starlette cancels the request task on
+``http.disconnect``). The ``except`` arm logs the structured
+``feed_subscriber_disconnected`` event for operator triage, then
+re-raises so the cancellation unwinds the task tree per asyncio's
+contract (Sonar S7497; Python 3.13+ asyncio re-issues cancellation
+if it goes unpropagated). The audit row at session end still records
+a clean 200 close because ``http.response.start`` was sent on the
+first yield — before any cancellation point — so AuditMiddleware's
+buffered status_code is already locked at 200 by the time the
+cancellation propagates. Valkey's BLOCKing XREAD releases the
+connection from the pool the next event-loop tick.
```

**Source:** The iter-1 M1 fix's commit message + the `except asyncio.CancelledError: raise` at line 354 are the canonical statement of the actual behavior; this docstring just needs to catch up.

zdamir (commented):
> **Note:** Posting as `--comment` because GitHub blocks formal `REQUEST_CHANGES` on one's own PR. Formal verdict below is **REQUEST_CHANGES**; orchestrator reads the result file, not this badge.

**Anchor task:** #310 — G6.1-T4 SSE endpoint `/api/v1/feed`
**Surface area:** backend (Python — FastAPI + asyncio + redis-py asyncio + SSE)
**Best-practices packs audited:** python (async-generator-cancellation, Valkey XREAD cursor semantics, SSE invariants)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `9c7082c1fa070bb6559dc777e40e7ebdd2326ff5`
**AI reviewer (CodeRabbit):** received in ~3 min on iter-2 push; **2 NEW findings** (cursor-no-advance + docstring-stale) plus 1 nitpick about regression-testing the cursor advance (folded into B1). iter-1 B1+M1+M2 threads are auto-resolved by CR (one shows resolved=true; the others moved to outdated as line numbers shifted).

### Acceptance criteria (iter-2 state)
- [x] AC #1-2 — 200 + `text/event-stream`, per-tenant stream key from JWT — unchanged, still passing.
- [x] AC #3 — Tenant boundary — by construction, unchanged.
- [x] AC #4-5 — Filter by `op_class` / `principal` — unchanged, still passing.
- [⚠️] AC #6-7 — Replay via `since=<entry-id>` + `Last-Event-Id` — **broken under the busy-but-filtered scenario by B1**: cursor doesn't advance past filtered/malformed entries, so the same batch gets re-read indefinitely and matching events that arrive later sit behind the never-advancing cursor.
- [x] AC #8 — Heartbeat every 30s during quiet periods — iter-1 B1 fix is sound; `test_heartbeat_emitted_when_all_entries_filter_out` pins it.
- [x] AC #9 — Client disconnect cleanly closes — iter-1 M1 fix landed correctly (`raise` instead of `return`). Caveat: docstring still describes the old behavior — M1 below.
- [x] AC #10-11 — Unauthenticated → 401, read_only → 403 — unchanged.
- [⚠️] AC #12 — CI green — SonarCloud Quality Gate down to **4.6% Duplication** only (Reliability rating C → A — iter-1 M1+M2 fixes resolved that side). Duplication is the test JWT-minting boilerplate, marked adjacent-improvement on iter-1; not a code defect. Three other CI lanes still fail on Docker Hub 429 rate-limit (infrastructure flakes, see Integration verification).

### Project-board alignment
|---|---|---|---|---|---|
| #310 (T4) | open | `Closes #310` | ✅ | ⚠️ partial (B1 cursor-advance contract gap) | close on merge after B1 fix |
| #228 (Initiative G6.1) | open | parent | — | ⚠️ T4 of 6 | tick T4 checkbox post-merge |

### Predicted vs. actual diff (iter-2 delta from iter-1)
- ✅ `feed.py` — B1 heartbeat-clock fix (emitted_any flag) lands cleanly; M2 `_process_entries` helper extraction reduces cognitive complexity well under 15; M1 `raise` replaces `return`. But the **cursor-advance bug** (new iter-2 B1) was introduced as a side-effect of the M2 refactor: the helper now consumes filtered entries internally, so the main loop's `cursor = entry_id` never observes those entries.
- ✅ `test_api_v1_feed.py` — `test_heartbeat_emitted_when_all_entries_filter_out` pins the iter-1 B1 contract correctly; `_collect_n_frames` now catches `TimeoutError` from `asyncio.timeout` (M1 propagation).
- ⚠️ `feed.py` module docstring — describes old `return` cancellation behavior; needs updating to match the iter-1 M1 fix (M1 below).

### Integration verification
- [x] `ruff format --check src tests` — clean (109 files)
- [x] `mypy src` strict — 56 source files, no issues
- [x] `pytest tests/test_api_v1_feed.py` (mocked layer) — 14 passed
- [x] `pytest --ignore=tests/integration` — 553 passed locally (+15 vs T3 baseline), 1 skipped, 1 deselected
- [⚠️ infrastructure-fail] **CI Python lane** — still failed pulling `pgvector/pgvector:pg16` (Docker Hub 429). Same root cause as iter-1; the failing test (`test_mcp_inspector.py::test_full_mcp_lifecycle_succeeds`) is pre-existing G0.5-T6 code unrelated to this PR.
- [⚠️ infrastructure-fail] **CI Semgrep / image build** — same Docker Hub 429 wall.
- [⚠️ partial-pass] **SonarCloud Quality Gate** — down to ONE failed condition (4.6% Duplication on New Code; 3% threshold). Reliability rating bumped back to A (iter-1 M1+M2 fixes worked). The Duplication is in test JWT-minting boilerplate — marked adjacent improvement on iter-1, not a blocker. Could be cleared by folding the setup into a shared fixture, but that's a follow-up.

- 🛑 **B1** — `feed.py:329-337` — `_process_entries` consumes filtered/malformed entries internally; `cursor = entry_id` only updates on yielded frames. Replay paths (`since=<id>` or `Last-Event-Id`) re-read the same batch indefinitely. Required change: advance `cursor = items[-1][0]` for every consumed entry batch (before the `for ... yield` loop), regardless of which entries the filter accepted. Add regression test pinning forward progress across a filtered-then-matching pair of XREAD batches.
- ⚠️ **M1** — `feed.py:77-82` (module docstring) — describes the old `return` cancellation behavior; iter-1 M1 fix now `raise`s. Required change: rewrite the disconnect-handling paragraph to match the actual `raise` mechanism + reaffirm the audit-row-200 invariant (still holds via `http.response.start` being sent on first yield).

Same as iter-1 (still applicable, none introduced by iter-2):

- Sonar S8410 (Annotated FastAPI type hints) — codebase-wide style decision; existing chassis routes don't use it. Defer to a focused refactor PR.
- Sonar S7483 (timeout parameter on `_collect_n_frames`) — false positive; the parameter forwards to `asyncio.timeout(timeout)` which IS the rule's recommended idiom. Defer.
- Sonar 4.6% Duplication — JWT-minting boilerplate. Folding into `tests/_oidc_jwt_helpers` factory closes the gap; worth doing in T5/T6 if those tests grow similar boilerplate.
- CI Docker Hub rate-limit (3 lanes) — infrastructure flake; durable fix is Docker Hub auth on the runner (separate ops issue).

- Adopted: 1 (CR-iter2 cursor-no-advance → B1 with severity bumped to blocker because it breaks AC #6/#7 under the busy-filtered scenario, not just sub-optimal).
- Adopted: 1 (CR-iter2 docstring-stale → M1 as Major doc-precision).
- Adopted: 1 (CR-iter2 nitpick about cursor regression test — folded into B1's required change).
- Disputed: 0.
- Out of scope: 0.

iter-1 threads (B1/M1/M2) — auto-marked outdated by CR as the lines shifted post-fix; one thread shows resolved=true (M1).

### Docs
- ⚠️ `backend/src/meho_backplane/api/v1/feed.py` module docstring — describes pre-M1 `return` behavior (see M1 above).
- [x] `docs/codebase/backend.md` — T4 paragraph unchanged from iter-1, still accurate.

- **Iter-1 M2 refactor (`_process_entries` helper extraction) was clean** — cognitive complexity is well under 15 and the main loop reads naturally. The cursor-advance bug it incidentally introduced is a refactor-side-effect, not a structural problem with the refactor itself.
- **`test_heartbeat_emitted_when_all_entries_filter_out`** is a tight pinning test for the iter-1 B1 contract: time-compressed via `monkeypatch.setattr(_HEARTBEAT_INTERVAL_SECONDS, 0.05)` so it runs in <1 s while exercising the exact code path the production 30 s window does.

backend/src/meho_backplane/api/v1/feed.py
zdamir: Fixed in `3d2b0fb` — cursor now advances to `items[-1][0]` BEFORE the post-helper yield loop, regardless of which entries the filter accepted. Regression test `test_cursor_advances_past_filtered_entries` records cursors passed to xread on each call and asserts the second call's cursor is the LAST entry id of the first batch (not the original subscriber cursor); the yielded frame comes from the second batch, proving forward progress.

backend/src/meho_backplane/api/v1/feed.py
zdamir: Fixed in `3d2b0fb` — rewrote the disconnect-handling paragraph of the module docstring to describe the actual `raise` mechanism after the structured log, reaffirming the audit-row-200 invariant via `http.response.start` being sent on first yield before any cancellation point. Doc now matches the iter-1 M1 fix's behavior verbatim.

zdamir:
@coderabbitai review

backend/src/meho_backplane/api/v1/feed.py
```diff
@@ -0,0 +1,426 @@
...
+    machinery (e.g. server-side bridges that consume the feed via
+    plain HTTP). When both are present the header wins because it
+    encodes "I'm continuing a session I already had"; the query
+    parameter is the explicit override.
+    """
+    if last_event_id_header:
+        return last_event_id_header
+    if since:
+        return since
+    return _LIVE_TAIL_CURSOR
```
zdamir, line 288: **[🛑 B1]** — Invalid `Last-Event-Id` / `since` cursors cause SSE-client reconnect loops (also flagged by CodeRabbit)

**What:** Lines 211-230 in `feed_endpoint`:
```python
cursor = _resolve_cursor(
    last_event_id_header=request.headers.get("Last-Event-Id"),
    since=since,
)
```
The cursor is passed verbatim to `XREAD`. Both `Last-Event-Id` (header) and `since` (query param) are operator-controlled strings; nothing validates them as Valkey stream IDs before they reach `xread`.

When a client sends a malformed cursor (anything other than `$` or `<int>(-<int>)?`):

1. `xread({stream: "garbage"}, block=...)` propagates a `redis.exceptions.ResponseError` ("Invalid stream ID specified as stream command argument").
2. The generator's `try / except CancelledError` doesn't catch it — the unhandled exception bubbles up to Starlette's `StreamingResponse`.
3. Because `http.response.start` was sent on the FIRST yield attempt (which is *after* the bad cursor was fed to xread), the response was either never sent (XREAD raised on cycle 1 before any yield) or terminated mid-stream with a connection drop.
4. Browsers' built-in `EventSource` auto-reconnects on stream drops with the **same `Last-Event-Id`** the server last sent — but the bad cursor came from the client side, so the reconnect replays the same garbage and the loop tightens.

**Why it matters:** This is an SSE-EventSource-spec-mandated reconnect loop with no recovery path. A misbehaving client (or a malformed value cached in a downstream LB header) can hot-loop reconnections at the SSE reconnect interval (~3s default). On a busy backplane this multiplies into hundreds of XREAD-errors-per-second across all bad-cursor clients. The audit_log layer doesn't catch this either — the request never reaches the per-cycle code that would write a row, since XREAD blows up before the first yield.

The fix is to validate cursors **at the route boundary** (before the generator starts) and return a clean 400 so SSE clients escalate (EventSource flips `readyState=CLOSED` on 4xx-class responses and stops auto-reconnecting, per the spec).

**Required change:** Add a regex validator and call it in the route handler. CodeRabbit's diff is the minimal shape:

```python
import re

_VALKEY_STREAM_ID_RE: Final[re.Pattern[str]] = re.compile(r"^\d+(?:-\d+)?$")

def _validate_cursor_or_400(cursor: str) -> str:
    """Accept ``$`` or a Valkey stream id; reject anything else with HTTP 400."""
    if cursor == _LIVE_TAIL_CURSOR or _VALKEY_STREAM_ID_RE.fullmatch(cursor):
        return cursor
    raise HTTPException(
        status_code=400,
        detail="invalid_cursor: expected Valkey stream id or '$'",
    )


@router.get("/feed")
async def feed_endpoint(...) -> StreamingResponse:
    cursor = _validate_cursor_or_400(
        _resolve_cursor(
            last_event_id_header=request.headers.get("Last-Event-Id"),
            since=since,
        )
    )
    ...
```

Add a regression test in `TestFeedEndpoint`:

```python
async def test_invalid_last_event_id_returns_400(
    self, _feed_env, monkeypatch
) -> None:
    """Malformed cursor → 400 (no reconnect loop)."""
    install_fake_vault(monkeypatch)
    key = make_rsa_keypair("kid-feed-bad-cursor")
    token = mint_token(
        key, sub="op", tenant_id=str(_TENANT_A),
        tenant_role=TenantRole.OPERATOR.value,
    )
    app = _build_app()
    with respx.mock as mock_router:
        mock_discovery_and_jwks(mock_router, public_jwks(key))
        async with _make_client(app) as client:
            response = await client.get(
                "/api/v1/feed",
                headers={
                    "Authorization": f"Bearer {token}",
                    "Last-Event-Id": "abc-not-a-valkey-id",
                },
            )
    assert response.status_code == 400
    assert "invalid_cursor" in response.json()["detail"]
```

**Source:** WHATWG SSE spec §"Connection re-establishment" — https://html.spec.whatwg.org/multipage/server-sent-events.html#reestablish-the-connection (EventSource auto-reconnects on stream end; 4xx-class responses set `readyState=CLOSED` and stop auto-reconnect). Valkey XREAD argument validation — https://valkey.io/commands/xread/.

backend/src/meho_backplane/api/v1/feed.py
```diff
@@ -0,0 +1,426 @@
...
+need real-time feed.
+
+Heartbeat
+=========
+
+The generator emits ``: heartbeat\\n\\n`` (SSE comment line) every
+``_HEARTBEAT_INTERVAL_SECONDS`` of inactivity to keep intermediaries
+from idle-timing-out the connection. The heartbeat fires when the
+``XREAD BLOCK`` call returns no entries within the block window —
+the natural quiet-time signal Valkey gives the loop.
```
zdamir: **[⚠️ M1]** — Heartbeat module docstring is narrower than the implemented behavior (also flagged by CodeRabbit)

**What:** Lines 68-72 in the module docstring:
> "The generator emits `: heartbeat\n\n` (SSE comment line) every `_HEARTBEAT_INTERVAL_SECONDS` of inactivity to keep intermediaries from idle-timing-out the connection. **The heartbeat fires when the `XREAD BLOCK` call returns no entries within the block window** — the natural quiet-time signal Valkey gives the loop."

The "fires when XREAD BLOCK returns no entries" claim is now WRONG since iter-1 B1's outbound-silence fix. The actual implementation (lines 339-344, verified) emits a heartbeat in TWO scenarios:

1. `XREAD BLOCK` returned no entries (the original case — Valkey quiet).
2. `XREAD BLOCK` returned entries but **all filtered out for this subscriber's `op_class` / `principal` / `target`** (the iter-1 fix — Valkey busy, subscriber's outbound silent).

The second scenario is what the iter-1 `emitted_any` flag + `elif now - last_heartbeat >= _HEARTBEAT_INTERVAL_SECONDS` branch adds. The module docstring still describes only the first case.

**Why it matters:** Doc-vs-code precision matters for the same reason it mattered on iter-2 M1: T5 (CLI `--watch`) and T6 (MCP resource) implementers reading this docstring inherit the wrong mental model. A T5 implementer assuming "heartbeat = Valkey-quiet" might add their own keep-alive layer over the SSE stream, doubling heartbeats for the busy-tenant case. The iter-1 B1 fix is the most important behavioral guarantee the broadcast subsystem ships; the canonical docstring should describe it accurately.

**Required change:** Rewrite the Heartbeat section to describe the outbound-silence rule, naming both trigger paths:

```diff
 Heartbeat
 =========

 The generator emits ``: heartbeat\\n\\n`` (SSE comment line) every
-``_HEARTBEAT_INTERVAL_SECONDS`` of inactivity to keep intermediaries
-from idle-timing-out the connection. The heartbeat fires when the
-``XREAD BLOCK`` call returns no entries within the block window —
-the natural quiet-time signal Valkey gives the loop.
+``_HEARTBEAT_INTERVAL_SECONDS`` of **outbound silence** (no event
+frame yielded to the subscriber) to keep intermediaries from
+idle-timing-out the connection. Two scenarios trigger the heartbeat:
+
+1. **Valkey-quiet** — ``XREAD BLOCK`` returns no entries within the
+   block window.
+2. **Subscriber-filtered** — ``XREAD BLOCK`` returns entries but they
+   all fail the operator's ``op_class`` / ``principal`` / ``target``
+   filter, so the generator yields nothing outbound. Without this
+   second path, a noisy tenant with a narrow-filtered subscriber
+   would emit zero outbound bytes (events filtered, no heartbeat
+   either) and the nginx/ALB/CloudFront ~60 s idle timeout would
+   drop the connection.
+
+``last_heartbeat`` tracks the wall-clock of the last outbound yield
+(event frame or heartbeat); inbound XREAD activity that doesn't
+produce an outbound frame does NOT reset it.
```

**Source:** The iter-1 B1 fix's commit message and the `emitted_any` flag at the generator implementation (current head, around lines 333-344) are the canonical statement of the actual behavior; this docstring just needs to catch up.

backend/tests/test_api_v1_feed.py
```diff
@@ -0,0 +1,722 @@
...
+    post-M1 contract; ``asyncio.timeout`` converts that to
+    ``TimeoutError`` on its way out of the context.
+    """
+    frames: list[str] = []
+    try:
+        async with asyncio.timeout(timeout):
+            async for chunk in gen:
+                frames.append(chunk)
+                if len(frames) >= n:
+                    break
```
zdamir, line 246: **[💡 m1]** — `_collect_n_frames` consumes one frame when `n=0` (also flagged by CodeRabbit)

**What:** Lines 230-241 in `_collect_n_frames`:
```python
frames: list[str] = []
try:
    async with asyncio.timeout(timeout):
        async for chunk in gen:
            frames.append(chunk)              # ← runs BEFORE the n check
            if len(frames) >= n:
                break
```
The helper's contract is "up to *n* frames", but with `n=0` the `async for` body still runs once before the `len(frames) >= 0` break condition evaluates. A test passing `n=0` with an event-producing mock would unexpectedly collect 1 frame.

**Why it matters:** Today no test trips this — every `n=0` call site uses a mock that returns empty / sleeps idle, so the loop body never executes (timeout fires first). But the helper's contract is documented as "up to n frames" and a future test that combines `n=0` with an event-producing mock would silently accept 1 frame, masking a real test regression. The helper is shared infrastructure across `TestFeedGenerator`'s six body-shaping tests; precision matters.

**Required change:** Early-return when `n <= 0` (CodeRabbit's diff verbatim):

```python
async def _collect_n_frames(
    gen: AsyncIterator[str],
    *,
    n: int,
    timeout: float = 1.0,
) -> list[str]:
    """Read up to *n* SSE frames from *gen* and aclose() it cleanly."""
    frames: list[str] = []
    if n <= 0:
        await gen.aclose()
        return frames
    try:
        async with asyncio.timeout(timeout):
            async for chunk in gen:
                frames.append(chunk)
                if len(frames) >= n:
                    break
    except TimeoutError:
        pass
    finally:
        await gen.aclose()
    return frames
```

The `if n <= 0` branch also cleans up the generator (calls `aclose`) so cursor-pass-through tests that pass `n=0` to capture the xread call's cursor without iterating frames stay symmetric with the n>0 path.

**Source:** Helper docstring's own "up to *n* frames" contract — the early-return makes the contract honest at every input boundary.

zdamir (commented):
> **Note:** Posting as `--comment` because GitHub blocks formal `REQUEST_CHANGES` on one's own PR. Formal verdict below is **REQUEST_CHANGES**.

**Anchor task:** #310 — G6.1-T4 SSE endpoint `/api/v1/feed`
**Surface area:** backend (Python — FastAPI + asyncio + redis-py asyncio + SSE)
**Best-practices packs audited:** python (operator-input validation at route boundary, doc-vs-code precision, test helper contract)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `3d2b0fb364a83f67a177a76b8f27a791aff3e742`
**AI reviewer (CodeRabbit):** received in ~3 min; **3 new findings** (cursor-validation Major bumped to Blocker, heartbeat-docs Minor escalated to Major doc-precision, helper-n<=0 Minor) + 1 nitpick about a 200 happy-path HTTP test (adjacent improvement).

### iter-2 fixes verified
- ✅ Cursor advances past filtered entries — `test_cursor_advances_past_filtered_entries` records cursors per xread call; asserts forward progress.
- ✅ Module docstring describes the `raise` mechanism — updated at lines 77-89.

### Acceptance criteria (iter-3 state)
- [x] AC #1-5, #8-11 — unchanged from iter-2 verified state.
- [⚠️] AC #6-7 — Replay via `since=<entry-id>` + `Last-Event-Id` — iter-2 B1 (cursor advance) is fixed, but **iter-3 B1 surfaces a new gap**: cursors aren't validated against the Valkey ID format. Garbage cursors propagate to xread, which raises mid-stream → SSE auto-reconnect loop with no recovery path. See B1.
- [⚠️] AC #12 — CI green — SonarCloud QG down to 4.1% Duplication (down from 4.6% iter-2 / 5.2% iter-1) — only ONE failed condition. Reliability still A. The 3 Docker Hub 429 lanes still fail (infra; iter-1/2/3 stance unchanged).

### Project-board alignment
Unchanged from iter-2: #310 closes on merge; #228 tick T4 post-merge.

### Predicted vs. actual diff (iter-3 delta from iter-2)
- ✅ `feed.py` — cursor advance fix (B1) lands cleanly; docstring update (iter-2 M1) lands cleanly. Both verified end-to-end.
- ⚠️ `feed.py` — new iter-3 findings (cursor validation missing, heartbeat docstring outdated post-iter-1 B1).
- ✅ `test_api_v1_feed.py` — `test_cursor_advances_past_filtered_entries` correctly pins the iter-2 contract.
- ⚠️ `test_api_v1_feed.py` — `_collect_n_frames` `n=0` precision (m1 Minor).

### Integration verification
- [x] `ruff format --check src tests` — clean (109 files)
- [x] `mypy src` strict — 56 source files, no issues
- [x] `pytest tests/test_api_v1_feed.py` (mocked layer) — 15 passed
- [x] `pytest --ignore=tests/integration` — **554 passed** (+15 vs T3 baseline), 1 skipped, 1 deselected
- [⚠️ infra-fail] CI Python / Semgrep / image build — same Docker Hub 429 wall as iter-1/2 (pre-existing test_mcp_inspector unrelated to T4).
- [⚠️ partial-pass] SonarCloud QG — 4.1% Duplication only (Reliability A); test JWT-minting boilerplate, adjacent improvement.

- 🛑 **B1** — `feed.py:211-230` — `Last-Event-Id` header and `since` query param pass to xread without validation. Malformed cursors (anything other than `$` or `<int>(-<int>)?`) raise `redis.ResponseError` mid-stream; SSE EventSource clients auto-reconnect with the same bad cursor → reconnect loop. Required change: add `_validate_cursor_or_400` regex validator + raise HTTPException(400) on bad input; regression test in TestFeedEndpoint asserts 400 on `Last-Event-Id: abc-not-a-valkey-id`.
- ⚠️ **M1** — `feed.py:68-72` (Heartbeat module docstring) — narrower than implemented behavior. Iter-1 B1 fix added the busy-but-filtered trigger path; docstring still describes only the Valkey-quiet path. Required change: rewrite Heartbeat section to name both trigger paths explicitly (Valkey-quiet + Subscriber-filtered) and the outbound-silence invariant.
- 💡 **m1** — `tests/test_api_v1_feed.py:230-241` (`_collect_n_frames`) — when `n=0`, the loop body still consumes one frame before the break check. Helper contract is "up to *n* frames" — currently honest only for `n>=1`. Required change: `if n <= 0: await gen.aclose(); return frames` early-return before the `async for`.

- CodeRabbit nitpick: add a 200 happy-path HTTP test in `TestFeedEndpoint`. The integration test (`TestFeedIntegration.test_publish_then_generator_read`) already exercises the full 200 + content-type chain under Docker; a non-Docker version would lock it in for CI environments without Docker. Worth a quick test in T5/T6 if mocked-xread + httpx.stream + early-close-no-iterate pattern proves stable; defer for now.
- Sonar S8410 (4 minors) — Annotated FastAPI type hints — codebase-wide style decision; defer. Iter-1/2/3 stance unchanged.
- Sonar S7483 — timeout parameter on `_collect_n_frames` — likely false positive. Iter-1/2/3 stance unchanged.
- Sonar Duplication 4.1% — test JWT-minting boilerplate. T5/T6 may grow the same shape; fold into shared fixture then. Not blocking on iter-3.
- CI Docker Hub rate-limit — ops issue; durable fix is Docker Hub auth on runners.

- Adopted: 3 (CR's cursor-validation Major → B1 with severity bump to blocker because invalid cursors cause SSE reconnect loops per the EventSource spec; CR's heartbeat-docs Minor → M1 because doc precision matters for T5/T6 implementers inheriting this docstring; CR's `n<=0` Minor → m1 verbatim).
- Adopted: 1 nitpick → recorded as adjacent improvement (200 happy-path HTTP test in `TestFeedEndpoint`).
- Disputed: 0.

### Docs
- ⚠️ `backend/src/meho_backplane/api/v1/feed.py` Heartbeat module docstring — needs the iter-3 M1 update.
- [x] iter-2 M1 (disconnect-handling docstring) — fixed correctly in this iter, verified.
- [x] `docs/codebase/backend.md` T4 paragraph — unchanged from iter-1, still accurate.

- **iter-2 B1 fix (cursor advance) lands cleanly.** The `cursor = items[-1][0]` placement BEFORE the yield loop is the right shape — symmetric with the heartbeat path's "always evaluate the wall-clock check after one inbound cycle, regardless of outbound yields" invariant. The regression test records cursors per xread call and asserts forward progress, which is the right level of granularity.
- **iter-2 M1 fix (disconnect docstring) precisely matches the iter-1 code.** Captures both the `raise` mechanism AND the audit-row-200 invariant (via `http.response.start` on first yield).

backend/src/meho_backplane/api/v1/feed.py
zdamir: **Resolved in `ed4be59`.**

- Added `_VALKEY_STREAM_ID_RE = re.compile(r'^\d+(?:-\d+)?$')` and `_validate_cursor_or_400(cursor)` in `feed.py`; the helper accepts `$` (live-tail) or a matching Valkey stream id and raises `HTTPException(400, 'invalid_cursor: ...')` otherwise.
- Wrapped `_resolve_cursor(...)` in `feed_endpoint` so both `Last-Event-Id` header and `since` query param pass through the same gate.
- Two regression tests in `TestFeedEndpoint`:
  - `test_invalid_last_event_id_returns_400` — `Last-Event-Id: abc-not-a-valkey-id` → 400 + detail starts with `invalid_cursor`.
  - `test_invalid_since_query_param_returns_400` — `?since=drop-table;` → 400 (same gate).

Per the WHATWG SSE spec, a 4xx flips `EventSource.readyState=CLOSED`, which breaks the reconnect loop on bad input.

backend/src/meho_backplane/api/v1/feed.py
zdamir: **Resolved in `ed4be59`.**

Rewrote the Heartbeat section of the module docstring (`feed.py` lines 65–85) to name both trigger paths and the outbound-silence invariant:

1. **Valkey-quiet** — `XREAD BLOCK` returns no entries.
2. **Subscriber-filtered** — entries arrive but every one is rejected by the `op_class` / `principal` / `target` filter, so `_process_entries` yields nothing outbound.

Both paths reach the same wall-clock check (`monotonic() - last_heartbeat > HEARTBEAT_SECONDS`). `last_heartbeat` is updated on every outbound yield (events, heartbeats, comments) and on the once-per-cycle `monotonic()` reset at the top of the loop — it tracks outbound silence, never inbound XREAD activity.

backend/tests/test_api_v1_feed.py
zdamir: **Resolved in `ed4be59`.**

Added the early-return verbatim:

```python
if n <= 0:
    await gen.aclose()
    return frames
```

The helper's "up to *n* frames" contract is now honest at every input boundary. `test_cursor_passes_through_to_xread` was using `n=0` as a hack to drive one xread cycle without iterating frames — switched it to `n=1` with a real event entry, same cursor assertion, cleaner shape.

zdamir:
@coderabbitai review

zdamir:
@coderabbitai review

backend/tests/test_api_v1_feed.py
```diff
@@ -0,0 +1,819 @@
...
+        with respx.mock as mock_router:
+            mock_discovery_and_jwks(mock_router, public_jwks(key))
+            async with _make_client(app) as client:
+                response = await client.get(
+                    "/api/v1/feed",
+                    headers={"Authorization": f"Bearer {token}"},
+                )
+        assert response.status_code == 403
+        assert response.json() == {"detail": "insufficient_role"}
+
```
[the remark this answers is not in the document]
zdamir: This is like ping pong, If I refactor this one I will get Code Quality fail. DENIED

zdamir (commented):
> **Note:** Posting as `--comment` because GitHub blocks formal `APPROVE` on one's own PR. Formal verdict below is **APPROVE**.

## /review-pr-slim summary (iter-5)



**Anchor task:** #310 — G6.1-T4 SSE endpoint `GET /api/v1/feed`
**Surface area:** backend (Python — FastAPI + asyncio + redis-py asyncio + SSE)
**Verdict:** APPROVE
**Head SHA reviewed:** `7aaa2d97d4bb4a872bee86f012d2eae6bf8744dd`

### Iteration history
- **iter-1** (commit `70c6058`): initial PR — B1 heartbeat (busy-but-filtered tenants), M1 disconnect re-raise (S7497), M2 cognitive-complexity helper extraction.
- **iter-2** (commit `9c7082c`): iter-1 fixes verified — B1 cursor-advance gap surfaced (XREAD re-reads filtered batch forever under explicit-cursor replay), M1 disconnect-handling docstring narrower than implementation.
- **iter-3** (commit `3d2b0fb`): iter-2 fixes verified — B1 route-boundary cursor validation gap (malformed cursors propagated to xread → SSE EventSource reconnect loop), M1 heartbeat docstring still narrower than iter-1 implementation, m1 `_collect_n_frames` n≤0 contract.
- **iter-4** (commit `ed4be59`): iter-3 fixes verified — `_VALKEY_STREAM_ID_RE` + `_validate_cursor_or_400` lands cleanly at route boundary, heartbeat docstring names both trigger paths, helper early-returns on n≤0.
- **iter-5 follow-up** (commits `067f182`, `7aaa2d9`): SonarCloud Duplication on New Code (4.6% → 3.7% → 2.0% across iter-3/4/5 follow-ups). Two helpers extracted: `_request_feed_authenticated` (HTTP-edge scaffold) and `_drive_generator_with_one_batch` (skip-path scaffold). QG now green.

### Acceptance criteria
- [x] AC #1-2 — 200 + `text/event-stream` for authenticated operator + JWT-derived stream key — verified end-to-end via `TestFeedIntegration.test_publish_then_generator_read` (Docker-gated); unit-level `TestFeedEndpoint` covers auth gates (401/403/400) and unit-level `TestFeedGenerator` covers generator output. The pure-Python 200-happy-path HTTP test was attempted (iter-3, then iter-5) but the `httpx.AsyncClient.stream` + ASGITransport cancellation handshake is racy enough to tight-loop the mocked generator; CodeRabbit acknowledged this and withdrew the recommendation at 14:54Z ("Acknowledged, no further push on this one").
- [x] AC #3-5 — Tenant boundary + filter by `op_class` / `principal` / `target` — TestFeedGenerator parametric cases.
- [x] AC #6-7 — Replay via `Last-Event-Id` / `since` cursors — TestResolveCursor + TestFeedGenerator. Iter-4 added the `_validate_cursor_or_400` gate; cursors are now Valkey-stream-id-shape-checked at the route boundary so malformed input returns 400 (flips SSE `readyState=CLOSED` per the spec) instead of dropping into a reconnect loop.
- [x] AC #8 — Heartbeat `: heartbeat\n\n` — verified via `test_heartbeat_emitted_when_all_entries_filter_out` with `_HEARTBEAT_INTERVAL_SECONDS` monkeypatched to 0.05.
- [x] AC #9 — Disconnect cleanup — verified by every test that drains the generator (which calls `aclose()`); the `except asyncio.CancelledError: log + raise` arm is exercised on every termination path. Iter-1 fixed the original swallow-and-return (S7497).
- [x] AC #10-11 — Unauthenticated → 401, read_only → 403, defensive parse — TestFeedEndpoint + `test_malformed_event_skipped` + `test_unknown_field_shape_skipped`.
- [x] AC #12 — CI green: **all 10 checks pass**: Python lane, SonarCloud, DCO, CodeRabbit, Semgrep SAST + OSS, Build (backplane image), Helm, Go, TruffleHog. The Docker Hub 429 wall that broke iter-1/2/3 has cleared.

### Project-board alignment
|---|---|---|---|---|---|
| #310 (evoila/meho) | OPEN | `Closes #310` | ✅ same-repo, default branch | ✅ full | close on merge |
| #228 (parent initiative) | OPEN | parent | manual | ✅ T4 complete | tick T4 checkbox post-merge, leave open for T5/T6 |

### Predicted vs. actual diff (iter-5 delta from iter-4)
- ✅ `backend/tests/test_api_v1_feed.py` — extracted `_drive_generator_with_one_batch` helper; `test_malformed_event_skipped` and `test_unknown_field_shape_skipped` collapse from ~28 lines each to ~7. Net `-3` lines, removes the last 24-line SonarCloud duplication block. Behavior-preserving (both tests still assert `id: 1715600000001-0` on the surviving frame).
- ✅ `backend/src/meho_backplane/api/v1/feed.py` — unchanged since iter-4.

### Integration verification
- [x] `ruff format --check src tests` — clean (109 files)
- [x] `mypy src` strict — 56 source files, 0 issues
- [x] `pytest tests/test_api_v1_feed.py` (mocked layer) — 17 passed
- [x] `pytest --ignore=tests/integration` — **556 passed**, 1 skipped, 1 deselected
- [x] CI Python lane — pass
- [x] CI SonarCloud — pass (duplication 1.6%, below the 3% threshold)
- [x] CI image build, Semgrep, Go, Helm, TruffleHog — all pass
- [⚠️ not run locally] `TestFeedIntegration.test_publish_then_generator_read` — Docker-gated; runs in CI when Docker Hub is reachable (was unreachable on iter-1/2/3 due to 429s, reachable on iter-5).
- [⚠️ not run] `docker compose up` end-to-end — not part of this PR's surface (T5/T6 will wire the consumers).
- [⚠️ not run] browser UI flow — not applicable (server-side route only).

### Findings (inline)
- 🛑 0 blockers
- ⚠️ 0 majors
- 💡 0 minors
- ❓ 0 questions

- Pure-Python 200 + `text/event-stream` HTTP test in `TestFeedEndpoint` — CodeRabbit raised this three iterations in a row, then withdrew at 14:54Z. The cleanest pattern that sidesteps the httpx/ASGITransport cancellation race is `patch("meho_backplane.api.v1.feed._feed_generator", return_value=<finite_async_iter>)` so `StreamingResponse` drains a finite iterator naturally (no `client.stream` context-manager exit race). Worth trying in T5/T6 once the broader mocked-xread + httpx pattern stabilizes; the Docker-gated integration test continues to cover the full chain.
- CI Docker Hub rate-limit was the iter-1/2/3 wall; durable fix is Docker Hub auth on runners (DOCKERHUB_USER + DOCKERHUB_PAT secrets + `docker/login-action` step). It cleared on iter-5 organically, but the underlying ops gap remains.
- Sonar S8410 (4 minors, Annotated FastAPI hints) — codebase-wide style decision, defer.
- Sonar S7483 (timeout parameter on `_collect_n_frames`) — likely false positive; the parameter forwards to `asyncio.timeout`.

### Docs
- [x] `backend/src/meho_backplane/api/v1/feed.py` module docstring — Heartbeat (iter-3 M1) and Disconnect-handling (iter-2 M1) sections accurately describe the iter-4 implementation; verified end-to-end.
- [x] `docs/codebase/backend.md` T4 paragraph — unchanged from iter-1, still accurate.

**Auto-close (should happen automatically)**
- #310 — `Closes #310` keyword on PR body, default-branch base, same-repo. Verify it actually closed via `gh issue view 310 --repo evoila/meho`.

- #228 parent initiative — tick the T4 checkbox in the body, leave the issue OPEN for T5 (#311) and T6 (#312).

- #310: "Closed by #353. SSE `GET /api/v1/feed` with per-tenant scoping, op_class/principal/target filters, Last-Event-Id replay, route-boundary cursor validation (400 on bad input, breaks SSE auto-reconnect loops), heartbeat-on-outbound-silence, defensive parse of malformed entries, RBAC operator-minimum. 17 unit tests + 1 Docker-gated end-to-end."

- **Iter-1 → iter-5 trajectory.** Five iterations, each closing the specific gap the prior review opened — no churn, no scope creep, no abandoned fixes. The cursor-advance fix (iter-3), cursor-validation fix (iter-4), and helper extractions (iter-5) are surgical: each one is a minimal change that addresses one named contract.
- **Iter-4 M1 heartbeat docstring** captures both trigger paths (Valkey-quiet + Subscriber-filtered) and the outbound-silence invariant. Future T5/T6 implementers inheriting this code get the contract on the first read.
- **Iter-5 helper extraction** is the cleanest possible shape for the duplication: keyword-only filter args, hardcoded `cursor="$"` because callers don't test cursor behavior here (cursor-specific tests stay on the direct-generator path). Net `-3` lines, gain is structural.

kr3s0 (approved):

PR #362, opened 2026-05-13
feat(cli): targets import verb — bulk-import targets.yaml with --update/--dry-run/--json (G0.3-T6)

### LINKED ISSUE #257: Task: G0.3-T6 Migration tool — meho targets import <file> with --update / --dry-run / --json (bulk-imports targets.yaml)
## ⚠️ Re-scope notice (2026-05-14)

Parent Initiative [#224 G0.3 was amended](https://github.com/evoila/meho/issues/224). Two new columns on the `targets` table:
- `fingerprint JSONB` — cached fingerprint result from the last probe (vendor, product, version, build, reachable, probed_at, probe_method, extras). Consumed by [#388 G0.6](https://github.com/evoila/meho/issues/388) resolver to pick connector implementation.
- `preferred_impl_id TEXT` — optional operator override for which connector implementation to use when multiple impls advertise overlapping `(product, version)` ranges (G0.6 resolver tie-break).

This task's schema work (migration / Pydantic / API / CLI) should reflect both new columns. The probe verb writes `fingerprint` so G0.6's resolver can read it without re-probing. Body below preserved for reference; re-walk against the amended Initiative when executing.

---

Parent initiative: #224

## Target

Migration tool `meho targets import <file>` — ingests the consumer's existing `targets.yaml` shape into the backplane:

- `cli/internal/cmd/targets/import.go` — new cobra subcommand; reads YAML, posts one `POST /api/v1/targets` per entry.
- `backend/src/meho_backplane/api/v1/targets.py` — extends T3's POST endpoint with optional `--update` semantics for the import tool (idempotency).
- `cli/internal/cmd/targets/import_test.go` — Go unit tests for YAML parsing + dry-run logic.
- `backend/tests/test_api_v1_targets_import.py` — integration test of the bulk-import flow against the consumer's real `targets.yaml` file (downloaded in CI).

## Context

T6 closes the loop: a consumer with an existing `targets.yaml` runs `meho targets import rdc-hetzner-dc/targets.yaml --update` and their targets land in the backplane. Mandatory acceptance: a round-trip from `targets.yaml` → backplane → `meho targets describe --json` matches the input semantically.

**YAML schema parsed.** The consumer's `targets.yaml` shape per [docs/cross-repo/targets-yaml.md](docs/cross-repo/targets-yaml.md):

```yaml
targets:
  - name: rdc-vcenter
    aliases: [rdc, host-vcenter]
    product: vcenter
    host: vc-dc.evba.lab
    port: 443
    secret_ref: secret/rdc-hetzner-dc/vsphere/rdc-vcenter
    sso_realm: evba.lab    # → goes into extras.sso_realm
    vpn_required: true
    notes: |
      Host vCenter for the 4 physical Hetzner hosts.
      vCenter 8.0.3.
```

**Mapping rules:**
- `name` → `name` (required, validates min_length 1).
- `aliases` → `aliases` (array, optional, defaults to []).
- `product` → `product` (required).
- `host` → `host` (required).
- `port` → `port` (optional).
- `secret_ref` → `secret_ref` (optional).
- `vpn_required` → `vpn_required` (defaults to false).
- `notes` → `notes` (preserved as-is, multi-line YAML scalar handled).
- Every other field (e.g. `sso_realm`, `kubeconfig_field`, `account`, `project_id`, `impersonate_sa`) → spilled into `extras` JSONB.
- `auth_model` — if not specified in YAML, defaults to `shared_service_account` (the v0.2 default for tier-1 targets).

**Idempotency.** Three modes:
- **No flag** (default): if a target with the same name exists, abort the entire import (no partial write) and tell the operator about the conflict.
- **`--update`**: existing targets are updated via PATCH; new targets are POSTed. Mixed-mode safe.
- **`--dry-run`**: print the plan (CREATE / UPDATE / SKIP per entry); no API calls. Combined with `--json` for machine-parseable output.

**Plan output (dry-run).** Per consumer-needs.md style ("verification commands per step"):

```
$ meho targets import rdc-hetzner-dc/targets.yaml --dry-run
Plan: 25 targets in input, 0 existing in tenant 'rdc-internal'

  CREATE  rdc-vcenter        (product=vcenter,    host=vc-dc.evba.lab)
  CREATE  rdc-vault          (product=vault,      host=vault.evba.lab)
  CREATE  rke2-meho          (product=kubernetes, host=10.5.50.153)
  ... (22 more)

Run without --dry-run to apply.
```

**Bulk endpoint?** No — T6 uses individual POST/PATCH per entry. Bulk-write endpoints are a v0.2.next polish. The 25-target consumer import is ~25 HTTP requests; runtime is ~5s; acceptable.

**Tenant scoping.** Import uses the operator's tenant from their JWT. If an operator wants to import into a different tenant, they need to `meho focus <tenant>` first (a CLI concept that doesn't exist yet — for v0.2, the operator's JWT defines the tenant). Document this clearly.

## Current state

After T1-T5:
- API + CLI work end-to-end for individual targets.
- No bulk-import path. Operators would have to write a shell loop calling `meho targets create` for every YAML entry.

## Desired state

`cli/internal/cmd/targets/import.go` sketch (Go pseudocode):

```go
func newImportCmd() *cobra.Command {
    var update, dryRun, asJSON bool
    cmd := &cobra.Command{
        Use:   "import <file>",
        Short: "Import targets from a targets.yaml file",
        Args:  cobra.ExactArgs(1),
        RunE: func(cmd *cobra.Command, args []string) error {
            data, err := os.ReadFile(args[0])
            if err != nil {
                return fmt.Errorf("read %s: %w", args[0], err)
            }
            entries, err := parseTargetsYaml(data)
            if err != nil {
                return err
            }
            plan, err := buildPlan(cmd.Context(), entries, update)
            if err != nil {
                return err
            }
            if dryRun {
                return printPlan(plan, asJSON)
            }
            return executePlan(cmd.Context(), plan)
        },
    }
    cmd.Flags().BoolVar(&update, "update", false, "PATCH existing targets instead of erroring")
    cmd.Flags().BoolVar(&dryRun, "dry-run", false, "print plan; no API calls")
    cmd.Flags().BoolVar(&asJSON, "json", false, "output plan as JSON (use with --dry-run)")
    return cmd
}
```

YAML parsing extracts the fields per the mapping rules above; spills unknown fields into `extras`.

## Acceptance criteria

- [ ] `meho targets import rdc-hetzner-dc/targets.yaml` (default mode) succeeds against a real `targets.yaml` (consumer's 25-entry file, fetched via `curl https://raw.githubusercontent.com/evoila-bosnia/claude-rdc-hetzner-dc/main/rdc-hetzner-dc/targets.yaml` in CI).
- [ ] Re-running without `--update` errors with the conflict list.
- [ ] Re-running with `--update` succeeds; existing targets get PATCHed (notes updates verified).
- [ ] `--dry-run` prints the plan; no API calls happen (verified by audit_log row count unchanged).
- [ ] `--dry-run --json` outputs structured plan: `{create: [...], update: [...], skip: [...]}`.
- [ ] Round-trip: import a target → `meho targets describe <name> --json` → fields match the YAML input semantically (notes preserved, extras populated with sso_realm/etc.).
- [ ] Malformed YAML → CLI exits non-zero with parse error message.
- [ ] Missing required field (e.g. `host`) → CLI exits non-zero before any API call.
- [ ] Unknown YAML field → spilled into `extras`; not a parse error.
- [ ] CI: integration test against the consumer's real `targets.yaml` (or a snapshot of it pinned to a specific git SHA).
- [ ] Go unit tests + Python integration tests pass.

## Out of scope

- Export (`meho targets export > file.yaml`) — out of scope for v0.2 (operators can use `meho targets list --json | jq` for now).
- Bulk delete via YAML — out of scope.
- Migration *between* tenants — out of scope; operators import into their JWT's tenant.
- Schema validation against a Pydantic-equivalent on the CLI side — CLI does minimal local validation; API does the strict validation. Errors propagate from API.
- Watching `targets.yaml` for changes (auto-sync) — out of scope.

## References

- Parent Initiative: [#224](https://github.com/evoila/meho/issues/224).
- Depends on: T3 (API routes the import calls).
- Consumer's source-of-truth file: https://github.com/evoila-bosnia/claude-rdc-hetzner-dc/blob/main/rdc-hetzner-dc/targets.yaml (25+ entries; T6's primary test corpus).
- Existing cross-repo doc for the YAML schema: [docs/cross-repo/targets-yaml.md](docs/cross-repo/targets-yaml.md) (T6 may extend this doc with the import recipe).
- YAML parsing in Go: https://pkg.go.dev/gopkg.in/yaml.v3
- DevOps best-practices: `.claude/skills/implement-issue/devops_best_practices.md` — idempotent import tooling, dry-run mode discipline.
- Go best-practices for cobra: subcommand structure, flag conventions, error wrapping.

ddzafic:
## Summary

- Adds `meho targets import <file>` cobra subcommand (`cli/internal/cmd/targets/import.go`)
- Parses `targets.yaml` (top-level `targets:` list); unknown fields spill into `extras` JSONB
- Three modes: default (abort on conflict), `--update` (PATCH existing), `--dry-run` (print plan, no writes)
- `--dry-run --json` emits `{create: [...], update: [...], skip: []}` for machine-readable pipelines
- Adds `CreateTarget` / `UpdateTarget` to `cli/internal/api/targets.go` with JSON body + 401-retry
- Fixes `ListTargets` to pass `limit` query param (was declared but unused)
- Adds `gopkg.in/yaml.v3 v3.0.1` direct dependency to `cli/go.mod` (already in go.sum as transitive)
- 13 Go unit tests covering YAML parsing, dry-run, conflict abort, `--update`, no-creds
- 13 Python integration tests for POST/PATCH round-trip, duplicate 409, extras preservation, cross-tenant isolation
- Updates `docs/codebase/cli.md` module layout

Closes #257

Parent Initiative: #224

## Test plan

- [x] `uv run pytest backend/tests/test_api_v1_targets_import.py -q` → 13 passed
- [x] `uv run pytest backend/ -q --ignore=tests/test_observability.py --ignore=tests/integration/` → 609 passed, 1 skipped (no regressions)
- [ ] Go tests: `go test ./cli/internal/cmd/targets/... ./cli/internal/api/...` — verified by CI (Go not installed locally)

### DIFF
--- a/cli/internal/api/targets.go
+++ b/cli/internal/api/targets.go
@@ -4,12 +4,14 @@
 package api
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 )
 
@@ -65,6 +67,34 @@ type ListTargetsParams struct {
 	Cursor  *string
 }
 
+// TargetCreateRequest is the POST /api/v1/targets body.
+type TargetCreateRequest struct {
+	Name        string         `json:"name"`
+	Aliases     []string       `json:"aliases"`
+	Product     string         `json:"product"`
+	Host        string         `json:"host"`
+	Port        *int           `json:"port,omitempty"`
+	SecretRef   *string        `json:"secret_ref,omitempty"`
+	AuthModel   string         `json:"auth_model"`
+	VPNRequired bool           `json:"vpn_required"`
+	Extras      map[string]any `json:"extras"`
+	Notes       *string        `json:"notes,omitempty"`
+}
+
+// TargetUpdateRequest is the PATCH /api/v1/targets/{name} body.
+// Nil pointer fields serialize as JSON null, which instructs the
+// server to clear that column (Pydantic exclude_unset behaviour).
+type TargetUpdateRequest struct {
+	Aliases     []string       `json:"aliases"`
+	Host        string         `json:"host"`
+	Port        *int           `json:"port"`
+	SecretRef   *string        `json:"secret_ref"`
+	AuthModel   string         `json:"auth_model"`
+	VPNRequired bool           `json:"vpn_required"`
+	Extras      map[string]any `json:"extras"`
+	Notes       *string        `json:"notes"`
+}
+
 // ListTargets calls GET /api/v1/targets with a one-shot 401-retry.
 // Returns the typed slice on success. On error the response body is
 // parsed for a structured error detail where possible; otherwise
@@ -74,6 +104,9 @@ func (c *AuthedClient) ListTargets(ctx context.Context, params *ListTargetsParam
 	if params != nil {
 		if params.Product != nil && *params.Product != "" {
 			q.Set("product", *params.Product)
+		}
+		if params.Limit != nil {
+			q.Set("limit", strconv.Itoa(*params.Limit))
 		}
 		if params.Cursor != nil {
 			q.Set("cursor", *params.Cursor)
@@ -151,6 +184,44 @@ func (c *AuthedClient) ProbeTarget(ctx context.Context, name string) (*ProbeResu
 	}
 }
 
+// CreateTarget calls POST /api/v1/targets with a one-shot 401-retry.
+// Returns the newly created target on 201, or an error on any other status.
+func (c *AuthedClient) CreateTarget(ctx context.Context, req TargetCreateRequest) (*Target, int, error) {
+	resp, err := c.doRequestWithBody(ctx, http.MethodPost, "/api/v1/targets", req)
+	if err != nil {
+		return nil, 0, err
+	}
+	defer resp.Body.Close()
+	body, _ := io.ReadAll(resp.Body)
+	if resp.StatusCode == http.StatusCreated {
+		var t Target
+		if jerr := json.Unmarshal(body, &t); jerr != nil {
+			return nil, resp.StatusCode, fmt.Errorf("meho: decode created target: %w", jerr)
+		}
+		return &t, resp.StatusCode, nil
+	}
+	return nil, resp.StatusCode, errFromBody(body, resp.StatusCode)
+}
+
+// UpdateTarget calls PATCH /api/v1/targets/{name} with a one-shot 401-retry.
+// Returns the updated target on 200, or an error on any other status.
+func (c *AuthedClient) UpdateTarget(ctx context.Context, name string, req TargetUpdateRequest) (*Target, int, error) {
+	resp, err := c.doRequestWithBody(ctx, http.MethodPatch, "/api/v1/targets/"+url.PathEscape(name), req)
+	if err != nil {
+		return nil, 0, err
+	}
+	defer resp.Body.Close()
+	body, _ := io.ReadAll(resp.Body)
+	if resp.StatusCode == http.StatusOK {
+		var t Target
+		if jerr := json.Unmarshal(body, &t); jerr != nil {
+			return nil, resp.StatusCode, fmt.Errorf("meho: decode updated target: %w", jerr)
+		}
+		return &t, resp.StatusCode, nil
+	}
+	return nil, resp.StatusCode, errFromBody(body, resp.StatusCode)
+}
+
 // doGet makes an authenticated GET to backplaneURL+path with optional query
 // parameters, retrying once with a refreshed token on 401.
 func (c *AuthedClient) doGet(ctx context.Context, path string, query url.Values) (*http.Response, error) {
@@ -192,6 +263,42 @@ func (c *AuthedClient) rawRequest(ctx context.Context, method, path string, quer
 		req.Header.Set("Authorization", authorizationHeader(bearer))
 	}
 	req.Header.Set("Accept", "application/json")
+	return c.httpClient.Do(req)
+}
+
+// doRequestWithBody sends method+path with a JSON-marshalled body,
+// retrying once on 401 after a token refresh.
+func (c *AuthedClient) doRequestWithBody(ctx context.Context, method, path string, body any) (*http.Response, error) {
+	resp, err := c.rawRequestWithBody(ctx, method, path, body)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusUnauthorized {
+		return resp, nil
+	}
+	_ = resp.Body.Close()
+	if rerr := c.box.refresh(ctx); rerr != nil {
+		return nil, rerr
+	}
+	return c.rawRequestWithBody(ctx, method, path, body)
+}
+
+func (c *AuthedClient) rawRequestWithBody(ctx context.Context, method, path string, body any) (*http.Response, error) {
+	rawURL := strings.TrimRight(c.backplaneURL, "/") + path
+	b, err := json.Marshal(body)
+	if err != nil {
+		return nil, fmt.Errorf("meho: marshal request: %w", err)
+	}
+	req, err := http.NewRequestWithContext(ctx, method, rawURL, bytes.NewReader(b))
+	if err != nil {
+		return nil, fmt.Errorf("meho: build request: %w", err)
+	}
+	bearer := c.box.snapshot()
+	if bearer != "" {
+		req.Header.Set("Authorization", authorizationHeader(bearer))
+	}
+	req.Header.Set("Accept", "application/json")
+	req.Header.Set("Content-Type", "application/json")
 	return c.httpClient.Do(req)
 }
 
--- a/cli/internal/api/targets.go
+++ b/cli/internal/api/targets.go
@@ -81,20 +81,6 @@ type TargetCreateRequest struct {
 	Notes       *string        `json:"notes,omitempty"`
 }
 
-// TargetUpdateRequest is the PATCH /api/v1/targets/{name} body.
-// Nil pointer fields serialize as JSON null, which instructs the
-// server to clear that column (Pydantic exclude_unset behaviour).
-type TargetUpdateRequest struct {
-	Aliases     []string       `json:"aliases"`
-	Host        string         `json:"host"`
-	Port        *int           `json:"port"`
-	SecretRef   *string        `json:"secret_ref"`
-	AuthModel   string         `json:"auth_model"`
-	VPNRequired bool           `json:"vpn_required"`
-	Extras      map[string]any `json:"extras"`
-	Notes       *string        `json:"notes"`
-}
-
 // ListTargets calls GET /api/v1/targets with a one-shot 401-retry.
 // Returns the typed slice on success. On error the response body is
 // parsed for a structured error detail where possible; otherwise
@@ -204,9 +190,11 @@ func (c *AuthedClient) CreateTarget(ctx context.Context, req TargetCreateRequest
 }
 
 // UpdateTarget calls PATCH /api/v1/targets/{name} with a one-shot 401-retry.
+// body is a sparse map — only keys present in the map are sent to the server,
+// so the backend's exclude_unset logic leaves unmentioned fields untouched.
 // Returns the updated target on 200, or an error on any other status.
-func (c *AuthedClient) UpdateTarget(ctx context.Context, name string, req TargetUpdateRequest) (*Target, int, error) {
-	resp, err := c.doRequestWithBody(ctx, http.MethodPatch, "/api/v1/targets/"+url.PathEscape(name), req)
+func (c *AuthedClient) UpdateTarget(ctx context.Context, name string, body map[string]any) (*Target, int, error) {
+	resp, err := c.doRequestWithBody(ctx, http.MethodPatch, "/api/v1/targets/"+url.PathEscape(name), body)
 	if err != nil {
 		return nil, 0, err
 	}
--- a/cli/internal/cmd/targets/import.go
+++ b/cli/internal/cmd/targets/import.go
@@ -1,0 +1,411 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 evoila Group
+
+package targets
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+
+	"github.com/evoila/meho/cli/internal/api"
+	"github.com/evoila/meho/cli/internal/output"
+)
+
+// knownYAMLFields are extracted into dedicated importEntry fields;
+// everything else is spilled into importEntry.Extras.
+var knownYAMLFields = map[string]bool{
+	"name":         true,
+	"aliases":      true,
+	"product":      true,
+	"host":         true,
+	"port":         true,
+	"secret_ref":   true,
+	"vpn_required": true,
+	"notes":        true,
+	"auth_model":   true,
+}
+
+// importEntry is a single target parsed from a targets.yaml file.
+type importEntry struct {
+	Name        string
+	Aliases     []string
+	Product     string
+	Host        string
+	Port        *int
+	SecretRef   *string
+	VPNRequired bool
+	Notes       *string
+	AuthModel   string
+	Extras      map[string]any
+}
+
+// importAction describes what the import will do with a single entry.
+type importAction struct {
+	Action string // "CREATE" or "UPDATE"
+	Entry  importEntry
+}
+
+// importPlan holds the full set of actions resolved before any writes.
+type importPlan struct {
+	Actions   []importAction
+	Conflicts []string // existing names when --update is not set
+}
+
+func newImportCmd() *cobra.Command {
+	var (
+		update    bool
+		dryRun    bool
+		asJSON    bool
+		backplane string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "import <file>",
+		Short: "Bulk-import targets from a targets.yaml file",
+		Long: `Read a targets.yaml file and import each entry into the backplane.
+
+Default: if any target in the file already exists the entire import is
+aborted — no partial write. Use --update to PATCH existing targets
+instead of aborting.
+
+Use --dry-run to preview the plan without making write API calls.
+Combine with --json for machine-parseable output.
+
+Tenant scoping: targets are imported into the tenant carried by your
+JWT. To target a different tenant, run "meho login <url>" with the
+appropriate credentials first.`,
+		Args:          cobra.ExactArgs(1),
+		SilenceUsage:  true,
+		SilenceErrors: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			data, err := os.ReadFile(args[0])
+			if err != nil {
+				return fmt.Errorf("read %s: %w", args[0], err)
+			}
+			entries, err := parseTargetsYAML(data)
+			if err != nil {
+				return err
+			}
+			if len(entries) == 0 {
+				fmt.Fprintln(cmd.OutOrStdout(), "No targets found in file.")
+				return nil
+			}
+
+			bpURL, err := resolveURL(backplane)
+			if err != nil {
+				return output.RenderError(cmd.ErrOrStderr(), output.AuthExpired(err.Error()), asJSON)
+			}
+			client, err := api.NewAuthedClient(cmd.Context(), bpURL, api.AuthedClientOptions{})
+			if err != nil {
+				if api.IsTokenNotFound(err) {
+					return output.RenderError(cmd.ErrOrStderr(),
+						output.AuthExpired(fmt.Sprintf("no stored credentials for %s; run `meho login %s`", bpURL, bpURL)),
+						asJSON)
+				}
+				return output.RenderError(cmd.ErrOrStderr(),
+					output.Unexpected(fmt.Sprintf("build client: %v", err)),
+					asJSON)
+			}
+
+			plan, err := buildImportPlan(cmd.Context(), client, entries, update)
+			if err != nil {
+				if api.IsNoRefreshToken(err) {
+					return output.RenderError(cmd.ErrOrStderr(),
+						output.AuthExpired(fmt.Sprintf("token expired; run `meho login %s`", bpURL)),
+						asJSON)
+				}
+				return output.RenderError(cmd.ErrOrStderr(),
+					output.Unreachable(fmt.Sprintf("list existing targets: %v", err)),
+					asJSON)
+			}
+
+			if dryRun {
+				return printImportPlan(cmd, plan, asJSON)
+			}
+
+			if !update && len(plan.Conflicts) > 0 {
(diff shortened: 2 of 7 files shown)


cli/internal/cmd/targets/import.go
```diff
@@ -0,0 +1,411 @@
...
+		Aliases:     e.Aliases,
+		Host:        e.Host,
+		Port:        e.Port,
+		SecretRef:   e.SecretRef,
+		AuthModel:   e.AuthModel,
+		VPNRequired: e.VPNRequired,
+		Extras:      e.Extras,
+		Notes:       e.Notes,
+	}
+}
```
damir-topic, line 417: **🛑 Blocker** — `--update` PATCH always sends every column and clobbers existing values

**What:** `entryToUpdateRequest` populates every field of `api.TargetUpdateRequest` from `e`, including the defaulted-when-absent values (`AuthModel: "shared_service_account"`, `VPNRequired: false`, `Extras: map[string]any{}`, `Aliases: []string{}`, plus nil pointers for `Port` / `SecretRef` / `Notes`). The struct (`cli/internal/api/targets.go:85-94`) has no `omitempty` tags, so `json.Marshal` always emits all 8 keys. Nil pointers serialize as JSON `null`, empty maps as `{}`, empty slices as `[]`, zero bool as `false`, zero string as `""`.

**Why it matters:** The doc comment on `TargetUpdateRequest` claims "Pydantic exclude_unset behaviour" makes nulls clear columns — that's the wrong model. Per Pydantic v2 docs and verified locally: `exclude_unset=True` excludes fields that were **not provided in the input**, not fields whose value is `None`. A JSON `null` is "explicitly provided" and ends up in `model_fields_set`, so the server's `setattr(t, k, v)` loop (`backend/src/meho_backplane/api/v1/targets.py:232-233`) applies it.

Concrete consequence on the happy-path consumer flow: import a YAML entry that only specifies `name`, `product`, `host`, `notes`. The PATCH body Go emits is:

```json
{"aliases":[],"host":"...","port":null,"secret_ref":null,"auth_model":"shared_service_account","vpn_required":false,"extras":{},"notes":"..."}
```

If the existing target had `auth_model=impersonation, vpn_required=true, port=6443, secret_ref="secret/foo", extras={sso_realm:"evba.lab"}, aliases=["rdc","host-vc"]` — **every one of those is wiped**. The test that supposedly proves the `--update` flow (`TestImport_Update_PatchesExisting`) only asserts that the stdout contains the word "updated"; it never inspects the request body or the resulting target state. `test_import_update_only_supplied_fields` on the Python side tests the *server* correctly when given a `{"host":"..."}`-only body — but the CLI never sends that body.

**Required change:** Make the PATCH body sparse — only include keys the YAML actually specified. Two viable shapes; pick the minimal-diff one:

*Option A (minimal — drop the typed struct for PATCH):* Track field presence in `parseYAMLEntry` (add a `Set map[string]bool` to `importEntry`) and build a `map[string]any` body in `entryToUpdateRequest`:

```go
func entryToUpdateRequest(e importEntry) map[string]any {
	body := map[string]any{"host": e.Host} // host is always required-by-parse
	if e.Set["aliases"] {
		body["aliases"] = e.Aliases
	}
	if e.Set["port"] {
		body["port"] = e.Port
	}
	if e.Set["secret_ref"] {
		body["secret_ref"] = e.SecretRef
	}
	if e.Set["auth_model"] {
		body["auth_model"] = e.AuthModel
	}
	if e.Set["vpn_required"] {
		body["vpn_required"] = e.VPNRequired
	}
	if e.Set["extras"] {
		body["extras"] = e.Extras
	}
	if e.Set["notes"] {
		body["notes"] = e.Notes
	}
	return body
}
```

Then change `client.UpdateTarget` to accept `any` (or `map[string]any`) as the body. (`TargetUpdateRequest` can be deleted, or kept and used by callers that want the full PUT-style semantics — but the importer should not be one of those callers.)

*Option B (keep typed surface):* Rebuild `TargetUpdateRequest` with `*string`, `*bool`, `*[]string`, `*map[string]any` for every field plus `omitempty`. Default the pointer to nil unless the YAML set the field. More work, more invasive.

Then add a test that asserts the PATCH body shape (or the resulting target state):

```go
// Capture the PATCH body in fakeImportServer and assert it does NOT
// contain auth_model / vpn_required / extras / aliases when the YAML
// didn't specify them.
```

and mirror that on the Python side: insert a target with rich existing state, PATCH only `notes`, assert `auth_model` / `vpn_required` / `extras` / `aliases` survive.

**Source:** Pydantic v2 serialization docs — https://docs.pydantic.dev/latest/concepts/serialization/#modelmodel_dump (`exclude_unset` excludes fields that were not explicitly provided; an explicit `None` counts as provided). Verified locally against the project's pinned Pydantic version: `model_fields_set` contains all 8 keys when the import body is parsed, and `model_dump(exclude_unset=True)` returns the full dict.

cli/internal/api/targets.go
```diff
@@ -63,6 +65,34 @@ type ListTargetsParams struct {
...
+type TargetUpdateRequest struct {
+	Aliases     []string       `json:"aliases"`
+	Host        string         `json:"host"`
+	Port        *int           `json:"port"`
+	SecretRef   *string        `json:"secret_ref"`
+	AuthModel   string         `json:"auth_model"`
+	VPNRequired bool           `json:"vpn_required"`
+	Extras      map[string]any `json:"extras"`
+	Notes       *string        `json:"notes"`
+}
```
damir-topic: **🛑 Blocker** — doc comment misstates server semantics; struct needs to change shape

**What:** The doc comment claims "Nil pointer fields serialize as JSON null, which instructs the server to clear that column (Pydantic exclude_unset behaviour)." Two things are wrong: (a) Pydantic v2's `exclude_unset=True` excludes fields **not provided in the input**, not fields whose value is `None` — an explicit JSON `null` ends up in `model_fields_set` and IS applied via `setattr`. (b) The fields without pointer types (`Aliases`, `Host`, `AuthModel`, `VPNRequired`, `Extras`) cannot signal absence at all — they always serialize to a real value (`[]`, `""`, `""`, `false`, `{}`).

**Why it matters:** Anchors the PATCH clobber bug detailed in the inline comment on `cli/internal/cmd/targets/import.go` `entryToUpdateRequest`. Any future caller that copies this struct's shape ("hey, this is the typed PATCH body") will silently inherit the same destructive PUT semantics.

**Required change:** Tied to the `entryToUpdateRequest` fix. If you pick Option A (drop the typed struct in favour of `map[string]any` for the PATCH path), delete `TargetUpdateRequest` and update `UpdateTarget`'s signature:

```go
func (c *AuthedClient) UpdateTarget(ctx context.Context, name string, body map[string]any) (*Target, int, error)
```

If you pick Option B (keep typed surface), change every field to a pointer/`omitempty` and rewrite the doc comment to match:

```go
// TargetUpdateRequest is the PATCH /api/v1/targets/{name} body.
// Fields are pointer-typed so a nil value omits the key from the
// JSON body (omitempty). The server applies only the keys that
// appear (Pydantic exclude_unset). To explicitly clear a nullable
// column, set the pointer to a typed null sentinel — currently
// unused; document if added.
type TargetUpdateRequest struct {
    Aliases     *[]string       `json:"aliases,omitempty"`
    Host        *string         `json:"host,omitempty"`
    Port        *int            `json:"port,omitempty"`
    SecretRef   *string         `json:"secret_ref,omitempty"`
    AuthModel   *string         `json:"auth_model,omitempty"`
    VPNRequired *bool           `json:"vpn_required,omitempty"`
    Extras      *map[string]any `json:"extras,omitempty"`
    Notes       *string         `json:"notes,omitempty"`
}
```

Either way, the existing doc comment about "nil pointer fields serialize as null to clear the column" must go — it describes behaviour the server does not implement.

**Source:** Pydantic v2 docs — https://docs.pydantic.dev/latest/concepts/serialization/#modelmodel_dump; server route at `backend/src/meho_backplane/api/v1/targets.py:230-234` (`body.model_dump(exclude_unset=True)` then `setattr(t, k, v)` for every k in the dump).

cli/internal/cmd/targets/import.go
```diff
@@ -0,0 +1,411 @@
...
+				}
+			}
+		}
+	}
+
+	if raw["port"] != nil {
+		if p, ok := raw["port"].(int); ok {
+			e.Port = &p
+		}
+	}
```
damir-topic, line 215: **⚠️ Major** — `port` silently dropped if YAML decodes it to anything other than `int`

**What:** The type-assertion `raw["port"].(int)` only succeeds when yaml.v3 decoded the scalar into a Go `int`. yaml.v3 returns `int64` for integer scalars that don't fit in `int` (rare for port values), but more importantly returns `string` for a quoted value like `port: "443"` — a perfectly reasonable thing a hand-written YAML file might contain. In either case the `ok` branch is false, `e.Port` stays nil, and the import proceeds silently with no port for that target.

**Why it matters:** The Task acceptance criteria require "malformed YAML → non-zero with parse error" and "missing required field → non-zero before any API call." Port isn't required, but a present-but-untyped port is *user error in the YAML*, which the import should report — not swallow. Worse, when combined with the PATCH clobber bug above, importing a YAML with `port: "443"` against an existing target with `port: 6443` will (after that bug is fixed) leave the existing port alone (good) but on a fresh CREATE, the target is created with no port (bad — silently degraded).

**Required change:** Either widen the assertion to accept `int` / `int64` / `float64` (yaml.v3 is occasionally lenient) and report a parse error on any other type:

```go
if v, ok := raw["port"]; ok && v != nil {
	switch p := v.(type) {
	case int:
		e.Port = &p
	case int64:
		pp := int(p)
		e.Port = &pp
	default:
		return e, fmt.Errorf("entry %d (%s): port must be an integer, got %T", i, name, v)
	}
}
```

Add a test that exercises a quoted port (`port: "443"`) and asserts the parser returns an error.

**Source:** yaml.v3 scalar tag resolution — https://pkg.go.dev/gopkg.in/yaml.v3#Unmarshal. `!!int` tags decode to `int` when they fit and `int64` otherwise; `!!str` tags (anything quoted, or anything that doesn't match the integer regex) stays a string.

backend/tests/_targets_helpers.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def _admin_token(key: Any, tenant_id: str = DEFAULT_TENANT_ID) -> str:
+    return mint_token(
+        key, sub="admin-1", tenant_role=TenantRole.TENANT_ADMIN.value, tenant_id=tenant_id
+    )
+
+
+def _operator_token(key: Any, tenant_id: str = DEFAULT_TENANT_ID) -> str:
+    return mint_token(
```
damir-topic: **⚠️ Major** — `ruff format --check` fails on this file (and on `test_api_v1_targets_import.py`)

**What:** Running the same `uv run ruff format --check src/ tests/` that `.github/workflows/ci.yml`'s `python-lint-test` job runs flags two files for reformatting:

```
Would reformat: tests/_targets_helpers.py
Would reformat: tests/test_api_v1_targets_import.py
2 files would be reformatted, 114 files already formatted
```

The diff is mechanical: this line (`_operator_token` call) is split across three lines but fits on one under `line-length = 100`; several inline dict literals in `test_api_v1_targets_import.py` exceed 100 chars and ruff wants them in vertical form.

**Why it matters:** CI's format gate would block this merge if it had run. It didn't run **because this PR's base is `feat/256-targets-cli`, not `main`** (`.github/workflows/ci.yml` is gated on `pull_request: branches: [main]`). When the stacked T5 PR lands and this one is rebased onto `main` — or when a future PR cherry-picks these helpers — the gate will fire and fail.

**Required change:** Run `uv run ruff format src/ tests/` and commit the result (or apply the diff in `ruff format --check --diff`'s output). Two files change; the diff is whitespace-only.

**Source:** `.github/workflows/ci.yml` python-lint-test job, step "Format gate — non-zero exit if any file would be reformatted".

damir-topic (changes requested):
## /review-pr-slim summary

**Anchor task:** #257 — G0.3-T6 `meho targets import <file>`
**Surface area:** backend (test helpers only) + CLI (Go)
**Best-practices packs audited:** python (test helpers), devops (idempotency / dry-run)
**Verdict:** REQUEST_CHANGES

### Acceptance criteria
- [x] Default mode aborts with conflict list when names exist — `TestImport_Conflict_AbortsByDefault`.
- [❌] `--update` PATCHes existing targets without clobbering — **see inline blocker on `entryToUpdateRequest`**. The current PATCH body always includes every column; combined with Pydantic v2's `model_fields_set` semantics (a field present-as-null in JSON is "set"), the server's `setattr` loop overwrites `auth_model`, `vpn_required`, `extras`, `aliases`, `port`, `secret_ref`, `notes` on every update regardless of whether the YAML provided them. The load-bearing feature of this PR does the opposite of what the task and the struct's own comment claim.
- [x] `--dry-run` prints plan, no API calls — `TestImport_DryRun_NoWrites`.
- [x] `--dry-run --json` emits `{create, update, skip}` — `TestImport_DryRun_JSON`.
- [partial] Round-trip from import → describe — verified at the API level (`test_import_round_trip_fields_match`); end-to-end CLI round-trip with the actual import path is not tested.
- [x] Malformed YAML → non-zero with parse error.
- [x] Missing required field → non-zero before any API call.
- [x] Unknown YAML field → spilled into extras.
- [❌] **CI integration test against the consumer's real `targets.yaml`** (Task #257 explicit acceptance line). No test fetches `evoila-bosnia/claude-rdc-hetzner-dc/rdc-hetzner-dc/targets.yaml` (or a pinned snapshot) and feeds it through the import path.
- [partial] Go unit + Python integration tests — Python tests pass locally (13 new + 609 total). Go tests not verifiable locally (no Go toolchain); central CI workflow did not run for this PR (see "CI gates" below).

### Project-board alignment
|---|---|---|---|---|---|
| #257 | In Progress | `Closes #257` in PR body | ❌ will **not** fire | partial (PATCH clobber) | rebase onto `main` before merge or close manually; resolve blockers first |
| #224 (Initiative) | In Progress | parent — `Parent Initiative: #224` | n/a | not yet (T6 is the final task; partial) | once T6 lands + the rest, mark Initiative done |

**`Closes #257` will not auto-close on merge.** The PR base is `feat/256-targets-cli`, not `main`. GitHub's auto-close keyword only fires on merge into the default branch. Either (a) rebase onto `main` once the T5 stack lands, or (b) close `#257` manually after merge with a completion comment.

### CI gates didn't run on this PR
`.github/workflows/ci.yml` is gated on `pull_request: branches: [main]`. Because this PR targets `feat/256-targets-cli`, the central Python + Go + Helm CI job — including `ruff format --check`, `mypy`, `pytest`, `golangci-lint`, `go test` — **never ran**. The only checks that did run are DCO, SonarCloud, CodeRabbit. SonarCloud's "Quality Gate Passed" is not a stand-in for the language-toolchain gates. Two of the new files fail `ruff format --check` locally (inline finding below); CI would have caught this had it run.

### Status checks
- ❌ **DCO: ACTION_REQUIRED** — commits carry `Co-Authored-By: Claude` trailers but no `Signed-off-by:`. DCO requires the latter. Author needs to amend each commit with `--signoff` (or re-author with sign-off) before merge.
- ✅ SonarCloud Code Analysis: Quality Gate Passed, but **18 new issues** including 3 cognitive-complexity criticals — addressed in inline findings.
- ⚠️ CodeRabbit: skipped because base ≠ `main`.

### Predicted vs. actual diff
- ✅ New `meho targets import` cobra subcommand wired into `targets.NewCommand()`; docs/codebase/cli.md updated.
- ✅ `CreateTarget` / `UpdateTarget` AuthedClient methods + 401-retry helper.
- ✅ `ListTargets` now sends `limit` query param (was declared-unused — confirmed in diff).
- ✅ Shared test helpers extracted to `_targets_helpers.py` (3 commits — extract, then consolidate `_insert_target`).
- ⚠️ **PATCH body always sends every column** — see inline `entryToUpdateRequest` blocker.
- ⚠️ **port silently dropped if YAML type is not `int`** — see inline `parseYAMLEntry` finding.
- 🕳️ **No integration test wired to the real consumer `targets.yaml`** — Task acceptance line.

### Integration verification (run inside `.claude/worktrees/pr-362-review/`)
- [❌] `ruff format --check src/ tests/` — **2 files would be reformatted**: `tests/_targets_helpers.py`, `tests/test_api_v1_targets_import.py`. CI gate (had it run) would have failed.
- [x] `mypy src/` (strict) — clean, 58 source files.
- [x] `pytest tests/test_api_v1_targets_import.py -q` — 13 passed.
- [x] `pytest tests/ -q --ignore=tests/test_observability.py --ignore=tests/integration/` — 609 passed, 1 skipped.
- [⚠️ not run] `golangci-lint`, `go build`, `go test ./cli/...` — no Go toolchain locally; central CI did not run (see "CI gates" above).
- [⚠️ not run] End-to-end CLI run against a live backplane (`meho targets import …`) — no live backplane.

- 🛑 2 blockers — see inline (PATCH clobber + missing struct tags)
- ⚠️ 2 majors — see inline (ruff format gate + port silent-drop)

- SonarCloud flags cognitive complexity > 15 on three functions: `newImportCmd.RunE` (26), `parseYAMLEntry` (27), `executeImportPlan` (17). Splitting the RunE auth/list/plan/execute pipeline into named helpers and the YAML parse into per-field setters would tidy this up — a follow-up refactor PR rather than churn on this one.
- SonarCloud's "duplicate literal" findings on `cli/internal/api/targets.go` (`"/api/v1/targets/"` × 3, `"application/json"` × 3) — extracting a `const targetsPath` and using `http.Header.Set("Content-Type", contentTypeJSON)` would clear them, again better as a small follow-up.
- `executeImportPlan` continues importing after individual failures (accumulating errMsgs), reports a non-zero exit at the end. The behaviour is reasonable but the "partial import on failure" semantics aren't documented in the long help or tested.
- `listAllTargets` paginates with `limit=500` and keyset by name — fine for the 25-entry consumer file; worth revisiting if a tenant ever has thousands of targets. The CLI does a full enumeration even when YAML has 1 entry.

### Docs
- [x] `docs/codebase/cli.md` — module layout updated to list the new `import.go` file and its tests.
- [⚠️] `docs/cross-repo/targets-yaml.md` — Initiative #224 §7 calls for a "migration recipe + dual-read overlap" section. This PR doesn't extend that doc. The Task #257 References section links it but the Acceptance Criteria don't require the doc update. The Initiative does. Either add the recipe in this PR or file a follow-up.

### What worked well (review-body only — not posted as inline comments)
- The shared `_targets_helpers.py` refactor is clean — pulls the autouse fixtures, mint-token helpers, and `_insert_target` out of three duplicating test modules into one source of truth.
- `parseTargetsYAML` test coverage is genuinely thorough for the parser (happy path, defaults, missing required fields, malformed YAML, unknown fields → extras, empty aliases, empty file) — the parser-level testing is in good shape.
- The dry-run + `--json` plan output shape (`{create, update, skip}`) is exactly the machine-readable contract the Task asked for.
- Tenant scoping is handled the right way: import uses the operator's JWT-bound tenant, no body override; `_insert_target` cross-tenant test confirms isolation.

PR #447, opened 2026-05-14
ci(security-scan): authenticate Docker Hub pulls via container.credentials (G2.9-T0)

damir-topic:
## Summary

- Add a `check-secrets` gate job + conditional `container:` construction to `.github/workflows/security-scan.yml` so the Semgrep job runs an authenticated Docker Hub pull when `DOCKERHUB_USERNAME`/`DOCKERHUB_TOKEN` org secrets exist, and degrades to an anonymous pull when they don't — without failing GH Actions template validation in the unset state. Authenticated pull doubles the runner pool's shared 6h pull bucket from 100 to 200, unsticking the recurring `unauthenticated pull rate limit` failure on `meho-runners`.
- Tactical fix only — `G2.9-T1..T3` (initiative #67) build the permanent in-cluster pull-through cache that retires this block.

## Implementation notes

The naive `credentials: { username: ${{ secrets.DOCKERHUB_USERNAME }}, password: ${{ secrets.DOCKERHUB_TOKEN }} }` block — what the issue body proposed — hard-fails GH Actions template validation at parse time when either secret is unset (the bootstrap state):

```
The template is not valid. .github/workflows/security-scan.yml (Line: 62, Col: 19): Unexpected value '',
.github/workflows/security-scan.yml (Line: 63, Col: 19): Unexpected value ''
```

This fires during `Set up job`, **before** any image pull is attempted. It is not an "auth failure during pull" — the workflow never reaches the pull. (Empirical confirmation: run [25864725292](https://github.com/evoila/meho/actions/runs/25864725292/job/76003627858).) An earlier iteration of the PR body claimed an auth-failure fallback mode; that was wrong and is corrected here.

CodeRabbit's suggested top-level `container: ${{ ... secrets.X ... }}` fromJSON ternary doesn't work either: per GitHub's documented [context-availability table](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability), `jobs.<job_id>.container` only allows `github, inputs, matrix, needs, strategy, vars` — `secrets` is excluded. `actionlint` correctly rejects the reference.

The fix is the documented workaround: a small `check-secrets` gate job runs first on `ubuntu-latest`, where `secrets` IS available inside a step's `run:` block. It computes a JSON container spec with `jq` — with `credentials` when both secrets are non-empty, without when either is unset — and exposes it on a job output. The `semgrep` job consumes it via `needs.check-secrets.outputs.container_spec` + `fromJSON()`, both of which ARE allowed at top-level `container:`.

```yaml
jobs:
  check-secrets:
    runs-on: ubuntu-latest
    outputs:
      container_spec: ${{ steps.compute.outputs.container_spec }}
    steps:
      - id: compute
        env:
          DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
          DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
        run: |
          if [ -n "$DOCKERHUB_USERNAME" ] && [ -n "$DOCKERHUB_TOKEN" ]; then
            spec=$(jq -c -n --arg user "$DOCKERHUB_USERNAME" --arg pass "$DOCKERHUB_TOKEN" \
              '{image: "semgrep/semgrep:1.160.0", credentials: {username: $user, password: $pass}}')
          else
            spec='{"image":"semgrep/semgrep:1.160.0"}'
          fi
          echo "container_spec=$spec" >> "$GITHUB_OUTPUT"

  semgrep:
    needs: check-secrets
    runs-on: meho-runners
    container: ${{ fromJSON(needs.check-secrets.outputs.container_spec) }}
    # ...
```

`credentials:` is still the right knob for the consumer (not `docker/login-action`): GHA pulls `container.image` *before* any step runs, so a login step inside the job cannot affect that pull. The gate job is a small price for the parse-time-validity property.

Docs: https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container, https://docs.github.com/en/actions/learn-github-actions/contexts.

## Deviation from the issue body — chart.yml is NOT in scope

Issue #68 lists `.github/workflows/chart.yml`'s `services:` block for `pgvector/pgvector:pg16` as a second target. **There is no such block in chart.yml.** The pgvector image is pulled inside a kind cluster spun up by the `helm-test` job, which:

1. Runs on `ubuntu-latest` (GitHub-hosted), NOT `meho-runners`, so it does not share the ARC cluster's NAT egress and does not consume the 100/6h ARC-shared bucket that triggered this fire.
2. Pulls the image via the kind cluster's node, not via a GHA `services:`/`container:` block. `container.credentials:` does not apply to images pulled inside a kind cluster.

Acceptance criterion #4 ("Both workflows reference `credentials:` under their respective `container:`/`services:` blocks") therefore cannot be literally satisfied for chart.yml without restructuring the workflow (e.g., introducing a real `services:` container in place of the in-cluster pod). That restructuring is a larger change with its own design questions and is **out of scope for this tactical unstick**. The actual failure mode the initiative is solving (`semgrep` in `security-scan.yml`) is fixed by this PR alone.

If the chart.yml pgvector pull ever does need authenticated access, the right fix is the permanent in-cluster cache from G2.9-T2/T3 (it covers any Docker Hub pull transparently, including pulls originating inside kind on a GitHub-hosted runner once those runners point at the cache — though the GH-hosted runner case is moot for our ARC-rate-limit concern).

## Org-level secrets — manual ops precondition

This PR's YAML edits expect `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` to exist as **org-level GitHub Secrets** on the `evoila` org, accessible to `evoila/meho`. I was unable to verify their presence from this session (`gh secret list --org evoila` → HTTP 403, requires org-admin permission).

Required before this PR's runs benefit from authenticated pulls:

1. Create a Docker Hub access token (NOT the account password) scoped `Public Repo Read-only` per https://docs.docker.com/security/for-developers/access-tokens/.
2. Store username + token in 1Password using `op-cli` skill conventions; record the item name in a follow-up comment on this PR (acceptance criterion #2).
3. Mirror them to `evoila` org secrets:
   ```
   gh secret set DOCKERHUB_USERNAME --org evoila --visibility all
   gh secret set DOCKERHUB_TOKEN    --org evoila --visibility all
   ```
4. Verify with `gh secret list --org evoila` (acceptance criterion #1).

Until the secrets are provisioned, `check-secrets` outputs the anonymous-pull spec and `semgrep` continues to share the 100/6h anonymous bucket. After provisioning, the next workflow run sees the populated secrets and the spec carries `credentials:`, doubling the bucket to 200/6h.

## Test plan

- [x] `actionlint .github/workflows/security-scan.yml` — exit 0
- [x] `python3 -c "import yaml; yaml.safe_load(...)"` — YAML parses; both `check-secrets` and `semgrep` jobs declared; `semgrep.container` is a `${{ fromJSON(...) }}` expression.
- [x] Both fromJSON shapes (with and without `credentials`) parse as valid GH Actions container maps via local `python -c "import json; ..."` sanity check.
- [x] Diff is targeted: one workflow file, one new gate job, one consumer-side `container:` expression swap.
- [ ] (Post-merge) Verify the next run of Security Scan reaches the `semgrep` job successfully — i.e. `Set up job` no longer hard-fails template validation regardless of secret state (acceptance criterion #5).
- [ ] (Post-merge, ops-gated) Once secrets are provisioned, re-run Security Scan and verify no `unauthenticated pull rate limit` line appears in the Semgrep job logs (acceptance criterion #5 second-order).
- [ ] (Post-merge) Re-run every workflow on `main` once to confirm no regression (acceptance criterion #6).

## Out of scope

- `.github/workflows/chart.yml` — no `services:`/`container:` block to attach `credentials:` to (see deviation note above).
- G2.9-T1 (#69): Docker Hub service account → 1Password → K8s Secret.
- G2.9-T2 (#70): registry:2 pull-through cache deployment.
- G2.9-T3 (#71): ARC `registry-mirrors` injection + smoke test.

Closes #68

---

## Follow-up (auto-implement-issue, iteration 2, 2026-05-14)

Addressed review findings from `.claude/auto/review-results/pr-447-iter-1.json`:

- **B1** `a2af1e6` — Replaced direct `container.credentials: { username: secrets.X }` with a `check-secrets` gate job that publishes the JSON container spec on an output; consumer `semgrep` job reads it via `needs.check-secrets.outputs.container_spec` + `fromJSON()`. Both parse-time-validates when secrets are unset AND carries credentials when secrets are set. CodeRabbit's suggested top-level `${{ ... secrets.X ... }}` ternary was rejected by `actionlint` because `secrets` is not allowed at `jobs.<job_id>.container` per [GH context-availability docs](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability); the gate-job pattern is the documented workaround.

Adjacent improvement folded in (per review's `adjacent_improvements`):

- PR body updated to reflect the empirically-observed failure mode (parse-time template validation), correcting the earlier "auth failure during pull" claim that conflated parse-time and pull-time failure modes.

## Summary by CodeRabbit

* **Chores**
  * CI security-scan workflow now supports authenticated container image pulls when credentials are available, falling back to anonymous pulls otherwise. The workflow dynamically selects the image specification while preserving existing scan and SARIF upload steps for consistent reporting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/447)

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -45,6 +45,22 @@ jobs:
     container:
       # Bump cadence: monthly review or on advisory.
       image: semgrep/semgrep:1.160.0
+      # Authenticated Docker Hub pull. The ARC scale-set runs every job
+      # as an ephemeral pod sharing one cluster-egress NAT IP, which
+      # exhausts Docker Hub's anonymous 100-pulls-per-6h limit fast and
+      # has tripped this Semgrep job in production. Authenticated
+      # Personal-tier doubles the bucket to 200/6h
+      # (https://docs.docker.com/docker-hub/usage/pulls/) — enough
+      # short-term while G2.9-T1..T3 stand up the in-cluster
+      # pull-through cache that retires this block.
+      #
+      # `credentials:` is consumed by the runner BEFORE any step runs,
+      # so `docker/login-action` cannot substitute. Secrets are
+      # provisioned at the `evoila` org level so all sibling repos
+      # inherit them without per-repo duplication.
+      credentials:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
 
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -38,29 +38,72 @@ permissions:
   security-events: write  # required for SARIF upload to Code Scanning
 
 jobs:
+  # Gate job: reads the optional Docker Hub secrets and surfaces their
+  # presence as a JSON-encoded container spec on a job output. This
+  # indirection is required because the `secrets` context is NOT
+  # available at `jobs.<job_id>.container` (top-level) per
+  # https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability
+  # — only `github, inputs, matrix, needs, strategy, vars` are allowed
+  # there. `needs.<job>.outputs.*` IS available, so the gate publishes
+  # the full container map (with or without `credentials:`) and the
+  # consumer job parses it via `fromJSON()`.
+  #
+  # A previous iteration attempted `container: ${{ ... secrets.X ... }}`
+  # directly; actionlint correctly rejected the `secrets` references at
+  # that scope, and the bare `credentials: { username: ${{ secrets.X }} }`
+  # before that failed GH Actions template validation at parse time
+  # with "Unexpected value ''" (run 25864725292, 'Set up job'). The
+  # gate-job pattern is the documented workaround.
+  check-secrets:
+    name: Check Docker Hub credentials
+    runs-on: ubuntu-latest
+    timeout-minutes: 1
+    outputs:
+      container_spec: ${{ steps.compute.outputs.container_spec }}
+    steps:
+      - id: compute
+        # Build the container spec as a JSON string. When both secrets
+        # are set, include `credentials`; otherwise emit the bare image
+        # spec so the downstream pull is anonymous. The JSON is emitted
+        # via the secret-aware shell (where `secrets` IS available) and
+        # passed to the consumer job through `outputs:` (allowed in
+        # `needs` context).
+        env:
+          DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+          DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+        run: |
+          if [ -n "$DOCKERHUB_USERNAME" ] && [ -n "$DOCKERHUB_TOKEN" ]; then
+            spec=$(jq -c -n \
+              --arg user "$DOCKERHUB_USERNAME" \
+              --arg pass "$DOCKERHUB_TOKEN" \
+              '{image: "semgrep/semgrep:1.160.0", credentials: {username: $user, password: $pass}}')
+          else
+            spec='{"image":"semgrep/semgrep:1.160.0"}'
+          fi
+          echo "container_spec=$spec" >> "$GITHUB_OUTPUT"
+
   semgrep:
     name: Semgrep SAST
+    needs: check-secrets
     runs-on: meho-runners
     timeout-minutes: 15
-    container:
-      # Bump cadence: monthly review or on advisory.
-      image: semgrep/semgrep:1.160.0
-      # Authenticated Docker Hub pull. The ARC scale-set runs every job
-      # as an ephemeral pod sharing one cluster-egress NAT IP, which
-      # exhausts Docker Hub's anonymous 100-pulls-per-6h limit fast and
-      # has tripped this Semgrep job in production. Authenticated
-      # Personal-tier doubles the bucket to 200/6h
-      # (https://docs.docker.com/docker-hub/usage/pulls/) — enough
-      # short-term while G2.9-T1..T3 stand up the in-cluster
-      # pull-through cache that retires this block.
-      #
-      # `credentials:` is consumed by the runner BEFORE any step runs,
-      # so `docker/login-action` cannot substitute. Secrets are
-      # provisioned at the `evoila` org level so all sibling repos
-      # inherit them without per-repo duplication.
-      credentials:
-        username: ${{ secrets.DOCKERHUB_USERNAME }}
-        password: ${{ secrets.DOCKERHUB_TOKEN }}
+    # Authenticated Docker Hub pull. The ARC scale-set runs every job
+    # as an ephemeral pod sharing one cluster-egress NAT IP, which
+    # exhausts Docker Hub's anonymous 100-pulls-per-6h limit fast and
+    # has tripped this Semgrep job in production. Authenticated
+    # Personal-tier doubles the bucket to 200/6h
+    # (https://docs.docker.com/docker-hub/usage/pulls/) — enough
+    # short-term while G2.9-T1..T3 stand up the in-cluster
+    # pull-through cache that retires this block.
+    #
+    # `credentials:` is consumed by the runner BEFORE any step runs,
+    # so `docker/login-action` cannot substitute. Secrets are
+    # provisioned at the `evoila` org level so all sibling repos
+    # inherit them without per-repo duplication.
+    #
+    # `container` is materialised from `check-secrets`' output (see the
+    # comment on that job). Bump cadence: monthly review or on advisory.
+    container: ${{ fromJSON(needs.check-secrets.outputs.container_spec) }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
 


damir-topic:
@coderabbitai review

.github/workflows/security-scan.yml
```diff
@@ -45,6 +45,22 @@ jobs:
...
+      # short-term while G2.9-T1..T3 stand up the in-cluster
+      # pull-through cache that retires this block.
+      #
+      # `credentials:` is consumed by the runner BEFORE any step runs,
+      # so `docker/login-action` cannot substitute. Secrets are
+      # provisioned at the `evoila` org level so all sibling repos
+      # inherit them without per-repo duplication.
+      credentials:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
```
damir-topic: **[🛑 B1]** — Empty-secret expressions cause GH Actions template-validation hard-fail; degrade gracefully

**What:** When `secrets.DOCKERHUB_USERNAME`/`secrets.DOCKERHUB_TOKEN` are not yet provisioned at the org level, GitHub Actions evaluates each `${{ secrets.X }}` expression to an empty string. Empty-string values under `container.credentials.{username,password}` are rejected by the GitHub Actions template parser at workflow-load time with `Unexpected value ''` — BEFORE any step runs and BEFORE the image pull is attempted.

This is not theoretical: it is the current failure mode on this PR's own head SHA. Run `25864725292`, Semgrep SAST job → "Set up job" step:

```
##[error]The template is not valid. .github/workflows/security-scan.yml (Line: 62, Col: 19): Unexpected value '',.github/workflows/security-scan.yml (Line: 63, Col: 19): Unexpected value ''
```

**Why it matters:**
- The PR body documents the expected post-merge state as `Error: Container image '...' could not be pulled (auth failure)`. The observed state is stricter: the workflow never even starts the container init, so no graceful retry, no useful diagnostic in the pull step, no observability into what's actually wrong. Empirically a regression from main's previous "fails at retry-with-rate-limit-error" mode, which at least reached the pull step.
- Acceptance criterion #6 ("No other workflow regressed") is currently FALSE at this SHA — the Semgrep workflow is regressed by this exact diff in a more severe way than before.
- If the secrets ever get unset later (rotation timing, key expiry, misconfig), every Semgrep run hard-fails at parse time with no fallback. The same fragility applies to any sibling repo that inherits these org-level secrets and finds them stale.
- CodeRabbit independently flagged this with a 🟠 Major severity at the same lines.

**Required change:** Make the `credentials:` block conditional on the secrets being non-empty so the workflow gracefully falls back to anonymous pull when secrets aren't provisioned (and gets the authenticated bucket once they are). The CodeRabbit-recommended `fromJSON` pattern is one valid shape:

```yaml
container: >-
  ${{ (secrets.DOCKERHUB_USERNAME != '' && secrets.DOCKERHUB_TOKEN != '')
      && fromJSON(format('{{"image":"semgrep/semgrep:1.160.0","credentials":{{"username":"{0}","password":"{1}"}}}}', secrets.DOCKERHUB_USERNAME, secrets.DOCKERHUB_TOKEN))
      || fromJSON('{"image":"semgrep/semgrep:1.160.0"}') }}
```

This evaluates to a `container` map with `credentials:` present only when both secrets are non-empty; otherwise it evaluates to a `container` map with just `image:` (which is the same parse-valid shape as `main` today). Anonymous pull then resumes — still subject to the rate-limit but at least the workflow validates and reaches the pull step.

An equivalent alternative: keep the `credentials:` block inline as today but guard the entire job behind a step-level check that fails fast with a clear error message before the container init runs (less elegant but more readable).

Whichever shape: the workflow must validate at parse-time when secrets are unset, even if the eventual pull then fails on rate-limit.

**Source:** Confirmed empirically by GitHub Actions runner output on this PR's failed run `25864725292`. GitHub's expression behavior on missing secrets documented at https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions (secrets evaluate to empty string when unset). `fromJSON` + ternary-style `&&`/`||` pattern documented at https://docs.github.com/en/actions/reference/workflows-and-actions/expressions.

damir-topic:




**Anchor task:** evoila-bosnia/meho-internal#68 — G2.9-T0 tactical Docker Hub `container.credentials:` unstick
**Surface area:** devops (CI/CD workflow)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `0dd806b466dff9e20c10a0ff2888f179217fb249`
**AI reviewer (CodeRabbit):** received (~3min after PR open) — 1 finding, 🟠 Major, aligned with this review

### Acceptance criteria

- [ ] AC#1 — `DOCKERHUB_USERNAME`/`DOCKERHUB_TOKEN` org secrets exist. **Not in scope of this PR** (manual ops). Documented as precondition. Acceptable scope split.
- [ ] AC#2 — Credentials in 1Password under documented item name. **Not in scope of this PR** (manual ops). Acceptable.
- [ ] AC#3 — Token is Personal Access Token, `Public Repo Read-only`. **Not in scope of this PR** (manual ops). Acceptable.
- [x] AC#4 (partial) — `security-scan.yml` references `credentials:` under `container:`. ✅ Verified at L61-63. `chart.yml` clause: **the issue body's premise is wrong** (verified by reading `chart.yml` end-to-end — there is no `services:`/`container:` block for pgvector; the image is pulled inside a kind cluster on `ubuntu-latest`, not via GH Actions container syntax, and not on `meho-runners`). The implementer's deviation note is factually correct. Partial AC is reasonable.
- [ ] AC#5 — Re-run of Semgrep on main completes without rate-limit error. **Currently FAIL at the PR's own head SHA** for a different reason (template-parse error — see B1).
- [ ] AC#6 — No other workflow regressed. **Currently FAIL — the Semgrep workflow itself is regressed at this SHA** (parse-time template validation rejects empty-string secrets — see B1).

### Project-board alignment

|---|---|---|---|---|---|
| evoila-bosnia/meho-internal#68 | OPEN | `Closes #68` in body | manual (cross-repo) | partial — chart.yml deviation justified, secrets pending | Will require manual close on merge (cross-repo `Closes` does not auto-link) |

GitHub's `Closes #N` only auto-links within the same repo. `closingIssuesReferences` is empty for this PR. Manual issue-close after merge required. **Not a blocker** — known platform constraint.

### Predicted vs. actual diff

- ✅ `.github/workflows/security-scan.yml` — `container.credentials:` added with `DOCKERHUB_USERNAME`/`DOCKERHUB_TOKEN` (16 additions, expected).
- 🕳️ Missing: A defensive guard for the case where the org secrets are not yet provisioned. The implementer's PR body anticipates this state ("Required before this PR's runs go green") and the issue body lists secret provisioning as AC#1, but at this head SHA the workflow hard-fails at template-validation time before any step runs. CodeRabbit independently flagged this. See B1.

No undisclosed scope changes. Diff is minimally invasive. Only `security-scan.yml` touched (matches the implementer's "out of scope" list for `chart.yml`).

### Integration verification

- [x] `actionlint .github/workflows/security-scan.yml` — clean (static check passes, but static check does NOT model GH's runtime template-parse rejection of empty-string expressions).
- [x] `python3 -c "import yaml; yaml.safe_load(...)"` — YAML parses; `container.credentials.{username,password}` keys present.
- [x] Read full `chart.yml` end-to-end to verify implementer's chart.yml deviation claim — **confirmed correct** (no `services:`/`container:` block for pgvector; image pulled via `docker pull` + `kind load docker-image` on `ubuntu-latest`, outside GH Actions container syntax).
- [x] GitHub Actions empirical check via the PR's own failing CI run `25864725292` — **the workflow fails at template validation** with `Unexpected value ''` at Lines 62, 63 because `secrets.DOCKERHUB_USERNAME`/`secrets.DOCKERHUB_TOKEN` resolve to empty strings (org secrets not yet provisioned). This is a stricter failure than the previous rate-limit failure on main, which at least let the runner attempt the pull.
- [⚠️ not run] Post-merge Semgrep re-run with provisioned secrets — gated on AC#1/#2/#3 manual ops; cannot verify from this session.

For `/auto-implement-issue --continue-pr`'s punch list:

- 🛑 B1 — Empty-secret expressions in `container.credentials:` cause GH Actions template-validation hard-fail; degrade gracefully. **Required change:** make the `credentials:` keys conditional on the secrets being non-empty (CodeRabbit's `fromJSON` pattern, or an equivalent shape). See inline.

- The PR body's "Without these secrets, the Semgrep job will still fail — but with `Error: Container image '...' could not be pulled` (auth failure)" is empirically inaccurate at this SHA (the actual failure is template-validation at parse-time, before any pull is attempted). Updating the PR body to reflect the observed failure mode would help reviewers of future similar PRs.
- Once G2.9-T2/T3 land and the in-cluster pull-through cache is live for ~1 week, the `credentials:` block can be removed (issue #67 Initiative body documents this as a cleanup follow-up).
- A documented runbook for rotating `DOCKERHUB_TOKEN` (referenced in issue #67 DoD) is out of scope here but should be filed as a follow-up against T1 (#69) deliverable.

### CodeRabbit cross-check

- Adopted: 1 — `🟠 Major` finding on lines 62-63 (empty secret resolution breaks template validation). Same root cause as B1; CodeRabbit's recommended fix (`fromJSON` conditional construction of the `container` object) is one valid shape.
- Disputed: 0
- Out of scope: 0

### Docs

- [n/a] `docs/codebase/devops.md` — verified-unchanged; the file describes the broader CI/CD landscape and the `credentials:` knob is a localized workflow detail. No doc gap.

- Sharp scope discipline. The deviation from issue #68's chart.yml clause is justified end-to-end by reading the actual workflow shape, and the PR body explains it clearly enough that future-me reading the trail will understand why AC#4 is partial.
- Diff is minimally invasive (16 additions in one file, zero deletions, zero adjacent logic touched).
- Commit message follows conventional commits, explains the runner-pre-step rationale for `credentials:` vs `docker/login-action`, names the parent initiative explicitly.
- Org-secrets-vs-repo-secrets choice is correct per devops best-practices pack §4 (least necessary scope; org-level shared across `evoila/meho` and future sibling repos).
- No `.planning/` references anywhere in PR body or commit message; clean.
- No secret values exposed in any echo/log statements; clean handling.

damir-topic:
@coderabbitai review

damir-topic:




**Anchor task:** evoila-bosnia/meho-internal#68 — G2.9-T0 tactical Docker Hub `container.credentials:` unstick
**Surface area:** devops (CI/CD workflows)
**Best-practices packs audited:** devops (§2 image/config/secret separation, §4 GitHub Actions / Security, §7 secrets & supply chain)
**Verdict:** APPROVE
**Head SHA reviewed:** `a2af1e6bbf9d84d5345dbeedc701c4d23fbcdacc`
**AI reviewer (CodeRabbit):** not received within 15min (triggered at 14:27:42Z; reply at 14:27:48Z acknowledged trigger, no review body landed within wait window)

### Acceptance criteria
- [x] AC#4 (`security-scan.yml` references `credentials:` under its `container:` block) — `container:` is now materialised via `fromJSON(needs.check-secrets.outputs.container_spec)`, and the gate-job computes the spec **with** `credentials: {username, password}` when both org secrets are non-empty. Verified by reading lines 57-83 (gate job) + 85-106 (consumer).
- [x] AC#4 deviation for `chart.yml` is justified — re-verified at this SHA. No `services:`/`container:` block exists for pgvector; image is pulled inside a kind cluster on `ubuntu-latest` (different egress, different bucket).
- [x] AC#5 (Set up job no longer hard-fails template validation) — empirically verified at SHA `a2af1e6`: workflow run [25865487081](https://github.com/evoila/meho/actions/runs/25865487081) shows `check-secrets`/Set up job: SUCCESS and `semgrep`/Set up job: SUCCESS. The template-validation hard-fail from iter-1's `0dd806b` is gone.
- [ ] AC#1/#2/#3 (org-level secrets provisioned, mirrored to 1Password, Docker Hub access token scoped Public Repo Read-only) — gated on manual ops; PR body documents the precondition. Cannot verify from this session.
- [ ] AC#5 second-order (no `unauthenticated pull rate limit` after secrets provisioned) — gated on AC#1/#2/#3.
- [ ] AC#6 (no other workflow regressed) — post-merge verification.

### Project-board alignment
|---|---|---|---|---|---|
| evoila-bosnia/meho-internal#68 | OPEN | `Closes` in body | manual (cross-repo) | full (scope: security-scan.yml; chart.yml deviation justified) | Manual close on merge |

Cross-repo `Closes #68` does not populate `closingIssuesReferences` — that's the platform constraint, not a PR-side defect. Initiative #67 status is unchanged; T0 (#68) is one of four tasks.

### Predicted vs. actual diff
- ✅ `.github/workflows/security-scan.yml` — expected and correct. The fix introduces a new gate job `check-secrets` (62 lines added) and rewrites the `semgrep` job's `container:` map to consume the gate's output via `fromJSON()`.
- ❓ Gate-job indirection — unexpected at first read (iter-1's blocker recommended a simple inline ternary) but **justified**. I empirically verified actionlint rejects CodeRabbit's suggested inline pattern: `context "secrets" is not allowed here. available contexts are "github", "inputs", "matrix", "needs", "strategy", "vars"` (per [GH context-availability docs](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability)). The gate-job is the documented workaround for the secrets-not-available-at-container-scope constraint.
- 🚫 None.
- 🕳️ None.

### Integration verification
- [x] `actionlint .github/workflows/security-scan.yml` → exit 0 (re-verified inside worktree at `a2af1e6`).
- [x] YAML parse via `python3 -c "yaml.safe_load(...)"` → both `check-secrets` and `semgrep` jobs declared; `semgrep.needs == 'check-secrets'`; `semgrep.container == '${{ fromJSON(needs.check-secrets.outputs.container_spec) }}'`.
- [x] **GH Actions runtime validation** (the only check that actually exercises the parse-time validator) → SUCCESS at `a2af1e6`. Workflow run [25865487081](https://github.com/evoila/meho/actions/runs/25865487081), both `check-secrets`/Set up job and `semgrep`/Set up job: SUCCESS. This is the regression test for B1.
- [x] **Adversarial actionlint check on CodeRabbit's suggested inline ternary** (run on a throwaway workflow under `/tmp/test-actionlint/`) → `exit 1` with two context-availability errors at the exact positions of `secrets.X` references. Confirms the fixer's claim that the inline pattern is rejected.
- [x] **Gate-job output: no secret-value leak** verified by inspecting [run 25865487081 check-secrets log](https://github.com/evoila/meho/actions/runs/25865487081/job/76006328977). The step's `env:` line displays `DOCKERHUB_USERNAME:` and `DOCKERHUB_TOKEN:` as empty (correct — secrets unset). When secrets ARE provisioned, GitHub Actions auto-masks them in step output by default; jq's `--arg` invocation is also the safe pattern (avoids shell-injection from the secret value).
- [x] `pytest` / `ruff` / `mypy` → n/a (workflow-only change).
- [ ] (Post-merge, ops-gated) Re-run with secrets provisioned → verify the spec contains `credentials:` and the `docker pull` succeeds with the 200/6h authenticated bucket. Cannot run from this session.

### Findings (inline)
None. iter-1's single blocker (B1) is **resolved at `a2af1e6`**. No new blockers, majors, or required-change findings introduced.

- After G2.9-T2/T3 lands (in-cluster pull-through cache) and the cache is green for ~1 week, the entire `check-secrets` gate + `container.credentials:` block becomes redundant. Initiative #67 documents this as a planned cleanup follow-up.
- The gate-job's `runs-on: ubuntu-latest` is deliberate (GitHub-hosted, separate IP/bucket — does not consume the ARC NAT-shared anonymous bucket). Worth a brief inline comment in the workflow at some later refresh if a future contributor wonders why the gate doesn't run on `meho-runners`.
- Runbook for rotating `DOCKERHUB_TOKEN` (Initiative #67 DoD requirement) should be filed as a follow-up against T1 (#69) deliverable — out of scope for this tactical PR.

- CodeRabbit re-review at the new head SHA did not land within the 15-minute wait window (triggered explicitly at 14:27:42Z by this reviewer). Iter-1 CodeRabbit review (at `0dd806b`) is now stale; its single finding (B1: inline `fromJSON` ternary) is functionally addressed even though the mechanism differs — the deviation is grounded in the actionlint context-availability constraint, which I verified empirically and which CodeRabbit's pattern would have tripped over had it landed.
- Adopted from iter-1 CodeRabbit: B1 is resolved (gate-job mechanism, not inline ternary).
- Disputed from iter-1 CodeRabbit: 0 — CodeRabbit's *intent* (degrade gracefully when secrets unset) is satisfied; only the *exact mechanism* differs, and the deviation is justified.
- Out of scope from iter-1 CodeRabbit: 0.

### Docs
- [x] `docs/codebase/devops.md` — verified unchanged; the file does not currently document the security-scan workflow's container-pull strategy. After G2.9-T2/T3 lands and the cache is the canonical pull path, that's the right time to add a paragraph there; not now.

- The fix correctly identified that CodeRabbit's suggested top-level inline pattern would not lint, **before** trying to apply it — pulling up the GH context-availability docs is the right reflex.
- The gate-job uses `jq --arg "$DOCKERHUB_TOKEN"` (the safe pattern) rather than string-interpolating the secret into the `jq` filter — prevents shell-injection if the token ever contains characters jq's filter language would parse.
- The PR body was updated to correct the earlier "auth failure during pull" framing to the empirically-observed "parse-time template validation" framing, addressing iter-1's `adjacent_improvements[0]`.
- The commit message `fix(ci): gate Docker Hub credentials behind a precheck job (B1)` follows conventional commits and cites the blocker ID, which makes the audit trail self-documenting.

---

**Note on review post:** GitHub blocks `gh pr review` self-author for both `--request-changes` and `--approve` because the auth user is the PR author. Verdict signal lives in this comment + the result file at `.claude/auto/review-results/pr-447-iter-2.json`. The orchestrator parses the result file.

Issue #464, opened 2026-05-14
Task: G4.3-T5b CLI verb `meho retrieval usage` + retrieval cobra command tree (Go follow-up to #444)


damir-topic:
Parent initiative: #373

## Target

Ship the **Go CLI verb `meho retrieval usage`** + tests + regenerated typed API client for the `/api/v1/retrieve/usage` endpoint. Deferred from #444 because the Go toolchain was unavailable on the implementing author's machine.

The backend API is fully shipped + tested in #444 (merged in `d201bc4`) and is operationally usable via `curl`; this Task ships the thin Cobra wrapper around it so operators can run `meho retrieval usage --since 30d` from their shell. Once T2 (#441) lands its `meho retrieval eval` verb, both verbs share the `retrieval` parent command this Task introduces.

## Files touched

- `cli/internal/api/client.gen.go` — regenerated from the existing `cli/api/openapi.json` (committed in #444). Mechanical: run `cd cli && make generate` and commit the diff. The snapshot already contains the new `/api/v1/retrieve/usage` path + `UsageReport` + `DailyUsageBucket` schemas.
- `cli/internal/cmd/retrieval/root.go` — `retrieval` parent cobra command. Will be reused by T2 (#441) for `meho retrieval eval`; file this Task ships the scaffolding, T2 just adds its `eval` subcommand on top.
- `cli/internal/cmd/retrieval/usage.go` — the `usage` subcommand. Flags: `--since`, `--surface`, `--tenant`, `--json`, `--backplane`. Calls the oapi-codegen-generated client method (`UsageEndpointApiV1RetrieveUsageGetWithResponse` or similar — confirm against the regenerated `client.gen.go`).
- `cli/internal/cmd/retrieval/usage_test.go` — mirrors `cli/internal/cmd/connector_test.go` patterns: parsed-args + status-code branches + redacted-error rendering.
- `cli/internal/cmd/root.go` — register the `retrieval` subtree onto the root cobra command.

## CLI shape

```bash
meho retrieval usage \
  [--since 30d]                       # default 30d; accepts 7d, 24h, 2026-04-01
  [--surface kb|memory|operations|all]
  [--tenant <uuid>]                   # tenant_admin only; backplane returns 403 otherwise
  [--json]                            # raw UsageReport JSON; default is text table
  [--backplane https://...]           # backplane URL override; same flag as `meho connector`
```

[...]

Issue #445, opened 2026-05-14
Task: G4.3-T6 Retire-checklist verb — combines eval + usage against 5 decision-#2 criteria; green/yellow/red per surface


damir-topic:
Parent initiative: #373
Depends-on: #441, #444

## Target

Ship the **retire-decision checklist** verb — `meho retrieval retire-checklist [--surface kb|memory|operations|all]`. Combines T2's eval results + T5's usage telemetry against decision #2 criteria, prints a per-surface checklist with green/yellow/red per criterion + an overall verdict (**READY TO RETIRE** / **REVIEW MANUALLY** / **NOT YET**). The load-bearing decision-support output.

Depends-on assigned to T2 only because T5 doesn't yet have a number at filing time; T5 dep added if needed when both numbers are known.

## Files touched

- `backend/src/meho_backplane/retrieval/retire.py` — checklist service.
- `cli/internal/cmd/retrieval/retire_checklist.go` — CLI wrapper.
- `backend/src/meho_backplane/api/v1/retrieve_retire.py` — new router.
- Tests at `backend/tests/test_retrieval_retire.py`.

## The 5 criteria (per surface)

1. **≥1 month elapsed since first daily-use date** (from usage telemetry).
2. **≥3 distinct operators with ≥1 search-per-week for ≥4 consecutive weeks** (from usage telemetry).
3. **Eval precision@5 ≥ 0.80** (green threshold from T2's threshold contract).
4. **MEHO ranking ≥ baseline on every metric** (kb: vs `grep -r kb/`; operations: vs `grep paths.txt + yq`).
5. **Zero open issues with `label:retrieval-migration-blocker`** filtered to the surface (read via `gh issue list`; T7 ships the label automation).

Each criterion gets green/yellow/red:
- **Green:** criterion satisfied.
- **Yellow:** within 70% of green threshold (manual review case).
- **Red:** criterion not met (block retire).

**Overall verdict:**
- READY TO RETIRE: all 5 criteria green.
- REVIEW MANUALLY: any yellow + no red.
- NOT YET: any red.

## CLI shape

```bash
meho retrieval retire-checklist \
  [--surface kb|memory|operations|all]   # default: all (per-surface results in one report)
  [--json]                                # machine-readable verdict + per-criterion breakdown
```

Default render: per-surface checklist table + bottom-line verdict. `--json` returns structured `RetireChecklistReport`.

## Acceptance criteria

[...]

Issue #443, opened 2026-05-14
Task: G4.3-T4 Memory eval corpus YAML — 10 queries across 5 scope shapes


damir-topic:
Parent initiative: #373
Depends-on: #440

## Target

Ship the **memory eval corpus YAML** — 10 queries with `expected_(scope, slug)` ground truth. Calibrates `search_memory` ranking quality. Soft-depends on G5.1 for the memory module to exist (the YAML schema is corpus-library-typed from T1 #440; entries can be drafted now).

## Files touched

- `backend/src/meho_backplane/retrieval/eval/memory_queries.yaml` — new 10-query corpus.
- Tests at `backend/tests/test_retrieval_eval_memory_corpus.py`.

## Corpus content

10 queries covering the 5 scope shapes proportionally (~2 per scope: user / user-tenant / user-target / tenant / target). Each entry:

```yaml
- query: "what kubectl flags do I prefer"
  expected_hits:
    - ["user", "kubectl-preferences"]              # ordered pairs (scope, slug)
  notes: "user-scope; tests basic personal-preference recall"

- query: "VPN setup for rdc-vcenter"
  expected_hits:
    - ["target", "rdc-vcenter-vpn-setup"]
    - ["user-target", "<operator>-rdc-vcenter-notes"]
  notes: "target-scope first; user-target as fallback"

# ... 8 more
```

Queries sourced from sample memory entries that a real operator workflow would generate (kubectl preferences, target-specific VPN notes, tenant convention reminders, etc.).

## Acceptance criteria

- [ ] `memory_queries.yaml` exists with 10 entries; all validate against `MemoryCorpusQuery` Pydantic schema from T1 #440.
- [ ] Coverage: ~2 queries per scope (user / user-tenant / user-target / tenant / target).
- [ ] `load_corpus("memory")` returns 10 queries after this Task lands.
- [ ] T2 #441 eval runner produces precision@5 + MRR + coverage against this corpus when seeded memories exist.
- [ ] CI green; YAML schema-validated.

## Out of scope

- Eval runner: T2 #441 (supports memory surface).
- The seeded test memory data: lives in T2's integration test fixtures.
- LLM-judge: deferred.

## References

- Parent Initiative: [#373 G4.3](https://github.com/evoila/meho/issues/373).
- Prerequisite: [#440 G4.3-T1 corpus library](https://github.com/evoila/meho/issues/440).
[...]

Issue #442, opened 2026-05-14
Task: G4.3-T3 Operations eval corpus YAML — 10 queries against ingested vendor surfaces (govc-parity benchmark)


damir-topic:
Parent initiative: #373
Depends-on: #440

## Target

Ship the **operations eval corpus YAML** — 10 representative queries against the ingested vendor surfaces (post-G0.6 + G0.7), each with `expected_connector_id` + ordered `expected_op_ids` ground truth. Calibrates the "is `search_operations` ranking govc-parity?" question quantitatively. Soft-depends on [#408 G0.7-T8 vSphere canary](https://github.com/evoila/meho/issues/408) for the operations to exist — but the YAML can be drafted now from the consumer-side analysis benchmark.

## Files touched

- `backend/src/meho_backplane/retrieval/eval/operation_queries.yaml` — new 10-query corpus.
- Tests at `backend/tests/test_retrieval_eval_operation_corpus.py` (schema validation + slug shape).

## Corpus content

Per the consumer-side govc-vs-meho analysis (cited in master amendment plan), 10 representative govc workflows that `search_operations` should rank in the top-3. Each entry:

```yaml
- query: "list VMs in cluster X"
  expected_connector_id: "vmware-rest-9.0"
  expected_op_ids:
    - "GET:/api/vcenter/vm"
  govc_equivalent: "govc vm.info -dc <dc> -cluster <cluster>"
  notes: "operator's most-common daily lookup"

- query: "revert vSphere VM snapshot"
  expected_connector_id: "vmware-rest-9.0"
  expected_op_ids:
    - "vmware.composite.vm.snapshot.revert"
    - "vi-json:.../VirtualMachine/{moId}/RevertToSnapshot_Task"
  govc_equivalent: "govc snapshot.revert"
  notes: "composite preferred; vi-json acceptable fallback"

# ... 8 more, covering inventory + lifecycle + performance + events + snapshot + storage + network + host
```

The 10 queries should map 1:1 to the govc-parity benchmark in [#408 G0.7-T8 vSphere canary](https://github.com/evoila/meho/issues/408). When canary acceptance runs, it asserts precision@3 on this same corpus.

## Acceptance criteria

- [ ] `operation_queries.yaml` exists with 10 entries; all validate against `OperationCorpusQuery` Pydantic schema from T1 #440.
- [ ] Each query maps to a govc workflow (per `govc_equivalent` field); covers inventory + VM + host + cluster + datastore + network + snapshot + event + performance + composite.
[...]

PR #473, opened 2026-05-14
feat(retrieval): G4.3-T2 eval runner — meho retrieval eval CLI + /api/v1/retrieve/eval API + precision@5/MRR/coverage + grep baseline + CI gate

### LINKED ISSUE #441: Task: G4.3-T2 Eval runner — meho retrieval eval CLI + /api/v1/retrieve/eval API + precision@5/MRR/coverage + grep baseline + CI gate
Parent initiative: #373

Depends-on: #440

## Target

Ship the **eval runner** — `meho retrieval eval [--surface kb|memory|operations|all] [--baseline grep] [--json]` CLI + `/api/v1/retrieve/eval` API route. Computes precision@5 + MRR + coverage per query and overall; outputs green/yellow/red per the threshold contract.

The runner is corpus-agnostic — same code path drives all 3 surfaces; per-surface differences (slug vs (scope, slug) vs op_id ground truth) handled by the dispatcher inside the runner.

**Depends-on assigned at filing time** based on T1's actual number.

## Files touched

- `backend/src/meho_backplane/retrieval/eval/runner.py` — eval engine.
- `backend/src/meho_backplane/retrieval/eval/metrics.py` — precision@k / MRR / coverage compute.
- `backend/src/meho_backplane/retrieval/eval/baseline_grep.py` — grep -r kb/ subprocess wrapper.
- `backend/src/meho_backplane/api/v1/retrieve_eval.py` — new router.
- `cli/internal/cmd/retrieval/eval.go` — CLI wrapper.
- Modify `backend/src/meho_backplane/main.py` to mount new router.
- Tests at `backend/tests/test_retrieval_eval_runner.py` + `test_retrieval_eval_metrics.py`.

## Threshold contract (per surface)

```
Green:    precision@5 >= 0.80  AND  MRR >= 0.50  AND  coverage >= 0.90
Yellow:   any metric below green but >= 70% of green
Red:      any metric below 70% of green threshold

For operations surface, "MEHO ≥ baseline" replaces grep-baseline with
"the operator's pre-MEHO grep + yq workflow against locally-cloned spec".
```

## CLI shape

```bash
meho retrieval eval \
  [--surface kb|memory|operations|all]   # default: all
  [--baseline grep]                       # run grep -r baseline alongside (kb only in v0.2)
  [--save-baseline <file>]                # save this run's metrics for future regression detection
  [--compare-baseline <file>]             # compare this run against saved baseline; exit 1 on regression
  [--json]                                # machine-readable output
```

Default render: per-surface text table with verdict + per-query breakdown. `--json` returns the full `EvalResult` Pydantic model.

## API route

```python
@router.post("/api/v1/retrieve/eval")
async def eval(body: EvalRequest, operator: Operator = _require_operator):
    """Body: {surface: "kb"|"memory"|"operations"|"all", baseline?: "grep"}
    Returns: EvalResult (per-query + overall metrics + green/yellow/red verdict).
    operator role; tenant-scoped automatically."""
```

## CI gate (optional but landed in T2)

The eval runner exits non-zero on **red** verdict. A GitHub Actions workflow under `.github/workflows/eval-gate.yml` runs `meho retrieval eval --surface kb --compare-baseline ci/eval-baseline.json` on every backend PR; failures gate merge. Baseline file checked into the repo + updated on intentional ranking-improvement PRs.

## Acceptance criteria

- [ ] `meho retrieval eval --surface kb` runs the 10-query KB corpus against `/api/v1/retrieve` + reports precision@5 + MRR + coverage + verdict.
- [ ] `--baseline grep` runs `grep -r kb/` on the consumer kb snapshot + reports same metrics; the runner asserts MEHO ≥ baseline as a precondition for green.
- [ ] `--json` returns machine-readable EvalResult.
- [ ] `--save-baseline` / `--compare-baseline` round-trip works.
- [ ] Surface=`memory` runs against memory corpus (no-ops cleanly if T4 corpus isn't filed yet — returns empty).
- [ ] Surface=`operations` runs against ops corpus (no-ops cleanly if T3 corpus isn't filed yet).
- [ ] Surface=`all` runs the union; per-surface verdicts in the output.
- [ ] `/api/v1/retrieve/eval` route works for all of the above via HTTP; tenant-scoped.
- [ ] Audit row emitted per eval call (`op_id="meho.retrieval.eval"`); broadcast classifier: `op_class="audit_query"` (aggregate-only — the queries themselves can be sensitive).
- [ ] CI workflow `.github/workflows/eval-gate.yml` exists; runs on backend PRs; gates on red verdict.
- [ ] CI green; ruff + mypy clean.

## Out of scope

- Memory corpus YAML: T4.
- Operations corpus YAML: T3.
- Usage telemetry: T5.
- Retire-checklist: T6.
- LLM-judge evaluation: deferred (v0.2 keyword-exact ground truth only).

## References

- Parent Initiative: [#373 G4.3](https://github.com/evoila/meho/issues/373).
- Prerequisite: T1 (corpus library + KB corpus).
- Retrieve endpoint to evaluate: [backend/src/meho_backplane/api/v1/retrieve.py](backend/src/meho_backplane/api/v1/retrieve.py) (G0.4-T5).
- Best-practices anchor: `.claude/skills/implement-issue/ai_engineering_best_practices.md` (retrieval evaluation metrics; CI-gated regression).

damir-topic:
## Summary

Ships the corpus-driven retrieval-eval runner that turns Goal #215's "10 sampled queries return equivalent answers to `grep -r kb/`" criterion into a falsifiable, CI-gated check. Without this, the retire-decision for the consumer's `kb/` directory remained an unfalsifiable claim that defaulted to whoever spoke last in the team-of-4 review.

- **Pure metrics + verdict bands** — `precision_at_k`, `reciprocal_rank`, `coverage_at_k`, `verdict` (green/yellow/red per the threshold contract: `precision@5 >= 0.80 AND MRR >= 0.50 AND coverage >= 0.90` = green, below 70% of any green = red, worst-metric-wins).
- **Async runner** — `eval_all` / `eval_surface` dispatch the same code path across kb / memory / operations; surfaces with no shipped corpus return `query_count=0` + `verdict=green` so the CI gate doesn't flip red on absent corpora (T6 #445's retire-checklist asserts the corpus actually exists before trusting the green).
- **Subprocess grep baseline** — literal-string `grep -r -l -i -F --include=*.md`; MEHO-≥-baseline overlay downgrades to red on any per-metric loss (Initiative #373 retire-criterion #4).
- **HTTP route** — `POST /api/v1/retrieve/eval`, operator-role minimum, tenant-scoped, with `audit_op_id="meho.retrieval.eval"` + `audit_op_class="audit_query"` overrides so the broadcast publisher emits aggregate-only events (eval queries can leak operator intent — same posture T5's `retrieve_usage` adopted).
- **Go CLI** — `meho retrieval eval [--surface] [--baseline grep] [--save-baseline] [--compare-baseline] [--json]`, exits 1 on red verdict OR baseline regression. Uses direct `net/http` on three new `AuthedClient` helpers (`HTTPClient`, `AccessToken`, `Refresh`) since the typed wrapper for the new endpoint lands in a follow-up `make generate` pass.
- **CI gate** — `.github/workflows/eval-gate.yml` runs `scripts/ci/run_eval_gate.py` against the runner + checked-in baseline at `ci/eval-baseline.json`. Path-scoped to retrieval-touching PRs.

## Test plan

- [x] `ruff check src/ tests/` — clean (192 files)
- [x] `ruff format --check src/ tests/` — clean
- [x] `mypy src/` (strict) — clean (107 source files)
- [x] `pytest backend/tests/ --ignore=tests/integration` — 1156 passed, 3 skipped (including 80 new tests across the 4 new test files: `test_retrieval_eval_metrics.py`, `test_retrieval_eval_runner.py`, `test_retrieval_eval_baseline_grep.py`, `test_api_v1_retrieve_eval.py`)
- [x] `python3 .claude/hooks/code-quality.py --diff` — 0 blockers, 7 warnings (mostly PLR0913 on internal helpers + one pre-existing warning on `main.py::lifespan`)
- [x] `scripts/ci/run_eval_gate.py --compare-baseline ci/eval-baseline.json` — `GATE PASSED: no regression vs baseline`
- [x] `make snapshot-openapi` re-run; `cli/api/openapi.json` includes the new `/api/v1/retrieve/eval` operation
- [x] Surface=`kb` with mocked `retrieve` returns the expected 10-query result + green verdict; `--baseline grep` populates baseline_* fields and applies the MEHO-≥-baseline overlay
- [x] Surface=`memory` / `operations` return `query_count=0` + verdict=`green` per the absent-corpus rule (T3/T4 corpora ship in #442/#443)
- [x] `--save-baseline` / `--compare-baseline` round-trip preserves the EvalResult shape; per-metric regressions > epsilon (default 0.02) flagged
- [x] Audit row payload carries `op_id`/`op_class`/`eval_surface`/`eval_baseline`/`row_count`
- [x] RBAC — `read_only` JWT → 403, `operator`/`tenant_admin` → 200, unauthenticated → 401
- [x] DCO sign-off present on the commit (`Signed-off-by: damir-topic <damir.topic@ed-consulting.at>`)
- [ ] CI: full `eval-gate` workflow runs green on the PR
- [ ] CI: `pr-smoke` + `python-lint-test` + `go-lint-test` runs green
- [ ] Post-merge: `meho retrieval eval --surface kb` end-to-end against the live deployed backplane (T6 #445 wires this as the inner ring of the live-substrate gate)

## Out of scope

- Memory + operations corpus YAML — T4 (#443) / T3 (#442) (in flight in Wave 2 alongside this Task per Initiative #373's DAG).
- Server-side baseline corpus snapshot for the API route — v0.2 CLI runs grep locally; v0.2.next can wire a server-side path.
- LLM-judge evaluation — v0.2 keyword-exact ground truth only (per Initiative body).
- `client.gen.go` regeneration — happens in a follow-up `make generate` pass; the eval CLI uses direct `net/http` on the existing `AuthedClient` helpers in the meantime (matches the T5b #464 deferral pattern).

Closes #441

## Summary by CodeRabbit

* **New Features**
  * POST /api/v1/retrieve/eval endpoint for corpus-driven retrieval evaluation with per-surface metrics (precision@5, MRR, coverage) and overall verdict.
  * meho retrieval eval CLI: run evaluations, save/compare baselines, JSON output, and CI-gate exit codes.
  * CI workflow to run the retrieval-quality gate automatically.

* **Tests**
  * Broad end-to-end and unit tests covering API, CLI, runner, metrics, baseline I/O, and gate script.

* **Documentation**
  * Backend and API docs updated for the retrieval-eval flow, thresholds, and CLI/CI behavior.


damir-topic:
@coderabbitai review

cli/internal/cmd/retrieval/eval.go
```diff
@@ -0,0 +1,481 @@
...
+	BaselineMRR          *float64          `json:"baseline_mrr,omitempty"`
+	BaselineCoverage     *float64          `json:"baseline_coverage,omitempty"`
+	BaselineVerdict      *string           `json:"baseline_verdict,omitempty"`
+	Queries              []EvalQueryResult `json:"queries"`
+}
+
+// EvalResult mirrors the backend EvalResult model — the top-level
+// shape returned by POST /api/v1/retrieve/eval.
+type EvalResult struct {
+	RanAt           string              `json:"ran_at"`
```
damir-topic: **[🛑 B1] — Go CI lint FAILS: gofmt formatting violation in EvalResult struct (PR-introduced)**

**What:** `golangci-lint v1.64.8` reports `cli/internal/cmd/retrieval/eval.go:63:1: File is not properly formatted (gofmt)`. The struct field alignment in `EvalResult` (lines 62-67) has extra column padding inconsistent with gofmt's tab-stop rules — `RanAt`, `Surfaces`, `OverallVerdict`, `Thresholds` are over-padded relative to their tag column.

**Why it matters:** The Go CI check is **RED** on this PR — branch protection blocks merge while CI fails on a check this PR introduced. The file is new; the lint failure is 100% on this diff.

**Required change:** Run `gofmt -w cli/internal/cmd/retrieval/eval.go` (or `make fmt` if the CLI Makefile exposes it) and commit the result. The fix collapses the extra-padding columns in the `EvalResult` struct so gofmt's tab-stop alignment is satisfied. Verify `cd cli && golangci-lint run` exits 0 before pushing.

**Source:** golangci-lint v1.64.8 `gofmt` linter — https://golangci-lint.run/usage/linters/#gofmt

backend/src/meho_backplane/retrieval/eval/baseline_grep.py
```diff
@@ -0,0 +1,291 @@
...
+        "-i",
+        "-F",
+        "--include=*.md",
+        "--",
+        query,
+        str(corpus_root),
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.PIPE,
+    )
+    stdout_bytes, stderr_bytes = await proc.communicate()
```
damir-topic: **[🛑 B2] — grep subprocess has no timeout; eval routes can hang indefinitely (also flagged by CodeRabbit)**

**What:** `_spawn_grep` calls `await proc.communicate()` with no timeout. A wedged `grep` (network FS, large symlink loop, runaway corpus root) hangs the eval call forever — and through it, `/api/v1/retrieve/eval` and the CI gate workflow.

**Why it matters:** Today's corpus is small + checked-in so practical risk is low, but the API route is a request path operators hit; a single hung grep blocks the FastAPI worker indefinitely with no kill path. Defense-in-depth around `asyncio.create_subprocess_exec` calls is a python_best_practices invariant. This becomes critical the moment T7 / v0.2.next wires a server-side baseline corpus snapshot per the docstring.

**Required change:** Wrap the communicate call in `asyncio.wait_for(..., timeout=15.0)`; on `TimeoutError`, call `proc.kill()` then `await proc.communicate()` to drain, and raise `BaselineConfigError("grep baseline timed out after 15s")`. CodeRabbit's suggested diff is correct — apply as-is, or tighten the timeout (5s is reasonable for a small kb corpus).

```python
try:
    stdout_bytes, stderr_bytes = await asyncio.wait_for(
        proc.communicate(),
        timeout=15.0,
    )
except TimeoutError as exc:
    proc.kill()
    await proc.communicate()
    raise BaselineConfigError("grep baseline timed out after 15s") from exc
```

backend/src/meho_backplane/retrieval/eval/baseline_grep.py
```diff
@@ -0,0 +1,291 @@
...
+
+    # grep exit codes: 0 = matches found, 1 = no matches, 2 = error.
+    # Treat 0/1 as success-with-different-result-shapes; anything else
+    # is an operator-actionable failure.
+    if proc.returncode not in (0, 1):
+        structlog.get_logger().warning(
+            "baseline_grep_nonzero_exit",
+            returncode=proc.returncode,
+            stderr=stderr_bytes.decode("utf-8", errors="replace")[:500],
+            query=query,
```
damir-topic: **[🛑 B3] — Raw eval query logged on grep failure path — leaks operator-sensitive search intent (also flagged by CodeRabbit)**

**What:** `structlog.get_logger().warning("baseline_grep_nonzero_exit", ..., query=query, ...)` emits the raw query string verbatim into the structured log.

**Why it matters:** `retrieve_eval.py:27-29` and the audit-classifier choice on this PR both explicitly call out that "eval queries themselves can be operator-sensitive — 'is the operator searching for the disk-failure runbook?' leaks intent." The audit row uses `audit_op_class=audit_query` aggregate-only mode for that exact reason. This log line bypasses the protection: cluster log aggregation (Loki / CloudWatch / Splunk) inherits the leak. Operationally inconsistent with the rest of the route's PII posture.

**Required change:** Replace `query=query` with `query_len=len(query)` (CodeRabbit's exact suggestion), or with a stable hash:

```python
import hashlib
query_sha256=hashlib.sha256(query.encode()).hexdigest()[:16]
```

Either preserves the operator's ability to correlate the failure to a specific input (length / hash) without leaking the content. Don't both log and audit-redact — pick one redaction strategy and apply it consistently.

scripts/ci/run_eval_gate.py
```diff
@@ -0,0 +1,195 @@
...
+        print("\nGATE FAILED: overall verdict is red", file=sys.stderr)
+        return 1
+
+    if not args.compare_baseline.exists():
+        print(
+            f"\nNo baseline at {args.compare_baseline} — "
+            "run with --save-baseline to create one. Skipping regression check.",
+            file=sys.stderr,
+        )
+        return 0
```
damir-topic: **[🛑 B4] — CI gate silently passes when baseline file is missing — gate-bypass surface (also flagged by CodeRabbit)**

**What:** When `args.compare_baseline` doesn't exist, `_run` prints a "No baseline ... Skipping regression check" warning and `return 0` (exit success). The workflow's whole purpose is to gate on baseline regressions; a missing baseline file makes the gate a silent no-op.

**Why it matters:** Gate-bypass via deletion: if the baseline file gets renamed, deleted, or path-shadowed (a refactor moves `ci/eval-baseline.json`, a `.gitignore` change drops it, etc.), the workflow goes silently green. The `--compare-baseline` invocation in `.github/workflows/eval-gate.yml` is the load-bearing call — there is no other gate. This **contradicts the file's own docstring** (lines 58-63: "Exit codes: 0 — green/yellow, no regression. 1 — red verdict, OR baseline regression detected, OR runtime exception") by adding a fourth case (exit 0 — baseline missing) that's not advertised and silently disables the only regression check.

**Required change:** Fail loud when running in CI. CodeRabbit's gate via `os.getenv("CI")` preserves local-dev ergonomics:

```python
if not args.compare_baseline.exists():
    if os.getenv("CI"):
        print(f"\nGATE FAILED: baseline missing at {args.compare_baseline}", file=sys.stderr)
        return 1
    print(f"\nNo baseline at {args.compare_baseline} — ...", file=sys.stderr)
    return 0
```

Unconditional `return 1` is also acceptable (and simpler) since the baseline file is checked-in — the local-dev "skipping" case is rare.

backend/src/meho_backplane/api/v1/retrieve_eval.py
```diff
@@ -0,0 +1,177 @@
...
+    # The grep baseline isn't wired to a server-side corpus root in
+    # v0.2 — operators run the baseline locally via the CLI verb
+    # against a checked-in snapshot. The API surface accepts the flag
+    # for forward-compat (T7 / v0.2.next can wire a server-side
+    # snapshot path) but ignores it today; the response carries
+    # ``baseline_kind=null`` per surface to signal that.
+    result = await eval_all(
+        tenant_id=operator.tenant_id,
+        surfaces=surfaces,
+    )
```
damir-topic, line 211: **[⚠️ M1] — `baseline` body field accepted but silently dropped (also flagged by CodeRabbit)**

**What:** `EvalRequest.baseline` (line 108) is accepted by the request schema, but `eval_endpoint` calls `await eval_all(tenant_id=..., surfaces=surfaces)` without passing `baseline` (lines 167-170). The docstring above intentionally documents this as forward-compat scaffolding.

**Why it matters:** Documented-but-bad API contract. A caller that POSTs `{"surface": "kb", "baseline": "grep"}` reasonably expects the baseline to run (or to get a 422). Today they get a 200 with `baseline_kind: null` per surface, which they may not notice — and the load-bearing CI signal "MEHO ≥ baseline" (Initiative #373 retire-criterion #4) silently does not fire on the API path. The CLI runs the baseline locally so the gate works there, but API-only consumers (e.g., a future external dashboard, a non-Go integration) get false confidence.

**Required change:** Two acceptable paths:

1. **Reject the field at validation** — drop `baseline` from `EvalRequest` until T7 / v0.2.next wires the server-side corpus snapshot. Cleanest fix; the CLI doesn't send the field to the API today anyway (it runs grep locally).
2. **Reject `baseline="grep"` with 501 Not Implemented** + a body pointing to the CLI as the supported path. Preserves the forward-compat shape but makes the contract honest.

Silently accepting + ignoring is the worst option — that's what's there today. Pick (1) for minimum diff.

cli/internal/cmd/retrieval/eval.go
```diff
@@ -0,0 +1,481 @@
...
+		fmt.Fprintf(w, "%-12s %-7s %10d %10.3f %10.3f %10.3f\n",
+			s.Surface, s.Verdict, s.QueryCount, s.PrecisionAt5, s.MRR, s.Coverage)
+		if s.BaselineKind != nil && s.BaselinePrecisionAt5 != nil {
+			fmt.Fprintf(w, "%-12s %-7s %10s %10.3f %10.3f %10.3f\n",
+				"  baseline:"+*s.BaselineKind,
+				strDeref(s.BaselineVerdict),
+				"-",
+				*s.BaselinePrecisionAt5,
+				*s.BaselineMRR,
+				*s.BaselineCoverage,
```
damir-topic, line 510: **[⚠️ M2] — Nil-pointer dereference risk on baseline metric pointers (also flagged by CodeRabbit)**

**What:** The gate checks `s.BaselineKind != nil && s.BaselinePrecisionAt5 != nil` then dereferences `*s.BaselineMRR` and `*s.BaselineCoverage` unconditionally on lines 462-463. The backend `SurfaceResult` Pydantic model declares each baseline_* field as independently nullable (`float | None`).

**Why it matters:** The runner currently always populates the trio together via `_aggregate_baseline`, so today's panic risk is zero. But the JSON contract from the API permits any subset to be set; future code that emits a partial baseline (e.g., only `precision_at_5` because the others didn't compute, or a manually-edited baseline file fed via `--compare-baseline`) would crash the CLI with a nil-deref runtime panic. Cheap to defend.

**Required change:** Extend the gate to all three pointers:

```go
if s.BaselineKind != nil &&
    s.BaselinePrecisionAt5 != nil &&
    s.BaselineMRR != nil &&
    s.BaselineCoverage != nil {
    fmt.Fprintf(w, "%-12s %-7s %10s %10.3f %10.3f %10.3f\n",
        "  baseline:"+*s.BaselineKind,
        strDeref(s.BaselineVerdict),
        "-",
        *s.BaselinePrecisionAt5,
        *s.BaselineMRR,
        *s.BaselineCoverage,
    )
}
```

Alternative: per-field placeholder (`"-"`) when any is nil, mirroring `strDeref(s.BaselineVerdict)`. Pick whichever matches the broader nil-handling style in the CLI.

cli/internal/cmd/retrieval/eval.go
```diff
@@ -0,0 +1,481 @@
...
+	BackplaneOverride   string
+}
+
+// runEval is the eval command's RunE body. Kept separate so the
+// flag-parsing boilerplate stays in newEvalCmd and the request-
+// response logic stays testable in isolation.
+func runEval(cmd *cobra.Command, opts evalOptions) error {
+	backplaneURL, err := resolveBackplane(opts.BackplaneOverride)
+	if err != nil {
+		return output.RenderError(cmd.ErrOrStderr(), output.AuthExpired(err.Error()), opts.JSONOut)
```
damir-topic: **[💡 m1] — All `resolveBackplane` errors mapped to AuthExpired — exit-code semantics broken (also flagged by CodeRabbit)**

**What:** `output.RenderError(cmd.ErrOrStderr(), output.AuthExpired(err.Error()), ...)` is called for every error from `resolveBackplane`, including invalid `--backplane` URLs (`normaliseURL` errors), file-system errors from `auth.LoadConfig`, and parse errors. The function's documented exit codes (lines 89-93) explicitly name `2 = auth_expired` vs `3 = unreachable` vs `4 = unexpected`.

**Why it matters:** Operators (and CI scripts) keying off the exit code see `auth_expired = 2` when the actual cause was a typo in `--backplane http:/example` (no slashes). They run `meho login` instead of fixing the URL. Same regression on the `auth.LoadConfig` path. Minor because it's a developer-experience papercut, not a correctness bug.

**Required change:** Inspect the error and pick the right renderer:

```go
backplaneURL, err := resolveBackplane(opts.BackplaneOverride)
if err != nil {
    if errors.Is(err, auth.ErrConfigNotFound) {
        return output.RenderError(cmd.ErrOrStderr(), output.AuthExpired(err.Error()), opts.JSONOut)
    }
    return output.RenderError(
        cmd.ErrOrStderr(),
        output.Unexpected(fmt.Sprintf("resolve backplane: %v", err)),
        opts.JSONOut,
    )
}
```

Refining the taxonomy further (separating `normaliseURL` parse errors from other `auth.LoadConfig` errors) is a v0.2.next polish.

damir-topic (commented):




**Anchor task:** #441 — G4.3-T2 Eval runner (precision@5 / MRR / coverage + grep baseline + CI gate)
**Surface area:** backend + go-cli + devops
**Best-practices packs audited:** python + ai + devops
**Verdict:** REQUEST_CHANGES (note: GitHub blocks REQUEST_CHANGES state when reviewer == author, so verdict is recorded here in the body; result file at `.claude/auto/review-results/pr-473-iter-1.json` is authoritative)
**Head SHA reviewed:** `97d545cd228152652c81e9c1dab87178637a349e`
**AI reviewer (CodeRabbit):** received in 60s — 6 actionable inline findings, all valid

### Acceptance criteria
- [x] `meho retrieval eval --surface kb` reports precision@5 + MRR + coverage + verdict — `test_eval_surface_kb_*` + Go `eval_test.go` cover the path
- [x] `--baseline grep` reports same metrics + asserts MEHO ≥ baseline — `_apply_baseline_check` covered, but **see M1** for the API-route gap
- [x] `--json` returns machine-readable EvalResult — Pydantic v2 frozen + Go struct mirror
- [x] `--save-baseline` / `--compare-baseline` round-trip — `baseline_io.py` + matching Go tests
- [x] Surface=`memory` / Surface=`operations` no-op cleanly (empty corpus → green + query_count=0) — `test_eval_surface_memory_empty_corpus_is_green` + ops mirror
- [x] Surface=`all` runs union with per-surface verdicts — `test_eval_all_*` covers
- [x] `/api/v1/retrieve/eval` route + tenant-scoped — handler reads `operator.tenant_id`, never accepts body tenant id
- [x] Audit row emitted with `op_id=meho.retrieval.eval` + `op_class=audit_query` — `test_eval_route_audit_payload_*` covers
- [x] CI workflow `.github/workflows/eval-gate.yml` exists, path-scoped — verified; but **see B4** on the gate-bypass surface
- [❌] CI green; ruff + mypy clean — ruff/mypy clean locally, but **Go CI is RED** on a PR-introduced gofmt violation (B1) and Python CI is RED on a pre-existing broken-main test (separate, not this PR's fault — recorded as adjacent)

### Project-board alignment
|---|---|---|---|---|---|
| #441 | In Review | Closes | ✅ | partial — code complete, CI not green | block on B1–B4 + 2 majors; **do not close** until fix-up iteration |

### Predicted vs. actual diff
- ✅ 17 files match the predicted blast-radius (metrics / runner / result_models / API route / corpus integration + CI workflow + baseline JSON + docs)
- ⚠️ 4 files have substantive issues — see inline B1-B4 + M1-M2
- ❓ `cli/internal/api/client.go` — 3 new helpers (HTTPClient, AccessToken, Refresh) for unwrapped HTTP calls; documented as a pre-codegen-pass bridge; reasonable surface addition
- 🚫 None undisclosed
- 🕳️ None missing

### Integration verification
- [x] `ruff check` on `retrieval/eval/` + `retrieve_eval.py` — clean (worktree)
- [x] `mypy` on 8 changed source files — clean (worktree)
- [x] `pytest` on the 4 new test files — 80 passed (worktree, with `RETRIEVAL_MODEL_CACHE_DIR=/tmp/fastembed-cache`)
- [x] `scripts/ci/run_eval_gate.py --compare-baseline ci/eval-baseline.json` — `GATE PASSED: no regression vs baseline`
- [❌] `golangci-lint run` (CI) — **FAIL** on PR-introduced gofmt violation (B1)
- [⚠️ broken main] `pytest tests/` (CI) — collection error on `tests/test_db_broadcast_override.py` (pre-existing from PR #458 / f340d50; not this PR's fault but blocks merge until fixed on main — see Adjacent improvements)
- [⚠️ not run] migrations replay — n/a, no migration in this PR
- [⚠️ not run] `docker compose up` — n/a
- [⚠️ not run] live agent trace — `eval_all` against the real PG+pgvector retriever is the T6 #445 outer ring; this PR's gate exercises the stub-retrieve runner end-to-end and that test passed

- 🛑 **B1** — Go CI FAIL: gofmt formatting violation at `cli/internal/cmd/retrieval/eval.go:63` (PR-introduced; CI red). Required: run `gofmt -w` on the file.
- 🛑 **B2** — `_spawn_grep` has no timeout; hung grep blocks the eval API + CI gate. Required: wrap `proc.communicate()` in `asyncio.wait_for(..., timeout=15.0)` + `proc.kill()` fallback. (Also CodeRabbit.)
- 🛑 **B3** — `baseline_grep_nonzero_exit` log emits raw `query=query`, bypassing the route's PII-aware aggregate-only audit posture. Required: replace with `query_len=len(query)` or a SHA256 prefix. (Also CodeRabbit.)
- 🛑 **B4** — `run_eval_gate.py` returns 0 on missing baseline file; gate-bypass via deletion. Required: `return 1` (optionally CI-gated via `os.getenv("CI")`). (Also CodeRabbit.)
- ⚠️ **M1** — `EvalRequest.baseline` accepted by API but silently dropped before `eval_all`. Required: drop the field from the schema (or 501 on `baseline="grep"`). (Also CodeRabbit.)
- ⚠️ **M2** — Nil-pointer deref risk on `s.BaselineMRR` / `s.BaselineCoverage` in Go CLI table renderer. Required: extend the nil-gate to all three baseline pointers. (Also CodeRabbit.)
- 💡 **m1** — `runEval` collapses all `resolveBackplane` errors to `AuthExpired`, breaking documented exit-code taxonomy. Required: `errors.Is(err, auth.ErrConfigNotFound)` discrimination. (Also CodeRabbit.)

- **Pre-existing broken main**: `tests/test_db_broadcast_override.py` (from PR #458, f340d50, "G6.3-T1 BroadcastOverride ORM model + Alembic migration 0007") imports `BroadcastOverride` from `meho_backplane.db.models`, but the ORM class was **never added** in that PR — only the test + migration shipped. Every PR opened against current main fails Python CI at collection on the same import error. This blocks merge of PR #473 until separately resolved on main (either land the missing ORM class as a follow-up, or revert the test). Not this PR's responsibility.
- Per-surface concurrency: `eval_all` runs surfaces sequentially. Fine today (stub retrieve = ms); once T3/T4 ship live corpora and the real PG-backed retriever runs (100-500ms per query), `asyncio.gather`-ing the surfaces cuts wall clock 3x. v0.2.next.
- `EvalResult.thresholds` gets baked into every saved baseline file — `compare_baseline` would use the baseline's saved thresholds, which is probably correct (regression detection should use the baseline's contract) but worth a docstring note.

- Adopted: 6 of 6 (all 6 inline findings carry concrete required-changes and were merged into the punch list above with attribution)
- Disputed: 0
- Out of scope: 0

### Docs
- [x] `docs/codebase/backend.md` — updated (+1 line per the PR diff). Sufficient as a minimum-update; a dedicated eval-surface section is a future polish.

- **Metrics module is textbook-correct**: precision@k uses `len(top_k)` as denominator (not `k`) so a system returning fewer results than `k` isn't penalised for missing slots — the load-bearing IR-correctness call; documented with the rationale. `reciprocal_rank` is 1-based and returns 0.0 on miss. `verdict` is single-pass worst-wins. `k <= 0` raises `ValueError` instead of returning 0.0 (typo-at-call-site surfaces immediately).
- **Subprocess safety done right (modulo the timeout gap)**: `asyncio.create_subprocess_exec` (not shell), `--` separator before pattern so a leading-dash query doesn't become a grep flag, argv-not-stdin for the query, `shutil.which` upfront so missing-binary fails with a clear message before the spawn.
- **Audit posture is on the right side**: `audit_op_class=audit_query` + aggregate-only broadcast event for eval queries — the issue body's PII concern is honoured at the audit layer (B3 is about the log layer separately leaking it). Override contextvars bound *before* the runner kicks off so a handler exception still produces a partial audit row.
- **Empty-corpus discipline**: `query_count=0 + verdict=green` so an absent T3/T4 corpus doesn't flip the gate red on shipping order. T6 retire-checklist becomes the corpus-actually-shipped gate (correct division of responsibility).
- **Pydantic v2 freezing**: every result model has `model_config = ConfigDict(frozen=True)` + `extra="forbid"` on the request schema. Prevents the runner from mutating rows mid-aggregation and catches typos at the framework boundary (422 instead of silent default-surface).
- **CI workflow shape**: SHA-pinned actions, fail-loud (no `continue-on-error`), path-scoped to retrieval changes, `meho-runners` (per the team's runner-pool convention), `--locked` on `uv sync` so dep drift surfaces at the gate rather than silently bumping.

damir-topic:
@coderabbitai review

damir-topic:




**Anchor task:** #441 — G4.3-T2 eval runner (meho retrieval eval CLI + /api/v1/retrieve/eval API + precision@5/MRR/coverage + grep baseline + CI gate)
**Surface area:** backend, go-cli, devops
**Best-practices packs audited:** python, ai, devops
**Verdict:** APPROVE
**Head SHA reviewed:** `99907d75f3d0ea9f67705b50b14547938309a6b0`
**AI reviewer (CodeRabbit):** received in ~6 min (post iter-1-fix re-trigger)

### Iter-1 punch-list closure (7/7 closed)

| ID | Finding | Disposition | Verification |
|---|---|---|---|
| B1 | Go CI gofmt FAIL | ✅ FIXED | Go CI green on iter-2 head SHA; struct/var blocks re-aligned cleanly |
| B2 | grep subprocess no timeout | ✅ FIXED | `GREP_TIMEOUT_SECONDS=15.0` + `asyncio.wait_for` + `proc.kill()` + best-effort drain; test pins the timeout path |
| B3 | Raw query in error logs | ✅ FIXED | Replaced `query=query` with `query_len`+`query_sha256[:16]` in both `baseline_grep_nonzero_exit` and the new `baseline_grep_timeout` warnings; test pins the redaction |
| B4 | CI gate silent-bypass on missing baseline | ✅ FIXED | Missing-baseline branch returns 1 + `GATE FAILED: baseline missing` on stderr; docstring updated; smoke-verified locally (exit 1 confirmed) |
| M1 | API silently drops `baseline` field | ✅ FIXED | Returns 501 with clear detail; audit context bound *before* the gate so rejection still records operator intent; OpenAPI regenerated to expose 501; two tests pin the contract |
| M2 | Go nil-deref on baseline pointers | ✅ FIXED | Gate extended from 2 → 4 pointers (`BaselineKind && BaselinePrecisionAt5 && BaselineMRR && BaselineCoverage`) |
| m1 | Go error taxonomy collapsed to AuthExpired | ✅ FIXED | `errNoBackplaneConfigured` wrapper + `classifyBackplaneError` routes `ErrConfigNotFound → AuthExpired`; everything else → `Unexpected`; test pins both branches |

### Acceptance criteria (#441)
- [x] `meho retrieval eval` CLI verb runs eval corpus + emits precision@5 / MRR / coverage
- [x] `/api/v1/retrieve/eval` POST route mirrored shape (now also rejects baseline with 501)
- [x] grep baseline available (CLI-side; API rejects loud)
- [x] CI gate workflow with regression detection (now fails loud on missing baseline)
- [x] Per-surface verdict (green / yellow / red) + overall verdict

### Predicted vs. actual diff (iter-2 delta)
- ✅ Expected and correct: every file in the iter-2 delta (5 commits, 10 files) maps to one of the 7 iter-1 findings or to the docs row update
- ❓ Unexpected but justified: `test_ci_run_eval_gate.py` — brand new in iter-2, smoke-tests the B4 fail-loud branch via subprocess (good)
- 🚫 Unexpected and undisclosed: none
- 🕳️ Missing: none

### Integration verification (worktree-local, iter-2 head)
- [x] `ruff check` on eval module + retrieve_eval.py — clean
- [x] `mypy` on eval module + retrieve_eval.py — Success: no issues found in 8 source files
- [x] `pytest` on 5 affected test files (retrieval_eval_baseline_grep + api_v1_retrieve_eval + ci_run_eval_gate + retrieval_eval_runner + retrieval_eval_metrics) — 87 passed
- [x] `scripts/ci/run_eval_gate.py --compare-baseline /tmp/does-not-exist.json` — exit 1 with `GATE FAILED: baseline missing` on stderr (B4 manual verification)
- [x] `scripts/ci/run_eval_gate.py --compare-baseline ci/eval-baseline.json` — exit 0, `GATE PASSED` (regression-free)
- [x] Go CI on PR (golangci-lint + go test) — SUCCESS
- [x] Helm / Semgrep / TruffleHog / DCO / Retrieval-eval-gate — SUCCESS
- ⚠️ Python CI on PR — FAILURE (pre-existing broken-main from PR #458: `tests/test_db_broadcast_override.py` imports `BroadcastOverride` from `meho_backplane.db.models` but the ORM class was never added in #458). Verified by reading the failing log (collect-time `ImportError`, `pytest -x` stops there, no other tests run) + confirmed `BroadcastOverride` absent from `origin/main:backend/src/meho_backplane/db/models.py`. NOT introduced by iter-2; blocks merge until a follow-up fix on main lands the ORM class or reverts the test. Not this PR's responsibility.

Items the author may pick up in a future PR. CodeRabbit's iter-2 review surfaced these on top of the iter-1 punch list; all are real but none are merge blockers under the team's all-comments-resolved policy.

- **AuditLog test queries lack `ORDER BY`** (`backend/tests/test_api_v1_retrieve_eval.py:339, 413, 449`) — three tests load every `AuditLog` row with `session.execute(select(AuditLog))`, filter in Python, then take `audit_rows[-1]`. SQL row order without `ORDER BY` is undefined; the tests work today because SQLite happens to return insertion order. The pattern existed in iter-1 (lines 339, 449) and iter-2 added one more occurrence (line 413, the new `test_eval_route_baseline_request_still_writes_audit`). On a Postgres-backed test run the `audit_rows[-1]` selection could flake. Fix: `select(AuditLog).where(AuditLog.operator_sub == "<sub>").order_by(AuditLog.occurred_at.desc())` per CodeRabbit's diff. CodeRabbit tagged this 🟠 Major / ⚡ Quick win; agree with the assessment.
- **Subprocess helper in `test_ci_run_eval_gate.py:48-55` has no `timeout=`** — `subprocess.run([sys.executable, GATE_SCRIPT, *args], ...)` without a wall-clock cap means a wedged gate script could hang CI indefinitely. The script under test now has its own grep timeout (B2), so the practical risk is low, but adding `timeout=30` + a `subprocess.TimeoutExpired` catch is a one-line defense-in-depth.
- **`_build_perfect_retrieve_fn` return annotation** (`scripts/ci/run_eval_gate.py:73-74`) — declares `-> Awaitable[list[RetrievalHit]]` and suppresses the return-value mismatch with `# type: ignore[return-value]`. The function returns an async callable, not an Awaitable instance. Correct annotation is `Callable[..., Awaitable[list[RetrievalHit]]]`; that removes the type:ignore.
- **`test_eval_route_baseline_request_still_writes_audit` doesn't assert `eval_all.assert_not_awaited()`** — the companion 501 test on line 355 (`test_eval_route_baseline_request_returns_501`) does patch `eval_all` with `AsyncMock` and asserts it wasn't called, but the audit-row test omits that guard. Both tests exercise the same 501 path; pinning the no-runner behaviour on both would prevent a regression that quietly re-enabled the runner.
- **Test redaction-positive assertion** (`backend/tests/test_retrieval_eval_baseline_grep.py:286-291`) — the new test asserts `secret_query not in combined_text` but doesn't assert `query_len` / `query_sha256` *are* present. If a future refactor removes the redaction fields entirely (no log emission at all), the test still passes. Minor hardening; the redaction contract is still pinned by the corresponding production code.
- **`contextlib.suppress(Exception)` in `baseline_grep.py:282`** — the drain after `proc.kill()` swallows `Exception` (very wide). The comment correctly explains why (kernel may have already reaped the child between kill() and communicate()), but a narrower `(BrokenPipeError, ConnectionResetError, OSError)` would be tighter. Non-blocking; matches asyncio guidance well enough.
- **Python CI broken-main blocker** — Per `findings` above, the `BroadcastOverride` ImportError comes from PR #458 (G6.3-T1, commit f340d50). Needs a separate fix on main. Tracked.

- Adopted as adjacent: 5 (3 inline + 2 nitpick / outside-diff)
- Disputed: 0
- Out of scope: 0
- CodeRabbit's review profile is CHILL; the two `🟠 Major / ⚡ Quick win` findings are real test-isolation concerns surfaced after the iter-1 dance. No code path is broken today; both fit the "adjacent polish for a follow-up PR" framing this skill uses for unresolved ⚠️ majors.

### Docs
- [x] `docs/codebase/backend.md` — eval row updated to describe 501-on-baseline, fail-loud gate, grep timeout, redacted log fields. Reasonable.

- Audit context bound **before** the 501 gate is the right move — rejection still records operator intent. The skill's incident-postmortem hook (#368) is honored.
- `errNoBackplaneConfigured` with `Unwrap()` + composable `errors.Is(err, auth.ErrConfigNotFound)` is idiomatic Go 1.13+ error wrapping; the typed wrapper carries a user-facing message while letting downstream classifiers still pattern-match the sentinel.
- B2's `kill()` → `with contextlib.suppress(Exception): await proc.communicate()` is the canonical asyncio "don't leak zombies on timeout" pattern; the comment correctly notes the kernel-already-reaped race.
- B4's docstring update + three behavioral tests is the right shape — operator-facing exit-code spec moves in lockstep with the script's behavior, and the docstring contract is itself pinned by `test_run_eval_gate_docstring_mentions_missing_baseline_exit_code`.

### Verdict rationale

All seven iter-1 findings closed cleanly. Iter-2's new CodeRabbit findings are real test-isolation polish items, not blockers — they don't break any acceptance criterion or any integration path. Per this skill's Phase 6: "Unresolved ⚠️ majors do not auto-bump to REQUEST_CHANGES." The Python CI red is exclusively the pre-existing broken-main from PR #458 (verified at collect-time; `pytest -x` stops on the BroadcastOverride ImportError before any iter-2 tests run). Go CI is green (B1 closed). Helm / Semgrep / TruffleHog / DCO / Retrieval-eval-gate all green.

**APPROVE** — iteration 2 delivers the closure the punch list asked for; the adjacent items can land in a follow-up polish PR.

Issue #595, opened 2026-05-17
Task: G9.2-T3 annotate/unannotate service + §6 conflict detection + traversal exclusion + audit/broadcast


damir-topic:
Parent initiative: #364
Depends-on: #593, #594

## Target

Ship the annotate/unannotate **service layer** with the §6 conflict-resolution rules, the §3 unannotate safety rule, and the §10 audit + broadcast integration. This is the load-bearing task of G9.2 — it is what makes "k8s SA → Vault role → namespace" knowable, and its conflict semantics are what make a wrong annotation recoverable.

## Blast radius

This task modifies the **hot read path**: all three traversal verbs' recursive CTEs (`find_dependents`/`find_dependencies`/`find_path`) gain a `properties->>'superseded_by' IS NULL` guard so superseded auto edges drop out of every blast-radius query. Regression risk: a malformed guard silently drops *non-superseded* edges from every traversal. It also modifies the refresh write path (marker preservation, below). Both are in scope and must be regression-tested.

## Files touched

- **Add** `backend/src/meho_backplane/topology/annotate.py` — `annotate_edge(...)` / `unannotate_edge(...)` async service functions + `resolve_edge(...)` lookup helper + conflict-detection logic.
- **Modify** `backend/src/meho_backplane/topology/query.py` — add a `properties->>'superseded_by' IS NULL` guard to the edge-pulling CTEs: reverse traversal `query.py:192`, forward traversal `query.py:240`, and the `bi_edge` CTE in `_PATH_SQL` — **both** legs, the forward half at `query.py:418` and the reversed half at `query.py:422` (NOT `query.py:442`, which is the `WHERE w.hops < :max_hops` recursive hop-bound — wrong site; missing the reversed `bi_edge` leg leaves superseded edges walkable backwards).
- **Modify** `backend/src/meho_backplane/topology/refresh.py` — `_reconcile_edges` at `refresh.py:339` today does `existing_edge.properties = dict(hint.properties)`, a **wholesale overwrite** that erases a `superseded_by`/`conflicts_with` marker on the next probe. Change the edge-update path to **merge, not replace**, for the reserved conflict-marker keys (`superseded_by`, `conflicts_with`) when refreshing an auto edge's properties from a hint. Without this the sticky-supersede invariant is unimplementable.
- **Modify** `backend/src/meho_backplane/topology/__init__.py` — export the two service functions.

## Service contract

[...]

Issue #594, opened 2026-05-17
Task: G9.2-T2 Public resolve_node(tenant,name,kind?) helper — extract topology anchor resolution for non-target nodes


damir-topic:
Parent initiative: #364

## Target

Add a public, reusable `resolve_node(session, tenant_id, name, kind=None)` helper for the topology module so the annotate/unannotate verbs (T3) and `list-edges` (T4) can resolve an edge endpoint by name (optionally pinned by `kind`) within a single tenant.

This is **net-new resolution behavior, not a pure extraction.** G9.1's query verbs apply only an *ambiguity* check inline (`_assert_anchor_unambiguous`, `query.py:271` — it raises `AmbiguousNodeError` on a multi-kind bare name but **returns `None`, never fetches a row, and does NOT error on not-found**: a name resolving to zero kinds is a silent no-op today). `resolve_node` additionally (a) fetches and returns the `GraphNode` row and (b) raises a brand-new `NodeNotFoundError` on no match. Neither behavior exists in the codebase today (there is no `NodeNotFoundError` anywhere). The annotation flow also needs this for **non-target** nodes (a curated edge can reference a `vault-role` or `keycloak-realm` node that is not a managed target).

## Files touched

- **Add** `backend/src/meho_backplane/topology/resolvers.py` — `resolve_node(...)` returning the `GraphNode` row (id, kind, name); raises `AmbiguousNodeError` when a bare name maps to >1 `kind` in the tenant and no `kind` was given; raises a new `NodeNotFoundError` (→ HTTP 404 / structured tool error at the call layer) when nothing matches. Tenant scoping is mandatory and non-optional.
- **Add** `NodeNotFoundError` — new exception (no equivalent exists today).
- **Relocate** `AmbiguousNodeError` from `topology/query.py` (currently defined there and in its `__all__` at `query.py:98`) into `resolvers.py`, and **re-import it back into `query.py`** so existing `from meho_backplane.topology.query import AmbiguousNodeError` call sites keep working (blast-radius: grep all importers before/after).
[...]

Issue #364, opened 2026-05-13
Initiative: G9.2 Curated cross-system edges + annotation flow — closed v0.2 edge-kind vocabulary + meho topology annotate/unannotate/list-edges + conflict resolution (decision #6, ~30% curated half)


damir-topic:
Parent goal: #220

## Scope

Ship the **curated cross-system edges + annotation flow** — the ~30% of the topology graph that auto-discovery cannot infer per [decision #6](docs/planning/v0.2-decisions.md). Locks the **typed edge-kind vocabulary** (the load-bearing modeling decision called out in decision #6's body), ships the `meho topology annotate` / `unannotate` / `list-edges` verbs for tenant-admins, handles **conflict resolution** when a curated edge contradicts (or is contradicted by) an auto-discovered one, and writes one audit row per annotation. The annotation surface is what makes "k8s SA → Vault role → namespace" knowable — auto-discovery sees the two ends, not the semantic edge between them.

**Area:** extends [backend/src/meho_backplane/topology/](backend/src/meho_backplane/topology/) (the module G9.1 stands up), new API routes for the annotation verbs on `/api/v1/topology/edges/*`, CLI verbs under `cli/internal/cmd/topology/annotate.go`, MCP tools registered against G0.5's registry, no new tables (extends `graph_edge` from G9.1).

The work items grouped under this Initiative:

1. **Edge-kind vocabulary lock — closed v0.2 enum.** G9.1 ships the auto-discoverable subset (`runs-on`, `mounts`, `routes-through`, `belongs-to`). G9.2 extends with the **curated-only edge kinds** that auto-discovery cannot infer:
   - `authenticates-via` — principal → identity-provider node (`k8s-sa-foo` → `vault-role-bar` → `keycloak-realm-baz`). The canonical example from consumer-needs.md L258.
   - `depends-on` — cross-system functional dependency (`service-X` → `database-Y`, where neither side knows about the other in its own probe output).
   - `replicates-to` — operator-asserted replication relationship between two storage / database nodes.
   - `backed-up-by` — operator-asserted backup relationship.
   - `routes-via` — operator-asserted network path through an intermediary (e.g., `vm-A` → `firewall-X` → `vm-B` when the probes only see point-to-point reachability).
   - `policy-binds` — RBAC / policy attachment that crosses connector boundaries (e.g., `kubernetes-namespace-prod` → `vault-policy-prod-read`).
[...]

PR #650, opened 2026-05-18
feat(topology): G9.2-T4 list_edges tenant-scoped filterable read helper (#596)

### LINKED ISSUE #596: Task: G9.2-T4 list_edges read helper — tenant-scoped filterable edge listing + --conflicts view
Parent initiative: #364
Depends-on: #593, #594

## Target

Ship the `list_edges` read helper backing `meho topology list-edges` (§4) — a tenant-scoped, filterable edge listing with a dedicated `--conflicts` view that surfaces only edges carrying a `conflicts_with` marker (the recoverability surface for a wrong annotation).

## Files touched

- **Modify** `backend/src/meho_backplane/topology/query.py` — add `list_edges(session, tenant_id, *, kind=None, source=None, from_ref=None, to_ref=None, conflicts_only=False, limit, offset)` alongside the traversal verbs; import `resolve_node` from wherever T2/#594 landed it. Returns a bounded list of edge summaries: `id`, `from` (name+kind), `to` (name+kind), `kind`, `source`, `properties` (note/evidence/conflict markers), `last_seen`.
- **Modify** `backend/src/meho_backplane/topology/__init__.py` — export `list_edges`.

## Current state

`query.py` has traversal verbs (`find_dependents`/`find_dependencies`/`find_path`) but **no flat edge-listing primitive**. Edges are only reachable via graph walks today. The unique/index set on `graph_edge` (`graph_edge_tenant_endpoints_kind_idx`, `graph_edge_tenant_from_idx`, `graph_edge_tenant_to_idx`) supports efficient tenant-scoped filtered listing.

## Desired state

- `list_edges(tenant)` → all edges in the tenant, joined to both endpoint nodes for human-readable `from`/`to`, bounded by `limit`/`offset`.
- Filters compose: `kind=` (one of the 10 `GraphEdgeKind`), `source='curated'|'auto'`, `from_ref`/`to_ref` (resolved via `resolve_node` #594), `conflicts_only=True` (rows where `properties->'conflicts_with'` is a non-empty array).
- Result is tenant-scoped unconditionally — no cross-tenant leakage even with crafted filters.
- `from_ref`/`to_ref` resolve via `resolve_node` (#594); an ambiguous bare name (multiple kinds) raises the same `AmbiguousNodeError` contract the traversal verbs use (caller passes a kind-qualified ref to disambiguate). A ref that resolves to no node yields an **empty result**, not an error.

## Acceptance criteria

- [ ] Unit test on a seeded fixture: unfiltered list returns all tenant edges; `source='curated'` excludes auto rows; `kind=` filter restricts correctly; `from_ref`/`to_ref` filter to incident edges.
- [ ] `conflicts_only=True` returns exactly the edges with a non-empty `properties.conflicts_with` array and nothing else (seeded against a conflict fixture).
- [ ] Cross-tenant isolation: a tenant-A call never returns a tenant-B edge regardless of filter combination.
- [ ] Result ordered by a stable total key (`ORDER BY last_seen DESC, id`); `limit`/`offset` paginate deterministically against that key — asserted by a two-page fixture test that reassembles to the unpaged set with no gaps or duplicates.
- [ ] CI green; `ruff` + `mypy` clean.

## Out of scope

- The REST `GET /api/v1/topology/edges` route + role gating — T5 (this is the service primitive it calls).
- CLI rendering / table formatting — T6.
- The MCP `query_topology(kind="edges")` facet — T7 (it calls this same helper).
- Conflict *creation* — T3 writes the markers; this task only reads them.

## References

- Parent Initiative: [#364](https://github.com/evoila/meho/issues/364) §4 (`meho topology list-edges [--kind] [--source] [--from] [--to] [--conflicts]`).
- Prerequisites: [#593](https://github.com/evoila/meho/issues/593) (kind enum for validating `kind=`), [#594](https://github.com/evoila/meho/issues/594) (`resolve_node` for `from_ref`/`to_ref`).
- Precedent to copy:
  - [backend/src/meho_backplane/topology/query.py:377](backend/src/meho_backplane/topology/query.py#L377) — `find_dependencies` raw-SQL + named-bind + tenant-scope + bound pattern.
- Framework docs: PostgreSQL 16 JSON functions (`jsonb_array_length`, `->`, containment) — https://www.postgresql.org/docs/16/functions-json.html .
- Best-practices anchor (hard rule #11): Python best-practices (per `implement-issue-slim`'s audit pack) — bounded queries by default, fully-parameterized SQL (named binds, nothing interpolated — mirror the `text()` pattern in `query.py`), tenant scoping non-optional.

damir-topic:
## Summary

- Add `list_edges(session, tenant_id, *, kind, source, from_ref, to_ref, conflicts_only, limit, offset)` to `meho_backplane.topology.query` — the tenant-scoped filter-composable flat edge-listing service primitive backing the G9.2 `list-edges` surface (T5 REST, T6 CLI, T7 MCP all dispatch through it).
- Joins both endpoint nodes so `from_endpoint` / `to_endpoint` carry `(id, kind, name)` without a second round trip; `properties` bag is deep-frozen like `TopologyNode.properties` so conflict markers can't leak via mutation.
- `from_ref`/`to_ref` go through `resolve_node` (#594): unresolved → empty list (not an error); ambiguous bare name → `AmbiguousNodeError`. `conflicts_only=True` returns exactly the rows whose `properties.conflicts_with` JSONB is a non-empty array — the recoverability surface #595 will populate.
- Stable pagination via `ORDER BY last_seen DESC NULLS LAST, id`; soft-deleted edges (`last_seen IS NULL`) excluded by default.

## Test plan

- [x] `ruff check` — clean
- [x] `mypy backend/src/meho_backplane/topology/` — clean
- [x] Targeted regression suite (116 tests across topology query/schema/refresh/scheduler + api/v1/topology + targets/discover + connectors_topology + list_edges import-smoke / validation) — passed locally in 84s
- [ ] Integration tests in `backend/tests/integration/test_topology_list_edges.py` — Docker-gated skip locally; runs in CI against the `pgvector/pgvector:pg16` testcontainer. Covers: unfiltered listing; `source=` / `kind=` filters; `from_ref`/`to_ref` restriction + unresolved → empty + ambiguous → raise; `conflicts_only=True` returns exactly the non-empty-array rows; cross-tenant boundary holds against every filter combo; two-page sweep reassembles to the unpaged set with no gaps or duplicates.

## Out of scope

- REST route `GET /api/v1/topology/edges` + role gating — T5.
- CLI verb `meho topology list-edges` rendering — T6.
- MCP `query_topology(kind='edges')` facet — T7.
- Annotation write path + conflict-detection that populates `properties.conflicts_with` — T3 (#595).

## Coupling with #595 (parallel-implementation note)

#595 is implementing the write side (annotate/unannotate + conflict-detection) in parallel. The `conflicts_only` predicate doesn't depend on a schema change from #595 — `properties` is already JSONB and the `jsonb_typeof(...) = 'array' AND jsonb_array_length(...) > 0` guard is safe whether the key is absent, NULL, an empty array, or a non-array value. When #595 lands and starts writing the markers, this filter becomes operationally useful with no further changes here.

The two tasks share `query.py` for the file-level diff; if #595 also modifies it (it adds a `superseded_by` exclusion guard to the traversal CTEs), merge order is the orchestrator's call. Re-using `resolve_node` and the existing `properties` JSONB column keeps the surface area minimal.

Closes #596

## Summary by CodeRabbit

* **New Features**
  * Added edge listing capability with filtering options for kind, source, and endpoint references; includes conflict detection support.

* **Documentation**
  * Updated topology documentation to reflect the new edge listing API and associated data models.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/650?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/meho_backplane/topology/query.py
+++ b/backend/src/meho_backplane/topology/query.py
@@ -89,14 +89,22 @@ from uuid import UUID
 
 from sqlalchemy import text
 from sqlalchemy.engine import Row
+from sqlalchemy.ext.asyncio import AsyncSession
 
 from meho_backplane.auth.operator import Operator
 from meho_backplane.db.engine import get_sessionmaker
 from meho_backplane.topology.resolvers import (
     AmbiguousNodeError,
+    NodeNotFoundError,
     _collect_distinct_kinds,
+    resolve_node,
 )
-from meho_backplane.topology.schemas import TopologyNode, TopologyPath
+from meho_backplane.topology.schemas import (
+    TopologyEdge,
+    TopologyEdgeEndpoint,
+    TopologyNode,
+    TopologyPath,
+)
 
 # ``AmbiguousNodeError`` was historically defined in this module and is
 # part of the G9.1 read-half public surface; Task #594 (G9.2-T2) moved
@@ -111,6 +119,7 @@ __all__ = [
     "find_dependencies",
     "find_dependents",
     "find_path",
+    "list_edges",
 ]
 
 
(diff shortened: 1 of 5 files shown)


damir-topic:
@coderabbitai review

backend/src/meho_backplane/topology/query.py
```diff
@@ -545,3 +554,271 @@ async def find_path(
...
+      AND e.last_seen IS NOT NULL
+      AND (CAST(:kind AS text) IS NULL OR e.kind = :kind)
+      AND (CAST(:source AS text) IS NULL OR e.source = :source)
+      AND (CAST(:from_node_id AS uuid) IS NULL OR e.from_node_id = :from_node_id)
+      AND (CAST(:to_node_id AS uuid) IS NULL OR e.to_node_id = :to_node_id)
+      AND (
+          NOT :conflicts_only
+          OR (
+              jsonb_typeof(e.properties -> 'conflicts_with') = 'array'
+              AND jsonb_array_length(e.properties -> 'conflicts_with') > 0
```
damir-topic, line 630: **[⚠️ M1]** — `conflicts_only` predicate relies on Postgres AND short-circuit (not guaranteed)

**What:** Lines 150-156 use `jsonb_typeof(e.properties -> 'conflicts_with') = 'array' AND jsonb_array_length(e.properties -> 'conflicts_with') > 0` as a guard. The PG16 docs (https://www.postgresql.org/docs/16/sql-expressions.html §4.2.14) explicitly state "the order of evaluation of subexpressions is not defined" and "Boolean expressions (AND/OR/NOT combinations) in those clauses can be reorganized in any manner allowed by the laws of Boolean algebra." If the planner ever evaluates the right side first on a row where `properties.conflicts_with` is a non-array, non-null scalar, `jsonb_array_length` raises `cannot get array length of a non-array`.

**Why it matters:** Today this is safe in practice — the column is only written by G9.2-T3 (#595, not yet merged) which writes arrays only, and the test suite covers absent-key, empty-array, and non-empty-array cases against real `pgvector/pgvector:pg16`. The hazard is future-T3 (or a manual UPDATE, or a buggy connector landing a scalar marker) creating a row where this clause is evaluated and raises mid-listing — surfacing as a 500 on `GET /api/v1/topology/edges` and a hard error in the CLI/MCP fronts. Not a blocker because no such write path exists today; flagged as defensive hardening.

**Suggested fix:** Wrap in a `CASE` to force evaluation order — same idiom the PG docs recommend for exactly this pattern:

```suggestion
      AND (
          NOT :conflicts_only
          OR (
              CASE
                  WHEN jsonb_typeof(e.properties -> 'conflicts_with') = 'array'
                  THEN jsonb_array_length(e.properties -> 'conflicts_with') > 0
                  ELSE FALSE
              END
          )
      )
```

The PG docs note `CASE` can defeat optimization, but here the predicate is gated behind `NOT :conflicts_only` (so the planner skips it on the common path) and marker rows are a tiny minority — the optimizer impact is negligible.

**Source:** https://www.postgresql.org/docs/16/sql-expressions.html §4.2.14.

damir-topic (commented):




**Anchor task:** #596 — G9.2-T4 `list_edges` read helper
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `d720c2a`
**AI reviewer (CodeRabbit):** received in ~180s (4 actionable findings, all minor/test-tightening — none block correctness)

### Acceptance criteria
- [x] Unit/integration test: unfiltered listing returns all tenant edges; `source='curated'` excludes auto rows; `kind=` filter restricts correctly; `from_ref`/`to_ref` filter to incident edges — covered in `test_list_edges_{unfiltered,source_filter,kind_filter,from_ref,to_ref}_*`.
- [x] `conflicts_only=True` returns exactly the edges with a non-empty `properties.conflicts_with` array — covered in `test_list_edges_conflicts_only_returns_marker_rows` (fixture exercises absent-key, empty-array, and reciprocal non-empty cases).
- [x] Cross-tenant isolation: tenant-A call never returns tenant-B edge — covered in `test_list_edges_tenant_boundary_holds`. SQL enforces `e.tenant_id = :tenant_id` unconditionally; no filter combination can leak across.
- [x] Stable pagination via `ORDER BY last_seen DESC NULLS LAST, id`; two-page sweep reassembles — covered in `test_list_edges_pagination_is_deterministic` (current fixture is 6 rows so the prefix-only assertion happens to be equivalent to full-set; flagged below as `m3` for future-proofing).
- [x] `ruff` + `mypy` clean — verified locally (`backend/src/meho_backplane/topology/`).

### Project-board alignment
|---|---|---|---|---|---|
| #596 | In Review (PR open) | `Closes #596` | yes | yes | close on merge |

Parent Initiative #364 §4 (`list-edges` surface) — this PR ships the substrate primitive only. T5 (REST), T6 (CLI), T7 (MCP) remain. Sibling PR #651 (T3 #595 write-side) is the announced parallel branch; both touch `topology/__init__.py`, `topology/query.py`, `docs/codebase/topology.md` — whichever merges second rebases, as the orchestrator flagged.

### Predicted vs actual diff
- ✅ `backend/src/meho_backplane/topology/query.py` — `list_edges` + `_LIST_EDGES_SQL` + `_resolve_ref_or_empty` + `_row_to_edge` + limit/offset constants
- ✅ `backend/src/meho_backplane/topology/__init__.py` — re-exports + docstring section
- ✅ `backend/src/meho_backplane/topology/schemas.py` — `TopologyEdge` + `TopologyEdgeEndpoint`, deep-freeze discipline matching `TopologyNode`
- ✅ `backend/tests/integration/test_topology_list_edges.py` — 3 collection-time + 10 Docker-gated covering all acceptance criteria
- ✅ `docs/codebase/topology.md` — entry-point block + key-type tables

No undisclosed scope. No missing predicted changes.

### Integration verification
- [x] `ruff check src/meho_backplane/topology/ tests/integration/test_topology_list_edges.py` — clean
- [x] `ruff format --check` — 7 files already formatted
- [x] `mypy src/meho_backplane/topology/` — `Success: no issues found in 6 source files`
- [x] `pytest tests/integration/test_topology_list_edges.py` — 3 passed, 10 skipped (Docker-gated, expected on no-Docker sandbox)
- [x] `pytest tests/ -k topology` regression sweep — 128 passed, 33 skipped (Docker-gated), no regressions
- [⚠️ not run] `pytest` against `pgvector/pgvector:pg16` testcontainer — Docker not available locally; queued in CI ("Python (integration testcontainers)" still pending at review time). The fixture shape is well-formed; CodeRabbit's pre-merge checks accepted the suite. K8s e2e flake on main is known and treated as expected.
- [x] Tenant-scope correctness — verified by reading `_LIST_EDGES_SQL`: `WHERE e.tenant_id = :tenant_id` is unconditional; `from_node_id`/`to_node_id` casts narrow only, never widen. Cross-tenant leakage impossible by construction.
- [x] `resolve_node` integration — `_resolve_ref_or_empty` correctly maps `NodeNotFoundError → None → empty list`, lets `AmbiguousNodeError` propagate. Matches `find_dependents`/`find_path` contract.
- [x] Deep-freeze discipline — `TopologyEdge._freeze_properties` mirrors `TopologyNode._freeze_properties` exactly (same `_deep_freeze` helper, same `MappingProxyType`/`tuple` recursion, same `field_serializer` thaw on egress).

- ⚠️ M1 — `conflicts_only` predicate relies on Postgres AND-clause short-circuit which the docs explicitly do not guarantee; safe today, brittle to future T3 writes — see inline.
- 💡 m1 — unfiltered-set test assertion is self-referential (RHS includes `edges - known`), so it's effectively `{e.id for e in edges} == {e.id for e in edges}` — see inline (also flagged by CodeRabbit).
- 💡 m2 — `conflicts_only` test asserts truthiness of `cw` rather than `isinstance(cw, (list, tuple)) and len(cw) > 0` — see inline (also flagged by CodeRabbit). Low practical risk because the substrate filter only returns arrays.
- 💡 m3 — pagination test asserts a prefix match (`unpaged[:len(paged_ids)]`); coincidentally exact today (fixture is 6, two pages of 3) but future-fragile — see inline (also flagged by CodeRabbit).
- 💡 m4 — `docs/codebase/topology.md` lines 19-26 (Read half — entry points) and 28-42 (Resolver) describe two different session-ownership conventions without an explicit "see also `list_edges` for the same caller-owned shape" note (also flagged by CodeRabbit). Cosmetic.
- 👏 p1 — Excellent acceptance-criterion-per-test mapping; the test file's module docstring lays out the coverage matrix directly against the issue body. The `_resolve_ref_or_empty` short-circuit comment explains why CAST-IS-NULL-OR semantics would silently broaden the listing if not handled — that kind of pre-emptive correctness note is what makes the helper safe.
- 👏 p2 — `_DEFAULT_EDGE_LIMIT` / `_MAX_EDGE_LIMIT` substrate guard is the right defensive layer — the REST cap (T5) is the operator-facing surface, but a CLI/MCP/REPL caller hitting `list_edges` directly cannot accidentally request a billion rows.

- Adopted: 3 (m1, m2, m3 — same findings, recorded for the punch list)
- Adopted with caveat: 1 (m4 — the doc nit is real but cosmetic; the existing read-half block isn't *contradicted* by `list_edges`, just doesn't cross-reference)
- Disputed: 0
- Out of scope: 0

### Docs
- [x] `docs/codebase/topology.md` — updated with `list_edges` entry-point block (lines 44-68) and `TopologyEdge` / `TopologyEdgeEndpoint` key-type tables (lines 123-147). Verified accurate against code.

### Why APPROVE despite M1
Postgres planner reordering of AND clauses is a documented language-spec concern, not an active bug. The column write path is gated by T3 (#595, not yet merged) and T3 writes only proper arrays. The fixture exercises absent-key + empty-array + non-empty-array cases against real `pgvector/pgvector:pg16` and the suite is green. The defensive-hardening fix (wrap in `CASE WHEN jsonb_typeof = 'array' THEN jsonb_array_length > 0 ELSE FALSE END`) belongs in a follow-up if T3 ever opens a non-array write surface — not as a blocker on a read-only helper that has zero failure cases in any current code path.

PR #671, opened 2026-05-18
fix(ci): restore main CI green — Go gofmt/revive (G9.2 CLI) + k8s e2e meta-tool target contract (#580)

damir-topic:
## Summary

`evoila/meho` `main` was red on two CI jobs, blocking a clean v0.2.1 tag. This is a direct, un-ticketed CI-unbreak: two deterministic regressions that landed on `main` (the DCO admin-bypass is what let them through — this PR ships a real `Signed-off-by` trailer so it passes DCO without a bypass, deliberately). Two separate commits, one per regression.

### Regression 1 — Go (`golangci-lint` job), trivial hygiene

The `Go (golangci-lint + go test)` job (golangci-lint **v1.64.8** / Go **1.23**, `run --path-prefix=cli`) failed on three findings introduced by **#656 (G9.2-T6)** and **#667 (G9.2-T8)**:

- `cli/internal/cmd/topology/list_edges.go:85` — `revive` `unused-parameter`: the cobra `RunE` closure declared `args` but never used it. Renamed `args` → `_` (matches the package convention: siblings that don't consume positional args use `_`; the ones that do use `args[0]`).
- `cli/internal/cmd/topology/bulk_import.go:89` — `gofmt`: doc-comment + struct-tag alignment.
- `cli/internal/cmd/topology/e2e_test.go:403` — `gofmt`: struct-literal field alignment.

Formatting + the one parameter rename only — **no behavioral change**.

### Regression 2 — Python (`Python (integration testcontainers)` job): k8s e2e meta-tool target contract (#580)

`backend/tests/integration/test_connectors_k8s_e2e.py::test_call_operation_about_dispatches_through_meta_tool` failed deterministically:

```
AttributeError: '_K3sTarget' object has no attribute 'get'
  at backend/src/meho_backplane/operations/meta_tools.py:455
```

[...]

PR #580, opened 2026-05-17
feat(connectors): G3.2-T6 K8s CLI alias verbs + k3d E2E acceptance + onboarding doc

damir-topic:
## Summary

- **CLI alias verbs** — new `meho k8s …` verb tree under `cli/internal/cmd/k8s/`, pre-baking `connector_id="k8s-1.x"` on the existing `POST /api/v1/operations/call` route. Covers every registered K8s op (about, ls, namespace.list, node.list, pod.{list,info}, deployment.{list,info}, service.list, ingress.list, configmap.{list,info}, event.list, logs) with the kubectl-style flag shape operators expect (`--namespace` XOR `--all-namespaces`, `--label-selector` / `--field-selector` / `--limit` / `--continue-token`, `--tail` / `--since` / `--previous` / `--container` for logs). Mirrors the `cli/internal/cmd/vault/` (G3.3-T6 PR #568) shape; replaces the consumer's `kubectl-vcf.sh` read workflows.
- **k3d E2E acceptance harness** — `backend/tests/integration/test_connectors_k8s_e2e.py` exercises the agent meta-tool flow (`search_operations` + `call_operation`) against a real k3s container through the real G0.6 dispatcher and a real Postgres audit store. Pins descriptor lookup on the canonical `("k8s","1.x","k8s")` triple, end-to-end round-trip through every op, and synchronous audit-row write per CLAUDE.md postulate 7. CI's `Python (integration testcontainers)` job picks it up; `MEHO_TEST_K3S_IMAGE` added to `.github/workflows/ci.yml` for Harbor-proxy pulls.
- **Operator-facing onboarding doc** — `docs/cross-repo/kubernetes-onboarding.md` is the operator cookbook for retiring `kubectl-vcf.sh` (kubeconfig-in-Vault staging recipe, per-verb table, agent meta-tool walkthrough, `kubectl-vcf.sh` ↔ `meho k8s` migration table, troubleshooting). Mirrors `docs/cross-repo/vault-onboarding.md` (G3.3-T8 PR #573).

[...]

PR #672, opened 2026-05-18
fix(test): deterministic running-pod selection in k8s e2e logs test (#580)

damir-topic:
## Summary

`backend/tests/integration/test_connectors_k8s_e2e.py::test_dispatch_logs_against_running_pod` flakes (and now fails deterministically on `main` after #671) because it selects the **first** pod from `k8s.pod.list` against `kube-system` blindly and calls `k8s.logs` on it. On a freshly-booted k3s container that first pod is frequently still `Pending` / `ContainerCreating`, so `k8s.logs` raises a generic kube `ApiException` ("container … is waiting to start: ContainerCreating") wrapped by the handler as `connector_error: ApiException` — which is neither the `status=="ok"` branch nor the structured `MultiContainerAmbiguityError` (`"container"`/`"ambig"`) branch the test asserts. The assertion `assert "container" in result.error or "ambig" in result.error.lower()` then fails on `result.error == 'connector_error: ApiException'`.

This makes the test name ("against_running_pod") a lie — it never ensured the pod was actually running with a ready container.

## Root cause

- `backend/tests/integration/test_connectors_k8s_e2e.py:772` (`test_dispatch_logs_against_running_pod`) — `pod_name = listing.result["rows"][0]["name"]` picks the first pod with no readiness check.
- Latent since the test was introduced (G3.2 **#580**, same file). `pytest -x` short-circuited on an earlier failure until **#671** fixed the meta-tool target-contract regression and let CI reach this test, unmasking the flake.

## Fix approach

Poll `k8s.pod.list` through the **same `dispatch()` path** (registered op only — the row already carries `status` (phase string) and `ready` ("`<ready>/<total>`"), per `backend/src/meho_backplane/connectors/kubernetes/ops_workload.py:249` `pod_row`) until a `kube-system` pod is `phase == "Running"` with **every** container Ready and at least one container. Bounded to 90 s with a 3 s sleep between attempts; `pytest.skip` only if no ready pod appears in time, matching the existing skip style in this file. k3s always brings up coredns / metrics-server / local-path-provisioner, so a ready pod deterministically appears well within the budget. No new production op, no raw kube-client reach-around.

## FORBIDDEN alternative — explicitly NOT taken

The negative-path assertion was **not** weakened. A correctly-selected Running pod still lands in exactly one of the two pre-existing branches (ok payload + audit row, or the structured multi-container-ambiguity error), and a bare `ApiException` / `connector_error: ApiException` is still **not** accepted — the test still fails loudly if `k8s.logs` genuinely errors against a healthy running pod. Both post-`k8s.logs` branches are untouched.

No production code under `backend/src/meho_backplane/` was touched; the handler erroring on a not-ready pod is correct behavior. Diff is one test file (+56 / −12).

## Verification

- `ruff check` + `ruff format --check` on the touched test path: clean.
- CI's mypy gate is `uv run mypy src/` (`[tool.mypy] files = ["src"]`) — it does not analyze test files, and no `src/` file changed; `uv run mypy src/` is clean (`no issues found in 180 source files`).
- This sandbox happened to have Docker available, so the k3s testcontainer actually ran: the target test and all 17 tests in the file **pass**. The k3d integration job is still the authoritative CI-verified gate; the change is reasoned to be deterministic because k3s' default control-plane pods reliably reach Running+Ready well inside the 90 s budget, and the selection predicate only accepts `Running` + all-containers-Ready (`X/Y`, `X==Y`, `X>=1`).
- Scope is limited to this one test's pod-selection; the other tests in the file are unchanged and still pass.

Refs #580, #671. Un-ticketed CI-unbreak fix — no `Closes`.

## Summary by CodeRabbit

* **Tests**
  * Improved stability of Kubernetes integration tests by enhancing pod selection logic in E2E test scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/672?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/tests/integration/test_connectors_k8s_e2e.py
+++ b/backend/tests/integration/test_connectors_k8s_e2e.py
@@ -65,8 +65,10 @@ in the ``Python (integration testcontainers)`` job.
 
 from __future__ import annotations
 
+import asyncio
 import contextlib
 import os
+import time
 from collections.abc import AsyncIterator, Iterator
 from dataclasses import dataclass
 from datetime import UTC, datetime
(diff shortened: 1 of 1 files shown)


damir-topic:
@coderabbitai review

PR #645, opened 2026-05-18
feat(migrate): G5.3-T2 frontmatter parser + scope-suggestion table + machine-local opt-out (#609)

ddzafic:
## Summary

- Adds `cli/internal/migrate/scan.go`: `ResolveSourceDir` (3-step resolution including `$HOME/.claude/projects/<sanitized-cwd>/memory` fallback), `ScanDir` (non-recursive `*.md` walker), and `MemoryFile` struct with frontmatter parser, `BodySHA256` (hex SHA-256 of post-frontmatter body for T5 idempotency), and `MachineLocalOptOut` detection via whitespace-tolerant `<!-- meho:machine-local -->` HTML comment scan.
- Adds `cli/internal/migrate/suggest.go`: exported `Scope*` constants (single source of truth for T4 picker and T5 submit) and `SuggestScope` deterministic mapping table — no AI inference, per Initiative #375 and `ai_engineering_best_practices §Architecture-&-boundaries`.
- No cobra import in `cli/internal/migrate/`; `go vet` clean; `gofmt -d` empty; uses only `gopkg.in/yaml.v3` v3.0.1, already a direct dep.

## Test plan

- [x] `go vet ./internal/migrate/...` — clean
- [x] `gofmt -d ./internal/migrate/` — empty
- [x] `go test ./internal/migrate/...` — 27 passed: well-formed frontmatter for each `type:` value; missing frontmatter → `FrontmatterOK=false` + `ParseWarning` set, no error; malformed YAML (bad indentation, unterminated `---`) → same graceful degradation; `<!-- meho:machine-local -->` present/absent toggles `MachineLocalOptOut`; whitespace-tolerant variant; `BodySHA256` stable across calls and changes when body changes; hex shape (64 lowercase chars); `ScanDir` non-recursive + `.md`-only; `ResolveSourceDir` explicit/`CLAUDE_PROJECT_DIR`/home-dir fallback with leading-dash assertion.
- [x] `SuggestScope` full mapping table including `tenantConfigured` branch for `type: project` and unknown-type fallback.
- [x] `go test ./...` — all 16 packages pass (no regressions).
- [x] `python3 .claude/hooks/code-quality.py --diff` — exit 0

## Out of scope

- Machine-local *heuristic* (path/hostname/username regex sweep) — T3's `machinelocal.go`
- Interactive picker and dry-run formatter — T4 (#610)
- Submission / `source_id` construction — T5 (#611)
- Sibling tasks T3–T5 (#610–#612)

Closes #609

Issue #610, opened 2026-05-18
Task: G5.3-T3 Machine-local detector (path/hostname/username heuristic + per-pattern tests)


damir-topic:
Parent initiative: #375 (G5.3 Laptop-local memory migration UX)
Parent goal: #216
Depends on: #608 (the `cli/internal/migrate/` package scaffold).
Independent of T2 (parser) — can be built and merged in parallel.

## Current state

After T1, `cli/internal/migrate/` exists but has no machine-local
heuristic. Nothing flags memory bodies that contain laptop-specific
paths, local-only hostnames, or the operator's username — so a naive
migration would leak `/Users/<name>/...` strings into `tenant`-scoped
memory where they mislead other operators (the failure mode #375's
Motivation section calls out).

## Desired state

`cli/internal/migrate/machinelocal.go` provides pure, cobra-free,
unit-testable detection:

- `DetectMachineLocal(body string) MachineLocalResult` where
  `MachineLocalResult` is `{ Flagged bool; Matches []MachineLocalMatch }`
  and `MachineLocalMatch` is `{ Category string; Sample string }`
  (category = which rule fired, sample = the matched substring,
  truncated to ~80 chars, for the T4 picker badge tooltip).
- Rules (each a distinct `Category`):
  - `unix-home`: `/Users/<seg>/` and `/home/<seg>/` (POSIX home
    prefixes; `<seg>` = one path segment).
  - `windows-home`: `C:\Users\<seg>\` (case-insensitive drive letter;
    backslash-escaped in the regex).
  - `tilde-home`: a `~/` token at a path boundary (start of string or
    preceded by whitespace/quote) to avoid matching `~/` inside prose
    like "approx ~/2 days".
  - `local-hostname`: hostnames ending `.local` or `.lan`, and the
    bare tokens `localhost`, `host.docker.internal`.
  - `operator-username`: the basename of `os.UserHomeDir()` appearing
    `>= 3` times in the body as a whole-word match (guard against a
    common-word username inflating false positives — document the
    3-occurrence threshold rationale in a comment; it mirrors the
    Initiative's spec).
- Whole-word / boundary matching uses Go `regexp` with explicit
  anchors — no third-party regex dep.
- `Flagged` is true iff `len(Matches) > 0`. The function is the
  single source of truth the T4 picker calls to pre-mark a file
  "Skip — machine-local"; `--include-machine-local` (T4) overrides
  the pre-mark, it does not change detection.

## Acceptance criteria

[...]

Issue #608, opened 2026-05-18
Task: G5.3-T1 Scaffold meho migrate command tree + charm.land/huh/v2 dependency


damir-topic:
Parent initiative: #375 (G5.3 Laptop-local memory migration UX)
Parent goal: #216

## Current state

The `meho` CLI has no `migrate` command and no interactive-prompt
dependency. Every verb tree today is a package under
`cli/internal/cmd/<name>/` exposing a `NewRootCmd()` constructor that
`cli/internal/cmd/root.go` registers via `root.AddCommand(...)` *before*
`registerDynamicSubcommands(root)` (so the backplane manifest cannot
shadow built-in verb names). `cli/go.mod` (module
`github.com/evoila/meho/cli`, `go 1.23.0`) pins `spf13/cobra v1.10.2`
and `gopkg.in/yaml.v3 v3.0.1` as direct deps. There is no TUI / form
library.

## Desired state

The `migrate` command tree exists and is wired into the root, with the
`memory` subcommand present as a skeleton (flags parsed, no flow logic
yet — that lands in T2–T5). The `charm.land/huh/v2` form library is a
declared dependency. `cli/README.md` records the dependency
justification and that this pick locks Charm in as the CLI's TUI stack.

**Layout (codebase-aligned — supersedes the path sketch in #375 which
predates this convention):**

- Add `cli/internal/cmd/migrate/migrate.go` — `package migrate`,
  `NewRootCmd() *cobra.Command` returning the `migrate` parent command
  (`Use: "migrate"`, short/long describing it as the laptop→backplane
  migration verb tree), with the `memory` subcommand attached.
- Add `cli/internal/cmd/migrate/memory.go` — `newMemoryCmd()
  *cobra.Command` (`Use: "memory"`, `SilenceUsage: true`), all flags
  registered, `RunE` a stub returning a "not yet implemented" error so
  the skeleton compiles and is discoverable. Flags:
  - `--source <path>` (string, default empty → resolver added in T2)
  - `--dry-run` (bool)
  - `--non-interactive` (bool)
  - `--include-machine-local` (bool)
  - `--mark-migrated` (bool)
- Add an empty `cli/internal/migrate/` package (`doc.go` with the
  package comment only) for the pure-logic helpers T2–T5 fill in. This
  package is imported only by `cli/internal/cmd/migrate/`, never by
  `cli/internal/cmd` root, so it introduces no import cycle (the
  cycle the kb.go header documents only applies to a *shared* HTTP
[...]

PR #644, opened 2026-05-18
feat(migrate): G5.3-T3 machine-local detector — path/hostname/username heuristic (#610)

ddzafic:
## Summary

- Adds `DetectMachineLocal(body string, homeFn HomeDirFunc) MachineLocalResult` to `cli/internal/migrate/` — pure stdlib, no cobra import, no third-party regex dep.
- Five heuristic rules (unix-home, windows-home, tilde-home, local-hostname, operator-username) flag laptop-specific content so the T4 picker can pre-mark files "Skip — machine-local".
- `HomeDirFunc` seam makes the operator-username rule deterministic in CI — tests inject a fake home path rather than depending on the runner's `$HOME`.
- `MachineLocalMatch.Sample` is truncated to ≤80 runes for the T4 badge tooltip.

## Test plan

- [x] `cd cli && go test ./internal/migrate/...` — all 21 subtests pass
- [x] `go vet ./...` — clean
- [x] `gofmt -d` — empty diff
- [x] `go test ./...` — all 18 packages pass (0 failures)
- [x] `python3 .claude/hooks/code-quality.py --diff` — exit 0
- [x] unix-home: `/Users/alice/x` hit; `/Usersalice` miss
- [x] windows-home: `C:\Users\bob\` hit; no trailing backslash miss
- [x] tilde-home: `~/foo` at line start hit; `approx ~/3 weeks` mid-prose miss
- [x] local-hostname: `db.local` / `box.lan` / `localhost` / `host.docker.internal` hits; `mylocaldb` miss
- [x] operator-username: 3 occurrences hit; 2 occurrences miss; error from homeFn miss; nil homeFn no panic
- [x] Each `MachineLocalMatch.Sample` non-empty and ≤80 runes

## Out of scope

- T4 picker wiring (`--include-machine-local` flag, per-file UI) — that is #612.
- Frontmatter parser / scanner (`scan.go`) — that is #609 (parallel sibling).
- `<!-- meho:machine-local -->` HTML-comment opt-out — that is T2 (#609).

Closes #610

## Release Notes

* **New Features**
  * Added machine-local pattern detection to identify home directory paths, usernames, and hostname indicators in document text.

* **Tests**
  * Added comprehensive test coverage for detection patterns and edge cases.

* **Documentation**
  * Updated module documentation for the new detection capability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/644?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #680, opened 2026-05-19
feat(migrate): G5.3-T4 interactive huh picker + --dry-run/--non-interactive (#611)

### LINKED ISSUE #609: Task: G5.3-T2 Frontmatter parser + scope-suggestion table + machine-local opt-out comment
Parent initiative: #375 (G5.3 Laptop-local memory migration UX)
Parent goal: #216
Depends on: #608 (the `cli/internal/migrate/` package scaffold).

## Current state

After T1, `cli/internal/migrate/` exists but is empty. There is no
code that locates the operator's laptop-local memory directory, parses
the YAML frontmatter Claude Code writes, suggests a backplane scope, or
honours an opt-out marker. `gopkg.in/yaml.v3 v3.0.1` is already a
**direct** dependency in `cli/go.mod` — no new dependency is needed for
frontmatter parsing.

## Desired state

`cli/internal/migrate/scan.go` and `cli/internal/migrate/suggest.go`
provide pure, cobra-free, unit-testable logic:

**`scan.go`:**
- `ResolveSourceDir(override string) (string, error)` — if `override`
  is non-empty, use it; else if `$CLAUDE_PROJECT_DIR` is set use
  `$CLAUDE_PROJECT_DIR/memory/`; else resolve
  `<os.UserHomeDir()>/.claude/projects/<sanitized-abs-cwd>/memory/`
  (the layout the repo CLAUDE.md auto-memory section documents — note
  the base is `$HOME/.claude`, **not** `os.UserConfigDir()` which is
  `~/Library/Application Support` on macOS / `~/.config` on Linux).
  `<sanitized-abs-cwd>` is the **absolute** working directory with
  every `/` replaced by `-`, so the leading `/` becomes a **leading
  `-`** (e.g. `/Users/x/repos/meho` → `-Users-x-repos-meho`); the
  leading separator is **retained**, not dropped. State the exact
  transform in a doc comment and pin it with a test.
- `ScanDir(dir string) ([]MemoryFile, error)` — walk `*.md` files
  (non-recursive — the memory dir is flat). For each:
  - Parse leading YAML frontmatter delimited by `---` lines using
    `gopkg.in/yaml.v3` into `name string`, `description string`,
    `type string`, `metadata` passthrough. A file with **no** or
    **malformed** frontmatter is not an error: return it with
    `FrontmatterOK=false`, `Body` = whole file, and a `ParseWarning`
    string; the caller (T4 picker) shows the warning and still lets
    the operator migrate it.
  - `BodySHA256` — hex SHA-256 of the body bytes (post-frontmatter),
    used by T5 for the idempotency `source_id`.
  - `MachineLocalOptOut bool` — true iff the body contains the literal
    HTML comment `<!-- meho:machine-local -->` anywhere (operator's
    explicit per-file opt-out; case-sensitive, whitespace-tolerant
    inside the comment delimiters).
- `MemoryFile` struct fields: `Path`, `Name`, `Description`, `Type`,
  `Body`, `BodySHA256`, `FrontmatterOK`, `ParseWarning`,
  `MachineLocalOptOut`.

**`suggest.go`:**
- `SuggestScope(m MemoryFile, tenantConfigured bool) string`:
  - `type: user` or `type: feedback` → `"user"`.
  - `type: project` → `"user×tenant"` if `tenantConfigured`, else
    `"user"`.
  - `type: reference` → `"user"`.
  - any other / empty `type` → `"user"` (safest default; T4 surfaces
    it as overridable).
- The returned string is one of the canonical scope tokens the T4
  picker and T5 submission share; define them as exported consts
  (`ScopeUser`, `ScopeUserTenant`, `ScopeUserTarget`, `ScopeTenant`,
  `ScopeTarget`) so there is one source of truth.

## Acceptance criteria

- [ ] `cd cli && go test ./internal/migrate/...` passes with table-driven tests covering: (a) well-formed frontmatter for each `type:` value; (b) missing frontmatter → `FrontmatterOK=false`, `ParseWarning` set, no error; (c) malformed YAML (unterminated `---`, bad indentation) → same graceful degradation; (d) `<!-- meho:machine-local -->` present/absent toggles `MachineLocalOptOut`; (e) `BodySHA256` stable across calls and changes when the body changes.
- [ ] `SuggestScope` unit test asserts the full mapping table including the `tenantConfigured` branch for `type: project` and the unknown-`type` fallback.
- [ ] `ResolveSourceDir("/explicit/path")` returns it verbatim; `$CLAUDE_PROJECT_DIR` (when set) wins over the home-dir fallback; with neither set a test pins the exact transform on a known abs path (`/a/b/c` → `<home>/.claude/projects/-a-b-c/memory/`), asserting the leading `-` is retained.
- [ ] No cobra import in `cli/internal/migrate/`; `go vet` clean; `gofmt -d` empty.

## Out of scope

- The machine-local *heuristic* (path/hostname/username regex sweep) —
  that is T3's `machinelocal.go`. T2 only handles the explicit
  `<!-- meho:machine-local -->` opt-out comment.
- The interactive picker and dry-run formatter — T4.
- Submission / `source_id` construction — T5 (T2 only exposes
  `BodySHA256`).
- Inherits all Out-of-scope bullets of Initiative #375.

## References

- Parent: Initiative #375, Goal #216.
- Claude Code auto-memory file shape (`name`/`description`/`type` +
  body) — documented in the repo CLAUDE.md auto-memory section; v0.2
  reads the same shape.
- yaml.v3 v3.0.1 — https://pkg.go.dev/gopkg.in/yaml.v3 (already a
  direct dep in `cli/go.mod`; the Initiative's "likely transitively
  present, confirm" note is resolved: it is a declared direct dep).
- Precedent for cobra-free, table-driven internal logic +
  `httptest`-free unit tests: `cli/internal/auth/config_test.go`,
  `cli/internal/cmd/kb/add_test.go`.
- Best practices: **ai_engineering_best_practices**
  §Architecture-&-boundaries (scope suggestion is a deterministic
  table, never an AI inference — every promotion stays
  operator-initiated, per #375 motivation and consumer-needs.md §G5
  "Out of scope: AI-suggested promotion").

### LINKED ISSUE #611: Task: G5.3-T4 Interactive huh picker + --dry-run/--non-interactive paths
Parent initiative: #375 (G5.3 Laptop-local memory migration UX)
Parent goal: #216
Depends on: #608 (scaffold), #609 (scanner + scope suggestion), #610 (machine-local detector).

## Current state

After T1–T3, `cli/internal/migrate/` can scan a directory, parse
frontmatter, suggest a scope, detect machine-local content, and honour
the opt-out comment — but `meho migrate memory`'s `RunE` is still the
T1 stub. There is no interactive picker and no `--dry-run` /
`--non-interactive` behaviour.

## Desired state

`cli/internal/cmd/migrate/memory.go`'s `RunE` builds and runs an
interactive `huh` form over the scanner output, or takes the
non-interactive / dry-run path. Form-building logic lives in a
cobra-free helper (e.g. `cli/internal/migrate/picker.go` returning a
`*huh.Form` plus a results accessor) so it is testable without a TTY.

**Per-file group** (one `huh.NewGroup` per `MemoryFile`,
`huh.NewForm(groups...)`):

- A `huh.NewNote()` header: filename + first 200 chars of the body
  (newlines stripped, truncated) + a machine-local badge when
  `DetectMachineLocal().Flagged` **or** `MachineLocalOptOut`.
- Action `huh.NewSelect[string]()`: options `Migrate to <suggested
  scope>` / `Migrate to a different scope…` / `Skip (keep
  laptop-local)` / `Skip — machine-local`. Default selection: the
  suggested-migrate option normally; `Skip — machine-local` when the
  file is flagged or opted-out (unless `--include-machine-local`, which
  flips the default back to migrate but keeps the badge).
- Scope picker as a **separate `huh.NewGroup` hidden via
  `(*huh.Group).WithHideFunc`** (NOTE: `WithHideFunc` is a *Group*
  method, not a field method — the Initiative body's
  "`.WithHideFunc(...)` when the operator chooses" wording must be
  realised as a hidden group bound to the action value). Shown only
  when the action is "different scope…". Its `huh.NewSelect` options
  are filtered by the operator's role read from local auth claims:
  `user` and `user×tenant`/`user×target` always offered;
  `tenant` and `target` offered only if the operator has
  `tenant_admin` (mirrors G5.2 #374's RBAC); `user×tenant` requires a
  resolvable tenant, `user×target` a resolvable target, else those
  options are omitted with an inline note.
- Slug `huh.NewInput()` defaulting to the file basename lowercased
  with non-`[a-z0-9-]` collapsed to `-`; `Validate` rejects anything
  not matching `^[a-z0-9][a-z0-9-]*$`.
- A body-edit `huh.NewText()` in a hidden group, revealed when the
  operator picks an "Edit before migrating" action variant; pre-filled
  with the body so machine-local snippets can be stripped pre-send.
- A final `huh.NewConfirm()` group: "Migrating N entries (M skipped).
  Proceed?" with the per-entry summary.

**Accessibility:** call `form.WithAccessible(...)` driven by an
env var the CLI defines — `MEHO_ACCESSIBLE` (non-empty → accessible
mode). huh v2 has **no** built-in `HUH_ACCESSIBLE` variable
(accessibility is the `WithAccessible(bool)` method only); the
Initiative's "Out of scope" reference to `HUH_ACCESSIBLE=1` is
inaccurate — document `MEHO_ACCESSIBLE` in the help text instead.

**`--dry-run`:** run the full scan + suggestion + detection, skip the
form, and print one JSON envelope **per line** (machine-parseable,
`jq`-friendly) of exactly what T5 would POST — `{scope, slug, body,
metadata, source_id}` with `source_id` =
`"laptop-migration/<sha256-prefix>"` (first 12 hex chars of
`BodySHA256`; the prefix length is fixed here and reused verbatim by
T5). No network call.

**`--non-interactive`:** skip the form; migrate only `type: user` and
`type: feedback` at their suggested scope; for any `project` /
`reference` entry, print a refusal line to stderr naming the file and
exit non-zero if any such entry was present and not skipped (the
operator must run interactively for those). Honours the T2 opt-out
comment and T3 flag (machine-local files are skipped, not migrated,
in non-interactive mode regardless of `--include-machine-local`).

The actual POST is T5; T4's non-interactive/interactive paths produce
the same in-memory `[]SubmitPlan` that T5 consumes, so T4 wires a
`submitFn` seam (defaulting to a not-yet-implemented stub) that T5
replaces.

## Acceptance criteria

- [ ] `cd cli && go test ./internal/migrate/... ./internal/cmd/migrate/...` passes, covering: (a) `--dry-run` emits one valid JSON object per migrate-selected entry, parseable line-by-line, with the `source_id` prefix scheme; (b) `--non-interactive` includes `user`/`feedback`, excludes `project`/`reference` with a stderr refusal and non-zero exit when such entries exist; (c) role filter omits `tenant`/`target` scope options without `tenant_admin`; (d) slug validation regex accepts `daily-routine`, rejects `Daily Routine`, `-x`, `x_y`; (e) the picker builder produces the expected group/field tree from a fixture file set (assert structure, no TTY).
- [ ] `go run ./cmd/meho migrate memory --dry-run --source testdata/...` prints envelopes and makes no HTTP request (assert via a failing/refusing fake transport in the test).
- [ ] `--help` documents `MEHO_ACCESSIBLE`.
- [ ] `go vet` clean; `gofmt -d` empty; CI green.

## Out of scope

- The HTTP POST, spinner, retry loop, post-login nudge, marker file,
  and the cross-repo doc — all T5. T4 stops at producing the
  `[]SubmitPlan` and the dry-run print.
- Conflict-resolution UI for same-slug-different-machine — explicitly
  out per #375 (v0.2 = update-on-change in T5).
- Inherits all Out-of-scope bullets of Initiative #375.

## References

- Parent: Initiative #375 (work items 3 + 5), Goal #216.
- Sibling RBAC source: #374 (G5.2) — the `tenant_admin` gate the scope
  filter mirrors.
- huh v2.0.3 — `huh.NewForm`/`NewGroup`/`NewSelect`/`NewInput`/
  `NewText`/`NewNote`/`NewConfirm`; `(*huh.Group).WithHideFunc(func()
  bool) *huh.Group` for conditional groups; `Form.WithAccessible(bool)`
  (no env var built in). Docs: https://pkg.go.dev/charm.land/huh/v2 ,
  https://github.com/charmbracelet/huh .
- Precedent for cobra-thin RunE delegating to a testable helper +
  output discipline (prompts → stdout, errors → stderr,
  machine-readable JSON mode): `cli/internal/cmd/login.go`,
  `cli/internal/output/format.go`.
- Best practices: **ai_engineering_best_practices** §Safety +
  §Architecture-&-boundaries (every migrate decision is an explicit
  operator choice on familiar content; machine-local + project/reference
  default to the conservative path; no AI-suggested bulk promotion —
  mirrors consumer-needs.md §G5 "Out of scope").

ddzafic:
Closes #611
Closes #609

Depends-on: #608, #610

> **Note on #609 provenance.** PR #645 (T2) was merged to `feat/608-…` rather than `main`, so T2's code (`scan.go` / `suggest.go` / `scan_test.go` / `suggest_test.go`) only reached `main` via this PR's diff. Adding `Closes #609` here so issue auto-close fires on merge; the unusual provenance is recorded in this note.

## What this does

Wires `meho migrate memory`'s `RunE` stub with the full G5.3-T4 picker flow:

**Interactive path** (default): builds a `huh.NewForm` over the `ScanDir` output. Per-file group structure:
- `huh.NewNote` header (filename + first 200 chars of body + machine-local badge)
- `huh.NewSelect[string]` action: Migrate to `<suggested-scope>` / Migrate to a different scope… / Migrate (edit body first) / Skip (keep laptop-local) / Skip — machine-local
- Hidden `huh.NewGroup` scope picker (via `WithHideFunc`) with role-aware options: `user`/`user×tenant`/`user×target` always present; `tenant`/`target` only when `IsTenantAdmin=true`
- `huh.NewInput` slug (validates `^[a-z0-9][a-z0-9-]*$`)
- Hidden `huh.NewGroup` body-edit text, revealed for the "edit" action variant
- Final `huh.NewConfirm`

`MEHO_ACCESSIBLE=1` activates `form.WithAccessible(true)`.

**`--dry-run`**: prints one JSON envelope per migrate-selected file (default plan, no form, no network). `source_id = "laptop-migration/<12 hex chars of SHA-256 body>"` — same constant used by T5 POST for idempotency.

**`--non-interactive`**: migrates only `type:user` and `type:feedback`; prints stderr refusal + exits non-zero for `project`/`reference`; always skips machine-local regardless of `--include-machine-local`.

**`submitFn` seam**: `newMemoryCmdWithSubmit(fn)` lets T5 inject the real backplane POST without touching `RunE`.

## New files
- `cli/internal/migrate/picker.go` — `BuildForm`, `SubmitPlan`, `FinalizeSkip`, `DefaultPlan`, `slugFromPath`, scope/action option builders
- `cli/internal/migrate/picker_test.go` — slug, validate, BuildForm structure, role-filtered scopes, FinalizeSkip, DefaultPlan
- `cli/internal/migrate/marker.go` — `TouchMarker`/`MarkerExists` stubs (full impl T5)
- `cli/internal/cmd/migrate/memory_test.go` — dry-run envelope + source_id, non-interactive filter, machine-local skip, empty-dir guard

## Test plan
- [x] `go build ./...` clean
- [x] `go test ./internal/migrate/... ./internal/cmd/migrate/...` — all pass
- [ ] CI green (depends on T1/T2/T3 merging; branch based on T3 branch)

## Summary by CodeRabbit

* **New Features**
  * Memory migration CLI: preview JSON (--dry-run) with stable source IDs, non-interactive submission, and interactive form-based selection; optional include-machine-local flag and migration-complete marker touch.
* **Bug Fixes / Behavior**
  * Machine-local entries are skipped by default; parsing is tolerant of malformed frontmatter with warnings.
* **Tests**
  * Extensive unit/integration tests for scanning, parsing, ID stability, picker logic, and CLI modes.
* **Documentation**
  * CLI docs updated to cover migration flows, dry-run output, and interactive picker.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/680?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cli/internal/cmd/migrate/memory.go
+++ b/cli/internal/cmd/migrate/memory.go
@@ -5,10 +5,12 @@ package migrate
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
 
+	"charm.land/huh/v2"
 	"github.com/spf13/cobra"
 
 	"github.com/evoila/meho/cli/internal/migrate"
@@ -90,7 +92,7 @@ func newMemoryCmdWithSubmit(submitFn SubmitFn) *cobra.Command {
 				plans, refused := nonInteractivePlans(files, opts)
 				if len(refused) > 0 {
 					for _, f := range refused {
-						fmt.Fprintf(os.Stderr,
+						fmt.Fprintf(cmd.ErrOrStderr(),
 							"meho migrate memory: skipping %s (type %q requires interactive review)\n",
 							f.Path, f.Type,
 						)
@@ -99,21 +101,27 @@ func newMemoryCmdWithSubmit(submitFn SubmitFn) *cobra.Command {
 				if err := submitFn(plans); err != nil {
 					return err
 				}
-				if markMigrated {
-					_ = migrate.TouchMarker(dir) // non-fatal in T4 stub
-				}
 				if len(refused) > 0 {
 					return fmt.Errorf("meho migrate memory: %d entries require interactive review", len(refused))
+				}
+				if markMigrated {
+					_ = migrate.TouchMarker(dir) // non-fatal in T4 stub
 				}
 				return nil
 			}
 
 			// ── Interactive path ──────────────────────────────────────
 			form, plans := migrate.BuildForm(files, opts)
-			accessible := os.Getenv("MEHO_ACCESSIBLE") != ""
+			accessible := os.Getenv("MEHO_ACCESSIBLE") == "1"
 			form = form.WithAccessible(accessible)
 
 			if err := form.Run(); err != nil {
+				// Operator aborted (Ctrl+C / Esc, or said "No" on the
+				// batch-confirm then aborted) — surface as clean exit.
+				if errors.Is(err, huh.ErrUserAborted) {
+					fmt.Fprintf(cmd.OutOrStdout(), "Migration cancelled.\n")
+					return nil
+				}
 				return fmt.Errorf("picker: %w", err)
 			}
 
--- a/cli/internal/cmd/migrate/memory.go
+++ b/cli/internal/cmd/migrate/memory.go
@@ -4,76 +4,230 @@
 package migrate
 
 import (
-	"errors"
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+	"time"
 
+	"charm.land/huh/v2"
 	"github.com/spf13/cobra"
+
+	"github.com/evoila/meho/cli/internal/migrate"
 )
 
-// newMemoryCmd returns the `meho migrate memory` subcommand. The cobra
-// surface is fully wired — all flags registered, short/long help text
-// describing the future interactive migration flow. The RunE stub returns
-// a "not yet implemented" error so the skeleton compiles and is discoverable
-// via `meho migrate memory --help` without accidentally running a partial
-// flow.
-//
-// Flow logic (local memory file scanner, interactive huh form picker, dry-run
-// preview, backplane submit, mark-migrated) lands in G5.3-T2–T5 (#609–#612).
+// SubmitFn is the seam T5 replaces with the real backplane POST.
+// T4 wires a no-op stub so the command tree compiles and the dry-run
+// path works end-to-end without a network call.
+type SubmitFn func(plans []migrate.SubmitPlan) error
+
+// dryRunEnvelope is the machine-readable shape emitted by --dry-run
+// (one JSON object per line). Matches the T5 POST body exactly.
+type dryRunEnvelope struct {
+	Scope    string         `json:"scope"`
+	Slug     string         `json:"slug"`
+	Body     string         `json:"body"`
+	Metadata map[string]any `json:"metadata"`
+	SourceID string         `json:"source_id"`
+}
+
+// SourceIDPrefix is the number of hex chars taken from the SHA-256 body
+// hash to build a source_id. Shared constant between T4 dry-run output
+// and T5 POST (keeping them bit-for-bit identical ensures --dry-run
+// preview = exactly what would be sent).
+const SourceIDPrefix = 12
+
 func newMemoryCmd() *cobra.Command {
+	return newMemoryCmdWithSubmit(stubSubmitFn)
+}
+
+// newMemoryCmdWithSubmit is the testable variant: callers inject a submitFn.
+func newMemoryCmdWithSubmit(submitFn SubmitFn) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:          "memory",
 		Short:        "Migrate laptop-local memory entries to the backplane",
 		SilenceUsage: true,
 		Long: "Scan a local memory directory (or the default XDG location), " +
 			"present an interactive picker for entries to migrate, preview " +
-			"changes in dry-run mode, and submit selected entries to the " +
-			"backplane knowledge base. Full flow lands in G5.3-T2–T5 " +
-			"(Initiative #375).",
-		RunE: func(_ *cobra.Command, _ []string) error {
-			return errors.New("meho migrate memory: not yet implemented (G5.3-T2–T5)")
+			"changes in --dry-run mode, or run headlessly with " +
+			"--non-interactive.\n\n" +
+			"Set MEHO_ACCESSIBLE=1 for screen-reader-friendly output.",
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			flags := cmd.Flags()
+			source, _ := flags.GetString("source")
+			dryRun, _ := flags.GetBool("dry-run")
+			nonInteractive, _ := flags.GetBool("non-interactive")
+			includeML, _ := flags.GetBool("include-machine-local")
+			markMigrated, _ := flags.GetBool("mark-migrated")
+
+			// ── Resolve source directory ──────────────────────────────
+			dir, err := migrate.ResolveSourceDir(source)
+			if err != nil {
+				return fmt.Errorf("resolve source dir: %w", err)
+			}
+
+			// ── Scan ──────────────────────────────────────────────────
+			files, err := migrate.ScanDir(dir)
+			if err != nil {
+				return fmt.Errorf("scan %s: %w", dir, err)
+			}
+			if len(files) == 0 {
+				fmt.Fprintf(cmd.OutOrStdout(), "No memory files found in %s.\n", dir)
+				return nil
+			}
+
+			opts := migrate.BuildFormOpts{
+				IncludeMachineLocal: includeML,
+				// TenantConfigured / IsTenantAdmin resolved from auth in T5;
+				// T4 leaves these at their zero values (conservative defaults).
+			}
+
+			// ── --dry-run path ────────────────────────────────────────
+			if dryRun {
+				return runDryRun(cmd, files, opts)
+			}
+
+			// ── --non-interactive path ────────────────────────────────
+			if nonInteractive {
+				plans, refused := nonInteractivePlans(files, opts)
+				if len(refused) > 0 {
+					for _, f := range refused {
+						fmt.Fprintf(os.Stderr,
+							"meho migrate memory: skipping %s (type %q requires interactive review)\n",
+							f.Path, f.Type,
+						)
+					}
+				}
+				if err := submitFn(plans); err != nil {
+					return err
+				}
+				if markMigrated {
+					_ = migrate.TouchMarker(dir) // non-fatal in T4 stub
+				}
+				if len(refused) > 0 {
+					return fmt.Errorf("meho migrate memory: %d entries require interactive review", len(refused))
+				}
+				return nil
+			}
+
+			// ── Interactive path ──────────────────────────────────────
+			form, plans := migrate.BuildForm(files, opts)
+			accessible := os.Getenv("MEHO_ACCESSIBLE") != ""
+			form = form.WithAccessible(accessible)
+
+			if err := form.Run(); err != nil {
+				return fmt.Errorf("picker: %w", err)
+			}
+
+			migrate.FinalizeSkip(plans)
+
+			toSubmit := filterMigrate(plans)
+			if len(toSubmit) == 0 {
+				fmt.Fprintf(cmd.OutOrStdout(), "No entries selected for migration.\n")
+				return nil
+			}
+
+			if err := submitFn(toSubmit); err != nil {
+				return err
+			}
+			if markMigrated {
+				_ = migrate.TouchMarker(dir)
+			}
+			return nil
 		},
 	}
 
-	// --source: local memory directory to scan. Empty default defers to
-	// the XDG-aware resolver added in T2 (#609).
-	cmd.Flags().String(
-		"source",
-		"",
-		"path to the local memory directory to scan (default: XDG-resolved in T2)",
-	)
-
-	// --dry-run: preview which entries would be migrated without sending
-	// anything to the backplane.
-	cmd.Flags().Bool(
-		"dry-run",
-		false,
-		"preview entries that would be migrated without submitting",
-	)
-
-	// --non-interactive: skip the huh picker and migrate all discovered
-	// entries that pass the filter criteria.
-	cmd.Flags().Bool(
-		"non-interactive",
-		false,
-		"skip the interactive picker and migrate all discovered entries",
-	)
-
-	// --include-machine-local: include entries tagged as machine-local
-	// (hostname-scoped) which are excluded by default because they rarely
-	// transfer meaningfully across machines.
-	cmd.Flags().Bool(
-		"include-machine-local",
-		false,
-		"include machine-local (hostname-scoped) entries in the migration set",
-	)
-
-	// --mark-migrated: after successful submission, write a sentinel
-	// marker in the source directory so a re-run skips already-migrated
-	// entries.
-	cmd.Flags().Bool(
-		"mark-migrated",
-		false,
-		"write a migration-complete marker after successful submission",
-	)
+	cmd.Flags().String("source", "", "path to the local memory directory to scan (default: XDG-resolved)")
+	cmd.Flags().Bool("dry-run", false, "preview entries that would be migrated without submitting")
+	cmd.Flags().Bool("non-interactive", false, "skip the interactive picker; migrates only user/feedback entries")
+	cmd.Flags().Bool("include-machine-local", false, "include machine-local entries (default: skip them)")
+	cmd.Flags().Bool("mark-migrated", false, "touch the migration-complete marker after successful submission")
 
 	return cmd
 }
+
+// runDryRun prints one JSON envelope per migrate-selected file (default
+// plan, no form) and makes no network call.
+func runDryRun(cmd *cobra.Command, files []migrate.MemoryFile, opts migrate.BuildFormOpts) error {
+	enc := json.NewEncoder(cmd.OutOrStdout())
+	enc.SetEscapeHTML(false)
+	for _, f := range files {
+		plan := migrate.DefaultPlan(f, opts)
+		if plan.Skip {
+			continue
+		}
+		env := dryRunEnvelope{
+			Scope:    plan.Scope,
+			Slug:     plan.Slug,
+			Body:     plan.Body,
+			Metadata: map[string]any{"tags": []string{}},
+			SourceID: sourceID(plan),
+		}
+		if err := enc.Encode(env); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// nonInteractivePlans returns plans for user/feedback files (at their
+// suggested scope) and the refused list (project/reference).
+// Machine-local files are always skipped in non-interactive mode,
+// regardless of IncludeMachineLocal (spec requirement: operator must
+// review machine-local content interactively).
+func nonInteractivePlans(files []migrate.MemoryFile, opts migrate.BuildFormOpts) (plans []migrate.SubmitPlan, refused []migrate.MemoryFile) {
+	safeOpts := opts
+	safeOpts.IncludeMachineLocal = false
+	for _, f := range files {
+		plan := migrate.DefaultPlan(f, safeOpts)
+		if plan.Skip {
+			continue
+		}
+		switch f.Type {
+		case "user", "feedback":
+			plans = append(plans, plan)
+		default:
+			refused = append(refused, f)
+		}
+	}
+	return plans, refused
+}
+
+// filterMigrate returns only the non-skip plans.
+func filterMigrate(plans []migrate.SubmitPlan) []migrate.SubmitPlan {
+	var out []migrate.SubmitPlan
+	for _, p := range plans {
+		if !p.Skip {
+			out = append(out, p)
+		}
+	}
+	return out
+}
+
+// sourceID builds the stable source_id string from a plan's BodySHA256.
+// Uses the first SourceIDPrefix hex chars of the SHA-256 hash.
+func sourceID(plan migrate.SubmitPlan) string {
+	prefix := plan.File.BodySHA256
+	if len(prefix) > SourceIDPrefix {
+		prefix = prefix[:SourceIDPrefix]
+	}
+	return "laptop-migration/" + prefix
+}
+
+// stubSubmitFn is the T4 no-op seam. T5 replaces it with the real
+// backplane POST implementation.
+func stubSubmitFn(plans []migrate.SubmitPlan) error {
+	_ = time.Now() // suppress unused import if huh brings in time
+	if len(plans) == 0 {
+		return nil
+	}
+	scopes := make([]string, 0, len(plans))
+	for _, p := range plans {
+		scopes = append(scopes, fmt.Sprintf("%s (%s)", p.Slug, p.Scope))
+	}
+	return fmt.Errorf("meho migrate memory: submit not yet implemented (T5); would send: %s",
+		strings.Join(scopes, ", "))
+}
+
+// withHuhForm is used in tests to replace the form runner.
+var withHuhForm = func(form *huh.Form) error { return form.Run() }
--- a/cli/internal/cmd/migrate/memory.go
+++ b/cli/internal/cmd/migrate/memory.go
@@ -8,9 +8,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
-	"time"
 
-	"charm.land/huh/v2"
 	"github.com/spf13/cobra"
 
 	"github.com/evoila/meho/cli/internal/migrate"
@@ -156,11 +154,15 @@ func runDryRun(cmd *cobra.Command, files []migrate.MemoryFile, opts migrate.Buil
 		if plan.Skip {
 			continue
 		}
+		md := plan.File.Metadata
+		if md == nil {
+			md = map[string]any{}
+		}
 		env := dryRunEnvelope{
 			Scope:    plan.Scope,
 			Slug:     plan.Slug,
 			Body:     plan.Body,
-			Metadata: map[string]any{"tags": []string{}},
+			Metadata: md,
 			SourceID: sourceID(plan),
 		}
 		if err := enc.Encode(env); err != nil {
@@ -217,7 +219,6 @@ func sourceID(plan migrate.SubmitPlan) string {
 // stubSubmitFn is the T4 no-op seam. T5 replaces it with the real
 // backplane POST implementation.
 func stubSubmitFn(plans []migrate.SubmitPlan) error {
-	_ = time.Now() // suppress unused import if huh brings in time
 	if len(plans) == 0 {
 		return nil
 	}
@@ -228,6 +229,3 @@ func stubSubmitFn(plans []migrate.SubmitPlan) error {
 	return fmt.Errorf("meho migrate memory: submit not yet implemented (T5); would send: %s",
 		strings.Join(scopes, ", "))
 }
-
-// withHuhForm is used in tests to replace the form runner.
-var withHuhForm = func(form *huh.Form) error { return form.Run() }
--- a/cli/internal/migrate/picker.go
+++ b/cli/internal/migrate/picker.go
@@ -272,16 +272,6 @@ func buildScopeOptions(opts BuildFormOpts) []huh.Option[string] {
 	return options
 }
 
-func countMigrate(plans []SubmitPlan) int {
-	n := 0
-	for _, p := range plans {
-		if !p.Skip {
-			n++
-		}
-	}
-	return n
-}
-
 func countMigrateFromPlans(plans []SubmitPlan) int {
 	n := 0
 	for _, p := range plans {
--- a/cli/internal/migrate/picker.go
+++ b/cli/internal/migrate/picker.go
@@ -1,0 +1,306 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 evoila Group
+
+package migrate
+
+import (
+	"fmt"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"charm.land/huh/v2"
+)
+
+// Action tokens for the per-file action select field.
+const (
+	// ActionMigrateSuggested uses the scope SuggestScope returned.
+	ActionMigrateSuggested = "migrate:suggested"
+	// ActionMigrateDifferent shows the scope picker group.
+	ActionMigrateDifferent = "migrate:different"
+	// ActionMigrateEdit shows the body-edit group before submit.
+	ActionMigrateEdit = "migrate:edit"
+	// ActionSkipManual keeps the file laptop-local by operator choice.
+	ActionSkipManual = "skip:manual"
+	// ActionSkipMachineLocal keeps a machine-local flagged file local.
+	ActionSkipMachineLocal = "skip:machine-local"
+)
+
+// SubmitPlan is one entry in the operator's migration decision. T5
+// consumes a []SubmitPlan from the interactive, --dry-run, or
+// --non-interactive paths.
+type SubmitPlan struct {
+	// File is the source MemoryFile.
+	File MemoryFile
+	// Scope is the chosen scope token (one of the Scope* constants).
+	Scope string
+	// Slug is the entry identifier validated against ^[a-z0-9][a-z0-9-]*$.
+	Slug string
+	// Body is the body text to POST (may differ from File.Body if the
+	// operator edited it via the body-edit group).
+	Body string
+	// Skip is true when the operator chose to keep the file laptop-local.
+	Skip bool
+	// Action records the raw action selection so callers can distinguish
+	// between the different migrate/skip variants. Set by BuildForm; also
+	// set to ActionMigrateSuggested by DefaultPlan for non-interactive
+	// and --dry-run paths.
+	Action string
+}
+
+// BuildFormOpts controls scope availability and picker defaults.
+type BuildFormOpts struct {
+	// IsTenantAdmin allows the tenant and target scope options. When false
+	// those options are omitted from the scope picker (the server would
+	// reject them with 403 anyway, but hiding them prevents confusion).
+	IsTenantAdmin bool
+	// TenantConfigured is true when the operator's meho config has a
+	// default tenant slug. Used by SuggestScope for the project type.
+	TenantConfigured bool
+	// TargetConfigured is true when the operator's config has a default
+	// target. Used to enable the user×target scope option.
+	TargetConfigured bool
+	// IncludeMachineLocal overrides the default-to-skip behaviour for
+	// files flagged by DetectMachineLocal or MachineLocalOptOut. The
+	// machine-local badge is still shown; only the default action flips.
+	IncludeMachineLocal bool
+}
+
+// BuildForm creates an interactive huh form over files. The returned
+// plans slice is populated as the operator advances through the form —
+// callers must not read plans until form.Run() returns nil.
+//
+// After form.Run(), call FinalizeSkip(plans) to propagate each plan's
+// Action field into its Skip bool.
+//
+// The form is designed to be run interactively (terminal); for
+// --dry-run and --non-interactive paths use DefaultPlan / NonInteractivePlans.
+func BuildForm(files []MemoryFile, opts BuildFormOpts) (*huh.Form, []SubmitPlan) {
+	plans := make([]SubmitPlan, len(files))
+	// One action string per file — kept outside SubmitPlan so we can
+	// take its address for the Select binding without the struct escaping
+	// to heap unintentionally.
+	actions := make([]string, len(files))
+
+	var groups []*huh.Group
+
+	for i, f := range files {
+		fi := i
+		ff := f
+		ml := DetectMachineLocal(ff.Body, nil)
+		isMachineLocal := ml.Flagged || ff.MachineLocalOptOut
+
+		// Initialise the plan defaults.
+		plans[fi] = SubmitPlan{
+			File:   ff,
+			Scope:  SuggestScope(ff, opts.TenantConfigured),
+			Slug:   slugFromPath(ff.Path),
+			Body:   ff.Body,
+			Action: defaultAction(isMachineLocal, opts.IncludeMachineLocal),
+		}
+		actions[fi] = plans[fi].Action
+
+		// ── Group 1: note header + action select ──────────────────────
+		noteDesc := buildNoteDescription(ff, isMachineLocal)
+		mainGroup := huh.NewGroup(
+			huh.NewNote().
+				Title(filepath.Base(ff.Path)).
+				Description(noteDesc),
+			huh.NewSelect[string]().
+				Title("Action").
+				Options(buildActionOptions(plans[fi].Scope, isMachineLocal)...).
+				Value(&actions[fi]),
+		)
+		groups = append(groups, mainGroup)
+
+		// ── Group 2: scope picker (hidden unless "migrate:different") ──
+		scopeGroup := huh.NewGroup(
+			huh.NewSelect[string]().
+				Title("Scope").
+				OptionsFunc(func() []huh.Option[string] {
+					return buildScopeOptions(opts)
+				}, nil).
+				Value(&plans[fi].Scope),
+		).WithHideFunc(func() bool {
+			return actions[fi] != ActionMigrateDifferent
+		})
+		groups = append(groups, scopeGroup)
+
+		// ── Group 3: slug input (hidden when skipping) ─────────────────
+		slugGroup := huh.NewGroup(
+			huh.NewInput().
+				Title("Slug").
+				Description("Identifier stored in the backplane (^[a-z0-9][a-z0-9-]*$)").
+				Value(&plans[fi].Slug).
+				Validate(validateSlug),
+		).WithHideFunc(func() bool {
+			return strings.HasPrefix(actions[fi], "skip:")
+		})
+		groups = append(groups, slugGroup)
+
+		// ── Group 4: body edit (hidden unless "migrate:edit") ──────────
+		editGroup := huh.NewGroup(
+			huh.NewText().
+				Title("Edit body before migrating").
+				Description("Strip machine-local snippets before sending.").
+				Value(&plans[fi].Body),
+		).WithHideFunc(func() bool {
+			return actions[fi] != ActionMigrateEdit
+		})
+		groups = append(groups, editGroup)
+	}
+
+	// ── Final confirm group ────────────────────────────────────────────
+	migrateCount := countMigrate(plans)
+	var confirmed bool
+	confirmGroup := huh.NewGroup(
+		huh.NewConfirm().
+			TitleFunc(func() string {
+				mc := countMigrateFromActions(actions)
+				sc := len(actions) - mc
+				return fmt.Sprintf("Migrate %d entries (%d skipped). Proceed?", mc, sc)
+			}, &actions).
+			Value(&confirmed),
+	)
+	_ = migrateCount // used via TitleFunc closure
+	groups = append(groups, confirmGroup)
+
+	return huh.NewForm(groups...), plans
+}
+
+// FinalizeSkip propagates each plan's Action field into its Skip bool.
+// Call after form.Run() returns nil.
+func FinalizeSkip(plans []SubmitPlan) {
+	for i := range plans {
+		plans[i].Skip = strings.HasPrefix(plans[i].Action, "skip:")
+		if plans[i].Action == ActionMigrateSuggested {
+			// Re-derive suggested scope in case opts changed after form
+			// construction; for well-formed calls this is a no-op.
+			// (plans[i].Scope was already set to SuggestScope at build time.)
+		}
+	}
+}
+
+// DefaultPlan returns the default SubmitPlan for a file without running
+// the interactive form. Used by --dry-run (all files) and
+// --non-interactive (user/feedback files only).
+func DefaultPlan(f MemoryFile, opts BuildFormOpts) SubmitPlan {
+	ml := DetectMachineLocal(f.Body, nil)
+	isMachineLocal := ml.Flagged || f.MachineLocalOptOut
+	action := defaultAction(isMachineLocal, opts.IncludeMachineLocal)
+	return SubmitPlan{
+		File:   f,
+		Scope:  SuggestScope(f, opts.TenantConfigured),
+		Slug:   slugFromPath(f.Path),
+		Body:   f.Body,
+		Skip:   strings.HasPrefix(action, "skip:"),
+		Action: action,
+	}
+}
+
+// slugFromPath derives a default slug from the file's base name:
+//   - strip the .md extension
+//   - lowercase
+//   - replace any run of characters not in [a-z0-9] with a single "-"
+//   - trim leading/trailing "-"
+//   - if the result is empty or starts with a digit, prepend "entry-"
+func slugFromPath(path string) string {
+	base := filepath.Base(path)
+	base = strings.TrimSuffix(base, filepath.Ext(base))
+	base = strings.ToLower(base)
+	re := regexp.MustCompile(`[^a-z0-9]+`)
+	slug := strings.Trim(re.ReplaceAllString(base, "-"), "-")
+	if slug == "" || (slug[0] >= '0' && slug[0] <= '9') {
+		slug = "entry-" + slug
+	}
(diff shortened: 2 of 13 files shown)


cli/internal/migrate/picker.go
```diff
@@ -0,0 +1,306 @@
...
+		// ── Group 1: note header + action select ──────────────────────
+		noteDesc := buildNoteDescription(ff, isMachineLocal)
+		mainGroup := huh.NewGroup(
+			huh.NewNote().
+				Title(filepath.Base(ff.Path)).
+				Description(noteDesc),
+			huh.NewSelect[string]().
+				Title("Action").
+				Options(buildActionOptions(plans[fi].Scope, isMachineLocal)...).
+				Value(&actions[fi]),
```
damir-topic: **[🛑 Blocker]** — operator's interactive action choice is discarded after `form.Run()`

**What:** the action `Select` binds to the parallel `actions[fi]` slice (`Value(&actions[fi])`), but `plans[fi].Action` is set once at build time (line 99-101) and never re-synced from `actions[fi]`. `FinalizeSkip` (line 173-182) then reads the stale `plans[i].Action`.
**Why it matters:** a file the operator selects "Skip (keep laptop-local)" / "Skip — machine-local" still has `plans.Action == "migrate:suggested"`, so `FinalizeSkip` sets `Skip=false` and `filterMigrate` submits it — laptop-local content the operator explicitly declined is sent to the backplane. This directly violates #611's explicit-operator-consent mandate (ai_engineering §Safety). The inverse (operator chooses to migrate a machine-local file → silently dropped) also holds. No test catches it: nothing drives the form runtime.
**Required change:** bind the action `Select` directly to `&plans[fi].Action`; change the `WithHideFunc`/`TitleFunc` closures (lines 124-126, 136-138, 147-149, 158-162) to read `plans[fi].Action`; delete the `actions` slice and `countMigrateFromActions`. `plans` is fixed-length and never re-sliced, so `&plans[fi].Action` is stable for the form's lifetime (verified against `huh@v2.0.3` `Select.Accessor`/`selectOption`, which honour a pre-set bound value).
**Source:** `charm.land/huh/v2@v2.0.3/field_select.go` `Accessor`/`selectValue`/`selectOption`.

cli/internal/migrate/scan.go
```diff
@@ -0,0 +1,278 @@
+// SPDX-License-Identifier: Apache-2.0
```
damir-topic, line 1: **[🛑 Blocker]** — this file (and `suggest.go`/`scan_test.go`/`suggest_test.go`) is #609's already-"merged" T2 work, absent from `main`

**What:** PR #645 (T2 / #609) was merged with `baseRefName = feat/608-migrate-cmd-huh-v2`, not `main`. Its merge commit `289fd97` is reachable only from `origin/feat/608-migrate-cmd-huh-v2`; `origin/main` HEAD is `a92d48c` (the #644/T3 merge). So `scan.go`/`suggest.go` are **not on `main`**, this PR's diff-vs-`main` carries them (byte-identical to the #645-merged copies — `git diff origin/feat/608-migrate-cmd-huh-v2 pr-680-review -- cli/internal/migrate/scan.go …` is empty), and #680 is `mergeable=false`/`dirty`.
**Why it matters:** merging #680 as-is makes it the de-facto delivery of #609 while #609/#645 already read merged/closed — broken provenance and a project-board inconsistency. The PR also cannot merge cleanly until rebased.
**Required change:** re-land #609 on `main` (re-merge or cherry-pick #645 → `main`), then rebase `feat/611-huh-picker` onto current `main` so #680's diff is T4-only. If the team instead accepts #680 carrying T2, disclose it (`Refs #609` in the PR body) and correct #609/#645 status. Not a T4 code defect, but it blocks this PR.

cli/internal/cmd/migrate/memory.go
```diff
@@ -4,76 +4,230 @@
...
+	for _, f := range files {
+		plan := migrate.DefaultPlan(f, opts)
+		if plan.Skip {
+			continue
+		}
+		env := dryRunEnvelope{
+			Scope:    plan.Scope,
+			Slug:     plan.Slug,
+			Body:     plan.Body,
+			Metadata: map[string]any{"tags": []string{}},
```
damir-topic: **[⚠️ Major]** — `--dry-run` metadata is hardcoded, so the preview is not "what T5 would POST"

**What:** `Metadata: map[string]any{"tags": []string{}}` discards `plan.File.Metadata`, `Type`, `Name`, `Description`.
**Why it matters:** #611's desired state says the `--dry-run` envelope is "exactly what T5 would POST", and the shared `SourceIDPrefix` constant exists precisely to keep preview == real POST bit-for-bit. A hardcoded empty placeholder breaks that contract — operators previewing a migration see metadata that won't match what actually gets sent.
**Required change:** derive `Metadata` from `plan.File.Metadata` (carry the parsed frontmatter, including `type`). If the exact POST metadata schema is genuinely T5's to define, say so explicitly in the PR body and the task, and document that `--dry-run` metadata is provisional — a documented gap, not a silent fake.

cli/internal/migrate/picker_test.go
```diff
@@ -0,0 +1,175 @@
...
+	if form == nil {
+		t.Fatal("BuildForm returned nil form")
+	}
+	// 4 groups per file + 1 confirm group
+	want := len(files)*4 + 1
+	got := len(plans) // plans length equals len(files)
+	if got != len(files) {
+		t.Errorf("plans length = %d; want %d", got, len(files))
+	}
+	_ = want // form.Groups() is not exported; we verify via plans length
```
damir-topic: **[⚠️ Major]** — acceptance criterion (e) is not satisfied: this asserts nothing structural

**What:** the test computes `want := len(files)*4 + 1`, then `_ = want` and asserts only `len(plans) == len(files)`. The group/field tree is never inspected (the comment concedes "form.Groups() is not exported").
**Why it matters:** #611 acceptance criterion (e) explicitly requires "the picker builder produces the expected group/field tree from a fixture file set (assert structure, no TTY)". `len(plans)` is not structure; the criterion is unmet, and the structural regressions that Blocker 1's fix touches would not be caught.
**Required change:** expose a test-observable structure (e.g. return a small group/field descriptor from `BuildForm`, or add an unexported inspector used by the test) and assert the per-file group/field shape (4 groups/file + 1 confirm, hidden-group transitions on action change). Then drop the dead `want`/`_ = want`.

cli/internal/cmd/migrate/memory.go
```diff
@@ -4,76 +4,230 @@
...
+	if len(prefix) > SourceIDPrefix {
+		prefix = prefix[:SourceIDPrefix]
+	}
+	return "laptop-migration/" + prefix
+}
+
+// stubSubmitFn is the T4 no-op seam. T5 replaces it with the real
+// backplane POST implementation.
+func stubSubmitFn(plans []migrate.SubmitPlan) error {
+	_ = time.Now() // suppress unused import if huh brings in time
```
damir-topic: **[💡 Minor]** — `time` imported only to be discarded; misleading comment

**What:** `_ = time.Now() // suppress unused import if huh brings in time` — `time` is imported solely for this no-op. The comment's premise is wrong: an unused import in *this* file is unused regardless of `huh`'s transitive imports.
**Required change:** remove the `time` import and the `_ = time.Now()` line.

cli/internal/migrate/picker.go
```diff
@@ -0,0 +1,306 @@
...
+	confirmGroup := huh.NewGroup(
+		huh.NewConfirm().
+			TitleFunc(func() string {
+				mc := countMigrateFromActions(actions)
+				sc := len(actions) - mc
+				return fmt.Sprintf("Migrate %d entries (%d skipped). Proceed?", mc, sc)
+			}, &actions).
+			Value(&confirmed),
+	)
+	_ = migrateCount // used via TitleFunc closure
```
damir-topic: **[💡 Minor]** — dead `migrateCount` + false comment

**What:** `migrateCount := countMigrate(plans)` then `_ = migrateCount // used via TitleFunc closure`. It is not used in the closure — the `TitleFunc` closure calls `countMigrateFromActions(actions)`. `migrateCount` (and `countMigrate`) are dead here and the comment is false.
**Required change:** remove the dead `migrateCount` line and false comment; reconcile the counting helper as part of Blocker 1's rewrite (a single `Action`-based count over `plans`).

cli/internal/cmd/migrate/memory.go
```diff
@@ -4,76 +4,230 @@
...
+	scopes := make([]string, 0, len(plans))
+	for _, p := range plans {
+		scopes = append(scopes, fmt.Sprintf("%s (%s)", p.Slug, p.Scope))
+	}
+	return fmt.Errorf("meho migrate memory: submit not yet implemented (T5); would send: %s",
+		strings.Join(scopes, ", "))
+}
+
+// withHuhForm is used in tests to replace the form runner.
+var withHuhForm = func(form *huh.Form) error { return form.Run() }
```
damir-topic: **[💡 Minor]** — dead `withHuhForm` seam

**What:** `var withHuhForm = func(form *huh.Form) error { return form.Run() }` is never referenced — `RunE` calls `form.Run()` directly (line 118) and tests inject via `submitFn`, not this var.
**Required change:** remove it, or route the interactive path through it if it was intended as the form-runner test seam (in which case add the test that uses it).

cli/internal/migrate/picker.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// FinalizeSkip propagates each plan's Action field into its Skip bool.
+// Call after form.Run() returns nil.
+func FinalizeSkip(plans []SubmitPlan) {
+	for i := range plans {
+		plans[i].Skip = strings.HasPrefix(plans[i].Action, "skip:")
+		if plans[i].Action == ActionMigrateSuggested {
+			// Re-derive suggested scope in case opts changed after form
+			// construction; for well-formed calls this is a no-op.
```
damir-topic: **[💡 Minor]** — dead empty `if` block in `FinalizeSkip`

**What:** `if plans[i].Action == ActionMigrateSuggested { … }` has a body consisting only of comments — it does nothing.
**Required change:** remove the empty block (fold into Blocker 1's `FinalizeSkip` rewrite).

damir-topic (changes requested):
## /review-pr-slim summary

**Anchor task:** #611 — G5.3-T4 interactive `huh` picker + `--dry-run`/`--non-interactive`
**Surface area:** backend (Go CLI)
**Best-practices packs audited:** ai_engineering (§Safety / §Architecture-&-boundaries — explicit operator consent), Go idiom
**Verdict:** REQUEST_CHANGES

Two blockers (one code, one merge-state), two majors. The dry-run and non-interactive logic is largely sound and well-tested; the **interactive path is broken** and the PR **cannot merge cleanly** because of an upstream merge-target mistake on its T2 dependency.

### Acceptance criteria (#611)
- [x] `--dry-run` emits one JSON object per migrate-selected entry, line-parseable, `laptop-migration/<12 hex>` source_id — verified (`go test ./internal/cmd/migrate/...`, `TestDryRun_*`).
- [x] `--non-interactive` includes `user`/`feedback`, refuses `project`/`reference` to stderr + non-zero exit — verified (`TestNonInteractive_*`).
- [x] role filter omits `tenant`/`target` without `tenant_admin` — verified (`TestBuildScopeOptions_*`).
- [x] slug regex accepts `daily-routine`, rejects `Daily Routine`/`-x`/`x_y` — verified (`TestValidateSlug`).
- [ ] **picker builder produces the expected group/field tree (assert structure, no TTY)** — ❌ not satisfied. `TestBuildForm_GroupCount` computes `want` then `_ = want` and asserts only `len(plans)`. No structural assertion (see inline, Major).
- [x] `--help` documents `MEHO_ACCESSIBLE` — present in `Long`.
- [ ] **`go vet` clean; `gofmt -d` empty; CI green** — `go vet` clean; `gofmt -d` **not empty** (`scan_test.go`); PR is `CONFLICTING`. Both are consequences of Blocker 2, not T4 code (see below).
- [ ] **operator's interactive selections honoured** — ❌ broken (Blocker 1).

### Project-board alignment (project — evoila/meho)
|---|---|---|---|---|---|
| #611 (T4) | In Progress | `Closes #611` | ✅ if rebased | ⚠️ partial (interactive path broken) | blocker — fix wiring, then close on merge |
| #609 (T2) | "merged" via #645 | none | — | 🚫 **#645 merged to `feat/608-…`, not `main`; its code is absent from `main` and carried undisclosed by this PR** | **blocker** — re-land #609 on `main`, correct #609/#645 state |
| #375 (Initiative) | In Progress | parent | — | ⚠️ not completed by this PR (T5 #612 remains) | none |

### Predicted vs. actual diff
- ✅ `picker.go`, `marker.go`, `memory.go`, `memory_test.go`, `picker_test.go`, `docs/codebase/cli.md` — expected T4 surface.
- ⚠️ `picker.go` action wiring — expected but wrong (Blocker 1, inline).
- 🚫 `scan.go` / `suggest.go` / `scan_test.go` / `suggest_test.go` — unexpected vs `main`, byte-identical to #645-merged T2; present only because #645 was merged to the wrong base (Blocker 2, inline on `scan.go`).

### Integration verification (run in `.claude/worktrees/pr-680-review/cli`)
- [x] `go vet ./internal/migrate/... ./internal/cmd/migrate/...` — clean
- [⚠️] `gofmt -d` — **not empty**: `internal/migrate/scan_test.go` (inherited T2 #645 code, not T4-introduced; will need fixing when #609 is re-landed)
- [x] `go test ./internal/migrate/... ./internal/cmd/migrate/...` — all pass (but the suite does **not** exercise the interactive form runtime, so Blocker 1 is uncaught)
- [⚠️ not run] interactive `huh` form exercised in a TTY — not feasible headless; reasoned through against vendored `charm.land/huh/v2@v2.0.3` source (`field_select.go`)

- 🛑 2 blockers — interactive action choices discarded (`picker.go`); T2 (#609) not on `main` / PR carries it undisclosed & is `CONFLICTING` (`scan.go`)
- ⚠️ 2 majors — `--dry-run` metadata hardcoded (`memory.go`); acceptance criterion (e) test asserts nothing structural (`picker_test.go`)
- 💡 4 minors — `time.Now()` import hack (`memory.go`); dead `migrateCount` + false comment (`picker.go`); dead `withHuhForm` (`memory.go`); empty dead `if` in `FinalizeSkip` (`picker.go`)

### Blocker detail

**🛑 1 — interactive operator choices are silently discarded.**
The action `Select` binds to a parallel `actions []string` slice; `plans[].Action` is set once at build time and **never updated from `actions[]`**. After `form.Run()`, `FinalizeSkip` reads the stale `plans[].Action`. Net effect:
- A file the operator selects "Skip (keep laptop-local)" / "Skip — machine-local" is still **submitted** (its default was `migrate:suggested`) — a direct violation of #611's explicit-operator-consent mandate (ai_engineering §Safety: every migrate decision must be an explicit operator choice).
- A machine-local file the operator chooses to migrate is silently **dropped** (default `skip:machine-local` persists).
Uncaught because no test drives the form runtime. Fix: bind the action `Select` to `&plans[fi].Action`; have the `WithHideFunc`/`TitleFunc` closures read `plans[fi].Action`; delete the parallel `actions` slice and `countMigrateFromActions`. (`plans` is fixed-length and never re-sliced, so `&plans[fi].Action` is stable for the form's lifetime — verified against `huh` v2.0.3 `Select.Accessor`/`selectOption`, which honour a pre-set bound value when options exist before `.Value()`/on option-load.)

**🛑 2 — T2 (#609) is not on `main`; this PR carries it undisclosed and cannot merge cleanly.**
`gh pr view 645 --json baseRefName` → `feat/608-migrate-cmd-huh-v2`. #645's merge commit `289fd97` is reachable only from `origin/feat/608-migrate-cmd-huh-v2`; `origin/main` HEAD is `a92d48c` (the #644/T3 merge). So `scan.go`/`suggest.go`/`scan_test.go`/`suggest_test.go` (832 LOC) are **absent from `main`**. This PR's diff-vs-`main` carries them (byte-identical to the #645-merged copies — `git diff origin/feat/608-migrate-cmd-huh-v2 pr-680-review -- …` is empty), `#680` is `mergeable=false` / `dirty`, and merging as-is makes #680 the de-facto delivery of #609 while #609/#645 already read merged/closed — broken provenance + board inconsistency. Required before #680 can land: re-land #609 on `main` (re-merge or cherry-pick #645 → `main`), then rebase `feat/611-huh-picker` onto current `main` so #680's diff is T4-only. If the team instead accepts #680 carrying T2, disclose it (`Refs #609`) and correct #609/#645 status. This is upstream, not a T4 code defect — but it blocks this PR.

### Docs
- [x] `docs/codebase/cli.md` — updated, accurate (documents T2/T3/T4 layout incl. new `migrate/` files); no .NET analogies. Verified.

### What worked well
- `--dry-run` / `--non-interactive` decision logic is clean and the safety posture is right: machine-local always skipped in `--non-interactive` regardless of `--include-machine-local`; `project`/`reference` refused with a clear stderr message + non-zero exit; conservative `SuggestScope` defaults.
- Test coverage of the non-form paths is genuinely good (source_id stability, machine-local skip, empty-dir guard, role filter).
- `SourceIDPrefix` shared-constant discipline so `--dry-run` preview stays bit-identical to the T5 POST is the right call (modulo the metadata Major).

- `nonInteractivePlans` calls `submitFn(plans)` before emitting the refusal-driven non-zero error; with the real T5 `submitFn` the ordering is fine, but a short comment explaining "submit succeeds, then signal partial refusal" would prevent a future reader misreading it as a bug.

cli/internal/migrate/picker.go
```diff
@@ -0,0 +1,283 @@
...
+
+	// ── Final confirm group ────────────────────────────────────────────
+	var confirmed bool
+	confirmGroup := huh.NewGroup(
+		huh.NewConfirm().
+			TitleFunc(func() string {
+				mc := countMigrateFromPlans(plans)
+				return fmt.Sprintf("Migrate %d entries (%d skipped). Proceed?", mc, len(plans)-mc)
+			}, nil).
+			Value(&confirmed),
```
damir-topic: **[🛑 Blocker]** — final "Migrate N entries… Proceed?" choice is captured but never enforced

**What:** `huh.NewConfirm().Value(&confirmed)` at line 156 binds the operator's Yes/No into local `confirmed`, but nothing reads it — `form.Run()` returns nil for *either* answer, and the caller (`memory.go:120-130`) goes straight from `FinalizeSkip` to `filterMigrate` to `submitFn`. The operator picking "No" on the batch-confirm still submits every non-skip plan.

**Why it matters:** same explicit-operator-consent class as the prior iteration's Select-not-bound blocker, one layer up — the operator's final "stop, don't send anything" gate is silently overridden. Per #611's ai_engineering §Safety anchor and the Initiative's "every migrate decision is an explicit operator choice" rule, the consent has to be enforceable at the batch level too, not just per file. Not caught by tests because the suite doesn't drive `form.Run()` (an artifact of `huh.Form` not exposing introspection — covered separately).

**Required change:** add a `Validate` callback on the confirm so the form refuses to submit when `confirmed == false`. `huh@v2.0.3` `field_confirm.go:150` runs the validator on `Blur()` and `form.go:577` gates `nextGroupMsg` on `group.Errors()` — combined, a non-nil validator error stalls submission, and the operator's only paths out are `Toggle` back to Yes or `Quit` (which surfaces as `ErrUserAborted` to `form.Run()` and is already handled by the `if err := form.Run(); err != nil` branch in `memory.go:116`).

```diff
-	var confirmed bool
 	confirmGroup := huh.NewGroup(
 		huh.NewConfirm().
 			TitleFunc(func() string {
 				mc := countMigrateFromPlans(plans)
 				return fmt.Sprintf("Migrate %d entries (%d skipped). Proceed?", mc, len(plans)-mc)
 			}, nil).
-			Value(&confirmed),
+			Validate(func(b bool) error {
+				if !b {
+					return errors.New("migration cancelled by operator")
+				}
+				return nil
+			}),
 	)
```

(Plus a corresponding `"errors"` import.) Drop the now-unused `confirmed` local. Alternative if the team wants a softer "No exits cleanly with `No entries selected for migration.`" behaviour: return `confirmed` from `BuildForm` and have `memory.go` check it before `filterMigrate`, but the `Validate` shape above is the smaller diff and matches huh's native idiom.

**Source:** `charm.land/huh/v2@v2.0.3/field_confirm.go:58` (`Validate`), `:150` (validator invoked on `Blur`); `charm.land/huh/v2@v2.0.3/form.go:577` (`if len(group.Errors()) > 0 { return f, nil }` on `nextGroupMsg`).

cli/internal/migrate/scan.go
damir-topic, line 1: **[🛑 Blocker — re-asserted from iteration 1]** — this file (and `suggest.go` / `scan_test.go` / `suggest_test.go`) is still #609's already-"merged" T2 work, absent from `main`

**What:** unchanged since the previous review iteration. `gh pr view 645 --json baseRefName` → `feat/608-migrate-cmd-huh-v2`. #645's merge commit `289fd97` is reachable only from `origin/feat/608-migrate-cmd-huh-v2`; `origin/main` HEAD is `547223c` and `git ls-tree -r origin/main --name-only | grep internal/migrate` shows only `doc.go` / `machinelocal.go` / `machinelocal_test.go` — T2's four files are not on `main`. PR #680's diff-vs-`main` still carries them (832 LOC, byte-identical to #645). PR is `mergeable=MERGEABLE` (no Git conflict) but `mergeStateStatus=BLOCKED` for separate reasons (required-review + red CI).

**Why it matters:** merging #680 as-is makes it the de-facto delivery of #609 while issue #609 stays OPEN and #645 reads merged — broken provenance and a project-board inconsistency. Issue #609 won't auto-close on merge (no `Closes #609` in PR body).

**Required change:** pick one of these two paths and apply it on this PR. No third option leaves the board consistent.

1. **Accept the carry (smaller diff path).** Add `Closes #609` to this PR's body so auto-close fires when #680 lands on `main`. Document the unusual provenance in the merge commit message ("T2 #609 reached main via this PR because #645 was merged to `feat/608-…`; original PR #645 stays in its current state").
2. **Split it out.** Cherry-pick `8cf376e` / `7de7cdc` / `9747886` to a new PR against `main`, land that first, then rebase `feat/611-huh-picker` so #680's diff is T4-only. Update #609/#645 state with a completion comment naming the new PR.

Option 1 is the cheaper path and what the diff already implies. Either way, the PR body must reflect what this merge will actually do to the issue graph.

(Pure project-board / merge-state defect — not a T4 code issue. Re-asserted because it is unchanged since iteration 1.)

cli/internal/cmd/migrate/memory.go
```diff
@@ -4,76 +4,228 @@
...
+				}
+				if len(refused) > 0 {
+					return fmt.Errorf("meho migrate memory: %d entries require interactive review", len(refused))
+				}
+				return nil
+			}
+
+			// ── Interactive path ──────────────────────────────────────
+			form, plans := migrate.BuildForm(files, opts)
+			accessible := os.Getenv("MEHO_ACCESSIBLE") != ""
```
damir-topic: **[⚠️ Major]** — `MEHO_ACCESSIBLE` non-empty check contradicts every adjacent contract

**What:** `os.Getenv("MEHO_ACCESSIBLE") != ""` activates accessible mode for any non-empty value, including `MEHO_ACCESSIBLE=0` and `MEHO_ACCESSIBLE=false`.

**Why it matters:** every other contract for this env var says `=1`. The `Long` help text three lines up: "Set MEHO_ACCESSIBLE=1 for screen-reader-friendly output." The PR description: "`MEHO_ACCESSIBLE=1` activates `form.WithAccessible(true)`." Task #611's body specifies the same. An operator setting `MEHO_ACCESSIBLE=0` to *disable* the variable today will instead enable it. CodeRabbit also flagged this on `2026-05-20`; reasserting because the contract divergence is real.

(Note: task #611 also says "`MEHO_ACCESSIBLE` (non-empty → accessible mode)" — but the help, PR body, and CodeRabbit-recommended fix all say `=1`. The `=1` shape is the one used elsewhere in `meho_app` env handling; align there.)

**Required change:**

```diff
-			accessible := os.Getenv("MEHO_ACCESSIBLE") != ""
+			accessible := os.Getenv("MEHO_ACCESSIBLE") == "1"
```

(If the team genuinely wants non-empty-truthy semantics, update the `Long` help text and the PR body to match — but then `MEHO_ACCESSIBLE=0` is a usability footgun.)

cli/internal/cmd/migrate/memory.go
```diff
@@ -4,76 +4,228 @@
...
+
+			// ── --non-interactive path ────────────────────────────────
+			if nonInteractive {
+				plans, refused := nonInteractivePlans(files, opts)
+				if len(refused) > 0 {
+					for _, f := range refused {
+						fmt.Fprintf(os.Stderr,
+							"meho migrate memory: skipping %s (type %q requires interactive review)\n",
+							f.Path, f.Type,
+						)
```
damir-topic, line 98: **[⚠️ Major]** — write refusal diagnostics to `cmd.ErrOrStderr()`, not `os.Stderr`

**What:** `fmt.Fprintf(os.Stderr, …)` writes the `--non-interactive` refusal lines to the process's stderr directly, bypassing `cmd.SetErr(...)`.

**Why it matters:** every other Cobra command in this repo uses `cmd.ErrOrStderr()` / `cmd.OutOrStdout()` (e.g. `cli/internal/cmd/login.go`, `cli/internal/output/format.go`); the rest of this same file already does (`cmd.OutOrStdout()` at line 73, 124). The mixed pattern breaks two things: (1) test capture — any `cmd.SetErr(&bytes.Buffer{})` in a test won't see the refusal lines; (2) consistent shell-redirection behaviour for callers wrapping `meho`. CodeRabbit also flagged this on `2026-05-20`.

**Required change:**

```diff
 				if len(refused) > 0 {
 					for _, f := range refused {
-						fmt.Fprintf(os.Stderr,
+						fmt.Fprintf(cmd.ErrOrStderr(),
 							"meho migrate memory: skipping %s (type %q requires interactive review)\n",
 							f.Path, f.Type,
 						)
 					}
 				}
```

(`os` stays imported via `os.Getenv` two lines below.)

cli/internal/cmd/migrate/memory.go
```diff
@@ -4,76 +4,228 @@
...
+					}
+				}
+				if err := submitFn(plans); err != nil {
+					return err
+				}
+				if markMigrated {
+					_ = migrate.TouchMarker(dir) // non-fatal in T4 stub
```
damir-topic: **[⚠️ Major]** — `TouchMarker` runs before the refusal error returns; a partial non-interactive run is marked complete

**What:** in the non-interactive path, the order is `submitFn(plans)` → `if markMigrated { TouchMarker(dir) }` → `if len(refused) > 0 { return error }`. When `--non-interactive --mark-migrated` runs against a directory with refused entries (`project` / `reference` files), the migration-complete marker is touched even though the command exits non-zero because some entries weren't migrated.

**Why it matters:** the marker exists so the post-login nudge (T5 #612) stops nagging once the operator has migrated; touching it on a partial run means the operator never sees the nudge again even though refused entries are still sitting unmigrated on their laptop. CodeRabbit flagged this on `2026-05-20` and the fix is straightforward — gate `TouchMarker` on "no refused entries".

**Required change:**

```diff
 				if err := submitFn(plans); err != nil {
 					return err
 				}
-				if markMigrated {
-					_ = migrate.TouchMarker(dir) // non-fatal in T4 stub
-				}
 				if len(refused) > 0 {
 					return fmt.Errorf("meho migrate memory: %d entries require interactive review", len(refused))
 				}
+				if markMigrated {
+					_ = migrate.TouchMarker(dir) // non-fatal in T4 stub
+				}
 				return nil
```

(The interactive path's `TouchMarker` call at line 132 already runs after all `Skip` filtering and the empty-plan early return, so it's correctly gated there.)

docs/codebase/cli.md
```diff
@@ -159,9 +160,10 @@ cli/
     │   │   ├── usage_test.go           # query-param + wire-shape + 403/400 routing tests.
     │   │   ├── retire_checklist.go     # `meho retrieval retire-checklist` (POST /api/v1/retrieve/retire-checklist) — G4.3-T6 #445.
     │   │   └── retire_checklist_test.go # surface-bucket + table-render + marshal tests.
-    │   ├── migrate/           # G5.3-T1 #608 — `meho migrate …` laptop-local migration verb tree (Initiative #375).
-    │   │   ├── migrate.go        # NewRootCmd + _ import charm.land/huh/v2 (declares dep; used by T2–T5).
-    │   │   └── memory.go         # `meho migrate memory` skeleton — 5 flags wired, RunE stub.
+    │   ├── migrate/           # G5.3 #608–#611 — `meho migrate …` laptop-local migration verb tree (Initiative #375).
+    │   │   ├── migrate.go        # NewRootCmd + _ import charm.land/huh/v2.
+    │   │   ├── memory.go         # `meho migrate memory` RunE — interactive picker / --dry-run / --non-interactive; submitFn seam for T5 (#611).
```
damir-topic: **[💡 Minor]** — T5 reference says `#611` but T5 is `#612`

**What:** "submitFn seam for T5 (`#611`)" — `#611` is **this** task (T4); T5 is `#612`. Lines 163 ("G5.3 #608–#611") and the file-tree section at line 184 ("full impl `#612`") both reference `#612` correctly, so this is a single-line typo.

**Required change:**

```diff
-    │   │   ├── memory.go         # `meho migrate memory` RunE — interactive picker / --dry-run / --non-interactive; submitFn seam for T5 (`#611`).
+    │   │   ├── memory.go         # `meho migrate memory` RunE — interactive picker / --dry-run / --non-interactive; submitFn seam for T5 (`#612`).
```

(Flagged inline rather than as a non-blocking note because the doc gets read by the next T5 implementer; the wrong issue number sends them to this PR's task instead of T5's.)

damir-topic (dismissed):
## /review-pr-slim summary — iteration 2

**Anchor task:** #611 — G5.3-T4 interactive `huh` picker + `--dry-run` / `--non-interactive`
**Surface area:** backend (Go CLI)
**Best-practices packs audited:** ai_engineering (§Safety / §Architecture-&-boundaries — explicit operator consent), Go idiom
**Verdict:** REQUEST_CHANGES

The previous iteration's biggest blocker (interactive Select wiring discarded operator choices) is **fixed cleanly** — `huh.NewSelect` now binds to `&plans[fi].Action`, the `WithHideFunc` closures read from the same source, and the parallel `actions` slice plus `countMigrateFromActions` are gone. `--dry-run` metadata now derives from `plan.File.Metadata` (previously hardcoded), and the four prior minor items (`time.Now()` hack, dead `migrateCount`, dead `withHuhForm`, empty `if` in `FinalizeSkip`) are all gone.

What remains: one **new blocker** (the final `huh.NewConfirm` Value is captured but never enforced — the same consent class as the previously-fixed Select bug, one layer up) and the prior **T2-not-on-`main` blocker**, plus three majors and one doc nit surfaced by CodeRabbit. CI Go + Python integration failures on this PR are **inherited from `main`** (unrelated `sddc-manager` gofmt failures + Python testcontainers regression), not T4 code.

### Acceptance criteria (#611)
- [x] `--dry-run` emits one JSON envelope per migrate-selected entry, line-parseable, `laptop-migration/<12 hex>` `source_id` — verified (`TestDryRun_*`, metadata now carried).
- [x] `--non-interactive` includes `user`/`feedback`, refuses `project`/`reference` to stderr + non-zero exit — verified (`TestNonInteractive_*`).
- [x] role filter omits `tenant`/`target` without `tenant_admin` — verified (`TestBuildScopeOptions_*`).
- [x] slug regex accepts `daily-routine`, rejects `Daily Routine` / `-x` / `x_y` — verified (`TestValidateSlug`).
- [⚠️] picker builder produces the expected group/field tree — **partially**. `TestBuildForm_GroupCount` now asserts default `plans[i].Action` values per file (good), but `huh.Form.Groups()` is unexported so the per-file 4-groups-and-1-confirm structure plus hidden-group transitions are still not inspected. Calling this satisfied-enough given the API surface; remaining gap is recorded in adjacent improvements.
- [x] `--help` documents `MEHO_ACCESSIBLE` — present in `Long`.
- [⚠️] `go vet` clean; `gofmt -d` empty; CI green — locally **clean** on T4 surface (vet, gofmt, golangci-lint v2.12.2, `go test ./internal/migrate/... ./internal/cmd/migrate/...` all pass). CI red: Go lane fails on `cli/internal/cmd/sddc-manager/{domain,network_pool}.go` gofmt (landed broken by #709, currently red on `main`); Python integration testcontainers fails (Python lane — untouched by this PR). Both pre-existing in `main`, not T4 regressions.
- [ ] **operator's final "Migrate N entries… Proceed?" confirm honoured** — ❌ broken (Blocker 1).

### Project-board alignment (evoila/meho)
| Issue | State | Link | Auto-close on merge | Delivered | Action |
|---|---|---|---|---|---|
| #611 (T4) | OPEN, In Progress | `Closes #611` | ✅ | ⚠️ partial (confirm not enforced) | fix Blocker 1, close on merge |
| #609 (T2) | **OPEN** despite #645 "merged" | none | won't fire — #645 merged to `feat/608-…`, never on `main` | 🚫 **832 LOC of T2 carried by this PR undisclosed** | **blocker** — add `Closes #609` to PR body, OR re-land #645 onto `main` separately and rebase #680 down to T4-only |
| #375 (Initiative) | OPEN, In Progress | parent | — | not completed (T5 #612 still open) | none |
| #608 (T1) | CLOSED | — | — | n/a (already on `main`) | none |
| #610 (T3) | CLOSED | — | — | n/a (already on `main`) | none |

### Predicted vs. actual diff
- ✅ `picker.go`, `marker.go`, `memory.go`, `memory_test.go`, `picker_test.go`, `docs/codebase/cli.md` — expected T4 surface; correct shape.
- ⚠️ `memory.go` lines 93, 102-107, 113 — expected but wrong (Majors, inline).
- ⚠️ `picker.go` lines 149-157 — expected but wrong (Blocker 1, inline).
- 🚫 `scan.go` / `suggest.go` / `scan_test.go` / `suggest_test.go` — unexpected vs `origin/main`, byte-identical to #645-merged T2; present only because #645 was merged to the wrong base (Blocker 2; inline on `scan.go` re-asserted from the prior iteration, still unresolved).

### Integration verification (in `.claude/worktrees/pr-680-review/cli`, head `e0a2e7a`)
- [x] `go vet ./...` — clean
- [x] `gofmt -d ./internal/migrate ./internal/cmd/migrate` — empty
- [x] `go test ./internal/migrate/... ./internal/cmd/migrate/...` — pass (`ok` 0.371s + 0.699s)
- [x] `golangci-lint run ./internal/migrate/... ./internal/cmd/migrate/...` — `0 issues`
- [⚠️ inherited red] CI Go lane — fails on `sddc-manager/domain.go:38` + `network_pool.go:37` gofmt (verified: `gofmt -l internal/cmd/sddc-manager/` flags both at same SHA on `origin/main`). Pre-existing main breakage from #709.
- [⚠️ inherited red] CI Python (integration testcontainers) — fails. This PR touches no Python; failure is in `backend/tests/`. Pre-existing main breakage.
- [⚠️ not run] interactive `huh` form exercised in TTY — not feasible headless; reasoned through against vendored `charm.land/huh/v2@v2.0.3` (`field_confirm.go:58/91/150` + `form.go:577` `group.Errors()` gate on `nextGroupMsg`) — that path is precisely what makes Blocker 1's `Validate(...)` fix work.

- 🛑 1 blocker — confirm Value bound but never enforced (`picker.go:149-157`)
- 🛑 1 blocker re-asserted — T2 (#609) carried undisclosed (`scan.go:1`)
- ⚠️ 3 majors — `MEHO_ACCESSIBLE != "" → == "1"`; refusal stream `os.Stderr → cmd.ErrOrStderr()`; `TouchMarker` runs before refusal-error return
- 💡 1 minor — `cli.md:165` T5 reference says `(#611)`; should be `(#612)`

- Acceptance criterion (e) is bounded by `huh.Form` not exposing `Groups()`/`Fields()`. The current test asserts default `plans[i].Action` which is a meaningful proxy (the hidden-group conditions all read `plans[fi].Action`). A future fully-structural assertion could either (a) return a small descriptor (e.g. `[]groupShape{Action, HiddenWhen, NextFieldKey}`) alongside the form from `BuildForm`, or (b) drive a goroutined `form.Run()` against a scripted `vt10x` PTY — but neither belongs in this PR.
- `nonInteractivePlans` calls `submitFn(plans)` before returning the refusal-driven non-zero error; with the real T5 `submitFn` the ordering is fine, but a one-line comment explaining "submit the safe subset, then signal partial refusal" would prevent a future reader misreading it as a bug.

### Docs
- [x] `docs/codebase/cli.md` — updated accurately (documents the new `migrate/` files, T1–T5 staging, `huh` dependency); no .NET analogies. One typo at line 165 (`#611` → `#612`) — minor, fixable inline.

### What worked well
- **Blocker 1 fix from iteration 1 is clean** — the `&plans[fi].Action` rebinding plus closure rewrite plus removal of `actions[]`/`countMigrateFromActions` matches the prescribed fix exactly. `huh@v2.0.3` honours the pre-set bound value as expected.
- **`--dry-run` metadata derivation** — `plan.File.Metadata` (with empty-map fallback) is exactly right; preserves the bit-for-bit `--dry-run` ↔ T5 POST contract.
- **`SourceIDPrefix` shared-constant discipline** — re-confirmed; still the right call.
- **Safety posture intact** — machine-local always skipped in `--non-interactive` regardless of `--include-machine-local`; `project`/`reference` refused with clear stderr + non-zero exit (modulo Major 2's `cmd.ErrOrStderr()` plumbing); conservative `SuggestScope` defaults.
- **Lint hygiene** — CodeRabbit's `scan.go close → closeDelim` shadowing fix + `scan_test.go` gofmt + S1039 cleanups all landed; vendored linter agrees.

**Auto-close (should happen automatically)**
- #611 — verify state CLOSED and `closedByPullRequestsReferences` contains this PR.

**Manual close needed**
- #609 — `Closes #609` requires Blocker 2's resolution first. If the team accepts the "PR carries T2" path, the PR body must say `Closes #609` so auto-close fires; otherwise close manually with a completion comment naming the merge commit.

- #611 → Done
- #609 → Done (only if delivery path resolved)

- #611: "Closed by #680. T4 huh picker + `--dry-run` / `--non-interactive` shipped. Verified: scope/action wiring, dry-run envelope shape, non-interactive refusal path."
- #609 (if applicable): "Delivered as part of #680's stacked diff. Original PR #645 merged to `feat/608-…` rather than `main`; T2 code reached `main` via this PR."

Issue #375, opened 2026-05-13
Initiative: G5.3 Laptop-local memory migration UX — meho migrate memory per-file picker + machine-local heuristic


damir-topic:
Parent goal: #216
Depends on: G5.1 server-side memory storage + verbs (#332) — specifically its `POST /api/v1/memory` endpoint, which this Initiative drives from the CLI.

## Scope

Ship the **first-time migration UX** from laptop-local `~/.claude/projects/<project>/memory/` markdown files into the server-side memory table, per consumer-needs.md §G5 ("Migration from laptop. First time I log in via `meho login`, the CLI offers to sync my existing `~/.claude/.../memory/` files into MEHO at the appropriate scopes. Operator picks per-file (defaults shown but reviewable). Some stay laptop-local (e.g. machine-specific paths); most migrate.") and §"Open questions" #3 ("The CLI should ask per-file at first sync, not bulk-migrate. Worth a small UX spec.").

The flow is an **interactive, re-runnable CLI verb** (`meho migrate memory`), **not** an implicit step inside `meho login`. Rationale: coupling auth and bulk data migration in one command (a) makes login non-idempotent, (b) buries the picker behind the auth flow on every subsequent login, (c) makes operating headless / scripted login (CI, jump-host) suddenly carry an interactive prompt. Per consumer-needs.md L151 the operator "offers" reading is satisfied by a one-line nudge printed at the end of a successful `meho login` ("Tip: run `meho migrate memory` to sync your laptop-local memories.") rather than auto-running.

**Area:** new `github.com/charmbracelet/huh` dependency in [cli/go.mod](cli/go.mod) (MIT-licensed, v2.0.3+), new Cobra subcommand at [cli/internal/cmd/migrate_memory.go](cli/internal/cmd/migrate_memory.go), new helpers under [cli/internal/migrate/](cli/internal/migrate/) (file scanning + frontmatter parsing + machine-local heuristic + scope-suggestion), small addition to [cli/internal/cmd/login.go](cli/internal/cmd/login.go) for the one-line post-login nudge, new docs at [docs/cli/memory-migration.md](docs/cli/memory-migration.md).

## Work items grouped under this Initiative

[...]

Issue #427, opened 2026-05-14
Task: G5.1-T6 Operator runbook docs/cross-repo/memory-migration.md + docs/architecture/memory.md (5-scope matrix + decision tree)


damir-topic:
Parent initiative: #332
Depends-on: #421, #422, #423, #424, #426

## Target

Ship `docs/cross-repo/memory-migration.md` — the operator-facing recipe for transitioning from laptop-local `~/.claude/.../memory/` files to MEHO's server-side memory. Documents the per-file picker UX (which G5.3 #375 ships as the actual CLI verb) + the 5-scope decision matrix + the rollback procedure if the migration regresses.

Plus `docs/architecture/memory.md` — architecture explainer for the memory module, sister to `docs/architecture/kb.md` (G4.1-T6).

## Files touched

- `docs/cross-repo/memory-migration.md` — new (canonical operator runbook).
- `docs/architecture/memory.md` — new architecture doc.
- Cross-link from `docs/cross-repo/README.md` if it exists.

## Operator runbook outline

```markdown
# Migrating laptop-local memory to MEHO

## Why this matters
Today: ~15 markdown files on each operator's laptop in
`~/.claude/.../memory/`. What one operator's Claude learns is invisible
to every other operator's Claude. Lost if the laptop dies.

After: server-side memory across 5 scopes; the team becomes the unit of
memory (per consumer-needs.md L131).

## The 5-scope decision matrix
| Scope | Visible to | Use for |
|---|---|---|
| `user` | Just you, across all tenants | Personal behavioral preferences ("I prefer kubectl over k9s") |
| `user-tenant` | You within this tenant | Personal context scoped to a specific lab |
| `user-target` | You scoped to one target | Personal notes about one vCenter / k8s cluster |
| `tenant` | Everyone in your tenant (tenant_admin write only) | Team-wide conventions ("we use GitOps for everything") |
| `target` | Anyone with access to this target | Target-specific gotchas ("rdc-vcenter requires VPN") |

## Per-file migration via G5.3 (when shipped)
G5.3 #375 ships `meho migrate memory` — an interactive picker that walks
through each local memory file and prompts the operator to choose scope.

In the meantime, manual migration via CLI:
```
# For each laptop-local memory file:
cat ~/.claude/.../memory/<file>.md | meho remember --scope <chosen-scope> --slug <slug>
```

## Default TTL behavior (when G5.2 ships)
G5.2 #374 ships default-7-day-TTL injection for user-scoped writes
[...]

PR #685, opened 2026-05-19
feat(migrate): G5.3-T5 submission layer + post-login nudge + marker file + docs

### LINKED ISSUE #612: Task: G5.3-T5 Submission layer + post-login nudge + marker file + docs/cli/memory-migration.md
Parent initiative: #375 (G5.3 Laptop-local memory migration UX)
Parent goal: #216
Depends on: #611 (produces the `[]SubmitPlan`).
Depends on: #332 (G5.1) for the **live** `POST /api/v1/memory` endpoint
— T5 is independently buildable and testable against a fake backplane
(httptest) without #332 merged; only end-to-end against a real
backplane requires #332.

## Current state

After T4, `meho migrate memory` produces a `[]SubmitPlan` (interactive
or non-interactive) and can `--dry-run` print envelopes, but the
`submitFn` seam is a stub — nothing is POSTed. There is no post-login
nudge, no migration marker file, and no `docs/cli/` documentation. The
established per-tree HTTP convention is an in-package
`resolveBackplane` / `doAuthedRequest` / `renderRequestError` trio
(see `cli/internal/cmd/kb/kb.go` header — a shared client would close
an import cycle).

## Desired state

**Submission — `cli/internal/migrate/submit.go` + the `submitFn`
seam in `cli/internal/cmd/migrate/memory.go`:**

- For each migrate entry, `POST /api/v1/memory` with body
  `{scope, slug, body, metadata:{tags:[]}, source_id:
  "laptop-migration/<sha256-prefix>"}` where `<sha256-prefix>` is the
  first 12 hex chars of `BodySHA256` (same scheme T4's dry-run prints
  — keep them identical; one shared constant for the prefix length).
  Do **not** set `expires_at`: the G5.2 server-side default-TTL
  injection owns that for `user`-scoped entries.
- Idempotency: re-running with unchanged body for the same
  `source_id` is a server-side upsert no-op; a changed body updates
  the row and bumps `updated_at`. The CLI does not dedupe locally — it
  relies on the G5.1 upsert-by-`source_id` contract; document this
  reliance.
- HTTP plumbing uses the in-package helper trio (copy the kb pattern;
  do **not** import `cli/internal/api`). Auth piggybacks on the token
  `meho login` stored, same as `meho kb` / `meho audit`.
- Progress via `charm.land/huh/v2/spinner` (one tick per file). On a
  transient error pause and prompt `Retry / Skip / Abort`
  (`huh.NewSelect`); `--non-interactive` maps transient errors to a
  bounded auto-retry then Skip (no prompt). Final summary line:
  `Migrated: N, Skipped: M, Errored: K (retried R)`.

**Post-login nudge — edit `cli/internal/cmd/login.go`:**

- After the existing `"Logged in to %s; token stored in %s.\n"` line,
  print a one-line tip iff: the resolved source dir
  (`migrate.ResolveSourceDir("")`) exists and is non-empty, **and** the
  per-source-dir marker file is absent. Marker path:
  `<XDG_CONFIG_HOME>/meho/migrated-from/<sanitized-source-dir>`. Tip
  text: `Tip: you have N memory files at <dir>. Run `+"`"+`meho migrate
  memory`+"`"+` to sync them to MEHO.` Keep the nudge non-fatal — any
  error resolving the dir or stat-ing the marker degrades to "print
  nothing" (never block or fail login). This is the
  consumer-needs.md §G5 "the CLI offers" reading: a nudge, not an
  auto-run.

**Marker management — `cli/internal/migrate/marker.go`:**

- Touch the marker on a successful migration run, and on explicit
  `meho migrate memory --mark-migrated` (which only writes the marker
  and exits — lets an operator who migrated by other means silence the
  nudge). Deleting the marker re-enables the nudge (documented).

**Docs — `docs/cli/memory-migration.md` (creates `docs/cli/`):**

- End-to-end flow (what is scanned, what the picker shows, what is
  sent), the exact T3 machine-local regex categories +
  `--include-machine-local` override, the T2 scope-suggestion table +
  when to override, re-run/idempotency semantics
  (`source_id`, update-on-change), the `--dry-run` / `--non-interactive`
  / `--mark-migrated` / `MEHO_ACCESSIBLE` matrix, and the
  `<!-- meho:machine-local -->` opt-out.
- Scope strictly to the **CLI migration flow**. The server-side
  5-scope matrix + decision tree is #427's `docs/cross-repo/
  memory-migration.md` (G5.1-T6) — same filename, different dir and
  audience. Cross-link: this doc links to #427's for the scope model;
  do not duplicate that content here.

## Acceptance criteria

- [ ] `cd cli && go test ./internal/migrate/... ./internal/cmd/migrate/...` passes, including an integration test against an `httptest.Server` fake backplane + in-memory `auth.TokenStore` (the `cli/internal/api/client_test.go` pattern) proving: (a) each migrate entry POSTs to `/api/v1/memory` with the correct `{scope,slug,body,source_id}`; (b) same body re-run = identical `source_id`, fake records an upsert no-op; (c) changed body re-run = same `source_id`, body updated; (d) transient 503 then 200 → one retry, success counted.
- [ ] Post-login nudge: unit test with a fake source dir + token store asserts the tip prints when the marker is absent and a non-empty dir exists, and prints **nothing** when the marker exists, when the dir is empty, or when dir resolution errors; login never fails because of the nudge.
- [ ] `meho migrate memory --mark-migrated` writes the marker and exits 0 without POSTing; subsequent simulated login prints no tip.
- [ ] `docs/cli/memory-migration.md` exists, documents the heuristic regexes, the scope-suggestion table, the flag matrix, the opt-out comment, and links to #427's `docs/cross-repo/memory-migration.md` for the 5-scope model (no duplication).
- [ ] No `cli/internal/api` import in the migrate packages (in-package helper trio per the kb convention); `go vet` clean; `gofmt -d` empty.
- [ ] CLI cross-compiles for linux/amd64, linux/arm64, darwin/amd64, darwin/arm64; CI green.
- [ ] Initiative #375 "Execution tasks" T5 line and Goal #216's G5.3 checklist line are ticked once this merges.

## Out of scope

- Reverse migration (`meho backup memory`) — flagged as a future G5.4
  per #375.
- Conflict-resolution UI for same-slug-different-machine — v0.2 is
  update-on-change (newer `updated_at` wins) per #375; an explicit
  merge UI is additive future work.
- Re-implementing the server-side memory storage / RBAC / TTL — owned
  by #332 (G5.1) and #374 (G5.2).
- The server-side operator runbook + 5-scope decision tree — owned by
  #427 (G5.1-T6). T5 only cross-links it.
- Inherits all Out-of-scope bullets of Initiative #375.

## References

- Parent: Initiative #375 (work items 4, 6, 7), Goal #216.
- Depends on: #332 (G5.1 `POST /api/v1/memory` upsert-by-`source_id`).
- Sibling: #374 (G5.2 default 7-day TTL on `user`-scope — why T5 omits
  `expires_at`).
- Adjacent doc, cross-linked not duplicated: #427 (G5.1-T6
  `docs/cross-repo/memory-migration.md` — server-side 5-scope matrix +
  decision tree).
- Related command tree (no collision): #424 (G5.1-T4 `meho
  remember/recall/forget/list`).
- HTTP-helper convention: `cli/internal/cmd/kb/kb.go` header +
  `cli/internal/cmd/kb/add.go` (the `resolveBackplane` /
  `doAuthedRequest` / `renderRequestError` trio to copy).
- Test harness: `cli/internal/api/client_test.go` (httptest +
  in-memory `auth.TokenStore`), `cli/internal/cmd/kb/add_test.go`.
- Post-login edit point: `cli/internal/cmd/login.go:119` (the
  `"Logged in to %s; token stored in %s."` line).
- huh spinner: https://pkg.go.dev/charm.land/huh/v2/spinner (v2.0.3).
- Best practices: **devops_best_practices** §CI (cross-compile matrix,
  idempotent re-run) and **ai_engineering_best_practices** §Safety
  (machine-local content never silently leaves the laptop; the nudge
  is an offer, never an auto-run — login stays idempotent and
  headless-safe, per #375 rationale).

ddzafic:
## Summary

- **`cli/internal/cmd/migrate/submit.go`** — real HTTP `POST /api/v1/memory` using the in-package helper trio (copied from kb pattern per the established convention). `huh/v2` spinner progress per entry; transient 5xx auto-retry in `--non-interactive`, `Retry/Skip/Abort` prompt interactively. Final summary: `Migrated: N, Skipped: M, Errored: K (retried R)`.
- **`cli/internal/cmd/migrate/memory.go`** — wire real `doSubmit` (nil `submitFn` → closure calling `doSubmit`); add `--backplane` flag; remove T4 `stubSubmitFn`; update `sourceID` to use `migrate.SourceIDPrefix`.
- **`cli/internal/migrate/picker.go`** — promote `SourceIDPrefix` constant to the internal package so both dry-run output and POST share it (bit-for-bit identical `source_id`).
- **`cli/internal/migrate/marker.go`** — full XDG marker implementation (`TouchMarker` writes `$XDG_CONFIG_HOME/meho/migrated-from/<sanitized-dir>`; `MarkerExists` degrades gracefully).
- **`cli/internal/cmd/login.go`** — `printMigrationNudge` after successful login; never blocks login.
- **`docs/cli/memory-migration.md`** — new; documents scan, picker, heuristics, scope table, flag matrix, idempotency, nudge, opt-out. Cross-links to `docs/cross-repo/memory-migration.md` for the 5-scope model.

## Test plan

- [x] `go test ./internal/migrate/... ./internal/cmd/migrate/... ./internal/cmd/...` — all pass
- [x] Integration tests against `httptest.Server` fake backplane: POST body shape, source_id stability, upsert re-run, transient retry (503→201), summary line
- [x] Post-login nudge: prints when files exist + marker absent; silent when marker exists, dir empty, or dir resolution errors
- [x] `--mark-migrated` writes marker; subsequent nudge is silent
- [x] `go vet` clean; cross-compiles for linux/amd64, linux/arm64, darwin/amd64, darwin/arm64
- [x] No `cli/internal/api` import in `cli/internal/migrate/` packages

Closes #612
Parent initiative: #375 (G5.3)

## Summary by CodeRabbit

* **New Features**
  * `meho login` now displays tips about pending memory migrations
  * Memory migration submission now fully implemented with automatic retry for transient failures
  * Added `--backplane` flag to override backplane URL
  * Added `--mark-migrated` flag for migration tracking
  * Interactive mode includes retry/skip/abort prompts for submission errors

* **Documentation**
  * Added comprehensive memory migration command reference

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/685?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cli/internal/cmd/migrate/memory.go
+++ b/cli/internal/cmd/migrate/memory.go
@@ -225,4 +225,3 @@ func sourceID(plan migrate.SubmitPlan) string {
 	}
 	return "laptop-migration/" + prefix
 }
-
--- a/cli/internal/cmd/migrate/memory.go
+++ b/cli/internal/cmd/migrate/memory.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"strings"
 
 	"charm.land/huh/v2"
 	"github.com/spf13/cobra"
@@ -16,9 +15,8 @@ import (
 	"github.com/evoila/meho/cli/internal/migrate"
 )
 
-// SubmitFn is the seam T5 replaces with the real backplane POST.
-// T4 wires a no-op stub so the command tree compiles and the dry-run
-// path works end-to-end without a network call.
+// SubmitFn is the seam tests replace with a mock; production code uses
+// the real doSubmit wired in newMemoryCmd.
 type SubmitFn func(plans []migrate.SubmitPlan) error
 
 // dryRunEnvelope is the machine-readable shape emitted by --dry-run
@@ -31,14 +29,10 @@ type dryRunEnvelope struct {
 	SourceID string         `json:"source_id"`
 }
 
-// SourceIDPrefix is the number of hex chars taken from the SHA-256 body
-// hash to build a source_id. Shared constant between T4 dry-run output
-// and T5 POST (keeping them bit-for-bit identical ensures --dry-run
-// preview = exactly what would be sent).
-const SourceIDPrefix = 12
-
 func newMemoryCmd() *cobra.Command {
-	return newMemoryCmdWithSubmit(stubSubmitFn)
+	// Real submit fn is wired inside RunE via newMemoryCmdWithSubmit(nil):
+	// a nil submitFn causes RunE to call doSubmit (the real HTTP POST path).
+	return newMemoryCmdWithSubmit(nil)
 }
 
 // newMemoryCmdWithSubmit is the testable variant: callers inject a submitFn.
@@ -59,6 +53,16 @@ func newMemoryCmdWithSubmit(submitFn SubmitFn) *cobra.Command {
 			nonInteractive, _ := flags.GetBool("non-interactive")
 			includeML, _ := flags.GetBool("include-machine-local")
 			markMigrated, _ := flags.GetBool("mark-migrated")
+			backplaneOverride, _ := flags.GetString("backplane")
+
+			// Resolve the effective submit function: injected for tests,
+			// real HTTP POST for production (nil submitFn).
+			eff := submitFn
+			if eff == nil {
+				eff = func(plans []migrate.SubmitPlan) error {
+					return doSubmit(cmd, backplaneOverride, plans)
+				}
+			}
 
 			// ── Resolve source directory ──────────────────────────────
 			dir, err := migrate.ResolveSourceDir(source)
@@ -78,8 +82,6 @@ func newMemoryCmdWithSubmit(submitFn SubmitFn) *cobra.Command {
 
 			opts := migrate.BuildFormOpts{
 				IncludeMachineLocal: includeML,
-				// TenantConfigured / IsTenantAdmin resolved from auth in T5;
-				// T4 leaves these at their zero values (conservative defaults).
 			}
 
 			// ── --dry-run path ────────────────────────────────────────
@@ -98,7 +100,7 @@ func newMemoryCmdWithSubmit(submitFn SubmitFn) *cobra.Command {
 						)
 					}
 				}
-				if err := submitFn(plans); err != nil {
+				if err := eff(plans); err != nil {
 					return err
 				}
 				if len(refused) > 0 {
@@ -133,7 +135,7 @@ func newMemoryCmdWithSubmit(submitFn SubmitFn) *cobra.Command {
 				return nil
 			}
 
-			if err := submitFn(toSubmit); err != nil {
+			if err := eff(toSubmit); err != nil {
 				return err
 			}
 			if markMigrated {
@@ -148,6 +150,7 @@ func newMemoryCmdWithSubmit(submitFn SubmitFn) *cobra.Command {
 	cmd.Flags().Bool("non-interactive", false, "skip the interactive picker; migrates only user/feedback entries")
 	cmd.Flags().Bool("include-machine-local", false, "include machine-local entries (default: skip them)")
 	cmd.Flags().Bool("mark-migrated", false, "touch the migration-complete marker after successful submission")
+	cmd.Flags().String("backplane", "", "backplane URL override (default: from meho login config)")
 
 	return cmd
 }
@@ -215,25 +218,11 @@ func filterMigrate(plans []migrate.SubmitPlan) []migrate.SubmitPlan {
 }
 
 // sourceID builds the stable source_id string from a plan's BodySHA256.
-// Uses the first SourceIDPrefix hex chars of the SHA-256 hash.
 func sourceID(plan migrate.SubmitPlan) string {
 	prefix := plan.File.BodySHA256
-	if len(prefix) > SourceIDPrefix {
-		prefix = prefix[:SourceIDPrefix]
+	if len(prefix) > migrate.SourceIDPrefix {
+		prefix = prefix[:migrate.SourceIDPrefix]
 	}
 	return "laptop-migration/" + prefix
 }
 
-// stubSubmitFn is the T4 no-op seam. T5 replaces it with the real
-// backplane POST implementation.
-func stubSubmitFn(plans []migrate.SubmitPlan) error {
-	if len(plans) == 0 {
-		return nil
-	}
-	scopes := make([]string, 0, len(plans))
-	for _, p := range plans {
-		scopes = append(scopes, fmt.Sprintf("%s (%s)", p.Slug, p.Scope))
-	}
-	return fmt.Errorf("meho migrate memory: submit not yet implemented (T5); would send: %s",
-		strings.Join(scopes, ", "))
-}
--- a/cli/internal/cmd/migrate/submit.go
+++ b/cli/internal/cmd/migrate/submit.go
@@ -58,9 +58,8 @@ const maxAutoRetry = 3
 // to avoid opening /dev/tty in a headless environment.
 var runSpinnerFn = func(sp *spinner.Spinner) error { return sp.Run() }
 
-// doSubmit orchestrates spinner progress, per-entry POST with
-// retry/skip/abort handling, and the final summary line. nonInteractive
-// controls whether a transient error prompts the user or auto-retries.
+// doSubmit orchestrates per-entry submission. The interactive error prompt runs
+// outside the spinner to avoid concurrent tea programs sharing the same TTY.
 func doSubmit(cmd *cobra.Command, backplaneOverride string, plans []migrate.SubmitPlan) error {
 	backplaneURL, err := resolveBackplaneURL(backplaneOverride)
 	if err != nil {
@@ -68,41 +67,36 @@ func doSubmit(cmd *cobra.Command, backplaneOverride string, plans []migrate.Subm
 	}
 
 	nonInteractive, _ := cmd.Flags().GetBool("non-interactive")
-
 	var res submitResult
 
-	sp := spinner.New().
-		Title(fmt.Sprintf("Migrating %d entries…", len(plans))).
-		ActionWithErr(func(ctx context.Context) error {
-			for i := range plans {
-				p := plans[i]
-				slug := p.Slug
-				if err := postWithRetry(ctx, cmd, backplaneURL, p, nonInteractive, &res); err != nil {
-					if errors.Is(err, errAborted) {
-						return nil
-					}
-					return fmt.Errorf("entry %s: %w", slug, err)
-				}
+	for i := range plans {
+		p := plans[i]
+		if err := postWithRetry(cmd, backplaneURL, p, nonInteractive, &res); err != nil {
+			if errors.Is(err, errAborted) {
+				break
 			}
-			return nil
-		})
-
-	if err := runSpinnerFn(sp); err != nil {
-		return err
+			return fmt.Errorf("entry %s: %w", p.Slug, err)
+		}
 	}
 
 	fmt.Fprintln(cmd.OutOrStdout(), res.String())
+	if res.Errored > 0 {
+		noun := "entries"
+		if res.Errored == 1 {
+			noun = "entry"
+		}
+		return fmt.Errorf("meho migrate memory: %d %s failed to migrate", res.Errored, noun)
+	}
 	return nil
 }
 
 var errAborted = errors.New("migration aborted by user")
-var errSkip = errors.New("entry skipped")
 
-// postWithRetry POSTs one plan entry, retrying on transient errors.
-// In interactive mode a prompt offers Retry / Skip / Abort.
-// In non-interactive mode, auto-retries up to maxAutoRetry then skips.
+// postWithRetry submits one plan entry. In non-interactive mode it auto-retries
+// on transient errors (up to maxAutoRetry attempts). In interactive mode, each
+// attempt runs inside its own spinner; the Retry/Skip/Abort prompt is presented
+// outside the spinner so no two tea programs run concurrently.
 func postWithRetry(
-	ctx context.Context,
 	cmd *cobra.Command,
 	backplaneURL string,
 	plan migrate.SubmitPlan,
@@ -125,7 +119,17 @@ func postWithRetry(
 	attempt := 0
 	for {
 		attempt++
-		_, postErr := doAuthedRequest(ctx, backplaneURL, http.MethodPost, "/api/v1/memory", raw)
+		var postErr error
+		sp := spinner.New().
+			Title(fmt.Sprintf("Migrating %q…", plan.Slug)).
+			ActionWithErr(func(ctx context.Context) error {
+				_, postErr = doAuthedRequest(ctx, backplaneURL, http.MethodPost, "/api/v1/memory", raw)
+				return nil
+			})
+		if err := runSpinnerFn(sp); err != nil {
+			return err
+		}
+
 		if postErr == nil {
 			res.Migrated++
 			return nil
@@ -135,12 +139,12 @@ func postWithRetry(
 			res.Errored++
 			fmt.Fprintf(cmd.ErrOrStderr(),
 				"meho migrate memory: permanent error for %s: %v\n", plan.Slug, postErr)
-			return nil
+			return fmt.Errorf("permanent error for %s: %w", plan.Slug, postErr)
 		}
 
-		// Transient error handling.
+		// Transient error: auto-retry in non-interactive mode.
 		if nonInteractive {
-			if attempt < maxAutoRetry {
+			if attempt <= maxAutoRetry {
 				res.Retried++
 				continue
 			}
@@ -151,7 +155,7 @@ func postWithRetry(
 			return nil
 		}
 
-		// Interactive prompt.
+		// Interactive mode: prompt runs outside the spinner (no concurrent tea programs).
 		var choice string
 		prompt := huh.NewSelect[string]().
 			Title(fmt.Sprintf("Error migrating %q: %v", plan.Slug, postErr)).
--- a/cli/internal/cmd/migrate/submit.go
+++ b/cli/internal/cmd/migrate/submit.go
@@ -75,7 +75,15 @@ func doSubmit(cmd *cobra.Command, backplaneOverride string, plans []migrate.Subm
 			if errors.Is(err, errAborted) {
 				break
 			}
-			return fmt.Errorf("entry %s: %w", p.Slug, err)
+			fmt.Fprintln(cmd.OutOrStdout(), res.String())
+			// Route the permanent error through renderSubmitError so the
+			// exit code reflects the failure class (auth_expired=2 for
+			// expired/missing/refresh-rejected tokens, unexpected=4 for
+			// other non-2xx, unreachable=3 for transport errors). Without
+			// this, every permanent error would exit 1 — same code as a
+			// generic CLI error — and `meho migrate memory` in a cron job
+			// would be indistinguishable from an unrelated CLI flag bug.
+			return renderSubmitError(cmd, backplaneURL, err)
 		}
 	}
 
@@ -137,9 +145,11 @@ func postWithRetry(
 
 		if !isTransient(postErr) {
 			res.Errored++
-			fmt.Fprintf(cmd.ErrOrStderr(),
-				"meho migrate memory: permanent error for %s: %v\n", plan.Slug, postErr)
-			return fmt.Errorf("permanent error for %s: %w", plan.Slug, postErr)
+			// Bubble the raw error up — doSubmit routes it through
+			// renderSubmitError for proper exit-code classification.
+			// fmt.Errorf with %w preserves errors.As(_, *httpError) so
+			// renderSubmitError can pick the right exit class.
+			return fmt.Errorf("entry %s: %w", plan.Slug, postErr)
 		}
 
 		// Transient error: auto-retry in non-interactive mode.
--- a/cli/internal/cmd/migrate/submit.go
+++ b/cli/internal/cmd/migrate/submit.go
@@ -1,0 +1,387 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 evoila Group
+
+package migrate
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"charm.land/huh/v2"
+	"charm.land/huh/v2/spinner"
+	"github.com/spf13/cobra"
+
+	"github.com/evoila/meho/cli/internal/api"
+	"github.com/evoila/meho/cli/internal/auth"
+	"github.com/evoila/meho/cli/internal/migrate"
+	"github.com/evoila/meho/cli/internal/output"
+)
+
+// memoryPostBody is the request body POSTed to POST /api/v1/memory.
+// source_id is the deduplication key on the server side — the G5.1
+// upsert-by-source_id contract means that re-running with the same
+// body is a no-op (updated_at unchanged) and with a changed body is
+// an update (updated_at bumped). expires_at is deliberately omitted:
+// the G5.2 server-side default-TTL injection owns it for user-scoped
+// entries (#374).
+type memoryPostBody struct {
+	Scope    string         `json:"scope"`
+	Slug     string         `json:"slug"`
+	Body     string         `json:"body"`
+	Metadata map[string]any `json:"metadata"`
+	SourceID string         `json:"source_id"`
+}
+
+// submitResult tallies the outcome of a submitPlans call.
+type submitResult struct {
+	Migrated int
+	Skipped  int
+	Errored  int
+	Retried  int
+}
+
+func (r submitResult) String() string {
+	return fmt.Sprintf("Migrated: %d, Skipped: %d, Errored: %d (retried %d)",
+		r.Migrated, r.Skipped, r.Errored, r.Retried)
+}
+
+const maxAutoRetry = 3
+
+// runSpinnerFn is the spinner seam: tests replace it with accessible mode
+// to avoid opening /dev/tty in a headless environment.
+var runSpinnerFn = func(sp *spinner.Spinner) error { return sp.Run() }
+
+// doSubmit orchestrates spinner progress, per-entry POST with
+// retry/skip/abort handling, and the final summary line. nonInteractive
+// controls whether a transient error prompts the user or auto-retries.
+func doSubmit(cmd *cobra.Command, backplaneOverride string, plans []migrate.SubmitPlan) error {
+	backplaneURL, err := resolveBackplaneURL(backplaneOverride)
+	if err != nil {
+		return renderSubmitError(cmd, "", err)
+	}
+
+	nonInteractive, _ := cmd.Flags().GetBool("non-interactive")
+
+	var res submitResult
+
+	sp := spinner.New().
+		Title(fmt.Sprintf("Migrating %d entries…", len(plans))).
+		ActionWithErr(func(ctx context.Context) error {
+			for i := range plans {
+				p := plans[i]
+				slug := p.Slug
+				if err := postWithRetry(ctx, cmd, backplaneURL, p, nonInteractive, &res); err != nil {
+					if errors.Is(err, errAborted) {
+						return nil
+					}
+					return fmt.Errorf("entry %s: %w", slug, err)
+				}
+			}
+			return nil
+		})
+
+	if err := runSpinnerFn(sp); err != nil {
+		return err
+	}
+
+	fmt.Fprintln(cmd.OutOrStdout(), res.String())
+	return nil
+}
+
+var errAborted = errors.New("migration aborted by user")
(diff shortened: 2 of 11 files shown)


ddzafic:
@copilot review

docs/cli/memory-migration.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Scope suggestion table
+
+The picker pre-selects a scope based on the file's `type` front-matter field:
+
+| `type` | Suggested scope | Rationale |
+|--------|----------------|-----------|
+| `user` | `user` | Personal workflow knowledge — scoped to you. |
+| `feedback` | `user` | Coding preferences — personal by default. |
+| `project` | `user×target` | Project context — tied to the current target. |
```
damir-topic: **[🛑 Blocker]** — scope-suggestion table contradicts `SuggestScope` code in two rows

**What:** the table says `project → user×target` and `reference → user×tenant`. The actual code in [`cli/internal/migrate/suggest.go:44-58`](cli/internal/migrate/suggest.go#L44-L58) maps `project → user×tenant` (when `tenantConfigured`, else `user`) and `reference → user`. The Initiative #375 spec body matches the code, not the table.

**Why it matters:** operators reading this doc to predict what `meho migrate memory` will do will get the wrong scope for two of the four declared `type:` values. `project` going to `user×target` instead of `user×tenant` would put tenant-shared knowledge under a target boundary; `reference` going to `user×tenant` instead of `user` would over-share a personal reference. Since the picker shows the suggested scope pre-selected, an operator scanning past the picker would land on a wrong default they didn't notice.

**Required change:**

```diff
 | `user` | `user` | Personal workflow knowledge — scoped to you. |
 | `feedback` | `user` | Coding preferences — personal by default. |
-| `project` | `user×target` | Project context — tied to the current target. |
-| `reference` | `user×tenant` | Stable reference — shared within the tenant. |
+| `project` | `user×tenant` (when a tenant is configured; otherwise `user`) | Project context — tied to the operator within their tenant. |
+| `reference` | `user` | References are system-level by default; tenant-level reference is an operator-initiated choice in the picker. |
 | *(unset / other)* | `user` | Conservative default. |
```

**Source:** `cli/internal/migrate/suggest.go:44-58`; Initiative #375 body §"Work items grouped under this Initiative" item 2.

cli/internal/cmd/migrate/submit.go
```diff
@@ -0,0 +1,387 @@
...
+	if err := runSpinnerFn(sp); err != nil {
+		return err
+	}
+
+	fmt.Fprintln(cmd.OutOrStdout(), res.String())
+	return nil
+}
+
+var errAborted = errors.New("migration aborted by user")
+var errSkip = errors.New("entry skipped")
```
damir-topic: **[🛑 Blocker]** — `errSkip` is declared but never referenced; `golangci-lint` `unused` fails CI

**What:** `var errSkip = errors.New("entry skipped")` is declared next to `errAborted`, but the "skip" branch of `postWithRetry` at lines 167-173 just increments `res.Skipped` and `return nil` — it never returns `errSkip`. `errAborted` IS used (line 81, 165, 175) so that one stays.

**Why it matters:** `golangci-lint run ./internal/cmd/migrate/...` flags it as `var errSkip is unused`, which currently blocks the Go CI lane (verified locally on `017d0125`).

**Required change:**

```diff
 var errAborted = errors.New("migration aborted by user")
-var errSkip = errors.New("entry skipped")
```

Either delete the sentinel or use it — but the "skip" branch's current control flow (`return nil` + caller continues to the next plan) is correct, so deletion is the minimum diff.

cli/internal/cmd/migrate/memory.go
```diff
@@ -207,25 +210,11 @@ func filterMigrate(plans []migrate.SubmitPlan) []migrate.SubmitPlan {
...
-// Uses the first SourceIDPrefix hex chars of the SHA-256 hash.
 func sourceID(plan migrate.SubmitPlan) string {
 	prefix := plan.File.BodySHA256
-	if len(prefix) > SourceIDPrefix {
-		prefix = prefix[:SourceIDPrefix]
+	if len(prefix) > migrate.SourceIDPrefix {
+		prefix = prefix[:migrate.SourceIDPrefix]
 	}
 	return "laptop-migration/" + prefix
 }
```
damir-topic: **[⚠️ Major]** — trailing blank line fails `gofmt` and `golangci-lint`

**What:** `memory.go` ends with a blank line after `}` of `sourceID`. `gofmt -d` flags it; `golangci-lint` surfaces the same diagnostic. Verified at `017d0125`:

```
$ gofmt -d cli/internal/cmd/migrate/memory.go
diff internal/cmd/migrate/memory.go.orig internal/cmd/migrate/memory.go
@@ -217,4 +217,3 @@
 	}
 	return "laptop-migration/" + prefix
 }
-
```

**Why it matters:** required Go CI lane fails on this; merge is blocked.

**Required change:** delete the trailing newline. `gofmt -w cli/internal/cmd/migrate/memory.go` fixes it.

cli/internal/cmd/migrate/submit.go
```diff
@@ -0,0 +1,387 @@
...
+			res.Errored++
+			fmt.Fprintf(cmd.ErrOrStderr(),
+				"meho migrate memory: skipping %s after %d retries: %v\n",
+				plan.Slug, attempt, postErr)
+			return nil
+		}
+
+		// Interactive prompt.
+		var choice string
+		prompt := huh.NewSelect[string]().
```
damir-topic, line 170: **[⚠️ Major]** — interactive Retry/Skip/Abort prompt runs *inside* the live spinner action; two tea programs share one TTY

**What:** `doSubmit` (lines 74-92) sets `sp.ActionWithErr(func(ctx) error { for ... postWithRetry(...) })` and then `runSpinnerFn(sp)` runs the spinner. While the spinner is rendering on stdout, `postWithRetry` (lines 154-176) calls `huh.NewForm(huh.NewGroup(prompt)).Run()` for the Retry/Skip/Abort prompt. Both spinner and form are `charm.land/bubbletea/v2` programs (see `charm.land/huh/v2@v2.0.3/spinner/spinner.go:248` — spinner.Run starts a `tea.NewProgram(...)`; same for `huh.Form.Run` in `form.go:687`).

**Why it matters:** two concurrent tea programs on the same TTY cause output corruption, lost keystrokes, and unpredictable cursor behaviour. The interactive prompt is the operator's escape hatch for transient errors; making it unusable in the very scenario it exists for is a real UX regression. Tests don't catch this: `TestMain` swaps `runSpinnerFn` to `WithAccessible(true).WithOutput(io.Discard)`, and no test exercises a transient error in interactive mode that hits the prompt.

**Required change:** stop the spinner before prompting, restart after — or skip the spinner entirely in interactive mode (the loop is a small number of files; per-entry println is fine). Minimum diff:

```diff
-	sp := spinner.New().
-		Title(fmt.Sprintf("Migrating %d entries…", len(plans))).
-		ActionWithErr(func(ctx context.Context) error {
-			for i := range plans {
-				p := plans[i]
-				slug := p.Slug
-				if err := postWithRetry(ctx, cmd, backplaneURL, p, nonInteractive, &res); err != nil {
-					if errors.Is(err, errAborted) {
-						return nil
-					}
-					return fmt.Errorf("entry %s: %w", slug, err)
-				}
-			}
-			return nil
-		})
-
-	if err := runSpinnerFn(sp); err != nil {
-		return err
-	}
+	ctx := cmd.Context()
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	for i := range plans {
+		p := plans[i]
+		if nonInteractive {
+			// Spinner per entry — no concurrent form runs in this mode.
+			sp := spinner.New().
+				Title(fmt.Sprintf("Migrating %s…", p.Slug)).
+				Context(ctx).
+				ActionWithErr(func(ctx context.Context) error {
+					return postWithRetry(ctx, cmd, backplaneURL, p, true, &res)
+				})
+			if err := runSpinnerFn(sp); err != nil {
+				if errors.Is(err, errAborted) {
+					break
+				}
+				return fmt.Errorf("entry %s: %w", p.Slug, err)
+			}
+		} else {
+			// Interactive: plain progress line so the Retry/Skip/Abort form
+			// owns the TTY when it has to appear.
+			fmt.Fprintf(cmd.OutOrStdout(), "Migrating %s (%d/%d)…\n", p.Slug, i+1, len(plans))
+			if err := postWithRetry(ctx, cmd, backplaneURL, p, false, &res); err != nil {
+				if errors.Is(err, errAborted) {
+					break
+				}
+				return fmt.Errorf("entry %s: %w", p.Slug, err)
+			}
+		}
+	}
```

(Or simpler: drop the spinner entirely and always print a per-entry progress line. The spec said "spinner per file" but the current single-spinner-for-the-batch shape already doesn't match that — there's no per-tick visualisation. Plain progress lines preserve the operator-visible "what's happening now" signal without the TTY contention.)

**Source:** `charm.land/huh/v2@v2.0.3/spinner/spinner.go:226-258` (Spinner.Run → tea.NewProgram); same library's `form.go:687` (Form.Run → tea.NewProgram).

cli/internal/cmd/migrate/submit.go
```diff
@@ -0,0 +1,387 @@
...
+		attempt++
+		_, postErr := doAuthedRequest(ctx, backplaneURL, http.MethodPost, "/api/v1/memory", raw)
+		if postErr == nil {
+			res.Migrated++
+			return nil
+		}
+
+		if !isTransient(postErr) {
+			res.Errored++
+			fmt.Fprintf(cmd.ErrOrStderr(),
```
damir-topic: **[⚠️ Major]** — permanent errors silently skip with `return nil`; auth failures look like a successful run (exit 0)

**What:** when `doAuthedRequest` returns a non-transient error (HTTP 401/403/404/422, or anything `isTransient` returns false for), `postWithRetry` (lines 134-138) prints to stderr, increments `res.Errored`, and `return nil`. `doSubmit` (line 94) then prints the summary line and returns `nil`. So even when every entry fails because the token is expired and refresh failed, the command exits 0.

**Why it matters:** (1) the docs at `docs/cli/memory-migration.md:152` tell the operator "A 401 means your token has expired — run `meho login` again" — but the CLI doesn't surface that distinction in either the exit code or via `renderSubmitError`, which is currently only reachable from the URL-resolve path at line 67. (2) Automation that runs `meho migrate memory --non-interactive` in a cron will see exit 0 even when the run errored every entry — silent failure of a sync job. (3) The kb / audit / topology siblings exit with `auth_expired (exit 2)` for the same condition; T5 should match.

**Required change:** route HTTP errors through `renderSubmitError` and let `doSubmit` return a non-nil error when all (or any) entries failed permanently. Two sketches:

(a) hardest stop — break out of the loop on the first permanent error and let `renderSubmitError` exit-code-classify it:

```diff
 		if !isTransient(postErr) {
-			res.Errored++
-			fmt.Fprintf(cmd.ErrOrStderr(),
-				"meho migrate memory: permanent error for %s: %v\n", plan.Slug, postErr)
-			return nil
+			res.Errored++
+			return postErr // bubbles to doSubmit → renderSubmitError → correct exit code
 		}
```

…and in `doSubmit`:

```diff
-		if err := postWithRetry(ctx, cmd, backplaneURL, p, nonInteractive, &res); err != nil {
+		if err := postWithRetry(ctx, cmd, backplaneURL, p, nonInteractive, &res); err != nil {
 			if errors.Is(err, errAborted) {
 				return nil
 			}
-			return fmt.Errorf("entry %s: %w", slug, err)
+			return renderSubmitError(cmd, backplaneURL, err)
 		}
```

(b) softer — keep going past per-entry permanent errors but exit non-zero with a classified error at the end:

```diff
+	if res.Errored > 0 {
+		return output.RenderError(cmd.ErrOrStderr(),
+			output.Unexpected(fmt.Sprintf("%d entries failed", res.Errored)),
+			false,
+		)
+	}
 	return nil
```

(a) is closer to kb's semantics (a 401 on any entry stops the loop) and is the smaller diff. (b) is gentler if the team prefers "migrate what we can, fail the run." Either works — pick one explicitly. Currently it's neither.

cli/internal/cmd/migrate/memory.go
```diff
@@ -96,11 +98,11 @@ func newMemoryCmdWithSubmit(submitFn SubmitFn) *cobra.Command {
 						)
 					}
 				}
-				if err := submitFn(plans); err != nil {
+				if err := eff(plans); err != nil {
 					return err
 				}
```
damir-topic, line 105: **[⚠️ Major]** — `--mark-migrated` touches the marker even when zero entries migrated; silences the nudge after a failed run

**What:** in the `--non-interactive` path (lines 100-107), `submitFn` (now `eff` calling `doSubmit`) always returns nil even when every entry errored. So the `markMigrated` branch unconditionally calls `TouchMarker(dir)` and the post-login nudge is silenced. The interactive path at line 131-133 has the same shape.

**Why it matters:** the marker exists precisely to silence the nudge once "the operator has actually migrated their files." Touching it after a run that migrated nothing (e.g. all entries 401'd, or every entry was transient-retried out) lies to the next login. The operator loses the reminder that they still need to migrate.

This is the same consent-class issue as the T4 fix I just landed on #680 (PR #680's `a9b80b9`), which moved `TouchMarker` after the refused-entries check. T5 needs the analogous fix — gate on `res.Errored == 0 && res.Migrated > 0` (or similar) before touching.

**Required change:** thread the submit result back to memory.go's RunE and gate the `TouchMarker` call:

```diff
-eff := submitFn
-if eff == nil {
-    eff = func(plans []migrate.SubmitPlan) error {
-        return doSubmit(cmd, backplaneOverride, plans)
-    }
-}
+// effSubmit returns (any-migrated, error) so the caller can decide
+// whether to touch the marker. submitFn injected by tests returns
+// (len(plans)>0, nil) for the same semantics.
+effSubmit := func(plans []migrate.SubmitPlan) (migrated bool, err error) {
+    if submitFn != nil {
+        return len(plans) > 0, submitFn(plans)
+    }
+    return doSubmit(cmd, backplaneOverride, plans)
+}
```

…with `doSubmit` returning `(res.Migrated > 0, nil)` (or `res.Migrated > 0, errIfAnyPermFailed`). Then in the non-interactive branch:

```diff
-if err := eff(plans); err != nil {
-    return err
-}
-if markMigrated {
-    _ = migrate.TouchMarker(dir)
-}
-if len(refused) > 0 {
-    return fmt.Errorf("meho migrate memory: %d entries require interactive review", len(refused))
-}
+migrated, err := effSubmit(plans)
+if err != nil {
+    return err
+}
+if len(refused) > 0 {
+    return fmt.Errorf("meho migrate memory: %d entries require interactive review", len(refused))
+}
+if markMigrated && migrated {
+    _ = migrate.TouchMarker(dir)
+}
```

Same shape for the interactive branch. If you'd rather not change the `SubmitFn` signature, the equivalent is to expose `submitResult` from `doSubmit` instead of returning a bool — the goal is "don't touch the marker when nothing was migrated."

Worth a focused test: `TestMarkMigrated_NotTouchedOnAllErrored` (httptest returns 401, set `--mark-migrated`, assert marker absent after run).

cli/internal/cmd/migrate/submit.go
```diff
@@ -0,0 +1,387 @@
...
+				"meho migrate memory: permanent error for %s: %v\n", plan.Slug, postErr)
+			return nil
+		}
+
+		// Transient error handling.
+		if nonInteractive {
+			if attempt < maxAutoRetry {
```
damir-topic: **[💡 Minor]** — `maxAutoRetry` name and behaviour disagree

**What:** `const maxAutoRetry = 3`, used at line 143 as `if attempt < maxAutoRetry`. With attempts counted as 1, 2, 3…, the first two attempts retry on transient error, the third skips. So the constant named "max auto retry" effectively caps at *2 retries after the first attempt*, not 3. The PR body and `docs/cli/memory-migration.md:148` say "up to 3 attempts per entry" — which IS what the code does, but `maxAutoRetry` reads as "the max number of retries" not "the max number of attempts."

**Why it matters:** future readers will read `maxAutoRetry = 3` and conclude "3 retries after the initial = 4 total attempts" — off by one from reality. Easy to fix.

**Required change:** rename for clarity. The smaller-diff form:

```diff
-const maxAutoRetry = 3
+// maxAttempts is the total number of POST attempts per entry in
+// --non-interactive mode (including the first one). Two transient
+// retries beyond the initial attempt = 3 total.
+const maxAttempts = 3
```

…and update the conditional:

```diff
-		if attempt < maxAutoRetry {
+		if attempt < maxAttempts {
 			res.Retried++
 			continue
 		}
```

(or `if attempt < maxAttempts - 1` and rename to `maxRetries` — either is fine, just pick the one whose name matches the conditional).

damir-topic (dismissed):
## /review-pr-slim summary

**Anchor task:** #612 — G5.3-T5 submission layer + post-login nudge + marker file + docs
**Surface area:** backend (Go CLI), docs
**Best-practices packs audited:** ai_engineering (§Safety — operator-consent / silencing nudges), Go idiom, devops (idempotent re-run)
**Verdict:** REQUEST_CHANGES

The T5 surface is mostly there — `submit.go` wires the in-package HTTP helper trio per the kb convention, `marker.go` is a clean XDG implementation, `login.go` adds a non-fatal nudge with focused tests, and the docs file is well-organised. What blocks merge: two of the four `type:` rows in the docs' scope-suggestion table contradict the actual `SuggestScope` code (silent operator-facing misinformation); `errSkip` + a trailing-newline in `memory.go` keep the Go lane red; the interactive Retry/Skip/Abort path runs `huh.NewForm(...).Run()` inside the spinner's `ActionWithErr` so the two tea programs race the TTY in production (tests bypass it via accessible+discard); permanent errors silently skip with exit 0 (auth failures look like success); and `--mark-migrated` touches the marker even when every entry errored.

### Acceptance criteria (#612)
- [x] POSTs `{scope, slug, body, metadata, source_id}` per entry — verified (`TestSubmit_PostsCorrectBody`).
- [x] same body re-run = same `source_id` — verified (`TestSubmit_SourceIDStable`, `TestSubmit_SameBodyRerun`).
- [x] changed body = same `source_id`-prefix scheme (per-body SHA-256, 12 hex) — verified (`TestSubmit_ChangedBody_DifferentSourceID`).
- [x] transient 503 then 200 → retry success — verified (`TestSubmit_TransientRetryThenSuccess`).
- [x] post-login nudge: tip when marker absent + dir non-empty; silent when marker present, dir empty, or dir resolve errors — verified (`TestPrintMigrationNudge_*`).
- [x] `meho migrate memory --mark-migrated` writes the marker — verified (`TestMarkMigrated_WritesMarker`).
- [⚠️] `docs/cli/memory-migration.md` exists and documents the heuristics + table + matrix + opt-out — **but two scope-table rows are wrong** (Blocker 1).
- [x] no `cli/internal/api` import in `cli/internal/migrate/` packages — verified (`submit.go` is in `cli/internal/cmd/migrate/`, where the import is allowed per task wording).
- [ ] `go vet` clean; `gofmt -d` empty; CI green — ❌ Two `golangci-lint` failures (Blocker 2 + Major 1).
- [⚠️] CLI cross-compiles for the four targets — not verified locally; CI's pre-existing Go-lane red blocks confirmation regardless.

### Project-board alignment (evoila/meho)
| Issue | State | Link | Auto-close on merge | Delivered | Action |
|---|---|---|---|---|---|
| #612 (T5) | OPEN, In Progress | `Closes #612` | ✅ once base is `main` | ⚠️ partial (5 findings) | fix blockers + majors, then close on merge |
| #375 (Initiative) | OPEN | parent | — | T5 was the last sibling — closing this completes G5.3 | bump status on merge |
| #216 (G5 Goal) | OPEN | grandparent | — | G5.3 was the final G5 lane | check checklist on merge |

PR base is `feat/611-huh-picker` not `main`. Once #680 merges, this PR's base should switch to `main` (auto-close requires same-repo default-branch base — currently it won't fire). I've been working on #680 in parallel; pushing this review while CI on #680 completes.

### Predicted vs. actual diff
- ✅ `submit.go`, `submit_test.go`, `marker.go`, `marker_test.go`, `login.go`, `login_test.go`, `docs/cli/memory-migration.md`, `docs/codebase/cli.md` — expected T5 surface.
- ✅ `picker.go` moves `SourceIDPrefix` const from `cmd/migrate` → `internal/migrate` so both paths share one source of truth. Correct call.
- ⚠️ `memory.go` — wired `eff` (real `doSubmit` for production, injected `submitFn` for tests). Functional, but the `--mark-migrated` ordering bug (Major 4) returns.
- ⚠️ `submit.go:postWithRetry` interactive prompt placement (Major 2) — expected but wrong.

### Integration verification (in `.claude/worktrees/pr-685-review/cli`, head `017d0125`)
- [x] `go vet ./...` — clean
- [❌] `gofmt -d ./internal/cmd/migrate` — flags `memory.go:220` trailing newline (Major 1)
- [x] `go test ./internal/migrate/... ./internal/cmd/migrate/... ./internal/cmd/` — all pass
- [❌] `golangci-lint run ./internal/cmd/migrate/...` — 2 issues: `submit.go:99 var errSkip is unused` + `memory.go:220 gofmt` (Blocker 2 + Major 1)
- [⚠️ not run] interactive spinner + prompt exercised in TTY — not feasible headless; reasoned through against vendored `charm.land/huh/v2@v2.0.3/spinner/spinner.go:226` (action runs as a tea program / goroutine concurrent with the spinner) → Major 2

- 🛑 2 blockers — docs scope-table wrong vs. code (`docs/cli/memory-migration.md:80-81`); `errSkip` unused fails `golangci-lint` (`submit.go:99`)
- ⚠️ 4 majors — `memory.go:220` trailing newline gofmt; interactive prompt inside live spinner (`submit.go:154-176`); permanent errors silently skip → exit 0 masks auth failures (`submit.go:134-138`); `--mark-migrated` touches marker even on all-errored runs (`memory.go:103-107`)
- 💡 1 minor — `maxAutoRetry` off-by-one: name suggests 3 retries, behaviour is 2 attempts after the first then skip (`submit.go:55, 143`)

- `renderSubmitError` is wired only for the URL-resolve path (`submit.go:67`); all the auth_expired branches live for the path that's never reached because POST errors bubble through `postWithRetry`. Once Major 3 is fixed, the dead branches become live — no separate work needed, but it's worth noting that the current shape carried auth-classification logic without actually using it.
- `submit.go` declares an in-package `httpError` but does not export it. That's fine; consistent with kb's pattern. Just flagging in case a future review thinks it should be shared.
- `sanitizeDirName` docstring says "absolute path" but the helper handles relative paths too (`{"relative/path", "relative_path"}` test passes); either tighten the docstring or rename to `sanitizeAnyPath`. Cosmetic.

### Docs
- [⚠️] `docs/cli/memory-migration.md` — covers everything the task asks for, well-organised, cross-links to #427 for the 5-scope model correctly. **But the scope-suggestion table at lines 80-81 contradicts the actual `SuggestScope` code in two rows** — see Blocker 1. No .NET analogies, no `.planning/` refs.
- [x] `docs/codebase/cli.md` — updated for the new T5 files; accurate against the diff.

### What worked well
- **`SourceIDPrefix` promotion** to `cli/internal/migrate/` so dry-run + POST share one source of truth — exactly the right move; eliminates the bit-for-bit drift risk T4 carried.
- **Marker implementation** (`TouchMarker` / `MarkerExists` / `sanitizeDirName`) is clean, XDG-aware with a `$HOME/.config` fallback, and **degrades gracefully** on any I/O error so the post-login nudge can never fail login. The `nolint:nilerr` comments on the swallow paths are honest about what's happening.
- **Post-login nudge tests** cover the four states (files+no-marker, files+marker, empty dir, dir-resolve-error) crisply via `seedNudgeDir` + `t.Setenv` + `httptest`-style isolation. Good shape to copy elsewhere.
- **Test isolation discipline** — `TestMain` swaps `runSpinnerFn` to accessible+discard so the spinner doesn't open `/dev/tty` in CI; `seedXDGAndToken` mirrors the kb test helper. The seam choice (`runSpinnerFn`) is the right one given huh's spinner doesn't expose a "test mode" directly.
- **In-package helper trio convention** followed — `resolveBackplaneURL` / `doAuthedRequest` / `renderSubmitError` mirror the kb pattern. The duplication is acknowledged in a comment and is consistent with `audit/` / `topology/` / `operation/`.
- **Auth refresh wired** — `doAuthedRequest:333-344` re-tries on 401 after `authed.Refresh(ctx)`, matching the kb convention.

**Auto-close** (requires base=`main` first)
- #612 — verify CLOSED and `closedByPullRequestsReferences` contains this PR.

- #612 → Done
- #375 (Initiative G5.3) → Done — T5 was the final sibling
- #216 (Goal G5) — verify G5.3 checklist line is ticked

- #612: "Closed by #685. T5 submission layer + nudge + marker + docs shipped. Verified: POST body shape, source_id stability, transient-retry path, nudge gating, marker touch."
- #375: "Closed by #685. G5.3 complete — T1 #608, T2 #609 (via #680), T3 #610, T4 #611 (#680), T5 #612 (#685)."

PR #746, opened 2026-05-20
feat(api): G0.9-T2 extra="forbid" on every public v1 request schema (#729)

damir-topic:
## Summary

- Flips every public v1 **request** schema to `model_config = ConfigDict(extra="forbid", ...)` so unknown keys fail 422 `extra_forbidden` at the framework boundary instead of being silently dropped (Pydantic v2's default `extra="ignore"`).
- Fixes Signal #2 from the 2026-05-20 RDC in-lab dogfood: a v0.2.1 client sending `{"q": "vault", "top_k": 3}` to `/api/v1/retrieve` used to get a confusing missing-`query` error; now it gets a clear 422 with one `extra_forbidden` per unknown field.
- Adds a layered regression test (`tests/test_api_v1_schema_extras.py`): parametrised schema-level coverage for every touched request model + an integration test exercising the exact Signal #2 payload through `/api/v1/retrieve`.

## Inventory of v1 request schemas

Every `BaseModel` backing a public v1 POST/PATCH/PUT body, plus `CallOperationBody` outside `api/v1/` per the issue's "any other request-body schema" line:

| Schema | File | Pre-existing | This PR |
| --- | --- | --- | --- |
| `RetrieveRequest` | `backend/src/meho_backplane/api/v1/retrieve.py` | `frozen=True` | + `extra="forbid"` |
| `AuditQueryRequest` | `backend/src/meho_backplane/api/v1/audit_models.py` | `frozen=True` | + `extra="forbid"` |
| `CallOperationBody` | `backend/src/meho_backplane/operations/meta_tools.py` | (none) | new `ConfigDict(extra="forbid")` |
| `IngestRequest` | `backend/src/meho_backplane/operations/ingest/api_schemas.py` | `frozen=True` | + `extra="forbid"` |
| `SpecSource` (nested in `IngestRequest.specs`) | same | `frozen=True` | + `extra="forbid"` |
| `EditGroupBody` | same | `frozen=True` | + `extra="forbid"` |
| `EditOpBody` | same | `frozen=True` | + `extra="forbid"` |
| `EvalRequest` | `backend/src/meho_backplane/api/v1/retrieve_eval.py` | `frozen=True, extra="forbid"` | unchanged ✓ |
| `RetireChecklistRequest` | `backend/src/meho_backplane/api/v1/retrieve_retire.py` | `frozen=True, extra="forbid"` | unchanged ✓ |
[...]

Issue #729, opened 2026-05-20
Task: harden public v1 request schemas with extra="forbid" (Signal #2)


damir-topic:
Parent goal: #221
Parent initiative: #737

## Problem

Field renames in v0.3.0's public REST schemas (`/api/v1/retrieve` `q` → `query`; `/api/v1/operations/call` `target: str` → `target: dict`) silently dropped old field names instead of failing loud with 422. Default Pydantic v2 `extra="ignore"` means client-side typos and rename drift pass validation and the request executes with defaults (or required-field-missing errors that are weaker than they could be).

Surfaced by the RDC operator team after the v0.3.0 dogfood; their `top_k`/`q` v0.2.1 client calls produced confusing behaviour rather than a clear "you sent an unknown field" error.

## Root cause

Every public v1 request body subclass of `BaseModel` declares fields but never sets `extra` in its `model_config`. Default is `"ignore"`, which silently drops unknown keys. Examples:

- [backend/src/meho_backplane/api/v1/retrieve.py:115](backend/src/meho_backplane/api/v1/retrieve.py#L115) — `RetrieveRequest.model_config = ConfigDict(frozen=True)` (no `extra`)
- [backend/src/meho_backplane/operations/meta_tools.py:216](backend/src/meho_backplane/operations/meta_tools.py#L216) — `CallOperationBody(BaseModel)` (no `model_config` at all → all defaults)
- Same shape across every public v1 request schema

## Current state

Pre-existing schemas use Pydantic defaults. Future renames have no guard.

## Desired state

Every public v1 **request** schema sets `model_config = ConfigDict(extra="forbid", frozen=True)` (frozen where already set; `extra="forbid"` everywhere). Unknown fields → 422 with `extra_forbidden` detail. Response schemas keep their current shape (they're not validated against client input).

## Acceptance criteria

- [ ] Every `BaseModel` subclass that backs a `POST` / `PATCH` / `PUT` body in [backend/src/meho_backplane/api/v1/](backend/src/meho_backplane/api/v1/) sets `model_config = ConfigDict(extra="forbid", ...)`. Inventory commit-side or report inventory + diff in PR description
- [ ] Same for `CallOperationBody` in [meta_tools.py](backend/src/meho_backplane/operations/meta_tools.py) and any other request-body schema outside `api/v1/`
[...]

PR #767, opened 2026-05-21
fix(test): g81 audit acceptance asserts 422 for body tenant_id (extra=forbid)

damir-topic:
## Summary

`test_g81_audit_query_acceptance.py::test_scenario_1_body_tenant_id_is_silently_dropped` was asserting pre-G0.9-T2 behaviour (body `tenant_id` silently dropped → 200). [#746](https://github.com/evoila/meho/pull/746) (G0.9-T2, #729) set `extra="forbid"` on every public v1 request schema, so the field now → 422 `extra_forbidden`. The test's premise is obsolete; **main is currently red on it.**

- Rewrite to assert the new contract: body `tenant_id` → 422 with `extra_forbidden` on `body.tenant_id`.
- Rename `_silently_dropped` → `_rejected_422`.
- Update the module docstring's scenario-1 line.
- Drop the now-moot `seed_tenants_with_overlap` call (request rejected at validation before the query handler runs); import stays for sibling tests.

The tenant-isolation security property is unchanged and stronger — a cross-tenant override attempt now fails loudly at the validation boundary rather than being quietly ignored.

## Why main was red without anyone noticing

The break landed via #746 but never surfaced on a green main run: the `ci-main` concurrency group's `cancel-in-progress: true` cancels post-merge main CI when the next PR merges, so the combined-state breakage stayed invisible until a later PR rebased onto the red base. This is the third such inherited-red in two days (harbor + sddc-manager regex breaks were the first two). A **merge queue** would prevent the class — flagged separately.

## Test plan

- [x] `pytest ...::test_scenario_1_body_tenant_id_is_rejected_422` with real PG + valkey testcontainers — 1 passed
- [ ] CI green on this PR (unblocks main)

## Summary by CodeRabbit

* **Bug Fixes**
  * The audit-query endpoint now enforces stricter tenant boundary validation. Requests containing client-supplied cross-tenant identifiers are properly rejected with validation error responses instead of being silently ignored. This improves security by preventing unauthorized cross-tenant access attempts and provides clearer error feedback.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/767?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/tests/acceptance/test_g81_audit_query_acceptance.py
+++ b/backend/tests/acceptance/test_g81_audit_query_acceptance.py
@@ -14,7 +14,8 @@ Coverage matrix (mirrors the 8 scenarios in the issue body #469):
 1. **Tenant boundary** — overlapping target names / principal subs
    across two tenants; tenant-A operator's REST surface never returns
    tenant-B rows, returns 404 (not 403) on tenant-B's audit_id,
-   and silently drops a body-supplied cross-tenant ``tenant_id``.
+   and rejects a body-supplied cross-tenant ``tenant_id`` with 422
+   (``extra="forbid"`` per G0.9-T2 #729).
 2. **Cursor pagination correctness** — 250 rows in tenant A, paged
    at limit=100 across three pages; ``asyncio.gather`` runs a
    concurrent 50-row insert against page-2-read using the page-1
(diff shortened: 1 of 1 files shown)


damir-topic:
@coderabbitai review

Issue #771, opened 2026-05-21
Task: CI — MCP test fixtures boot full app lifespan (+embedding preload) per test; scope it down


damir-topic:
## Context

The #761 heavy-pool experiment (PR #765) showed doubling xdist workers (3→6) bought only ~10% wall, not ~2x — the unit job doesn't parallelize. A `--durations=25` diagnostic (full unit suite, `-n 6 --dist loadscope`, local, against main HEAD) located the real bottleneck, and it is **not** the testcontainer/db tests I originally hypothesized for an integration-split. It is the **MCP test fixtures booting the full production app lifespan — including the fastembed ONNX model preload — on every test.**

## Evidence

`--durations=25` (local run, 2854 passed / 47 skipped / `645.70s` = 10:45 wall with `-n 6`):

- The slowest **25 setups are all 22-25s**, dominated by `test_mcp_tools_*` modules (`test_mcp_tools_memory`, `test_mcp_tools_topology`, `test_mcp_broadcast_overrides_tools`, `test_mcp_tools_connector_admin`, `test_mcp_tools_operations`, `test_mcp_tool_query_audit`, `test_mcp_registry`, …) plus `test_g07_vsphere_canary` setup at 44s.
- The **same module appears 3×** in the slowest list (e.g. `test_mcp_tools_memory.py`, `test_mcp_tools_topology_annotate.py`) — proving the expensive fixture is **function-scoped (re-runs per test)**, not module/session-scoped.
- Local wall 10:45 vs CI 21:50 for the same `-n 6` — CI is ~2× slower per test, consistent with cold per-worker model/onnxruntime cost on the ARC pods amplifying the same root cause.

## Root cause

`client_with_operator` ([`backend/tests/mcp_test_fixtures.py:170`](backend/tests/mcp_test_fixtures.py#L170)) is **function-scoped** and enters `with TestClient(app)` where `app` is the **full production app** (`from meho_backplane.main import app`, line 54). The `with` runs the complete FastAPI lifespan ([`backend/src/meho_backplane/main.py:185`](backend/src/meho_backplane/main.py#L185)) on **every test**:

- `_eager_import_connectors()` (243)
- `run_typed_op_registrars()` (248)
- `eager_import_mcp_modules()` (252)
[...]

PR #793, opened 2026-05-21
DIAGNOSTIC (do not merge): per-lifespan-step + fixture timing for #771

damir-topic:
## DIAGNOSTIC — DO NOT MERGE

Definitive CI-side attribution for the #771 unit-suite wall-time
investigation. After six local mis-diagnoses, the goal here is to stop
guessing and let **one instrumented CI run** say exactly where the time
goes — because the cost does not reproduce locally.

### What the baseline already proves

The latest PR-path run of the unit job (a `pull_request` event, so
coverage is **off** — #726 already gated `--cov` to push-to-main) spent
**22m19s purely in `pytest`** under `-n 6` on the heavy pool. So the
wall is **not** coverage instrumentation.

`--durations=25` on that run blames the 25 slowest items on
`test_mcp_*` **setup**, 35–47s each — and crucially, **multiple tests
within the same module** appear. Under `--dist loadscope` a module's
higher-scoped fixtures are set up once per worker, so a per-module cost
would show each module once. Multiple hits per module ⇒ the 35–47s is a
**function-scoped** cost paid **per test** (the full FastAPI lifespan
runs every test via `client_with_operator`'s `with TestClient(app)`).

The one non-setup outlier — `42.26s call test_auth_config ::
test_auth_config_normalises_trailing_slash_on_issuer` — is the
signature of a **network connection hanging to timeout** on the request
path (real auth, no override, `.evba.lab` hosts). Same suspected
mechanism, different entry point.

### What this PR adds (env-gated, test-only, zero production change)

- `tests/conftest.py`: two no-op-unless-enabled hooks —
  - `MEHO_LIFESPAN_TIMING` → times each FastAPI lifespan step
    (`run_typed_op_registrars`, `_preload_embedding_model`,
    `get_broadcast_client`, `start_topology_refresh_scheduler`, …) by
    wrapping `meho_backplane.main`'s module-level callables.
  - `MEHO_FIXTURE_TIMING_FILE` → per-fixture setup duration (outer
    attribution: which fixture's setup the `--durations` report blames).
- `tests/_diag_aggregate.py`: prints both attribution tables.
- `.github/workflows/ci.yml` (PR branch only): sets the two env vars,
  runs the full suite, then prints the aggregate.

Both hooks are pure no-ops when the env vars are absent, so `main`
behaves identically.

### What we expect to learn

Which single lifespan step (or fixture) eats the 35–47s per test in CI.
If it confirms a network-bound step hanging to timeout, the fix is a
real, scoped change (short timeout / skip-in-tests / lazy-on-first-use)
— filed separately. **This branch is then closed, not merged.**

Refs #771

## Summary by CodeRabbit

* **Chores**
  * Enhanced CI diagnostic capabilities to collect and aggregate performance timing metrics during test runs, improving insight into system performance characteristics and test execution efficiency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/793?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -308,7 +308,15 @@ jobs:
             COVERAGE_CORE=sysmon uv run pytest -n 6 --dist loadscope --maxfail=1 --ignore=tests/integration \
               --cov=meho_backplane --cov-report=xml tests/
           else
-            uv run pytest -n 6 --dist loadscope --maxfail=1 --ignore=tests/integration tests/
+            # DIAGNOSTIC (#771, branch diag/ci-lifespan-timing, DO NOT MERGE):
+            # env-gated per-lifespan-step + per-fixture setup attribution, then
+            # aggregate. --maxfail dropped so the full suite yields complete
+            # timing; || true so the aggregator always runs.
+            export MEHO_LIFESPAN_TIMING=/tmp/lifespan
+            export MEHO_FIXTURE_TIMING_FILE=/tmp/fixt
+            uv run pytest -n 6 --dist loadscope --durations=30 --ignore=tests/integration tests/ || true
+            echo "######################## DIAGNOSTIC TIMING ########################"
+            uv run python tests/_diag_aggregate.py
           fi
 
       - name: Upload Python coverage artifact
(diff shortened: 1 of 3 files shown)


damir-topic:
Diagnostic complete. Root cause attributed definitively: per-test descriptor re-embedding in `run_typed_op_registrars` (mean 23s/test, 4253s cumulative) — not coverage, not model load, not containers. Fix in #799 dropped the unit-job pytest step **22m35s → 7m35s**. Closing — diagnostic served its purpose, not for merge.

PR #900, opened 2026-05-22
feat(topology): G9.3-T1 graph_node_history + graph_edge_history (Alembic 0012 + ORM)

zdamir:
## Summary

- Adds Alembic migration `0012_create_topology_history.py` -- two application-managed append-only history tables (`graph_node_history` + `graph_edge_history`) mirroring the live `graph_node` / `graph_edge` pair from `0007`, with temporal columns and an `audit_id` soft-link back to `audit_log.id`. Reversible.
- Adds SQLAlchemy 2.x ORM models `GraphNodeHistory`, `GraphEdgeHistory`, and the closed-vocabulary `GraphHistoryChangeKind` StrEnum to `backend/src/meho_backplane/db/models.py`, mirroring the chassis's `AuditLog` append-only recipe.
- Records the ADR-style rationale (application-managed history vs. PG system-period extensions) in the migration docstring -- decision #6 rejects extension-based temporal versioning for v0.2.
- Ships the schema half of Initiative #365 (G9.3 Discovery audit log -> graph history). The diff-on-write hook (T2 #857), temporal-query verbs (T3/T4/T5 #859/#860/#861), and retention prune task (T6 #858) are out of scope here.

## Test plan

- [x] `ruff check` -- clean on all touched files
- [x] `ruff format --check` -- clean
- [x] `mypy src/meho_backplane/` -- 220 source files, no issues
- [x] `pytest tests/test_topology_history_migration.py` -- **15 passed** (upgrade/downgrade round-trip back to `0011` and re-upgrade; index presence on both tables; DESC ordering on every `valid_from` index via `sqlite_master.sql` reflection; partial-WHERE `change_kind = 'removed'` predicate; CHECK constraint accepts the closed three-kind vocabulary and rejects out-of-vocabulary values; drift guard between `GraphHistoryChangeKind` and `_GRAPH_HISTORY_CHANGE_KINDS`; ORM round-trip on both tables; `ON DELETE SET NULL` clears `node_id` / `edge_id` on live-row deletion while every other field survives)
- [x] `pytest tests/test_topology_schema.py tests/test_alembic_probe.py tests/test_migration_compat.py tests/test_migration_rollback.py tests/test_db_models.py` -- 69 passed (no regressions on the live graph schema, the migration probe, or the forward-compat rollback test)
- [x] `scripts/ci/check_migration_compat.py` -- exit 0 (additive-upgrade discipline preserved)
- [x] `.claude/hooks/code-quality.py --diff` -- 0 blockers
[...]

Issue #859, opened 2026-05-21
Task: G9.3-T3 meho topology history verb + GET /topology/history + query_topology(kind=history) facet


damir-topic:
Parent initiative: #365
Parent goal: #220
Depends-on: #857 (history rows must be written before a read verb can return them)

## Target

- Add: `meho topology history` verb in `cli/internal/cmd/topology/` (mirror `dependencies.go`/`path.go`).
- Add: `GET /api/v1/topology/history` in [`backend/src/meho_backplane/api/v1/topology.py`](../blob/main/backend/src/meho_backplane/api/v1/topology.py).
- Add: `query_topology(kind="history", resource=...)` facet in [`backend/src/meho_backplane/mcp/tools/topology.py`](../blob/main/backend/src/meho_backplane/mcp/tools/topology.py).
- Add: query handler in [`backend/src/meho_backplane/topology/query.py`](../blob/main/backend/src/meho_backplane/topology/query.py).
- Add: `backend/tests/test_topology_history_query.py`.
- Precedent: `topology/query.py` + `api/v1/topology.py` + `mcp/tools/topology.py` (G9.1 dependents/dependencies/path verbs), `cli/internal/cmd/topology/dependencies.go`.

## Context

Per #365 work-items #3 + #7, the temporal read lands on the existing `query_topology` **parametric** meta-tool as `kind="history"` (no new MCP tools — postulate 5); CLI `meho topology history` is the operator convenience resolving to that. Operator role (read); tenant-scoped; emits one `audit_log` row per call with `op_class="audit_query"` (decision #3).

## Current state

History rows are written (#857) but unreadable — no history query verb/route/MCP facet.

## Desired state

`meho topology history <node-name|alias> [--since <ts>] [--until <ts>] [--include-edges] [--json]`: resolves node by name+alias (G9.1 resolver `topology/resolvers.py`), returns its history rows over the window; `--include-edges` joins `*_history` on `node_id IN (from,to)`; default structured table (`change_kind` + `valid_from` + 1-line diff summary), `--json` carries full `snapshot.before`/`after`. `GET /api/v1/topology/history` mirrors. `query_topology(kind="history", resource, since?, until?, include_edges?)` facet with the temporal use case in its parametric description. Tenant-scoped.

## Acceptance criteria

- [ ] `meho topology history <node> --since <ts>` returns the node's chronology; `--include-edges` adds incident edges; `--json` carries `snapshot.before`/`after` (e2e against a real backplane).
[...]

Issue #858, opened 2026-05-21
Task: G9.3-T6 History retention prune task (TOPOLOGY_HISTORY_RETENTION_DAYS, weekly, summary audit row, 0=keep-forever)


damir-topic:
Parent initiative: #365
Parent goal: #220
Depends-on: #856 (history tables must exist before a prune task can target them)

## Target

- Add: a weekly background prune task (APScheduler — the scheduler G9.1 refresh uses, [`topology/scheduler.py`](../blob/main/backend/src/meho_backplane/topology/scheduler.py)) deleting history rows older than `TOPOLOGY_HISTORY_RETENTION_DAYS`.
- Add: `TOPOLOGY_HISTORY_RETENTION_DAYS` setting (default 90; `0` = keep forever) + Helm knob.
- Add: `backend/tests/test_topology_history_retention.py`.
- Precedent: `topology/scheduler.py` (G9.1 APScheduler refresh task, [#450](https://github.com/evoila/meho/issues/450)); `MemoryExpirySweeper` lifespan task ([#623](https://github.com/evoila/meho/issues/623) G5.2-T1) for the prune-with-audit-row pattern.

## Context

Per #365 work-item #8, history grows unbounded without a retention policy. A weekly task deletes rows where `valid_from < now() - retention_window`, writes **one** audit row per run (`op_id='topology.history.prune'`, `op_class='write'`) noting rows dropped; broadcast carries the deletion count. Opt-out via `TOPOLOGY_HISTORY_RETENTION_DAYS=0` (keep forever — disk-growth tradeoff flagged in onboarding docs).

## Current state

History rows (after #856 / T2) accumulate with no pruning.

## Desired state

A scheduled prune task (default weekly) runs the bounded DELETE, emits the summary audit row, honors the setting incl. `0`=disabled. This is the **only** place history rows are deleted (the tables are otherwise append-only).

## Acceptance criteria

- [ ] The prune task deletes rows with `valid_from < now() - TOPOLOGY_HISTORY_RETENTION_DAYS` and leaves newer rows (test with seeded old+new rows).
- [ ] `TOPOLOGY_HISTORY_RETENTION_DAYS=0` → prune is a no-op (keep forever); asserted.
- [ ] Each run writes exactly one audit row (`op_id='topology.history.prune'`, `op_class='write'`) with the dropped-row count in `payload`.
- [ ] The setting + Helm knob are documented.
[...]

Issue #856, opened 2026-05-21
Task: G9.3-T1 Alembic migration — graph_node_history + graph_edge_history (append-only, temporal indexes incl. partial removed) + ORM models


damir-topic:
Parent initiative: #365
Parent goal: #220
Depends-on: (none — G9.1 graph_node/graph_edge tables shipped, #448 closed)

## Target

- Add: one Alembic migration under `backend/alembic/versions/` (next sequential after `0007_create_topology_graph.py` / `0008_*`) creating `graph_node_history` + `graph_edge_history`.
- Add: ORM models for both tables in `backend/src/meho_backplane/db/models.py` (mirror the `AuditLog` append-only shape at [models.py:331](../blob/main/backend/src/meho_backplane/db/models.py#L331); `graph_node` at L1300, `graph_edge` at L1418).
- Add: `backend/tests/test_topology_history_migration.py` (upgrade/downgrade round-trip + index presence).
- Precedent: G9.1-T1 migration [#448](https://github.com/evoila/meho/issues/448) (`graph_node`/`graph_edge`), `0007_create_topology_graph.py`, `AuditLog` model.

## Context

Per #365 work-item #1, topology history is **application-managed append-only tables** (Postgres has no first-class system-versioning). Two tables mirror the live `graph_node`/`graph_edge` with temporal columns. The chassis `AuditLog` (models.py:331) is the recipe: tenant-scoped, indexed by tenant+time, JSONB payload, append-only. This Task ships **only** the schema + ORM; the diff-on-write hook is T2.

## Current state

`graph_node` + `graph_edge` exist (G9.1). No history tables; no temporal query is possible.

## Desired state

[...]

PR #904, opened 2026-05-22
feat(topology): G9.3-T2 diff-on-write hook — *_history rows + audit_id in refresh + annotate

### LINKED ISSUE #857: Task: G9.3-T2 Diff-on-write hook — same-txn *_history rows + audit_id linkage in topology refresh + annotate paths
Parent initiative: #365
Parent goal: #220
Depends-on: #856 (history tables + ORM must exist before the hook can write to them)

## Target

- Modify: [`backend/src/meho_backplane/topology/refresh.py`](../blob/main/backend/src/meho_backplane/topology/refresh.py) (every `graph_node`/`graph_edge` INSERT/UPDATE/soft-remove emits a paired `*_history` INSERT in the same transaction) + [`backend/src/meho_backplane/topology/annotate.py`](../blob/main/backend/src/meho_backplane/topology/annotate.py) (annotate/unannotate handlers).
- Add: a shared history-writer helper in `backend/src/meho_backplane/topology/history.py` used by both call sites.
- Add: `backend/tests/test_topology_history_hook.py`.
- Precedent: `topology/refresh.py` (G9.1 diff/apply reconciliation, [#450](https://github.com/evoila/meho/issues/450)), `topology/annotate.py` (G9.2, [#595](https://github.com/evoila/meho/issues/595)), the `audit_id` pre-generation pattern at [`audit.py:208`](../blob/main/backend/src/meho_backplane/audit.py#L208).

## Context

Per #365 work-item #2, every live-state mutation emits exactly one history row in the **same transaction** (atomicity: both land or neither — a partial `topology.refresh` failure cannot leave history desynced). The `audit_id` on the history row is the `audit_id` AuditMiddleware generated for the request (pre-generation pattern). `snapshot` projects `before` (NULL for `created`; full row JSON for `updated`/`removed`) + `after` (full row JSON for `created`/`updated`; NULL for `removed`) — bidirectional, for diff reconstruction. History writes do **not** emit their own audit rows (each already references the causing op's `audit_id`).

## Current state

History tables exist (#856) but nothing writes to them; topology refresh + annotate mutate live tables with no history trail.

## Desired state

A history-writer helper records `(change_kind, snapshot{before,after}, audit_id, tenant_id)` per node/edge mutation, called within the refresh + annotate transactions. refresh: `created`/`updated`/`removed` (soft `last_seen=NULL` → `removed`). annotate/unannotate: `created`/`removed` for curated edges. Same-transaction atomicity guaranteed.

## Acceptance criteria

- [ ] A `topology.refresh` adding 3 nodes + 2 edges produces exactly 5 history rows in the same txn, all sharing the refresh's `audit_id` (integration test).
- [ ] A `topology.annotate` adding 1 curated edge produces 1 history row sharing the annotate's `audit_id`.
- [ ] A refresh dropping a previously-discovered node produces 1 `removed` history row (`snapshot.before` = full row, `after` = NULL) with a NEW `audit_id`.
- [ ] The hook is transactional: a forced failure after the live-table write but before commit leaves NEITHER the live change nor the history row (assert rollback).
- [ ] History writes emit no `audit_log` rows of their own.
- [ ] `ruff` + `mypy` clean; `pytest -n auto backend/tests/test_topology_history_hook.py` passes.

## Out of scope

- History tables/migration — #856. Query verbs — T3/T4/T5. Retention — T6.

## References

- Precedent: `topology/refresh.py` ([#450](https://github.com/evoila/meho/issues/450)), `topology/annotate.py` ([#595](https://github.com/evoila/meho/issues/595)), `audit_id` pre-generation [`audit.py:208`](../blob/main/backend/src/meho_backplane/audit.py#L208).
- Pinned: SQLAlchemy 2.x async transactions.
- Best-practices anchors (by pack name): **python_best_practices** — SQLAlchemy 2.x async unit-of-work, same-transaction atomicity; **devops_best_practices** — append-only history discipline.

zdamir:
## Summary

- Wires the live-mutation diff onto the append-only `graph_node_history` + `graph_edge_history` substrate from G9.3-T1 (#856). Every live `graph_node` / `graph_edge` INSERT / UPDATE / soft-remove in the refresh service, and every annotate / unannotate of a curated edge (including the §6 supersede / conflict marker writes), emits a paired history row **in the same transaction** as the live write. Atomicity is load-bearing: a partial-failure rollback can never leave the live graph and the history table disagreeing about which mutations committed.
- Shared writer is `backend/src/meho_backplane/topology/history.py` (`record_node_change` / `record_edge_change` + `node_snapshot` / `edge_snapshot`), projecting a fresh row dict into the `{before, after}` JSONB shape the temporal-query verbs T3/T4/T5 read. Snapshot column selection is deliberately narrow (a constant tuple) so a future derived column does not leak into the history projection.
- Every history row carries the **causing operation's** `audit_log.id` via the soft-FK `audit_id` column. The refresh / annotate service pre-allocates one `audit_id` per request and threads it down to every emitted history row, preserving the chassis's "one operation, one audit row" invariant (acceptance criterion #5). `bulk_import_edges` (G9.2-T8 #600) composes cleanly because it already calls `annotate_edge_in_txn`.

Stacked on #900 (G9.3-T1). Review/merge #900 first.

## Test plan

- [x] `ruff check` -- clean on all touched files
- [x] `ruff format --check` -- clean
- [x] `mypy src/meho_backplane/` -- 221 source files, no issues
- [x] `.claude/hooks/code-quality.py --diff` -- 0 blockers
- [x] `pytest tests/test_topology_history_hook.py` -- **9 passed** (one per acceptance criterion + tenant boundary + §6 marker history + no-op no-emit + iter-2 B2 idempotent-annotate regression)
- [x] `pytest tests/ --ignore=tests/integration` -- **3287 passed, 47 skipped** (no regressions in any topology / annotate / bulk-import / migration suite; acceptance suite re-greens after the T1 TRUNCATE gap fix)
- [x] **AC#1**: a `topology.refresh` adding 3 nodes + 2 edges produces exactly 5 history rows in the same txn, all sharing the refresh's `audit_id` -- proven by `test_refresh_3n2e_emits_5_history_rows_sharing_audit_id`.
- [x] **AC#2**: a `topology.annotate` adding 1 curated edge produces 1 history row sharing the annotate's `audit_id` -- proven by `test_annotate_one_edge_emits_one_edge_history_row_with_shared_audit_id`.
- [x] **AC#3**: a refresh dropping a previously-discovered node produces 1 `removed` history row (`snapshot.before` = full row, `after` = NULL) with a NEW `audit_id` -- proven by `test_refresh_dropping_node_emits_removed_history_with_new_audit_id`.
- [x] **AC#4**: a forced failure after the live-table write but before commit leaves NEITHER the live change nor the history row -- proven by `test_refresh_failure_rolls_back_both_live_and_history_rows`.
- [x] **AC#5**: history writes emit no `audit_log` rows of their own -- proven by the `len(audits) == 1` assertion in the AC#1 test (one refresh = one audit row, 5 history rows reference it).
- [x] **AC**: `pytest -n auto backend/tests/test_topology_history_hook.py` passes (uses the standard per-worker template DB cache).

## Out of scope

- T3/T4/T5 (#859/#860/#861) temporal-query verbs (`meho topology history / diff / timeline`) and the `query_topology(kind=...)` MCP routing -- this PR ships only the write half; reads land on top.
- T6 (#858) retention prune task.
- `topology/nodes.py` (`create_or_get_node` operator-seeded nodes) -- the task body explicitly scopes T2 to the `topology/refresh.py` + `topology/annotate.py` write paths.
- PG-side integration verification (Docker absent in agent sandbox); the unit suite exercises every code path against `sqlite+aiosqlite` via the standard worker-template cache, and `tests/integration/conftest.py`'s TRUNCATE list is updated so the testcontainers PG suite runs cleanly when CI provisions Docker.

## Adjacent findings (recorded, fix not in this PR)

- `backend/src/meho_backplane/topology/annotate.py` is now 1289 lines and `topology/refresh.py` is 1013 lines. The code-quality hook flags both as pre-existing file-size warnings (block limit 600). The task body mandates editing both files; an in-PR split is out of scope. The four critical-path functions (`_reconcile_nodes`, `_reconcile_edges`, `annotate_edge_in_txn`, `unannotate_edge`) were extracted into per-branch / per-phase helpers as part of this PR to clear the function-size block-limit -- a follow-up refactor task to split the topology service by responsibility (`refresh` / `annotate` / `history` / `bulk` modules) is a candidate for the orchestrator to file.
- `tests/test_migration_0011_backfill_when_to_use.py::test_re_running_migration_is_idempotent` was failing on this branch because the test stamps the DB back to `0010` then runs `upgrade("head")`, which re-executes 0012's non-idempotent `CREATE TABLE graph_node_history`. Fixed in-PR by targeting `upgrade("0011")` so 0012 is not re-executed; the test's intent (re-run 0011 against an already-curated DB) is preserved.

Closes #857

---

## Follow-up (auto-implement-issue, iteration 2, 2026-05-22)

Addressed every iter-1 `/auto-review-pr` REQUEST_CHANGES finding in commit `0f3ba8b`:

- **B1** (blocker) `0f3ba8b` -- accept the migration's intended `ON DELETE SET NULL` cascade on `graph_edge_history.edge_id`: the curated row's hard-delete in `unannotate_edge` nulls the tombstone's `edge_id` on PG, which is correct (history rows must survive live-row deletion). Identity recovery now documented to flow through `snapshot.before.id` + the `graph_edge_history_tenant_removed_idx` partial index. Unit test asserts `removed[0].edge_id is None` + `snapshot.before.id == str(curated_id)`. Added an opt-in `MEHO_SQLITE_FOREIGN_KEYS=1` env-var gate in `create_engine_for_url` plus an autouse fixture in the topology history hook test module that flips it, so SQLite enforces the FK and the test catches the divergence going forward. The gate is module-scoped because a blanket SQLite-FK-on flip surfaces a large pre-existing tape of test fixtures that insert FK-referencing rows without seeding the parent (chassis-wide refactor, out of scope for #857).
- **B2** (blocker, CodeRabbit-flagged) `0f3ba8b` -- guarded the idempotent re-annotate path so it no longer emits phantom `UPDATED` history rows. `_mark_same_kind_different_endpoint_superseded` skips rows whose `superseded_by` already equals the curated id; `_mark_incompatible_kinds_conflict` skips rows with complete bidirectional `conflicts_with` linkage; the curated row's own history emission is gated on `_annotate_curated_is_meaningful` (excludes the `last_seen` / `properties.annotated_at` heartbeat fields, mirroring refresh's `_properties_differ` contract). Added `test_idempotent_annotate_emits_no_history_rows` regression covering both shapes.
- **M1** (major, CodeRabbit-flagged) `0f3ba8b` -- unannotate audit / broadcast payload partitions the cleared edges into two distinct buckets: `superseded` (where `snapshot.before.properties.superseded_by == removed_id`) and `conflicts` (where the removed_id appears in `snapshot.before.properties.conflicts_with`). Read from the captured snapshot so the partition is immune to the upcoming `_clear_reciprocal_markers` mutation.
- **m1** (minor, CodeRabbit-flagged) `0f3ba8b` -- `node_snapshot` / `edge_snapshot` now use `copy.deepcopy` instead of shallow `dict()` on `properties`, forward-compat against any nested-container mutation (the §6 `conflicts_with` is a list).
- **m2** (minor, CodeRabbit-flagged) `0f3ba8b` -- `test_refresh_failure_rolls_back_both_live_and_history_rows` materializes `.scalars().all()` inside the `AsyncSession` context, asserts on the bound lists outside.

Also repairs a T1 (#856) gap in `backend/tests/acceptance/conftest.py`: the per-test TRUNCATE list was missing `graph_edge_history` / `graph_node_history`, so every PG-backed acceptance test was erroring at setup with `cannot truncate a table referenced in a foreign key constraint`. The integration conftest's TRUNCATE list already includes both.

Disputed: none.

Deferred: none.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive diff-on-write history tracking for topology operations (node/edge creation, updates, deletions).
  * History rows now capture before/after snapshots linked to audit trails for refresh, annotation, and unannotation operations.

* **Tests**
  * Added extensive test suite validating history emission across all topology operations and failure scenarios.

* **Documentation**
  * Added detailed guide describing history hook behavior and snapshot structure for topology changes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/904?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/meho_backplane/db/engine.py
+++ b/backend/src/meho_backplane/db/engine.py
@@ -47,8 +47,11 @@ References
 
 from __future__ import annotations
 
+import os
 from collections.abc import AsyncIterator
+from typing import Any
 
+from sqlalchemy import event
 from sqlalchemy.ext.asyncio import (
     AsyncEngine,
     AsyncSession,
@@ -95,7 +98,47 @@ def create_engine_for_url(url: str, *, pool_size: int, pool_timeout: float) -> A
     if not url.startswith("sqlite"):
         pool_kwargs["pool_size"] = pool_size
         pool_kwargs["pool_timeout"] = pool_timeout
-    return create_async_engine(url, future=True, **pool_kwargs)
+    engine = create_async_engine(url, future=True, **pool_kwargs)
+
+    if url.startswith("sqlite") and os.environ.get("MEHO_SQLITE_FOREIGN_KEYS") == "1":
+        # SQLite enforces foreign-key constraints only when ``PRAGMA
+        # foreign_keys = ON`` is issued on every connection (the default
+        # is OFF, even on aiosqlite). Production runs on PG where FKs
+        # are always enforced; the dev/test driver is SQLite, where the
+        # default-off behaviour means a test suite cannot catch PG-vs-
+        # SQLite divergences like the G9.3-T2 ``graph_edge_history.
+        # edge_id`` ``ON DELETE SET NULL`` cascade, which silently
+        # nulled the column on PG and left it untouched on SQLite.
+        #
+        # The PRAGMA is gated on an **opt-in env var** rather than
+        # being on by default for SQLite. A blanket enable surfaces a
+        # large pre-existing tape of test fixtures that insert
+        # FK-referencing rows without seeding the parent (the
+        # SQLite-doesn't-enforce-FKs default has masked them since the
+        # chassis adopted the per-test SQLite template at #793).
+        # Tearing that out is a chassis-wide refactor; the diff-on-
+        # write hook does not need it. Tests that *do* exercise the
+        # cascade semantics (currently :mod:`tests.test_topology_history_hook`)
+        # opt in via the ``_enforce_sqlite_foreign_keys`` autouse
+        # fixture, which sets the env var, resets the engine cache,
+        # and tears down on yield.
+        #
+        # Registering on the underlying sync engine fires once per new
+        # connection in the async pool. ``sync_engine`` is the
+        # documented async-side event-attachment surface (SQLAlchemy
+        # 2.x async docs § "Synopsis - Core").
+        @event.listens_for(engine.sync_engine, "connect")
+        def _sqlite_enable_foreign_keys(
+            dbapi_connection: Any,
+            _connection_record: Any,
+        ) -> None:
+            cursor = dbapi_connection.cursor()
+            try:
+                cursor.execute("PRAGMA foreign_keys=ON")
+            finally:
+                cursor.close()
+
+    return engine
 
 
 def get_engine() -> AsyncEngine:
--- a/backend/src/meho_backplane/topology/annotate.py
+++ b/backend/src/meho_backplane/topology/annotate.py
@@ -82,7 +82,17 @@ import structlog
 from sqlalchemy import select
 
 from meho_backplane.broadcast import BroadcastEvent, publish_event
-from meho_backplane.db.models import AuditLog, GraphEdge, GraphEdgeKind, GraphNode
+from meho_backplane.db.models import (
+    AuditLog,
+    GraphEdge,
+    GraphEdgeKind,
+    GraphHistoryChangeKind,
+    GraphNode,
+)
+from meho_backplane.topology.history import (
+    edge_snapshot,
+    record_edge_change,
+)
 from meho_backplane.topology.resolvers import resolve_node
 
 if TYPE_CHECKING:
@@ -278,7 +288,7 @@ async def _mark_same_kind_different_endpoint_superseded(
     from_id: uuid.UUID,
     to_id: uuid.UUID,
     kind: str,
-) -> list[uuid.UUID]:
+) -> list[tuple[GraphEdge, dict[str, Any]]]:
     """Stamp ``superseded_by`` on auto edges that the curated row replaces.
 
     Conflict §6 rule 1: a curated edge ``runs-on(vm-A → host-Y)``
@@ -289,9 +299,13 @@ async def _mark_same_kind_different_endpoint_superseded(
     the change when the attribute is reassigned, so we reassign
     rather than mutating the inner dict).
 
-    Returns the list of marked auto-edge ids — caller uses it for the
-    audit/broadcast payload's ``superseded_count`` so the visibility
-    of the conflict is not buried in a side effect.
+    Returns a list of ``(edge, pre_mutation_snapshot)`` pairs — the
+    caller emits one ``GraphEdgeHistory`` row per marked auto edge
+    (G9.3-T2 #857 diff-on-write hook), using the pre-mutation
+    snapshot as the history row's ``snapshot.before``. The audit /
+    broadcast payload's ``superseded`` array is built from
+    ``edge.id`` of each returned pair so the visibility of the
+    conflict is not buried in a side effect.
     """
     stmt = select(GraphEdge).where(
         GraphEdge.tenant_id == tenant_id,
@@ -301,12 +315,16 @@ async def _mark_same_kind_different_endpoint_superseded(
         GraphEdge.to_node_id != to_id,
     )
     rows = (await session.execute(stmt)).scalars().all()
-    marked: list[uuid.UUID] = []
+    marked: list[tuple[GraphEdge, dict[str, Any]]] = []
     for row in rows:
+        # Capture the pre-mutation snapshot **before** reassigning
+        # ``properties`` so the history row's ``before`` reflects the
+        # row as the operator last saw it without the supersede mark.
+        before = edge_snapshot(row)
         props = dict(row.properties or {})
         props[_SUPERSEDED_BY] = str(curated_edge_id)
         row.properties = props
-        marked.append(row.id)
+        marked.append((row, before))
     return marked
 
 
@@ -318,7 +336,7 @@ async def _mark_incompatible_kinds_conflict(
     from_id: uuid.UUID,
     to_id: uuid.UUID,
     kind: str,
-) -> list[uuid.UUID]:
+) -> list[tuple[GraphEdge, dict[str, Any]]]:
     """Stamp bidirectional ``conflicts_with`` on edges of other kinds.
 
     Conflict §6 rule 2: a curated edge ``depends-on(svc → db)``
@@ -328,8 +346,12 @@ async def _mark_incompatible_kinds_conflict(
     marker is a list so a single edge with several conflicting kinds
     accumulates every id (dedupe preserved).
 
-    Returns the list of conflicting edge ids — payload counterpart of
-    the superseded-list above.
+    Returns ``(edge, pre_mutation_snapshot)`` pairs for every
+    conflicting edge — caller emits one ``GraphEdgeHistory`` row per
+    pair (G9.3-T2 #857). The curated row's own marker write is
+    captured separately by the calling :func:`annotate_edge_in_txn`
+    so the curated row's history snapshot reflects the post-conflict
+    state.
     """
     stmt = select(GraphEdge).where(
         GraphEdge.tenant_id == tenant_id,
@@ -338,17 +360,21 @@ async def _mark_incompatible_kinds_conflict(
         GraphEdge.kind != kind,
     )
     rows = (await session.execute(stmt)).scalars().all()
-    conflicting_ids: list[uuid.UUID] = []
+    conflicting: list[tuple[GraphEdge, dict[str, Any]]] = []
     for row in rows:
-        conflicting_ids.append(row.id)
+        # Capture the pre-mutation snapshot before appending the marker.
+        before = edge_snapshot(row)
         _append_conflict_marker(row, curated_edge.id)
+        conflicting.append((row, before))
 
-    if conflicting_ids:
+    if conflicting:
         # Bidirectional: the curated row points back at every
-        # conflicting edge.
-        for other_id in conflicting_ids:
-            _append_conflict_marker(curated_edge, other_id)
-    return conflicting_ids
+        # conflicting edge. The curated row's own history row is
+        # written by the caller after both conflict scans complete,
+        # so we do not capture an intermediate snapshot here.
+        for other_edge, _ in conflicting:
+            _append_conflict_marker(curated_edge, other_edge.id)
+    return conflicting
 
 
 def _append_conflict_marker(edge: GraphEdge, other_id: uuid.UUID) -> None:
@@ -694,6 +720,251 @@ class AnnotatePlan:
     was_created: bool
 
 
+async def _insert_curated_edge(
+    session: AsyncSession,
+    *,
+    operator: Operator,
+    from_node: GraphNode,
+    to_node: GraphNode,
+    canonical_kind: str,
+    properties: dict[str, Any],
+    now: datetime,
+) -> GraphEdge:
+    """Insert a fresh ``source='curated'`` :class:`GraphEdge` and flush.
+
+    Flushing inside the helper ensures the row has its server-side
+    state (``id`` is already Python-side) before the §6 conflict
+    scans run -- they may need to read the freshly-inserted row back.
+    """
+    edge = GraphEdge(
+        id=uuid.uuid4(),
+        tenant_id=operator.tenant_id,
+        from_node_id=from_node.id,
+        to_node_id=to_node.id,
+        kind=canonical_kind,
+        source="curated",
+        properties=properties,
+        discovered_by=operator.sub,
+        first_seen=now,
+        last_seen=now,
+    )
+    session.add(edge)
+    await session.flush()
+    return edge
+
+
+def _merge_onto_existing_edge(
+    *,
+    existing: GraphEdge,
+    operator: Operator,
+    properties: dict[str, Any],
+    now: datetime,
+) -> tuple[GraphEdge, dict[str, Any]]:
+    """Idempotent upsert path: merge annotation onto an existing edge.
+
+    Covers two cases the single insert/upsert branch must handle:
+
+    1. **Re-annotate of an already-curated row.** Merge new free-text
+       fields onto the existing properties; the ``source='curated'``
+       and any reciprocal ``conflicts_with`` / ``superseded_by``
+       markers are preserved so the re-annotate does not silently
+       drop the §6 back-references.
+
+    2. **Annotate over an existing ``source='auto'`` edge with the
+       same triple.** The operator's intent is to take ownership of
+       that edge going forward — set notes / evidence, run §6
+       conflict detection from it, eventually revoke via
+       :func:`unannotate_edge`. Leaving the row as ``source='auto'``
+       would (a) make the triple-form unannotate raise
+       :class:`AutoEdgeDeletionError` so the operator cannot revoke
+       their own annotation; (b) let the next refresh's
+       :func:`_merge_edge_properties` overwrite the free-text props;
+       (c) make :func:`_mark_same_kind_different_endpoint_superseded`
+       on a future annotate mis-target the row as another auto
+       edge. The promotion to ``source='curated'`` and
+       ``discovered_by=operator.sub`` makes the operator the
+       canonical author from this call onward.
+
+    Returns ``(edge, pre_mutation_snapshot)`` so the caller can use
+    the snapshot as the history row's ``snapshot.before``.
+    """
+    # Capture the pre-mutation snapshot **before** rewriting
+    # ``properties`` / ``source`` so the history row's ``before``
+    # reflects the state the operator would have seen on a
+    # ``list-edges`` immediately prior to the re-annotate.
+    edge_before = edge_snapshot(existing)
+    merged = dict(existing.properties or {})
+    for key, value in properties.items():
+        merged[key] = value
+    promoting = existing.source != "curated"
+    if promoting:
+        # §6 invariant: a curated edge must never carry an
+        # ``superseded_by`` marker — that marker is meaningful only
+        # on ``source='auto'`` rows that the conflict-detection scan
+        # stamped against a different-endpoint curated assertion.
+        # When the same triple gets re-annotated, the auto row's
+        # stale marker would otherwise persist onto the promoted
+        # curated row and hide it from :func:`find_dependents` /
+        # :func:`find_dependencies` — the traversal CTE filters
+        # ``properties->>'superseded_by' IS NULL``. Drop the key
+        # here so the promoted curated edge is immediately visible.
+        #
+        # ``conflicts_with`` (bidirectional, list-shaped) is kept:
+        # the entries point at *other-kind* edges over the same
+        # endpoint pair, which still exist and still conflict; the
+        # incompatible-kinds scan below re-stamps the curated side
+        # without dropping the legitimate back-references.
+        merged.pop(_SUPERSEDED_BY, None)
+    existing.properties = merged
+    existing.last_seen = now
+    if promoting:
+        existing.source = "curated"
+        existing.discovered_by = operator.sub
+    return existing, edge_before
+
+
+def _emit_annotate_history(
+    session: AsyncSession,
+    *,
+    operator: Operator,
+    edge: GraphEdge,
+    edge_before: dict[str, Any] | None,
+    was_created: bool,
+    superseded_pairs: list[tuple[GraphEdge, dict[str, Any]]],
+    conflict_pairs: list[tuple[GraphEdge, dict[str, Any]]],
+    audit_id: uuid.UUID,
+    now: datetime,
+) -> None:
+    """Emit one history row per mutated edge in the annotate transaction.
+
+    Diff-on-write hook (G9.3-T2 #857): one row per mutated edge, all
+    sharing this annotate's pre-allocated ``audit_id`` and
+    ``valid_from=now`` so the temporal-query verbs (T3 / T4 / T5)
+    can reconstruct the operation as a single point-in-time event.
+    The curated row's history is emitted with the post-conflict-scan
+    snapshot so the ``conflicts_with`` marker is visible in
+    ``after`` -- the bidirectional view ``meho topology diff``
+    reconstructs.
+    """
+    edge_change_kind = (
+        GraphHistoryChangeKind.CREATED if was_created else GraphHistoryChangeKind.UPDATED
+    )
+    record_edge_change(
+        session,
+        edge_id=edge.id,
+        tenant_id=operator.tenant_id,
+        change_kind=edge_change_kind,
+        before=edge_before,
+        after=edge_snapshot(edge),
+        audit_id=audit_id,
+        valid_from=now,
+    )
+    for other_edge, other_before in (*superseded_pairs, *conflict_pairs):
+        # Each marked auto / other-kind edge survives -- the marker
+        # write is the only column that changed -- so the history
+        # row is ``updated`` with the marker visible in ``after``.
+        record_edge_change(
+            session,
+            edge_id=other_edge.id,
+            tenant_id=operator.tenant_id,
+            change_kind=GraphHistoryChangeKind.UPDATED,
+            before=other_before,
+            after=edge_snapshot(other_edge),
+            audit_id=audit_id,
+            valid_from=now,
+        )
+
+
+async def _upsert_curated_edge(
+    session: AsyncSession,
+    *,
+    operator: Operator,
+    from_node: GraphNode,
+    to_node: GraphNode,
+    canonical_kind: str,
+    note: str | None,
+    evidence_url: str | None,
+    now: datetime,
+) -> tuple[GraphEdge, dict[str, Any] | None, bool]:
+    """Resolve / upsert the curated edge; return ``(edge, before, was_created)``.
+
+    Wraps the insert-or-merge branch so :func:`annotate_edge_in_txn`
+    stays linear. ``before`` is ``None`` for a fresh insert, the
+    pre-mutation snapshot for the merge branch.
+    """
+    properties = {
+        "note": note,
+        "evidence_url": evidence_url,
+        "annotated_by": operator.sub,
+        "annotated_at": now.isoformat(),
+    }
+    existing = await _find_existing_edge(
+        session,
+        tenant_id=operator.tenant_id,
+        from_id=from_node.id,
+        to_id=to_node.id,
+        kind=canonical_kind,
+    )
+    if existing is None:
+        edge = await _insert_curated_edge(
+            session,
+            operator=operator,
+            from_node=from_node,
+            to_node=to_node,
+            canonical_kind=canonical_kind,
+            properties=properties,
+            now=now,
+        )
+        # ``before`` is None for the curated row's history -- the row
+        # did not exist prior to this transaction. ``after`` is captured
+        # by :func:`_emit_annotate_history` after both conflict scans
+        # have stamped any ``conflicts_with`` markers on this row.
+        return edge, None, True
+    edge, edge_before = _merge_onto_existing_edge(
+        existing=existing,
+        operator=operator,
+        properties=properties,
+        now=now,
+    )
+    return edge, edge_before, False
+
+
+async def _run_conflict_scans(
+    session: AsyncSession,
+    *,
+    edge: GraphEdge,
+    operator: Operator,
+    from_node: GraphNode,
+    to_node: GraphNode,
+    canonical_kind: str,
+) -> tuple[
+    list[tuple[GraphEdge, dict[str, Any]]],
+    list[tuple[GraphEdge, dict[str, Any]]],
+]:
+    """Run the §6 same-kind-supersede and incompatible-kinds scans.
+
+    Returns ``(superseded_pairs, conflict_pairs)``. Each pair is
+    ``(mutated_edge, pre_mutation_snapshot)`` for the history hook.
+    """
+    superseded_pairs = await _mark_same_kind_different_endpoint_superseded(
+        session,
+        tenant_id=operator.tenant_id,
+        curated_edge_id=edge.id,
+        from_id=from_node.id,
+        to_id=to_node.id,
+        kind=canonical_kind,
+    )
+    conflict_pairs = await _mark_incompatible_kinds_conflict(
+        session,
+        tenant_id=operator.tenant_id,
+        curated_edge=edge,
+        from_id=from_node.id,
+        to_id=to_node.id,
+        kind=canonical_kind,
+    )
+    return superseded_pairs, conflict_pairs
+
+
 async def annotate_edge_in_txn(
     session: AsyncSession,
     operator: Operator,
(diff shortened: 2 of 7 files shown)


zdamir:
@coderabbitai review

zdamir:




**Anchor task:** #857 — G9.3-T2 diff-on-write hook (same-txn `*_history` rows + `audit_id` linkage in refresh + annotate)
**Surface area:** backend
**Best-practices packs audited:** python (SQLAlchemy 2.x async unit-of-work, transactional atomicity), devops (append-only history discipline)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `f6499c7`
**Stacked on:** #900 (G9.3-T1 #856 migration + ORM) — reviewed delta only via `git diff origin/feat/856-graph-history-migration..HEAD`
**AI reviewer (CodeRabbit):** received in 61s (7 actionable + 2 nitpicks)

### Acceptance criteria
- [x] AC#1: refresh adds 3 nodes + 2 edges → 5 history rows + 1 audit row, all sharing `audit_id` — verified by `test_refresh_3n2e_emits_5_history_rows_sharing_audit_id`.
- [x] AC#2: annotate adds 1 curated edge → 1 history row sharing the annotate's `audit_id` — verified by `test_annotate_one_edge_emits_one_edge_history_row_with_shared_audit_id`.
- [x] AC#3: second refresh dropping a node → 1 `removed` history row with `before=<full row>`, `after=None`, NEW `audit_id` — verified by `test_refresh_dropping_node_emits_removed_history_with_new_audit_id`.
- [x] AC#4: forced failure rolls back live + history rows together — verified by `test_refresh_failure_rolls_back_both_live_and_history_rows` (patches `_write_audit_and_broadcast` to raise).
- [x] AC#5: history writes emit no own `audit_log` rows — verified inline in AC#1 test (`len(audits) == 1`).
- [⚠️] `pytest -n auto backend/tests/test_topology_history_hook.py` passes against SQLite; **PG-side semantics for unannotate (hard delete + FK SET NULL) is unverified and contradicts the unit-test assertion** — see B1.

### Project-board alignment
|---|---|---|---|---|---|
| #857 | In Review | Closes | ✅ | full (modulo B1/B2) | close on merge after blockers fixed |

### Predicted vs. actual diff (#857 delta only)
- ✅ `backend/src/meho_backplane/topology/history.py` (new, shared writer + snapshot helpers)
- ✅ `backend/src/meho_backplane/topology/refresh.py` (diff-on-write wired into `_reconcile_nodes` + `_reconcile_edges`, audit_id pre-allocated at top)
- ⚠️ `backend/src/meho_backplane/topology/annotate.py` (hook wired into `annotate_edge_in_txn` + `_unannotate_in_txn`; see B1, B2, M1)
- ✅ `backend/tests/test_topology_history_hook.py` (new, 8 ACs covered)
- ❓ `backend/tests/integration/conftest.py` (TRUNCATE list extended — justified: PG can't truncate `graph_node` while `graph_node_history` references it without listing both)
- ❓ `backend/tests/test_migration_0011_backfill_when_to_use.py` (idempotency test fix — justified: stamping back to 0010 + `upgrade("head")` re-runs 0012's non-idempotent CREATE TABLE; targeting 0011 preserves test intent while avoiding 0012 collision)

### Integration verification
- [x] `ruff check src/meho_backplane/topology/ tests/test_topology_history_hook.py` — clean (ran in worktree)
- [x] `mypy src/meho_backplane/topology/` — clean, 10 source files (ran in worktree)
- [x] `pytest tests/test_topology_history_hook.py` — 8 passed (ran in worktree)
- [x] `pytest tests/ --ignore=tests/integration --ignore=tests/acceptance --ignore=tests/test_api_v1_feed.py` — 3192 passed, 11 skipped (ran in worktree; 1 unrelated testcontainers env error in `test_api_v1_feed.py` skipped)
- [⚠️ not run] integration PG suite — Docker absent in agent sandbox; **B1 below is PG-specific and not exercised by current tests**
- [⚠️ not run] migrations replayed against fresh PG — same Docker constraint
- [n/a] frontend, agent eval, docker compose

For `/auto-implement-issue --continue-pr`:

- 🛑 **B1** — `annotate.py:1335` unannotate hard-deletes the curated edge in the same txn as its `removed` history insert. The `graph_edge_history.edge_id` column has a real `REFERENCES graph_edge(id) ON DELETE SET NULL` FK (migration 0012). On PostgreSQL the SET NULL action fires immediately when the live row is deleted in the same transaction; the freshly-inserted history row's `edge_id` is nulled on commit. The unit test `test_unannotate_emits_removed_history_and_clears_marker_history` asserts `removed[0].edge_id == curated_id` but only passes on SQLite (which does not enforce FKs by default — no `PRAGMA foreign_keys=ON` in `db/engine.py`). On PG production the assertion fails and downstream T3 `meho topology history --include-edges` queries that filter by `edge_id` (the `(tenant_id, edge_id, valid_from DESC)` index) cannot recover the curated edge's tombstone. Required change: either (a) document this PG behavior, change the unit-test assertion to `removed[0].edge_id is None` and adjust T3's tombstone-walk to use `snapshot.before.id` instead of the live `edge_id` column; OR (b) defer the FK action via `DEFERRABLE INITIALLY DEFERRED` and check the order at commit; OR (c) split the unannotate into two transactions (history insert + audit, then live delete) — but this breaks the same-txn atomicity contract of AC#4. The cleanest fix is (a): the migration's design intent is already that tombstones survive live-row deletion via SET NULL, so the test assertion is the wrong contract and T3 should be designed to query tombstones by `tenant_id + valid_from + change_kind='removed'` (matching `graph_edge_history_tenant_removed_idx`) and read identity from `snapshot.before`. Add an integration test against PG to lock the post-unannotate `edge_id IS NULL` behavior.
- 🛑 **B2** — `annotate.py:319-327` (`_mark_same_kind_different_endpoint_superseded`) and `annotate.py:363-376` (`_mark_incompatible_kinds_conflict`) unconditionally re-stamp `superseded_by` / `conflicts_with` markers and append every candidate to `marked` / `conflicting` — even when the existing marker value is already correct. `_emit_annotate_history` (`annotate.py:862-875`) then writes `UPDATED` history rows with byte-identical `before` and `after` snapshots. An idempotent re-annotate of the same triple thus inflates `graph_edge_history` with phantom no-op rows. The negative test `test_unchanged_refresh_emits_no_history_rows` proves the refresh path has the no-op contract; the annotate path silently violates it. Required change: in both scan helpers, compare the proposed marker value to the existing one before mutating + appending to the return list. For `_SUPERSEDED_BY` skip the row when `row.properties.get(_SUPERSEDED_BY) == str(curated_edge_id)`. For `_CONFLICTS_WITH` skip the row when `str(other_edge.id)` is already in the existing `conflicts_with` list AND the curated row's own `conflicts_with` already references the other id (the bidirectional check). Add a `test_idempotent_annotate_emits_no_history_rows` regression covering both shapes.
- ⚠️ **M1** — `annotate.py:1338-1348` (`_unannotate_in_txn`) builds the audit payload with `superseded=[]` hard-coded and `conflicts=[r.id for r in referencing]` flattening every cleared marker into the conflicts bucket. `referencing` is the union of edges whose `superseded_by` matched the removed edge AND edges whose `conflicts_with` listed it — they're distinct semantic categories per §6 of #364. The downstream audit-query verb (#219 G8) and broadcast watcher (`meho status --watch`) attribute the unannotate's cleared edges to the wrong category. Required change: extract two ID lists from `referencing_befores` before `_clear_reciprocal_markers` mutates them — `superseded_ids = [edge.id for edge, snap in referencing_befores if snap.get("properties", {}).get("superseded_by") == str(removed_id)]` and `conflict_ids = [edge.id for edge, snap in referencing_befores if str(removed_id) in (snap.get("properties", {}).get("conflicts_with") or [])]` — then pass them as `superseded=superseded_ids, conflicts=conflict_ids` to `_audit_payload`.
- 💡 **m1** — `topology/history.py:154-160, 175-179` (`node_snapshot` / `edge_snapshot`) use `dict(raw or {})` which is a shallow copy. Nested list / dict values inside `properties` (e.g. `conflicts_with: [...]`, future tag lists, free-text evidence objects) stay aliased to the live ORM row. The §6 conflict-detection passes that follow (`_append_conflict_marker` in `annotate.py:388-395`) reassign the whole `edge.properties` dict so the current code is safe, but the forward-compat hazard is real: any future code that does `edge.properties["conflicts_with"].append(...)` (in-place mutation) instead of the reassign-the-whole-dict idiom corrupts a previously-captured snapshot. Required change: `from copy import deepcopy` and use `deepcopy(raw or {})` for the `properties` column. Single-line fix, the test suite stays green.
- 💡 **m2** — `tests/test_topology_history_hook.py:454-474` (`test_refresh_failure_rolls_back_both_live_and_history_rows`) calls `.scalars()` on each SELECT inside `async with sessionmaker() as session:` but iterates the result objects (`list(live_nodes)` etc.) OUTSIDE the context manager. SQLAlchemy 2.0 async treats the result cursor's lifetime as tied to the session; the iteration works against `sqlite+aiosqlite` but is undefined across drivers. Required change: materialize with `.scalars().all()` inside the `async with` block, assign to lists, then assert outside. CodeRabbit flagged the same shape.

- `refresh.py:760-771` (`_reconcile_edges` soft-remove loop) does not skip `source="curated"` edges before calling `_soft_remove_edge`. This means a curated edge whose `from_node_id` is in this target's scope but isn't re-emitted by the connector (because no probe can emit curated edges — that's the whole point of annotation) gets soft-removed + a `removed` history row fires. **This is a pre-existing bug from G9.1/G9.2 not introduced by #857** — the diff shows the loop's structure is unchanged from the base; #857 only added the `_soft_remove_edge` history emission. CodeRabbit flagged it as if it were a #857 finding but the scope is wider — file a separate Task under G9.2 (#364) to add the `if row.source == "curated": continue` guard + a regression test. The downstream behavioral implication for #857 is that history rows for these spurious soft-removes will be emitted; the cleanup ticket should land before T3/T4/T5 (#859/#860/#861) so the temporal-query verbs aren't polluted with phantom curated-edge removals.
- `node_snapshot` / `edge_snapshot` round-trip uses `datetime.isoformat()` without an explicit `timezone.utc` normalization. Live rows always carry `tzinfo=UTC` (the chassis convention) so the round-trip is correct today; if a future write path persists a naive datetime the snapshot loses the timezone marker. Defensive guard worth considering when T6 retention prune lands (#858).
- The `_NODE_SNAPSHOT_COLUMNS` / `_EDGE_SNAPSHOT_COLUMNS` tuples omit `tenant_id` (documented rationale: the history row's own `tenant_id` carries it) but include `id`. For tombstone replay the omission is fine; for diff-rendering UIs that display the row's tenant alongside the change, the consumer reads from the history row's column. Worth a one-line cross-reference in the T4 `meho topology diff` verb docs when it lands.

- Adopted: 4 (B1 is original to me; B2 = CodeRabbit's "phantom UPDATED rows"; M1 = CodeRabbit's "superseded vs conflicts buckets"; m1 = CodeRabbit's "deepcopy for properties"; m2 = CodeRabbit's "scalars after session close")
- Out of scope: 4 (CodeRabbit findings on `models.py:1226` `_ck_in` nested f-string, `test_topology_history_migration.py:716-753` `valid_from` assertion, `docs/codebase/topology.md:602` markdown lint — these belong to #856's review concern; the curated-edge soft-remove finding at `refresh.py:760-771` is pre-existing from G9.1/G9.2 and recorded in Adjacent improvements above)
- Disputed: 0

### Docs
- [x] `docs/codebase/topology.md` — updated with the "Diff-on-write history hook" section (+129 lines covering snapshot shape, audit_id linkage, refresh path, annotate path, atomicity contract). Markdown lint MD018 at line 602 (CodeRabbit nit) is cosmetic — recommend fix in continue-mode pass but not a blocker.

- **Atomicity discipline is exemplary.** The same-txn contract is honored everywhere: `_apply_reconcile` opens one `session.begin()`, the audit row is added to the same session, every history row goes via `session.add(...)` not a separate session. The AC#4 test mocks `_write_audit_and_broadcast` to raise and asserts both live + history tables are empty post-rollback — exactly the right shape.
- **`audit_id` pre-allocation pattern reused correctly.** `refresh.py:975` allocates one `audit_id = uuid.uuid4()` at the top of `refresh_target_topology` and threads it through `_apply_reconcile` → `_reconcile_nodes` / `_reconcile_edges` → every `record_*_change` call. Same shape `annotate.py:1027` follows. The "one operation, one audit row, N history rows linked back" invariant is preserved as the issue body specifies.
- **Snapshot column tuples (`_NODE_SNAPSHOT_COLUMNS` / `_EDGE_SNAPSHOT_COLUMNS`) are deliberately narrow.** The docstring at `history.py:44-54` correctly identifies the forward-compat hazard of reflection-based projection (a future derived column leaking into history) and chooses the explicit-opt-in tuple. The omission of `tenant_id` (already on the history row's own column) is right.
- **Pre-mutation snapshot capture ordering is correct in the easy cases.** `_update_existing_node` (`refresh.py:333`), `_soft_remove_node` (`refresh.py:366`), `_soft_remove_edge` (`refresh.py:683`), `_mark_same_kind_different_endpoint_superseded` (`annotate.py:323`), `_mark_incompatible_kinds_conflict` (`annotate.py:366`), and `_merge_onto_existing_edge` (`annotate.py:795`) all capture `before = edge_snapshot(row)` / `node_snapshot(row)` BEFORE the in-place mutation. The pattern is consistent and well-commented.
- **The reserved-marker preservation in `_merge_edge_properties` (`refresh.py:173-206`) is the right answer to the #595 sticky-supersede contract.** The PR threads `audit_id` into this path without altering the merge semantics, which is the minimum-change discipline.

zdamir:
auto-review-pr: pr=904 iter=1 verdict=REQUEST_CHANGES blockers=2 majors=1 minors=2 bot=coderabbitai received(61s) next_action=continue-implement-issue

zdamir:
@coderabbitai review

zdamir:




**Anchor task:** #857 — G9.3-T2 diff-on-write hook (history rows + audit_id in refresh + annotate)
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `0f3ba8bef7654ae7a90d1244611c39b3c15a1772`
**AI reviewer (CodeRabbit):** not received within 15min (iter-2 re-trigger at 09:23:41Z; no new review by 09:38Z; iter-1 CodeRabbit findings B2/m1/m2 were already addressed in iter-2)

### Iter-1 punch-list re-verification

All five iter-1 findings addressed in commit `0f3ba8b`:

- **B1** (blocker — unannotate hard-delete + FK SET NULL nulls edge_id on PG) — FIXED. Worker chose to embrace the migration's intended `ON DELETE SET NULL` cascade (history rows must survive live-row deletion per `migration 0012` docstring). Identity recovery now routes through `snapshot.before.id` + the `graph_edge_history_tenant_removed_idx` partial index, which is consistent with T3's (#859) API: `meho topology history <node-name>` resolves by node name, NOT by `edge_id`. The `_EDGE_SNAPSHOT_COLUMNS` tuple already includes `id` for this purpose. Test assertion updated to `removed[0].edge_id is None` + `snapshot.before.id == str(curated_id)`. The new opt-in `MEHO_SQLITE_FOREIGN_KEYS=1` gate (with `@event.listens_for(engine.sync_engine, "connect")` PRAGMA listener — the documented SQLAlchemy 2.x async event-attachment surface) plus the module-scoped autouse fixture ensures SQLite catches the divergence going forward. The module-scoped opt-in (rather than chassis-wide flip) is correctly justified: a blanket SQLite-FK-on would surface a large pre-existing tape of FK-orphan fixtures.

- **B2** (blocker — idempotent re-annotate emits phantom UPDATED rows) — FIXED. All three call sites guarded:
  - `_mark_same_kind_different_endpoint_superseded`: `if (row.properties or {}).get(_SUPERSEDED_BY) == target_marker: continue` (line 319+).
  - `_mark_incompatible_kinds_conflict`: bidirectional check on both `row.properties.conflicts_with` AND `curated_edge.properties.conflicts_with` (line 372+) — the bidirectional guard correctly handles the half-stamped state.
  - `_emit_annotate_history`: gates on `_annotate_curated_is_meaningful` (strips `last_seen` top-level + `properties.annotated_at` heartbeat), mirroring `refresh._properties_differ`. `was_created=True` always emits.
  - New `test_idempotent_annotate_emits_no_history_rows` exercises both shapes (curated row + §6 markers) and passes.

- **M1** (major — audit payload conflation) — FIXED. `_unannotate_in_txn` partitions `referencing_befores` into `superseded_ids` (where `before_props.get(_SUPERSEDED_BY) == removed_marker`) and `conflicts_ids` (where `removed_marker in before_conflicts`). Critically, the partition reads from the captured `ref_before` snapshot **before** `_clear_reciprocal_markers(referencing, ...)` mutates the live rows — immune to mutation as required. The `_audit_payload` helper already accepts both keys; the unannotate side was the only path conflating them.

- **m1** (shallow → deepcopy) — FIXED. `from copy import deepcopy` added to `history.py`; both `node_snapshot` (line 153+) and `edge_snapshot` (line 180+) now `deepcopy(raw or {})` on properties.

- **m2** (scalars after context close) — FIXED. `test_refresh_failure_rolls_back_both_live_and_history_rows` materializes `.scalars().all()` inside the `AsyncSession` context manager; assertions on the bound lists outside.

### Bonus fix verification: `tests/acceptance/conftest.py` TRUNCATE list

Worker added `graph_edge_history` and `graph_node_history` to `_TRUNCATE_TABLES`. This is a **legitimate T1 (#856) gap**, not silent re-litigation: T1 created the history tables with FK to live tables (`ON DELETE SET NULL`), but only the `tests/integration/conftest.py` `_TRUNCATE_TABLES` was updated; the acceptance conftest was missed. Without these entries, PG rejects per-test TRUNCATE on the parent with `cannot truncate a table referenced in a foreign key constraint`. Repairing this upstream gap is in-scope for #857 since #856 is the parent T1 and the failure mode is surfaced by stacking T2 on top. The change is minimal (two entries + explanatory comment) and well-commented.

### PR body iter-2 follow-up section

The "Follow-up (auto-implement-issue, iteration 2)" section in the PR body accurately summarizes every fix. Each B1/B2/M1/m1/m2 disposition matches the code diff. All referenced symbols (`_annotate_curated_is_meaningful`, `test_idempotent_annotate_emits_no_history_rows`, `MEHO_SQLITE_FOREIGN_KEYS`) are present.

### Project-board alignment
|---|---|---|---|---|---|
| #857 | In Review | Closes | yes | full (all 5 ACs met, iter-1 punch list cleared) | close on merge |

### Predicted vs. actual diff (iter-1→iter-2 delta)
- E `backend/src/meho_backplane/db/engine.py` — new opt-in SQLite FK PRAGMA gate
- E `backend/src/meho_backplane/topology/annotate.py` — B1 docstring + B2 guards (×3) + M1 partition
- E `backend/src/meho_backplane/topology/history.py` — m1 deepcopy on both snapshot helpers
- E `backend/tests/test_topology_history_hook.py` — B1 assertion update + autouse FK fixture + new B2 regression test + m2 materialize
- E `backend/tests/acceptance/conftest.py` — bonus T1 gap fix

### Integration verification (against iter-2 head SHA inside review worktree)
- [x] `ruff check` — clean on touched files
- [x] `mypy` — clean on touched src files (annotate.py, history.py, db/engine.py)
- [x] `pytest backend/tests/test_topology_history_hook.py` — **9 passed** (8 original + new B2 regression)
- [x] `pytest backend/tests/ --ignore=tests/integration --ignore=tests/acceptance` — **3211 passed, 11 skipped** (no regressions)
- [⚠️ not run] migrations replayed against fresh DB — substrate-only fix on the test side; B1's PG-specific assertion is exercised transitively via the autouse SQLite FK fixture
- [⚠️ not run] `docker compose up` boots clean — sandbox lacks Docker; the change is test-only + opt-in FK gate
- [n/a] touched UI flow — backend-only PR
- [n/a] agent trace through changed reasoning path

### Findings

No blockers. No majors. No minors with required changes.

- The `_annotate_curated_is_meaningful` heartbeat list is currently `("annotated_at",)`. If a future curated edge property is added that legitimately stamps every call but isn't operator intent (e.g. a `last_validated_at` mirror), it must be added here in lock-step with `refresh._properties_differ`. Worth a one-line cross-reference comment in both helpers when T6 (#858) retention prune lands.
- The autouse `_enforce_sqlite_foreign_keys` fixture is correctly module-scoped, but a follow-up task under #364 (chassis-wide) could widen the gate once the FK-orphan fixture tape is repaired. Captured in iter-1's adjacent improvements; no action needed for #857.
- The pre-existing G9.1/G9.2 bug from iter-1 (refresh.py:760-771 `_reconcile_edges` does not skip `source='curated'` edges before `_soft_remove_edge`) is unchanged in iter-2 and remains for a separate task. Confirmed not regressed.

- iter-2 re-trigger posted at 09:23:41Z; CodeRabbit did NOT post a new review within the 15-minute window (poll exhausted at 09:38Z, 30 attempts of 30s each)
- iter-1 CodeRabbit findings (B2/m1/m2) were already verified addressed above
- bot_review.received=false for iter-2; orchestrator may choose to ignore or escalate

### Docs
- [x] `docs/codebase/topology.md` — unchanged in iter-2 (was updated in iter-1; iter-2 fixes are mechanical, no doc surface drift)

### What worked well
- The B1 disposition (embrace SET NULL rather than re-order/change FK) is the right architectural call: the migration's documented intent IS that history rows survive live-row deletion. The fix correctly threads through to T3's read-side design (resolves by node name + snapshot.before.id, not by FK edge_id).
- The opt-in `MEHO_SQLITE_FOREIGN_KEYS=1` gate plus PRAGMA event listener is a textbook SQLAlchemy 2.x async pattern — listener on `sync_engine` is the documented attachment surface — and the module-scoped opt-in is well-justified.
- The M1 fix correctly captures the partition from the snapshot BEFORE `_clear_reciprocal_markers` mutates the live rows; the comment explicitly explains the ordering.
- The new B2 regression test exercises both shapes (curated row no-op + §6 marker no-op) in a single test, mirroring `test_unchanged_refresh_emits_no_history_rows` on the annotate side.
- The bonus T1 conftest fix is in-scope and well-commented; stacking T2 on T1 correctly surfaced the gap, and the worker repaired it rather than working around it.

zdamir (commented):
## /review-pr-slim summary

**Anchor task:** #857 — G9.3-T2 diff-on-write hook (same-txn `*_history` rows + `audit_id` linkage in topology refresh + annotate)
**Surface area:** backend (topology service / SQLAlchemy 2.x async unit-of-work / append-only history substrate)
**Verdict:** APPROVE (posted as `COMMENT` — GitHub blocks formal approval on your own PR; a third-party `APPROVE` is still required to flip `reviewDecision`.)
**Head SHA reviewed:** `93a21b15289a679929e7778db52049c9b2fd4b39`

### Reviewer disclosure

This reviewer touched PR #904's commits twice this session before this review: first to resolve a rebase-conflict in `backend/tests/integration/conftest.py` + `backend/tests/test_migration_0011_backfill_when_to_use.py` (both files functionally equivalent between branches; I took main's version, which is a strict superset of #904's intent), then to amend both commits with `Signed-off-by: Damir Zubanovic <dzubanovic@evoila.com>` and rebase onto current main. The PR's substantive diff content is preserved (post-rebase tree matches what auto-review-pr iter-2 reviewed at `0f3ba8b`).

A prior `/auto-review-pr` iter-2 (2026-05-22 09:56Z) independently approved at SHA `0f3ba8b`. Since the iter-2 SHA and the current head `93a21b1` are content-equivalent (only commit metadata + a doc-comment region differ; the `--ours` resolution dropped wording also present elsewhere in #904), iter-2's verdict transfers. This review primarily attests to (a) the conflict resolution was non-substantive, (b) acceptance criteria still pass on the rebased tree, (c) CI is green on the rebased SHA.

### Acceptance criteria

All 5 from #857 satisfied by tests:

- [x] **AC#1** — refresh adding 3 nodes + 2 edges produces exactly 5 history rows in the same txn sharing the refresh's `audit_id` — `test_refresh_3n2e_emits_5_history_rows_sharing_audit_id` (`backend/tests/test_topology_history_hook.py:287`).
- [x] **AC#2** — annotate adding 1 curated edge produces 1 history row sharing the annotate's `audit_id` — `test_annotate_one_edge_emits_one_edge_history_row_with_shared_audit_id` (line 338).
- [x] **AC#3** — refresh dropping a previously-discovered node produces 1 `removed` history row (`snapshot.before` = full row, `after` = NULL) with a NEW `audit_id` — `test_refresh_dropping_node_emits_removed_history_with_new_audit_id` (line 400).
- [x] **AC#4** — forced failure after live-table write but before commit leaves NEITHER the live change nor the history row — `test_refresh_failure_rolls_back_both_live_and_history_rows` (line 460). Atomicity via shared `AsyncSession` in `record_node_change` / `record_edge_change` is documented in `topology/history.py:12-30` and exercised end-to-end.
- [x] **AC#5** — history writes emit no `audit_log` rows of their own — implicit in AC#1's `len(audits) == 1` assertion (one refresh = one audit row, 5 history rows reference it via soft-FK `audit_id`).

Plus four iter-1 / iter-2 regression tests: `test_unannotate_emits_removed_history_and_clears_marker_history`, `test_annotate_supersedes_auto_emits_two_history_rows`, `test_unchanged_refresh_emits_no_history_rows`, `test_idempotent_annotate_emits_no_history_rows` (the B2 fix), `test_history_rows_scoped_to_tenant` — 9 hook tests in total.

### Project-board alignment

|---|---|---|---|---|---|
| #857 (T2) | In Review | `Closes #857` | ✅ same-repo, default base | ✅ full | close on merge, comment |
| #365 (Initiative G9.3) | In Progress | parent of #857 | — | ⚠️ partial (T1 merged + T2 here; T3-T6 still open) | leave open |
| #220 (Goal Topology) | In Progress | grandparent | — | — | leave open |

No T3/T4/T5/T6 sibling tasks claimed-closed. `closingIssuesReferences` matches Task body scope exactly.

### Predicted vs. actual diff

- ✅ `backend/src/meho_backplane/topology/history.py` (NEW, 267 lines) — expected and correct. Shared writer for both refresh + annotate. `_serialise` handles UUID + datetime portable to PG/SQLite JSONB; `deepcopy` on `properties` is the m1 fix.
- ✅ `backend/src/meho_backplane/topology/refresh.py` (+350/-71) — expected and correct. Helper extraction (`_soft_remove_edge` etc.) cleared the function-size block-limit; history emission lands inside the existing `session.begin()`.
- ✅ `backend/src/meho_backplane/topology/annotate.py` (+662/-210) — expected and correct. Per-branch / per-phase helpers, `_clear_reciprocal_markers`, `_annotate_curated_is_meaningful` heartbeat guard (B2), partition-by-marker-shape in unannotate audit/broadcast (M1).
- ✅ `backend/src/meho_backplane/db/engine.py` (+44/-1) — `MEHO_SQLITE_FOREIGN_KEYS=1` opt-in gate. Module-scoped autouse fixture in the new test file flips it; rationale documented (chassis-wide blanket flip is out of scope, B1 fix).
- ✅ `backend/tests/test_topology_history_hook.py` (NEW, 870 lines) — expected and correct. 9 tests covering 5 ACs + 4 regression cases. Mock target_uuid → AsyncSession path is consistent with existing topology test fixtures.
- ✅ `backend/tests/acceptance/conftest.py` (+10) — T1 (#856) gap fix: per-test TRUNCATE missing history tables. Same fix landed in main via PR #900's squash-commit `a1c0354`; #904's branch was based on pre-fix `f7b2473`, so this delta is needed.
- ✅ `docs/codebase/topology.md` (+129) — diff-on-write hook semantics, snapshot shape, no-own-audit rationale. Updates the existing topology doc rather than creating a new one (correct — topology surface is the same).
- ✅ `backend/tests/integration/conftest.py` (resolved during my rebase to main's version, equivalent to #904's intent).
- ✅ `backend/tests/test_migration_0011_backfill_when_to_use.py` (resolved during my rebase to main's version, strict superset of #904's pin-line-444 fix).
- ❓ None unexpected.
- 🚫 None undisclosed.
- 🕳️ None missing.

### Integration verification

Run in `$WORKTREE = .claude/worktrees/pr-904-review/` against head SHA `93a21b1`:

- [x] `ruff check src/meho_backplane/topology/ src/meho_backplane/db/engine.py tests/test_topology_history_hook.py tests/acceptance/conftest.py` — All checks passed
- [x] `ruff format --check` — implicitly clean (part of CI's Python job which is SUCCESS on `93a21b1`)
- [x] `mypy src/meho_backplane/topology/ src/meho_backplane/db/engine.py` — `Success: no issues found in 11 source files`
- [x] `pytest tests/test_topology_history_hook.py tests/test_topology_history_migration.py tests/test_topology_refresh.py tests/test_topology_annotate.py -x -q` — **54 passed** in 8.11s
- [x] CI on this SHA — every Python job (ruff+mypy+pytest, integration testcontainers), backward-compat migration guard, build, SAST/SecretScan, CodeQL — all SUCCESS.
- [x] DCO on this SHA — SUCCESS (post-amend with `Signed-off-by: Damir Zubanovic <dzubanovic@evoila.com>`).
- [⚠️ reasoned-only] PG-side same-txn-rollback verification — covered by the testcontainers suite at CI; locally exercised on SQLite which the new `MEHO_SQLITE_FOREIGN_KEYS=1` gate makes FK-enforcing for this test module specifically (B1 fix mechanism).
- [⚠️ not run] `docker compose up` — no service code touched; not needed.
- [n/a] frontend / agent — no `meho_frontend/` or AI reasoning files touched.

### Findings (inline)

None added by this review. Two open items from prior CodeRabbit reviews; not duplicating them inline.

- **CodeRabbit iter-2 MD018 finding at `docs/codebase/topology.md:602` is still open** (flagged twice across iter-1 + iter-2). The line `#5 of #857 — re-emitting an audit row per history row would balloon` starts with `#5`, which markdownlint MD018 parses as an unclosed ATX heading. Fix: wrap `#5` in backticks or move it off line-start. CodeRabbit's suggested diff is the cleanest resolution. I'd address before merge for docs cleanliness, but CI doesn't currently enforce markdownlint so it's not strictly merge-blocking.
- **CodeRabbit iter-1 refresh.py:760-771 suggestion** — guard `source == "curated"` in the soft-remove loop. **Pre-existing on main** (line-history confirms #904 only changed `row.last_seen = None` → `_soft_remove_edge(...)` helper call; the loop-without-curated-filter existed before #857). Genuine concern (auto-discovery refresh would soft-remove operator-curated edges that aren't in the connector's probe results), but adjacent to #857's hook-emission scope. Candidate for a separate follow-up task.
- **CodeRabbit iter-1 nitpicks** that the author deferred:
  - Tenant-scoped cleanup deletes in `test_topology_history_hook.py:323/536/608` — test cleanup currently uses unscoped `delete(GraphEdgeHistory)`. Minor test-isolation hygiene; existing test runs are passing.
  - `_ck_in` nested f-string in `db/models.py:1226-1228` — CodeRabbit's "syntax error on Python 3.11" concern is moot (`requires-python = ">=3.12"` per `pyproject.toml`).
  - `docs/codebase/topology.md:584-590` doc-wording clarity around `ON DELETE SET NULL` — improvement, not bug.

### Docs

- [x] `docs/codebase/topology.md` — updated with 129 lines covering: hook semantics (atomicity, snapshot shape, no-own-audit), `audit_id` threading, snapshot column selection rationale, bulk-import composition. Open MD018 lint issue (see Adjacent improvements). No `.NET` analogies introduced.

### What worked well (review-body only, never inline)

- **Same-transaction atomicity contract is exemplary.** The `topology/history.py:12-30` docstring states the invariant ("the live graph and the history table can never disagree about which mutations committed") and AC#4's `test_refresh_failure_rolls_back_both_live_and_history_rows` proves it. The `record_node_change` / `record_edge_change` API takes the caller's `AsyncSession` (not its own); that's the right shape for "atomic with the caller's unit of work."
- **Snapshot column selection is narrow by design** (`_NODE_SNAPSHOT_COLUMNS` / `_EDGE_SNAPSHOT_COLUMNS` tuples) rather than `vars()` / `sqlalchemy.inspect`. Two genuine wins documented in the docstring: (a) prevents autoflush at the wrong time on rolled-back rows, (b) forward-compat opt-in for future derived columns. This is exactly the right discipline for an append-only substrate.
- **Idempotent re-annotate guard (B2 fix) is well-anchored.** `_annotate_curated_is_meaningful` mirrors refresh's existing `_properties_differ` contract — the `last_seen` / `properties.annotated_at` heartbeat fields are explicitly excluded so an annotate that just refreshes the timestamp doesn't emit a phantom UPDATED history row. The regression test `test_idempotent_annotate_emits_no_history_rows` covers both the same-kind-different-endpoint and the conflict-marker shapes.
- **`deepcopy` on `properties` (m1 fix)** is the right call. The §6 `conflicts_with` is a list — a shallow `dict(properties or {})` would alias the live row's list and any in-place mutation (`.append`, `.remove`) would retroactively rewrite a captured `before` snapshot. The docstring on `node_snapshot` makes the forward-compat rationale explicit.
- **Audit-id threading is consistent with chassis precedent.** Pre-allocating one `audit_id` per request and threading it down to every history row preserves "one operation, one audit row" while still letting each history row link back to the causing operation. Same discipline `audit_log.parent_audit_id` (migration 0006) and `publisher.publish` (G6 broadcast) already established.

## Post-merge actions (do after merge)

### Auto-close (should happen automatically)
- #857 — `Closes #857` in PR body, same repo, default base. Verify after merge: `gh issue view 857 --json state,closedByPullRequestsReferences`.

### Completion comment to post on close
- #857: "Closed by #904 — diff-on-write hook lands `*_history` rows in the same transaction as the live `graph_node`/`graph_edge` mutation, with `audit_id` threaded from the refresh/annotate operation. T3/T4/T5 (#859/#860/#861) temporal-query verbs and T6 (#858) retention prune are tracked separately under Initiative #365."

### Project-board status bump
- #857 → `Done` after merge.

### Initiative #365 still open
- T1 (#856) merged, T2 (#857) lands here; T3-T6 are separate PRs (#902 is T6, currently open). Initiative stays `In Progress`.

kr3s0 (approved):

Issue #882, opened 2026-05-21
Task: G10.5-T3 Topology dependents/dependencies + path query overlays (subgraph + highlight) + 30s polling-refresh


damir-topic:
Parent initiative: #342
Parent goal: #336
Depends-on: #881 (the overlays + polling render onto the Cytoscape graph)
Related: #363 (G9.1 dependents/dependencies/path APIs)

## Target

- Add: `/ui/topology?from=<name>&depth=N` (dependents/dependencies subgraph) + `/ui/topology?from=A&to=B` (path) query handling + HTMX `hx-trigger="every 30s"` polling-refresh on the graph container.
- Add: `backend/tests/test_ui_topology_queries.py`.
- Precedent: #881 graph view; the G9.1 `dependents`/`dependencies`/`path` query verbs (`api/v1/topology.py` + `cli/internal/cmd/topology/dependencies.go`/`path.go`, [#363](https://github.com/evoila/meho/issues/363)).

## Context

Per #342 work-items #4, #5, #8 — the query overlays on the graph. Dependents/dependencies (`?from=<name>&depth=N`) wrap the G9.1 APIs; the server returns the subgraph JSON, Cytoscape re-renders. Path (`?from=A&to=B`) wraps `meho topology path`; the server returns the path, Cytoscape highlights it (`cy.$('#A, #B, ...edges').addClass('highlight')`). v0.2 ships polling-refresh (`hx-trigger="every 30s"`); live SSE update is v0.3.

## Current state

The graph (#881) renders the full tenant graph but has no subgraph/path query overlays and no auto-refresh.

## Desired state

`?from=<name>&depth=N` renders the dependents/dependencies subgraph; `?from=A&to=B` renders + highlights the path; the graph container polls every 30s for changes.

## Acceptance criteria

- [ ] `?from=<name>&depth=N` returns the correct dependents/dependencies subgraph (matches the G9.1 `dependents`/`dependencies` API) and Cytoscape re-renders it.
- [ ] `?from=A&to=B` returns + highlights the path (matches `meho topology path`); edge labels shown on the highlighted path.
- [ ] The "show dependents" link from the node drawer (#880) drives the `?from=` subgraph view.
- [ ] Polling-refresh (`hx-trigger="every 30s"`) re-pulls the current view's graph JSON without losing pan/zoom.
- [ ] Cross-tenant isolation holds for all query overlays.
[...]

Issue #881, opened 2026-05-21
Task: G10.5-T2 Topology Cytoscape.js graph view (vendored, layouts cose-bilkent/dagre/circle) + node-click drawer + tabular↔graph cross-link + 500-node cap


damir-topic:
Parent initiative: #342
Parent goal: #336
Depends-on: #880 (graph shares the node/edge data + cross-links with the table)
Related: #363 (G9.1 topology graph APIs)

## Target

- Add: `/ui/topology?view=graph` route + Jinja2 template emitting a Cytoscape.js island (`<div id="cy">` + init script) + vendored Cytoscape.js + cose-bilkent + dagre (SHA256-pinned like the chassis JS).
- Add: `backend/tests/test_ui_topology_graph.py`.
- Precedent: #880 table/drawer; chassis vendored-JS SHA-pin (#863); topology node/edge APIs (`api/v1/topology.py`, G9.1 #363).

## Context

Per #342 work-items #2, #7, #6 — the interactive graph. The Jinja2 template emits the node/edge JSON the backend already rendered into the page; the Cytoscape init script builds the graph (node shape per kind; edge labels by relationship type). Layout switcher (Alpine `<select>` → `cy.layout({name}).run()`): cose-bilkent (default), dagre, circle. Tabular↔graph cross-link preserves selection via query params. Frontend render capped at 500 nodes (documented); beyond that, a server-side subgraph query (T3) is required.

## Current state

The table + drawer (#880) exist but there is no graph visualization.

## Desired state

`/ui/topology?view=graph` renders the Cytoscape graph (pan/zoom/click) from server-embedded node/edge JSON; layout switcher works; clicking a node opens the same drawer as the table (`#880`); selecting a table row centers+selects that node in the graph and vice versa (selection preserved via query params); a 500-node cap is enforced + documented.

## Acceptance criteria

- [ ] `/ui/topology?view=graph` renders a Cytoscape graph (vendored Cytoscape + cose-bilkent + dagre, SHA256-pinned) for up to 500 nodes; pan + zoom + node-click work.
- [ ] Node click opens the detail drawer (reusing the #880 `/ui/topology/node/<id>` partial).
- [ ] Layout switcher toggles cose-bilkent / dagre / circle via `cy.layout(...).run()`.
- [ ] Tabular ↔ graph cross-link: selecting a table row centers+selects the node in the graph; a graph tap navigates back to the table with the row scrolled into view; selection preserved via query params.
[...]

PR #974, opened 2026-05-23
feat(ui): G10.5-T1 Topology tabular view + node detail drawer

### LINKED ISSUE #880: Task: G10.5-T1 Topology tabular view (sortable/filterable node table + multi-select) + node detail drawer
Parent initiative: #342
Parent goal: #336
Depends-on: #866 (G10.0 chassis — base layout + `/ui/topology` stub route)
Related: #363 (G9.1 topology graph + query APIs), #365 (G9.3 history — time-travel is OOS here)

## Target

- Add: `backend/src/meho_backplane/ui/routes/topology/` + Jinja2 templates (node table, node detail drawer).
- Add: `backend/tests/test_ui_topology_table.py`.
- Precedent: G10.0 chassis #866; topology query APIs [`api/v1/topology.py`](../blob/main/backend/src/meho_backplane/api/v1/topology.py) + `cli/internal/cmd/topology/` (G9.1 [#363](https://github.com/evoila/meho/issues/363)); `/api/v1/feed` for recent-ops on a node (G6.1 [#228](https://github.com/evoila/meho/issues/228)).

## Context

Per #342 work-items #1, #3 — the tabular surface + node detail. (Note: the #342 body's "G9 not yet filed" is **stale** — G9.1/G9.2 shipped in v0.3; the topology nodes/edges + dependents/dependencies/path APIs already exist in `api/v1/topology.py`.) Table: rows per node (name, type, product, parents, children count, last_seen); HTMX column-sort + filter dropdowns; multi-row checkbox select. Node detail drawer (`/ui/topology/node/<id>`): properties + incoming/outgoing edges + recent ops (audit_log filtered) + "show dependents" button (hands off to the graph in T2/T3). This Task ships table + drawer; graph (T2) + query overlays (T3) follow.

## Current state

`/ui/topology` is a chassis stub (#866); no table/detail.

## Desired state

`GET /ui/topology?view=table`: sortable + filterable node table with multi-select. `/ui/topology/node/<id>`: server-rendered drawer (properties, edge list, recent ops, show-dependents link).

## Acceptance criteria

- [ ] `/ui/topology?view=table` lists tenant nodes (name, type, product, parents badges, children count, last_seen); HTMX sort + filter re-render the table; multi-row checkbox select works.
- [ ] `/ui/topology/node/<id>` drawer shows properties + incoming/outgoing edges + recent ops on the node; "show dependents" link present (targets the graph view, T3).
- [ ] Cross-tenant isolation: another tenant's node never renders.
- [ ] `ruff` + `mypy` clean; `pytest -n auto backend/tests/test_ui_topology_table.py` passes.

## Out of scope

- Cytoscape graph view + cross-link — T2. dependents/path query overlays + polling — T3. Topology editing, time-travel, 3D — per #342 OOS.

## References

- Precedent: G10.0 chassis #866; [`api/v1/topology.py`](../blob/main/backend/src/meho_backplane/api/v1/topology.py) + `cli/internal/cmd/topology/` (G9.1 [#363](https://github.com/evoila/meho/issues/363)); `/api/v1/feed` (G6.1 [#228](https://github.com/evoila/meho/issues/228)).
- HTMX: https://htmx.org
- Best-practices anchors (by pack name): **frontend_best_practices** — HTMX server-side sort/filter + drawer; **python_best_practices** — FastAPI read routes + Jinja2, topology API reuse.

zdamir:
## Summary

- **Tabular surface** at `/ui/topology` -- server-rendered HTML table of the tenant's `graph_node` inventory. Sort + HTMX filter (name substring + kind dropdown) re-render only the `<tbody>` partial; per-row checkboxes drive client-side multi-select.
- **Per-node detail drawer** at `/ui/topology/node/{id}` -- properties + incoming/outgoing edges + recent `audit_log` ops on the node's target + a "Show dependents" link handing off to the future T2/T3 graph view.
- **Substrate primitive** `topology.query.list_nodes` -- tenant-scoped, ORM-based (SQLite + PG portable), closed sort-column enum as the SQL-injection guard. UI / CLI / MCP layers share it.
- **Cross-tenant isolation** enforced at the SQL layer: every UI handler reads `tenant_id` from the session-bound `UISessionContext`; no query parameter or path segment carries one. Cross-tenant node ids surface as 404.

## Test plan

- [x] `ruff check src/ tests/` -- clean.
- [x] `ruff format --check src/ tests/` -- clean.
- [x] `mypy src/` -- 0 issues across 256 source files.
- [x] `pytest -n auto backend/tests/test_ui_topology_table.py` -- 17 passed.
- [x] `pytest -n auto backend/tests/test_ui_chassis_smoke.py` -- 18 passed (the `/ui/topology` stub assertion is updated to skip the slug now that T1 ships the real surface; the four remaining stubs still pin).
- [x] Cross-tenant isolation -- `test_table_isolates_other_tenants_nodes` + `test_drawer_isolates_other_tenants_node_id` cover both surfaces.
- [x] HTMX fragment behaviour -- `test_table_htmx_request_returns_fragment_only` asserts the `<tbody>`-only response on `HX-Request: true`.
- [x] Sort + filter -- ascending default, `direction=desc` flips, unknown sort 422, kind + name-substring filters narrow results.
- [x] Drawer -- properties + edges + recent ops + show-dependents link + inner-node empty-ops message + 404 fragment.
- [x] Route ordering -- `test_real_topology_route_wins_over_stub` pins the include-router precedence so a future re-shuffle in `meho_backplane.ui.routes.build_router` can't silently re-route to the stub.
- [x] Adjacent topology suites (`test_topology_query_schemas`, `test_api_v1_topology`, `test_topology_create_node`, `test_topology_schema`) -- all green.

## Out of scope

- **Cytoscape.js graph view + table-graph cross-link + 500-node cap** -- T2 (#881).
- **Dependents / dependencies / path query overlays + 30s polling-refresh** -- T3 (#882).
- **Bulk-action endpoints** consuming the multi-row selection -- not in the work-item list; client-side selection state is exposed via Alpine.js so a future bulk-action button can read it without a template rewrite.
- **Parents column projection** -- the column header + empty cell reserves the space; the immediate-parent projection lands when T3's dependents overlay is wired or the substrate helper grows a `parents` projection.

Closes #880

---

## Follow-up (auto-implement-issue, iteration 1, 2026-05-23)

Addressed every finding from `/auto-review-pr` iteration-1
REQUEST_CHANGES verdict (see `.claude/auto/review-results/pr-974-iter-1.json`):

- **B1** `f1eada9` — Drop phantom `hx-*` attrs on Close button; inline `onclick` is sole handler.
- **B2** `f1eada9` — Wrap `node.name` + `node.kind` in `urllib.parse.quote(safe='')` for `dependents_href`; regression test on `"vm prod & staging"`.
- **B3** `f1eada9` — Regenerated `cli/api/openapi.json` + `cli/internal/api/client.gen.go` via `make snapshot-openapi && make generate`.
- **M1** `f1eada9` — Pipe `kind_filter` / `name_filter` through Jinja2 `| urlencode` on sort-column hrefs + `hx-get`.
- **M2** `f1eada9` — Kind dropdown sourced from `_GRAPH_NODE_KINDS` closed enum; regression test seeding 51 vm + 1 target row.

Disputed: none.

Deferred: none.

## Summary by CodeRabbit

* **New Features**
  * Topology table view: searchable, paginated, sortable columns, kind filter, HTMX fragment updates, and persistent selection
  * Node detail drawer: server-rendered node properties, incoming/outgoing edges, recent activity, and encoded “dependents” link
* **Security / Privacy**
  * Tenant-scoped access enforced for table and drawer; CSRF token/cookie enabled for HTMX flows
* **Tests**
  * New UI/acceptance tests covering table, drawer, sorting, filtering, encoding, and tenant isolation
* **Documentation**
  * Added Topology UI docs and usage/limits notes

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/974?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


zdamir:
@coderabbitai review

backend/src/meho_backplane/ui/templates/topology/_drawer.html
```diff
@@ -0,0 +1,173 @@
...
+          <span class="badge badge-ghost badge-sm">{{ node.kind }}</span>
+          <span class="font-mono">{{ node.name }}</span>
+        </h2>
+        <p class="text-xs opacity-70">Node detail</p>
+      </div>
+      <button
+        type="button"
+        class="btn btn-ghost btn-xs"
+        aria-label="Close node detail drawer"
+        hx-get="/ui/topology/node-drawer-empty"
```
zdamir: **[🛑 B1]** — Phantom HTMX GET to undefined route `/ui/topology/node-drawer-empty`

**What:** The Close button declares both `hx-get="/ui/topology/node-drawer-empty"` and an inline `onclick="...; return false;"` that swaps the drawer DOM client-side. There is no route handler registered for `/ui/topology/node-drawer-empty` anywhere in the codebase (`rg -n "node-drawer-empty" backend/` finds only this single template reference).
**Why it matters:** HTMX listens for the click via `document.body.addEventListener` (event delegation, set up at HTMX init) and the inline `onclick`'s `return false;` does **not** cancel a separately-registered listener. So every drawer-close click also fires a GET to `/ui/topology/node-drawer-empty`. Best case it hits a 404; with `UISessionMiddleware` in the chain, the unmatched `/ui/*` path is more likely to 302-redirect to `/ui/auth/login`. Either way the network panel shows a 4xx on every Close, and a future maintainer will see the failed request and chase a phantom bug.
**Required change:** Pick one path —
  - **Option A (smaller):** drop the `hx-get` / `hx-target` / `hx-swap` attributes from the Close button. The inline `onclick` already does the swap.
  - **Option B (cleaner):** remove the inline `onclick`, add a real `GET /ui/topology/node-drawer-empty` route in `routes/topology/detail.py` (or a new `node_drawer_empty.py`) that returns the "Select a row to inspect a node" empty fragment. Then the swap happens via HTMX semantics instead of inline JS.
Add a test that asserts the close path does not 4xx (either no network request, or 200 with the placeholder).

backend/src/meho_backplane/ui/routes/topology/detail.py
```diff
@@ -0,0 +1,356 @@
...
+            "outgoing_edges": outgoing,
+            "incoming_edges": incoming,
+            "recent_ops": recent_ops,
+            # The "show dependents" link target. T3 (#882) replaces
+            # the surface with a real handler; the URL contract
+            # ships today so a future hand-off does not need a URL
+            # rename. The dependents view rooted at this node lives
+            # at ``/ui/topology?view=graph&root=<name>`` per the
+            # Initiative #342 work-item #4 / #7 design.
+            "dependents_href": (f"/ui/topology?view=graph&root={node.name}&kind={node.kind}"),
```
zdamir: **[🛑 B2]** — `dependents_href` interpolates raw node name/kind into a URL query string without URL encoding

**What:** Line 345 builds the URL as `f"/ui/topology?view=graph&root={node.name}&kind={node.kind}"`. Per `backend/src/meho_backplane/db/models.py:1319`, `graph_node.name` is `Text NOT NULL` with no character-class CHECK constraint. Connectors populate it from arbitrary external sources (VM names, pod names, principal slugs, target slugs) which legitimately contain URL-reserved characters: `&`, `?`, `#`, `+`, `=`, `%`, and space.
**Why it matters:** A VM named `app&env=prod` produces `?view=graph&root=app&env=prod&kind=vm`, which `URLSearchParams` (or FastAPI's `Query`) parses as three separate params: `view=graph`, `root=app`, `env=prod`, `kind=vm`. The downstream T2/T3 graph view sees `root=app` instead of `root=app&env=prod` — the dependents link silently routes the operator to the wrong root node. The same issue applies for `#` (anchor truncation), `+` (decoded to space), and `%` (start of a percent-escape that may parse to garbage).
**Required change:**
```python
from urllib.parse import quote

# ...
"dependents_href": (
    f"/ui/topology?view=graph"
    f"&root={quote(node.name, safe='')}"
    f"&kind={quote(node.kind, safe='')}"
),
```
Add a test that seeds a node with a name containing `&` (or `+` / space) and asserts the rendered `href` contains the percent-encoded form (e.g. `root=app%26env%3Dprod`). The drawer test in `tests/test_ui_topology_table.py::test_drawer_renders_node_properties_and_edges` already asserts `"root=vm-on-host-1"` — extend with a second case using a reserved-char name.

backend/src/meho_backplane/ui/templates/topology/table.html
```diff
@@ -0,0 +1,180 @@
...
+            ("kind", "Kind"),
+            ("last_seen", "Last seen"),
+            ("first_seen", "First seen"),
+          ] -%}
+          {% for col_value, col_label in columns %}
+            {%- set next_dir = next_direction_for(col_value) -%}
+            <th>
+              <a
+                href="/ui/topology?sort={{ col_value }}&direction={{ next_dir }}{% if kind_filter %}&kind={{ kind_filter }}{% endif %}{% if name_filter %}&q={{ name_filter }}{% endif %}"
+                hx-get="/ui/topology?sort={{ col_value }}&direction={{ next_dir }}{% if kind_filter %}&kind={{ kind_filter }}{% endif %}{% if name_filter %}&q={{ name_filter }}{% endif %}"
```
zdamir: **[⚠️ M1]** — Sort/filter URL builders interpolate `kind_filter` / `name_filter` without URL encoding

**What:** Lines 138-139 build the column-header `href` and `hx-get` URLs by interpolating `{{ kind_filter }}` and `{{ name_filter }}` directly into the query string:
```
href="/ui/topology?sort={{ col_value }}&direction={{ next_dir }}{% if kind_filter %}&kind={{ kind_filter }}{% endif %}{% if name_filter %}&q={{ name_filter }}{% endif %}"
```
`name_filter` carries operator-controllable substring input (capped at `max_length=256`); `kind_filter` echoes the URL `?kind=` value back. Jinja2 autoescape converts `&` to `&amp;` for the HTML attribute, but the browser URL-decodes the attribute on click — so a value containing `&`, `+`, `#`, `?`, or `%` poisons the next request's query string (same class of bug as the drawer's `dependents_href`, different site).
**Why it matters:** An operator who pastes a `&`-containing search term, or who arrives via a URL with a doctored `?q=foo&drop_kind=...` query, sees the next sort click silently rewrite the query in unintended ways. For `name_filter` the practical blast radius is small (the substring filter just gets a different value); for `kind_filter` an attacker who can influence the `?kind=` value can force the link to carry an arbitrary additional query parameter.
**Required change:** Use Jinja2's built-in `urlencode` filter on every interpolated value inside the URL:
```jinja
href="/ui/topology?sort={{ col_value | urlencode }}&direction={{ next_dir | urlencode }}{% if kind_filter %}&kind={{ kind_filter | urlencode }}{% endif %}{% if name_filter %}&q={{ name_filter | urlencode }}{% endif %}"
```
Apply the same to the `hx-get` attribute on the same lines. (The `col_value` / `next_dir` interpolations are bound by closed enums, but applying `| urlencode` uniformly keeps the rule simple — "every URL segment passes through urlencode".)

backend/src/meho_backplane/ui/routes/topology/table.py
```diff
@@ -0,0 +1,284 @@
...
+
+def _node_kind_options(nodes: Iterable[TopologyNodeListEntry]) -> list[str]:
+    """Derive the kind-filter dropdown options from the current rows.
+
+    The dropdown shows only kinds the operator's tenant actually has
+    -- a v0.2-onboarding tenant with one connector should not see
+    eight irrelevant filter options. Sorted alphabetically for stable
+    rendering across page loads.
+    """
+    return sorted({node.kind for node in nodes})
```
zdamir: **[⚠️ M2]** — Kind-filter dropdown options are derived from the current page's rows, not the tenant's full kind set

**What:** `_node_kind_options(nodes)` returns the distinct kinds present in `nodes`, which is the **post-filter, post-limit** result of `list_nodes`. The default page size is 50 (`_LIMIT_DEFAULT`, table.py:91); the substrate caps the helper at `_LIMIT_MAX = 500`. So on any tenant with more than 50 (or more than 500 if the caller bumps `limit`) nodes that spans multiple kinds, the dropdown will silently miss kinds that exist in the tenant but aren't represented in the first page.
**Why it matters:** Two operator-visible bugs:
1. **Hidden kinds.** A tenant with 60 `vm` rows + 10 `target` rows alphabetically sorted will render the dropdown with only `vm` — the operator literally cannot filter to `target` from the dropdown (they have to know the URL or guess the value).
2. **Self-collapsing dropdown.** Once the operator filters by a kind (`?kind=vm`), `list_nodes` returns only `vm` rows, so the dropdown re-renders with `[All kinds, vm]` only. The operator now has no path back to other kinds via the dropdown — they have to manually clear the URL or page-refresh.
**Required change:** Source the dropdown from a tenant-stable kind set, not from the current page. Two reasonable options:
  - **Cheapest:** use the closed `_GRAPH_NODE_KINDS` tuple from `backend/src/meho_backplane/db/models.py:1136` (the same vocabulary the CHECK constraint enforces). Drop `_node_kind_options(nodes)` and pass `_GRAPH_NODE_KINDS` directly into the template context. The dropdown will always show every legal kind, regardless of which page is rendered or how the operator has filtered.
  - **Better signal-to-noise:** add a `list_node_kinds(session, tenant_id) -> list[str]` substrate helper that runs `SELECT DISTINCT kind FROM graph_node WHERE tenant_id = :tenant_id AND last_seen IS NOT NULL` and call it once per render. The dropdown then shows only kinds the operator's tenant actually has, but is independent of the current page / filter.
Add a test that seeds 51 `vm` nodes + 1 `target` node and asserts the dropdown contains both kinds.

backend/src/meho_backplane/ui/routes/__init__.py
```diff
@@ -33,20 +33,32 @@
...
+    sub-path; topology lands ``/ui/topology`` and
+    ``/ui/topology/node/{id}`` which would otherwise hit the
+    ``/ui/topology`` stub. Once G10.1-G10.4 land their surface
+    routers, each includes itself ahead of the stubs the same way.
     """
     router = APIRouter()
     router.include_router(build_dashboard_router())
+    # Surface routers ahead of stubs -- their concrete paths win
+    # the match against the stubs' placeholder ``/ui/{slug}``.
+    router.include_router(build_topology_router())
```
zdamir, line 62: **[🛑 B3]** — CLI snapshot + Go client are stale; `CLI API snapshot freshness` CI check is FAILED

**What:** The PR adds the `/ui/topology` and `/ui/topology/node/{node_id}` routes to the FastAPI app (registered at line 62 here), which widens the auto-generated OpenAPI document. The committed `cli/api/openapi.json` and the generated `cli/internal/api/client.gen.go` were not regenerated to reflect the new paths. The PR's `CLI API snapshot freshness` job (run 26329502687/job/77513007140) failed with:
```
cli/api/openapi.json and/or cli/internal/api/client.gen.go are out of date relative to the backend OpenAPI document. Run 'cd cli && make snapshot-openapi && make generate' and commit the result.
```
The job log shows the missing `UiTopologyNodeDetailUiTopologyNodeNodeIdGetResponse` types the regeneration would have added.
**Why it matters:** This check is a merge gate; it stays red until the snapshot + client are regenerated. The CLI / MCP layers that will eventually wrap `/ui/topology/node/{node_id}` rely on the generated types staying in sync with the backend OpenAPI doc — a stale snapshot is a silent contract drift waiting to surprise a follow-up PR.
**Required change:** In the same branch, run from a clean working tree:
```bash
cd cli && make snapshot-openapi && make generate
```
Commit the two regenerated files (`cli/api/openapi.json` and `cli/internal/api/client.gen.go`). Re-push; the freshness job should turn green.

zdamir (commented):
## /auto-review-pr summary (iteration 1) — VERDICT: REQUEST_CHANGES

**Note on review type:** GitHub blocks `REQUEST_CHANGES` on a self-authored PR (the gh token belongs to the PR author `zdamir`), so this review is posted as `COMMENT`. The binding verdict for the orchestrator is **REQUEST_CHANGES** — both in this body and in the machine-readable result file at `.claude/auto/review-results/pr-974-iter-1.json`. Three blockers + two majors below.




**Anchor task:** #880 — G10.5-T1 topology tabular view + node detail drawer
**Surface area:** backend + frontend
**Best-practices packs audited:** python (FastAPI/Jinja2/SQLAlchemy), frontend (HTMX/Alpine/Jinja escaping)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `0a9593c538622431ef05b3e85f95a73248ebfed6`
**AI reviewer (CodeRabbit):** not received within 15 minutes (poll window completed; the PR has been open 12+ hours since the `@coderabbitai review` trigger at 09:41 UTC and CodeRabbit has not posted)

### Acceptance criteria
- [x] `/ui/topology?view=table` lists tenant nodes with sort + HTMX filter + multi-select — verified by `test_table_full_page_renders_seeded_nodes`, `test_table_htmx_request_returns_fragment_only`, sort/filter tests.
- [x] `/ui/topology/node/{id}` drawer with properties + edges + recent ops + "show dependents" — verified by `test_drawer_renders_node_properties_and_edges` (but **see B2 below — the dependents-link URL is malformed for names containing URL-reserved chars**).
- [x] Cross-tenant isolation — verified by `test_table_isolates_other_tenants_nodes` + `test_drawer_isolates_other_tenants_node_id`.
- [x] `ruff`/`mypy` clean; `pytest tests/test_ui_topology_table.py` passes — 17 passed locally in the review worktree.

### Project-board alignment
|---|---|---|---|---|---|
| #880 | In Review | Closes | ✅ | ⚠️ (functional but with the B1/B2/B3/M1/M2 fixes) | hold close until blockers addressed |

### Predicted vs. actual diff
- ✅ Substrate `list_nodes` in `topology/query.py`; routes `topology/table.py` + `topology/detail.py` + `topology/__init__.py`; templates `topology/table.html` + `_table_rows.html` + `_drawer.html` + `_drawer_not_found.html`; tests `test_ui_topology_table.py`; chassis-smoke + `routes/__init__.py` updated to register topology router before stubs; `docs/codebase/ui.md` updated.
- 🚫 No undisclosed scope changes.
- 🕳️ **Missing**: regenerated `cli/api/openapi.json` and `cli/internal/api/client.gen.go` — adding the new route invalidated the committed snapshot. The `CLI API snapshot freshness` CI check is FAILURE on this PR (see **B3**).

### Integration verification
- [x] `ruff check` on changed paths — clean
- [x] `mypy` on changed paths — clean (4 source files)
- [x] `pytest tests/test_ui_topology_table.py -x -q` — 17 passed in 13s
- [x] `pytest tests/test_ui_chassis_smoke.py -x -q` — 18 passed (regression check after topology was inserted ahead of stubs)
- [⚠️ CI failure] `CLI API snapshot freshness` job — FAILED in PR CI: `cli/api/openapi.json and/or cli/internal/api/client.gen.go are out of date relative to the backend OpenAPI document`. See **B3**.
- [⚠️ not run] HTMX flow exercised in a browser (manual UI traversal) — reasoned-about only

- 🛑 **B1** — `_drawer.html` Close button fires a phantom HTMX GET to `/ui/topology/node-drawer-empty` (no such route); inline `onclick="...; return false;"` does NOT suppress HTMX. Required change: drop the `hx-*` attributes from Close, OR add a real route for the empty fragment. Pick one path.
- 🛑 **B2** — `detail.py:345` builds `dependents_href` without URL-encoding raw `node.name`/`node.kind`. A VM name containing `&` / `+` / `#` / `%` / space silently corrupts the next URL's query parsing. Required change: `urllib.parse.quote(..., safe='')` on both segments + a reserved-char test.
- 🛑 **B3** — `cli/api/openapi.json` and `cli/internal/api/client.gen.go` are stale because the new route widened the OpenAPI document; the `CLI API snapshot freshness` CI check is RED. Required change: `cd cli && make snapshot-openapi && make generate`, commit the diff.
- ⚠️ **M1** — `table.html:138-139` builds column-header URLs by interpolating `kind_filter` / `name_filter` directly into the query string. Same URL-encoding class as B2, different site. Required change: pipe every interpolated URL segment through `| urlencode`.
- ⚠️ **M2** — `table.py:218-226` derives the kind-filter dropdown from the current page's `nodes`, so kinds beyond the first 50 rows silently vanish and the dropdown collapses to one kind once a filter is applied. Required change: source the dropdown from `_GRAPH_NODE_KINDS` (closed enum) or a separate `SELECT DISTINCT kind` substrate helper.

- `_drawer.html` renders edge endpoints with a `&rarr;` arrow between name and kind; the incoming-edges section reads `from name → kind` which is a slight visual mismatch (looks like the edge kind is the destination). A clearer shape would be `{from_name} ({from_kind}) →[edge.kind]→ this`. Pick up in a UI-polish PR.
- The `<select>` "kind" filter does not re-submit the form when the user changes via keyboard arrow-up/arrow-down + space — only `change` fires. Acceptable for v0.2; reconsider when a daisy-ui combobox lands.
- Consider documenting the 500-row hard cap on the table page in the surfaced UI (a small "showing 50 of N" footer would be useful once tenants grow). Adjacent to T2/T3.

- Adopted: 0 (no findings — bot did not post)
- Disputed: 0
- Out of scope: 0
- **Note:** CodeRabbit was triggered at PR open time (`@coderabbitai review` comment at 2026-05-23T09:41:24Z) and has not posted any review in the 12+ hours since. The 15-minute wait window in this skill elapsed without a review. The orchestrator may want to re-trigger CodeRabbit if a fresh review is desired on iteration 2.

### Docs
- [x] `docs/codebase/ui.md` — updated with a full "Topology surface (Task #880)" section covering routes / module layout / substrate primitive / templates / HTMX wiring / cross-tenant isolation / known limits. 92-line addition. Verified accurate against the code.

- Substrate-first factoring: `list_nodes` lands in `topology/query.py` with a closed sort-column enum, ORM-based dialect portability, unconditional `WHERE tenant_id = :tenant_id`, and per-call validation — exactly the right home for a primitive UI / CLI / MCP layers will share.
- Cross-tenant isolation is enforced at the SQL layer in every helper, not just the route — `_fetch_node`, `_fetch_edges`, `_fetch_recent_ops` all start with `tenant_id` match. The two cross-tenant tests are well-targeted.
- The `topology/__init__.py` umbrella + `routes/__init__.py` re-ordering (topology router included **before** stubs router) is the right way to preserve first-match-wins; `test_real_topology_route_wins_over_stub` pins the invariant explicitly so a future reshuffle can't silently re-route to the stub.
- Comprehensive docstrings on every new function explain rationale (closed-enum SQL injection guard, SQLite-vs-PG dialect portability, soft-delete exclusion default) — easier to review and easier to maintain.

zdamir:
@coderabbitai review

cli/api/openapi.json
```diff
@@ -3723,13 +3723,13 @@
...
-    "/ui/broadcast": {
+    "/ui/topology": {
       "get": {
         "tags": [
           "ui-stubs"
         ],
-        "summary": "Ui Stub Broadcast",
-        "operationId": "ui_stub_broadcast_ui_broadcast_get",
+        "summary": "Ui Stub Topology",
+        "operationId": "ui_stub_topology_ui_topology_get",
```
damir-topic: **[🛑 B1]** — `GET /ui/topology` is documented as the chassis **stub**, not the real table route.

**What:** `_SURFACE_STUBS` in `backend/src/meho_backplane/ui/routes/stubs.py` still registers a `/ui/topology` stub. Runtime routing is first-match-wins so the real table route (registered first) serves traffic — correctly pinned by `test_real_topology_route_wins_over_stub`. But FastAPI's `get_openapi()` keys paths by `(path, method)` and the **later-registered stub overwrites the real route** in the generated schema. Reproduced in-process against the real `ui.routes.build_router()` on FastAPI 0.136.1: the generated `GET /ui/topology` is `operationId=ui_stub_topology_ui_topology_get`, `tags=['ui-stubs']`, **0 parameters, responses `[200]` only** — exactly what this committed snapshot shows here.

**Why it matters:** The generated Go CLI client (`cli/internal/api/client.gen.go`) therefore exposes the topology table to the CLI with **no** `sort` / `direction` / `kind` / `q` / `limit` / `view` params and **no** 404. The Task #880 / dual-front-end contract ("UI / CLI / MCP share the same surface") is broken for the CLI front-end. `CLI API snapshot freshness` is green only because it checks `snapshot == generated`, not `generated == correct`.

**Required change:** Remove the `slug="topology"` entry from `_SURFACE_STUBS` in `stubs.py` (T1 ships the real surface; `_STUB_SURFACE_ROUTES` in the smoke test already excludes it), then `cd cli && make snapshot-openapi && make generate` and commit the corrected `openapi.json` + `client.gen.go`. After the fix the generated `/ui/topology` GET must carry the real query params. Add a regression assertion that the generated OpenAPI for `GET /ui/topology` is **not** tagged `ui-stubs` (schema-level analogue of `test_real_topology_route_wins_over_stub`).

**Source:** verified in-process against `meho_backplane.ui.routes.build_router()`; FastAPI 0.136.1 `fastapi.openapi.utils.get_openapi`. (Also flagged by CodeRabbit.)

backend/src/meho_backplane/ui/routes/topology/detail.py
```diff
@@ -0,0 +1,369 @@
...
+            from_alias.name.label("from_name"),
+            to_alias.id.label("to_id"),
+            to_alias.kind.label("to_kind"),
+            to_alias.name.label("to_name"),
+        )
+        .join(from_alias, from_alias.id == GraphEdge.from_node_id)
+        .join(to_alias, to_alias.id == GraphEdge.to_node_id)
+        .where(
+            GraphEdge.tenant_id == tenant_id,
+            GraphEdge.last_seen.is_not(None),
```
damir-topic, line 200: **[⚠️ M1]** — `_fetch_edges` endpoint joins are not tenant-scoped.

**What:** The endpoint joins constrain only on `id` (`from_alias.id == GraphEdge.from_node_id`, `to_alias.id == GraphEdge.to_node_id`) and this `WHERE` filters tenant only on `GraphEdge.tenant_id`. `GraphEdge.from_node_id` / `to_node_id` are FKs to `graph_node(id)` — the globally-unique PK — with **no** schema-level tenant constraint. So a same-tenant edge row that points at a cross-tenant node is not prevented by the DB; safety rests entirely on the refresh-service write-path invariant the docstring cites.

**Why it matters:** It is a defense-in-depth gap, not an active leak today, but it is the same belt-and-suspenders posture this PR already adopts for `_fetch_node` (which scopes by `tenant_id` rather than trusting the id). A future write path / manual fixup / partial migration that violates the invariant would silently surface another tenant's node `(kind, name)` in the drawer.

**Required change:** Add `from_alias.tenant_id == tenant_id` and `to_alias.tenant_id == tenant_id` to the `_fetch_edges` `WHERE` clause. Add a regression test that seeds a tenant-A edge whose endpoint node belongs to tenant B and asserts the drawer does not surface the foreign endpoint's `(kind, name)`.

(Also flagged by CodeRabbit.)

cli/api/openapi.json
```diff
@@ -3744,13 +3744,57 @@
...
+            "description": "Successful Response",
+            "content": {
+              "text/html": {
+                "schema": {
+                  "type": "string"
+                }
+              }
+            }
+          },
+          "422": {
```
damir-topic, line 3860: **[⚠️ M2]** — node-detail route omits `404` in its OpenAPI responses.

**What:** This operation declares `responses` `[200, 422]` only, but the handler returns **404** for a missing / cross-tenant node id (`detail.py:308-313`, rendering `_drawer_not_found.html` with `status_code=404`).

**Why it matters:** The generated CLI client has no knowledge of the 404 the route actually emits, so it can't model the not-found path — the contract under-describes the route's real behaviour.

**Required change:** Declare `responses={404: {...}}` on the detail route (`add_api_route(..., responses=...)` in `detail.py:build_detail_router`), then regenerate — this folds into B1's `cd cli && make snapshot-openapi && make generate` step.

(Also flagged by CodeRabbit.)

damir-topic (changes requested):




**Anchor task:** #880 — G10.5-T1 topology tabular view + node detail drawer
**Surface area:** backend + frontend
**Best-practices packs audited:** python (FastAPI / Jinja2 / SQLAlchemy), frontend (HTMX / Alpine / Jinja escaping)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `04b7ec1`
**AI reviewer (CodeRabbit):** received — review posted 2026-05-23T10:06Z against `f1eada9` (the post-fix head). The 2026-05-23T16:45Z merge with `main` introduced **zero** topology / UI / OpenAPI delta (it only pulled in connector + docs commits), so CodeRabbit did not re-review the refreshed head and its `f1eada9` review covers the identical PR content. Cross-checked in full, including all collapsed `<details>` blocks.

Iteration-1 fixes (B1 / B2 / B3 / M1 / M2) all verified landed — see "What worked well". This iteration's findings are **new**: they were missed in iter-1 because CodeRabbit was unavailable during that window, and CodeRabbit surfaced them on the post-fix head.

### Acceptance criteria
- [x] `/ui/topology?view=table` lists tenant nodes with sort + HTMX filter + multi-select — verified by `test_table_full_page_renders_seeded_nodes`, `test_table_htmx_request_returns_fragment_only`, sort/filter tests (20 passed locally).
- [x] `/ui/topology/node/{id}` drawer with properties + edges + recent ops + "show dependents" — verified; B2 URL-encoding fix confirmed with `quote(safe='')` + reserved-char test.
- [x] Cross-tenant isolation (table + drawer) — verified by `test_table_isolates_other_tenants_nodes` + `test_drawer_isolates_other_tenants_node_id`. **See M1** — the edge-endpoint projection has a defense-in-depth gap (not an active leak under write-path invariants).
- [❌] "UI / CLI / MCP layers share it" (substrate primitive contract, Task #880 + CLAUDE.md dual-front-end postulate) — **not satisfied for the CLI front-end**: the committed OpenAPI snapshot + generated Go client document the **stub**, not the real table route, for `GET /ui/topology` (no `sort`/`direction`/`kind`/`q`/`limit`/`view` params, no 404). See **B1**.

### Project-board alignment
|---|---|---|---|---|---|
| #880 | In Review | Closes | ✅ | ⚠️ (UI complete; CLI contract broken — B1) | hold close until B1 addressed |

### Predicted vs. actual diff
- ✅ Substrate `list_nodes` (`topology/query.py`); routes (`topology/table.py` + `detail.py` + `__init__.py`); templates; tests; `routes/__init__.py` topology-before-stubs ordering; `docs/codebase/ui.md`.
- ⚠️ `cli/api/openapi.json` + `cli/internal/api/client.gen.go` — regenerated (B3 from iter-1) but the regenerated artefact is **wrong** for `/ui/topology` because the still-registered stub shadows the real route in OpenAPI generation. See **B1**.
- 🚫 No undisclosed scope changes.
- 🕳️ Missing: removal of the now-superseded `/ui/topology` stub registration (the test was relaxed to skip it, but the route itself was left registered). See **B1**.

### Integration verification
- [x] `ruff check` (changed paths) — clean
- [x] `ruff format --check` (changed paths) — clean (4 files)
- [x] `mypy` (changed topology source) — clean (4 source files)
- [x] `pytest backend/tests/test_ui_topology_table.py` — **20 passed** in 1.84s (SQLite, no Docker)
- [x] CI (post-merge head `04b7ec1`): `Python (ruff + mypy + pytest)` pass, `Python (integration testcontainers)` pass, `Build backplane image` pass, `CLI API snapshot freshness` pass, CodeQL / Semgrep SAST+OSS / TruffleHog / Go / Helm / DCO pass. Deploy/smoke skipping (expected).
- [x] OpenAPI generation reproduced **in-process against the real app router** (FastAPI 0.136.1): confirms `/ui/topology` resolves to `ui_stub_topology_ui_topology_get` (stub) in the schema. See **B1**.
- [⚠️ not run] Browser HTMX traversal — reasoned-about only.

> Note on `CLI API snapshot freshness` being green: that check verifies `snapshot == generated-output`. It cannot catch that the generated output is itself wrong — the stub overwrites the real route during generation, and the committed snapshot faithfully captures that broken state. CI green is necessary but not sufficient here.

- 🛑 **B1** — `GET /ui/topology` is registered twice (the real table route in `topology/table.py` **and** the chassis stub in `ui/routes/stubs.py:_SURFACE_STUBS`). Runtime routing is first-match-wins so the real route serves traffic (correctly pinned by `test_real_topology_route_wins_over_stub`), but FastAPI's `get_openapi()` keys paths by `(path, method)` and the **later-registered stub overwrites the real route** in the schema. Confirmed in-process against the real `build_router()` on FastAPI 0.136.1: the generated `/ui/topology` GET is `operationId=ui_stub_topology_ui_topology_get`, `tags=['ui-stubs']`, **0 parameters, responses `[200]` only**. The committed `cli/api/openapi.json` and `cli/internal/api/client.gen.go` therefore expose the topology table to the CLI with **no** `sort` / `direction` / `kind` / `q` / `limit` / `view` params and **no** 404 — the CLI front-end of the delivered feature is wrong, violating the Task #880 / CLAUDE.md "UI / CLI / MCP share the same surface" contract. Required change: remove the `slug="topology"` entry from `_SURFACE_STUBS` in `backend/src/meho_backplane/ui/routes/stubs.py` (T1 now ships the real surface; `_STUB_SURFACE_ROUTES` in the smoke test already excludes it), then `cd cli && make snapshot-openapi && make generate` and commit the corrected snapshot + client. After the fix, the generated `/ui/topology` GET must carry the real query params (and the `/ui/topology/node/{node_id}` route should pick up its 404 — see M2). Add a regression assertion that the generated OpenAPI for `GET /ui/topology` is **not** tagged `ui-stubs` (mirrors the spirit of `test_real_topology_route_wins_over_stub`, but for the schema rather than the router).

- ⚠️ **M1** — `topology/detail.py` `_fetch_edges` (≈L193–198): the endpoint joins (`from_alias`, `to_alias`) are constrained only on `id`, and the `WHERE` filters tenant only on `GraphEdge.tenant_id`. `GraphEdge.from_node_id` / `to_node_id` are FKs to `graph_node(id)` (globally-unique PK) with **no** schema-level tenant constraint, so a same-tenant edge row pointing at a cross-tenant node is not prevented by the DB — the current safety rests entirely on the refresh-service write-path invariant the docstring cites. This is a defense-in-depth gap, not an active leak today, but it is the same belt-and-suspenders posture this PR already adopts for `_fetch_node` (which scopes by `tenant_id` rather than trusting the id). Required change: add `from_alias.tenant_id == tenant_id` and `to_alias.tenant_id == tenant_id` to the `_fetch_edges` `WHERE` clause, and add a regression test that seeds a tenant-A edge whose endpoint row belongs to tenant B and asserts the drawer does not surface the foreign endpoint's `(kind, name)`. (Also flagged by CodeRabbit.)

- ⚠️ **M2** — `cli/api/openapi.json` `/ui/topology/node/{node_id}` GET declares responses `[200, 422]` only; the handler returns **404** for a missing / cross-tenant node id (`detail.py:308-313`). Add a `responses[404]` declaration on the detail route (`add_api_route(..., responses={404: {...}})` or a `response` annotation) so the generated CLI client can parse the 404 the route actually emits. This regenerates as part of the B1 `make snapshot-openapi && make generate` step. (Also flagged by CodeRabbit.)

- `docs/codebase/ui.md` (≈L418-421): the prose mixes `request.headers["hx-request"]` (lowercase, correct for Starlette's normalized access) with "HTMX sets `HX-Request: true`" (the wire-form header name). Both are accurate but the casing shift reads as inconsistent. Optionally clarify that HTMX sends `HX-Request` on the wire and Starlette exposes it lowercased. (CodeRabbit nitpick — non-blocking; no code change.)
- Edge-direction rendering polish carried over from iter-1 (incoming-edges `name → kind` visual) — pick up in a UI-polish PR.

- Adopted: 3 — `/ui/topology` stub-shadow + missing real params (B1), node-detail missing 404 (M2), `_fetch_edges` cross-tenant endpoint filter (M1). All independently verified against the code; B1 reproduced in-process against the real app.
- Disputed: 0
- Out of scope: 0
- Nitpick (non-blocking, surfaced in Adjacent improvements): 1 — `ui.md` HX-Request header casing.

### Docs
- [x] `docs/codebase/ui.md` — updated with the topology surface section; accurate against the code. (One non-blocking wording nit in Adjacent improvements.)

- All five iteration-1 findings landed cleanly and correctly: B1 (Close button — `hx-*` dropped, inline `onclick` sole handler), B2 (`quote(safe='')` on `dependents_href` + reserved-char regression test), M1 (`| urlencode` on every sort-header href segment), M2 (kind dropdown sourced from the `_GRAPH_NODE_KINDS` closed enum + 51-row regression test). B3 ran the snapshot regen — the residual B1 here is a pre-existing route-registration issue the regen surfaced, not a regression in the fix.
- Substrate-first factoring and SQL-layer tenant scoping in every helper (`_fetch_node`, `_fetch_edges`, `_fetch_recent_ops`, `list_nodes`) remain exemplary; M1 is a tightening of an already-strong posture, not a hole.
- `test_real_topology_route_wins_over_stub` correctly pins the *runtime* first-match-wins invariant — the gap is that the analogous *schema*-generation invariant wasn't pinned, which is exactly what B1's regression assertion closes.

Issue #984, opened 2026-05-24
Task: Thread the dispatched Operator into Harbor robot_create/robot_delete (production-callable credential read; un-mask the stubbed test)


damir-topic:
Parent initiative: #944
Parent goal: #214
Related: #621 (G3.5-T9 — built the robot lifecycle ops, pre-credential-broker), #972 (PR that flagged this under "Adjacent findings"), #939 (credential-broker threading + helper)

## Scope

Make Harbor's typed write/mint ops `harbor.robot.create` and `harbor.robot.delete` production-callable under operator-context auth. Today they authenticate with a synthesised **system** operator (empty `raw_jwt`), which the post-#944 live Vault read rejects fail-closed — so the ops error against any real Harbor target, while CI stays green because the test injects a stub loader.

**Area:** [connectors/harbor/connector.py](backend/src/meho_backplane/connectors/harbor/connector.py), [operations/_branches.py](backend/src/meho_backplane/operations/_branches.py) (signature contract — no change expected), [tests/test_connectors_harbor_robot.py](backend/tests/test_connectors_harbor_robot.py).

## Current state

- `HarborConnector.robot_create` ([connector.py:389](backend/src/meho_backplane/connectors/harbor/connector.py#L389)) and `robot_delete` ([connector.py:461](backend/src/meho_backplane/connectors/harbor/connector.py#L461)) have signature `(self, target, params)` — **no `operator` parameter**.
- The typed dispatcher `_dispatch_typed` ([_branches.py:206-233](backend/src/meho_backplane/operations/_branches.py#L206)) threads the operator **by parameter name**: `"operator"` in the signature → invoke `(operator, target, params)`; otherwise `(target, params)`. Because the robot handlers omit `operator`, they never receive the dispatched one.
- Lacking an operator, the handlers call the auth surface with a synthesised system operator: `auth_headers(target, synthesise_system_operator())` ([connector.py:501](backend/src/meho_backplane/connectors/harbor/connector.py#L501)) and `_post_form(..., operator=synthesise_system_operator())` ([connector.py:453](backend/src/meho_backplane/connectors/harbor/connector.py#L453)).
[...]

PR #980, opened 2026-05-24
fix(connectors): G3.10 hygiene — defense-in-depth fail-closed on cache fast-path + architecture-doc carve-out

damir-topic:
## Summary
- Defense-in-depth fail-closed: every connector credential-cache fast-path now rejects an empty `operator.raw_jwt` before returning a cached entry (mirrors the loader-path guard at `vault_creds.py:188`). Covers `CredentialsCache.get`, vRLI `_session_token`, vcf-automation provider/tenant token caches, and the vmware-rest session-token cache.
- New architecture-doc section in `docs/architecture/connector-auth.md` ("Cache scoping under `shared_service_account`") formalises the per-target cache-key contract so future review doesn't re-flag it as a cross-operator leak — the cache mirrors the shared-service-account credential reality by design.
- `synthesise_system_operator()` now carries a non-empty placeholder JWT; live Vault still rejects it (preserving the architectural "system-initiated calls cannot read per-target vendor credentials" carve-out), but the cache fast-path no longer short-circuits every probe path before the wire call. Defense-in-depth still fires on actually-empty raw_jwt values (the four new regression tests prove it).
- Adjacent sibling-test default-operator updates: 5 connector `_make_operator()` factories changed default from `raw_jwt=""` to `raw_jwt="op.test.jwt"` (matching the existing vcf-shared convention); tests that exercise the empty-jwt rejection pass `raw_jwt=""` explicitly.

## Test plan
- [x] `ruff check` / `ruff format --check` — clean on all 16 touched files
- [x] `mypy` — clean (no issues found in 5 source files)
- [x] `pytest tests/test_connectors_vcf_operations_credread.py tests/test_connectors_vcf_logs_credread.py tests/test_connectors_vcf_automation_credread.py tests/test_connectors_vmware_rest_credread.py` — 13 passed (includes the 4 new fail-closed regression tests, one per cache helper)
- [x] `pytest tests/test_connectors_vcf_operations_auth.py tests/test_connectors_vcf_logs_auth.py tests/test_connectors_vcf_fleet_auth.py tests/test_connectors_vcf_shared_auth.py tests/test_connectors_vmware_rest_auth.py tests/test_connectors_vcf_automation_auth.py tests/test_connectors_vcf_fleet_credread.py` — 190 passed
- [x] Acceptance criteria mapped:
  - Each of the 4 cache helpers validates `operator.raw_jwt` BEFORE returning a cached entry; empty value raises `VaultCredentialsReadError` — verified by the 4 new regression tests
[...]

Issue #942, opened 2026-05-22
Task: G3.9-T3 Wire vmware-rest credential loader (live KV-v2 read) + recorded-fixture E2E + opt-in lab-vCenter smoke + onboarding doc (State 2)


damir-topic:
Parent initiative: #939
Parent goal: #214
Depends-on: #940 (operator threading), #941 (shared Vault-creds helper)

## Target

- Modify: [`connectors/vmware_rest/session.py`](backend/src/meho_backplane/connectors/vmware_rest/session.py) — replace the `load_session_credentials_from_vault` stub (L99) with a live read via the T2 helper.
- Modify: [`tests/test_connectors_vmware_rest_auth.py`](backend/tests/test_connectors_vmware_rest_auth.py) — replace `test_default_session_loader_raises_deliberate_stub` (L143) with a live-read test.
- Add: `backend/tests/integration/test_connectors_vmware_rest_credread.py` (recorded-fixture E2E of the full dispatch→session→read chain).
- Add: `docs/cross-repo/vmware-rest-onboarding.md`.
- Precedent: [`connectors/vault/ops.py:294`](backend/src/meho_backplane/connectors/vault/ops.py#L294), existing respx integration test [`tests/integration/test_connectors_vmware_rest_vcsim.py`](backend/tests/integration/test_connectors_vmware_rest_vcsim.py).

## Context

This is the load-bearing proof: with T1 (operator threaded to the loader) and T2 (shared KV-v2 helper) in place, the vmware default loader becomes live and `vmware-rest-9.0` executes against a real vCenter at rubric **State 2**. `target.secret_ref` is the Vault path holding `{"username","password"}` for the vSphere service account; the connector then establishes `POST /api/session` and reuses the `vmware-api-session-id` token ([connector.py:309](backend/src/meho_backplane/connectors/vmware_rest/connector.py#L309)). **vcsim cannot be the E2E** — it 404s on the modern `/api` surface ([vcsim test:4-16](backend/tests/integration/test_connectors_vmware_rest_vcsim.py)); the proof is a recorded-fixture replay in CI plus an opt-in live lab-vCenter smoke.

## Current state

`load_session_credentials_from_vault(target)` raises `NotImplementedError` (L99); `test_default_session_loader_raises_deliberate_stub` asserts it raises. vmware executes only with an injected stub loader (State 1).

## Desired state

[...]

PR #1006, opened 2026-05-24
fix(connectors): fail-fast secret_ref shape guard in _resolve_secret_ref (#989)

### LINKED ISSUE #989: Task: Fail-fast secret_ref shape guard in _resolve_secret_ref (reject KV-v2 API-prefixed values) + normalize offending test fixtures
Parent initiative: #944
Parent goal: #214
Depends-on: #984 (so this comprehensive guard + fixture sweep lands after the harbor robot-op fix and normalizes any `secret_ref` fixtures #984 adds; avoids a guard-vs-fixture CI conflict)
Related: #941 (built the shared `vault_creds` helper), #942 (added the operator-facing onboarding warning this guard enforces in code)

## Scope

Add a **fail-fast shape guard** for `secret_ref` in the shared credential-read path, so a target configured with a Vault KV-v2 **API-path** value (`secret/data/...`, `kv/data/...`, leading `data/...`) is rejected at read time with a clear, actionable error — instead of silently resolving to a doubled path (`secret/data/secret/data/...`) and returning a confusing 404. Then normalize the existing test fixtures that encode the wrong form.

**Why now:** there is **no programmatic guard** today — the only safeguard is the prose warning #942 added to `docs/cross-repo/vmware-rest-onboarding.md`. That the test suite itself contains ~100 fixtures using the wrong `kv/data/...` form (see below) is direct evidence that the mistake is easy and recurring, even for developers.

**Area:** [connectors/_shared/vault_creds.py](backend/src/meho_backplane/connectors/_shared/vault_creds.py), plus the connector test fixtures that currently use the API-prefixed form.

## Current state

- `secret_ref` is the **logical KV-v2 path relative to the mount root** — no `secret/` mount prefix, no `/data/` segment. The loader passes it straight to hvac `read_secret_version(path=secret_ref, mount_point=<mount>)` (default mount `"secret"`), and hvac inserts `/data/` itself. This is documented in the helper ([vault_creds.py:50-57](backend/src/meho_backplane/connectors/_shared/vault_creds.py#L50)) and the onboarding doc (#942).
- `_resolve_secret_ref` ([vault_creds.py:172-199](backend/src/meho_backplane/connectors/_shared/vault_creds.py#L172)) already runs the fail-closed precondition checks (empty `raw_jwt` → :189; unset `secret_ref` → :194-198) and returns `target.secret_ref.strip()` at :199 — but it does **not** validate the path *shape*. A mis-prefixed value passes straight through.
- Result: a target with `secret_ref="kv/data/vsphere/vcenter-a"` resolves (mount `secret`) to `secret/data/kv/data/vsphere/vcenter-a` → 404, with no hint that the prefix is the cause.
- The test suite encodes the wrong form widely: `grep -rn secret_ref backend/tests | grep -cE 'kv/data|secret/data|/data/'` → ~100 occurrences across `test_connectors_vmware_rest_auth.py`, `test_connectors_vcf_automation_auth.py`, `test_connectors_vcf_shared_auth.py`, `test_connectors_gcloud_*.py`, `test_connectors_hetzner_robot_auth.py`, and more. They pass only because they use pass-through fakes that return the secret regardless of path — so they also fail to prove the real logical-path → API-path resolution.

## Desired state

- `_resolve_secret_ref` rejects an API-path-shaped `secret_ref` with a `VaultCredentialsReadError` naming the target and stating the fix: "secret_ref must be the logical KV-v2 path relative to the mount (e.g. `targets/<id>`); drop the `secret/data/` / `kv/data/` / `data/` prefix — Vault adds `/data/` for KV-v2." Guard runs after the existing precondition checks, before the Vault read.
- The predicate is **specific** to avoid false positives: it trips on a leading mount-style `…/data/` API signature (`^(secret|kv)/data/`, `^data/`, or a second path segment equal to `data`), NOT on a logical segment legitimately named `data` deeper in the path (e.g. `targets/data-center-01/host` must remain valid).
- Test fixtures that exercise the guarded path are normalized to the logical form so they'd pass the guard; a new test asserts the guard **rejects** a representative API-prefixed value with the actionable message.

## Acceptance criteria

- [ ] `_resolve_secret_ref` raises `VaultCredentialsReadError` for `secret_ref` values `secret/data/foo`, `kv/data/foo`, and `data/foo`, with a message naming the target and the logical-path fix. Verified by a unit test.
- [ ] A logical path containing `data` as a non-signature segment (e.g. `targets/data-center-01/host`, `vsphere/vcenter-a`) is **accepted** (no false positive). Verified by a unit test.
- [ ] All test fixtures on the guarded `load_basic_credentials` path are normalized to the logical form (`grep -rn secret_ref backend/tests | grep -E 'kv/data|secret/data'` returns only fixtures that intentionally assert rejection). Connectors with bespoke loaders that do **not** go through `_resolve_secret_ref` (verify gcloud SA-key path, vcf-automation dual-plane) are either normalized too or explicitly noted as not-on-this-path.
- [ ] No credential/secret value appears in the new error message or any log event.
- [ ] `ruff check` + `ruff format --check` + `mypy backend/src/meho_backplane/` clean; full unit suite green under `pytest -n auto`.

## Out of scope

- **Registration-time / `targets.yaml`-load validation** (a Pydantic field validator on `Target.secret_ref` that fails even earlier). The read-path guard gives fail-fast-with-clear-error today; a registration-time variant can be filed separately if earlier feedback is wanted — not added speculatively here.
- Changing the logical-path convention itself, the default mount, or hvac usage.
- Auto-"fixing" a mis-prefixed value by stripping the prefix — reject with a clear error rather than silently guessing operator intent.

## References

- Guard home + existing precondition pattern: [vault_creds.py:172-199](backend/src/meho_backplane/connectors/_shared/vault_creds.py#L172); error type [vault_creds.py:124-165](backend/src/meho_backplane/connectors/_shared/vault_creds.py#L124).
- The convention this enforces: [vault_creds.py:50-57](backend/src/meho_backplane/connectors/_shared/vault_creds.py#L50) and the onboarding warning added in #942 (`docs/cross-repo/vmware-rest-onboarding.md`).
- hvac KV-v2 `read_secret_version` semantics (logical `path` relative to `mount_point`; `/data/` inserted by the client) — pin the installed hvac version in the implementation; introspect `help(hvac.api.secrets_engines.kv_v2.KvV2.read_secret_version)` if docs and behavior disagree.
- Best practices — **python_best_practices**: validate-at-the-edge / fail-fast with an actionable error message over silent-wrong-result; specific predicates over greedy matches to avoid false positives; a test must assert both the accept and reject paths (the ~100 wrong fixtures are the cautionary example of pass-through tests that prove nothing). Pydantic v2 + `pytest`/`capture_logs` are the repo conventions.

damir-topic:
## Summary
- Add a fail-fast shape guard in `_resolve_secret_ref` (`backend/src/meho_backplane/connectors/_shared/vault_creds.py`): an API-path-shaped `secret_ref` (`secret/data/…`, `kv/data/…`, leading `data/…`) is now rejected with a `VaultCredentialsReadError` naming the target and the logical-path fix, **before** the Vault read — instead of silently double-resolving to a 404. Until now the only safeguard was the prose onboarding warning (#942); the ~96 wrong test fixtures were direct evidence the mistake recurs.
- The predicate (`_is_api_path_shaped`) is **specific**: it trips only when the first path segment is `data` or the second is `data`, so a logical segment legitimately named `data` deeper in the path (`targets/data-center-01/host`) stays valid. No auto-stripping (no guessing operator intent); no credential value in the error.
- Composes correctly with #980's non-empty placeholder system JWT: the new guard runs independently of (after) the empty-JWT precondition.
- Normalized 97 test fixtures from the API-prefixed form to the logical form (incl. the harbor fixtures from the just-merged #984/#998, the `provider_secret_ref` override assertion, and the topology DB-seed value). gcloud's loader is a deliberate stub / injected pass-through that never reaches `_resolve_secret_ref` — its placeholder fixtures are normalized too for a correct convention example and to satisfy the acceptance grep. Vault ACL-policy paths (which correctly use the `secret/data/` API form) and the credread tests (already logical) are unchanged.

## Test plan
- [x] `ruff check` — clean
- [x] `mypy backend/src/meho_backplane/` — clean (269 files)
- [x] New unit tests assert the guard **rejects** `secret/data/foo` / `kv/data/foo` / `data/foo` (parametrized, incl. post-strip) and **accepts** `targets/data-center-01/host` / `vsphere/vcenter-a` / `targets/vc-lab-01` (no false positive) — `pytest tests/test_connectors_vault_creds.py` 25 passed
- [x] No credential value in the guard error (dedicated test) and no secret in logs (existing `capture_logs` test)
- [x] `grep -rn secret_ref backend/tests | grep -E 'kv/data|secret/data'` returns only the new guard test's intentional-rejection params
- [x] Normalized fixtures still green — connector auth (357 passed), e2e + topology + broadcast (329 passed, 5 skipped), credread + harbor robot (52 passed), all under `pytest -n auto`
- [x] `code-quality.py --diff` — 0 blockers (1 pre-existing warning: `load_basic_credentials` is 92 lines, mostly docstring; not agent-introduced)

## Out of scope
- Registration-time / `targets.yaml`-load validation (a Pydantic validator on `Target.secret_ref`) — the read-path guard gives fail-fast-with-clear-error today; a registration-time variant can be filed separately.
- Changing the logical-path convention, the default mount, or hvac usage.

Closes #989


damir-topic:
@coderabbitai review

damir-topic (commented):




**Anchor task:** #989 — fail-fast `secret_ref` shape guard in `_resolve_secret_ref` + normalize offending test fixtures
**Surface area:** backend (connectors / credential read path)
**Verdict:** APPROVE
**Head SHA reviewed:** `b56ed43791b5a7f55aad381e9681b32e120ccfec`
**AI reviewer (CodeRabbit):** not received within window — CodeRabbit returned a hard **rate-limit** notice ("Review limit reached … 1 review/hour, refill in ~26 min"); no review ran. Proceeded without it (sanctioned Phase 1.7 outcome). Static + integration analysis below stands on its own.

### Acceptance criteria
- [x] `_resolve_secret_ref` raises `VaultCredentialsReadError` for `secret/data/foo`, `kv/data/foo`, `data/foo` with a message naming the target + the logical-path fix — verified by `test_api_path_shaped_secret_ref_is_rejected` (parametrized, includes a whitespace-padded variant rejected post-strip). Ran locally: passing.
- [x] A logical path with `data` as a non-signature segment (`targets/data-center-01/host`, `vsphere/vcenter-a`) is accepted, no false positive — verified by `test_logical_secret_ref_with_data_substring_is_accepted` (asserts the read addressed the verbatim stripped path, proving it passed the guard rather than a pass-through fake hiding it). Also covers `data-store/primary` and `secret-data/foo` boundary cases.
- [x] All fixtures on the guarded `load_basic_credentials` path normalized to logical form. Bespoke loaders correctly handled: **gcloud** default loader is a deliberate `NotImplementedError` stub (not on the `_resolve_secret_ref` path) — its fixtures were normalized for consistency anyway, harmless; **vcf-automation** `provider_secret_ref` *does* flow to the guard via `load_credentials_with_override` → `load_basic_credentials`, and its fixture was correctly normalized (`vcfa/provider`).
- [x] No credential value in the error message or any log — guard fires before the Vault read; verified by `test_api_path_guard_message_carries_no_credential_value` + existing `capture_logs` no-leak test.
- [x] `ruff check` + `ruff format --check` + `mypy` clean (ran locally on changed files); full unit suite + integration testcontainers green in CI.

### Guard correctness
- Predicate `_is_api_path_shaped` = "first `/`-segment == `data`" OR "second `/`-segment == `data`". Correctly rejects `secret/data/…`, `kv/data/…`, `data/…`, and a leading-slash `/data/…` (split yields `['', 'data', …]` → second segment trips). Correctly accepts logical paths with a deeper/substring `data`.
- Placed **after** the two existing preconditions (empty `raw_jwt`, unset `secret_ref`) and after `.strip()`, **before** the Vault read. Raises `VaultCredentialsReadError` naming the target + the logical-path fix; does **not** auto-strip.
- Composes cleanly with #980: the guard is independent of the empty-JWT branch (separate, earlier check).

### ACL-policy paths correctly LEFT in API form (normalization not over-broad)
Confirmed the diff did **not** rewrite any real Vault API path:
- `test_connectors_k8s_live_vault.py:311` — `path "secret/data/{...}"` is an HCL **ACL policy** string; the matching `secret_ref` (line 226) is already the logical `k8s/k3d-live-vault`. Both correct, untouched.
- `test_connectors_vault.py` / `test_operations_*_register.py` — `secret/data` there are literal vault-ops API args / `when_to_use` prose, not `secret_ref` values. Untouched.
- `kv/dev/bind9/e2e` placeholder left as-is (only its comment updated); it short-circuits the Vault resolve and isn't API-path-shaped (second segment `dev`).

### Predicted vs. actual diff
- ✅ `vault_creds.py` — `_is_api_path_shaped` + guard wired into `_resolve_secret_ref`
- ✅ `test_connectors_vault_creds.py` — reject/accept/no-leak unit tests
- ✅ ~50 fixture files normalized `kv/data/<x>` → `<x>` (and one `vault:kv/data/x` → `targets/topology-x`)
- ✅ `docs/codebase/connectors-shared-vault-creds.md` — guard documented as step 3, Known-issues + tests sections updated, #989 reference added
- ❓ gcloud `_sa_key_loader` private-key fixture split to evade `detect-private-key` pre-commit hook — behavior-preserving (concatenation yields the identical literal); justified
- 🕳️ none missing
- 🚫 none undisclosed

### Integration verification
- [x] `ruff check` / `ruff format --check` — clean (local, changed files)
- [x] `mypy src/.../vault_creds.py` — clean (local)
- [x] `pytest tests/test_connectors_vault_creds.py` — 25 passed (local, against worktree code via `PYTHONPATH=src`)
- [x] CI `Python (ruff + mypy + pytest)` — SUCCESS (authoritative full unit suite)
- [x] CI `Python (integration testcontainers)` — SUCCESS
- [⚠️ skipped by CI] `pr-smoke` deploy — SKIPPED (expected for a test-only/guard change)

### Findings
None. No blockers, no majors, no minors requiring a code change.

- A whitespace-only `secret_ref` (`"   "`) passes the `not target.secret_ref` check, strips to `""`, and is not flagged by the shape guard (empty string trips neither predicate) — it would reach Vault with an empty path. This is pre-existing behavior (the `.strip()` predates this PR) and out of #989's scope; the issue explicitly defers registration-time validation. Worth a follow-up only if earlier feedback on blank refs is wanted.

- Adopted: 0 — CodeRabbit was rate-limited and produced no review this iteration.
- Disputed: 0
- Out of scope: 0

### Docs
- [x] `docs/codebase/connectors-shared-vault-creds.md` — updated (guard documented, step renumbering correct, no .NET analogies). `docs_status: updated`.

- Predicate is segment-based (not a greedy regex), which is exactly the "specific predicate over greedy match" the issue asked for — and the accept-tests prove the no-false-positive contract rather than just asserting a happy path.
- The accept-tests assert the read addressed the verbatim logical path, closing the "pass-through fake proves nothing" gap the issue flagged as the cautionary example.
- The normalization correctly distinguished `secret_ref` values from real Vault ACL/API paths — the k8s-live-vault ACL policy line is the load-bearing case and it was left untouched.

Issue #1091, opened 2026-05-25
Task: G6.4-T1 broadcast_recent MCP tool — agent-facing read of recent broadcast events from meho:feed:{tenant_id}


damir-topic:
Parent initiative: #1090

## Target

Register MCP tool `meho.broadcast.recent` (agent-facing read of recent broadcast events):

- NEW `backend/src/meho_backplane/mcp/tools/broadcast.py` — tool registration module + shared helper.
- NEW `backend/tests/test_mcp_tool_broadcast_recent.py` — handler tests.

## Context

Reads the operator's tenant's recent broadcast events from the Valkey stream `meho:feed:{tenant_id}` and returns them as a poll-style snapshot. Mirrors what `meho status --watch` consumes over SSE, but in the request/response shape MCP tools support.

**Input schema:**

```json
{
  "since": "ISO-8601 timestamp OR Valkey stream cursor (e.g. '1747800000000-0'); optional, default last 30m",
  "filter": {
    "op_class": "read|write|credential_read|audit_query (optional)",
    "principal": "JWT sub claim (optional)",
    "target": "target_name / connector instance (optional)"
  },
  "limit": "integer 1..1000, default 100"
}
```

**Output:**

```json
{
  "events": [
    {"id": "1747800000000-0", "tenant_id": "...", "op_class": "read", "op_id": "vsphere.vm.list", "principal_sub": "...", "target_name": "...", "result_status": "ok", "ts": "2026-05-25T10:00:00Z", "payload": {"...": "redacted per decision #3"}, "audit_id": "..."}
  ],
  "next_cursor": "Valkey stream cursor (pass back as `since` for the next call)"
}
```

**Handler implementation pattern:** mirror the `client.xrange(stream_key, ...)` call in [backend/src/meho_backplane/ui/routes/broadcast/history.py:182](backend/src/meho_backplane/ui/routes/broadcast/history.py) which already iterates `meho:feed:{tenant_id}` and applies the same filter shape. Factor the xrange + filter + redact loop into a shared `list_recent_events_impl(operator, since, filter, limit)` helper that both this tool and the existing UI history route call. The MCP handler is a thin in-process wrapper, the same shape as `meho.broadcast.overrides.list` ([backend/src/meho_backplane/mcp/tools/broadcast_overrides.py:131](backend/src/meho_backplane/mcp/tools/broadcast_overrides.py)).

**RBAC:** `operator` role. Tenant_admin not required — every tenant member sees their tenant's feed.

[...]

Issue #1090, opened 2026-05-25
Initiative: G6.4 Broadcast meta-tools — broadcast_recent + broadcast_announce + broadcast_watch MCP tools (G6.1 amendment follow-up)


damir-topic:
Parent goal: #217

## Scope

Register the three broadcast meta-tools — `broadcast_recent`, `broadcast_announce`, `broadcast_watch` — the **2026-05-14 G6.1 amendment** (#228) added to scope: *"land as additional Tasks (TBD) on this Initiative's list."* G6.1 closed without filing those Tasks. Initiative #229 G7.1 then shipped a `CLAUDE.md` template (PR #1028 closing #318) whose **broadcast-discipline** section makes these tools load-bearing for cross-operator awareness. The contract is documented; the implementation is not. G6.4 is the catch-up.

**What exists today:** Valkey stream infrastructure ([backend/src/meho_backplane/broadcast/{events,publisher,client,history}.py](backend/src/meho_backplane/broadcast/)), per-tenant SSE feed at `/api/v1/feed`, `meho status --watch` CLI verb ([cli/internal/cmd/status.go:78](cli/internal/cmd/status.go)), MCP resource `meho://tenant/{id}/feed` ([backend/src/meho_backplane/mcp/resources/tenant_feed.py](backend/src/meho_backplane/mcp/resources/tenant_feed.py)), and three PII-override admin tools ([backend/src/meho_backplane/mcp/tools/broadcast_overrides.py](backend/src/meho_backplane/mcp/tools/broadcast_overrides.py) — `meho.broadcast.overrides.{list,set,remove}`). What's missing: the three agent-facing tools.

## Motivation

The G7.1 Layer-2 starter (`docs/examples/consumer-onboarding/CLAUDE.md`, merged in PR #1028) instructs every consumer agent session to follow a four-step contract — before-start `broadcast_recent`, intent `broadcast_announce`, in-flight `broadcast_watch`, completion `broadcast_announce`. The contract is now live in operator-facing docs but the tools the docs name are not registered anywhere. Consumers copying the template will either discover the gap at runtime and lose trust, or silently degrade to the substitute surfaces the template documents as fallbacks (`meho status --watch`, `meho audit who-touched`) — which is what the template currently does honestly, but the load-bearing tools should exist.

## Definition of done

- [ ] `broadcast_recent` MCP tool registered, tenant-scoped, tested (read from Valkey stream with `since` + `filter` args).
[...]

Issue #217, opened 2026-05-11
Goal: Activity broadcast — real-time SSE feed + meho status --watch so operators see colleagues' work as it happens


damir-topic:
## Goal

Ship a real-time activity broadcast across each tenant — a live, filterable stream of every operation flowing through MEHO, surfaced via Slack mirror + SSE web feed + terminal `meho status --watch` — so operators see their colleagues' work *while it happens*, not post-hoc via tickets or CHANGELOG entries.

This is the "shared sessions" feature the team explicitly wants, but built without a custom collaborative chat UI: Slack as the team-side UX, the SSE feed as the always-on dashboard, the audit log as the canonical retrospective trail.

## Why

Today an operator has zero real-time awareness of what colleagues are doing. They find out tomorrow via a ticket, a CHANGELOG bullet, or an INVENTORY entry. By then the moment to coordinate has already passed: collisions happen, redundant work ships, "I didn't know you were on that" is a frequent post-mortem refrain.

Activity broadcast is also the single biggest UX shift from "MEHO is a per-operator CLI" to "MEHO is the team's shared brain" — and it's largely orthogonal to the connector work (G3), so it can ship in parallel without a hard dependency.

## Done-when (testable from the consumer)

- SSE feed at the MEHO instance's `/feed` endpoint, showing every operation in real time, filterable by tenant / principal / target / op-type.
- Slack app posts every operation to a per-tenant configurable channel as a structured card (op-id, principal, target, op, result, link to audit row).
- `meho status --watch` provides a terminal-resident live view.
- At least two operators have used the feed to coordinate or avoid a collision in a real situation (not a contrived drill).
- Read-filter for sensitivity defaults are conservative: some op classes (credential reads, audit queries) broadcast only as aggregate counts in the feed unless explicit per-op opt-in flips them to full detail. Operators can opt their own ops out of broadcast for genuinely sensitive cases.

## Out of scope (explicit)

- Threaded `@meho-bot` mention-in-Slack-triggers-agent — that's part of the v0.2 Agent Runner Goal.
- Activity *aggregation* (heatmaps, weekly summaries, dashboards) — separate follow-up Goal.
- Cross-tenant visibility (a `meho-evoila` admin seeing all tenants' feeds at once) — explicitly disallowed; tenant boundaries hold.

## References

[...]

PR #1106, opened 2026-05-26
refactor(broadcast): G6.4-T4 share xrange+filter helper between MCP recent + UI history

### LINKED ISSUE #1103: Task: G6.4-T4 wire T1's shared list_recent_events helper into the UI history route (preserve fail-soft via fail_soft param)
Parent initiative: #1090

## Target

Wire T1's shared `_list_recent_events_impl` helper into the UI history route so the read-and-redact logic isn't duplicated between the MCP tool and the dashboard SSE bridge:

- MODIFIED `backend/src/meho_backplane/conventions/preamble.py` — N/A (this is the broadcast track, not conventions); the helper lives at `backend/src/meho_backplane/mcp/tools/broadcast.py` per T1.
- MODIFIED `backend/src/meho_backplane/mcp/tools/broadcast.py` — promote `_list_recent_events_impl` to a public `list_recent_events` (or move into a shared module under `backend/src/meho_backplane/broadcast/`) + add a `fail_soft: bool = False` parameter (or two thin wrappers) so callers can pick the failure semantics.
- MODIFIED `backend/src/meho_backplane/ui/routes/broadcast/history.py` — call the shared helper in `fail_soft=True` mode instead of its current in-route xrange + filter + redact loop.
- MODIFIED tests — extend or move tests to cover the shared helper directly; UI integration test asserts the route still degrades gracefully on Valkey errors.

## Context

T1 (G6.4-T1, #1091) introduced the helper as a deliberate factor-out so both the new MCP tool and the existing UI history route could share the xrange + filter + redact logic. The implement agent ultimately chose NOT to update the UI route, because the two callers have different failure semantics:

- **MCP tool (`broadcast.recent`)** — fail-loud. An agent calling the tool needs to know if Valkey is unreachable; a silent empty result would be a correctness bug.
- **UI history route** — fail-soft. A Valkey blip shouldn't 500 the dashboard; an empty event list with a small banner is the right UX.

This semantic divergence was documented in the `broadcast.py` module docstring (see PR #1097's Follow-up section in the PR body). The result is two code paths that read the same Valkey stream, apply the same filters, and call the same `redact_payload` — but live in different files.

This Task picks up the deferred wiring without regressing the UI's fail-soft contract. The simplest shape: extend the helper with a `fail_soft: bool` flag (or add a thin `list_recent_events_fail_soft` wrapper that catches `RedisError` and returns `[]`); update the UI route to call the wrapper.

**Why this is worth doing:** when the redaction policy evolves (which is likely — decision #3 is conservative-by-default and may add per-tenant opt-ins), the change needs to land in exactly one place. Today it would need to land in two and stay in sync, which is exactly the kind of drift that produces incidents.

## Current state

- `backend/src/meho_backplane/mcp/tools/broadcast.py` defines `_list_recent_events_impl` (private helper for the MCP `broadcast.recent` tool).
- `backend/src/meho_backplane/ui/routes/broadcast/history.py` has its own xrange + filter + redact loop (predates T1; left untouched per the documented divergence).
- Both reach the same Valkey stream + apply the same redaction + return the same event shape; the divergence is purely error-handling.

## Desired state

- One shared implementation; both callers use it.
- The fail-loud vs fail-soft choice is a parameter (or per-wrapper), not a duplicated function body.
- Tests cover the shared helper directly; UI integration test still asserts the dashboard degrades gracefully.

## Acceptance criteria

- [ ] The xrange + filter + redact logic is implemented in exactly one place (no duplicated function body across MCP tool + UI route).
- [ ] The MCP tool (`broadcast.recent`) keeps its fail-loud contract — Valkey unreachable still surfaces as an MCP error.
- [ ] The UI history route keeps its fail-soft contract — Valkey unreachable still returns an empty event list and a small UX banner (no 500).
- [ ] A unit test asserts the fail-soft caller catches `redis.exceptions.RedisError` and returns `([], None)` (or equivalent empty sentinel).
- [ ] A unit test asserts the fail-loud caller re-raises the same `RedisError`.
- [ ] Integration test: dashboard returns 200 + empty history when Valkey is unreachable (mirrors the existing test if one exists).
- [ ] No regression in `backend/tests/test_mcp_tool_broadcast_recent.py` (T1's full test suite still passes).
- [ ] PR body Decisions section documents the helper's new location + the `fail_soft` parameter contract for future maintainers.
- [ ] CI green; ruff + mypy clean.
- [ ] Parent Initiative #1090 body checklist line for this Task is ticked (add a new row under "Execution tasks").

## Out of scope

- Changing the redaction policy (decision #3 stays).
- Migrating the UI route to async iterators / streaming responses — the existing pagination shape is fine.
- Extracting the helper into a separate package — leave it in `backend/src/meho_backplane/broadcast/` or `mcp/tools/broadcast.py`; the right home is the implementer's call.
- Touching `broadcast_announce` or `broadcast_watch` helpers — those are write-side and BLOCK-read-side respectively; out of scope here.

## References

- Parent Initiative: #1090 (G6.4 Broadcast meta-tools).
- T1 origin: #1091 → PR #1097 — implement agent's Follow-up section in the PR body documents the deferral.
- UI route: [backend/src/meho_backplane/ui/routes/broadcast/history.py](backend/src/meho_backplane/ui/routes/broadcast/history.py).
- MCP tool: [backend/src/meho_backplane/mcp/tools/broadcast.py](backend/src/meho_backplane/mcp/tools/broadcast.py) (look for `_list_recent_events_impl`).
- Redaction primitive: [backend/src/meho_backplane/broadcast/events.py](backend/src/meho_backplane/broadcast/events.py) (`redact_payload`).
- Best-practices anchors:
  - `.claude/skills/explore-and-refactor/python_best_practices.md` (DRY discipline at the helper-extraction boundary)
  - `.claude/skills/implement-issue/ai_engineering_best_practices.md` (PII redaction policy locality — single source of truth)

damir-topic:
## Summary

- Collapses the duplicate `xrange` + filter + redact-aware parse body that previously lived in both the MCP `broadcast.recent` tool and the UI `/ui/broadcast/history` route into a single shared module at `backend/src/meho_backplane/broadcast/history.py`. T1 (#1091) had deferred this unification because the two callers' failure shapes differ; this Task lands the helper with that contract divergence handled explicitly via two named wrappers.
- The MCP tool keeps its fail-loud contract (`list_recent_events_strict` re-raises `RedisError`; the dispatcher maps to `-32603`). The UI route keeps its fail-soft contract (`list_recent_events_fail_soft` returns `{"events": [], "next_cursor": None}` on `RedisError`; the pane renders its empty state, not a 500).
- T1's full test suite (89 passed, 10 docker-gated skipped) still passes verbatim. The 15 UI replay tests pass (one new fail-soft case added). 10 new unit tests pin the helper-level contract.

### Helper location: `backend/src/meho_backplane/broadcast/history.py` (new shared module)

The issue body left the home open: "leave it in `backend/src/meho_backplane/broadcast/` or `mcp/tools/broadcast.py`; the right home is the implementer's call." Picked `broadcast/` because:

- It's a read-side counterpart to the existing `broadcast/publisher.py`. The `broadcast/` namespace is the proper home for stream-level primitives shared across consumers.
- A broadcast-read helper isn't MCP-specific. Forcing the UI to import from `mcp/tools/` would be a layering-wrong dependency (UI -> MCP).
- T2/T3 (the `broadcast.announce` and `broadcast.watch` tools, which share this file in `mcp/tools/`) still need the parse primitives (`_event_matches` / `_parse_entry` / `_stream_key`); they import them from the new module's location. No re-export shims.

### Failure-shape API: two named wrappers around a shared core, NOT `fail_soft: bool`

The issue body suggested either approach. Picked two wrappers because:

- A `fail_soft: bool` flag gates error-handling semantics on a parameter -- the "flag argument" anti-pattern Beck and Martin both flag as a code smell. The caller's intent ("I tolerate a missing read") would live in the value of an argument rather than the name of the function.
- Two named wrappers make grep ("show me every fail-soft call site") and diff review trivial. A one-character flip on a flag changes a load-bearing contract.
- The extra function definition cost is one body per wrapper -- trivial in this case because the wrappers are 3 lines each.

### Error-class split: `InvalidSinceError` (domain) vs `RedisError` (transport)

The shared helper raises a generic `InvalidSinceError` (a `ValueError` subclass) on a malformed `since` argument rather than the MCP-specific `McpInvalidParamsError`. This keeps the helper library-agnostic:

- The MCP handler catches `InvalidSinceError` and maps it to `-32602` Invalid Params at the wire layer.
- The UI fail-soft wrapper deliberately does NOT swallow `InvalidSinceError` -- a malformed `since` is a programmer error on the caller's part, not a Valkey teardown. Fail-soft is "the data store is unavailable", not "the caller passed garbage".
- Only `RedisError` (the redis-py 7.4 base class for `ConnectionError` / `TimeoutError` / etc. -- verified via `uv run python -c 'import redis.exceptions; ...'` against the installed library) is in the fail-soft catch clause. Catching bare `Exception` (which the previous UI body did) would swallow `AttributeError` / `TypeError` from a refactor bug.

### UI projection: history pane stays audit-event only

The shared helper returns both `BroadcastEvent` (audit-driven) and `AgentAnnouncementEvent` (agent-authored) -- T1's `_parse_entry` handles the dual-event shape via the `event_kind` discriminator. The UI route filters back to audit-only at the call site (`_is_audit_event`) because:

- The existing `_event_row.html` template binds to audit columns (`op_class` / `op_id` / `result_status` / `payload`) that announcements don't carry. Surfacing announcement-shape events through the existing template would render blank cells.
- The live SSE feed at `api/v1/feed.py` also drops announcements today (only `BroadcastEvent.model_validate_json` -- announcements fail validation and are skipped). Keeping the history pane symmetric with the live feed preserves the existing contract exactly.
- Adding announcement rendering is a follow-up that lives on the row template, not this route.

## Test plan

- [x] `uv run ruff check src/ tests/` -- clean
- [x] `uv run ruff format --check src/ tests/` -- 645 files already formatted
- [x] `uv run mypy src/` -- 314 source files, no issues
- [x] `uv run pytest tests/test_broadcast_history.py -x -q` -- 10 passed (new wrapper-contract suite)
- [x] `uv run pytest tests/test_mcp_tool_broadcast_recent.py -x -q` -- 33 passed, 4 docker-skipped (T1's suite, unchanged)
- [x] `uv run pytest tests/test_mcp_tool_broadcast_watch.py tests/test_mcp_tool_broadcast_announce.py -x -q` -- T2/T3 suites, all green
- [x] `uv run pytest tests/test_ui_broadcast_wall_replay.py -x -q` -- 15 passed (one new fail-soft case added)
- [x] Full broadcast-related sweep: 247 passed, 14 skipped across 10 test files

Acceptance criteria coverage:

- [x] `xrange` + filter + redact logic in EXACTLY one place -- only `_list_recent_events_core` does the loop
- [x] MCP tool keeps fail-loud contract (`list_recent_events_strict`)
- [x] UI history route keeps fail-soft contract (`list_recent_events_fail_soft`)
- [x] Unit test: fail-soft caller catches `RedisError` returns `([], None)` -- `test_fail_soft_swallows_redis_error_returns_empty` parametrised across `RedisError` / `ConnectionError` / `TimeoutError`
- [x] Unit test: fail-loud caller re-raises `RedisError` -- `test_strict_re_raises_redis_error`
- [x] Integration test: dashboard returns 200 + empty history when Valkey unreachable -- `test_history_returns_200_and_empty_when_valkey_unreachable`
- [x] T1's test suite still passes (89 passed in `test_mcp_tool_broadcast_recent.py`)
- [x] PR body Decisions section documents the helper's new location + the contract (this section)
- [x] CI green; ruff + mypy clean

## Out of scope

- Changing the redaction policy -- decision #3 stays. Helpers surface what's on the stream verbatim.
- Migrating to async iterators / streaming responses -- the existing pagination shape is fine.
- Adding announcement-event rendering to the UI history pane -- requires row-template work; logged as adjacent follow-up.
- Ticking the parent Initiative #1090's T4 checkbox in its body -- handled by auto-close on merge plus the orchestrator's post-merge sweep.

Closes #1103

---

## Follow-up (auto-implement-issue, iteration 1, 2026-05-26)

Addressed review findings from `/review-pr-slim` REQUEST_CHANGES verdict on PR #1106:

- **m1** `a686424` -- promote 5 cross-module helpers (`event_matches` / `parse_entry` / `stream_key` / `parse_since` / `default_since_ms`) from `_`-prefixed private to public; add to `broadcast.history.__all__`; update import sites in `mcp/tools/broadcast.py` + `tests/test_mcp_tool_broadcast_recent.py` + the `ui/routes/broadcast/history.py` docstring xref.
- **B1** `c26f505` -- regenerate `cli/api/openapi.json` after the prior commit's UI history docstring edit; `cli/internal/api/client.gen.go` byte-identical (the OpenAPI `description` field is metadata, not schema-bearing, so the generated Go client doesn't carry it). `CLI API snapshot freshness` CI now green; `make snapshot-openapi` idempotent on re-run; `go build ./...` clean.

Deferred (recorded as adjacent findings; orchestrator may file follow-up issues):

- **m2** -- the original commit message on 161a7db references a "re-export shim" that doesn't exist. Cosmetic only -- the contract documented in the PR body Decisions section is accurate. Not amending: published commits are immutable by hard-rule.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of broadcast history display by gracefully handling temporary Redis connectivity issues; the history view now displays the empty state instead of failing when connection errors occur.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/1106?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/meho_backplane/broadcast/__init__.py
+++ b/backend/src/meho_backplane/broadcast/__init__.py
@@ -23,6 +23,13 @@ from meho_backplane.broadcast.events import (
     classify_op,
     redact_payload,
 )
+from meho_backplane.broadcast.history import (
+    DEFAULT_WINDOW_MINUTES,
+    OP_CLASS_ENUM,
+    InvalidSinceError,
+    list_recent_events_fail_soft,
+    list_recent_events_strict,
+)
 from meho_backplane.broadcast.overrides import (
     compute_effective_broadcast_detail,
     invalidate_tenant_cache,
@@ -45,14 +52,19 @@ __all__ = [
     "BROADCAST_EVENTS_PUBLISHED_TOTAL",
     "BROADCAST_MAXLEN",
     "BROADCAST_PUBLISH_ERRORS_TOTAL",
+    "DEFAULT_WINDOW_MINUTES",
+    "OP_CLASS_ENUM",
     "AgentAnnouncementEvent",
     "BroadcastEvent",
+    "InvalidSinceError",
     "broadcast_readiness_probe",
     "classify_op",
     "compute_effective_broadcast_detail",
     "dispose_broadcast_client",
     "get_broadcast_client",
     "invalidate_tenant_cache",
+    "list_recent_events_fail_soft",
+    "list_recent_events_strict",
     "publish_agent_announcement",
     "publish_event",
     "read_request_override",
--- a/backend/src/meho_backplane/broadcast/history.py
+++ b/backend/src/meho_backplane/broadcast/history.py
@@ -96,8 +96,13 @@ __all__ = [
     "DEFAULT_WINDOW_MINUTES",
     "OP_CLASS_ENUM",
     "InvalidSinceError",
+    "default_since_ms",
+    "event_matches",
     "list_recent_events_fail_soft",
     "list_recent_events_strict",
+    "parse_entry",
+    "parse_since",
+    "stream_key",
 ]
 
 _log = structlog.get_logger(__name__)
@@ -138,7 +143,7 @@ OP_CLASS_ENUM: Final[tuple[str, ...]] = (
 
 
 class InvalidSinceError(ValueError):
-    """Raised when :func:`_parse_since` rejects the caller's input.
+    """Raised when :func:`parse_since` rejects the caller's input.
 
     Surfaced as a domain error so a wire-layer caller (the MCP handler)
     can map it to ``-32602`` Invalid Params, and a non-wire caller (the
@@ -147,7 +152,7 @@ class InvalidSinceError(ValueError):
     """
 
 
-def _stream_key(tenant_id: object) -> str:
+def stream_key(tenant_id: object) -> str:
     """Build the per-tenant Valkey stream key.
 
     Mirrors :func:`meho_backplane.broadcast.publisher._stream_key` and
@@ -157,7 +162,7 @@ def _stream_key(tenant_id: object) -> str:
     return f"meho:feed:{tenant_id}"
 
 
-def _default_since_ms() -> int:
+def default_since_ms() -> int:
     """Return the inclusive bare-millisecond cursor for the default window.
 
     Default window is :data:`DEFAULT_WINDOW_MINUTES` minutes before
@@ -170,7 +175,7 @@ def _default_since_ms() -> int:
     return max(now_ms - DEFAULT_WINDOW_MINUTES * _MS_PER_MINUTE, 0)
 
 
-def _parse_since(raw: str | None) -> str:
+def parse_since(raw: str | None) -> str:
     """Translate the caller-supplied ``since`` into an ``XRANGE`` ``min``.
 
     Three shapes accepted:
@@ -199,7 +204,7 @@ def _parse_since(raw: str | None) -> str:
     Invalid inputs raise :class:`InvalidSinceError`.
     """
     if raw is None:
-        return str(_default_since_ms())
+        return str(default_since_ms())
     if _looks_like_stream_cursor(raw):
         return f"({raw}"
     return _iso8601_to_min_cursor(raw)
@@ -238,7 +243,7 @@ def _iso8601_to_min_cursor(raw: str) -> str:
     return str(int(parsed.timestamp() * 1000))
 
 
-def _event_matches(
+def event_matches(
     event: BroadcastEvent | AgentAnnouncementEvent,
     *,
     op_class: str | None,
@@ -286,7 +291,7 @@ def _event_matches(
     return True
 
 
-def _parse_entry(
+def parse_entry(
     entry_id: str,
     fields: dict[str, str],
     *,
@@ -393,12 +398,12 @@ async def _list_recent_events_core(
     re-raise.
     """
     client = get_broadcast_client()
-    stream_key = _stream_key(operator.tenant_id)
-    min_cursor = _parse_since(since)
+    key = stream_key(operator.tenant_id)
+    min_cursor = parse_since(since)
     raw_entries = cast(
         "list[tuple[str, dict[str, str]]]",
         await client.xrange(
-            stream_key,
+            key,
             min=min_cursor,
             max=_XRANGE_END,
             count=limit,
@@ -407,10 +412,10 @@ async def _list_recent_events_core(
 
     matched: list[dict[str, Any]] = []
     for entry_id, fields in raw_entries:
-        event = _parse_entry(entry_id, fields, stream_key=stream_key)
+        event = parse_entry(entry_id, fields, stream_key=key)
         if event is None:
             continue
-        if not _event_matches(
+        if not event_matches(
             event,
             op_class=op_class,
             principal=principal,
@@ -512,6 +517,6 @@ async def list_recent_events_fail_soft(
         _log.warning(
             "broadcast_history_fetch_failed",
             error_class=type(exc).__name__,
-            stream_key=_stream_key(operator.tenant_id),
+            stream_key=stream_key(operator.tenant_id),
         )
         return {"events": [], "next_cursor": None}
--- a/backend/src/meho_backplane/broadcast/history.py
+++ b/backend/src/meho_backplane/broadcast/history.py
@@ -1,0 +1,517 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""Shared read-side helper for the per-tenant Valkey broadcast stream.
+
+G6.4-T4 (#1103) factors the xrange + filter + redaction-aware parse
+loop into a single implementation that both the MCP ``broadcast.recent``
+tool (Task #1091) and the UI ``/ui/broadcast/history`` route (Task #869)
+go through. Two callers had previously duplicated this body because
+their **failure shape** differs:
+
+* The MCP tool is **fail-loud** -- an agent calling ``broadcast.recent``
+  needs a Valkey teardown to surface as a JSON-RPC ``-32603`` Internal
+  Error; a silent empty result would be a correctness bug.
+* The UI history route is **fail-soft** -- a transient broadcast-subchart
+  blip should degrade the Last-24h pane to its empty state, not 500 the
+  fragment.
+
+The previous design solved this by writing the xrange + parse loop
+twice. Every redaction-policy or schema tweak then had to land in two
+places and stay in sync, which is exactly the kind of drift that
+produces incidents (see PR #1097 follow-up notes).
+
+This module factors the loop into a single core helper plus two named
+wrappers, one per failure shape. New callers should pick the wrapper
+whose name matches the failure contract their call site holds; the
+core helper is internal because a bare call elides the contract choice.
+
+Why two wrappers instead of ``fail_soft: bool``
+================================================
+
+A ``fail_soft: bool`` flag would compress the API to one function, but
+it gates the function's error-handling semantics on a parameter -- the
+"flag argument" anti-pattern Beck / Martin both flag as a code smell.
+The caller's intent ("I tolerate a missing read") would live in the
+*value* of an argument rather than the *name* of the function, which
+hurts grep ("show me every fail-soft call site"), hurts diff review
+(a one-character flip changes a load-bearing contract), and reads
+ambiguously at the call site. Two named wrappers cost one extra
+function definition each; that is a cheap trade for a clearer contract.
+
+PII inheritance
+===============
+
+Every entry on ``meho:feed:{tenant_id}`` was already redacted by
+:func:`~meho_backplane.broadcast.events.redact_payload` at publish time
+(the publish-on-write hook in
+:mod:`~meho_backplane.broadcast.publisher`). The helpers in this module
+surface what's on the stream verbatim -- they do NOT re-redact, do NOT
+mutate the payload field, and do NOT call ``redact_payload``. The
+contract flows: publisher redacts -> stream stores redacted -> every
+reader (SSE, UI history, MCP tools) surfaces redacted-as-is.
+
+Tenant scoping
+==============
+
+The stream key is derived exclusively from the :class:`Operator`'s
+verified ``tenant_id`` JWT claim. The helpers take no separate
+``tenant_id`` argument; cross-tenant reads are structurally impossible.
+
+References
+----------
+
+* Initiative: #1090 (G6.4 broadcast meta-tools).
+* T1 origin (the MCP tool that first factored the helper out):
+  Task #1091 / PR #1097.
+* T4 unifier (this module): Task #1103.
+* Sibling primitive: :mod:`meho_backplane.broadcast.publisher`
+  (write-side; the fail-loud / fail-open split there mirrors the
+  fail-loud / fail-soft split here).
+* Sibling read surface, unmodified: the SSE feed generator at
+  :func:`meho_backplane.api.v1.feed._feed_generator`. It uses XREAD
+  BLOCK (open-ended stream tail), not XRANGE (finite window), so it
+  shares the parse primitives in spirit but cannot reuse this body
+  verbatim.
+* Valkey XRANGE: https://valkey.io/commands/xrange/
+"""
+
+from __future__ import annotations
+
+import json
+import time
+from datetime import UTC, datetime
+from typing import Any, Final, cast
+
+import structlog
+from pydantic import ValidationError
+from redis.exceptions import RedisError
+
+from meho_backplane.auth.operator import Operator
+from meho_backplane.broadcast.agent_events import AgentAnnouncementEvent
+from meho_backplane.broadcast.client import get_broadcast_client
+from meho_backplane.broadcast.events import BroadcastEvent
+
+__all__ = [
+    "DEFAULT_WINDOW_MINUTES",
+    "OP_CLASS_ENUM",
+    "InvalidSinceError",
+    "list_recent_events_fail_soft",
+    "list_recent_events_strict",
+]
+
+_log = structlog.get_logger(__name__)
+
+#: Milliseconds per minute. The default ``since`` window is the last
+#: :data:`DEFAULT_WINDOW_MINUTES` minutes; computed as
+#: ``now_ms - DEFAULT_WINDOW_MINUTES * _MS_PER_MINUTE``.
+_MS_PER_MINUTE: Final[int] = 60_000
+
+#: Default look-back window when the caller omits ``since``. 30 minutes
+#: -- smaller than the 24h retention ceiling so the default response
+#: stays bounded on a busy tenant. The MCP tool's input schema documents
+#: this default; the UI history route does NOT use it (the UI always
+#: passes an explicit retention-bound ``since``).
+DEFAULT_WINDOW_MINUTES: Final[int] = 30
+
+#: ``XRANGE`` end anchor. ``"+"`` is Valkey's "latest available entry"
+#: sentinel, so the pull spans from the ``since`` cursor to the live
+#: tail. Mirrors :mod:`~meho_backplane.ui.routes.broadcast.stream` and
+#: :mod:`~meho_backplane.api.v1.feed`.
+_XRANGE_END: Final[str] = "+"
+
+#: Allowed ``op_class`` filter values. The four values the issue body
+#: example listed (``read|write|credential_read|audit_query``) plus
+#: ``credential_mint`` and ``other`` -- :func:`classify_op` emits those
+#: too. Restricting to the four would force the caller to omit the
+#: filter when chasing a freshly-minted credential (e.g.
+#: ``harbor.robot.create``) or an unmapped HTTP route, which defeats
+#: the filter's purpose.
+OP_CLASS_ENUM: Final[tuple[str, ...]] = (
+    "read",
+    "write",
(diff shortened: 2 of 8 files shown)


damir-topic:
@coderabbitai review

damir-topic (commented):
## /review-pr-slim summary

**Anchor task:** #1103 — G6.4-T4 wire T1's shared `list_recent_events` helper into the UI history route (preserve fail-soft)
**Surface area:** backend (Python / FastAPI / Pydantic v2 / redis-py 7.4)
**Best-practices packs audited:** python (DRY discipline at helper-extraction boundary, no flag-arg anti-pattern, named error classes, structlog discipline)
**Verdict:** REQUEST_CHANGES

> Posted as a `COMMENT` review because the GitHub web UI blocks own-author `APPROVE`. The explicit verdict above is the contract.

### Acceptance criteria

- [x] `xrange` + filter + redact logic in EXACTLY one place — only `_list_recent_events_core` in `broadcast/history.py:360-435` issues a `client.xrange(...)`; grep `\.xrange\(` across `backend/src/meho_backplane/` returns exactly one production hit (verified).
- [x] MCP tool (`broadcast.recent`) keeps fail-loud contract — `_handler_recent` calls `list_recent_events_strict`, lets `RedisError` propagate to the MCP dispatcher's `-32603` arm. `test_strict_re_raises_redis_error` pins it.
- [x] UI history route keeps fail-soft contract — `_fetch_history` calls `list_recent_events_fail_soft`, which catches `RedisError` and returns `{"events": [], "next_cursor": None}`. `test_history_returns_200_and_empty_when_valkey_unreachable` asserts 200 + empty-state copy at the HTTP edge.
- [x] Unit test: fail-soft caller catches `RedisError` → `([], None)` — `test_fail_soft_swallows_redis_error_returns_empty` is parametrised across `RedisError`, `ConnectionError`, `TimeoutError`. Verified redis-py 7.4 hierarchy (both subclasses inherit `RedisError`).
- [x] Unit test: fail-loud caller re-raises `RedisError` — `test_strict_re_raises_redis_error`.
- [x] Integration test: dashboard returns 200 + empty history when Valkey unreachable — `test_history_returns_200_and_empty_when_valkey_unreachable`.
- [x] T1's test suite still passes — `tests/test_mcp_tool_broadcast_recent.py` 33 passed, 4 docker-skipped after the import shift to `broadcast.history`.
- [x] PR body Decisions section documents helper location + `fail_soft` contract + error-class split.
- [ ] CI green; ruff + mypy clean — **ruff/mypy/pytest are clean locally, but the `CLI API snapshot freshness` CI job is FAILING.** See blocker below.
- [ ] Parent Initiative #1090's "Execution tasks" `#1103` checkbox ticked — PR body explicitly defers to the orchestrator's post-merge sweep. T1/T2/T3 checkboxes on #1090 are also still `[ ]` despite having merged, so this is a consistent pattern (post-merge sweep handles all four). Not a blocker for this PR.

|---|---|---|---|---|---|
| #1103 (Task) | In Progress | `Closes #1103` | yes | full once snapshot fixed | close on merge |
| #1090 (Initiative) | Todo | parent of #1103 | — | does not complete; T1/T2/T3 already merged but checklist deferred to post-merge sweep | bump status, post-merge tick T1-T4 |
| #217 (Goal) | Todo | grandparent | — | does not complete | no action |

### Predicted vs. actual diff

- ✅ `backend/src/meho_backplane/broadcast/history.py` (new) — predicted; single core + two named wrappers + dedicated `InvalidSinceError`.
- ✅ `backend/src/meho_backplane/broadcast/__init__.py` — both wrappers + `InvalidSinceError` exported, no stale `_list_recent_events_impl` re-export.
- ✅ `backend/src/meho_backplane/mcp/tools/broadcast.py` — `_handler_recent` is now a 9-line delegate that maps `InvalidSinceError` → `McpInvalidParamsError`.
- ✅ `backend/src/meho_backplane/ui/routes/broadcast/history.py` — `_fetch_history` delegates to the fail-soft wrapper + post-filters via `_is_audit_event` (consistent with live SSE feed's de-facto AgentAnnouncementEvent drop).
- ✅ `backend/tests/test_broadcast_history.py` (new, 10 cases) — happy-path shape symmetry, parametrised RedisError fail-soft swallow, strict re-raise, `InvalidSinceError` propagation through both wrappers, structural tenant scoping.
- ✅ `backend/tests/test_mcp_tool_broadcast_recent.py` — import shift + wire-level `test_invalid_since_iso_returns_invalid_params_at_wire`.
- ✅ `backend/tests/test_ui_broadcast_wall_replay.py` — `test_history_returns_200_and_empty_when_valkey_unreachable`.
- 🕳️ MISSING: `cli/api/openapi.json` regeneration. The UI history handler docstring changed; that docstring is the route's OpenAPI description, snapshotted in `cli/api/openapi.json`. The `CLI API snapshot freshness` CI job fails on the diff. See blocker below.

### Integration verification

- [x] `ruff check src/meho_backplane/broadcast/history.py …` — clean.
- [x] `ruff format --check` (4 files) — already formatted.
- [x] `mypy src/meho_backplane/broadcast/history.py …` — `Success: no issues found in 4 source files`.
- [x] `pytest tests/test_broadcast_history.py tests/test_mcp_tool_broadcast_recent.py tests/test_ui_broadcast_wall_replay.py` — **56 passed, 4 docker-skipped**.
- [x] `pytest tests/test_mcp_tool_broadcast_watch.py tests/test_mcp_tool_broadcast_announce.py` — **58 passed, 6 docker-skipped** (T2/T3 unaffected by the move).
- [x] `pytest tests/ -k broadcast` — **349 passed, 33 skipped** (full broadcast sweep, no regressions).
- [⚠️ not run] UI flow exercised in browser — integration test covers 200 + empty-state path at the HTTP edge; full browser exercise not done.
- [⚠️ not run] `docker compose up` — change is delegate-only; no compose verification needed for this slice.

### Findings

- 🛑 **1 blocker** — stale CLI API OpenAPI snapshot (CI red).
- ⚠️ **0 majors**.
- 💡 **2 minors** — cross-module private-name imports; commit-message "re-export" claim that the code doesn't implement.
- ❓ **0 questions**.
- 👏 **4 praise** — see below.

---

### 🛑 Blocker 1 — `CLI API snapshot freshness` CI check is FAILING; the committed OpenAPI snapshot does not match the regenerated one

**What:** The PR changed the docstring on the UI history handler (`backend/src/meho_backplane/ui/routes/broadcast/history.py`'s inner `_handler`) from
> "Pulls the session tenant's last-24h events via a finite ``XRANGE`` …"

to
> "Pulls the session tenant's last-24h events via the shared fail-soft read helper …"

That docstring is the route's OpenAPI `description` (FastAPI extracts the handler's first-line docstring into the OpenAPI route description). The CI job `CLI API snapshot freshness` (workflow `CI`, run 26440449640, job 77833287213) regenerates `cli/api/openapi.json` from the live backend and `git diff --exit-code`s the committed file against the regenerated one. The diff shows only the description string changed — but `--exit-code` fails on any diff. CI step "Fail if the committed snapshot / client are stale" exited 1; the workflow is red.

**Why it matters:** A red required check blocks merge under the repo's auto-merge / branch-protection rules. The PR cannot ship as-is. Independently, a stale OpenAPI snapshot drifts the Go client's metadata away from the actual backend response shape — not a correctness regression here (only a description string changed, no signatures), but the stale-snapshot guard exists exactly so the Go client's view of the backend stays honest.

**Suggested fix:**

```bash
cd cli
make snapshot-openapi
make generate
git add cli/api/openapi.json cli/internal/api/client.gen.go
git commit --signoff -m "refactor(broadcast): regenerate OpenAPI snapshot after UI history docstring update"
git push
```

(The `client.gen.go` re-gen will likely be a no-op for this change — the description doesn't show up in the typed client — but running both commands matches the CI workflow's expectation.)

**Source:** CI workflow log at https://github.com/evoila/meho/actions/runs/26440449640/job/77833287213 — the inline diff shows the route description swap.

---

### 💡 Minor 1 — Cross-module private-name imports for `_event_matches` / `_parse_entry` / `_stream_key` / `_parse_since` / `_default_since_ms`

**What:** `backend/src/meho_backplane/mcp/tools/broadcast.py` (lines 97-101) imports `_event_matches`, `_parse_entry`, `_stream_key` directly from `broadcast.history`. `backend/tests/test_mcp_tool_broadcast_recent.py` (lines 58-62) imports `_default_since_ms` and `_parse_since` similarly. PEP 8 / Python convention treats the leading underscore as "module-private"; cross-module imports of underscore-prefixed names are a code smell.

**Why it matters:** The contract a reader infers from the underscore is "internal to this module; don't import from outside". When external callers import these anyway, the underscore stops meaning "internal" and starts meaning "internal-but-shared-via-trust", which is a silent contract change. Future maintainers refactoring `broadcast.history` may rename or restructure those primitives confidently because the underscore signals freedom to do so — and then T2/T3's broadcast watch tool will break unexpectedly.

**Suggested fix:** Two clean shapes:

1. **Rename to public** — drop the leading underscore on the five names that are documented as shared: `event_matches`, `parse_entry`, `stream_key`, `parse_since`, `default_since_ms`. Add to `__all__`. Tests stay valid white-box references but no longer trip the smell.
2. **Move to a sub-private module** — put the five names in `backend/src/meho_backplane/broadcast/_parse.py` with a module docstring "package-private, shared between `broadcast.history` and `mcp.tools.broadcast`". The names keep their underscores, but the module name carries the "internal" signal instead.

Either is fine; #1 is the smaller diff. Suitable as a small follow-up; not a blocker.

**Source:** PEP 8 § "Public and Internal Interfaces" — https://peps.python.org/pep-0008/#public-and-internal-interfaces

---

### 💡 Minor 2 — Commit-message claim about "re-export" doesn't match the code

**What:** The commit body for `161a7db` ends with:
> "T1's primitives the shared module re-exports (`_event_matches` / `_parse_entry` / `_stream_key` / `_parse_since` / `_default_since_ms`) are also used by T3's `broadcast.watch` handler -- they stay importable from the MCP module via re-export so T3's XREAD path doesn't need to chase the move."

The code doesn't re-export those names through the MCP module — `mcp/tools/broadcast.py` imports them directly from `broadcast.history` (lines 97-101). There is no re-export shim anywhere in this PR. The implementation is sound (direct import is fine), but the commit message reads as if there's a shim that doesn't exist.

**Why it matters:** A future maintainer chasing "where do these primitives live now" reads the commit body, expects a re-export shim, and is confused when there isn't one. Low-impact discrepancy; correctness unaffected.

**Suggested fix:** None mandatory. If a fix-up commit lands for the CLI snapshot, the new commit body can rephrase as "imported directly from `broadcast.history`".

---

### Docs

- [x] `docs/codebase/ui.md` (lines 387, 442, 599-628) — describes the history route's XRANGE shape + fail-soft contract. Still accurate after the refactor (the helper does XRANGE under the hood, and the fail-soft contract is preserved). Doc verified unchanged-but-correct. A follow-up could add "Implementation delegates to `meho_backplane.broadcast.history.list_recent_events_fail_soft`" for traceability.
- [x] `docs/codebase/backend.md` — broadcast substrate section (lines 127-219) does not mention the read-side helper at all. The new shared helper module is a non-trivial new surface in `broadcast/`; adding a paragraph alongside the publisher one would be a small follow-up improvement (not a blocker).

### What was done well (👏)

- **👏 Single-implementation discipline is bulletproof.** `client.xrange(` grep across all of `backend/src/meho_backplane/` returns exactly one production hit (in the new core). The MCP tool and UI route are 100% delegate-only. The "factor out and use" contract is preserved end-to-end.
- **👏 The two-wrapper choice is correctly defended in the PR body and the helper docstring.** "Flag-argument anti-pattern (Beck / Martin) — value of an argument gates load-bearing semantics, hurts grep, hurts diff review" is the right diagnosis for `fail_soft: bool`. The wrappers are 9 lines each + share `_list_recent_events_core`; the cost is minimal.
- **👏 `InvalidSinceError` is the right shape.** A domain error (`ValueError` subclass) lets the MCP handler map to `-32602` and the UI fail-soft wrapper selectively propagate without taking the MCP error vocabulary as a dependency. The contract is symmetric — both wrappers propagate it; only `_fail_soft` swallows `RedisError`. Tests pin both directions.
- **👏 The parametrised RedisError swallow test.** `test_fail_soft_swallows_redis_error_returns_empty` runs against `RedisError`, `ConnectionError`, `TimeoutError` to lock the contract across all three teardown classes. A `ConnectionError`-only catch would still 500 the UI on a `TimeoutError`; this test prevents that regression.

---

## Post-merge actions (do after merge, once blocker is fixed)

### Auto-close (should happen automatically on merge)

- #1103 — verify it actually closed.

### Project-board status bumps

- #1103 → `Done`.
- #1090 (Initiative) → check that all four execution-task checkboxes (T1-T4) are now ticked; if T1/T2/T3 haven't been ticked yet either (consistent with the deferred-sweep pattern), tick all four in one edit.

### Completion comment to post on close

- #1103: "Closed by #1106. The xrange + filter + redact loop now lives in `meho_backplane.broadcast.history` with two named wrappers (`list_recent_events_strict` for MCP fail-loud, `list_recent_events_fail_soft` for UI fail-soft); 10 new unit tests pin the contract divergence. Verified: 349 passed in the broadcast sweep, ruff/mypy clean, CI snapshot regenerated."

damir-topic:
@coderabbitai review

Issue #1142, opened 2026-05-26
Task: G0.14-T1 dispatcher resolver error surfacing — catch NoMatchingConnector + AmbiguousConnectorResolution in typed branch + unify /probe with dispatch (signals 7, 8, 19)


damir-topic:
Parent goal: #221
Parent initiative: #1139

## Problem

Three related defects at the dispatcher-resolver boundary cause
operator-visible bare HTTP 500s — while the resolver itself emits
the most-diagnostic error messages in the entire MEHO surface.

**Signal 7 — typed branch silently swallows `NoMatchingConnector`.**
[dispatcher.py:216-244](backend/src/meho_backplane/operations/dispatcher.py#L216):

```python
if descriptor.source_kind == "ingested":
    try:
        connector_cls = resolve_connector(target)
    except NoMatchingConnector:
        return None, "no_connector"        # explicit label
    return get_or_create_connector_instance(connector_cls), None
if descriptor.source_kind in ("typed", "composite") and target is not None:
    try:
        optional_cls = resolve_connector(target)
    except NoMatchingConnector:
        return None, None                  # silent fallthrough!
    return get_or_create_connector_instance(optional_cls), None
```

The `ingested` branch returns explicit `"no_connector"`; the
`typed`/`composite` branch silently returns `(None, None)`.
`_maybe_bind_method` then sees `connector_instance is None`,
returns the handler unbound, dispatch invokes it without `self`,
and the operator sees:

> `connector_error: RuntimeError: typed handler
> 'KubernetesConnector.about' reached dispatch still unbound
> (first parameter 'self'): the dispatcher could not bind it
> to a connector instance ... do not mask it as ha...<truncated>`

The error pleads "do not mask it" because the implementer
correctly anticipated this — but the upstream resolver-miss
ALREADY exists in the `ingested` branch with the right label;
the `typed`/`composite` branch just doesn't mirror it.

**Signal 8 — `AmbiguousConnectorResolution` propagates uncaught.**
The `typed`/`composite` branch catches `NoMatchingConnector` but
NOT `AmbiguousConnectorResolution`. When the resolver finds
multiple candidates and the tie-break ladder can't pick (see T2
of this Initiative for the design-root-cause; happens for any
k8s target with `version=None`),
`AmbiguousConnectorResolution` propagates through the dispatcher,
through `dispatch_typed`, up to FastAPI — bare 500, no JSON body.

But the exception itself, in pod logs, is **the most diagnostic
[...]

Issue #1141, opened 2026-05-26
Task: G0.14-T11 error-message-shape convention + audit pass (docs/codebase artefact; lands first; cited by T1+T5+T7)


damir-topic:
Parent goal: #221
Parent initiative: #1139

## Problem

MEHO's error surfaces are inconsistent in how much they help the
operator recover. The consumer (`claude-rdc-hetzner-dc#697`)
flagged this explicitly: *"one positive pattern worth replicating
everywhere: `/ui/auth/login`'s 503 names the exact missing env
vars + points at the doc to read
(`docs/cross-repo/keycloak-web-client.md`). That's the error
shape operators want everywhere."*

A few surfaces already get it right — and the contrast is sharp:

**Gold-standard (consumer-cited):**
- `/ui/auth/login` 503:
  `"ui_oauth_not_configured: UI_KEYCLOAK_CLIENT_ID /
  UI_KEYCLOAK_CLIENT_SECRET are unset. Render the confidential
  client credentials from Vault per
  docs/cross-repo/keycloak-web-client.md before serving /ui/auth/*."`
- `POST /api/v1/targets/{name}/probe` 501:
  `"no connector registered for product='kubernetes'"`
- `POST /api/v1/connectors/ingest` 422 (mismatched version
  label): rich diagnostic with `values` +
  `remediation` + structured `data` payload with
  `kind=spec_label_mismatch` (consumer's "the shape we want
  more error surfaces to have").
- `AmbiguousConnectorResolution` log message:
  `"resolution ambiguous after tie-break ladder for
  (product='k8s', version=None); candidates=[...];
  set target.preferred_impl_id to one of them"` — but **this
  diagnostic never reaches the operator** because dispatcher
  swallows the exception (signal 8; fixed by [T1](placeholder
  will be linked to T1 once filed)).

**Non-compliant (consumer-cited):**
- `GET /api/v1/health` w/ bad bearer:
  `{"detail":"invalid_token"}` (bare) — partially addressed by
  G0.9.1-T12 #797 + G0.13-T1 #1131.
- `GET /api/v1/feed` on missing redis stream:
  `500 Internal Server Error` with no JSON body (signal 10;
  fixed by #1146).
- Dispatcher composite-handler failures: bare 500 envelope when
  `AmbiguousConnectorResolution` propagates (signal 8; fixed by
  T1).
- `POST /api/v1/agent-principals` 503:
  `{"detail":"keycloak_admin_not_configured"}` — names the
[...]

Issue #797, opened 2026-05-21
Task: G0.9.1-T12 specific token-validator errors at the decode stage (invalid_audience/missing_sub/token_expired/...) instead of opaque invalid_token (Addendum II Ask #1)


damir-topic:
Parent goal: #221
Parent initiative: #772

## Problem

When an inbound JWT fails validation at the decode stage, the backplane returns a single opaque `invalid_token` with no indication of *which* check failed. During the consumer's deployer-onramp walk this cost ~40 minutes of side-by-side claim comparison against a known-good `meho-backplane` token to discover, in sequence, (Wall #2) a missing/mismatched audience and (Wall #3) a missing `sub` claim — each surfaced identically as `invalid_token`. The consumer's #1 ask: *"Specific error messages from the backplane's token validator instead of generic `invalid_token`"*, e.g. `{"error":"invalid_token","detail":"missing required claim: sub"}` or `audience mismatch: expected meho-backplane, got [...]`.

## Root cause

The validator already distinguishes the *post-decode tenant-claim* failures — [_extract_tenant_id](backend/src/meho_backplane/auth/jwt.py) raises `missing_tenant_claim` / `malformed_tenant_claim` and [_extract_tenant_role](backend/src/meho_backplane/auth/jwt.py) raises `missing_tenant_role_claim` / `unknown_tenant_role`, each with a structured structlog event. **That part is good and is the pattern to follow.**

The gap is the *decode stage*: every authlib decode/claims failure (bad signature, `iss` mismatch, `aud` mismatch, expired, **missing `sub`**) collapses into one branch:

```python
except _AUTHLIB_DECODE_ERRORS as exc:
    raise _http_401("invalid_token") from exc
```

(plus the retry path's `_DECODE_ERRORS_WITH_VALUEERROR` → `invalid_token`). Because decode runs *before* the tenant-claim extractors, a token missing `sub` or with a wrong `aud` never reaches the specific tenant errors — the operator only ever sees `invalid_token`. RFC 9068 §2.2.1 makes `sub` REQUIRED on access tokens, so rejecting it is correct; the problem is purely diagnostic opacity.

## Current state

- Decode failures → opaque `invalid_token` ([jwt.py](backend/src/meho_backplane/auth/jwt.py), the `_AUTHLIB_DECODE_ERRORS` / `_DECODE_ERRORS_WITH_VALUEERROR` branches).
[...]

Issue #1148, opened 2026-05-26
Task: G0.14-T7 /ready features block + agent-runtime error-message symmetry + RELEASING.md post-deploy enablement (signals 16, 17; cites T11)


damir-topic:
Parent goal: #221
Parent initiative: #1139

## Problem

v0.6.0 ships **four major features that each require additional
deploy configuration not flagged anywhere operator-facing**
(`claude-rdc-hetzner-dc#697` signal 17 — cross-cutting; signal
16 — the specific case).

| feature | config dep | discovery mode | error shape |
|---|---|---|---|
| Audit replay (G8.2) | `MCP_REQUIRE_SESSION_ID` (or T6 fix lands and decouples it) | trip MCP `initialize` → look for header → grep `audit_log.agent_session_id` for null | (silent — column null) |
| Agent runtime (G11.1) | `KEYCLOAK_ADMIN_URL` + `_CLIENT_ID` + `_CLIENT_SECRET` | trip `POST /api/v1/agent-principals` | 503 `keycloak_admin_not_configured` |
| UI surface (G10) | `UI_KEYCLOAK_CLIENT_ID` + `_CLIENT_SECRET` | trip `GET /ui/auth/login` | 503 with **gold-standard message** |
| Approval queue (G11.2) | (transitive on agent runtime) | trip `POST /api/v1/agents/grants` | not yet exercised |

The pattern across all four: the feature's code ships, env vars
get added in `settings.py`, but no `/ready` check, no
`docs/RELEASING.md` callout, no release-body "what won't work
out of the box" section. Operators discover the dependency by
hitting the surface and reading the 503 — or worse, the silent
NULL column (audit replay before T6's fix).

**Signal 16 (the specific case):** The 503 from
`POST /api/v1/agent-principals` at
[api/v1/agent_principals.py:99](backend/src/meho_backplane/api/v1/agent_principals.py#L99)
says only `{"detail":"keycloak_admin_not_configured"}` — names
the domain but not the env vars, not the doc reference. Compare
with the gold-standard `/ui/auth/login` 503: *"UI_KEYCLOAK_CLIENT_ID
/ UI_KEYCLOAK_CLIENT_SECRET are unset. Render the confidential
client credentials from Vault per `docs/cross-repo/keycloak-web-
client.md` before serving /ui/auth/*."* — same shape opportunity.

## Current state

- `settings.py:767-769` declares the three Keycloak-admin env
  vars; nothing surfaces their state externally.
[...]

PR #1163, opened 2026-05-26
fix(feed): G0.14-T5 surface broadcast outage as T11 SSE feed_error (signal 10)

### LINKED ISSUE #1146: Task: G0.14-T5 /api/v1/feed bare-500 on Redis xread failure — catch RedisError, emit sentinel SSE event or structured 503 (signal 10; cites T11)
Parent goal: #221
Parent initiative: #1139

## Problem

`GET /api/v1/feed` returns `500 Internal Server Error` with no
JSON body when the broadcast Valkey stream key doesn't exist
yet (`claude-rdc-hetzner-dc#697` signal 10). The pod log
carries the actual failure:

```
File ".../meho_backplane/api/v1/feed.py:380", in _feed_generator
    entries = await client.xread(...)
File ".../redis/asyncio/client.py:781", in execute_command
    ...
```

The endpoint is supposed to be an SSE-style broadcast stream.
On a fresh deploy (or after a broadcast reset), the stream key
doesn't exist; the `xread` call propagates uncaught.

Same anti-pattern as the dispatcher's
`AmbiguousConnectorResolution` propagation (signal 8, fixed by
T1 of this Initiative): a clean, recoverable, well-classified
error reaches the FastAPI default handler and becomes a bare
500.

## Current state

[`_feed_generator` at api/v1/feed.py:344-472](backend/src/meho_backplane/api/v1/feed.py#L344):

```python
async def _feed_generator(...):
    ...
    while True:
        entries = await client.xread(...)
        # no try/except around xread
```

The `xread` is awaited bare. Any redis-side exception
propagates out of the generator, kills the SSE stream
mid-handshake, and FastAPI's default handler emits the bare
500.

The redis client's typical exception classes for this case:
`redis.exceptions.ResponseError` ("no such key"-like) or
`RedisError` / `ConnectionError` for transport failures. The
"stream doesn't exist yet" case specifically returns an empty
result in some redis versions and raises in others — depends
on the redis/Valkey version + the XREAD options.

## Desired state

**Catch the `RedisError` family inside `_feed_generator`.**
Two reasonable shapes — pick one or do both:

**Option A — sentinel SSE event.** Yield a structured
`feed_initialized` SSE event when the stream doesn't exist yet:

```
event: feed_status
data: {"status":"awaiting_events","stream":"meho:broadcast"}
```

…and continue the long-poll. The client sees a heartbeat-shaped
event meaning "feed is healthy, no data yet." On real transport
failures (connection refused), close the SSE stream cleanly
with a structured error event.

**Option B — structured 4xx on the initial response.** If the
client should know upfront that the feed is empty (e.g. no
broadcast subsystem at all in this deploy), return a structured
`503 {"detail": "broadcast_subsystem_unavailable", ...}` per
the convention codified by T11 of this Initiative.

A is closer to SSE semantics (the stream is the contract; events
flow). B is closer to "this feature isn't configured" (more like
the gated-feature surfaces in T7). The consumer's signal frames
this as a fresh-deploy / empty-stream case, which is more A —
but on a genuinely-broken redis connection, B is the right
response.

Compliant per T11: the response (whichever shape) names the
diagnostic value, the cause, and the recovery action.

## Acceptance criteria

- [ ] `_feed_generator` catches the redis exception family;
  empty-stream cases produce either a sentinel SSE event
  (Option A) or a structured 503 (Option B) — picked +
  documented in PR.
- [ ] Real-transport-failure cases (connection refused, host
  unreachable) close the SSE stream cleanly with a structured
  error event, not a bare 500.
- [ ] Test covers the empty-stream case against a fresh test
  fixture: GET /api/v1/feed returns the chosen-shape response
  (not a 500).
- [ ] The error / sentinel response complies with the
  convention codified by T11 (`docs/codebase/error-message-
  shape.md`); cite the doc in the Task's PR.
- [ ] `Python (ruff + mypy + pytest)` green.

## Out of scope

- Re-engineering the broadcast subsystem's SSE protocol —
  the SSE shape is settled.
- Persistent connection retry policy across XREAD failures —
  the existing while-True loop's retry/backoff is fine; this
  Task only fixes the bare-500 anti-pattern at the boundary.
- The broader bare-500 anti-pattern at dispatcher (T1) — same
  shape, different surface.

## References

- Consumer feedback: `claude-rdc-hetzner-dc#697` signal 10
  (`api-v1-feed-bare-500-on-redis-xread-failure`).
- Convention anchor: **T11 of this Initiative**.
- Sister anti-pattern Task: **T1 of this Initiative**
  (dispatcher bare-500s).
- Grounding:
  [api/v1/feed.py:344-472](backend/src/meho_backplane/api/v1/feed.py#L344).
- Best-practices: **python_best_practices** pack (catch
  expected-failure exceptions at the boundary; bare 500s are
  always a defect).

damir-topic:
## Summary

- `/api/v1/feed` no longer drops to a bare HTTP 500 when the broadcast subsystem is unreachable. `_feed_generator` now catches `redis.exceptions.RedisError` (the parent class of `ConnectionError`, `TimeoutError`, `ResponseError`) around the `XREAD` call, emits a single T11-compliant `event: feed_error` frame, and breaks the loop so the SSE stream closes cleanly. Signal 10 of `claude-rdc-hetzner-dc#697`.
- The error frame carries the convention from `docs/codebase/error-message-shape.md`: a stable `snake_case` code (`broadcast_subsystem_unavailable`), a human-readable message naming the affected stream key + the redis-py exception class, and a `doc` field pointing at the convention doc for the remediation procedure.
- Empty-stream behavior is unchanged on purpose — redis-py returns `None` (not an exception) for a missing stream key, which already falls through to the existing heartbeat path. The fix scope is the genuinely-broken case (broadcast pod down, network partition, transport timeout), which is what the consumer signal flagged.

## Design choice (Option A vs B from the issue body)

The issue body offered two shapes — Option A (sentinel SSE event, continue long-poll) and Option B (structured 4xx on the initial response). We landed on a third position the body anticipated as "doing both is also valid": an **SSE error event then clean close**, T11-compliant in content (B's shape) but emitted in the SSE channel (A's transport). The trade-off:

- **Option B is impossible mid-stream.** Once `_feed_generator` yields its first frame (event, heartbeat, or now an error frame), Starlette has already sent `http.response.start` and FastAPI cannot retroactively swap to a `503` body. The pre-fix bare 500 was the same shape — by the time `xread` raised, the response was already committed.
- **A's "continue long-poll on every error"** would keep an SSE connection open against a known-failing broadcast subsystem until the client gave up — a busy retry loop the client side cannot suppress because the SSE spec auto-reconnects on connection drop, not on error event.
- **The hybrid this PR ships** — emit one structured error frame, then break — gives the consumer the diagnostic value B wanted (operator-actionable code + message + doc reference) without the tight-loop A would produce. The browser `EventSource` reconnect machinery sees a clean close and decides whether to retry, the same way it would for any other natural stream end.

The empty-stream case (Valkey reachable, stream key has no entries yet) was misdiagnosed in the issue body as a `RedisError`-emitting path. redis-py 7.4 + Valkey return `None` from `XREAD` against a missing key — a normal empty result, not an exception. The existing `if entries` branch (now `_consume_xread_batch`'s `if not entries: return cursor, []`) already handles that with the heartbeat-keepalive path; no separate sentinel-event handling is needed.

## Convention compliance

Cites `docs/codebase/error-message-shape.md` (G0.14-T11 #1141, merged on `main` at `d7e9ac8`):

- **Code** (`broadcast_subsystem_unavailable`) — stable `snake_case`, domain-prefixed; added to the catalogue conceptually (the doc's audit table tracks this surface as `T5 #1146` and will move out of "non-compliant" once this PR lands and a follow-up audit sweep updates the row).
- **Message** — three clauses per the convention: diagnostic value (`stream='meho:feed:<tenant_id>'`, redis-py exception class) → remediation pointer (the doc reference) → doc path (`docs/codebase/error-message-shape.md`).
- **Doc** — relative path the operator can `cat` from their checked-out clone, per the convention's "operators new to MEHO need it" rule.
- **Info-leak boundary** — the redis-py transport-level message string is **not** echoed; only the exception class name lands in the response. The `_log.warning(... exc_info=True)` carries the full traceback to structlog so the operator with kubectl access has the underlying detail. Mirrors the G0.9.1-T12 #797 precedent (codes in body, values in log).

## Refactor

Extracted two helpers out of `_feed_generator` to stay under the code-quality function-size ceiling:

- `_format_feed_error(stream_key, exception_type)` — formats the T11-compliant SSE error frame.
- `_log_and_format_broadcast_unavailable(operator, stream_key, exc)` — pairs the structured log call with the frame formatting (they're always invoked together at the catch site).
- `_consume_xread_batch(entries, ...)` — replaces the inline `if entries:` block; returns `(new_cursor, [frames_to_yield])`. Same cursor-advance-past-filtered-batch semantics the iter-2 B1 fix added — the test `test_cursor_advances_past_filtered_entries` still pins the behavior.

## Test plan

- [x] `uv run ruff check src/meho_backplane/api/v1/feed.py tests/test_api_v1_feed.py` — clean
- [x] `uv run ruff format --check ...` — clean
- [x] `uv run mypy src/meho_backplane/api/v1/feed.py --ignore-missing-imports` — clean
- [x] `uv run pytest tests/test_api_v1_feed.py -x -q` — 20 passed, 1 skipped (Docker integration)
- [x] `python3 .claude/hooks/code-quality.py --diff` — 0 blockers, 2 pre-existing warnings (file size 646>600, function size 100>60; both pre-existing or unavoidable given the docstring-rich generator)
- [x] **AC1** (`_feed_generator` catches RedisError family; pick documented in PR) — covered by `test_xread_connection_error_emits_feed_error_then_closes`, `test_xread_response_error_emits_feed_error_then_closes`. Picked shape documented above.
- [x] **AC2** (real-transport-failure → structured error event, not bare 500) — covered by the three new tests; all assert the frame shape matches `event: feed_error\ndata: {"code":"broadcast_subsystem_unavailable", ...}` and the loop breaks (generator exhausts naturally, no further frames).
- [x] **AC3** (fresh-test-fixture: GET /api/v1/feed returns chosen-shape, not 500) — covered at the generator layer (the same layer the existing HTTP-edge tests delegate to for body shape, per the test-architecture comment in `test_api_v1_feed.py`); the HTTP edge wrapper is unchanged so the 200 + `text/event-stream` happy path still applies.
- [x] **AC4** (T11 compliance + doc cite) — message includes `meho:feed:<tenant_id>`, redis-py exception class, doc path; PR body cites the doc.
- [x] **AC5** (`Python (ruff + mypy + pytest)` green) — confirmed above.

## Out of scope (recorded as adjacent findings)

- `backend/src/meho_backplane/mcp/tools/broadcast.py:_xread_or_cancelled` has the same unguarded-xread pattern (it catches `asyncio.CancelledError` but not `RedisError`). Different surface (MCP JSON-RPC, not SSE), different remediation shape (JSON-RPC error with `data.code`), out of scope for T5 which is explicitly scoped to `/api/v1/feed` per the issue body's "Out of scope" section. Worth a follow-up.
- `backend/src/meho_backplane/ui/routes/broadcast/stream.py` has the same SSE pattern (it's the UI surface counterpart of `/api/v1/feed`). Same fix shape would apply; same out-of-scope reasoning.
- The audit table at `docs/codebase/error-message-shape.md` still lists `GET /api/v1/feed` 500 as "non-compliant (tracked by T5 #1146)". Updating that row to "compliant" is the audit-sweep follow-up that lands after all of T1/T5/T7 have merged — the precedent set by T1 #1142 which similarly did not touch the audit table inline.

Closes #1146


damir-topic:
@coderabbitai review

damir-topic (commented):




**Anchor task:** #1146 — G0.14-T5 `/api/v1/feed` bare-500 on Redis xread failure
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `ca78de9`
**AI reviewer (CodeRabbit):** rate-limited (review-limit reached; "more reviews available in 41 minutes"); polled ~12.5 min within the 15-min window before stopping early. Recorded as not received this iteration.

### Acceptance criteria
- [x] **AC1** — `_feed_generator` catches `RedisError` family around the `XREAD` call (`backend/src/meho_backplane/api/v1/feed.py:549`). Pick documented in the PR body (hybrid: SSE error frame + clean close, neither pure A nor pure B).
- [x] **AC2** — Real transport failure emits a structured `event: feed_error` frame then closes; verified by all three new tests.
- [x] **AC3** — Generator-layer tests cover the chosen-shape response (delegation to the existing test-architecture pattern in `tests/test_api_v1_feed.py` where HTTP-edge tests delegate to generator-layer tests for body shape).
- [x] **AC4** — T11 compliance verified against `docs/codebase/error-message-shape.md`: `code='broadcast_subsystem_unavailable'` (snake_case, domain-prefixed), `message` names stream key + redis-py exception class + doc reference, `doc` field carries relative path. PR body cites the doc.
- [x] **AC5** — `Python (ruff + mypy + pytest)` green on CI; locally `4786 passed, 185 skipped` across the full test suite.

### Project-board alignment
|---|---|---|---|---|---|
| #1146 | In Progress | Closes | yes | full | close on merge |
| #1139 (parent) | Todo | (parent) | — | (partial; sibling Tasks still open) | — |

No mismatches.

### Predicted vs. actual diff
- ✅ `backend/src/meho_backplane/api/v1/feed.py` — `_feed_generator` gains `try/except RedisError` around `xread`, three helpers extracted for function-size compliance.
- ✅ `backend/tests/test_api_v1_feed.py` — three new generator-layer tests covering `ConnectionError`, `ResponseError`, mid-stream `TimeoutError`.
- ✅ `CHANGELOG.md` — Unreleased / Fixed entry.

No undisclosed changes; no predicted-but-missing files.

### Integration verification
- [x] `ruff check` on changed files — clean
- [x] `mypy` on `feed.py` — clean (1 file); also clean across full backend (330 files)
- [x] `pytest tests/test_api_v1_feed.py -x -q` — 20 passed, 1 skipped (Docker integration; matches PR body)
- [x] `pytest tests/ -q --ignore=tests/integration` — 4786 passed, 185 skipped, 0 failed
- [x] CI status checks all green (`Python (ruff + mypy + pytest)`, `Go`, `Helm`, `CodeQL`, `Semgrep SAST`, `Secret Scan`, `image`, `DCO`)
- [⚠️ not run] Live SSE end-to-end against a real broadcast pod — out of scope for unit test; covered conceptually by the three new tests' direct exercise of `_feed_generator`.

### Findings (inline)
None.

- The same unguarded `xread` pattern exists at `backend/src/meho_backplane/mcp/tools/broadcast.py:563` (`_xread_or_cancelled`) and `backend/src/meho_backplane/ui/routes/broadcast/stream.py:138`. Both are explicitly out of scope for T5 per the issue body (which is scoped to `/api/v1/feed`). Worth a follow-up sweep alongside T7 #1148.
- The audit table at `docs/codebase/error-message-shape.md:240` still tracks this surface as "T5 #1146" — the implementer's recorded plan (audit-sweep follow-up that runs after T1/T5/T7 land) matches T1 #1142's precedent (also didn't flip the audit row inline). Confirmed against the convention doc's own guidance (lines 247-248: "Adding new error surfaces against the convention extends this table" — T5 fixes an existing surface).

- Adopted: 0 (no review produced — rate-limited; "more reviews available in 41 minutes" exceeded the 15-min wait window).
- Disputed: 0
- Out of scope: 0

### Docs
- [x] `docs/codebase/error-message-shape.md` — verified unchanged (no edit needed in this PR; audit-table row flip deferred to a sweep per the convention doc's own pattern and T1 precedent).

- The hybrid design choice is well-defended in the PR body: pure Option B (HTTP 503) is architecturally infeasible mid-stream (`http.response.start` already locked once `_feed_generator` yields), pure Option A (continue long-poll with heartbeats) would auto-reconnect against a known-failing backend, and the chosen "yield structured error frame + break" gives client-side JS the signal it needs to call `eventSource.close()` rather than relying on bare-drop detection.
- Empty-stream handling correctly identified — redis-py returns `None` (not an exception) for a missing stream key per [Redis XREAD docs](https://redis.io/commands/xread/) and [Valkey XREAD docs](https://valkey.io/commands/xread/), so the existing heartbeat fall-through path is sufficient; no separate sentinel-event handling was needed.
- T11 compliance is exact, not approximate: stable snake_case code, component-naming message with doc reference inline, separate `doc` field, info-leak boundary respected (the redis-py transport-level message string is not echoed in the body; only the exception class name lands, with the full traceback going to structlog `exc_info=True`).
- Three tests, three documented `RedisError` subclasses (`ConnectionError`, `ResponseError`, `TimeoutError`), one mid-stream-after-prior-events case — covers the bug classes from the issue body and verifies the info-leak boundary (the `WRONGTYPE` raw message is asserted NOT to appear in the response body).
- Function-size compliance achieved via three well-scoped helpers (`_format_feed_error`, `_consume_xread_batch`, `_log_and_format_broadcast_unavailable`) without losing readability; the cursor-advance-past-filtered-batch invariant from a prior iteration's B1 fix is preserved.

Issue #1144, opened 2026-05-26
Task: G0.14-T3 TargetCreate.product enum at boot + discoverable 422 — single typo at target creation should not silently break (signal 5)


damir-topic:
Parent goal: #221
Parent initiative: #1139

## Problem

`TargetCreate.product` at
[targets/schemas.py:134](backend/src/meho_backplane/targets/schemas.py#L134)
is a free-form string:

```python
product: str = Field(min_length=1, max_length=100)
```

But the resolver matches `target.product` against the connector
class's exact `product` token at registration time — `"k8s"`
not `"kubernetes"`, `"vmware"` not `"vsphere"`, etc. The valid
set is the union of `product` fields across registered connector
classes; nothing operator-facing enumerates it.

The consumer's `claude-rdc-hetzner-dc#697` walked this exact
trap:

> *"We made the mistake (registered `product='kubernetes'`,
> hit ambiguous downstream error). The discovery path is: hit
> `/api/v1/connectors`, read the `product` field of the
> connector you want, mirror it on the target."*

This is the first sharp edge an operator hits — and combined
with signal 6 (no DELETE / no PATCH on `product` — T4 of this
Initiative), a single typo at target creation creates a
permanent broken row.

## Current state

- `product: str = Field(min_length=1, max_length=100)`. No
  enum, no example, no description pointing at the source of
  valid values.
- The valid set is fully determined at app boot: the
  registered connector classes' `product` fields are knowable
  before the first request.
- `/api/v1/connectors` lists registered connectors with their
  `product` field, but the operator has to know to look there
  first.
- The probe endpoint (`POST /api/v1/targets/{name}/probe`)
  returns a great diagnostic for an unknown product:
  `501 {"detail":"no connector registered for product='kubernetes'"}`
  — but that's after the operator already created the target.

## Desired state

**Pick one (Option A is cleanest):**

- **Option A:** Generate the `product` enum at app boot from
  the registered connector classes. Expose as a JSON Schema
  enum in `TargetCreate.product`. Operators get autocomplete
  in OpenAPI tooling / Swagger; validation rejects unknown
  products with a structured 422 listing the valid set.
- **Option B:** Add `example` + `description` to the field
  pointing at `/api/v1/connectors` as the source of valid
  values, plus a one-line warning that the resolver matches
  exactly. Lower-effort; less ergonomic.
- **Option C:** When POSTing a target with a product no
  connector advertises, return a structured 422 with
[...]

PR #1162, opened 2026-05-26
feat(conventions): G0.13-T7 supersede rdc-internal seed with generic default tenant

damir-topic:
## Summary

- Replace the public OSS surface's baked-in `rdc-internal` tenant seed (migration `0018`, 8 conventions extracted verbatim from one consumer's `CLAUDE.md`) with a generic `default` tenant + 2 illustrative conventions. Closes the operational impact from v0.6.0 dogfood signal 12 — every adopting customer's MCP `initialize.instructions` field had been receiving the original consumer's operational discipline + repo references as their session-start instructions.
- Ship migration `0025_supersede_rdc_internal_seed.py` on top of `0018` (Alembic is forward-only). `upgrade()` cleans up the rows `0018` authored (narrowed on the seed `actor_sub` so operator edits under seeded slugs survive), upserts the `default` tenant, and seeds 2 short conventions that read as documentation about the convention feature itself rather than as someone's operational rules. `downgrade()` removes the default seed but does NOT restore `rdc-internal` (restoring consumer content would re-leak).
- The consumer-side template for re-applying the original 8 `rdc-internal`-specific conventions lives in `docs/architecture/conventions-seed.md`; the consumer applies it post-deploy in their own repo (`claude-rdc-hetzner-dc`) on their next deploy cycle — not part of this PR.

## Test plan

- [x] `ruff check` on touched files — clean
- [x] `ruff format --check` — clean
- [x] `mypy --ignore-missing-imports` on new `0025` + touched `conventions/__init__.py` — clean
- [x] `pytest backend/tests/test_alembic_seed_0025_supersede.py backend/tests/test_alembic_seed_rdc_conventions.py` — **14 passed** in 2.36s
- [x] `pytest backend/tests/test_db_agent_run.py backend/tests/test_db_models.py backend/tests/test_db_scheduled_trigger.py backend/tests/test_topology_bulk_import.py backend/tests/test_mcp_initialize_instructions.py` — **50 passed** in 9.42s
- [x] `code-quality.py --diff` — clean
- [x] AC: head replaces rdc-internal with default (`test_head_replaces_rdc_internal_with_default`)
- [x] AC: migration reversible (`test_downgrade_removes_default_seed_keeps_tenant_no_rdc_internal_restore`)
- [x] AC: tests cover both directions — 6 SQLite cases + 1 PG testcontainers idempotency replay case
[...]

Issue #1139, opened 2026-05-26
Initiative: G0.14 v0.6.0 post-validate extended dogfood hardening — 11 tasks from typed-connector + breadth + agent-runtime + vmware-rest exercises


damir-topic:
Parent goal: #221

## Why now

`claude-rdc-hetzner-dc#697` continued past the closed-loop
validate phase (which produced [G0.13 #1130](https://github.com/evoila/meho/issues/1130))
with four follow-up exercises against the same live v0.6.0
deploy:

1. **Typed-connector dispatch exercise** — first-ever live
   exercise of the G3.9/G3.10 wave against the rke2-infra cluster.
   The G3.9/G3.10 kubeconfig loader path is now **proven live**
   (`k8s.about` + `k8s.node.list` returned real cluster data via
   the operator-context Vault → kubeconfig → KubernetesConnector
   chain).
2. **Breadth pass across the unexplored v0.6.0 surfaces** —
   read-only sweep of ~50-60 of the 86 OpenAPI paths.
3. **Agent-runtime probe + convention write loop.**
4. **vmware-rest BasicCredentials parallel** — the second
   G3.9/G3.10 auth shape (Vault → vCenter `POST /api/session`),
   reached the composite handler successfully proving the loader
   path; downstream sub-op failed (signal 20).

These four exercises surface **16 new signals** (#5 through #20
in the consumer's signal directory) that the validate phase
didn't trip. Same release, same deploy, deeper exercise — a
distinct triage cycle.

The G0.9 → G0.9.1 precedent (#737 → #772) is exactly this shape:
a second-pass dogfood against the already-shipped release earns
its own Initiative rather than a re-open of the first one.

### Convention task (lands first; cited by T1, T5, T7)

- **T11** `error-message-shape convention + audit pass` — codify
  the gold-standard error pattern the consumer flagged
  (`/ui/auth/login` 503: *names exact missing env vars + points
  at the doc to read*) as a `docs/codebase/` artefact; audit the
  catalogued error surfaces (positives: `/probe`,
  `/retrieve/usage`, `/audit/who-touched`, the
  `AmbiguousConnectorResolution` log message; non-compliant:
  signal 8 the bare-500 envelope, signal 10 the feed bare-500,
  signal 16 the `keycloak_admin_not_configured` message, signal
  11's silent dependency). Per-surface fixes stay in T1 / T5 /
  T7; this Task lands the doc + audit, the others cite it. **MED
  / lands first.**

[...]

Issue #221, opened 2026-05-11
Goal: G0 Foundational substrate — tenant model + connector base + targets-as-data + retrieval substrate; the hidden Goal #214-#220 silently assume


damir-topic:
## Goal

Land the foundational backplane substrate every other v0.2 Goal silently assumes but the v0.1 chassis (Goal #11 in `evoila-bosnia/meho-internal`) does not provide: a tenant model, a connector base contract, the targets-as-data table, and the retrieval substrate. Without G0, every downstream Goal either silently leaks cross-tenant data, ships connectors as one-off bespoke modules, or has to retrofit tenancy in a breaking schema migration mid-rollout.

This is the **hidden Goal** the customer-requirements pass surfaced — none of #214–#220 name it, but each transitively depends on it.

## Why

Three structural gaps in the v0.1 chassis are load-bearing for v0.2:

1. **No tenancy.** The JWT validator never extracts a `tenant_id` claim; `audit_log` has no `tenant_id` column; there is no row-level scoping anywhere. Every G3–G9 Goal is tenant-scoped (`rdc-internal` first, `customer-a` next). Shipping any of them without tenancy makes them single-tenant-forever or a future breaking migration.

2. **No connector base.** G3 (#214) lands ~14 connectors. The chassis has Vault + Keycloak as vendor-specific modules with no shared `Connector` ABC, no fingerprint contract, no per-target identity-model field, no registry. Without a base contract, tier-1 (vSphere + Vault + bind9) ships as three one-offs and tier-2 onward becomes a refactor.

3. **No targets-as-data table.** G9 (#220) is the named owner, but every G3 connector op needs `--target <name>` resolution at the backplane on day 1. Today the consumer's `targets.yaml` is a file-on-disk resolved client-side. The *targets* half of G9 is actually a G3 prerequisite — the *topology graph* half stays G9.

Plus the retrieval substrate (pgvector + fastembed per v0.1-spec L382–401) is shared by G4 (#215, knowledge) and G5 (#216, memory). Building it once under G0.4 is the only way to avoid two parallel embedding pipelines.

## Done-when (testable from the consumer)

- JWT carries a `tenant_id` claim; `verify_jwt_and_bind` extracts it; `audit_log` rows include it; every authenticated handler scopes its queries by tenant.
- Backplane defines a `Connector` ABC + registry; the existing Vault integration has been re-shaped to fit (proving the abstraction).
[...]

PR #1164, opened 2026-05-26
feat(targets): G0.14-T4 DELETE /api/v1/targets/{name} + allow product in TargetUpdate

### LINKED ISSUE #1145: Task: G0.14-T4 DELETE /api/v1/targets/{name} + allow product in TargetUpdate — recover from product typo + close credential-hygiene gap (signal 6)
Parent goal: #221
Parent initiative: #1139

## Problem

`/api/v1/targets/{name}` has a missing-API problem in two
dimensions:

1. **No `DELETE` route.** Available verbs: `GET` (list), `GET`
   (show), `POST` (create), `PATCH` (update), `POST /probe`,
   `POST /refresh` (G9.1). No DELETE.
2. **`TargetUpdate` excludes `product`.** Per
   [targets/schemas.py:146-158](backend/src/meho_backplane/targets/schemas.py#L146)
   only certain fields are patchable; `product` is not among
   them. So PATCH can't recover from a `product` typo.

The combined effect (per `claude-rdc-hetzner-dc#697` signal 6):
**a single typo at target creation creates a permanent broken
row.** The consumer's rdc-internal tenant now has one
permanent `rke2-infra` row (product=kubernetes, broken) sitting
alongside the working `rke2-infra-k8s` row, taking up name +
alias slots.

Compound risk: targets hold credential references
(`secret_ref` pointing at Vault paths). Undeletable target rows
mean undeletable credential references over time — a
credential-hygiene concern that grows linearly with the
operator's mistake-rate.

The consumer also flags the asymmetric precedent — the
sibling `/api/v1/conventions` route family ships full CRUD +
history (`POST`, `GET`, `PATCH`, `DELETE` per slug, `GET
/history`). Targets deserve the same housekeeping shape.

## Current state

- `/api/v1/targets/{name}`: GET, POST, PATCH, `/probe`, `/refresh`.
  No DELETE. `TargetUpdate` excludes `product`.
- `/api/v1/conventions/{slug}`: full CRUD + `/history`. The
  reference shape.

## Desired state

**Pick one or both (the consumer recommended either; both is
strongest):**

**Part A — `DELETE /api/v1/targets/{name}`.**
Add a soft-delete (`deleted_at` timestamp; row stays for audit
purposes but is excluded from GET list / GET show / resolver
lookups). RBAC-gated (`tenant_admin` at minimum). Cascade-check
against active audit rows + topology graph rows — if the target
is referenced anywhere durable, the DELETE response includes
the references and the operator can confirm with a `?force=true`
query parameter (or the DELETE fails 409 with the reference
list — implementer's call between soft-delete-with-warnings
vs hard-fail-on-references).

**Part B — allow `product` in `TargetUpdate`.**
Add `product: str | None` to `TargetUpdate`. Gate the update
on resolver-validates-against-known-products (same enum / 422
shape T3 uses) so a PATCH-time typo doesn't silently break a
working target. After the PATCH, optionally re-run the probe
to assert the target now resolves correctly.

Together: an operator who misregisters `product=kubernetes`
can PATCH it to `product=k8s` (Part B) OR DELETE the broken
row and create a fresh one (Part A). Either recovery path works;
both available is strongest.

## Acceptance criteria

- [ ] **Part A:** `DELETE /api/v1/targets/{name}` returns 204
  on success; cascade-checks against `audit_log` (target_name
  references) and `graph_node`/`graph_edge` references; either
  soft-deletes (with `deleted_at`) or hard-fails 409 with the
  references — picked + documented in the PR.
- [ ] **Part A:** RBAC-gated to `tenant_admin` at minimum;
  cross-tenant DELETE returns the standard 404 conflation.
- [ ] **Part A:** Audit row written (`meho.targets.delete` or
  similar; consistent with the
  `meho.connector.edit_*` audit precedent).
- [ ] **Part B:** `TargetUpdate.product` accepts the same
  product-token enum T3 enforces; rejects unknown products
  with the structured 422 from T3.
- [ ] **Part B:** Optional: PATCH response includes a
  `probe_result` field if the new product was resolver-valid
  (so the operator sees the working state immediately).
- [ ] Tests cover: DELETE happy path, DELETE with cascade,
  PATCH product typo rejection, PATCH product success +
  resolver re-validation.
- [ ] `Python (ruff + mypy + pytest)` green.

## Out of scope

- Hard-deleting audit rows that reference a deleted target —
  audit is append-only per v0.1-spec §6; not in scope.
- Renaming the target row at delete — separate concern from
  cleanup.
- Cascading delete of `secret_ref` paths in Vault — out of
  scope (operator can `vault kv destroy` separately if desired;
  Vault paths aren't owned by this database).
- A full `/history` view for targets (the
  `/api/v1/conventions` parity goal would imply this) —
  separate Task if anyone wants it; not v0.6.1 priority.

## References

- Consumer feedback: `claude-rdc-hetzner-dc#697` signal 6
  (`target-product-immutable-no-delete`) + the consumer's
  v0.6.0 update positive callout about
  `/api/v1/conventions` shipping full CRUD + history as the
  reference shape.
- Sibling lifecycle Task: **T3 of this Initiative**
  (`product` enum + structured 422) — Part B reuses the
  validator T3 ships.
- Audit precedent: existing `meho.connector.edit_*` audit
  rows at
  [connectors/ingest/...](backend/src/meho_backplane/api/v1/connectors_ingest.py).
- Grounding:
  [targets/schemas.py:146-158](backend/src/meho_backplane/targets/schemas.py#L146),
  [api/v1/targets.py](backend/src/meho_backplane/api/v1/targets.py).
- Best-practices: **python_best_practices** pack (CRUD
  surfaces should be symmetric across entities of the same
  housekeeping class) + **devops_best_practices** pack
  (credential references must be removable / auditable).

damir-topic:
## Summary

- Closes the "misregistered target cannot be recovered" gap from
  signal 6 of `claude-rdc-hetzner-dc#697`: a single typo at target
  creation previously created a permanent broken row because there
  was no DELETE route and `TargetUpdate` excluded `product`.
- Adds `DELETE /api/v1/targets/{name}` (tenant_admin) — soft-delete
  by stamping `deleted_at`; every read path filters
  `deleted_at IS NULL`; cascade-check on `graph_node.target_id`
  references defaults to 409 + a `?force=true` hint when the target
  is wired into the topology graph.
- Allows `product` in `TargetUpdate` — operator can correct
  `product='kubernetes'` → `'k8s'` in-place; an unknown product
  yields a structured 422 mirroring the `/probe` 501 shape.

## Design choices

**Soft-delete vs hard-delete.** Chose soft-delete (`deleted_at`
timestamp). Hard-delete would orphan the `audit_log.target_id`
soft-FK (audit is append-only per v0.1-spec §6, so referenced ids
must keep pointing at *something*). Re-DELETEs naturally collapse
to 404 because the resolver filters `deleted_at IS NULL` once the
first DELETE flushes.

**Cascade strategy.** The DELETE counts `graph_node.target_id`
references only — not `audit_log.target_id` references. Audit
references would exist for *every authenticated request* against
the target and would always block the delete, defeating the
recovery purpose. Graph references are operational topology that
the operator may want to preserve, so the default is to fail-loud
with 409 (`kind=target_has_references`, count, message naming
`?force=true`); `?force=true` proceeds with the soft-delete anyway
(the `graph_node.target_id` FK is `ON DELETE SET NULL` so the
graph rows survive safely).

**Product validation strategy (T3 #1144 coordination).** T3 ships
the same enum at POST time, but T3 hasn't merged at the time of
writing. Per the coordination instructions on this issue, the
product-validation helper (`_registered_products()`) is inline in
`api/v1/targets.py` and computes the product set from
`all_connectors_v2().keys()`. A `# consolidate with #1144 once
both land` note in the docstring is the cleanup marker. The
validation only fires when `product` is set to a *different* value
than the existing row's product — a PATCH that re-asserts the
current product passes without registry lookup so the route is
robust to a mid-rolling-restart unregistered state.

**Error shape compliance.** Both new error sites comply with the
T11 #1141 convention codified in `docs/codebase/error-message-shape.md`:
structured `detail` with `kind`, machine-actionable fields, and a
`message` naming the offending value + the remediation + a doc
reference. The convention's audit table is extended with the two
new sites.

## Test plan

- [x] `uv run ruff format --check src/ tests/` — clean (700 files
  already formatted)
- [x] `uv run mypy src/` — Success: no issues found in 330 source
  files (strict mode)
- [x] `pytest tests/test_api_v1_targets.py` — 38 passed (10 new
  DELETE + PATCH product tests added)
- [x] `pytest tests/test_targets_schemas.py` — 23 passed (4 new
  schema tests added)
- [x] `pytest tests/test_targets_resolver.py` — 12 passed, 1
  skipped (3 new soft-delete-filter tests added)
- [x] `pytest tests/test_db_targets.py` — 7 passed (1 new
  `deleted_at` round-trip test added)
- [x] `pytest tests/test_targets_audit_integration.py` — 11
  passed (existing tests still green; audit pipeline unaffected)
- [x] `pytest tests/test_mcp_tools_topology.py` — passed (1 new
  MCP soft-delete-filter test added)
- [x] Targeted sweep: `pytest tests/test_api_v1_targets.py
  tests/test_targets_schemas.py tests/test_targets_resolver.py
  tests/test_db_targets.py tests/test_targets_audit_integration.py
  tests/test_targets_fingerprint.py
  tests/test_api_v1_targets_discover.py
  tests/test_api_v1_targets_import.py
  tests/test_mcp_tools_topology.py
  tests/test_operations_dispatcher.py` — **177 passed, 1 skipped**
- [x] Migration `0028` reversible: `alembic upgrade head` →
  `alembic downgrade -1` → `alembic upgrade head` clean on SQLite
- [x] OpenAPI snapshot regenerated (`make snapshot-openapi`); DELETE
  surfaces under `delete_target_api_v1_targets__name__delete`
- [x] `code-quality.py --diff`: 0 blockers, 13 warnings
  (pre-existing function-size warnings on adjacent code; no new
  blockers)

### Acceptance-criteria mapping

| Criterion | Evidence |
|---|---|
| Part A: DELETE returns 204 on success; cascade-checks audit + graph references | `test_delete_target_soft_deletes_and_returns_204` — passed. Soft-delete chosen (documented above); audit references intentionally excluded |
| Part A: cascade-check + `?force=true` semantics | `test_delete_target_with_graph_node_refs_returns_409` (409 + count + message) + `test_delete_target_with_graph_node_refs_force_true_succeeds` (force=true proceeds) — both passed |
| Part A: RBAC-gated to tenant_admin; cross-tenant DELETE returns 404 | `test_delete_target_operator_role_returns_403` + `test_delete_target_cross_tenant_returns_404` — both passed |
| Part A: Audit row written (`meho.targets.delete` or similar) | `audit_op_id='targets.delete'` bound on entry; mirrors the existing `targets.discover` precedent and the `meho.connector.edit_*` pattern |
| Part B: `TargetUpdate.product` accepts product-token enum; rejects unknown with structured 422 | `test_patch_product_unknown_returns_422` (422 + `kind`/`product`/`valid_products`/`message`) + `test_patch_product_valid_succeeds` — both passed |
| Part B: Optional probe_result in PATCH response | Not implemented — out of scope; operator runs `/probe` after PATCH if desired. The probe is idempotent + cheap |
| Tests cover DELETE happy path, cascade, PATCH product typo, PATCH product success | 10 new tests cover all four paths (see test plan) |
| `Python (ruff + mypy + pytest)` green | All three gates clean (see test plan) |

## Out of scope

- Sibling Task **T3 #1144** ships the same product-validation
  enum at POST time. The PATCH-time validator in this PR mirrors
  the same logic inline; once #1144 lands the two sites
  consolidate against a shared helper (note in the code).
- Full `/history` view for targets (parity with
  `/api/v1/conventions`) — out of scope per task body; separate
  task if anyone wants it.
- Hard-deleting `audit_log` rows that reference a deleted target —
  audit is append-only per v0.1-spec §6.
- Cascading delete of `secret_ref` paths in Vault — Vault paths
  aren't owned by this database (operator runs `vault kv destroy`
  separately if needed).
- Renaming the target row at delete — separate concern from
  cleanup.

## Adjacent findings

- `_handler_resolve`'s `get_or_create_connector_instance` is
  imported in `targets.py` but only used by `discover_targets`.
  Pre-existing; left as-is.
- Pre-existing ruff `RUF034` on
  `backend/alembic/versions/0023_create_approval_request.py` is
  not caught by CI (CI runs ruff on `src/` + `tests/` only, not
  `alembic/`). Unrelated to this PR.

Closes #1145

---

## Follow-up (auto-implement-issue, iteration 1, 2026-05-26)

Addressed review findings from #pullrequestreview-4365677478:

- **B1** `7111ca0` — Regenerated `cli/internal/api/client.gen.go`
  alongside `cli/api/openapi.json`. `DeleteTargetApiV1TargetsNameDelete`
  (request struct, response struct, request fn, parser — 6 references)
  now present in the Go client; CI `CLI API snapshot freshness`
  unblocked.
- **B2** `7111ca0` — Added `Target.deleted_at.is_(None)` to the
  topology scheduler's `_run_one_sweep` WHERE clause; soft-deleted
  targets no longer keep being probed every cadence (regression test
  `test_sweep_excludes_soft_deleted_targets`).
- **M1** `7111ca0` — Added `TargetORM.deleted_at.is_(None)` to
  `_target_names` in the broadcast feed dropdown; soft-deleted target
  names no longer leak into the filter bar (regression test
  `test_target_dropdown_excludes_soft_deleted_targets`). The dropdown
  is now in parity with `GET /api/v1/targets` + MCP `list_targets`.

Adjacent sweep result: a `grep` for `select(Target...)` /
`select(TargetORM...)` across `backend/src/meho_backplane/` surfaces
no other read paths missing the filter. The
`audit_query/replay.py` + `audit_query/query.py` LEFT JOINs
intentionally keep the asymmetry so historical `target_name` resolves
for tombstoned audit rows (called out by the reviewer in
adjacent_improvements; not part of this PR's scope).

Coordination with sibling PR #1166 (T3 #1144): `_registered_products()`
in `api/v1/targets.py` keeps its inline shape and now carries an
explicit `TODO: consolidate with #1166's registered_product_tokens()`
pointer for a Wave 5 follow-up. Consolidating now would race the
two parallel PRs.

## Release Notes

* **New Features**
  * Added ability to delete targets via `DELETE /api/v1/targets/{name}` endpoint with soft-delete support; deleted targets are excluded from listings.
  * Made target product field patchable via `PATCH /api/v1/targets/{name}` with validation against registered products.
  * Added `force` parameter to override deletion conflicts when targets have existing references.

* **Documentation**
  * Updated OpenAPI specification to reflect soft-delete behavior and new product update capabilities.

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/1164?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


damir-topic:
@coderabbitai review

cli/api/openapi.json
```diff
@@ -710,6 +710,60 @@
             }
           }
         }
+      },
+      "delete": {
```
damir-topic, line 714: **[🛑 B1]** — CLI Go client out of date relative to backend OpenAPI document

**What:** `cli/api/openapi.json` was regenerated with the new `DELETE /api/v1/targets/{name}` operation, but `cli/internal/api/client.gen.go` was not regenerated to match. The CI check `CLI API snapshot freshness` failed:

```
##[error]cli/api/openapi.json and/or cli/internal/api/client.gen.go are out of date relative to the backend OpenAPI document. Run 'cd cli && make snapshot-openapi && make generate' and commit the result.
```

**Why it matters:** Merging this PR with the failing freshness check would land a CLI binary that cannot call the new DELETE route — the Go `Client` would not expose `DeleteTargetApiV1TargetsNameDelete`, so `meho target delete <name>` (or whatever the CLI verb is) would fail at compile time of any downstream code that adds the verb, AND the snapshot-freshness invariant the team relies on for "OpenAPI is the contract" gets one drift hop into the future.

**Required change:** From the repo root, run:

```bash
cd cli && make snapshot-openapi && make generate
git add cli/api/openapi.json cli/internal/api/client.gen.go
git commit --signoff -m "feat(cli): regenerate Go client for DELETE /api/v1/targets/{name}"
```

Confirm `cli/internal/api/client.gen.go` now contains `DeleteTargetApiV1TargetsNameDelete*` (currently `grep -c` returns 0).

**Source:** Failing job log https://github.com/evoila/meho/actions/runs/26460215609/job/77905584974

damir-topic (commented):




**Anchor task:** #1145 — G0.14-T4 DELETE /api/v1/targets/{name} + allow product in TargetUpdate
**Surface area:** backend (REST + MCP + DB + migration)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `e055fbc`
**AI reviewer (CodeRabbit):** not received within 15min (rate-limited — bot reported "32 minutes and 9 seconds" until quota reset, beyond our wait budget)

### Acceptance criteria
- [x] **Part A:** `DELETE /api/v1/targets/{name}` returns 204 on success — verified by `test_delete_target_soft_deletes_and_returns_204`
- [x] **Part A:** Cascade-checks `graph_node.target_id` with 409 + `?force=true` hint; force-true proceeds — verified by `test_delete_target_with_graph_node_refs_returns_409` + `_force_true_succeeds`
- [⚠️] **Part A:** "Soft-deletes (with `deleted_at`) ... picked + documented in the PR" — soft-delete picked + documented, BUT the `deleted_at IS NULL` filter is **not** applied at every read path (B2 — topology scheduler bypasses it)
- [x] **Part A:** RBAC-gated to `tenant_admin`; cross-tenant DELETE returns 404 conflation — verified by `_operator_role_returns_403` + `_cross_tenant_returns_404`
- [x] **Part A:** Audit row written (`targets.delete` op id) — bound via structlog contextvars on entry; AuditMiddleware writes
- [x] **Part B:** `TargetUpdate.product` accepts product-token enum; rejects unknown with structured 422 (`kind=unknown_product`, `valid_products[]`) — verified by `test_patch_product_unknown_returns_422`
- [N/A] **Part B (optional):** PATCH response includes `probe_result` — out of scope per PR design
- [x] Test coverage: DELETE happy path, cascade, force, PATCH typo rejection, PATCH success, plus 13 supporting resolver + schema + DB tests
- [⚠️] `Python (ruff + mypy + pytest)` green — backend Python CI ✅, but **CI is BLOCKED** on the `CLI API snapshot freshness` check (B1)

### Project-board alignment
|---|---|---|---|---|---|
| #1145 | OPEN | Closes | ✅ | ⚠️ partial (see Findings) | close on merge after B1/B2 fixed |

No undisclosed scope creep — PR touches exactly the surface area the issue calls out plus the legitimate adjacent updates (`docs/codebase/error-message-shape.md` audit-table extension, `cli/api/openapi.json` regen).

### Predicted vs. actual diff
- ✅ migration `0028_add_targets_deleted_at.py` — additive nullable column, reversible
- ✅ `backend/src/meho_backplane/db/models.py` — `deleted_at` column added
- ✅ `backend/src/meho_backplane/targets/schemas.py` — `Target.deleted_at` + `TargetUpdate.product`
- ✅ `backend/src/meho_backplane/targets/resolver.py` — `deleted_at IS NULL` on all 5 internal selects
- ✅ `backend/src/meho_backplane/api/v1/targets.py` — DELETE route, PATCH `product` validation, list filter
- ✅ `backend/src/meho_backplane/mcp/tools/topology.py` — MCP list filter
- ✅ `docs/codebase/error-message-shape.md` — audit-table extended with two new compliant surfaces
- ⚠️ `cli/api/openapi.json` — regenerated, BUT `cli/internal/api/client.gen.go` NOT regenerated (B1)
- 🕳️ `backend/src/meho_backplane/topology/scheduler.py` — predicted to need a `deleted_at IS NULL` filter; not in PR (B2)
- 🕳️ `backend/src/meho_backplane/ui/routes/broadcast/feed.py` — predicted to need a `deleted_at IS NULL` filter for parity with REST/MCP lists; not in PR (M1)

### Integration verification
- [x] `ruff check` on changed files — clean
- [x] backend Python CI (`Python (ruff + mypy + pytest)`) — SUCCESS in upstream CI
- [x] All new + impacted tests — **104 passed, 1 skipped** locally in worktree (`test_api_v1_targets.py`, `test_targets_resolver.py`, `test_targets_schemas.py`, `test_db_targets.py`, `test_mcp_tools_topology.py`)
- [x] Migration replay: `alembic upgrade head` → `downgrade -1` → `upgrade head` clean on file-backed SQLite
- [❌] `CLI API snapshot freshness` — **FAILED** on the upstream CI; Go client out of date relative to backend OpenAPI document (B1)
- [⚠️ reasoned-only] Topology refresh scheduler — not exercised end-to-end, but read-trace confirms it bypasses the `deleted_at` filter (B2)

- 🛑 **B1** — `cli/internal/api/client.gen.go` not regenerated alongside `cli/api/openapi.json`; CI `CLI API snapshot freshness` FAILED. Required change: run `cd cli && make generate` and commit the result.
- 🛑 **B2** — `topology/scheduler.py:228` enumerates all targets per tenant without filtering `deleted_at IS NULL`; the background refresh loop will keep probing soft-deleted targets, generating audit rows, broadcast events, and connector calls forever. Required change: add `Target.deleted_at.is_(None)` to the `_run_one_sweep` WHERE clause + a test asserting a soft-deleted target is skipped by the sweep.
- ⚠️ **M1** — `ui/routes/broadcast/feed.py:178-183` (`_target_names`) does not filter soft-deleted rows; the broadcast feed dropdown will surface tombstoned target names alongside live ones, breaking parity with the REST + MCP list endpoints. Required change: add `TargetORM.deleted_at.is_(None)` to the WHERE clause.

Items the author may pick up in a future PR:

- The audit-replay/query paths (`audit_query/replay.py:144-152`, `audit_query/query.py:141-150`, 211) intentionally keep the LEFT JOIN against soft-deleted targets so `target_name` resolves for historical rows. This is the right call per the design intent ("audit references must keep pointing"), but a one-liner module docstring note in `audit_query/query.py` calling out the asymmetry with the live-read paths would save the next reviewer from re-deriving it.
- The inline `_registered_products()` helper in `api/v1/targets.py:247-264` is acknowledged in the docstring as a temporary inline copy pending T3 (#1144) — once both land, consolidate into a shared helper in `connectors/registry.py` (likely as `registered_product_set()`) so POST + PATCH share one source of truth for "valid product".

- Adopted: 0 (bot rate-limited and did not submit a review within 15-min budget; rate-limit message reported quota resets in 32+ minutes)
- Disputed: 0
- Out of scope: 0
- Recommendation: if the orchestrator re-spawns this skill against a later iteration, CodeRabbit may have recovered and its findings can be folded in then.

### Docs
- [x] `docs/codebase/error-message-shape.md` — extended with two new compliant rows (`PATCH /api/v1/targets/{name}` 422 unknown_product and `DELETE /api/v1/targets/{name}` 409 target_has_references); both messages follow the three-clause shape (code → value/remediation → doc reference). Verified against the convention text.

- Migration design is clean: additive nullable column, reversible verified, justified docstring explaining why no index yet and why no UNIQUE constraint changes.
- The cascade design (block graph_node refs, allow `?force=true`, exclude audit refs by design) is well-reasoned and documented in both the route docstring and the PR body.
- Test coverage on the new contracts is dense and behavioural — RBAC, cross-tenant, soft-delete-by-name, soft-delete-by-alias, near-miss exclusion, force-true, structured 422 shape with `valid_products`, and the "same-value PATCH passes without registry lookup" edge case (which catches the mid-rolling-restart connector-registry blip).
- The T11 error-shape convention is faithfully applied: structured `detail` with `kind`, `message` naming the offending value + the `?force=true` remediation + the doc reference, plus the audit-table is extended in the same PR per the convention's "extend at the same time the surface lands" rule.
- The `same-value PATCH` short-circuit (`new_product != t.product`) is a thoughtful guardrail against transient unregistered-state at request time.

damir-topic:
@coderabbitai review

backend/alembic/versions/0028_add_targets_deleted_at.py
```diff
@@ -0,0 +1,97 @@
...
+Alembic's batch-mode fallback isn't required.
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0028"
```
damir-topic: **[🛑 B1]** — Migration revision `0028` conflicts with `0028_supersede_rdc_internal_seed.py` on `main`

**What:** This PR's migration `backend/alembic/versions/0028_add_targets_deleted_at.py` declares `revision: str = "0028"` and `down_revision: str | None = "0027"`. But `main` already merged `backend/alembic/versions/0028_supersede_rdc_internal_seed.py` (PR #1162, G0.13-T7) which also declares `revision = "0028"` and `down_revision = "0027"`. Alembic sees two heads at `0028` and refuses to upgrade:

```
alembic.util.exc.CommandError: Multiple head revisions are present for given
  argument 'head'; please specify a specific target revision, '<branchname>@head'
  to narrow to a specific head, or 'heads' for all heads
```

This is the root cause of both red CI checks on iteration 2:
- `Python (ruff + mypy + pytest)` — 6 ERROR rows at test setup, all on the same `MultipleHeads: 0028, 0028` exception.
- `Python (integration testcontainers)` — same exception on `test_client_credentials_grant_end_to_end` setup.

**Why it matters:** The PR cannot merge in its current state. Alembic refuses to compute an upgrade path; every test that depends on the migration ledger (which is most of them) fails at setup. The fix is mechanical — re-anchor this PR's migration on top of `0028_supersede_rdc_internal_seed.py` so the head chain is linear again.

**Required change:**
1. Rename the file: `backend/alembic/versions/0028_add_targets_deleted_at.py` → `backend/alembic/versions/0029_add_targets_deleted_at.py`.
2. In the renamed file, update the revision header:
   ```python
   revision: str = "0029"
   down_revision: str | None = "0028"
   ```
3. Update the docstring's `Revision ID: 0028` line to `Revision ID: 0029`.
4. Re-run `alembic upgrade head → downgrade -1 → upgrade head` on a fresh SQLite to confirm the chain is linear.
5. Re-run `make snapshot-openapi && make generate` (the snapshot is migration-agnostic, but verify nothing in the openapi description text referenced `0028`).

**Source:** Alembic 1.13 docs, [Working with Branches](https://alembic.sqlalchemy.org/en/latest/branches.html#working-with-explicit-branches) — "in a fully linear migration sequence, there is only one head revision at any time". Confirmed in repo: `git ls-tree origin/main backend/alembic/versions/` lists `0028_supersede_rdc_internal_seed.py` as the current head; this PR adds a second `revision="0028"` against the same `down_revision="0027"`.

backend/src/meho_backplane/api/v1/targets.py
```diff
@@ -463,12 +545,45 @@ async def update_target(
...
+                    "message": (
+                        f"product={new_product!r} is not registered; "
+                        f"pick one of {valid_products!r} or register a "
+                        f"connector for it before retrying. "
+                        f"See docs/codebase/error-message-shape.md for "
+                        f"the convention."
+                    ),
+                },
+            )
     for k, v in updates.items():
```
damir-topic, line 669: **[🛑 B2]** — `PATCH /api/v1/targets/{name}` with `{"product": null}` falls through to DB IntegrityError

**What:** `TargetUpdate.product` is typed `str | None = Field(default=None, min_length=1, max_length=100)`, so a client can send `{"product": null}` and `body.model_dump(exclude_unset=True)` returns `{"product": None}`. The PATCH route reads:

```python
updates = body.model_dump(exclude_unset=True)
new_product = updates.get("product")
if new_product is not None and new_product != t.product:
    # validate against _registered_products(), 422 on mismatch
    ...
for k, v in updates.items():
    setattr(t, k, v)
```

The `if new_product is not None` guard skips validation when `new_product` is `None` — so the loop unconditionally runs `setattr(t, "product", None)`. But `targets.product` is `NOT NULL` at the DB layer (`db/models.py`, migration `0004`). On `await session.commit()`, SQLAlchemy raises `IntegrityError`, FastAPI converts that to a 500, and the operator gets a stack-trace surface instead of the structured 422 the rest of this PR's validation discipline establishes. (Also flagged by CodeRabbit at `api/v1/targets.py:589`.)

**Why it matters:** PATCH is the only recovery path for a misregistered product (the primary use-case driving this whole task per signal 6 of `claude-rdc-hetzner-dc#697`). A malformed PATCH from a CLI / agent — easy to send by forgetting to strip a `None` default before serializing — bypasses the structured-error contract every other validation site on this route honours. The whole point of `docs/codebase/error-message-shape.md` is that every error path returns a structured detail; a 500 from an unhandled IntegrityError breaks that contract.

**Required change:** Insert the null-check before the `for k, v in updates.items()` loop:

```python
updates = body.model_dump(exclude_unset=True)
new_product = updates.get("product")
if "product" in updates and updates["product"] is None:
    raise HTTPException(
        status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
        detail={
            "kind": "invalid_null",
            "field": "product",
            "message": (
                "product cannot be null; targets.product is NOT NULL. "
                "Pass a valid product token instead. "
                "See docs/codebase/error-message-shape.md for the convention."
            ),
        },
    )
if new_product is not None and new_product != t.product:
    ...
```

Add a regression test in `tests/test_api_v1_targets.py` that PATCHes `{"product": null}` and asserts `response.status_code == 422` with `detail["kind"] == "invalid_null"`.

**Source:** `backend/src/meho_backplane/db/models.py:Target.product` is `Mapped[str] = mapped_column(Text, nullable=False)`. `docs/codebase/error-message-shape.md` codifies the structured-422 convention this PR ships under (and which this PR extends — line 2 of the docs diff is exactly the new audit table entry).

cli/api/openapi.json
```diff
@@ -710,6 +710,60 @@
...
+            "description": "Validation Error",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/HTTPValidationError"
+                }
+              }
+            }
+          }
+        }
```
damir-topic, line 804: **[⚠️ M1]** — OpenAPI document advertises 204 + 422 but the DELETE route also returns 409

**What:** `DELETE /api/v1/targets/{name}` raises `HTTPException(status_code=409, detail={"kind": "target_has_references", ...})` when the target is referenced by `graph_node` rows and `?force=true` was not passed. The PR's prose documents this exhaustively (PR body + route docstring + 409-conflict regression test in `test_api_v1_targets.py`). But the generated `cli/api/openapi.json` only declares two responses for this operation:

```json
"responses": {
  "204": {"description": "Successful Response"},
  "422": {"description": "Validation Error", ...}
}
```

No `409` entry. Downstream, `cli/internal/api/client.gen.go`'s `DeleteTargetApiV1TargetsNameDeleteResponse` struct exposes only `Body` + `HTTPResponse` + `JSON422` — no typed `JSON409`. (Also flagged by CodeRabbit at `cli/api/openapi.json:766` and `client.gen.go:15362`.)

**Why it matters:** The whole point of a generated OpenAPI document is that it's the single source of truth for clients. Consumers of the Go CLI must hand-roll `if rsp.StatusCode == 409 { json.Unmarshal(bodyBytes, ...) }` against an untyped body to handle the cascade-conflict path — exactly the asymmetry "declare every error path" exists to prevent. Future Python clients generated from the same spec carry the same gap.

**Required change:** Two edits, both small:

1. In `backend/src/meho_backplane/api/v1/targets.py`, add `responses=` to the `@router.delete` decorator:
   ```python
   @router.delete(
       "/{name}",
       status_code=status.HTTP_204_NO_CONTENT,
       response_class=Response,
       responses={
           status.HTTP_409_CONFLICT: {
               "description": (
                   "Target is referenced by graph_node rows; retry "
                   "with ?force=true to soft-delete anyway."
               ),
           },
       },
   )
   ```
2. Regenerate the snapshot + Go client:
   ```bash
   make snapshot-openapi && make generate
   ```

Verify `cli/api/openapi.json` lists `"409"` under the DELETE operation's `responses`, and `cli/internal/api/client.gen.go`'s `DeleteTargetApiV1TargetsNameDeleteResponse` gains a `JSON409` field that the response reader populates.

**Source:** FastAPI [Additional Responses](https://fastapi.tiangolo.com/advanced/additional-responses/) — `responses=` on a route operation is the canonical way to surface non-default status codes in `openapi.json`. The existing `POST /api/v1/targets` route has a similar undocumented 409 (target-name conflict); fixing it in the same edit is reasonable but not strictly in scope.

backend/tests/test_api_v1_targets.py
```diff
@@ -699,3 +699,399 @@ def test_create_unauthenticated_returns_401(client: TestClient) -> None:
...
+            "/api/v1/targets/to-delete",
+            headers={"Authorization": f"Bearer {_admin_token(key, tenant_id)}"},
+        )
+    assert response.status_code == 204
+
+    # The row stays in the DB but is now soft-deleted.
+    sm = get_sessionmaker()
+    async with sm() as session:
+        row = (await session.execute(_select(TargetORM).where(TargetORM.id == t.id))).scalar_one()
+    assert row.deleted_at is not None
```
damir-topic, line 964: **[⚠️ M2]** — DELETE happy-path test does not assert the `targets.delete` audit row is written

**What:** The PR body claims `audit_op_id='targets.delete'` is bound on entry to the DELETE route, mirroring `targets.discover` precedent. The route code confirms this binding via `structlog.contextvars.bind_contextvars(audit_op_id=_DELETE_OP_ID, audit_op_class="write")`. But none of the DELETE tests (`test_delete_target_soft_deletes_and_returns_204`, `test_delete_target_with_graph_node_refs_force_true_succeeds`) query `AuditLog` to verify the row reaches the DB with the right `audit_op_id`. The current assertions stop at "row.deleted_at is not None" + "follow-up GET returns 404". (Also flagged by CodeRabbit at `tests/test_api_v1_targets.py:956`.)

**Why it matters:** The audit row is a load-bearing acceptance criterion on issue #1145 ("Audit row written (`meho.targets.delete` or similar)") — it's how a tenant_admin reviews who deleted what, and how G8's cross-tenant audit query surfaces deletions. Without an explicit assertion, a regression that swaps the `bind_contextvars` call to a different op_id, or moves the binding after a code path that short-circuits to 409, lands green. "I bound a contextvar" and "an audit row exists with audit_op_id='targets.delete'" are two different invariants — only the second is the contract.

The same pattern is already exercised cleanly by `test_targets_audit_integration.py` for the create/discover/probe/refresh verbs — the DELETE addition is the missing parallel.

**Required change:** In `tests/test_api_v1_targets.py`, after the `assert response.status_code == 204` line in both `test_delete_target_soft_deletes_and_returns_204` and `test_delete_target_with_graph_node_refs_force_true_succeeds`, add:

```python
from meho_backplane.db.models import AuditLog
from sqlalchemy import select as _select

async with sm() as session:
    audit_row = (
        await session.execute(
            _select(AuditLog).where(
                AuditLog.target_id == t.id,
                AuditLog.audit_op_id == "targets.delete",
            )
        )
    ).scalars().first()
assert audit_row is not None, "DELETE must write an audit row with op_id='targets.delete'"
```

(If the existing test suite has a `_assert_audit_row(target_id, op_id)` helper, prefer that for consistency.)

**Source:** Issue #1145 acceptance criterion: "Audit row written (`meho.targets.delete` or similar)". Audit-assertion shape modelled on `tests/test_targets_audit_integration.py`.

damir-topic:




**Anchor task:** #1145 — G0.14-T4 DELETE /api/v1/targets/{name} + allow product in TargetUpdate
**Surface area:** backend (api, db, scheduler, ui), cli (openapi.json + Go client), docs
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `7111ca08fb5a3b95a85db42e590fc163b6774d50`
**AI reviewer (CodeRabbit):** received in 1198s — 4 inline findings (3 of 4 adopted with attribution, 1 surfaced as M2 with refined context)

### Acceptance criteria

| Criterion | Status | Evidence |
|---|---|---|
| Part A: DELETE returns 204 on success; cascade-checks audit + graph references | x | `test_delete_target_soft_deletes_and_returns_204` — passed locally |
| Part A: cascade-check + `?force=true` semantics | x | `test_delete_target_with_graph_node_refs_returns_409` + `_force_true_succeeds` — both pass |
| Part A: RBAC-gated to tenant_admin; cross-tenant DELETE returns 404 | x | `test_delete_target_operator_role_returns_403` + `test_delete_target_cross_tenant_returns_404` |
| Part A: Audit row written (`meho.targets.delete` or similar) | partial — see M2 | `audit_op_id='targets.delete'` is bound on entry; no test asserts the AuditLog row reaches the DB |
| Part B: `TargetUpdate.product` accepts product-token enum; rejects unknown with structured 422 | x | `test_patch_product_unknown_returns_422` + `test_patch_product_valid_succeeds` — both pass for the non-null case (see B2 for null case) |
| Tests cover DELETE happy path, cascade, PATCH product typo, PATCH product success | x | 10 new tests on the route + 3 new soft-delete-filter tests on the resolver |
| `Python (ruff + mypy + pytest)` green | **fail** — see B1 | both Python CI jobs RED on `MultipleHeads: 0028, 0028` (migration conflict with merged PR #1162) |

### Project-board alignment

|---|---|---|---|---|---|
| #1145 (Task G0.14-T4) | In Progress | Closes | ✅ | partial — see B1/B2 | close on merge once blockers fixed |
| #1139 (Initiative) | Todo (rolls up from tasks) | parent | — | n/a | initiative-level, not gated by this PR |
| #221 (Goal G0) | Todo | grandparent | — | n/a | goal-level |

### Predicted vs. actual diff

- ✅ DELETE route + `deleted_at` column + soft-delete filter on `resolve_target`, `list_targets`, MCP `list_targets`, broadcast feed dropdown, topology scheduler — all expected and delivered.
- ✅ `TargetUpdate.product` + inline `_registered_products()` validator with structured 422.
- ✅ Migration `0028_add_targets_deleted_at.py` — but conflicts with `main` (see B1).
- ✅ docs/codebase/error-message-shape.md audit-table entry for the two new error sites.
- ✅ Iteration-1 fix sweep: scheduler + feed dropdown filters with regression tests verified locally (`test_sweep_excludes_soft_deleted_targets` and `test_target_dropdown_excludes_soft_deleted_targets` both pass — they exercise the production filter end-to-end, not just compile).
- ❓ Missing: `responses={409: ...}` on the DELETE decorator → openapi.json omits the 409 path → Go client lacks typed JSON409 (see M1).
- 🕳️ Missing: explicit null-rejection on `TargetUpdate.product` when client sends `{"product": null}` (see B2).
- 🕳️ Missing: audit-log row assertion in the DELETE happy-path tests (see M2).

### Integration verification

- [x] `uv run ruff check src/ tests/` (in worktree) — clean.
- [x] `uv run pytest tests/test_topology_scheduler.py tests/test_ui_broadcast_filters.py` — 34 passed (the two new regression tests for B2/M1 from iter-1 both green).
- [⚠️ not run] `pytest tests/ -x` full sweep — locally the worktree has only this PR's `0028`, so it would pass locally. The CI failure is the **rebase-on-main multiple-heads** path; reproducing requires merging origin/main into the branch first. Confirmed authoritative via the CI logs.
- [x] migrations replay reasoned about — `alembic upgrade head` works on the worktree alone (one `0028`) but fails on `git merge origin/main` (two `0028`). See B1 source citation.
- [❌] `Python (ruff + mypy + pytest)` CI job — **RED** on iteration 2 due to `MultipleHeads: 0028, 0028`. Root cause is B1.
- [❌] `Python (integration testcontainers)` CI job — **RED** on the same exception.
- [x] `CLI API snapshot freshness` CI job — **GREEN** (iteration-1 B1 cleanly resolved by the fix commit; the Go client is now in sync with `openapi.json` modulo the 409 gap surfaced as M1 here).
- [x] `Go (golangci-lint + go test)` — green.
- [x] `Helm (lint + template + kubeconform)` — green.
- [x] `CodeQL`, `Semgrep`, `TruffleHog`, `DCO` — all green.
- [⚠️ not run] `docker compose up` — out of scope for this PR (no compose changes).

- 🛑 **B1** — Migration `0028_add_targets_deleted_at.py` conflicts with `0028_supersede_rdc_internal_seed.py` (already on `main`); CI is RED on `MultipleHeads: 0028, 0028`. Required change: rename to `0029_add_targets_deleted_at.py`, set `revision="0029"` + `down_revision="0028"`.
- 🛑 **B2** — `PATCH /api/v1/targets/{name}` with `{"product": null}` bypasses validation and would set `targets.product` to NULL, hitting an unhandled DB IntegrityError → 500. Required change: insert an explicit-null check before the `setattr` loop, raise structured 422 with `kind=invalid_null`. (Also flagged by CodeRabbit at `targets.py:589`.)
- ⚠️ **M1** — `cli/api/openapi.json` advertises only 204 + 422 for DELETE; the route also returns 409 `target_has_references`. Required change: add `responses={409: ...}` to the `@router.delete` decorator, regenerate snapshot + Go client. (Also flagged by CodeRabbit at `openapi.json:766` and `client.gen.go:15362`.)
- ⚠️ **M2** — DELETE happy-path tests don't assert the `targets.delete` audit row reaches the DB. Required change: add an `AuditLog` query + assertion after `assert response.status_code == 204` in `test_delete_target_soft_deletes_and_returns_204` and `_force_true_succeeds`. (Also flagged by CodeRabbit at `test_api_v1_targets.py:956`.)

- The same `responses={409: ...}` pattern is missing on `POST /api/v1/targets` (which also returns 409 on duplicate name). Worth folding into the M1 edit if convenient.
- `_registered_products()` in `api/v1/targets.py` carries a TODO pointer to consolidate with sibling PR #1166's `registered_product_tokens()` — confirmed pragmatic for parallel-merge safety; the consolidation is captured for Wave 5 follow-up.
- The `_target_names` helper in `ui/routes/broadcast/feed.py` mirrors the resolver + scheduler + REST list filters; if a fifth read-path appears, a shared `_live_targets_stmt(tenant_id)` helper would reduce drift risk (out of scope here).

- Adopted: 4 (all 4 of CodeRabbit's inline findings → B2, M1, M2; the 4th is the Go-client JSON409 which folds into M1 as a downstream effect of regenerating the client after the openapi spec gets the 409).
- Disputed: 0.
- Out of scope: 0.
- CodeRabbit framed B2 as Major; promoted to Blocker here because PATCH is the primary recovery path the whole task ships for, and a 500 from an unhandled IntegrityError directly violates the structured-error contract this PR also extends `docs/codebase/error-message-shape.md` to cover.

### Docs

- [x] `docs/codebase/error-message-shape.md` updated with the two new error sites (`+2 lines` in this PR's diff).
- [⚠️] `docs/codebase/backend.md` documents `Target` but pre-dates `deleted_at`. Not edited in this PR — the soft-delete contract is documented via route + migration docstrings + comments at every read-path filter site. Either approach is acceptable; calling it out for future-PR consideration. Not a finding.

### What worked well

- The iteration-1 fix sweep is thorough and grounded: the scheduler test (`test_sweep_excludes_soft_deleted_targets`) exercises the production filter end-to-end via the real `_run_one_sweep` codepath rather than asserting on a mocked WHERE clause; same for the feed dropdown.
- Every read-path consumer of `targets` is updated coherently — resolver, REST list, MCP `list_targets`, broadcast feed dropdown, topology scheduler. The "five places + a sixth would drift" risk is the right call-out in the adjacent improvements.
- The cascade strategy (graph_node only, not audit) and the soft-delete-vs-hard-delete decision are documented exhaustively in both the PR body and the route docstring, with the v0.1-spec §6 cite intact.
- The `_registered_products()` inline + TODO to consolidate with sibling PR #1166 is the right pragmatic call — coordinating two parallel-merge branches via in-flight refactor is the rebase-storm shape we explicitly want to avoid.

damir-topic:
@coderabbitai review

damir-topic:




**Anchor task:** #1145 — G0.14-T4 DELETE /api/v1/targets/{name} + allow product in TargetUpdate
**Surface area:** backend (REST + MCP + DB + migration), devops (alembic), CLI (Go regen), docs
**Verdict:** APPROVE
**Head SHA reviewed:** `18dcc01`
**AI reviewer (CodeRabbit):** not received within 15 min — CodeRabbit acknowledged the `@coderabbitai review` trigger at 18:22:37Z but did not post a new review in the 15-min window. Its incremental-review mode (per the acknowledgement note) typically skips re-reviewing previously-reviewed commits when the new commits are direct responses to its own iter-1 findings. The four findings from CodeRabbit's iter-1 review (which fed iter-2) are all verified resolved in this iteration — see below.

### Acceptance criteria
- [x] **Part A**: DELETE /api/v1/targets/{name} returns 204; cascade-check against graph_node refs; soft-delete via `deleted_at`; `?force=true` semantics — verified via existing tests + audit-row assertions added in M2 fix.
- [x] **Part A**: RBAC-gated to tenant_admin; cross-tenant DELETE returns 404 — verified via existing tests.
- [x] **Part A**: Audit row written with `op_id='targets.delete'` — M2 fix landed the asserting test; both happy paths (unforced + forced) write the canonical audit row.
- [x] **Part B**: `TargetUpdate.product` accepts product-token enum; rejects unknown with structured 422 — verified via existing tests; B2 fix added rejection of explicit `null` with `kind="invalid_null"` (a fine T11-convention choice).
- [x] **Part B**: PATCH-time `product` validator short-circuits on equality so a re-asserting PATCH doesn't fail in a transient unregistered-state — verified via `test_patch_product_same_value_passes_without_validator`.
- [x] Tests cover DELETE happy path, cascade, PATCH product typo, PATCH product success — 39 tests in `test_api_v1_targets.py` all pass locally; 139 in the broader targets-area sweep.
- [x] `Python (ruff + mypy + pytest)` GREEN; `Python (integration testcontainers)` GREEN; all 16 CI lanes SUCCESS (1 SKIPPED deploy-to-rke2-infra by design).

### Project-board alignment
|---|---|---|---|---|---|
| #1145 | In Progress | Closes | yes | full | close on merge |
| #1139 | In Progress | Parent initiative | — | n/a | initiative-level; this is one of 11 tasks |
| #221 | In Progress | Parent goal | — | n/a | goal-level |

### Predicted vs. actual diff
- ✅ `backend/alembic/versions/0029_add_targets_deleted_at.py` (renamed from 0028; B1 fix)
- ✅ `backend/src/meho_backplane/api/v1/targets.py` (null guard at line 570; 409 in @router.delete responses; B2 + M1 fixes)
- ✅ `cli/api/openapi.json` + `cli/internal/api/client.gen.go` (regen with JSON409 typed wrapper; M1 fix)
- ✅ `backend/tests/test_api_v1_targets.py` (audit assertions + null-product regression test + slug-collision drive-by; M2 + B2 fixes)
- ✅ All other PR files retained from iter-2 (no unexpected drift)
- 🚫 None
- 🕳️ None

### Integration verification
- [x] `ruff check src/ tests/` — clean (worktree)
- [x] `mypy src/meho_backplane` — Success: no issues found in 335 source files (worktree)
- [x] `pytest tests/test_api_v1_targets.py` — 39 passed (worktree)
- [x] Targets-area sweep (`tests/test_api_v1_targets.py tests/test_targets_resolver.py tests/test_targets_schemas.py tests/test_db_targets.py tests/test_mcp_tools_topology.py tests/test_topology_scheduler.py tests/test_ui_broadcast_filters.py`) — 139 passed, 1 skipped (worktree)
- [x] `alembic heads` — `0029 (head)` (single head; worktree)
- [x] `alembic upgrade head → downgrade 0028 → upgrade head` — reversible cycle clean on aiosqlite (worktree)
- [x] `cd cli && make snapshot-openapi && make generate && git diff --exit-code cli/` — no diff (worktree); CLI artifacts are deterministically fresh
- [x] CI `Python (ruff + mypy + pytest)` — SUCCESS
- [x] CI `Python (integration testcontainers)` — SUCCESS
- [x] CI `CLI API snapshot freshness` — SUCCESS
- [x] CI `Go (golangci-lint + go test)` — SUCCESS
- [x] CI `Helm (lint + template + kubeconform)` — SUCCESS
- [x] CI `CodeQL`, `Semgrep SAST`, `Semgrep OSS`, `TruffleHog Secret Scan`, `DCO`, `Backward-compat migration guard` — all SUCCESS

### Iter-2 finding closure (B1 / B2 / M1 / M2)

**B1 — Migration revision `0028` conflicts with `0028_supersede_rdc_internal_seed.py`** → FIXED in `1b61203`. File renamed to `0029_add_targets_deleted_at.py`; `revision="0029"`, `down_revision="0028"`. `alembic heads` reports single head `0029`. Linear chain `0027 → 0028 → 0029` verified; upgrade/downgrade/upgrade cycle clean on aiosqlite. CI green.

**B2 — `PATCH {"product": null}` falls through to DB IntegrityError** → FIXED in `18dcc01`. Explicit guard at `api/v1/targets.py:570`:

```python
if "product" in updates and updates["product"] is None:
    raise HTTPException(
        status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
        detail={"kind": "invalid_null", "field": "product", "message": "..."},
    )
```

Placed before the setattr loop so the path is unreachable. The `invalid_null` kind discriminator is consistent with T11 — snake_case + machine-actionable field reference + remediation message + doc pointer. Regression test `test_patch_product_null_returns_422_not_500` asserts the structured 422.

**M1 — DELETE OpenAPI 409 advertisement** → FIXED in `18dcc01`. `responses={409: {...inline schema...}}` on `@router.delete` declares the `target_has_references` shape. Regen produces `cli/api/openapi.json` listing 204+409+422, and `cli/internal/api/client.gen.go` exposes `DeleteTargetApiV1TargetsNameDeleteResponse.JSON409` + the typed `DeleteTargetApiV1TargetsNameDelete409DetailKind` enum + the `ParseDeleteTargetApiV1TargetsNameDeleteResponse` handler that unmarshals 409 into the wrapper. `make snapshot-openapi && make generate` is deterministically idempotent on the worktree.

**M2 — AuditLog assertion in DELETE happy-path tests** → FIXED in `18dcc01`. Both `test_delete_target_soft_deletes_and_returns_204` and `test_delete_target_with_graph_node_refs_force_true_succeeds` now query `AuditLog` filtered by `method='DELETE'` + `target_id` and assert the canonical row with `payload['op_id']=='targets.delete'`, `payload['op_class']=='write'`, `status_code==204`, `operator_sub=='admin-1'`. The contextvar binding in the route handler is what surfaces the canonical op_id; a future refactor that drops it now fails closed.

### Drive-by review

**Test slug fix.** The merge of `origin/main` (which brought in `0028_supersede_rdc_internal_seed` seeding a tenant with `slug='default'`) collided with two pre-existing DELETE tests that constructed their own `Tenant(slug='default')` for graph_node FK setup. The drive-by switches to `slug=f"tenant-{tenant_uuid}"`. Verified no cross-test pollution risk: per-test SQLite file (`conftest.py:_default_database_url` copies the migrated template into `tmp_path / "default.db"` for every test), tenant lookup uses `session.get(Tenant, tenant_uuid)` before adding so re-runs are idempotent, and the per-tenant-UUID slug coexists with the migration seed's `default` slug. Pure test-idempotency repair.

- Adopted: n/a (CodeRabbit did not post a new review in the 15-min wait window after the 18:22:37Z trigger). Its iter-1 review's four findings (PATCH null, audit-log assertion, OpenAPI 409, Go JSON409) are all resolved this iteration and would have been the same findings to flag again.
- Disputed: 0
- Out of scope: 0

### Docs
- [x] `docs/codebase/error-message-shape.md` — already updated in iter-1; the new `invalid_null` site mirrors the existing `unknown_product` shape and is captured in the audit table.

### What worked well
- Migration renumbering executed correctly under the rebase-with-merge pattern; the `alembic heads` single-head invariant restored without leaving the branch open.
- `invalid_null` finding shape pins the absent-marker semantics with a precise docstring that explains *why* `model_dump(exclude_unset=True)` doesn't catch this and why the handler-level guard is load-bearing — future refactors that drop it fail closed.
- The audit-row tests now exercise the `audit_op_id` contextvar binding directly, which catches the silent-regression failure mode (route handler still returns 204 but the audit row carries the fallback `http.delete:...` op_id).
- CLI regen is deterministic: `make snapshot-openapi && make generate` produces no diff against the committed artifacts, confirming the inline 409 schema flows cleanly through oapi-codegen v2.5.0.

- `_registered_products()` and sibling PR #1166's `registered_product_tokens()` consolidation captured as Wave 5 follow-up; the TODO marker is in place.
- `POST /api/v1/targets` has the same undocumented 409 (duplicate-name conflict) — a parallel M1-style fix would be reasonable for v0.6.x parity, but out of scope here.

damir-topic (commented):
auto-review-pr iter-3: APPROVE — see https://github.com/evoila/meho/pull/1164#issuecomment-4547532376 for full review body. Own-author APPROVE is blocked by GitHub; verdict is authoritative in this iteration's review-results file and in the linked comment body. CodeRabbit did not post a new review in the 15-min wait window (incremental review skipped — its iter-1 findings are all resolved). All four iter-2 blockers/majors (B1 migration renumber, B2 null-product 422, M1 DELETE 409 OpenAPI, M2 audit-row assertions) verified resolved. CI all green.

Issue #1242, opened 2026-05-27
Task: G3.11-T8 Reconcile T1↔T3 connector version-string drift — catalog says "v3", registry says "3"


damir-topic:
Parent goal: #214
Parent initiative: #1220

## Problem

T1 #1221 (PR #1231) registered the GitHub REST connector in the v2
registry as:

    register_connector_v2(product="gh", version="3", impl_id="gh-rest", ...)

The `version="3"` value (digit-prefix, no `v`) is forced by the
dispatcher's `parse_connector_id` regex (`^[0-9][A-Za-z0-9._]*$` —
letter-prefixed versions like `"v3"` are rejected at parse time).
T1's commit message documents this trade-off and preserves the
upstream "v3" label in `FingerprintResult.version` and in docs.

T3 #1223 (PR #1228), shipped in parallel with T1 in Wave 1 of the
auto-implement run, wrote the catalog YAML at
`backend/src/meho_backplane/operations/ingest/catalog.yaml`:

```yaml
- product: gh
  version: v3
  impl_id: gh-rest
```

The catalog `version` field is `"v3"` (letter-prefixed). The
dispatcher's tuple-lookup expects `(product, version, impl_id)` to
match the registry triple exactly. **Mismatch**: catalog
`("gh", "v3", "gh-rest")` will not resolve to the registered
`("gh", "3", "gh-rest")`.

**Latent today** — live ingest is currently blocked by the G0.7
OpenAPI parser ref-bucket gap (separate Task: G3.11-T7). Once T7
lands and live ingest runs, this drift surfaces as the connector
class lookup failing post-ingest: the ~700 ingested rows would
carry the catalog's `version="v3"` but no connector class is
registered under that key.

T5 #1225's worker hit this directly when writing the
`requires_approval` annotation script and worked around with a
`(product, version, impl_id)` triple lookup directly against the
catalog + registry tables (rather than going through
`parse_connector_id`). The script ships, but the underlying drift
remains in the dispatcher routing path.

## Current state

- **Catalog YAML:**
  `backend/src/meho_backplane/operations/ingest/catalog.yaml` has
  `version: v3` for the `gh-rest` row.
- **Registry registration:**
  `backend/src/meho_backplane/connectors/github/__init__.py` has
  `version="3"`.
- **Parser regex:** `parse_connector_id` in
  `backend/src/meho_backplane/connectors/registry.py` (verify path)
  enforces `^[0-9][A-Za-z0-9._]*$` for the version field.
[...]

Issue #585, opened 2026-05-17
Task: test-isolation — global registries (_TOOLS/_REGISTRY/registrars) leak across modules, blocks pytest-xdist (#540-class)


damir-topic:
No fitting parent Initiative — cross-cutting test-infra hygiene spanning MCP registry (`mcp/registry.py`), connector registry (`connectors/registry.py`), typed-op registrars, and audit middleware. Filed standalone (same pattern as #578/#579). Related: #540 (the prior, narrower instance of this exact class), #578, #579.

## ⚠️ This is the HARD PREREQUISITE for parallelizing CI — not a cleanup

The slowest required gate (`ci.yml` `Python (ruff + mypy + pytest)`, ~49 min vs ~5 min locally) **cannot be parallelized with pytest-xdist until this is fixed.** Establishing that took a full investigation; recorded here so it isn't re-litigated:

- `pytest-xdist -n auto --dist loadscope` was implemented + measured: serial 4:43 → xdist **1:29** locally (~7 cores). The speedup is real and large.
- Controlled same-machine/same-commit (`587542c`) diff, identical flags: **serial = 2007 passed / 0 failed / 0 errors**; **xdist = 2006 passed / 1 failed / 2 errors**. Failure set is **xdist-only**:
  1. `FAILED tests/test_mcp_registry.py::test_tools_list_endpoint_returns_filtered_tools[read_only]`
  2. `ERROR  tests/test_audit_middleware.py::test_audit_write_failure_converts_request_to_500`
  3. `ERROR  tests/test_audit_middleware.py::test_handler_exception_writes_audit_row_with_status_500`
- A "split-pass" workaround (parallelize the bulk, run just these 2 modules in a dedicated **serial** pass) was implemented and **proven non-viable**: the 2 modules are **red even in an isolated serial pass** (`test_mcp_registry.py` alone serial = "1 failed, 25 passed"; the 2-file serial pass = "1 failed, 5 passed"). They pass **only** inside the full-suite serial collection order, because earlier modules' app-lifespan `TestClient` usage populates the global registries into the exact state their assertions expect. No partition reproduces that — so the ONLY robust fix is making the tests order-independent.

The perf PR therefore ships only the safe wins (fastembed model `actions/cache` + `HF_TOKEN`) and leaves the unit job **serial**, hard-blocked on this Task. Priority is high: this is the sole unlock for the ~49 min → ~10 min win the job header has flagged as pending for months.

## Current state

[...]

PR #1259, opened 2026-05-27
feat(connectors): G3.11-T10 validator asserts (product, version, impl_id) triple registration

### LINKED ISSUE #1253: Task: G3.11-T10 Extend validate_catalog_registry_coverage to assert (product, version, impl_id) triple registration
Parent goal: #214
Parent initiative: #1220

## Problem

`validate_catalog_registry_coverage` (location TBD — `grep -rn` for it; likely in `backend/src/meho_backplane/operations/ingest/` or `backend/src/meho_backplane/connectors/`) only checks that **every catalog row has a registered connector class**. It does NOT check that the catalog row's `(product, version, impl_id)` triple matches the registry's actual registered triple.

T8 #1242 (PR #1249, merged `c480d63`) had to fix exactly this class of mismatch — T1 #1221 registered `("gh", "3", "gh-rest")` and T3 #1223 wrote catalog `version: v3`. The validator passed both PRs (the class WAS registered; just under a different version key) and the drift only surfaced when T5 #1225 worker tried to do a triple lookup.

**The validator's gap let this drift ship in parallel-Wave Tasks.** Extending the validator to assert triple registration would catch the class of bug at CI time — before it lands.

## Current state

- **Validator location:** find via `grep -rn 'validate_catalog_registry_coverage\|def validate_catalog' backend/src/meho_backplane/`.
- **Current behavior:** asserts the catalog row's `impl_id` (or whatever class-reference field) resolves to a registered connector class. Does NOT assert the registered triple.
- **Catalog row example post-T8:**
  ```yaml
  - product: gh
    version: "3"
    impl_id: gh-rest
    requires_connector_class: GitHubRestConnector
  ```
  The validator should now also assert that `("gh", "3", "gh-rest")` exists in the v2 registry table.

- **Registry table:** `backend/src/meho_backplane/connectors/registry.py` exposes `_connector_table_v2` (or similar) keyed by `(product, version, impl_id)`. Per G0.15-T6, every typed connector ships dual v1+v2 registration.

- **T8's fix:** aligned the gh-rest catalog/registry tuples and added 4 resolver regression tests that specifically assert the catalog→registry resolution. But those tests are gh-specific; the validator change would be **project-wide**.

## Desired state

1. **Extend the validator** to walk each catalog row and assert two things:
   - `requires_connector_class` resolves to a registered class (existing check).
   - The catalog row's `(product, version, impl_id)` triple has an entry in the v2 registry table.

2. **Failure mode:** if the triple is missing, surface a T11-compliant envelope per G0.14-T11 #1141:
   - `code: catalog_registry_triple_mismatch`
   - Message includes the catalog row's triple AND the closest registered triple for the same product (so the operator can see which field drifted).

3. **CI integration:** the validator already runs as part of the chassis startup smoke test (verify path); the extension stays at startup-time so a misregistration fails fast on `meho serve` boot, not at first dispatch.

4. **Regression test:** a unit test that constructs a catalog row + registry pair where the version drifts (e.g. catalog `version="v3"` vs registry `version="3"`), runs the validator, asserts the envelope shape.

5. **Self-test against current state:** run the validator against the live `catalog.yaml` post-T8 and assert it passes for ALL existing rows (vmware, vault, k8s, harbor, sddc-manager, nsx, vcf-*, hetzner-robot, bind9, gh, plus any new since this Task ships).

## Acceptance criteria

- [ ] Validator extended to check `(product, version, impl_id)` triple registration.
- [ ] T11-compliant envelope with `catalog_registry_triple_mismatch` code; closest-registered-triple hint.
- [ ] Unit test asserts the failure envelope on a synthetic drift.
- [ ] Validator passes against the current live `catalog.yaml` (no false positives on existing rows).
- [ ] Validator runs at chassis startup; failure fails-fast at boot, not at first dispatch.
- [ ] No regression in existing chassis startup tests.

## Out of scope

- Auto-fixing the drift at boot (e.g. patching the catalog from the registry triple). Detection only — operators fix the source of truth.
- Validator coverage for connectors that ship NO catalog row (some still use exclusively code-side typed-op declaration). The triple-check applies only to catalog-driven rows.

## References

- **Initiative parent:** #1220 G3.11 github-rest typed connector.
- **Goal parent:** #214 Connector parity.
- **T8 #1242 PR:** evoila/meho#1249 — surfaced this gap; worker noted in PR body: "The `validate_catalog_registry_coverage` only checks class registration, not the (product, version, impl_id) triple — extending it would prevent recurrence project-wide."
- **T1 #1221 PR:** evoila/meho#1231 — the registration side of the drift.
- **T3 #1223 PR:** evoila/meho#1228 — the catalog side of the drift.
- **G0.14-T11 #1141 error-message-shape** — applies to the new envelope.
- **G0.15-T6** — dual v1+v2 registration mandatory pattern; the validator extension reinforces this contract.
- **Best practices** — Python: structural assertion + structured-error envelope; chassis startup hook for fail-fast detection; per-connector test fixture in the same shape the registry uses today.

damir-topic:
## Summary
- Extend `validate_catalog_registry_coverage` to assert each catalog row's `(product, version, impl_id)` triple is a key in the v2 connector registry (G3.11-T10 #1253). Closes the gap that let T8 #1242's catalog-`v3`-vs-registry-`3` drift ship — the class-presence-only check passed because `GitHubRestConnector` was registered, just under a different version key.
- T11-compliant envelope: `catalog_registry_triple_mismatch:` code prefix, message names the offending catalog triple, the closest registered triple for the same product (with global fallback when the product slug itself is typo'd), the source-of-truth files to reconcile, and the convention doc.
- Validator now also runs at chassis startup after `_eager_import_connectors` + `load_catalog`, so a triple drift fails-fast at boot instead of surfacing as `no_connector` on the first dispatch.
- New session-scoped `_force_import_connector_modules` conftest fixture mirrors the existing `_force_import_mcp_modules` pattern (#585) — guarantees connector registrations land before the per-test `_isolate_global_registries` snapshot, so lifespan-driving tests see a populated registry and the new boot-time validator passes under pytest-xdist.

## Self-test against the shipped catalog
Verified that the extended validator passes for ALL 8 currently-shipped catalog rows against the live registry:

```
vmware/9.0/vmware-rest         sddc-manager/9.0/sddc-rest    harbor/2.x/harbor-rest
nsx/4.2/nsx-rest               gh/3/gh-rest (post T8 #1242)  vault/1.x/vault
k8s/1.x/k8s                    bind9/9.x/bind9-ssh
```

No drift on any non-gh connector — no pushback to the issue needed.

## Test plan
- [x] `ruff check` — clean
- [x] `mypy src/meho_backplane/operations/ingest/catalog.py src/meho_backplane/main.py` — clean
- [x] `pytest tests/test_operations_ingest_catalog.py` — 23 passed (existing 20 + 2 new T10 drift envelope tests + 1 unchanged passing under the extended check)
- [x] `pytest tests/test_mcp_startup_guard.py tests/test_connectors_registry.py` — startup-guard `test_lifespan_boots_when_backplane_url_resolves` boots cleanly under the new validator after the conftest fixture lands; both `test_lifespan_*` tests updated to mock `load_catalog` + `validate_catalog_registry_coverage` for the mocked-registry path
- [x] Live shipped-catalog self-test — validator passes for every existing connector row (above)
- [x] Synthetic drift unit test — `("gh", "v3", "gh-rest")` catalog row + `("gh", "3", "gh-rest")` registry triple raises `CatalogError` with the exact `catalog_registry_triple_mismatch:` envelope shape, names both triples, references `catalog.yaml` + `register_connector_v2` + `error-message-shape.md`
- [x] Global-fallback hint test — typo'd product (`k8z` vs `k8s`) still produces a useful closest-triple hint

## Out of scope
- Auto-fixing catalog↔registry drift at boot (detection only — operators reconcile the source of truth).
- Validator coverage for connectors with no catalog row (some still use exclusively code-side typed-op declaration).
- Mechanizing the T11 envelope assertion across other startup-time errors (stretch goal listed in `docs/codebase/error-message-shape.md`).

Closes #1253

### DIFF
--- a/backend/src/meho_backplane/main.py
+++ b/backend/src/meho_backplane/main.py
@@ -116,7 +116,7 @@ from meho_backplane.memory import (
 from meho_backplane.metrics import render_metrics
 from meho_backplane.middleware import BroadcastDetailMiddleware, RequestContextMiddleware
 from meho_backplane.operations import run_typed_op_registrars, set_default_reducer
-from meho_backplane.operations.ingest import load_catalog
+from meho_backplane.operations.ingest import load_catalog, validate_catalog_registry_coverage
 from meho_backplane.operations.jsonflux_reducer import JsonFluxReducer
 from meho_backplane.retrieval.embedding import get_embedding_service
 from meho_backplane.scheduler import start_scheduler, stop_scheduler
@@ -255,11 +255,20 @@ async def _run_lifespan_startup() -> None:
     # field, non-PEP-440 spec_info_version, duplicate (product, version))
     # raises here and crashes the lifespan, so CI's app-boot smoke fails
     # instead of the bad catalog surfacing as a 500 on first
-    # GET /api/v1/connectors/catalog. Registry-coverage of each entry's
-    # requires_connector_class is a CI regression test, not a startup
-    # guard (the registry's populated-ness is import-order-dependent
-    # under pytest-xdist; see catalog.py).
+    # GET /api/v1/connectors/catalog.
     load_catalog()
+    # Catalog ↔ registry triple coverage (G3.11-T10 #1253). At chassis
+    # boot the registry is fully populated (_eager_import_connectors
+    # above ran every connectors/<product>/__init__.py registration),
+    # so the validator can assert both the class-presence check
+    # (#743 crit. b) and the (product, version, impl_id) triple-
+    # registration check (T10) without the pytest-xdist import-order
+    # hazard the test suite hits. A mismatch fails-fast here rather
+    # than surfacing as ``no_connector`` on the first dispatch — the
+    # T8 #1242 class of bug (catalog ``version: v3`` vs registry
+    # ``version: "3"``) would have crashed the lifespan instead of
+    # shipping silently.
+    validate_catalog_registry_coverage()
     # Typed-op registration (G0.6-T-Refactor-Vault #390). See the
     # docstring for the contract; runs registrars connectors appended
     # to during the import pass above so descriptor rows are populated
--- a/backend/src/meho_backplane/operations/ingest/catalog.py
+++ b/backend/src/meho_backplane/operations/ingest/catalog.py
@@ -30,16 +30,34 @@ connector-catalog.md`` is the operator-facing pointer.
   YAML crash startup (and therefore CI's app-boot smoke). It does not
   touch the connector registry, so it is safe to run inside the lifespan
   regardless of import-cache state.
-* :func:`validate_catalog_registry_coverage` cross-checks every entry's
-  ``requires_connector_class`` against
-  :func:`~meho_backplane.connectors.registry.all_connectors_v2`. That is
-  the #743 criterion-(b) guard; it runs as a CI regression test (where
-  the full connector set is registered) rather than at startup, where the
-  registry's populated-ness is import-order-dependent under pytest-xdist.
+* :func:`validate_catalog_registry_coverage` cross-checks every entry
+  against :func:`~meho_backplane.connectors.registry.all_connectors_v2`
+  on two axes:
+
+  - **Class presence** (#743 criterion (b)) — the
+    ``requires_connector_class`` resolves to a registered connector class.
+  - **Triple registration** (G3.11-T10 #1253) — the ``(product, version,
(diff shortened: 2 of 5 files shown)


damir-topic:
@coderabbitai review

Issue #1307, opened 2026-05-28
G0.16-T5 gh/3 catalog label-vs-spec version-drift reconciliation


damir-topic:
Parent initiative: #1302

## Problem (SEV-2, Finding 18 in `claude-rdc-hetzner-dc#771`)

The gh/3 catalog entry's `version` field disagrees with the upstream OpenAPI document's `info.version` field, and the ingest validator rejects the mismatch:

```
POST /api/v1/connectors/ingest {"catalog_entry":"gh/3"}
  → HTTP 422 spec_label_mismatch:
    {"requested_version":"3",
     "spec_info_versions":[
       {"spec_uri":"https://raw.githubusercontent.com/.../api.github.com.json",
        "info_version":"1.1.4"}],
     "message":"spec/label version mismatch: requested_version='3' is
                incompatible with ['...' -> info.version='1.1.4']; either
                re-ingest under version='1.1.4' to match the spec, or
                supply specs whose info.version matches version='3'"}
```

GitHub's REST API uses "v3" as the *product line* identifier (the path prefix on some legacy endpoints; most current endpoints have no version prefix), while the OpenAPI document's `info.version` tracks the *spec's documentation versioning* (currently 1.1.4 — and growing). The two semantically distinct version fields will continue to diverge as github/rest-api-description releases new spec versions.

The 422 envelope is well-shaped — it names the conflict precisely. **The underlying design is what's broken**: the catalog `version` is forced to mean the same thing as `info.version`, but for GitHub-shaped APIs they're orthogonal concepts.

## Why this matters

Blocks gh/3 catalog ingest. Even if T2 (auth_model reconciliation) lands, gh-rest L2 ops can't be ingested via the catalog entry. The same class issue (T1 commits-and-grouping pod-crash) would still apply if this were resolved — but T5 is independently fixable and shouldn't wait on the bigger ingest-pipeline rework.

Same class as Finding 22 in the doc — catalog-vs-spec connector_id divergence, but Finding 22 affects vmware (where rdc-vcenter resolves to `vmware-rest-9.0` while ingest would produce `vmware-rest-9.0.0.0`).

## Acceptance

Three valid resolutions; pick one:

**(a) Catalog uses spec-aligned version:** `version="1.1.4"`. Forces the connector_id to `gh-rest-1.1.4` — semantically correct but ugly for operators, and changes every release. Not recommended.

[...]

PR #1324, opened 2026-05-29
fix(connectors): G0.16-T5 gh/3 catalog label-vs-spec drift opt-in

damir-topic:
## Summary

- Catalog rows can declare an opt-in compatibility range via the new
  `spec_info_versions_compatible` field; the ingest validator widens
  its verbatim/major-band check to accept upstream `info.version`
  values inside the declared band even if they differ from the
  catalog's `version` label.
- Adopts the opt-in on the `gh/3` row with `["1.x.x"]` so
  `POST /api/v1/connectors/ingest {"catalog_entry":"gh/3"}` against
  the live upstream stops failing `spec_label_mismatch` (catalog
  `"3"` vs upstream `info.version="1.1.4"` and growing).
- Resilient to upstream `1.x` bumps without manual catalog edits;
  only a `2.x` breaking change forces an operator decision.

Closes #1307 — addresses RDC #771 Finding 18 (gh/3 catalog
label-vs-spec mismatch, SEV-2). Per
`docs/codebase/api-shape-conventions.md` §9 this is the recommended
resolution (b) for connector families whose catalog `version` is a
product-line label divergent from the spec's documentation version.
The same field is available for G0.16-T6 Finding 22 / Task #1312 H
to adopt for the vmware variant if that task chooses approach (b).

## Design notes

- Pattern syntax: glob (`"1.x"`, `"9.0.x"`) or PEP 440 specifier set
  (`">=1.0,<2.0"`, `"~=1.4"`). Globs translate to PEP 440 specifiers
  so the matcher reuses the well-tested `packaging.specifiers`
  implementation. Any-of semantics across multiple patterns.
- The opt-in flows from the catalog row through the catalog-entry
  resolver in the REST route to
  `IngestionPipelineService.ingest(spec_info_versions_compatible=...)`.
  The explicit-quadruple shape (MCP admin tool, historical callers)
  doesn't carry the catalog row and therefore can't opt in —
  operators using it implicitly accept the historical strict check.
- A successful bypass emits a structured
  `connector_ingest_version_label_decoupled` log event so the audit
  trail records the decision and the values that flowed through it.
- Malformed patterns fail at catalog-parse time (CI app-boot smoke),
  not at first ingest — same fail-loud-early posture as
  `spec_info_version` PEP 440 validation.

## Test plan

- [x] `ruff check src/ tests/` — clean
- [x] `ruff format --check src/ tests/` — clean
- [x] `mypy src/ --ignore-missing-imports` — Success: no issues
[...]

Issue #1215, opened 2026-05-27
Task: G0.15-T6 version editable on TargetCreate/Update + wildcard-registration fanout across typed connectors (extends signal 6)


damir-topic:
Parent goal: #221
Parent initiative: #1209

## Problem

`claude-rdc-hetzner-dc#753` § "Target version field is uneditable —
extends signal #6, blocks every non-wildcard typed-connector
dispatch on a target with version=null".

**Reproducer at v0.7.0:**

```
$ call_operation vmware-rest-9.0 vmware.composite.datastore.usage target={"name":"rdc-vcenter"}
→ {
    "status": "error",
    "error": "no_connector: no implementation for product='vmware' version='9.0'",
    "extras": {
      "error_code": "no_connector",
      "exception_message": "no connector advertises support for (product='vmware', version=None)"
    }
  }
```

The G0.14-T1 #1142 structured envelope lands the diagnostic
correctly — `version=None` on the target vs `version=9.0` on the
connector advertisement. Resolver is right; **the gap is no
operator-driven path exists to fix the underlying `version=null`
state on the target**.

**Schema inspection at v0.7.0:**

- `TargetCreate.properties` = aliases, auth_model, extras, fqdn,
  host, name, notes, port, preferred_impl_id, product, secret_ref,
  vpn_required — **no `version`**.
- `TargetUpdate.properties` = same minus `name` — **no `version`**.

So:
- A fresh target via `POST /api/v1/targets` can't specify version
  at create time → defaults to `null`.
- An existing target with `version=null` can't have version set via
  `PATCH /api/v1/targets/{name}` → stays `null`.
- Probe (which discovers version via fingerprint) requires the
  resolver to find a connector first — and the resolver bails on
  `(product, null)` for any connector that doesn't register a
  wildcard. Chicken-and-egg.

**The k8s exception:** the k8s connector self-registers BOTH
`("k8s", "", "")` wildcard AND `("k8s", "1.x", "k8s")` versioned
(per G0.14-T2 #1156). Target `(product=k8s, version=None)` matches
the wildcard candidate, resolver returns ≥1 candidate, tie-break
ladder + preferred_impl_id picks the versioned one. **K8s is the
only typed connector that dispatches on a fresh target out-of-box.**

The consumer also tried PATCH `preferred_impl_id="vmware-rest"` (and
the full `"vmware-rest-9.0"` form). PATCH accepts both (200 OK,
persisted), but the resolver IGNORES preferred_impl_id when no
[...]

PR #1325, opened 2026-05-29
fix(api): G0.16-T6 API shape conventions sweep — 8 findings across REST/MCP/catalog (#1312)

### LINKED ISSUE #1312: G0.16-T6 API shape conventions sweep — converge 8 existing endpoints onto the conventions doc
Parent initiative: #1302

## Problem (SEV-4 sweep, Findings 3/6/7/8/10/13/15/22 in `claude-rdc-hetzner-dc#771`)

Eight surface-shape mismatches between today's endpoints and the conventions codified in [`docs/codebase/api-shape-conventions.md`](../../docs/codebase/api-shape-conventions.md) (PR #1310). Individually small; collectively the kind of "left as opportunistic" backlog that gets re-flagged by every dogfood cycle until somebody runs the sweep deliberately.

**This Task converges existing endpoints onto the conventions doc.** The doc is the rulebook; this Task is the operational execution against today's surface so the next dogfood cycle (RDC #1090 or its v0.9 successor) doesn't find any of these eight findings still OPEN.

The acceptance is per-finding. Each finding cites the conventions-doc section it violates, the v0.8.0 shape, and the target shape.

### A. Finding 3 — List-endpoint envelope unification (anchors §2)

Five list endpoints today, three shapes:

| Endpoint | v0.8.0 | Target |
|---|---|---|
| `GET /api/v1/connectors` | `{"connectors": [...]}` | `{items, next_cursor?, ...sidecars}` |
| `GET /api/v1/targets` | `[...]` bare | `{items, next_cursor?, ...sidecars}` |
| `GET /api/v1/conventions` | `{budget_status, entries}` | `{items, next_cursor?, budget_status}` |
| `GET /api/v1/audit/my-recent` | `{rows, next_cursor}` | `{items, next_cursor}` (rename `rows`→`items`) |
| `GET /api/v1/broadcast/overrides` | `[...]` bare | `{items, next_cursor?}` |

- [ ] All five endpoints accept `?envelope=v2` query parameter returning the unified shape; default stays v0.8.0 (bare/keyed) shape.
- [ ] CLI + MCP tool sister surfaces forward `envelope=v2` opt-in.
- [ ] Acceptance test exercises `?envelope=v2` against all five endpoints; assert shape conformance.
- [ ] CHANGELOG bullet under **Added**: `?envelope=v2` opt-in available on `connectors`, `targets`, `conventions`, `audit/my-recent`, `broadcast/overrides`. v0.x bare/keyed shapes remain default.

Deprecation flip (v0.10 cut) and removal (v0.11) are out of scope for this Task — file follow-up Tasks then.

### B. Finding 6 — `product` enum reconciliation (anchors §3)

TargetCreate enum spells SDDC Manager as `"sddc-manager"`; the catalog connector advertises `product: "sddc"`. Pick one.

- [ ] Reconcile: TargetCreate enum and catalog `product` field hold identical strings for every product they both name.
- [ ] Recommendation: drop catalog `product: "sddc"` → `"sddc-manager"` (catalog matches enum). Lower blast radius than enum churn.
- [ ] Sweep the full product set: check that every catalog entry's `product` matches a TargetCreate enum value. Surface any other mismatch as a check item.
- [ ] CHANGELOG: under **Changed**, name the catalog rename(s) + cite consumer signal.

### C. Finding 7 — `preferred_impl_id` Create-vs-Update agreement (anchors §3)

TargetCreate validates against base impl-id names (`nsx-rest`); TargetUpdate accepts versioned forms (`nsx-rest-4.2`). Same field, two different enums.

- [ ] TargetCreate accepts versioned form (`nsx-rest-4.2`). The Pydantic schema validator widens to match TargetUpdate.
- [ ] Schema enum description names the versioned form as canonical, with a one-line "the resolver's wildcard fan-out handles unversioned lookups" callout.
- [ ] Acceptance test: POST a target with `preferred_impl_id="nsx-rest-4.2"` → 200.
- [ ] CHANGELOG **Changed** bullet.

### D. Finding 8 — List-vs-detail field consistency (anchors §5)

`GET /api/v1/targets` returns `version: null, secret_ref: null, preferred_impl_id: null` for targets whose detail endpoint returns actual values. Silent masking.

Two acceptable resolutions per §5:

- [ ] **Pick one and apply uniformly to all list endpoints:** either (a) list returns the same fields as detail, OR (b) keep the current projection but rename to `/api/v1/targets/summary` with explicit "projection; fall through to detail for full state" docs, and add a fully-shaped `/api/v1/targets` that mirrors detail. Recommendation: (a) plus pagination via §2's `?envelope=v2`.
- [ ] Document the choice in the conventions doc's §5 with the live endpoint shape.
- [ ] If (a) is chosen, acceptance test asserts list-row field set ⊇ detail field set (no silent masking).
- [ ] CHANGELOG **Fixed** bullet.

### E. Finding 10 — REST↔MCP envelope agreement on topology (anchors §4)

`GET /api/v1/topology/dependents/{target}` returns bare array; MCP `query_topology` returns `{kind, nodes: [...]}`.

- [ ] REST endpoint returns `{kind, nodes, ...}` matching MCP (§4: migration is REST-toward-MCP, not the other way).
- [ ] Sweep the topology traversal endpoint set (`dependents` / `dependencies` / `path` / `edges` / `timeline` / `diff` / `history`) — apply uniformly.
- [ ] Acceptance test: REST + MCP responses for the same `dependents` call share JSON-equivalent shape under `result`.
- [ ] CHANGELOG **Changed** bullet (note: this is a REST-side breaking change; gate behind `?envelope=v2` from finding A if the topology endpoints are in scope of that opt-in, otherwise call out separately as breaking).

### F. Finding 13 — Broadcast stream `kind` discriminator (anchors §6)

`meho:feed:{tenant_id}` carries two event shapes: `event_kind: "agent_announcement"` (with `activity`/`target`/`phase`) and audit-derived (with `op_id`/`op_class`/`payload`). Consumers that switch on `op_id` silently nullify the agent-authored half.

- [ ] Every write to `meho:feed:{tenant_id}` carries a top-level `kind` field with values `"agent_announcement"` or `"operation"`.
- [ ] `broadcast.recent` + `broadcast.watch` consumers normalize to a `kind`-keyed envelope on read; no `event_kind`-vs-`op_id` inference.
- [ ] Existing rows without `kind` (pre-migration) read as `kind: "operation"` (the audit-derived majority shape; backwards-compatible inference for the historical window).
- [ ] Acceptance test: write one of each shape; assert both surface under `broadcast.recent` with correct `kind` discrimination.
- [ ] CHANGELOG **Changed** bullet citing consumer signal.

### G. Finding 15 — `?since=` docs↔impl agreement (anchors §8)

MCP `broadcast.recent`'s `since` parameter description promises ISO-8601 + Valkey-id; impl rejects ISO with `invalid_cursor`.

- [ ] **Pick one and apply:** either (a) extend impl to accept ISO `since=2026-05-25T10:00:00Z` (recommended — operators type timestamps, not Valkey ids), OR (b) update tool descriptions + OpenAPI parameter docs to say "Valkey stream cursor only; use `since=$` for tail" and remove the ISO promise.
- [ ] If (a): parser accepts ISO + Valkey-id; tool description names both.
- [ ] If (b): docs purge every "or ISO" mention.
- [ ] Acceptance test: `broadcast.recent` with both forms succeeds (a) or fails-loud with structured 422 (b).
- [ ] CHANGELOG bullet.

### H. Finding 22 — vmware catalog `version` vs spec `info.version` (anchors §9)

vmware catalog registers as `vmware-rest-9.0`; explicit-quadruple ingest with `version="9.0.0.0"` (matching spec `info.version`) produces a separate `vmware-rest-9.0.0.0` connector_id with disjoint catalog. Two connector_ids for one logical connector.

- [ ] Pick one per §9: either (a) reconcile the catalog `version="9.0"` ↔ spec `info.version="9.0.0.0"` so catalog ingest and explicit-quadruple ingest land on the same connector_id, OR (b) catalog declares `spec_info_versions_compatible: ["9.0.x"]` so the validator accepts the divergence (companion to G0.16-T5 for gh/3).
- [ ] Acceptance test: ingest under both the catalog name and the explicit-quadruple shape land on the same connector_id (under either resolution).
- [ ] CHANGELOG bullet.

## Why one Task, not eight

These are eight small surface-shape mismatches against one rulebook. Filing them as eight individual Tasks creates ceremony for what's a single sweep of "walk the conventions doc, run each ⊂ violation through its fix". One Task with eight independent acceptance items lets the work be picked up in any order, partial-progress to be visible (item count complete), and the consumer-feedback rollup to reference one number on the next dogfood cycle.

## Definition of done

- [ ] All 8 finding acceptance items above are checked.
- [ ] Conventions doc §2-§9 each have at least one "code reference" link pointing at a now-conforming endpoint (the lived-in proof of the rule).
- [ ] CHANGELOG `[Unreleased]` carries 8 bullets (one per finding) before the v0.9 cut.
- [ ] Next RDC dogfood cycle's signal directory finds zero re-OPEN entries for findings 3/6/7/8/10/13/15/22.

## Source

- Consumer findings: `claude-rdc-hetzner-dc#771` Findings 3 + 6 + 7 + 8 + 10 + 13 + 15 + 22
- Rulebook: [`docs/codebase/api-shape-conventions.md`](../../docs/codebase/api-shape-conventions.md) (PR #1310)
- Sibling Initiative Tasks: T2 (#1304, gh-rest auth_model) + T4 (#1306, Vault OIDC) anchor §7 — probe↔dispatch agreement is enforced by them in parallel
- Related: T5 (#1307, gh/3 catalog label-vs-spec) anchors §9 for the gh-rest variant; this Task H handles the vmware variant

damir-topic:
## Summary

Lands all 8 sub-findings of the G0.16-T6 SEV-4 sweep — converging today's REST + MCP + CLI surfaces onto `docs/codebase/api-shape-conventions.md` (PR #1310). Each finding lands in its own commit so reviewers can walk the sweep one rule at a time.

Per-finding status — **8/8 landed**, with 2 partial deferrals flagged:

| Finding | Anchor §  | Status | Notes |
|---|---|---|---|
| **A** — `?envelope=v2` opt-in on list endpoints | §2 | **Partial — reference adoption on `/api/v1/targets`** | Helper module landed; 4 sister endpoints (conventions / audit / broadcast-overrides / connectors) + CLI/MCP sister-surface forwarding deferred to follow-up. Each remaining adoption is a 5-line patch. |
| **B** — Catalog ↔ TargetCreate `product` enum reconciliation | §3 | **Verified already aligned; structural regression test added** | The v0.7-era `"sddc"` vs `"sddc-manager"` mismatch was closed by intervening connector work; new test pins the convergence so future drift fails CI. |
| **C** — `preferred_impl_id` accepts versioned form on Create + Update | §3 | **Landed** | `TargetCreate.preferred_impl_id="nsx-rest-4.2"` now accepted; resolver matches both base + versioned forms; unknown-impl 422 lists both shapes. |
| **D** — List rows mirror detail field set | §5 | **Landed** | `TargetSummary` widened to mirror `Target` minus operator-authored `notes` / `extras`. Structural regression test guards future masking. |
| **E** — REST↔MCP envelope agreement on topology | §4 | **Partial — `dependents` + `dependencies` landed under `?envelope=v2`** | Sister endpoints (`path` / `edges` / `timeline` / `diff` / `history`) deferred — those already return typed dict envelopes; per-endpoint migration decisions still needed. |
| **F** — Broadcast stream top-level `kind` discriminator | §6 | **Landed** | `BroadcastEvent.kind="operation"`, `AgentAnnouncementEvent.kind="agent_announcement"`; consumers switch on top-level `kind` first, fall back to historical `event_kind`; pre-migration entries infer `operation`. |
| **G** — `/api/v1/feed?since=` accepts ISO-8601 | §8 | **Landed** | REST feed now mirrors the MCP `broadcast.recent` ISO acceptance (which already existed); bare dates still rejected as likely-typos. |
| **H** — vmware catalog `spec_info_versions_compatible` | §9 | **Landed** | New catalog field + matching helper + vmware entry declaring `["9.0.x"]`; pairs with T5 #1307 for the gh-rest variant where the divergence is load-bearing. |

The shared `_envelope.py` helper module makes the deferred Finding A / E sister-endpoint adoptions cheap follow-ups.

## Test plan

- [x] `ruff format --check src/ tests/` — clean (788 files)
- [x] `mypy` on every touched file — `Success: no issues found in 11 source files`
- [x] Backend unit tests across the 10 most-relevant files — **340 passed, 2 skipped**:
  - `test_api_v1_targets.py` — Findings C / D / A
  - `test_api_v1_topology.py` — Finding E
  - `test_api_v1_feed.py` — Finding G
  - `test_broadcast_history.py` — Finding F
  - `test_operations_ingest_catalog.py` — Findings B / H
  - `test_connectors_resolver.py` — Finding C
  - `test_targets_schemas.py` / `test_targets_resolver.py` — Finding D
  - `test_api_v1_broadcast_overrides.py` — Finding F regression coverage
- [x] DCO sign-off on every commit
- [x] CHANGELOG `[Unreleased]` carries 8 bullets, one per finding

## Out of scope (deferred follow-ups)

- Finding A: 4 sister endpoints (`conventions` / `audit/my-recent` / `broadcast/overrides` / `connectors`) + CLI/MCP forwarding of the `envelope=v2` query parameter.
- Finding E: 5 sister topology endpoints (`path` / `edges` / `timeline` / `diff` / `history`).
- Finding F: pre-migration data-migration sweep on `meho:feed:{tenant_id}` streams (the attribute-default fallback covers reads structurally without it; a one-off `XADD`-rewrite to backfill `kind` on historical entries was not in scope).
- Finding H: validator-side application of `spec_info_versions_compatible` to the gh-rest ingest path — that's T5 #1307's load-bearing surface; the helper here is ready to consume.

Closes #1312

## Release Notes

* **New Features**
  * Optional `?envelope=v2` query parameter for unified paginated response format on list endpoints
  * ISO-8601 timestamp support for feed endpoint `since` parameter
  * Catalog compatibility hints for connector version matching

* **Changed**
  * Target list responses now include expanded field set (version, timestamps, audit fields) matching detail view
  * Connector preference override now accepts versioned identifiers (e.g., `"impl-version"` form)

[Review Change Stack](https://app.coderabbit.ai/change-stack/evoila/meho/pull/1325?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


damir-topic:
@coderabbitai review

backend/src/meho_backplane/api/v1/targets.py
```diff
@@ -394,8 +420,29 @@ def _registered_impl_ids() -> set[str]:
...
+    when only one version is registered.
     """
-    return {impl_id for (_product, _version, impl_id) in all_connectors_v2() if impl_id}
+    base_ids = {impl_id for (_product, _version, impl_id) in all_connectors_v2() if impl_id}
+    versioned_ids = {
+        f"{impl_id}-{version}"
+        for (_product, version, impl_id) in all_connectors_v2()
+        if impl_id and version
+    }
+    return base_ids | versioned_ids
```
damir-topic, line 440: **[🛑 B1]** — `preferred_impl_id` validation builds a global allowlist across all connector products, reopening the silent-ignore foot-gun.

**What:** `_registered_impl_ids()` reads every `(_, version, impl_id)` triple from `all_connectors_v2()` without filtering by product. POST at L794 and PATCH at L914 both call it without passing `body.product` / the effective patched product.

**Why it matters:** A `product="k8s"` target with `preferred_impl_id="vmware-rest-9.0"` passes this validator cleanly. At dispatch time the resolver's tie-break ladder never sees `vmware-rest-9.0` as a candidate for k8s, so it silently ignores the override and falls through to the default tie-break — the exact "operator believed they pinned an impl, every dispatch ignored it" UX foot-gun G0.15-T6 #1215 was created to close. Finding C made this strictly worse by widening the accepted shape (versioned form) without scoping by product.

**Required change:** make `_registered_impl_ids` product-aware:

```python
def _registered_impl_ids(product: str | None = None) -> set[str]:
    triples = [
        (p, v, i) for (p, v, i) in all_connectors_v2()
        if i and (product is None or p == product)
    ]
    base_ids = {i for (_, _, i) in triples}
    versioned_ids = {f"{i}-{v}" for (_, v, i) in triples if v}
    return base_ids | versioned_ids
```

Then in `create_target` pass `body.product`; in `update_target` pass the effective patched product (`new_product if new_product is not None else t.product`). Surface the scoped allowlist in the 422 detail so the operator's diagnostic names only the impls valid for *their* product. (Also flagged by CodeRabbit.)

backend/src/meho_backplane/operations/ingest/catalog.py
```diff
@@ -99,9 +99,70 @@ class ConnectorSpecEntry(BaseModel):
...
+    #: matches every ``info.version`` with release tuple
+    #: ``(9, 0, ...)``). Bare versions (no ``.x`` suffix) are treated
+    #: as PEP 440 exact-or-prefix.
+    spec_info_versions_compatible: tuple[str, ...] | None = None
     sha256: str | None = Field(default=None, max_length=64)
     notes: str = Field(default="", max_length=2048)
 
+    @field_validator("spec_info_versions_compatible")
+    @classmethod
+    def _compatibility_specifiers_well_formed(
```
damir-topic: **[🛑 B2]** — Sibling-PR collision with #1324 on `spec_info_versions_compatible`.

**What:** PR #1324 (G0.16-T5, gh/3 catalog drift) also adds `spec_info_versions_compatible: tuple[str, ...] | None` to `ConnectorSpecEntry`, AND modifies the same `catalog.yaml` file under the gh entry (`["1.x.x"]`). This PR adds the field under vmware (`["9.0.x"]`). Two issues:

1. **catalog.py merge conflict.** The two PRs add adjacent field-validator methods with different names (`_compatibility_specifiers_well_formed` here vs `_compatibility_patterns_are_parseable` in #1324) and incompatible parsing grammars.
2. **Runtime mismatch on the matcher.** This PR's `spec_info_version_matches_compatibility_specifier` only handles `.x` globs and bare PEP 440 versions. #1324's helper handles globs AND PEP 440 specifier sets (`">=1.0,<2.0"`, `"~=1.4"`) via `_compatibility_pattern_to_specifier`. If #1324 lands first and a future catalog row adopts a specifier set, this PR's matcher returns `False` silently.

**Why it matters:** Whichever PR merges second will hit a guaranteed merge conflict on `backend/src/meho_backplane/operations/ingest/catalog.py` (and on `catalog.yaml` because both modify adjacent entries). Even if the conflict is mechanically resolved, the matcher helper grammar diverges. The two PRs are in the same wave per the orchestrator's plan; this is a coordination defect, not a future maintenance hazard.

**Required change:** before this PR lands, coordinate with #1324:

- **Preferred:** rebase this PR on top of #1324 once it merges and reuse `_compatibility_pattern_to_specifier` + the field validator name from #1324 unchanged. #1324's grammar is a strict superset of this PR's (it accepts everything this PR accepts plus specifier sets), so the test surface here doesn't need to change.
- **Alternative:** land both behind one shared helper module (e.g. `meho_backplane/operations/ingest/_compat_spec.py`) introduced in a precursor PR, then both PRs import it.

The conventions-doc §9 citation pointing at this field's helper should resolve to one canonical implementation, not two.

backend/src/meho_backplane/api/v1/targets.py
```diff
@@ -457,7 +529,15 @@ async def list_targets(
...
-    return [_to_summary(t) for t in result.scalars().all()]
+    rows = list(result.scalars().all())
+    summaries = [_to_summary(t) for t in rows]
+    if envelope is None:
+        return summaries
+    next_cursor = rows[-1].name if len(rows) >= limit else None
+    return wrap_v2_envelope(
+        [s.model_dump(mode="json") for s in summaries],
+        next_cursor=next_cursor,
+    )
```
damir-topic, line 544: **[⚠️ M1]** — `list_targets` `?envelope=v2` pagination emits a non-null `next_cursor` on the terminal page.

**What:** L534-540 sets `next_cursor = rows[-1].name if len(rows) >= limit else None`. When the matching set has exactly `limit` rows remaining (i.e. the current page IS the last), the SQL returns `limit` rows, the `>=` comparison is true, and the cursor is non-null even though the next call would return `[]`.

**Why it matters:** Contradicts the docstring's explicit §2 contract ("`None` when the page exhausted the matching set so callers see 'no more pages' without inspecting list length") and forces every client paginating to issue one wasted round-trip per result set to discover exhaustion. The whole point of the `next_cursor: null` convention is to surface exhaustion in-band; this implementation loses that signal exactly when the operator hit the last page cleanly.

**Required change:** over-fetch by one and infer next-page presence from the extra row:

```python
stmt = stmt.order_by(TargetORM.name).limit(limit + 1)
result = await session.execute(stmt)
rows = list(result.scalars().all())
has_more = len(rows) > limit
page_rows = rows[:limit]
summaries = [_to_summary(t) for t in page_rows]
if envelope is None:
    return summaries
next_cursor = page_rows[-1].name if has_more else None
return wrap_v2_envelope(
    [s.model_dump(mode="json") for s in summaries],
    next_cursor=next_cursor,
)
```

Add a unit test that asserts `next_cursor is None` when the page exhausts the set. (Also flagged by CodeRabbit.)

backend/src/meho_backplane/targets/resolver.py
```diff
@@ -287,10 +287,31 @@ async def _near_misses(
 
 
 def _to_summary(t: TargetORM) -> TargetSummary:
```
damir-topic: **[💡 m1]** — `_to_summary` is now byte-for-byte duplicated across `api/v1/targets.py` and `targets/resolver.py`.

**What:** Both files define a `_to_summary(t: TargetORM) -> TargetSummary` that builds the same TargetSummary with identical fields, identical `tuple(t.aliases)` conversion, and identical `AuthModel(t.auth_model)` wrapping. The list endpoint uses the api/v1 copy; `resolve_target`'s near-miss surface uses the resolver copy.

**Why it matters:** Finding D was caused by silent drift between list (TargetSummary) and detail (Target) projections. Re-introducing a different shape of the same drift class (two projectors that must stay in lockstep) sets up the next regression cycle. Adding `extras` to TargetSummary in a future PR — or correcting an AuthModel coercion bug — has to be done in two places, and a reviewer who patches one will not notice the other.

**Required change:** extract to one shared helper:

```python
# meho_backplane/targets/_summary.py
def project_target_to_summary(t: TargetORM) -> TargetSummary:
    return TargetSummary(...)  # canonical projection
```

Both call sites import this. (Also flagged by CodeRabbit.)

damir-topic (commented):




**Anchor task:** #1312 — G0.16-T6 API shape conventions sweep (8 sub-findings)
**Surface area:** backend
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `d432377de47fc02ecc4e7c1f588849bf5d6e79e6`
**AI reviewer (CodeRabbit):** received (~13 min wait, 2 inline findings + 2 nitpicks)

### Acceptance criteria

- [x] **A — `?envelope=v2` opt-in on `/api/v1/targets`.** Helper module landed at `_envelope.py` (`EnvelopeVersion`, `ENVELOPE_QUERY`, `wrap_v2_envelope`); the reference adoption on `list_targets` works under the opt-in and the bare-list default is byte-preserved. 4 sister endpoints + CLI/MCP forwarding deferred — flagged in PR body and conventions doc §2.
- [x] **B — `product` enum reconciliation.** Diff is structural (regression test pins catalog ↔ TargetCreate convergence); no enum value churn needed.
- [x] **C — `preferred_impl_id` versioned form on Create.** TargetCreate widens to accept versioned form; resolver `_run_tie_break_ladder` normalises both shapes; 422 lists both. **See B1 below — the validation set is product-blind, which re-opens the silent-ignore foot-gun the validator was supposed to close.**
- [x] **D — TargetSummary mirrors Target minus `notes`/`extras`.** Structural regression guard in `test_targets_schemas.py`. Two `_to_summary` projectors now duplicate byte-for-byte (api/v1/targets.py + targets/resolver.py) — m1 below.
- [x] **E — REST↔MCP topology envelope agreement under `?envelope=v2`.** `dependents`/`dependencies` return `{kind, nodes}` matching MCP; default response preserves the v0.8.0 bare list. 5 sister endpoints deferred (per-endpoint migration decisions needed because they already return typed dict envelopes).
- [x] **F — Broadcast top-level `kind` discriminator.** `BroadcastEvent.kind="operation"`, `AgentAnnouncementEvent.kind="agent_announcement"`. Consumer switches on `kind` first, falls back to `event_kind`. Pre-migration entries infer from the field default (genuine pydantic default + parser fallthrough — confirmed: `_is_audit_event` reads `event.get("kind") or event.get("event_kind")`, then defaults to audit because the parser's `model_cls` selects `BroadcastEvent` whenever `discriminator != "agent_announcement"`).
- [x] **G — REST `/api/v1/feed?since=` accepts ISO-8601.** `_validate_cursor_or_400` + `_normalize_iso_to_cursor` mirror the MCP broadcast.recent dual-acceptance contract.
- [x] **H — vmware catalog `spec_info_versions_compatible`.** New field + helper + `["9.0.x"]` on vmware entry. **See B2 below — sibling-PR collision with #1324.**

### Project-board alignment
|---|---|---|---|---|---|
| #1312 | In Review | Closes | ✅ | partial (A, E flagged) | close on merge — partials explicit in PR body + conventions doc §2/§4 |

### Predicted vs. actual diff
- ✅ `_envelope.py`, `targets.py`, `topology.py`, `events.py`, `agent_events.py`, `broadcast/history.py`, `ui/routes/broadcast/history.py`, `feed.py`, `targets/schemas.py`, `targets/resolver.py`, `connectors/resolver.py`, `operations/ingest/catalog.py`, `operations/ingest/catalog.yaml`, `mcp/tools/broadcast.py`, `api-shape-conventions.md`, CHANGELOG.md + 7 test files — all expected and consistent with the 8-finding scope.
- ⚠️ `_registered_impl_ids` and `list_targets` pagination — see inline.
- ❓ none.
- 🚫 none.
- 🕳️ none — deferred items disclosed.

### Integration verification
- [x] `ruff format --check src/ tests/` — 779 files already formatted
- [⚠️ not run] `pytest` — venv module-import path failed inside the worktree (`No module named 'meho_backplane.settings'`); test-runner config differs from the parent checkout. Author reports 340 passed / 2 skipped across 10 target modules; reasoning-only review of the test files confirms coverage (`test_list_targets_envelope_v2_returns_unified_shape`, `test_dependents_envelope_v2_returns_kind_nodes_shape`, `test_parse_entry_dispatches_on_top_level_kind`, `test_parse_entry_legacy_audit_event_infers_operation`, `test_target_summary_field_set_superset_of_target`). CI's `Python (ruff + mypy + pytest)` is **IN_PROGRESS** at review time.
- [⚠️ not run] migrations replay — no schema change in this PR (only response-shape opt-ins + Pydantic schemas + catalog YAML).
- [⚠️ not run] `docker compose up` — out of scope for a shape-sweep PR.
- [⚠️ not run] UI flow — `ui/routes/broadcast/history.py` change is parser-only (`_is_audit_event` discriminator); no template change.

For `/auto-implement-issue --continue-pr`:

- 🛑 **B1** — `_registered_impl_ids()` builds a global allowlist across all connector products; a `k8s` target validates `preferred_impl_id="vmware-rest-9.0"` cleanly and the resolver silently ignores it at dispatch time, re-opening the exact foot-gun G0.15-T6 #1215 closed. Required change: pass `body.product` / patched product into the helper, filter `all_connectors_v2()` by product before building the base/versioned sets. (Also flagged by CodeRabbit, inline at L445.)

- 🛑 **B2** — Sibling-PR collision with #1324 on `spec_info_versions_compatible`. Both PRs define the same field on `ConnectorSpecEntry` AND modify the SAME catalog.yaml block (this PR adds the field on vmware; #1324 adds it on gh). Field validators are incompatible: this PR's `_compatibility_specifiers_well_formed` accepts globs (`"9.0.x"`) and bare PEP 440 (`"9.0.3"`); #1324's `_compatibility_patterns_are_parseable` accepts globs AND specifier sets (`">=1.0,<2.0"`, `"~=1.4"`) via a richer parser; this PR's matcher helper `spec_info_version_matches_compatibility_specifier` does NOT recognise specifier sets at all. Whichever PR merges second will face a guaranteed merge conflict on `catalog.py` and a runtime mismatch on `catalog.yaml` (gh row uses `"1.x.x"` glob which this PR's helper accepts, but the helper is the LESS-capable one). Required change: coordinate with #1324 — either rebase to inherit #1324's helper unchanged (preferred — #1324's grammar is a strict superset), or land both with one shared canonical helper module before merging either.

- ⚠️ **M1** — `list_targets` pagination `next_cursor` is set whenever `len(rows) >= limit`. On the terminal page where row count is exactly `limit`, the cursor is non-null even though the next call returns []. Forces every client to make one wasted round-trip to discover exhaustion, and contradicts the §2 contract documented in the docstring ("`None` when the page exhausted the matching set"). Required change: `limit + 1` over-fetch, slice to `limit`, emit `next_cursor` only when the extra row exists. (Also flagged by CodeRabbit, inline at L540.)

- 💡 **m1** — `_to_summary` is now byte-for-byte duplicated in `api/v1/targets.py` and `targets/resolver.py`. Finding D itself was caused by silent drift between list and detail surfaces; two independent projectors invite the same regression. Required change: extract a shared `project_target_to_summary(t: TargetORM) -> TargetSummary` helper (e.g. `meho_backplane/targets/_summary.py`) and have both call sites use it. (Also flagged by CodeRabbit, nit.)

- Conventions doc §8 mixes `_normalise_since` (British) and `_normalize_iso_to_cursor` (American). The function name in `broadcast/history.py` is `_iso8601_to_min_cursor`, not `_normalise_since` — the latter doesn't exist in code, suggesting a stale planning-doc reference. Pick the canonical name + spelling in a follow-up doc fixup. (CodeRabbit nit.)
- The 4 deferred Finding A sister endpoints (`conventions`, `audit/my-recent`, `broadcast/overrides`, `connectors`) are not yet filed as follow-up Tasks visible from the issue tree — the conventions doc text says "ship in a follow-up Task" but there's no `#NNNN` reference. Filing them as a single follow-up Initiative or Task before merge would make the deferral discoverable from project 19.
- Finding F's "pre-migration entries infer `operation`" path is implemented via two layers (field default + parser fallthrough). The parser logic (`broadcast/history.py:363` and `ui/routes/broadcast/history.py:151`) is genuinely fail-open — verified literal, not a fragile guess. No migration script needed for in-flight stream data, since `MAXLEN ~` will trim pre-migration entries within retention window.

- Adopted: 3 (B1 product-aware impl_ids, M1 pagination cursor, m1 _to_summary duplication — cited inline)
- Disputed: 0
- Out of scope: 1 (normalize/normalise spelling — adjacent improvement)

### Docs
- [x] `docs/codebase/api-shape-conventions.md` — updated with §2/§3/§4/§5/§6/§8/§9 code references pointing at the now-conforming endpoints + helpers + tests. The convention doc is now the lived-in proof of the rule per the §10 framing.

- 9-commit granularity (one logical change per finding + a closing format-only commit) makes the sweep walkable one rule at a time.
- The `?envelope=v2` opt-in design preserves v0.8.0 byte-shape on the default path — verified by inspection: `if envelope is None: return summaries` on targets, `if envelope is None: return nodes` on topology. No client breaks.
- Finding F's discriminator design is genuinely backward-compatible: pydantic default on `kind`, parser fallthrough on `kind or event_kind`, no in-flight data migration needed. `AgentAnnouncementEvent.kind` is `Literal["agent_announcement"]` and the historical `event_kind` alias is retained — both sides round-trip cleanly.
- Conventions doc updates carry concrete `code reference:` blocks naming the test that pins each invariant — the convention is bolted to executable proof, not prose alone.

damir-topic:
@coderabbitai review

PR #1340, opened 2026-05-29
feat(runbooks): G12.3-T3 run service layer — start/next/abort/reassign/list + post-completion check

damir-topic:
## Summary

- New `runbooks/run_service.py` module exposing `RunbookRunService` with six async methods (`start_run`, `next_step`, `abort_run`, `reassign_run`, `list_runs`, `can_show_template_post_completion`) plus five typed exceptions, ties together T1 schemas (#1300), T2 engine (#1301), the G12.1-T2 dispatcher contextvars (#1294), and the G12.2-T2 template service (#1296).
- Load-bearing: this is the **only** module that writes to `runbook_runs` and `runbook_run_step_states`; every state-machine transition lives in one diff for review.
- Single-assignee enforcement on `next_step` is unconditional (admin-non-assignees get the same 403); `abort_run` permits assignee or admin (via `caller_is_admin`); `reassign_run` is service-role-agnostic (the admin gate is the route's job per T5/T6).
- `operation_call` verify dispatches via `call_operation()` with `run_id_var`/`step_id_var` bound — every dispatched call's audit row carries the run/step correlation columns. `abort_run` writes its own audit row directly with `path='runbook.abort'`, reason in payload, columns populated.

## Test plan

- [x] `uv run pytest backend/tests/test_runbooks_run_service.py -v` — 27 passed (5 start_run, 4 next_step happy + 5 refusals, 2 audit-correlation, 3 abort, 2 reassign, 3 list_runs, 3 post-completion predicate).
- [x] `uv run pytest backend/tests/test_runbooks_engine.py backend/tests/test_runbooks_runs_schemas.py -v` — 34 passed, no T1/T2 regression.
- [x] `uv run ruff check backend/src/meho_backplane/runbooks/` — clean.
- [x] `uv run ruff format --check backend/src/meho_backplane/runbooks/ backend/tests/test_runbooks_run_service.py` — clean.
- [x] `uv run mypy backend/src/meho_backplane/runbooks/run_service.py --ignore-missing-imports` — clean.

## Out of scope

- REST routes (T5 / #1311) and MCP tools (T6 / #1313) wrap this service.
- Post-completion `show_template` exception wiring (T4 / #1309) calls `can_show_template_post_completion()` from the route boundary.
- Architecture doc (T7 / #1314).
- `validate_substitutions` refactor to share T2's regex constants (recorded in the issue body as opportunistic, deferred to keep this PR scoped).

Closes #1308

## Summary by CodeRabbit

* **New Features**
[...]

PR #1377, opened 2026-05-31
fix(runbooks): G12.3-T3 follow-up — release DB session across verify dispatch + preserve falsy forensics

### LINKED ISSUE #1352: Task: G12.3-T3 follow-ups — run_service.py next_step DB session lifetime + audit forensics falsy collapse
Parent initiative: #1198
Parent goal: #1195

## Problem

PR #1340 (G12.3-T3, the run-service layer) shipped two adjacent improvements that the auto-review approved as non-blocking but flagged for follow-up. Both live in [`backend/src/meho_backplane/runbooks/run_service.py`](backend/src/meho_backplane/runbooks/run_service.py) and share the verify-dispatch path inside `RunbookRunService.next_step`. Folding them into one Task because they're literally one PR's worth of work and they overlap in the code being read.

**Finding 1 — outer DB session is held across the external connector dispatch.**

`RunbookRunService.next_step` ([`run_service.py:343-454`](backend/src/meho_backplane/runbooks/run_service.py#L343-L454)) opens a session at [`run_service.py:385`](backend/src/meho_backplane/runbooks/run_service.py#L385) with `async with sessionmaker() as session:`, does its pre-dispatch reads (`_require_run_assignee`, `_load_pinned_template_body`, `_load_step_states`), then calls `_resolve_verify_response` which — for `verify.type='operation_call'` — invokes `call_operation()` at [`run_service.py:910`](backend/src/meho_backplane/runbooks/run_service.py#L910) while the outer session is still checked out. After the dispatch returns, `_apply_outcome` writes the post-dispatch state on the same session.

SQLAlchemy 2.0's `AsyncSession` acquires a pooled connection on first execute and **holds it for the lifetime of the `async with` context** — external awaits inside the context do not release it back to the pool. (Confirmed via the SQLAlchemy 2.0 async docs; see References.) `call_operation` itself opens its own session(s) for audit writes ([`meta_tools.py:315-316`](backend/src/meho_backplane/operations/meta_tools.py#L315-L316), [`meta_tools.py:427-428`](backend/src/meho_backplane/operations/meta_tools.py#L427-L428)), so during a verify dispatch the pool serves **at least two connections per in-flight runbook step**, with the outer connection idle for the full duration of the external call.

With `DATABASE_POOL_SIZE` capped (defaults are typical for asyncpg + SQLAlchemy at 5-20), 10-50 concurrent in-flight steps + multi-second connector latencies exhaust the pool. The reviewer's M1 in [PR #1340 iter-1](https://github.com/evoila/meho/issues/1198#issuecomment-4579680885) framed it as connection-pinning under PG pooling.

**Finding 2 — `call_result.get("result") or {}` silently collapses falsy payloads.**

In the same `_resolve_verify_response` method, [`run_service.py:915`](backend/src/meho_backplane/runbooks/run_service.py#L915) extracts the dispatched call's payload for `OperationCallVerifyResponse.actual` (which the engine then re-uses to write `runbook_run_step_states.verify_response` and feeds into the audit-write trail):

```python
actual = call_result.get("result") or {}
```

The `or {}` short-circuits on any falsy result — `[]`, `0`, `""`, `False`. The lines immediately below ([`run_service.py:916-922`](backend/src/meho_backplane/runbooks/run_service.py#L916-L922)) handle the non-dict case explicitly by wrapping (`actual = {"result": actual}`) for forensics — but the `or {}` step has already destroyed the value by then. A typed handler returning `result: []` (empty list — a perfectly valid "no matches" answer from a list-shaped op) becomes indistinguishable in the persisted `verify_response.actual` from a missing `result` key.

This is the M3 finding from [PR #1340 iter-1](https://github.com/evoila/meho/issues/1198#issuecomment-4579680885). It doesn't break the engine's `matched` logic (the engine recomputes against `expect` after this method returns) but it does break the forensics promise of the surrounding comment ("Wrap the value so the persisted row still carries the raw payload for forensics" — when the raw payload is `[]`, the wrap step never fires because the `or {}` already replaced it with `{}`).

## Current state

- `next_step` ([`run_service.py:343-454`](backend/src/meho_backplane/runbooks/run_service.py#L343-L454)) opens one session at line 385 and keeps it open through `_resolve_verify_response` (line 400) and `_apply_outcome` (line 416). No commit/release happens between the pre-dispatch reads and the dispatch.
- `_resolve_verify_response` ([`run_service.py:880-927`](backend/src/meho_backplane/runbooks/run_service.py#L880-L927)) doesn't take `session` — it doesn't read or write the DB itself (the connector-id resolution at line 892 opens its own session via `_resolve_connector_id_for_op`). So the outer session is held purely as a hand-off vehicle for `_apply_outcome` post-dispatch.
- `run_id_var` / `step_id_var` ([`run_service.py:907-913`](backend/src/meho_backplane/runbooks/run_service.py#L907-L913)) are set/reset around the dispatch via standard `contextvars.Token`-based try/finally. These are `contextvars` and survive across session boundaries (they're task-local, not session-bound).
- `_apply_outcome` writes the new step state, the run state (completed if applicable), and commits the outer session at [`run_service.py:454`](backend/src/meho_backplane/runbooks/run_service.py#L454).
- No precedent in `backend/src/meho_backplane/` currently splits a service-layer method into "pre-dispatch session + dispatch + post-dispatch session". This Task sets the precedent.
- Engine and substitution modules are unchanged (no DB touch); the change is contained to `run_service.py`.

### Finding 1 — release the outer connection before the external dispatch

Split `next_step` into a pre-dispatch session block, a sessionless dispatch, and a post-dispatch session block. Sketch:

```python
async def next_step(self, tenant_id, operator_sub, run_id, request):
    sessionmaker = get_sessionmaker()

    # Phase A — read everything we need under session A
    async with sessionmaker() as session_a:
        run = await self._require_run_assignee(session_a, tenant_id, run_id, operator_sub)
        self._refuse_if_terminal(run)
        template_body = await self._load_pinned_template_body(session_a, run)
        step_states = await _load_step_states(session_a, run_id)
        current_step_id = self._current_step_id_or_raise(run, step_states)
        current_state = step_states[current_step_id].state
        if current_state == _STEP_FAILED:
            raise PreviousStepFailedError(...)
        current_step = _find_step(template_body, current_step_id)

    # Phase B — dispatch with NO session checked out
    verify_response = await self._resolve_verify_response(
        run=run, current_step=current_step, request_verify=request.verify_response,
    )

    # Phase C — write the outcome under a fresh session B
    async with sessionmaker() as session_b:
        # Re-fetch the run + step_states inside session_b so SQLAlchemy
        # doesn't fight us over identity-map ownership across the gap.
        run = await self._require_run_assignee(session_b, tenant_id, run_id, operator_sub)
        step_states = await _load_step_states(session_b, run_id)
        # Re-check terminal/failed state in case a sibling caller raced us
        # while we were dispatching (the only legal races are abort_run
        # or reassign_run from a TENANT_ADMIN).
        self._refuse_if_terminal(run)
        if step_states[current_step_id].state == _STEP_FAILED:
            raise PreviousStepFailedError(...)
        outcome = advance(template_body, current_step_id, ...)
        response = await self._apply_outcome(...)
        await session_b.commit()
    return response
```

Critical points:

- **Two `async with sessionmaker()` blocks**, not one. Between them, no pooled connection is held.
- **`contextvars.Token` (`run_id_var.set` / `step_id_var.set`) stays inside `_resolve_verify_response`** as it does today — context binding is task-local, independent of session lifetime.
- **Re-validate run + current step state at the start of session B.** Between A and B the run could have been aborted (`abort_run`) or reassigned (`reassign_run`) by a TENANT_ADMIN. The single-assignee invariant must still hold at write time. Surface the same exceptions (`NotRunAssigneeError` / `RunAlreadyTerminalError`) if the gap was racey.
- **`template_body` and `current_step` are stable across the gap** (immutable: the run is pinned to a specific template version at start_run; the template's `steps` are frozen as a JSONB blob). Safe to re-use without re-loading.
- **`step_states[current_step_id]` is also stable for the assignee** — only the assignee can advance it, and they're blocked here. The only state that can change is the terminal flip via `abort_run`, which the post-dispatch re-check catches.

### Finding 2 — use `dict.get("result", {})` not `... or {}`

[`run_service.py:915`](backend/src/meho_backplane/runbooks/run_service.py#L915), one-line change:

```python
# Before
actual = call_result.get("result") or {}

# After
actual = call_result.get("result", {})
```

The `, {}` default fires only when the `"result"` key is absent — preserving falsy-but-meaningful payloads (`[]`, `0`, `""`, `False`) for the forensics wrap below. The non-dict wrap at lines 916-922 already handles those cases correctly once the value isn't pre-collapsed.

## Acceptance criteria

- [ ] `next_step` no longer holds an `AsyncSession` across the `call_operation` await. Verified by reading the diff: two `async with sessionmaker()` blocks (or one block that explicitly commits + closes before the dispatch and opens a fresh one after). Grep test: `git diff main -- backend/src/meho_backplane/runbooks/run_service.py` shows the dispatch call outside any `async with` session block.
- [ ] Race coverage: new test in [`backend/tests/test_runbooks_run_service.py`](backend/tests/test_runbooks_run_service.py) simulates `abort_run` landing between phase A and phase C of `next_step` (e.g. by patching `_resolve_verify_response` to fire a concurrent `abort_run` mid-dispatch). The test asserts `next_step` raises `RunAlreadyTerminalError` instead of overwriting the aborted state.
- [ ] Race coverage: new test simulates `reassign_run` landing between A and C. Asserts `next_step` raises `NotRunAssigneeError` rather than advancing on behalf of the previous assignee.
- [ ] Connection-pinning regression test: spin up a sessionmaker with `pool_size=1, max_overflow=0`, call `next_step` with an `operation_call` verify whose mocked dispatch holds for 500ms, and concurrently issue a second simple DB query. The second query must complete inside that 500ms (it would block if the pool were exhausted). Test belongs in the same module.
- [ ] [`run_service.py:915`](backend/src/meho_backplane/runbooks/run_service.py#L915) — `call_result.get("result") or {}` is replaced with `call_result.get("result", {})`. Verified by grep: `grep -n 'or {}' backend/src/meho_backplane/runbooks/run_service.py` returns no results in `_resolve_verify_response`.
- [ ] Falsy-result forensics test: a stub `call_operation` returning `{"result": []}` (and another returning `{"result": 0}`) — `next_step` calls it, then the test asserts `runbook_run_step_states.verify_response.actual` round-trips as `{"result": []}` / `{"result": 0}` (wrapped by the non-dict branch), not `{}`.
- [ ] CI green: `ruff check`, `ruff format --check`, `mypy backend/src/`, `pytest backend/tests/test_runbooks_run_service.py -v` all pass.
- [ ] No regression: the existing 27+ run-service tests, the 12 priming tests (G12.4), the 30+ REST/MCP route tests still pass.

## Out of scope

- The post-dispatch session B's terminal-state re-check is per-call — building a generalised "optimistic-concurrency wrapper" for service methods is a separate refactor.
- Renaming `_resolve_verify_response` to make its sessionless nature explicit (`_resolve_verify_response_outside_session`?) is a doc/clarity concern, not load-bearing for this fix.
- Audit-side audit of `or {}` patterns elsewhere in the codebase. Finding 2 is scoped to the verify-dispatch path; a broader sweep belongs to its own refactor Initiative if the team wants one.
- Backend 422-shape normalisation (the `rawNextResponse` shim in `cli/internal/cmd/runbook/next.go` works around a separate issue: `HTTPException(detail=str(exc))` not matching the OpenAPI Pydantic validation-error shape). Tracked in a separate Task.
- Connector-pool sizing or async pool tuning — orthogonal infra concern.

## References

- PR #1340 (G12.3-T3) iter-1 review (M1 + M3 findings): https://github.com/evoila/meho/issues/1198#issuecomment-4579680885
- Initiative #1198 (G12.3 Runbook run lifecycle) — parent.
- Goal #1195 (G12 Runbooks) — grandparent.
- SQLAlchemy 2.0 async session + connection lifetime: https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html (confirms `AsyncSession` holds the pooled connection for the lifetime of `async with`).
- SQLAlchemy AsyncIO pooling discussion: https://github.com/sqlalchemy/sqlalchemy/discussions/5980 (community confirmation that external awaits inside an open session block don't release the connection).
- `call_operation` sessionmaker usage: [`meta_tools.py:315-316`](backend/src/meho_backplane/operations/meta_tools.py#L315-L316), [`meta_tools.py:427-428`](backend/src/meho_backplane/operations/meta_tools.py#L427-L428) — `call_operation` opens its own sessions; the outer one is purely a hand-off vehicle.
- Pool sizing config: [`db/engine.py:33`](backend/src/meho_backplane/db/engine.py#L33), [`db/engine.py:157`](backend/src/meho_backplane/db/engine.py#L157) — `DATABASE_POOL_SIZE` env var.
- Python best practices — async I/O isolation discipline, no DB session held across HTTP calls (cite the codebase's existing async-session conventions in `kb/service.py` and `runbooks/service.py` — both follow the open-do-commit pattern, this Task extends it with a "release-during-external-call" variant).

damir-topic:
## Summary
- `RunbookRunService.next_step` held its outer `AsyncSession` across the external `call_operation` verify dispatch. SQLAlchemy 2.0 pins a session's checked-out connection for the lifetime of its `async with` block, so the outer connection sat idle through the whole (multi-second) connector call — and `call_operation` opens its own session(s) for audit writes, so each in-flight step consumed ≥2 pooled connections, exhausting a capped pool under concurrency (M1 from the #1340 iter-1 review).
- Split `next_step` into a read-only **session A** (`_load_next_step_inputs`), a **sessionless** verify dispatch, and a write **session B** (`_write_next_step_outcome`). Session B re-loads and re-validates the run + step states, so a `TENANT_ADMIN` `abort_run` (terminal flip) or `reassign_run` (assignee change) landing in the dispatch gap is caught — the single-assignee / no-mutate-terminal invariants hold at write time, not just at call start.
- Fixed the forensics collapse in `_resolve_verify_response`: `call_result.get("result") or {}` discarded any falsy payload (`[]`, `0`, `""`, `False`) before the non-dict wrap could preserve it, so a list-shaped op returning an empty "no matches" result persisted as `{}` instead of `{"result": []}`. Now `.get("result", {})` — only an absent key defaults (M3 from the #1340 iter-1 review).
- The verify dispatch's run/step `contextvars` binding stays inside `_resolve_verify_response` (task-local, session-independent) — unchanged.

## Test plan
- [x] `ruff format --check src/ tests/` — clean (808 files)
- [x] `mypy src/` — clean (394 files)
- [x] `pytest tests/test_runbooks_run_service.py` — 31 passed (27 existing + 4 new)
- [x] No `AsyncSession` held across the dispatch — verified by diff: the `_resolve_verify_response` call sits in the `next_step` orchestrator between two separate session-scoped helpers; `grep 'or {}'` returns nothing in the module.
- [x] Race coverage — `test_next_step_aborted_during_dispatch_refuses` (abort mid-dispatch → `RunAlreadyTerminalError`, aborted state survives) and `test_next_step_reassigned_during_dispatch_refuses` (reassign mid-dispatch → `NotRunAssigneeError`, step not advanced).
- [x] Connection-pinning regression — `test_next_step_does_not_pin_connection_across_dispatch` uses a `pool_size=1, max_overflow=0` engine + a 500ms stub dispatch; a concurrent probe completes promptly. Confirmed it **fails** against the pre-split single-session shape (true regression guard).
- [x] Falsy-result forensics — `test_next_step_falsy_operation_result_preserved_in_forensics` asserts `verify_response.actual` round-trips as `{"result": []}` / `{"result": 0}`. Confirmed it **fails** against the old `or {}` code.
- [x] No regression — `test_runbook_runs_api.py`, `test_mcp_tools_runbook_runs.py`, `test_runbooks_priming.py`, `test_audit_runbook_correlation.py`, `test_runbooks_engine.py` — 88 passed.

## Out of scope
- A generalised optimistic-concurrency wrapper for service methods (the session-B re-check is per-call here), renaming `_resolve_verify_response`, and a codebase-wide sweep for other `or {}` patterns — all deferred per the issue body.
- No OpenAPI / route / schema surface changed (service-layer only), so no CLI snapshot regen needed.

Closes #1352

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of falsy operation results (e.g., empty arrays, zero) being incorrectly discarded from forensic records.

* **Refactor**
  * Improved internal session lifecycle management to prevent database connection retention during external dispatch operations.
  * Enhanced state validation and re-checks before persisting changes.

* **Tests**
  * Added race condition tests for admin abort and reassignment scenarios during concurrent operations.
  * Added connection pooling regression tests to ensure resources are released properly.

### DIFF
--- a/backend/src/meho_backplane/runbooks/run_service.py
+++ b/backend/src/meho_backplane/runbooks/run_service.py
@@ -102,6 +102,7 @@ from __future__ import annotations
 
 import time
 import uuid
+from dataclasses import dataclass
 from datetime import UTC, datetime
 from decimal import Decimal
 from typing import Any, Literal
@@ -238,6 +239,28 @@ class MissingParamsError(ValueError):
     catching the gap at start time gives the operator a typed error
     before the run row lands.
     """
+
+
+@dataclass(frozen=True, slots=True)
+class _NextStepInputs:
+    """Pre-dispatch reads :meth:`RunbookRunService.next_step` carries across the gap.
+
+    Captured under the read-only session A and consumed by the
+    sessionless verify dispatch + the write-phase session B. The
+    ``template_body`` / ``current_step`` are pinned at start_run and
+    immutable for the run, so they remain valid after session A closes;
+    ``run`` is re-loaded fresh inside session B (the captured instance
+    is detached once its session closes and is only used for the
+    contextvar-binding read in the dispatch phase).
+    """
+
+    run: RunbookRun
+    template_body: RunbookTemplateBody
+    current_step: Step
+    current_step_id: str
+    current_state: str
+    run_target: str
+    run_params: dict[str, Any]
 
 
 class RunbookRunService:
@@ -380,14 +403,57 @@ class RunbookRunService:
           its ``verify_response``. Raises :class:`PreviousStepFailedError`
           so the caller's next move is :meth:`abort_run` rather than a
           spurious retry on a step the state machine no longer accepts.
+
+        Session lifetime: the method runs in three phases — a read-only
+        session A (:meth:`_load_next_step_inputs`) for the pre-dispatch
+        reads, a **sessionless** verify dispatch (so no pooled connection
+        is pinned across the external ``operation_call``), and a write
+        session B (:meth:`_write_next_step_outcome`) for the outcome. The
+        run + step states are re-loaded and re-validated at the start of
+        session B because a TENANT_ADMIN could have raced the dispatch
+        with :meth:`abort_run` (→ :class:`RunAlreadyTerminalError`) or
+        :meth:`reassign_run` (→ :class:`NotRunAssigneeError`); the
+        single-assignee / no-mutate-terminal invariants must hold at the
+        moment of the write, not just at the start of the call.
+        """
+        inputs = await self._load_next_step_inputs(tenant_id, operator_sub, run_id)
+
+        # Dispatch the verify with NO session checked out.
+        # ``template_body`` / ``current_step`` are pinned at start_run and
+        # immutable for the run, so they survive the gap; the run/step
+        # contextvar binding inside ``_resolve_verify_response`` is
+        # task-local and independent of session lifetime.
+        verify_response = await self._resolve_verify_response(
+            run=inputs.run,
+            current_step=inputs.current_step,
+            request_verify=request.verify_response,
+        )
+
+        return await self._write_next_step_outcome(
+            tenant_id, operator_sub, run_id, inputs, verify_response
+        )
+
+    async def _load_next_step_inputs(
+        self,
+        tenant_id: uuid.UUID,
+        operator_sub: str,
+        run_id: uuid.UUID,
+    ) -> _NextStepInputs:
+        """Phase A: read everything the dispatch + write need, under session A.
+
+        Releases the pooled connection on block exit — SQLAlchemy 2.0's
+        :class:`AsyncSession` holds its checked-out connection for the
+        whole ``async with`` block, so awaiting the connector dispatch
+        inside it would pin the connection for the full (multi-second)
+        call and starve the pool under load.
         """
         sessionmaker = get_sessionmaker()
-        async with sessionmaker() as session:
-            run = await self._require_run_assignee(session, tenant_id, run_id, operator_sub)
+        async with sessionmaker() as session_a:
+            run = await self._require_run_assignee(session_a, tenant_id, run_id, operator_sub)
             self._refuse_if_terminal(run)
 
-            template_body = await self._load_pinned_template_body(session, run)
-            step_states = await _load_step_states(session, run_id)
+            template_body = await self._load_pinned_template_body(session_a, run)
+            step_states = await _load_step_states(session_a, run_id)
             current_step_id = self._current_step_id_or_raise(run, step_states)
             current_state = step_states[current_step_id].state
 
@@ -396,34 +462,62 @@ class RunbookRunService:
                     f"previous step {current_step_id!r} is in 'failed' state; abort the run",
                 )
 
-            current_step = _find_step(template_body, current_step_id)
-            verify_response = await self._resolve_verify_response(
+            return _NextStepInputs(
                 run=run,
-                current_step=current_step,
-                request_verify=request.verify_response,
+                template_body=template_body,
+                current_step=_find_step(template_body, current_step_id),
+                current_step_id=current_step_id,
+                current_state=current_state,
+                run_target=run.target,
+                run_params=dict(run.params),
             )
 
-            now = datetime.now(UTC)
+    async def _write_next_step_outcome(
+        self,
+        tenant_id: uuid.UUID,
+        operator_sub: str,
+        run_id: uuid.UUID,
+        inputs: _NextStepInputs,
+        verify_response: VerifyResponse | None,
+    ) -> NextStepResponse:
+        """Phase C: advance + persist the outcome under a fresh session B.
+
+        Re-loads and re-validates the run + step states: a TENANT_ADMIN
+        could have raced the dispatch with :meth:`abort_run` (terminal
+        flip) or :meth:`reassign_run` (assignee change). The
+        single-assignee / no-mutate-terminal invariants must still hold
+        at write time, not just at the start of the call.
+        """
+        now = datetime.now(UTC)
+        sessionmaker = get_sessionmaker()
+        async with sessionmaker() as session_b:
+            run = await self._require_run_assignee(session_b, tenant_id, run_id, operator_sub)
+            self._refuse_if_terminal(run)
+            step_states = await _load_step_states(session_b, run_id)
+            if step_states[inputs.current_step_id].state == _STEP_FAILED:
+                raise PreviousStepFailedError(
+                    f"previous step {inputs.current_step_id!r} is in 'failed' state; abort the run",
+                )
+
             outcome = advance(
-                template_body,
-                current_step_id,
-                current_state,
-                target=run.target,
-                params=dict(run.params),
+                inputs.template_body,
+                inputs.current_step_id,
+                inputs.current_state,
+                target=inputs.run_target,
+                params=inputs.run_params,
                 verify_response=verify_response,
                 completed_at=now,
             )
-            response = await self._apply_outcome(
-                session=session,
+            return await self._apply_outcome(
+                session=session_b,
                 run=run,
-                template_body=template_body,
+                template_body=inputs.template_body,
                 step_states=step_states,
-                current_step_id=current_step_id,
+                current_step_id=inputs.current_step_id,
                 outcome=outcome,
                 now=now,
                 operator_sub=operator_sub,
             )
-        return response
 
     async def _apply_outcome(
         self,
@@ -912,7 +1006,7 @@ class RunbookRunService:
             step_id_var.reset(step_token)
             run_id_var.reset(run_token)
 
-        actual = call_result.get("result") or {}
+        actual = call_result.get("result", {})
         if not isinstance(actual, dict):
             # The verify's _matches() compares dict-shaped expect against
             # dict-shaped actual; a list/scalar result from a typed handler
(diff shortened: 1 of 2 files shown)


damir-topic:
@coderabbitai review

damir-topic:
@coderabbitai review

PR #1418, opened 2026-05-31
feat(ingest): wire production grouping LLM client at lifespan startup

### LINKED ISSUE #1386: Task: G3.17-T1 — Wire a production ingest LLM client at lifespan startup so --catalog ingest works on deployed backplanes (reuse the agent runtime's client)
Parent goal: #214
Parent initiative: #1407

On any deployed backplane, `meho connector ingest --catalog <product>/<version>` fails the grouping pass with a 503 — `default_llm_client_factory` raises unconditionally and nothing wires `set_llm_client_factory` at FastAPI lifespan startup, so the chassis ships with no production `LlmClient`. `settings.anthropic_api_key` only reaches the agent runtime, not the ingest path. The G0.18-T7 docstring cleanup made this honest (and correctly killed the dead #405 reference — that was the CLI-verb task, never an LLM adapter), which is genuinely appreciated. But the practical effect is that the typed connectors (vmware-rest, nsx, sddc-manager, and the gh L2 surface) stay "registered, zero dispatchable ops" on every real deploy, because their L2 operations only appear after ingest groups them.

The thing that makes this feel like a small fix rather than a big one: the agent runtime already constructs a working Anthropic client (`agent/run.py` / `agent/models.py`, with the `_split_model_id` prefix handling) — we've run agents end-to-end against it. The ingest pipeline just doesn't reuse it. So the ask is really "reuse the agent runtime's client for the spec-ingestion grouping pass" — wire `set_llm_client_factory` at lifespan startup from the same `settings.anthropic_api_key`, rather than standing up a separate adapter.

From a consumer's seat this is the single dependency gating everything else: it's the line between meho being a read/governance plane and meho being able to actually drive vCenter/NSX/SDDC operations. We're set up to be the first validator — the API key is already in our Vault, we have the connector specs ready to hand in as `--spec` URIs (the VCF-family catalog upstreams are HTML/templated, so the explicit `--spec` path is what we'd use anyway), and a live deploy to prove it against the moment it lands. Happy to test a branch.

Context: this is what the v0.8.x → v0.9.0 catalog-ingest findings keep landing on. Related: the G0.7 spec-ingestion pipeline (#389) and the G11 agent runtime that already has the working client (#800).

damir-topic:
## Summary

- Wire a production `LlmClient` for the spec-ingestion grouping pass at FastAPI lifespan startup, so non-dry-run `meho connector ingest --catalog <product>/<version>` works on deployed backplanes instead of failing closed with HTTP 503.
- New `build_anthropic_ingest_llm_client` **reuses the agent runtime's Anthropic construction** — the same `settings.anthropic_api_key` and the same `_split_model_id(settings.agent_default_model)` prefix handling — talking to the Messages API directly for the one-shot `system + user -> raw JSON` shape the grouping pass needs (vs. the pydantic-ai `Model` the agent loop uses). No second provider/config stack.
- Fail-closed preserved: a deploy with **no** `ANTHROPIC_API_KEY` still raises `LlmClientUnavailable` -> 503 when the factory is invoked.
- Fixes a latent gap on the **MCP** surface: `meho.connector.ingest` hardcoded `default_llm_client_factory` and would have 503'd even after lifespan wiring. It now reads the active factory via the new public `get_llm_client_factory()` accessor — the same holder REST/CLI use (the module holder's own docstring says it exists to keep these consumers in sync).
- Updated the now-stale "build-time-only / no production adapter ships" framing across code docstrings, the `composite_l2_missing` operator message, and the codebase / architecture / cross-repo docs (the section rename also fixed every inbound anchor).

## Design notes

- **Direct Anthropic, not the G11.5 resolver.** Ingest grouping is a build-time operator action with no per-tenant tier/egress context, and the resolver returns pydantic-ai `Model`s shaped for the agent tool-use loop, not `generate_json`. Routing ingest through the resolver (so air-gapped, no-Anthropic-key deploys can group) is called out as the explicit out-of-scope follow-up in `docs/codebase/spec-ingestion.md`.
- Best-practices packs referenced by the skill are not vendored in this repo; applied the surface's standard practices from first principles (mypy-strict typing, structlog without secrets, SDK-native retries, function-local lazy imports to avoid an ingest↔agent import cycle — mirroring `anthropic_backend_builder`).

## Test plan

- [x] `ruff check` + `ruff format --check` — clean (all changed files)
- [x] `mypy` — clean (5 source modules)
- [x] New `tests/test_operations_ingest_anthropic_client.py` — factory fail-closed without key, provider-prefix strip, bare-id passthrough, `generate_json` Messages-API arg mapping + text-block concatenation + empty-content, and the lifespan helper installing the production factory
- [x] Strengthened MCP test: `meho.connector.ingest` passes exactly the active lifespan-wired factory (would fail if it still hardcoded the default)
- [x] Existing 503 + `set_llm_client_factory` contract tests stay green
- [x] Broad sweep `-k "ingest or catalog or connector or lifespan or composite_l2"` — 2377 passed, 101 skipped, 0 failed (no cross-test pollution from the global wiring)
- [x] OpenAPI snapshot regenerated — unchanged (the `Depends` rename is not part of the schema)
- [ ] Live end-to-end grouping against the real model — manual, `ANTHROPIC_API_KEY`-gated canary (`MEHO_G07_CANARY_LIVE_LLM=1`); skips in sandbox/CI without a key

## Out of scope

- Routing the grouping pass through the G11.5 per-tenant model resolver (multi-backend, egress-aware) — needed for air-gapped deploys with no Anthropic key.

Closes #1386


damir-topic:
@coderabbitai review

Issue #823, opened 2026-05-21
Task: G11.3-T2 cron + one-off triggers fire agent runs; replica-safe (advisory lock / SKIP LOCKED), restart-durable


damir-topic:
Parent goal: #800
Parent initiative: #804

## Context

The two simplest, highest-value triggers: fire an agent run on a **cron** expression and **once at time T**. Must persist across restart and never double-fire across replicas.

## Current state

- Trigger model + substrate land in 11.3-T1. The replica-safe firing precedent is the topology scheduler's `pg_try_advisory_lock` non-blocking claim ([topology/scheduler.py](backend/src/meho_backplane/topology/scheduler.py)). Agent invocation entry: the AgentRun seam (11.1-T1 #808) + run record (11.1-T6 #813).
- **Autonomous auth + audit seam shipped by 11.2-T2 (#816):** `AgentInvoker.run_scheduled(name, inputs, *, agent_client_id, agent_client_secret)` ([agent/invocation.py](backend/src/meho_backplane/agent/invocation.py)) authenticates the agent via the `client_credentials` grant (`get_client_credentials_token`, [auth/agent_token.py](backend/src/meho_backplane/auth/agent_token.py)), runs it as the agent principal (`operator_sub`=agent, `actor_sub`=NULL, `trigger=SCHEDULED`), and enforces `entry.identity_ref == agent_client_id` (cross-agent launch refused). **This Task is its caller** — and owns the one piece it leaves open: where the agent client secret comes from (Vault vs a Keycloak admin `GET .../clients/{internal_id}/client-secret` via the T1 `KeycloakAdminClient`).

## Desired state

- A poller/loop (Option A: extend the asyncio loop; Option B: DBOS `@scheduled` + durable queue) that, each tick, claims due `scheduled_trigger` rows (cron `next_run_at` reached, or one-off `run_at` reached) **replica-safely** (advisory lock / `FOR UPDATE SKIP LOCKED`, or DBOS uniqueness), fires the run via `AgentInvoker.run_scheduled(...)` as the trigger's agent identity (the 11.2-T2 #816 seam — `client_credentials`, `operator_sub`=agent, `actor_sub`=NULL), advances `next_run_at` (cron) or marks the one-off done, and records `last_fired_at`. Catch-up/missed-run + clock-skew semantics handled (don't storm on restart).

## Acceptance criteria

- [ ] A cron trigger fires its agent run on schedule; a one-off fires once at T and never again.
[...]

PR #1489, opened 2026-06-03
fix(scheduler): source agent client_credentials from Vault, not pod env (#1478)

### LINKED ISSUE #1478: Scheduler sources agent client_credentials secret from a pod env var, not Vault — an API-registered agent isn't schedulable without an operator env wire-up + redeploy
Parent goal: #221
Parent initiative: #1484

## Problem

While dogfooding the v0.10.0 governance surface we tried the canonical autonomous pattern — register an agent principal over the API, bind an agent definition to it, schedule a one-off trigger, let it fire. The trigger never fired; every scheduler tick logged `scheduler_credentials_unresolved`.

Root cause: the scheduler resolves a scheduled agent's `client_credentials` secret from a **pod environment variable** derived from the identity_ref (default `MEHO_AGENT_SECRET_{client_id}`, upper-cased with every non-alphanumeric collapsed to `_` — so `agent:reporter` → `MEHO_AGENT_SECRET_AGENT_REPORTER`). The Vault path setting `scheduler_agent_vault_path_pattern` exists but is **defined-but-never-read**. So an agent you can fully *register and define* over the API/MCP is **not runnable** by the scheduler until an operator separately wires its secret into the pod env and redeploys — the registration path and the run path disagree about where the credential lives, which breaks the self-service / autonomous loop the agent-registration API otherwise promises.

This is the explicitly-deferred "swap to the Vault path" follow-up that the module docstring and the settings comment both promise; no issue tracks it today. **It is not a one-line resolver swap** — see Desired state.

## Current state (verified in code, v0.10.0)

- [`scheduler/credentials.py:119-139`](backend/src/meho_backplane/scheduler/credentials.py#L119-L139) — `resolve_agent_credentials` reads the secret from `os.environ` via `settings.scheduler_agent_secret_env_pattern.format(client_id=…).upper()`; empty/unset → `AgentCredentialsUnresolvedError` naming the expected env var. **No Vault read anywhere.**
- [`settings.py:981-986`](backend/src/meho_backplane/settings.py#L981-L986) — `scheduler_agent_secret_env_pattern` default `MEHO_AGENT_SECRET_{client_id}` is the active source; `scheduler_agent_vault_path_pattern` default `secret/data/agents/{client_id}/credentials` is defined right below, commented "ships configured but unused". Grep confirms it is only ever env-wired ([`settings.py:1315`](backend/src/meho_backplane/settings.py#L1315)) and never read by resolution logic.
- [`scheduler/loop.py:381-393`](backend/src/meho_backplane/scheduler/loop.py#L381-L393) — `_prepare_invocation` catches the error, logs warning `scheduler_credentials_unresolved`, returns `None`; the trigger row is left active and retried next tick (loud, non-destructive — not parked).
- [`auth/agent_principals.py:183-236`](backend/src/meho_backplane/auth/agent_principals.py#L183-L236) — `register()` creates the Keycloak `agent:<name>` client then persists only `keycloak_client_id` + `keycloak_internal_id`; **no client secret is captured**.
- [`auth/keycloak_admin.py:237-308`](backend/src/meho_backplane/auth/keycloak_admin.py#L237-L308) — `create_client` returns only the internal UUID parsed from the `Location` header; Keycloak's generated `client_credentials` secret is **discarded**.
- The only Vault accessor is `vault_client_for_operator(operator)` ([`vault.py`](backend/src/meho_backplane/vault.py)), which authenticates with a per-operator JWT; the scheduler is **operator-less** (`credentials.py:9-12`), so it has no Vault read identity today.
- Docs corroborate: [`docs/codebase/scheduler.md:104-130`](docs/codebase/scheduler.md#L104-L130) ("ships configured but unused in v0.2").

## Desired state

An agent registered + defined over the API is schedulable without a manual pod-env wire-up + redeploy. Because the secret is neither captured at registration nor readable by the scheduler today, this is a **two-part** change:

1. **Capture the secret at registration.** On `agent_principals.register`, fetch the generated client secret from Keycloak (Admin REST `GET /admin/realms/{realm}/clients/{id}/client-secret` → `CredentialRepresentation`, or `POST` to (re)generate) and persist it to Vault at `scheduler_agent_vault_path_pattern` (KV v2), reusing the existing Vault write conventions.
2. **Give the scheduler a Vault read identity.** Add a scheduler service-token (or AppRole) Vault auth path so `resolve_agent_credentials` can read `secret/data/agents/{client_id}/credentials` without a per-operator JWT, then switch resolution to Vault-first (env var kept as a documented fallback / break-glass).

## Acceptance criteria

- [ ] Registering an agent principal persists its Keycloak client secret to Vault at the `scheduler_agent_vault_path_pattern` location (integration test against a Vault + Keycloak fixture, à la #1098).
- [ ] `resolve_agent_credentials` reads the secret from Vault using a scheduler service identity (no per-operator JWT); env-var lookup remains only as a documented fallback.
- [ ] An agent registered + defined + one-off-triggered purely over the API fires on the next scheduler tick with **no** pod env var set and no redeploy (e2e).
- [ ] `scheduler_credentials_unresolved` still fires (loud, trigger-preserving) when neither Vault nor the fallback env var yields a secret.
- [ ] `docs/codebase/scheduler.md` updated to Vault-first resolution; the "ships configured but unused" comments removed.

## Out of scope

- Agent secret rotation / Vault lease renewal (separate follow-up).
- Changing the env-var fallback's naming/derivation.
- Non-agent (operator) scheduled work — this is specific to `agent:<name>` identities.

## References

- Origin: `evoila-bosnia/claude-rdc-hetzner-dc#841` (v0.10.0 RDC governance-surface dogfood).
- Keycloak Admin REST API — client-secret endpoint (`GET`/`POST /admin/realms/{realm}/clients/{id}/client-secret`, returns `CredentialRepresentation`): https://www.keycloak.org/docs-api/latest/rest-api/index.html
- OAuth 2.0 client credentials grant (RFC 6749 §4.4): https://datatracker.ietf.org/doc/html/rfc6749#section-4.4
- Closed origin: #823 (G11.3-T2 — introduced env-var sourcing, deferred the Vault path as a follow-up). Adjacent: #1098 (G11.2-T7 — live-Keycloak `client_credentials` integration test + reusable Keycloak testcontainer fixture). Parent Goal #800 (G11 agentic ops runtime).

damir-topic:
## Summary

- The scheduler is operator-less and previously sourced an agent's `client_credentials` secret only from a pod env var (`MEHO_AGENT_SECRET_*`); the Vault path setting was defined-but-never-read. An agent registered + defined purely over the API was therefore **not schedulable** until an operator wired its secret into the pod env and redeployed — the registration path and the run path disagreed about where the credential lived.
- **Capture-at-registration:** `AgentPrincipalService.register` now fetches the Keycloak-generated client secret (`KeycloakAdminClient.get_client_secret` → `GET .../clients/{id}/client-secret`) and persists it to Vault at `scheduler_agent_vault_path_pattern` under a scheduler service token. A Vault-write failure rolls back the just-created Keycloak client (fail-closed); an *unset* token skips the write with a warning (backward-compatible with env-var-only deployments).
- **Scheduler Vault read identity:** new `scheduler/vault_credentials.py` reads/writes the secret under a static `VAULT_SCHEDULER_TOKEN` (hvac `Client(token=…)`) — no per-operator JWT. `resolve_agent_credentials` is now **Vault-first** with the env var kept as a documented break-glass fallback; `scheduler_credentials_unresolved` still fires when neither source yields a secret.
- A static service token (not AppRole) is the lowest-friction operator-less Vault identity — it reuses the primitive the live-Vault test harness already uses, with no `secret_id` bootstrap. The doc records AppRole-via-Vault-Agent as an additive, no-code-change alternative.

## Test plan

- [x] `mypy src/meho_backplane/` — no issues (430 files)
- [x] Unit: `test_scheduler_credentials.py` (Vault-first / fallback / both-miss), `test_scheduler_vault_credentials.py` (path splitter, write/read round-trip, not-configured, error mapping), `test_keycloak_admin.py` (`get_client_secret`), `test_api_v1_agent_principals.py` (persist-to-Vault, rollback-on-write-failure, skip-when-token-unset), `test_scheduler.py` — all green
- [x] Integration (live Vault dev container, `tests/integration/test_scheduler_vault_credentials_dev_e2e.py`): write→read round-trip, resolution from Vault with **no** pod env var set, raises when neither source yields. Skips-in-sandbox (no Docker); runs in CI.
- [x] `docs/codebase/scheduler.md` updated to Vault-first; "ships configured but unused" removed. CHANGELOG `[Unreleased]` bullet added.

## Out of scope

- Agent secret rotation / Vault lease renewal (deferred per the issue).
- Helm chart wiring of `VAULT_SCHEDULER_TOKEN` (env-driven; chart `secretKeyRef` surface is a separate devops follow-up — recorded as adjacent finding).
- Changing the env-var fallback naming/derivation.

Closes #1478

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,17 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Fixed
+
+- Scheduler now sources an agent's `client_credentials` secret from Vault
+  instead of a pod environment variable, so an agent registered + defined
+  purely over the API is schedulable with no `MEHO_AGENT_SECRET_*` env var
+  and no redeploy. Registration captures the Keycloak-generated client
+  secret and persists it to Vault under a scheduler service token
+  (`VAULT_SCHEDULER_TOKEN`); `resolve_agent_credentials` reads it
+  Vault-first, keeping the env var as a documented break-glass fallback
+  (#1478).
+
 ## [0.10.0] - 2026-06-01
 
 The **connector write-surface** release: MEHO connectors graduate from
--- a/backend/src/meho_backplane/api/v1/agent_principals.py
+++ b/backend/src/meho_backplane/api/v1/agent_principals.py
@@ -66,6 +66,7 @@ from meho_backplane.auth.keycloak_admin import (
 )
 from meho_backplane.auth.operator import Operator, TenantRole
 from meho_backplane.auth.rbac import require_role
+from meho_backplane.scheduler.vault_credentials import SchedulerVaultBrokerError
 
 __all__ = ["router"]
 
@@ -200,6 +201,16 @@ async def register_agent_principal(
         ) from exc
     except KeycloakAdminError as exc:
         raise _handle_admin_error(exc) from exc
+    except SchedulerVaultBrokerError as exc:
+        # ``VAULT_SCHEDULER_TOKEN`` is set but the Vault write failed
+        # (unreachable / denied). 502 upstream failure; the just-created
+        # Keycloak client is already rolled back. (An *unset* token is not
+        # an error — the service skips the write and the agent uses the
+        # env-var fallback.)
+        raise HTTPException(
+            status_code=http_status.HTTP_502_BAD_GATEWAY,
+            detail="scheduler_vault_write_error",
+        ) from exc
     except ValueError as exc:
         raise HTTPException(
             status_code=http_status.HTTP_422_UNPROCESSABLE_ENTITY,
--- a/backend/src/meho_backplane/auth/agent_principals.py
+++ b/backend/src/meho_backplane/auth/agent_principals.py
@@ -68,6 +68,10 @@ from meho_backplane.auth.keycloak_admin import (
 )
 from meho_backplane.db.engine import get_sessionmaker
 from meho_backplane.db.models import AgentPrincipal
+from meho_backplane.scheduler.vault_credentials import (
+    SchedulerVaultNotConfiguredError,
+    write_agent_secret,
+)
 
 __all__ = [
     "AgentPrincipalCreate",
@@ -156,20 +160,25 @@ class AgentPrincipalService:
     ) -> AgentPrincipalRead:
         """Register a new agent principal.
 
-        Creates the Keycloak client first; on success inserts the DB row.
-        If the Keycloak creation fails the DB row is never written.
+        Creates the Keycloak client first, captures its generated
+        ``client_credentials`` secret, and persists that secret to Vault
+        via :meth:`_persist_secret_to_vault` (so the operator-less
+        scheduler can read it back — G0.19-T2 #1478); on success inserts
+        the DB row. If any step fails the Keycloak client is rolled back
+        and the DB row is never written.
 
         Raises
         ------
         ValueError
             When *name* contains characters outside the safe alphabet.
         AgentPrincipalExistsError
-            When a principal with the same name already exists in this
-            tenant (unique-index violation on DB or Keycloak 409).
-        KeycloakAdminNotConfiguredError
-            When Keycloak admin credentials are not configured.
-        KeycloakAdminError
-            On any Keycloak Admin API failure.
+            Duplicate ``(tenant_id, name)`` (DB unique-index or Keycloak 409).
+        KeycloakAdminNotConfiguredError / KeycloakAdminError
+            Keycloak admin unconfigured / any Admin API failure.
+        SchedulerVaultBrokerError
+            Vault configured but the secret write failed — see
+            :meth:`_persist_secret_to_vault` (an *unset* token is a
+            skip-with-warning, not a raise).
         """
         if not _NAME_PATTERN.fullmatch(payload.name):
             raise ValueError(
@@ -180,7 +189,10 @@ class AgentPrincipalService:
         owner = payload.owner_sub or created_by_sub
         client_id = _keycloak_client_id(payload.name)
 
-        # Phase 1: create Keycloak client (fail before DB on error).
+        # Phase 1: create Keycloak client + capture its generated secret
+        # in the same admin session (create_client returns only the
+        # internal UUID; Keycloak never echoes the generated secret on
+        # create). Fail before any DB write on error.
         kc_client = KeycloakAdminClient.from_settings()
         try:
             async with kc_client:
@@ -190,14 +202,21 @@ class AgentPrincipalService:
                     tenant_id=str(tenant_id),
                     owner_sub=owner,
                 )
+                client_secret = await kc_client.get_client_secret(internal_id)
         except KeycloakClientConflictError as exc:
             raise AgentPrincipalExistsError(payload.name) from exc
 
-        # Phase 2: insert the DB row. If anything fails after the Keycloak
-        # client was created, delete that client before surfacing the error:
-        # a created client with no MEHO row is an orphaned, token-issuing
-        # identity that can never be listed or revoked through MEHO — exactly
-        # the unreachable-kill-switch failure this lifecycle exists to prevent.
+        # Phase 1b: persist the captured secret to Vault for the scheduler.
+        await self._persist_secret_to_vault(
+            client_id,
+            client_secret,
+            internal_id=internal_id,
+            tenant_id=tenant_id,
+            name=payload.name,
+        )
+
+        # Phase 2: insert the DB row (rolls back the Keycloak client on
+        # any failure — see _insert_row).
         row = AgentPrincipal(
             tenant_id=tenant_id,
             name=payload.name,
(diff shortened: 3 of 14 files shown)


damir-topic:
@coderabbitai review

damir-topic (commented):




**Anchor task:** #1478 — scheduler must source agent `client_credentials` from Vault, not a pod env var
**Surface area:** backend (auth / scheduler / secrets)
**Verdict:** APPROVE
**Head SHA reviewed:** `20d60ecdaeb6539dedafa3e8d69a8fcb867b5ebd`
**AI reviewer (CodeRabbit):** not received within 15 min — CodeRabbit hit its org review rate limit / out of usage credits and produced **no** findings (Run ID `2de5c255`). Treated as bot-unavailable for this iteration; does not block the verdict.

> Note: author is `damir-topic` (same as reviewer identity), so `gh pr review --approve` is rejected by GitHub. This APPROVE verdict is recorded here and in the machine-readable result file; the orchestrator merges via `--admin`.

### Acceptance criteria
- [x] Registration persists the Keycloak client secret to Vault at `scheduler_agent_vault_path_pattern` — `AgentPrincipalService.register` Phase 1b → `write_agent_secret` (KV-v2 `create_or_update_secret`). Verified by `test_register_persists_captured_secret_to_vault` + live-Vault `test_seeded_payload_field_shape`.
- [x] `resolve_agent_credentials` reads Vault under a scheduler service identity (no per-operator JWT); env var is documented fallback — `vault_credentials._scheduler_client` binds `VAULT_SCHEDULER_TOKEN` via the `_build_client(settings, token=…)` seam. Verified Vault-first ordering in `test_vault_first_secret_wins_over_env`.
- [x] API-registered agent fires with **no** pod env var (e2e) — live-Vault `test_resolve_is_vault_sourced_with_no_env_var` asserts resolution from Vault with `MEHO_AGENT_SECRET_*` deleted.
- [x] `scheduler_credentials_unresolved` still fires when neither source yields — `test_resolve_missing_both_raises` + loop.py unchanged catch path; `loop.py:384` correctly `await`s the now-async resolver.
- [x] `docs/codebase/scheduler.md` updated to Vault-first; "ships configured but unused" removed — verified in the doc text.

### Project-board alignment
|---|---|---|---|---|---|
| #1478 | In Review | Closes | ✅ | ✅ full | close on merge |

### Predicted vs. actual diff
- ✅ `auth/keycloak_admin.py` (`get_client_secret`), `auth/agent_principals.py` (capture+persist+rollback), `scheduler/vault_credentials.py` (new broker), `scheduler/credentials.py` (Vault-first resolver), `scheduler/loop.py` (await), `api/v1/agent_principals.py` (502 mapping), `settings.py` (`vault_scheduler_token`), `CHANGELOG.md`, `docs/codebase/scheduler.md`, 4 test modules + 1 integration test — all expected and correct.
- ❓ none unexpected.
- 🕳️ none missing.

### Integration verification
- [x] `ruff check` — clean (changed modules)
- [x] `mypy` — Success, no issues (5 changed modules, `--ignore-missing-imports`)
- [x] `pytest` — 99 passed (vault_credentials / credentials / keycloak_admin / api_v1_agent_principals / scheduler) + 2 passed (scheduler credential-path) against the PR worktree's `src` (PYTHONPATH-pinned; the repo's editable install points at a stale unrelated worktree).
- [⚠️ not run] live-Vault integration test (`test_scheduler_vault_credentials_dev_e2e.py`) — skips in this sandbox (no Docker socket); runs in CI. Reasoned: the suite asserts write→read round-trip, Vault-sourced resolution with no env var, and raise-when-neither.

- None. No blockers, no majors posted inline.

- **Orphaned Vault secret on a Phase-2 DB-insert failure.** Register writes the secret to Vault (Phase 1b) *before* the DB insert (Phase 2). If the DB insert fails, `_rollback_orphan_client` deletes the Keycloak client but leaves the Vault secret at `agents/<CLIENT_ID>/credentials`. Impact is low: the leftover secret is the `client_credentials` for a now-deleted Keycloak client, so it is **non-authenticating** (no client → no token issuance), and a same-name retry idempotently overwrites it via `create_or_update_secret`. This matches the module's already-accepted "best-effort cleanup, manual remediation" posture for the orphaned KC client. A future tightening could `delete_metadata`/destroy the Vault secret inside the rollback path (or reorder Vault-write after the DB commit), but it is not required for this PR's fail-closed correctness.
- **Helm chart does not yet wire `VAULT_SCHEDULER_TOKEN`** (implementer-flagged, out of scope). Confirmed: `deploy/charts/meho/` wires `KEYCLOAK_ADMIN_CLIENT_SECRET` (values.yaml / externalsecrets.yaml / deployment.yaml) but has no `VAULT_SCHEDULER_TOKEN` surface. Until that lands, a chart-deployed scheduler runs with the token unset and falls back to the env-var path — which is the documented, backward-compatible default and does **not** re-introduce #1478 for env-var-wired agents. Recommend the separate devops follow-up adds the `secretKeyRef` surface.

- Adopted: 0 — CodeRabbit was rate-limited (out of credits) and produced no review/findings.
- Disputed: 0
- Out of scope: 0

### Docs
- [x] `docs/codebase/scheduler.md` — updated to Vault-first; "ships configured but unused" removed; documents `VAULT_SCHEDULER_TOKEN`, the raw-API-path split, and the env-var break-glass fallback.

- **Fail-closed secret handling is correct and the leak posture holds.** No code path interpolates the secret *value* into a log event, error message, or HTTP response: `get_client_secret` errors carry only the HTTP status; `vault_credentials` errors carry only the path + `type(exc).__name__`; the 502 detail is a static `"scheduler_vault_write_error"`; `vault_scheduler_token` is `Field(..., repr=False)`. Surfacing the *absence* of a secret (empty `value` → `KeycloakAdminError`) rather than persisting an empty credential is the right call.
- **Break-glass cannot silently mask a Vault outage in a way that re-creates #1478.** Three distinct postures are kept separate: not-configured (`SchedulerVaultNotConfiguredError`) → silent env fallback; read error (`SchedulerVaultBrokerError`) → WARN `scheduler_vault_read_failed` + env fallback; secret-absent (`None`) → env fallback. A Vault outage falls back to the env var *only when that env var is set*; otherwise it still raises `AgentCredentialsUnresolvedError` (loud, trigger-preserving). The write path is stricter (fail-closed) than the read path (fall-back-with-warning), which is the correct asymmetry.
- **Backward-compat is genuinely preserved.** `test_register_skips_vault_when_token_unset` exercises the *real* broker (overriding the autouse stub) and asserts a 201 with the row persisted — an env-var-only deployment registers and schedules unchanged.
- **`split_kv_v2_api_path` honours the "#823 promised a code swap, not an env-var rename" contract** by reconciling the raw KV-v2 API path with hvac's `(mount, logical)` form, with rejects-empty-logical guards and good test coverage.

Issue #1487, opened 2026-06-03
Scheduled agent runs rejected fail-closed at JWT verify — agent_principals.register creates a Keycloak client with no audience mapper / default scopes / tenant-claim mappers (blocks the autonomous agent→approval loop)


damir-topic:
Parent goal: #221
Parent initiative: #1484

## Problem

While dogfooding the v0.10.0 governance surface we tried the canonical autonomous pattern — a scheduled agent requests a write, a human approves it. The scheduled run dies immediately at JWT verification, **before dispatching any operation** — so no operation, no parked approval, no write. This blocks the entire autonomous agent→approval path, which matters because it's effectively the only way the approval queue gets a *distinct* requester: a human-initiated agent run records the human as principal, so four-eyes then blocks them from approving their own request.

Root cause: an agent principal registered via `agent_principals.register` gets a Keycloak `agent:<name>` client created with **no protocol mappers and no default client scopes**. Its `client_credentials` token therefore carries neither `aud=meho-backplane`, nor `sub` (the `basic` scope's subject mapper), nor the `tenant_id`/`tenant_role` claims the backplane requires — and is rejected fail-closed at `verify_jwt_for_audience`.

**Corrections vs. the original dogfood note:**
- The mint **already** requests the audience (RFC 8707 resource indicator: `audience=settings.keycloak_audience`). Stock Keycloak does **not** honor that request param to inject an audience into a `client_credentials` token without a configured `oidc-audience-mapper` / client scope — so requesting it isn't enough; the mapper/scope must exist on the client.
- The failing claim isn't only `aud`. With no mappers/scopes the token is missing `aud`, `sub`, `tenant_id`, **and** `tenant_role` at once. Authlib validates essential claims in insertion order (`iss`→`aud`→`sub`), raising on the first absent one — so the first 401 is typically **`missing_audience`** (no `aud` at all), or `invalid_audience` if Keycloak emitted a default `aud=account`. Same root cause and fix either way.
- The fix is broader than "an aud mapper": clone the **full** working `meho-backplane` client mapper+scope set. The token-mint function is `get_client_credentials_token` (not `mint_agent_token`).

## Current state (verified in code, v0.10.0)

[...]

PR #1495, opened 2026-06-04
fix(scheduler): stop leaking agent client_credentials secret in failed-run logs (#1488)

### LINKED ISSUE #1488: Security: failed scheduled run leaks the agent client_credentials secret in plaintext — scheduler_fire_failed exc_info + structlog dict_tracebacks(show_locals=True) renders agent_client_secret frame locals
Parent goal: #221
Parent initiative: #1484

## Problem

**Security — credential disclosure.** When a scheduled agent run fails, the scheduler's per-row handler logs `scheduler_fire_failed` via `_log.exception(...)` (which sets `exc_info=True`), and structlog's `dict_tracebacks` processor renders **frame locals** (`show_locals=True` by default). The agent's `client_credentials` secret is held in plaintext as a frame local (`agent_client_secret`) on the traceback's frames — so **every failed scheduled run writes that agent's client secret into the JSON logs in cleartext** (stdout → the k8s log collector). We caught our test agent's secret in the logs during dogfooding and rotated / revoked the principal immediately.

It is reachable on exactly the fail-closed path the audience bug (sibling ticket) triggers: the `missing_audience` / `invalid_audience` `HTTPException` is not in `_dispatch_invocation`'s narrow `except` list, so it escapes to `run_one_tick`'s broad `except Exception:` → `_log.exception("scheduler_fire_failed", ...)`.

**Corrections vs. the original dogfood note:**
- The leak is via the **plaintext `agent_client_secret` frame locals** rendered by `show_locals=True` — both the bare `agent_client_secret` local (the kwarg at `loop.py:522` and the `run_scheduled` parameter) and `_PreparedInvocation`'s repr. A 36-char secret *inside* `_PreparedInvocation`'s repr is truncated past `locals_max_string=80` in some frames, but the **bare** local renders verbatim — so a repr-only mask on the dataclass is necessary-but-not-sufficient.
- The cited precedent is inaccurate: there is **no `SecretStr` anywhere** in the backplane (grep: zero hits), and the `credential_write` audit redaction works by hashing params (SHA-256 `params_hash`, never persisting raw params) — unrelated to log frame locals. The `redaction/` engine is a connector-**response** payload sanitizer, not a structlog processor. Neither is the right lever — see Desired state.

## Current state (verified in code, v0.10.0 — leak reproduced through the real processor chain)

- [`scheduler/loop.py:307-326`](backend/src/meho_backplane/scheduler/loop.py#L307-L326) — `_PreparedInvocation` (frozen dataclass) declares `agent_client_secret: str` (plain `str`, not `SecretStr`).
- [`scheduler/loop.py:494-523`](backend/src/meho_backplane/scheduler/loop.py#L494-L523) — `_dispatch_invocation(prepared)` reads `prepared.agent_client_secret` (:522); `prepared` and the bare kwarg value are live locals on this frame.
- [`scheduler/loop.py:598-635`](backend/src/meho_backplane/scheduler/loop.py#L598-L635) — `run_one_tick`'s per-row `except Exception:` calls `_log.exception("scheduler_fire_failed", trigger_id=...)` (:629-632); the captured traceback spans `_fire_cron`/`_fire_one_off`/`_dispatch_invocation`, each holding `prepared`. (`structlog`'s `FilteringBoundLogger.exception` sets `exc_info=True`.)
- [`logging.py:82-93`](backend/src/meho_backplane/logging.py#L82-L93) — the only `structlog.configure()` in the backplane uses `structlog.processors.dict_tracebacks` = `ExceptionRenderer(ExceptionDictTransformer())`; `ExceptionDictTransformer` defaults `show_locals=True`, `locals_max_string=80` (structlog 25.5.0, confirmed by introspection). `configure_logging()` runs at lifespan startup ([`main.py:273`](backend/src/meho_backplane/main.py#L273)) — this is the live prod config; no `show_locals` override.
- [`scheduler/credentials.py:131`](backend/src/meho_backplane/scheduler/credentials.py#L131), [`agent/invocation.py:746`](backend/src/meho_backplane/agent/invocation.py#L746) — the secret is a plain `str` from `os.environ` through to `run_scheduled`; **no `SecretStr` on the path**.
- **Reproduced:** running the exact `logging.py` processor chain over the real call shape renders the plaintext secret in the JSON log line in **three** frames.

**Second (latent) surface:** [`auth/agent_token.py:76-82`](backend/src/meho_backplane/auth/agent_token.py#L76-L82) holds `client_secret` and a `data` dict (containing it) as frame locals — currently masked only because `AgentTokenError` is caught early *without* `exc_info`, but any `exc_info` renderer on that frame would leak it too. The doc claim "the client secret is never logged" ([`docs/codebase/backend.md`](docs/codebase/backend.md)) is true for *direct* logging but does not cover the traceback-locals vector.

## Desired state

The agent `client_credentials` secret can never be rendered into logs, on any exception path.

- **Primary:** wrap it in `pydantic.SecretStr` on `_PreparedInvocation` (and the `run_scheduled` parameter), unwrapping with `.get_secret_value()` only at the `get_client_credentials_token` call site. `SecretStr`'s `repr`/`str` mask to `'**********'` even as a bare frame local (verified), so this closes every vector and establishes the first `SecretStr` precedent in the backplane.
- **And/or:** configure `dict_tracebacks` with `ExceptionDictTransformer(show_locals=False)` in `logging.py` — a one-shot fix that also closes the `agent_token.py` frame, at the cost of triage locals (which `logging.py:37-42` documents as "load-bearing for production triage"). Weigh against the `SecretStr` approach.

## Acceptance criteria

- [ ] A failed scheduled run no longer emits the agent `client_credentials` secret anywhere in the `scheduler_fire_failed` log output — test drives a failure through the real `logging.py` processor chain and asserts the secret string is absent.
- [ ] `agent_client_secret` is `SecretStr` (or otherwise non-rendering) from `_PreparedInvocation` through `run_scheduled`, unwrapped only at the token-mint call.
- [ ] The fix holds regardless of which exception escapes — assert with at least two distinct failure causes (e.g. the audience 401 and a budget error).

## Out of scope

- The audience/mapper bug that triggers this failure path — sibling ticket in this initiative.
- Where the secret is sourced (#1478).

## References

- Origin: `evoila-bosnia/claude-rdc-hetzner-dc#841`.
- structlog `dict_tracebacks` / `ExceptionDictTransformer` (`show_locals` renders frame locals; default `True`, `locals_max_string=80` — confirmed by introspection of structlog 25.5.0): https://www.structlog.org/en/stable/api.html
- pydantic v2 `SecretStr` — `repr()`/`str()` render `'**********'`; real value via `.get_secret_value()` (pydantic 2.13.4): https://docs.pydantic.dev/latest/api/types/#pydantic.types.SecretStr
- Best practice — CWE-532 "Insertion of Sensitive Information into Log File" (the defect class; mitigation: never write credentials to logs): https://cwe.mitre.org/data/definitions/532.html
- Related: #1478 (secret sourcing). Closed origin: #823 (G11.3-T2 — the scheduler fire path). Feature context: Goal #800 (G11).

damir-topic:
## Summary

- **Security (credential disclosure).** A failed scheduled agent run logged the agent's `client_credentials` secret into the JSON logs in cleartext. On the `scheduler_fire_failed` path the secret was a plain-`str` frame local, and structlog's `dict_tracebacks` renderer defaults to `show_locals=True`, so every failed fire serialised the secret to stdout (CWE-532).
- **Fix is defense in depth.** (1) The secret is threaded as a `pydantic.SecretStr` from `_PreparedInvocation` through `AgentInvoker.run_scheduled` — masking to `'**********'` even as a bare frame local, unwrapped via `.get_secret_value()` only at the single `get_client_credentials_token` call site (the first `SecretStr` in the backplane). (2) `configure_logging` runs the traceback transformer with `show_locals=False`, dropping every frame's locals dict — which also closes the latent `auth/agent_token.py` frame where the secret is an unavoidable plain `str` for the httpx form-post.
- The structured traceback (file / line / function / exception type) is retained for 5xx triage; only the per-frame locals dict is dropped.

## Test plan

- [x] `ruff check` — clean on all changed files
- [x] `mypy src/meho_backplane/` — Success, no issues (430 files)
- [x] New regression tests in `tests/test_secret_leak_checks.py`:
  - [x] Drives a failed fire through the **real** `configure_logging` chain for **two distinct** escaping exceptions (an audience 401 `HTTPException` and a bare `RuntimeError`) and asserts a dummy secret canary is absent while `scheduler_fire_failed` is present (non-vacuous).
  - [x] Pins the `SecretStr` layer independently by re-enabling `show_locals=True` and asserting the masked form renders (secret still absent).
  - [x] Confirmed non-vacuous: the pre-fix shape (plain `str` + `show_locals=True`) leaks the canary, so the assertions would fail against the unfixed code.
- [x] `pytest` — affected suites green: `test_secret_leak_checks`, `test_scheduler`, `test_scheduler_admin`, `test_scheduler_credentials`, `test_agent_invocation`, `test_auth_agent_token`, `test_observability` (no realistic credential value ever emitted; tests assert masking of an obvious dummy placeholder).
- [x] Docs: `docs/codebase/backend.md` updated for the `configure_logging` `show_locals=False` change and the `SecretStr` threading through `run_scheduled`.

## Out of scope

- The audience/mapper bug that triggers this failure path — sibling ticket #1487 in the same initiative.
- Where the secret is sourced (#1478, already merged).

Closes #1488

## Release Notes

* **Bug Fixes**
  * Fixed a vulnerability where agent client credentials were exposed in JSON logs during failed scheduled agent runs. Credentials are now properly masked in structured exception tracebacks.

* **Tests**
  * Added comprehensive tests to verify agent credentials are not inadvertently exposed in failure scenarios.

* **Documentation**
  * Updated documentation to reflect changes in structured exception traceback rendering.


damir-topic:
@coderabbitai review

damir-topic (commented):




**Anchor task:** #1488 — failed scheduled run leaks the agent `client_credentials` secret in plaintext (CWE-532)
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `5e0f849b4e0177f391a6d2625cbbad5f40ba26fd`
**AI reviewer (CodeRabbit):** received — "No actionable comments were generated" (CHILL profile, all 8 files, same head SHA); 0 findings

This is a security fix (credential disclosure) and was reviewed with extra rigor: the SecretStr thread was grepped end-to-end, the regression test's non-vacuity was independently reproduced, and the `show_locals=False` blast radius was checked against the existing traceback test.

### Acceptance criteria
- [x] A failed scheduled run no longer emits the agent secret anywhere in `scheduler_fire_failed` output — the regression test drives a real failed fire through the **production** `configure_logging` chain and asserts the canary is absent. Independently reproduced: pre-fix shape (plain `str` + `show_locals=True`) leaks the canary; the production chain does not.
- [x] `agent_client_secret` is `SecretStr` from `_PreparedInvocation` through `run_scheduled`, unwrapped only at the token-mint call — verified by grepping every read on the path; the single `.get_secret_value()` is at `invocation.py:782`.
- [x] Holds regardless of which exception escapes — the test is parametrized over an audience-401 `HTTPException` and a bare `RuntimeError`, both green.

### Security review (extra rigor)
- **SecretStr thread complete, no re-widening.** `_prepare_invocation` resolves a plain `str` then wraps it into `SecretStr` immediately (`loop.py:415`); that frame returns before any fire, so it is not on the failure traceback. The value stays `SecretStr` on `_PreparedInvocation` (`loop.py:335`), is passed through `_dispatch_invocation` as `prepared.agent_client_secret` (`loop.py:539`), and is the `SecretStr` param of `run_scheduled` (`invocation.py:747`). It is unwrapped exactly once, at the `get_client_credentials_token` call (`invocation.py:782`). No f-string, dict, log kwarg, exception message, `.model_dump()`, `repr`, audit field, or metrics label carries the raw value.
- **Mask holds as a bare frame local AND in the dataclass repr** — verified empirically: a bare `SecretStr` local and the frozen `_PreparedInvocation` repr both render `'**********'` (canary absent) under `show_locals=True`.
- **Residual leak vectors: none found.** The only irreducible plain-`str` frame is `get_client_credentials_token` (httpx needs the raw value for the form-post); it is covered globally by `show_locals=False`, its log lines carry only `client_id`/`status`/`error-type`, and its `AgentTokenError` messages never include the secret.
- **No realistic secret in the diff** — only the obvious dummy canary and fake `s3cr3t`/`dummy` placeholders.

### Predicted vs. actual diff
- ✅ `agent/invocation.py`, `scheduler/loop.py`, `logging.py`, `tests/test_secret_leak_checks.py`, `CHANGELOG.md`, `docs/codebase/backend.md` — expected and correct
- ❓ `tests/test_agent_invocation.py`, `tests/test_scheduler_admin.py` — justified: callers updated to construct `SecretStr(...)` for the new param/field type
- 🕳️ Missing: none. The latent `auth/agent_token.py` frame is correctly addressed by `show_locals=False` rather than a code change (httpx needs the raw `str` there).

### Integration verification
- [x] `ruff check` — All checks passed (changed files)
- [x] `mypy src/meho_backplane/` — Success, no issues in 430 files; no `# type: ignore` added for the SecretStr type change
- [x] `pytest` — 131 passed (`test_secret_leak_checks`, `test_scheduler`, `test_scheduler_admin`, `test_scheduler_credentials`, `test_agent_invocation`, `test_auth_agent_token`, `test_observability`)
- [x] Independent non-vacuity repro — confirmed the pre-fix shape leaks and the production chain masks
- [⚠️ not run] docker compose — not applicable to this logging/type change

### `show_locals=False` blast radius
Acceptable and documented. No existing test asserts on traceback frame *locals*. `test_observability.py::test_handler_exception_emits_structured_traceback` asserts only on `exc_type` / `exc_value` / `frames` (file/line/func) — exactly the triage data retained under `show_locals=False` — and it builds its own structlog config, so it is decoupled from this change and stays green. The tradeoff (dropping per-frame locals for triage) is documented in the `configure_logging` docstring and `docs/codebase/backend.md`.

### Findings (inline)
None. No blocker, major, minor, or question.

- Adopted: 0
- Disputed: 0
- Out of scope: 0
- CodeRabbit reviewed the exact head SHA and reported zero actionable comments; its walkthrough correctly captured the SecretStr thread + `show_locals=False` intent.

### Docs
- [x] `docs/codebase/backend.md` — updated for both the `configure_logging` `show_locals=False` change and the `SecretStr` threading.

### What worked well
- Defense-in-depth done right: the type-level `SecretStr` guard and the global `show_locals=False` are independent, and the test suite pins each layer separately (one drives the production chain; one re-enables `show_locals=True` to prove the mask survives a future triage re-enable).
- The regression test is explicitly non-vacuous — it asserts the failure *was* logged before asserting the secret is absent, and the canary is a module-level constant so it never lands in the test's own frame locals.

Issue #1503, opened 2026-06-04
Parked direct operator op approved via /decide or MCP-by-id is never executed — only REST /approve re-dispatches


damir-topic:
Parent goal: #221
Parent initiative: #1500

## Problem

A parked **direct operator op** (an operator calling a `requires_approval` op directly, not via an agent run) that is then approved is **never executed**. The only execute-after-approve path is REST `POST /approve` carrying the original params; approving via `/decide` or the MCP/CLI by-id approve commits the decision **without** re-dispatching, and simply re-calling the op re-parks it.

(For an *agent run*, resume works in-process — the agent loop awaits `approval.approved` and re-dispatches; this task is specifically the **direct** operator-op path, which has no awaiting loop.)

## Current state (verified in code, v0.10.1)

- [`api/v1/approvals.py:265-274`](backend/src/meho_backplane/api/v1/approvals.py#L265-L274) — `_resume_dispatch_after_approval` calls `dispatch(..., _approved=True)`; it is reached **only** from REST `POST /approve` with the original params.
- [`api/v1/approvals.py:586-609`](backend/src/meho_backplane/api/v1/approvals.py#L586-L609) — `/decide` commits `approve_request(params=None)` then publishes `approval.approved`, but never re-dispatches the op. MCP/CLI by-id approve is the same shape.
- [`operations/dispatcher.py:1046-1085`](backend/src/meho_backplane/operations/dispatcher.py#L1046-L1085) — `dispatch` bypasses the approval gate **only** on `_approved=True`; a normal re-call of the op hits `requires_approval` again and routes to `NEEDS_APPROVAL` (re-parks).

So an operator who approves their colleague's parked direct write via the activity feed / `/decide` / MCP sees it marked approved but the write never lands; the op silently has to be re-invoked through the exact REST `/approve` shape that carries the params.

## Desired state

Approving a parked direct operator op through any approval surface (`/decide`, MCP by-id, REST `/approve`) results in the approved op being executed exactly once — the approve decision drives the re-dispatch (using the stored original params) rather than only recording the decision.

## Acceptance criteria

- [ ] A direct operator op parked for approval and then approved via `/decide` (and via MCP by-id approve) is executed once — its effect lands — without re-invoking the op (integration test per surface).
- [ ] The op's stored params drive the post-approval re-dispatch; double-execution is prevented (the request transitions to a terminal executed state).
- [ ] Agent-run resume behavior is unchanged (no regression in the in-process resume path).

## Out of scope

[...]

Issue #1502, opened 2026-06-04
Scheduler tick blocks and holds the advisory lock until pod restart — run_scheduled awaits the run with no timeout


damir-topic:
Parent goal: #221
Parent initiative: #1500

## Problem

A long-running or hung scheduled run blocks the **entire** scheduler tick loop — and holds a process-wide advisory lock — until a pod restart. The consumer saw a second valid one-off trigger not fire for 3.5+ min after an earlier run, recovering only on restart.

Root cause: `run_scheduled` does a bare, unbounded `await task` with no timeout ([agent/invocation.py:835](backend/src/meho_backplane/agent/invocation.py#L835)), while `run_one_tick` holds a session-level `pg_try_advisory_lock` released only in its `finally` **after** the serial row loop ([scheduler/loop.py:604-659](backend/src/meho_backplane/scheduler/loop.py#L604-L659)). `_scheduler_loop` is strictly serial (sleep → `await run_one_tick`), so any run that blocks — a hung model HTTP call, or (realistically) a `requires_approval` op whose in-process wait blocks up to `agent_approval_wait_timeout_seconds` (default **1800s / 30 min**, [settings.py:913](backend/src/meho_backplane/settings.py#L913)) — stalls every later tick. Because the run never returns, the `finally` never runs and the advisory lock is held for the process lifetime, so other replicas also skip — only a pod restart recovers.

The human path already guards this with `asyncio.wait_for(asyncio.shield(task), timeout=...)` ([invocation.py:711-731](backend/src/meho_backplane/agent/invocation.py#L711-L731)); the scheduled path omits the equivalent.

**Correction vs. the original report:** it is **not** "a *failed* run wedges the loop." A fast failure (e.g. an Anthropic 400) is caught, finalized `failed`, and returns in seconds ([invocation.py:614-626](backend/src/meho_backplane/agent/invocation.py#L614-L626)); the per-row `try/except` already isolates raised exceptions ([loop.py:642-652](backend/src/meho_backplane/scheduler/loop.py#L642-L652)). The wedge is a blocking/hung `await`, not an escaping exception or a leaked lock on clean failure.

## Current state (verified in code, v0.10.1)

[...]

Issue #825, opened 2026-05-21
Task: G11.3-T4 in-flight resume-or-fail-into-audit policy (lease/heartbeat + reaper, or DBOS resume) — no run silently lost


damir-topic:
Parent goal: #800
Parent initiative: #804

## Context

A scheduled or event-fired agent run that gets killed mid-flight (pod restart, crash) must do one of two things, per the agent definition: **resume** where it left off, or **fail cleanly into the audit log**. This Task implements that policy, on whichever substrate 11.3-T1 selected.

## Current state

- No in-flight-resume mechanism. The roll-our-own pattern's failure isolation is per-tick try/except ([memory/expiry.py](backend/src/meho_backplane/memory/expiry.py)) — a killed process loses in-flight work entirely today. The run record (11.1-T6 #813) tracks `status`. DBOS (if chosen in 11.3-T1) provides automatic resume.

## Desired state

- An `in_flight_policy` on the trigger/run: `resume` or `fail_into_audit`.
- **Option A (roll-our-own):** a claimed run carries a lease + heartbeat; a reaper detects an expired lease (the worker died) and, per policy, either re-dispatches the run (best-effort resume from the run's recorded state) or marks it `failed` with a clean audit row explaining the interruption. At-least-once semantics documented; agent runs should be idempotent-friendly.
- **Option B (DBOS):** `resume` is automatic (checkpointed steps continue after restart); `fail_into_audit` is a definition that opts out of resume and records the interruption.
- Either way: no run is silently lost — every killed in-flight run ends in a terminal audited state.

## Acceptance criteria

- [ ] A run killed mid-flight ends in a terminal, audited state: either resumed to completion or `failed` with an interruption reason — never silently lost.
- [ ] The per-definition `in_flight_policy` is honored.
- [ ] (Option A) a lease/heartbeat + reaper detects dead workers and applies the policy; (Option B) DBOS resume verified by a kill-restart test.
- [ ] `Python (ruff + mypy + pytest)` + integration green; a test kills a running scheduled run and asserts the policy outcome.

## Out of scope

- Trigger firing (11.3-T2/T3) and the admin surface (11.3-T5).
- Cross-replica work-stealing beyond claim-and-reap.

## References

- Grounding: failure-isolation precedent [memory/expiry.py](backend/src/meho_backplane/memory/expiry.py), run record (#813), the 11.3-T1 substrate decision.
- DBOS durable resume; lease/heartbeat + reaper pattern; at-least-once semantics.
[...]

PR #1511, opened 2026-06-04
fix(agent): wire lease/heartbeat into the run fire path so the reaper can reclaim hung runs (#1501)

### LINKED ISSUE #1501: Agent-run lease/heartbeat never wired into the fire path — reaper can't reclaim a hung/crashed run (stuck status=running forever)
Parent goal: #221
Parent initiative: #1500

## Problem

Driving the autonomous agent→approve→**execute** loop on the v0.10.1 deploy, a scheduler-fired run parked under `sub=agent`; on approve (`approval_request_approved` logged) the run stayed `status=running` indefinitely and was **never resumed or reaped** — only `agent_run_reaper_started` was present.

The reported root cause ("the out-of-process resume trigger is missing") is **not** the bug: approval-resume runs in-process inside the agent loop's wrapped `call_operation` tool and is identical for human and scheduler-fired runs ([agent/approval_wait.py:522-540](backend/src/meho_backplane/agent/approval_wait.py#L522-L540), [agent/invocation.py:821-842](backend/src/meho_backplane/agent/invocation.py#L821-L842); 10/10 resume tests pass). The real defect the symptom exposes: **the in-flight lease/heartbeat lifecycle (G11.3-T4 #825) was never wired into the run path, so the reaper can never reclaim a hung or worker-killed run** — it stays `running` forever with no terminal/audited state, defeating #825's "no run silently lost" contract. If the run's asyncio task dies (worker recycle, unhandled crash), nothing is awaiting the approval to resume it **and** nothing reaps it — exactly the observed "approved but never re-dispatched, reaper started but never reaps."

This affects **all** agent runs, not only scheduler-fired ones — `claim_lease`/`heartbeat` have no production caller anywhere.

## Current state (verified in code, v0.10.1)

- [`agent/reaper.py:379-389`](backend/src/meho_backplane/agent/reaper.py#L379-L389) — the reaper scans `WHERE status='running' AND lease_expires_at IS NOT NULL AND lease_expires_at < now`. [`reaper.py:303-309`](backend/src/meho_backplane/agent/reaper.py#L303-L309) self-documents the gap: "T2/T3 will wire that filter when they fire runs."
- [`operations/agent_run.py:316-331,436-465`](backend/src/meho_backplane/operations/agent_run.py#L316-L331) — `_create_run_row` calls only `create_run` + `start_run`; neither sets `lease_owner`/`lease_expires_at`. `claim_lease`/`heartbeat`/`release_lease` ([:468-596](backend/src/meho_backplane/operations/agent_run.py#L468-L596)) are referenced **only by tests** (`git log -S 'await claim_lease'` is empty). So every run has `lease_expires_at = NULL` and is unreapable.
- Even if selected, the reaper only flips the DB row to `failed` ([reaper.py:329-330](backend/src/meho_backplane/agent/reaper.py#L329-L330)) — it does not cancel the in-process `asyncio.Task`, so a wedged `await task` keeps blocking regardless.

## Desired state

A hung, crashed, or worker-killed agent run is reliably driven to a terminal, audited state (the #825 "no run silently lost" contract). The run path stamps a lease on start and heartbeats while alive; the reaper reclaims runs whose lease has expired (and a reclaimed run lands `failed` with a clear reason, not stuck `running`).

## Acceptance criteria

- [ ] `_create_run_row` / the run loop calls `claim_lease` on start and `heartbeat` periodically, so a live run has a non-NULL `lease_expires_at` (unit test asserts the lease is stamped).
- [ ] A run whose lease has expired (simulated dead worker) is transitioned to `failed` by the reaper within one reap interval (test using the existing `test_agent_run_reaper` harness).
- [ ] An integration test exercises the scheduler-fired run → park → approve → resume path through the **real** broadcast delivery (the current resume test stubs the broadcast client, so a real Valkey delivery gap would be invisible).

## Out of scope

- The scheduler tick blocking on a long run (separate task — `run_scheduled` bounded-wait).
- The direct-operator-op execute-after-approve path (separate task).

## References

- Origin: `evoila-bosnia/claude-rdc-hetzner-dc#864` (v0.10.1 closed-loop dogfood).
- Partial-overlap / provenance: #825 (G11.3-T4 — built lease/heartbeat+reaper, left the "wire into the fire path" half to T2/T3), #804 (G11.3 Scheduler initiative), #1171 (G11.1-T9 resume-on-broadcast substrate). Predecessor wall: #1487 (fixed reaching the parked-approval point). Feature context: Goal #800 (G11).

damir-topic:
## Summary

- The in-flight lease/heartbeat lifecycle (G11.3-T4 #825) was built and tested but **never wired into the run path** — `_create_run_row` only called `create_run` + `start_run`, so every agent run committed with `lease_expires_at = NULL`. The reaper's claim query (`status='running' AND lease_expires_at IS NOT NULL AND < now`) could therefore never match a real run, leaving a hung/crashed/worker-killed run stuck `status=running` forever with no terminal, audited state (SEV-1; defeats #825's "no run silently lost" contract).
- This PR stamps a lease on every run at start (in the same committed transaction as `pending -> running`) and runs a heartbeat sidecar while the run is alive, across all three run shapes — the background loop, the inline SSE stream, and child (`invoke_agent`) runs. If the worker task dies, the sidecar dies with it, the lease lapses, and the reaper drives the row to `failed` (the default `fail_into_audit` policy) instead of leaving it `running`.
- A `LeaseLostError` (the reaper or an operator cancel already reclaimed the row) stops the heartbeat cleanly so a partitioned worker stops touching a row it no longer owns.
- The `ChildRunRecorder` protocol now returns `(run_id, lease_owner)` so `invoke_agent` can heartbeat the child for its loop's lifetime — child runs are leased too, per the "all agent runs" scope.

The reaper itself already reaped correctly once leases existed (its tests pass); the bug was purely that nothing stamped a lease. The reaper is left unchanged — its resume-dispatch path (`T2/T3 will wire that filter`) is out of this task's scope.

## Test plan

- [x] `ruff format --check src/ tests/` — clean (876 files)
- [x] `mypy src/` — clean (430 source files)
- [x] `pytest tests/ --ignore=tests/integration` — 6675 passed, 194 skipped
- [x] **AC1** — `_create_run_row` claims a lease so a live run has a non-NULL `lease_expires_at`: `test_create_run_row_stamps_lease_so_run_is_reapable`
- [x] **AC2** — a run whose lease expired (simulated dead worker) is transitioned to `failed` by the reaper within one tick, exercised end-to-end through the real fire path: `test_hung_run_with_expired_lease_is_reaped_to_failed`
- [x] **AC3** — an integration test exercises the approval-resume wait through **real** Valkey broadcast delivery (the existing unit test stubs the broadcast client, so a real delivery gap would be invisible): `tests/integration/test_agent_approval_resume_broadcast_e2e.py` (testcontainers Valkey; runs in CI, skips when no Docker socket)

## Out of scope

- The scheduler tick blocking on a long run (`run_scheduled` bounded-wait) — sibling task #1502.
- The direct-operator-op execute-after-approve path — sibling task #1503.
- The reaper cancelling the in-process `asyncio.Task` — the loop observes terminal status at its next turn boundary via the existing cancellation seam (documented on `cancel_run`).

Closes #1501

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,10 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Fixed
+
+- Wire the agent-run lease/heartbeat into the fire path so a hung, crashed, or worker-killed run is reliably reaped to a terminal `failed` state instead of staying `running` forever; the run loop now stamps a lease on start and heartbeats while alive, and child (`invoke_agent`) runs are leased too (#1501).
+
 ## [0.10.1] - 2026-06-04
 
 ### Fixed
--- a/backend/src/meho_backplane/agent/invocation.py
+++ b/backend/src/meho_backplane/agent/invocation.py
@@ -72,6 +72,9 @@ the operator's tenant owns (cross-tenant run ids surface as
 from __future__ import annotations
 
 import asyncio
+import contextlib
+import os
+import socket
 import uuid
 from collections.abc import AsyncIterator
 from dataclasses import dataclass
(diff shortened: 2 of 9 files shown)


damir-topic:
@coderabbitai review

damir-topic:
@coderabbitai review

Issue #1389, opened 2026-05-31
SSE feed delivers 0 bytes on deploy — AuditMiddleware buffers text/event-stream (/api/v1/feed + /ui/broadcast/stream dark; unmasked by #1354)


damir-topic:
Parent goal: #221
Parent initiative: #1575

## Problem

`AuditMiddleware` buffers the **entire** ASGI response before forwarding it, so every Server-Sent-Events endpoint (`/api/v1/feed`, `/ui/broadcast/stream`) delivers **zero bytes to the client until the stream ends** — which, for an open-ended SSE feed, is never. The live-activity UI and `meho status --watch` are dark on a real deploy.

This is a regression that was *masked* until recently: #1354 (G0.18-T1) removed a 5s `socket_timeout < XREAD block` that had been periodically tearing down the connection and flushing the buffer with a spurious error frame. Removing the spurious teardown unmasked the underlying defect — the response never streams because the chassis audit middleware holds it.

This is distinct from #1305/#1321 (G0.16-T3), which fixed SSE *fanout* + backlog-prelude at the broadcast layer. Those handlers now correctly *write* stream chunks; the bytes just never leave the middleware.

## Current state (verified in code, 2026-05-31, v0.9.0)

- `AuditMiddleware.__call__` ([`backend/src/meho_backplane/audit.py:638`](backend/src/meho_backplane/audit.py#L638)) runs the inner app through `_run_inner_app_buffered` ([`audit.py:467`](backend/src/meho_backplane/audit.py#L467)), whose `buffered_send` appends **every** `http.response.start` + `http.response.body` message to a list and forwards them only **after** the audit insert completes ([`audit.py:485-494`, `:683`, `:818`](backend/src/meho_backplane/audit.py#L485)).
- The only exemption is `_AUDIT_SKIP_PATH_PREFIXES = ("/mcp",)` ([`audit.py:622`](backend/src/meho_backplane/audit.py#L622), checked at [`:649`](backend/src/meho_backplane/audit.py#L649)). `/api/v1/feed` and `/ui/broadcast/stream` are **not** exempt → they are fully buffered.
- The middleware's own comment admits the stale assumption: *"v0.1 routes return single-shot JSON responses; the buffer is small"* ([`audit.py:657-660`](backend/src/meho_backplane/audit.py#L657)). SSE feeds added later (Goal #217) violate it.
[...]

PR #1585, opened 2026-06-07
fix(audit): stream text/event-stream through AuditMiddleware (#1389)

damir-topic:
## Summary

- `AuditMiddleware` buffered the **entire** ASGI response before forwarding it, so `text/event-stream` responses (`/api/v1/feed`, `/ui/broadcast/stream`) delivered **zero bytes** until the stream closed — which for an open-ended SSE feed is never. The live-activity UI and `meho status --watch` were dark on a real deploy. #1354 unmasked this by removing the periodic teardown that had been flushing the buffer with a spurious error frame.
- `_run_inner_app_buffered` now recognises a `text/event-stream` response (`content-type` on `http.response.start`) and forwards the start message + every body chunk **live** through the real `send` instead of buffering. The audit row is still written — at stream end (normal completion) or on a client-disconnect `CancelledError` — so the fail-closed "every authenticated action gets a row" contract holds for the streaming path.
- The fail-closed-500 swap cannot apply once a stream's `http.response.start` is on the wire; an audit-write failure is still logged loudly, and the feed handler surfaces transport faults to the subscriber as an `event: feed_error` frame. **Non-streaming JSON routes keep the buffered fail-closed-500 contract verbatim.**
- The post-inner-app audit/forward control flow was extracted from `__call__` into `_finalize` so the normal-return and SSE-disconnect paths share one finalizer.

## Test plan

- [x] `ruff check` / `ruff format --check` — clean
- [x] `mypy src/meho_backplane/audit.py` — clean
- [x] New ASGI-transport e2e coverage in `backend/tests/test_api_v1_feed.py::TestSseStreamsThroughMiddleware`:
  - `test_first_chunk_forwarded_before_inner_generator_completes` — drives `RequestContextMiddleware(AuditMiddleware(app))` as a raw ASGI callable with an instrumented `send`; asserts a body chunk reaches the transport **while the SSE generator is still suspended**. Deadlocks (times out → fails) against the old buffering behaviour. *(Note: `httpx.ASGITransport` runs the inner app to completion before exposing the body, so it cannot prove incremental delivery — the raw-ASGI seam is the discriminating one.)*
  - `test_streaming_request_writes_audit_row` — finite generator; asserts exactly one `audit_log` row for `GET /api/v1/feed` with `status_code=200` (fail-closed preserved on the streaming path).
- [x] Existing `test_audit_middleware.py` (non-streaming JSON + fail-closed-500 + handler-exception) and `test_api_v1_feed.py` suites — green (no regression).
[...]

Issue #1575, opened 2026-06-07
Initiative: G0.21 Runtime + operability hardening — post-ship still-live findings (SSE streaming through the chassis audit middleware + standing operability bucket)


damir-topic:
Parent goal: #221

## Why

Operability findings surfaced during `evoila-bosnia/claude-rdc-hetzner-dc` dogfooding were filed as loose issues and never landed under a tracked Initiative. They sit between the per-cycle dogfood Initiatives (G0.8 v0.2 #634 … G0.20 v0.11.0 #1500), **all now closed and shipped**. This Initiative is the home for **post-ship operability/runtime findings that are verified still-real on current `main`** (past `v0.11.0`) and worth driving to closure, rather than reopening shipped release records.

The seed finding is the live-SSE regression (#1389): the chassis audit middleware buffers `text/event-stream` responses, so `/api/v1/feed` and `/ui/broadcast/stream` deliver zero bytes until the (never-ending) stream closes — the live-activity UI and `meho status --watch` are dark on a real deploy.

## Outcome

The shipped backplane streams its SSE feeds incrementally through the full middleware stack, and this Initiative remains the standing bucket for cross-version operability/runtime findings that don't belong to an active dogfood cycle.

## Tasks

- [x] #1389 — SSE feed delivers 0 bytes on deploy (`AuditMiddleware` buffers `text/event-stream`; `/api/v1/feed` + `/ui/broadcast/stream` dark) — done in #1585
- [ ] #1599 — Audit row can be silently dropped on a double-cancelled SSE disconnect (shield the disconnect-path write + add coverage)
- [ ] #1785 — `meho targets import --update --dry-run` prints CREATE for an existing target (dry-run skips the existence GET); route dry-run through the live planner [CLI/Go]
- [ ] #1805 — onboarding guides show an invalid `meho targets import --<field>` flag form; sweep `docs/cross-repo/*` to the valid `targets.yaml` file form [docs]

## Definition of done

- [ ] #1389 closed: `/api/v1/feed` + `/ui/broadcast/stream` stream incrementally through the full middleware stack (ASGI-transport e2e test green), audit row still written (fail-closed preserved), non-streaming JSON routes unchanged.

## Out of scope

- New connector or feature work — this is operability/correctness of already-shipped surfaces only.
- (Operational doc-accuracy for shipped CLI verbs — e.g. #1805 — is in scope as a dogfood-surfaced operability finding; new feature docs are not.)

## References

[...]

PR #1600, opened 2026-06-08
fix(audit): shield SSE disconnect-path audit write against double cancellation (#1599)

### LINKED ISSUE #1599: Audit row can be silently dropped on a double-cancelled SSE disconnect — shield the disconnect-path write + add coverage
Parent goal: #221
Parent initiative: #1575

## Problem

PR #1585 (closed #1389) made `AuditMiddleware` stream `text/event-stream`
responses incrementally and write the audit row at stream-end **or** on a
client-disconnect `CancelledError`. The disconnect path works, but two
residual gaps surfaced in review:

1. **Silent audit-row drop under double-cancellation.** On an SSE client
   disconnect, `__call__` catches `asyncio.CancelledError` and calls
   `await self._finalize(...)` **unshielded** before re-raising. If a
   *second* cancellation is delivered while `_finalize` is awaiting
   `_write_audit_row` (task-tree teardown, server shutdown, an enclosing
   `timeout()`), that await raises `CancelledError` — which is a
   `BaseException`, not `Exception`, so the `except Exception` arm that
   logs `audit_write_failed` does **not** catch it. The audit row is
   dropped **with no log line at all** — a silent hole in the "every
   authenticated action gets exactly one row" fail-closed contract.

2. **The disconnect branch has no automated test.**
   `test_streaming_request_writes_audit_row` covers the *normal*
   stream-end audit row (it drains a finite generator "rather than
   depending on ASGI's racy disconnect-cancellation handshake"). The
   `except asyncio.CancelledError` branch in `__call__` — the one that
   writes the row for a *disconnected* session — is proven only by
   reasoning. That is the same class of coverage gap that let #1389
   itself ship undetected.

## Current state (verified in code on `main` @ 40a0e419, 2026-06-07)

- `AuditMiddleware.__call__` disconnect arm:
  [`backend/src/meho_backplane/audit.py:772-785`](backend/src/meho_backplane/audit.py#L772-L785)
  — `except asyncio.CancelledError as cancelled: await self._finalize(...) ... raise cancelled`.
  The `_finalize` call is **not** wrapped in `asyncio.shield`.
- `_finalize` writes the row at
  [`audit.py:888-900`](backend/src/meho_backplane/audit.py#L888-L900)
  (`await _write_audit_row(...)`), inside a `try` whose only handler is
  `except Exception:` ([`audit.py:901-914`](backend/src/meho_backplane/audit.py#L901))
  — logs `audit_write_failed`. `asyncio.CancelledError` (subclasses
  `BaseException`, confirmed for 3.12) bypasses this handler.
- For streamed responses the fail-closed-500 swap is correctly skipped
  because the bytes are already on the wire
  ([`audit.py:915-926`](backend/src/meho_backplane/audit.py#L915)); the
  only protection the streamed disconnect path has is the
  `audit_write_failed` log — which the second cancellation skips.
- Tests: `TestSseStreamsThroughMiddleware`
  ([`backend/tests/test_api_v1_feed.py:484`](backend/tests/test_api_v1_feed.py#L484))
  has `test_first_chunk_forwarded_before_inner_generator_completes`
  ([`:515`](backend/tests/test_api_v1_feed.py#L515) — drives the
  middleware as a raw ASGI app with an emulated `http.disconnect` at
  [`:588-598`](backend/tests/test_api_v1_feed.py#L588)) and
  `test_streaming_request_writes_audit_row`
  ([`:631`](backend/tests/test_api_v1_feed.py#L631) — normal stream-end
  drain). Neither asserts an audit row on the cancellation/disconnect
  branch.

## Desired state

- The disconnect-path audit write completes (or fails **loudly**) even
  when a second `CancelledError` arrives mid-write. The Python 3.12
  docs' recommended mechanism: schedule/`asyncio.shield()` the write
  (keeping a strong reference — the event loop holds only weak refs to
  tasks), and on `CancelledError` at the shield site still `await` the
  shielded write to completion before re-raising, e.g.:

  ```python
  write = asyncio.ensure_future(self._finalize(...))
  try:
      await asyncio.shield(write)
  except asyncio.CancelledError:
      await write          # let the audit commit finish
      raise                # honor cancellation (no uncancel() suppression)
  ```

  Net effect: on disconnect the row is committed even under teardown; an
  audit-write *failure* still emits a single structured log line
  (`audit_write_failed`, or a distinct `audit_write_cancelled`) — never
  a silent drop.
- Shielding applies **only** to the `CancelledError` arm; the normal
  (non-disconnect) return path is untouched.
- An automated test drives the middleware's disconnect/cancellation
  branch (reuse the raw-ASGI driver from
  `test_first_chunk_forwarded_before_inner_generator_completes`) and
  asserts exactly one `audit_log` row is written for the disconnected
  SSE request.

## Acceptance criteria

- [ ] On an SSE client disconnect, the audit row is written even if the
  task is cancelled a second time while the audit insert is in flight;
  under no interleaving is the row dropped *without at least one
  structured log line*.
- [ ] The `CancelledError` is still re-raised after the audit write
  (cancellation is honored; no suppression without `uncancel()`), so
  enclosing `TaskGroup` / `timeout()` semantics are unchanged.
- [ ] Non-streaming JSON routes and the normal stream-end path retain
  exact current behavior (existing `test_audit_middleware.py` +
  `TestSseStreamsThroughMiddleware` tests stay green).
- [ ] A new test in `TestSseStreamsThroughMiddleware` drives the
  disconnect/cancellation branch and asserts exactly one `audit_log` row
  for the disconnected `/api/v1/feed` request; it must **fail** against a
  variant where the disconnect-path write is skipped/dropped (i.e. it
  genuinely exercises the branch, mirroring how #1585's incremental-
  delivery test was made discriminating).
- [ ] `cd backend && uv run ruff check . && uv run ruff format --check . && uv run mypy . && uv run pytest tests/test_api_v1_feed.py tests/test_audit_middleware.py`
  all green.

## Out of scope

- Any change to non-streaming JSON route buffering or the fail-closed-500
  swap.
- Broadcast / publish-on-write behavior (`_publish_broadcast_event`) —
  unchanged.
- Re-architecting the buffered-vs-streamed split introduced by #1585.

## References

- Parent: Initiative #1575 (G0.21 operability bucket), Goal #221.
  Origin: review of PR #1585 (which closed #1389).
- In-repo precedent this hardens: `audit.py` `__call__` / `_finalize`
  and the raw-ASGI test driver in `test_api_v1_feed.py` (both from
  #1585).
- `asyncio.shield()` + Task cancellation semantics, Python 3.12 —
  `CancelledError` subclasses `BaseException`; `shield()` protects the
  inner task but the outer `await` still raises; suppressing
  cancellation requires `Task.uncancel()`:
  https://docs.python.org/3.12/library/asyncio-task.html#asyncio.shield
  (and the "Task Cancellation" section of the same page).
- Best-practices-pack anchors (skill Hard rule 11): the referenced packs
  are not present in this checkout; substituting the authoritative
  external cite above. Python-surface task.

damir-topic:
## Summary

- On an SSE client disconnect, `AuditMiddleware.__call__` catches `CancelledError` and writes the audit row via `_finalize` before re-raising (#1389). That await was **unshielded**: a *second* `CancelledError` arriving while the audit INSERT was in flight (task-tree teardown, server shutdown, an enclosing `timeout()`/anyio cancel scope) propagated straight out of the bare `await`, bypassed `_finalize`'s `except Exception` arm (`CancelledError` is a `BaseException`), and dropped the row **with no log line** — a silent hole in the "every authenticated action gets exactly one row" fail-closed contract.
- Fix: schedule the disconnect-path write as a task (strong reference kept) and drain it to completion under `asyncio.shield` in a loop, so redelivered/level-triggered cancellations interrupt only the *wait*, never the shielded write. The original `CancelledError` is re-raised once the row commits — cancellation is honored, enclosing `TaskGroup`/`timeout()` semantics unchanged (no `Task.uncancel()` suppression).
- Shielding applies **only** to the `CancelledError` arm; the normal (non-disconnect) return path is byte-for-byte unchanged. No FastAPI route/schema touched → no OpenAPI regen.

## Test plan

- [x] New `test_double_cancelled_disconnect_still_writes_audit_row` in `TestSseStreamsThroughMiddleware` drives the middleware as a raw ASGI app, reproduces a production disconnect by cancelling the ASGI task while the feed generator is suspended (how uvicorn signals a disconnect under `asgi.spec_version >= 2.4`), and delivers a **second** cancellation while the INSERT is in flight — asserts exactly one `audit_log` row for the disconnected `GET /api/v1/feed`, and that cancellation is still re-raised.
- [x] **Discriminating** (AC#4): adversarially verified — against an unshielded variant the second cancel drops the row and the test FAILS (`disconnect-path audit write did not run to completion`); restored, it passes.
- [x] `uv run ruff check src/ tests/` — clean (the one pre-existing RUF034 lives in `alembic/`, outside the `src/ tests/` CI scope, and is unchanged on `main`).
- [x] `uv run ruff format --check src/ tests/` — clean.
- [x] `uv run mypy src/` (CI gate, strict) — `Success: no issues found in 464 source files`.
- [x] `uv run pytest tests/test_api_v1_feed.py tests/test_audit_middleware.py` — 43 passed, 1 skipped (Docker-gated integration). Confirms AC#3: non-streaming JSON routes + the normal stream-end path retain exact current behavior.

## Acceptance criteria

- [x] On an SSE disconnect, the audit row is written even under a second cancellation mid-INSERT; never dropped without at least one structured log line.
- [x] `CancelledError` re-raised after the write (no suppression without `uncancel()`); `TaskGroup`/`timeout()` semantics unchanged.
- [x] Non-streaming JSON routes + normal stream-end path retain exact current behavior (existing tests green).
- [x] New `TestSseStreamsThroughMiddleware` test drives the disconnect/cancellation branch, asserts exactly one row, and fails against the skipped/dropped-write variant.
- [x] ruff check + ruff format --check + mypy + the two pytest files all green (scoped to the CI gate; see Test plan note on the pre-existing alembic-only RUF034).

## Out of scope

- Non-streaming JSON route buffering / the fail-closed-500 swap — untouched.
- Broadcast/publish-on-write (`_publish_broadcast_event`) — unchanged.
- The buffered-vs-streamed split from #1585 — not re-architected.

### Adjacent findings (deferred, not in scope)

- Pre-existing `RUF034` in `backend/alembic/versions/0023_create_approval_request.py:228` (`sa.text("'{}'") if is_postgres else sa.text("'{}'")`) — present on `main`, outside the `src/ tests/` CI lint scope.
- `backend/tests/test_api_v1_feed.py` carries pre-existing strict-mypy noise (`call-overload`/`unused-ignore` on the raw-ASGI `send`/`int(message["status"])` pattern, line 605). The new test mirrors that established in-file pattern; the test tree is intentionally outside the `mypy src/` CI gate.

Closes #1599

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of audit logging when SSE streams disconnect unexpectedly. Audit records are now guaranteed to be written even if cancellation errors occur during the write operation.

* **Tests**
  * Added regression test to verify proper audit logging behavior during stream disconnections under various cancellation scenarios.

### DIFF
--- a/backend/src/meho_backplane/audit.py
+++ b/backend/src/meho_backplane/audit.py
@@ -757,7 +757,9 @@ class AuditMiddleware:
         # out of the inner app; the audit row for the now-closed session
         # is still written (the ``result`` holder is fully populated by
         # then) before the cancellation propagates to unwind the task
-        # tree per asyncio's contract.
+        # tree per asyncio's contract. That disconnect-path write is
+        # shielded (#1599) so a second cancellation mid-INSERT cannot
+        # silently drop the row — see the ``except`` arm below.
         buffered: list[Message] = []
         result = _InnerAppResult()
         try:
@@ -775,13 +777,56 @@ class AuditMiddleware:
             # sent on the first yield; write the audit row for the closed
             # session, then re-raise to unwind the task tree (Sonar
             # S7497 / asyncio cancellation contract).
-            await self._finalize(
-                scope=scope,
-                send=send,
-                buffered=buffered,
-                result=result,
-                duration_ms=round((time.monotonic() - start) * 1000, 2),
+            #
+            # The write is run as a shielded task (#1599): a *second*
+            # ``CancelledError`` delivered while ``_finalize`` is awaiting
+            # the audit INSERT (task-tree teardown, server shutdown, an
+            # enclosing ``timeout()`` or anyio cancel scope) would
+            # otherwise propagate straight out of a bare ``await`` —
+            # ``CancelledError`` is a ``BaseException``, so ``_finalize``'s
+            # ``except Exception`` arm never sees it and the row is dropped
+            # with no log line at all.
+            #
+            # ``ensure_future`` schedules the write on the loop and we keep
+            # a strong reference (the loop holds only a weak one). We then
+            # drain it to completion under ``asyncio.shield``. The drain is
+            # a loop, not a single ``await``: cancellation can be delivered
+            # repeatedly (an ASGI server re-cancelling the task on a
+            # half-closed disconnect, an enclosing anyio cancel scope that
+            # is *level-triggered* and re-raises at every checkpoint while
+            # it stays cancelled), so a single ``await shield(write)`` may
+            # be interrupted before the shielded task finishes. Re-awaiting
+            # under a fresh shield each iteration lets the write run to
+            # completion regardless of how many cancellations are
+            # redelivered. The shielded task itself is never cancelled by
+            # these (that is what ``shield`` guarantees); only our wait is.
+            # Once the write is done we re-raise the original ``cancelled``
+            # so the disconnect is honored, without suppressing it (no
+            # ``Task.uncancel()``).
+            write = asyncio.ensure_future(
+                self._finalize(
+                    scope=scope,
+                    send=send,
+                    buffered=buffered,
+                    result=result,
+                    duration_ms=round((time.monotonic() - start) * 1000, 2),
+                ),
             )
+            while not write.done():
+                try:
+                    await asyncio.shield(write)
+                except asyncio.CancelledError:
+                    # A redelivered cancellation interrupted the wait, not
+                    # the shielded write. Loop to finish draining it; the
+                    # original ``cancelled`` is re-raised below to honor the
+                    # disconnect.
+                    continue
+            # Surface a ``_finalize`` failure (the handler itself raised);
+            # ``write.result()`` re-raises it. An SSE disconnect never
+            # populates ``handler_exc`` (the inner app only catches
+            # ``Exception``), so a cleanly-closed stream's write returns
+            # ``None`` here.
+            write.result()
             raise cancelled
 
         duration_ms = round((time.monotonic() - start) * 1000, 2)
--- a/backend/tests/test_api_v1_feed.py
+++ b/backend/tests/test_api_v1_feed.py
@@ -697,6 +697,197 @@ class TestSseStreamsThroughMiddleware:
         assert rows[0].method == "GET"
         assert rows[0].status_code == 200
 
+    async def test_double_cancelled_disconnect_still_writes_audit_row(
+        self,
+        _feed_env: None,
+        monkeypatch: pytest.MonkeyPatch,
+    ) -> None:
+        """A second cancellation mid-INSERT does not drop the disconnect-path row (#1599).
+
+        Drives ``RequestContextMiddleware(AuditMiddleware(app))`` as a raw
+        ASGI callable (the same seam as
+        ``test_first_chunk_forwarded_before_inner_generator_completes``)
+        and reproduces a production SSE disconnect by **cancelling the ASGI
+        task** while the feed generator is suspended — which is how a real
+        ASGI server (uvicorn, ``asgi.spec_version >= 2.4``) signals a
+        client disconnect: it cancels the task running the app. That
+        cancellation raises out of the suspended generator, past
+        ``_run_inner_app_buffered``'s ``except Exception`` (``CancelledError``
+        is a ``BaseException``), and into ``AuditMiddleware.__call__``'s
+        ``except asyncio.CancelledError`` arm — the disconnect-path audit
+        write under test.
+
+        The ``http.disconnect`` / ``receive``-listener route is *not* used:
+        under Starlette's ``spec_version < 2.4`` task-group branch the
+        disconnect listener cancels the stream inside the response's own
+        anyio cancel scope, which absorbs the ``CancelledError`` so the
+        response returns normally and the row is written on the *normal*
+        path — never exercising the ``except`` arm. Pinning
+        ``spec_version="2.4"`` selects the branch that streams directly, so
+        a task-level cancel propagates the way a real disconnect does.
+
+        While the disconnect-path INSERT is in flight, the instrumented
+        write delivers a **second** ``task.cancel()`` — the task-tree
+        teardown / server-shutdown / enclosing-``timeout()`` interleaving
+        the issue describes. With the shielded write the row still commits:
+        the cancellation hits ``await asyncio.shield(write)`` (not the write
+        task), the fix drains the write to completion, then re-raises so
+        cancellation is honored. The post-run ``audit_log`` table carries
+        exactly one row for ``GET /api/v1/feed``.
+
+        This test *discriminates*: against a variant where the
+        disconnect-path ``_finalize`` is awaited unshielded, the second
+        cancellation interrupts the in-flight ``_write_audit_row`` await
+        (``CancelledError`` bypasses ``_finalize``'s ``except Exception``
+        arm) and zero rows are written → ``write_committed`` stays False and
+        ``len(rows) == 1`` fails. Adversarially verified against that
+        broken variant during implementation.
+        """
+        from sqlalchemy import select
+
+        import meho_backplane.audit as audit_module
+        from meho_backplane.db.engine import get_sessionmaker
+        from meho_backplane.db.models import AuditLog
+        from tests._vault_fakes import install_fake_vault
+
+        first_frame = _sse_frame(_make_event(), "1715600000000-0")
+        first_body_seen = asyncio.Event()
+        # Set only in cleanup. The generator must NOT return on its own —
+        # the *only* way out of its post-yield suspension is the disconnect
+        # cancellation that drives ``__call__``'s ``except`` arm. A
+        # generator that could return would race the normal stream-end path
+        # and the test would no longer isolate the disconnect branch.
+        teardown = asyncio.Event()
+
+        async def _one_then_block(*_args: object, **_kwargs: object) -> AsyncIterator[str]:
+            yield first_frame
+            await teardown.wait()
+
+        import meho_backplane.api.v1.feed as feed_module
+
+        monkeypatch.setattr(feed_module, "_feed_generator", _one_then_block)
+
+        # The ASGI task is bound below; the in-flight write closes over it.
+        task_holder: dict[str, asyncio.Task[None]] = {}
+        write_committed = False
+
+        # Wrap the audit INSERT so the SECOND cancellation is delivered
+        # *while the disconnect-path write is in flight*. By the time this
+        # runs we are provably inside ``__call__``'s ``except CancelledError``
+        # arm — it is the only caller of ``_finalize`` for a generator that
+        # never returns — so cancelling the task here is the "second cancel
+        # mid-INSERT" the issue describes, with no race against the normal
+        # path.
+        real_write_audit_row = audit_module._write_audit_row
+
+        async def _instrumented_write(**kwargs: object) -> None:
+            nonlocal write_committed
+            # Deliver the second cancellation, then yield so it is raised at
+            # the parent's ``await asyncio.shield(write)`` before this
+            # shielded write proceeds. ``shield`` keeps the cancel off this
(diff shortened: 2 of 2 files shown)


damir-topic:
@coderabbitai review

damir-topic (commented):




**Anchor task:** #1599 — shield the SSE disconnect-path audit write against a double cancellation + add a discriminating test
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `c58712bf134ae986286e5d4f6e034b3fa6c05cd4`
**AI reviewer (CodeRabbit):** not received — CI `CodeRabbit` status resolved to `Review skipped` (incremental-review + residual rate-limit from a back-to-back auto-run); 0 reviews / 0 inline comments confirmed via GraphQL + REST

### Acceptance criteria
- [x] **AC#1** — row written even under a second cancel mid-INSERT, never dropped without a log line. *Verified empirically:* the shielded drain-loop drains the write to completion regardless of redelivered cancels; against an unshielded variant the row is dropped (see adversarial check below).
- [x] **AC#2** — `CancelledError` re-raised after the write, no `Task.uncancel()` suppression, `TaskGroup`/`timeout()` semantics unchanged. *Verified:* `raise cancelled` always runs on the disconnect path (`_finalize` returns `None` there), and the new test asserts `pytest.raises(asyncio.CancelledError)`.
- [x] **AC#3** — non-streaming JSON + normal stream-end path retain exact current behavior. *Verified:* the diff confines the change to the `except asyncio.CancelledError` arm (normal-return `_finalize` call byte-for-byte unchanged; `_finalize` itself untouched); `test_audit_middleware.py` + `test_api_v1_feed.py` → 43 passed, 1 skipped.
- [x] **AC#4** — new test drives the disconnect/cancellation branch, asserts exactly one row, and **fails against the skipped/dropped-write variant**. *Reproduced by the reviewer* (not taken on faith) — see adversarial check.
- [x] **AC#5** — CI lint/type gates green. `ruff check src/ tests/` clean, `ruff format --check src/ tests/` clean, `mypy src/` → `Success: no issues found in 464 source files`. The literal `ruff check .` in the AC text trips **one pre-existing** `RUF034` in `alembic/versions/0023_create_approval_request.py:228` — proven byte-identical on `main` and fired by `main`'s own copy; this PR touches no alembic file. Out of scope, not a regression.

### Project-board alignment
|---|---|---|---|---|---|
| #1599 | In Progress | Closes | ✅ | ✅ | close on merge |

`Closes #1599` present; PR touches exactly the two files #1599 scopes (`audit.py`, `test_api_v1_feed.py`). Parent goal #221 / parent initiative #1575 per the issue body. No delivered-but-unlinked or linked-but-undelivered drift.

### Predicted vs. actual diff
- ✅ `backend/src/meho_backplane/audit.py` — disconnect-arm shield (schedule `_finalize` as a task, drain under `asyncio.shield` in a loop, `write.result()`, `raise cancelled`)
- ✅ `backend/tests/test_api_v1_feed.py` — discriminating `test_double_cancelled_disconnect_still_writes_audit_row`
- 🚫 none
- 🕳️ none

### Integration verification
- [x] `ruff check src/ tests/` — clean (CI scope)
- [x] `ruff format --check src/ tests/` — 936 files already formatted
- [x] `mypy src/` (strict) — Success, 464 files
- [x] `pytest tests/test_api_v1_feed.py tests/test_audit_middleware.py` — 43 passed, 1 skipped (Docker-gated)
- [x] new test re-run with `-W error::RuntimeWarning` — no leaked tasks / unawaited coroutines
- [x] **Adversarial discrimination check (reproduced by reviewer)** — reverted the disconnect arm to a single **unshielded** `await self._finalize(...)` + a stderr probe in the `except` arm. Result: the probe **fired** (`entered except CancelledError arm` → the test genuinely exercises the disconnect branch, *not* the normal path) **and** the test **FAILED** (`AssertionError: disconnect-path audit write did not run to completion`). Restored to pristine PR head; shielded variant passes. The shield is load-bearing and the test discriminates.
- [x] **Drain-loop micro-benchmark (reviewer)** — standalone asyncio model of the exact except-arm loop: under 8 redelivered cancellations `loop_iterations=9` (≈ cancels+1, *not* a busy-spin), the shielded write ran to completion, the drainer returned normally. No 100%-CPU spin; termination guaranteed once the shielded write completes.
- [x] **Starlette 1.0.0 version claim verified against installed source** — `StreamingResponse.__call__`: `spec_version >= (2,4)` awaits `stream_response(send)` directly (a task-level cancel propagates straight into the middleware's `except` arm); `spec_version < (2,4)` wraps it in `anyio.create_task_group()` with a disconnect listener that `cancel_scope.cancel()`s, which **absorbs** the `CancelledError` so the response returns on the *normal* path. The test's `spec_version="2.4"` pin is the correct (and necessary) mechanism, and accurately models how uvicorn signals disconnects under ASGI ≥2.4.
- [ ] ⚠️ not run: migrations replay — n/a (no migration in this PR; alembic untouched)
- [ ] ⚠️ not run: `docker compose up` — not needed for an in-process middleware/asyncio change
- [ ] n/a: OpenAPI/CLI snapshot — no FastAPI route/schema touched; CI `CLI API snapshot freshness` green

None. No blockers, majors, minors, or questions. Nothing posted inline.

- **`write.result()` re-raise vs. `raise cancelled` (latent, unreachable today).** After the drain loop, `write.result()` re-raises any exception `_finalize` raised, which would *replace* `raise cancelled` (cancellation no longer honored). This is **unreachable on the disconnect path today**: `_run_inner_app_buffered` catches only `except Exception`, so a disconnect's `CancelledError` never populates `handler_exc`; and on the streamed branch `_finalize` returns `None` on both audit-write success and audit-write failure (it `raise`s only when `handler_exc is not None`). So `write.result()` is a no-op return and `raise cancelled` always wins — exactly as the inline comment documents. The coupling is *incidental*, not defended: if a future change ever let `_finalize` raise on the streamed path (e.g. `_publish_broadcast_event` made raising, or the inner buffer widened to `except BaseException`), this site would silently convert a disconnect into that exception and drop the `CancelledError`. A one-line guard (log a streamed-path `_finalize` failure but still let `raise cancelled` win) would make the invariant defensive. Out of scope for #1599.
- **No `docs/codebase/audit_middleware.md`** for the write-path/SSE-streaming internals (`docs/architecture/audit.md` + `docs/codebase/audit_query.md` cover the query/replay *read* surface only). Pre-existing gap — the streaming/disconnect mechanism shipped in #1389/#1585, and the module already carries a ~90-line docstring plus dense inline comments. Not introduced or worsened here.

- Adopted: 0
- Disputed: 0
- Out of scope: 0
- CodeRabbit produced no review for this PR (CI status `Review skipped`). Verdict rests on the reviewer's own analysis + reproduced empirical checks + green CI.

### Docs
- [x] `docs/architecture/audit.md` / `docs/codebase/audit_query.md` — verified unchanged-and-still-accurate (this PR strengthens the "exactly one row" / fail-closed write-path guarantee without altering the documented query/replay contract or the audit-row shape).

- The drain-loop is *more robust* than the mechanism #1599 itself sketched: the issue's suggested single `await write` would not survive a redelivered/level-triggered cancel, whereas the `while not write.done(): await shield(write)` loop does. The fix improved on its own brief and the inline comment explains exactly why.
- The test is genuinely discriminating and the harness reasoning is precise: pinning `spec_version="2.4"` to dodge Starlette's disconnect-listener cancel scope, and delivering the second cancel from *inside* the instrumented `_write_audit_row` (the only `_finalize` caller for a never-returning generator), is the right way to land squarely in the `except` arm with no race against the normal path. Verified — the probe lands in the except arm exactly as claimed.

PR #1706, opened 2026-06-12
fix(ui): echo the modal-minted CSRF token from the memory create form (#1693)

damir-topic:
## Summary
- Every Memory create-modal POST 403'd (`csrf_token_invalid`) in the v0.14.0 cycle-10 dogfood, invisibly to the operator (`hx-swap="none"` discards the error response). The form now declares `hx-headers='{"X-CSRF-Token": "{{ csrf_token }}"}'` on `<form id="memory-create-form">`, echoing the token minted by the same `GET /ui/memory/create` render that sets the `meho_csrf` cookie — the double-submit pair can no longer drift apart.
- Root-cause note (refines the issue text): htmx 2.0.9 **does** inherit `hx-headers` from ancestors (verified against the vendored `htmx.min.js` and https://htmx.org/attributes/hx-headers/ — "a child declaration of a header overrides a parent declaration"). The actual desync: the modal GET re-mints + re-sets the cookie (`create.py` / `set_csrf_cookie`), so the token inherited from `index.html`'s page-level directive was the stale list-render token. The prescribed fix is correct under either mechanism — the form-level child declaration wins and is minted token-for-token with the cookie. Same desync class previously fixed for the runbooks list fragment ("B1") and the KB editor modal.
- Bonus from the same attribute: the body textarea's debounced preview `hx-post` (a child of the form) now inherits the fresh token from the form instead of the stale page-level one.
- Stale comments that claimed the page-level directive covers the modal are rewritten (`_create_modal.html` header comment, `render_create_modal` docstring, `docs/codebase/ui.md` modal-conventions + tests bullets).

## Test plan
- [x] `uv run ruff check src/ tests/` — clean
- [x] `uv run ruff format --check src/ tests/` — 950 files already formatted
- [x] `uv run mypy src/` — no issues in 468 files
- [x] `uv run pytest tests/test_ui_memory_create_promote.py tests/test_ui_memory_list.py` — 53 passed
- [x] AC1 — `hx-headers` present on `<form id="memory-create-form">`: template diff + new test asserts exactly one `X-CSRF-Token` echo in the rendered modal fragment
[...]

Issue #1696, opened 2026-06-12
Task: Update dashboard feed endpoint — use session-gated proxy instead of Bearer-only API


damir-topic:
Parent goal: #221
Parent initiative: #1691

**Severity:** SEV-2 · **Area:** frontend · **Source:** claude-rdc-hetzner-dc v0.14.0 cycle-10 dogfood

## Problem

Dashboard's recent-activity widget displays a 'Connecting to live feed...' placeholder indefinitely instead of streaming live broadcast events. Root cause: the dashboard hardcodes `sse-connect=/api/v1/feed`, which is a Bearer-token-authenticated endpoint. Browser `EventSource` API cannot send custom `Authorization` headers (only cookies), so all SSE connections fail with HTTP 401 and tighten into reconnect loops.

## Current state

**Dashboard route** (`backend/src/meho_backplane/ui/routes/dashboard.py:168`):
```python
"feed_endpoint": "/api/v1/feed",
```

**Dashboard template** (`backend/src/meho_backplane/ui/templates/dashboard.html:118`):
```html
sse-connect="{{ feed_endpoint }}"
```

**`/api/v1/feed` endpoint** (`backend/src/meho_backplane/api/v1/feed.py:825-836`) requires:
- `Authorization: Bearer <jwt>` header (enforced by `verify_jwt` → `require_role(TenantRole.OPERATOR)`)
- Therefore, unauthenticated EventSource connections receive HTTP 401

**Session-gated bridge already exists** (`backend/src/meho_backplane/ui/routes/broadcast/stream.py`):
- Route: `/ui/broadcast/stream`
- Authentication: UISessionMiddleware (cookie-based, same as `/ui/`)
- SSE frame format: byte-compatible with `/api/v1/feed` per stream.py docstring (lines 41-52)
- Already deployed and working for `/ui/broadcast` surface (used since G10.1-T1 #867)

**Code comments acknowledge the limitation** (stream.py:25-29):
> "The chassis dashboard's recent-activity snippet (#866) wired `sse-connect="/api/v1/feed"` directly; that wiring is inert for the same reason...G10.1's live feed is the first surface that must actually stream, so it routes through this UI-owned bridge instead."

The comment was written but the actual wiring was never corrected.

## Desired state

[...]

Issue #869, opened 2026-05-21
Task: G10.1-T3 Broadcast wall-monitor mode (?wall=1) + Last-24h replay tab + long-display session refresh + cross-tenant isolation


damir-topic:
Parent initiative: #338
Parent goal: #336
Depends-on: #868 (wall-monitor + replay extend the filtered feed + drawer)

## Target

- Add: wall-monitor layout (`/ui/broadcast?wall=1`) + `/ui/broadcast/history` (replay tab) routes + templates.
- Add: `backend/tests/test_ui_broadcast_wall_replay.py`.
- Precedent: #867 feed + #868 filters; `/api/v1/feed?since=24h` replay (G6.1-T4); `base.html` chrome (#863) for the no-chrome variant.

## Context

Per #338 work-items #5, #6. Wall-monitor (`?wall=1`) renders a minimal no-sidebar/no-topbar maximized auto-scrolling feed for the team-room monitor; session middleware silent-refreshes the access token so the display never logs out. Replay-history "Last 24h" tab HTMX-pulls `/api/v1/feed?since=24h` and renders row fragments; clicking a row jumps to live at that timestamp.

## Current state

The feed + filters + drawer (#867/#868) exist but only as the live in-chrome view.

## Desired state

`?wall=1` → minimal layout, larger rows, auto-scroll. Replay tab pulls last 24h via `?since=24h`; scroll-back; row-click jumps to live at timestamp. Long-display token silent-refresh verified.

## Acceptance criteria

- [ ] `/ui/broadcast?wall=1` hides sidebar + top bar, maximizes the feed, auto-scrolls.
- [ ] Wall mode survives an access-token expiry without logging out (session silent-refresh — assert across a forced expiry).
- [ ] The "Last 24h" replay tab pulls `/api/v1/feed?since=24h` and renders historical rows; row-click jumps to live at that timestamp.
- [ ] Cross-tenant isolation holds in wall + replay (tenant-A events absent from tenant-B).
- [ ] Goal #336 **G10.1** checklist line ticked; `ruff` + `mypy` clean; `pytest -n auto backend/tests/test_ui_broadcast_wall_replay.py` passes.

## Out of scope

- Live feed/filters/drawer — #867/#868. Replay > 24h — G8 audit query. Notification rules — per #338 out-of-scope.

## References

[...]

Issue #864, opened 2026-05-21
Task: G10.0-T3 BFF session storage — web_session table + ORM + encrypted token custody + refresh-token rotation (RFC 9700)


damir-topic:
Parent initiative: #337
Parent goal: #336
Depends-on: #863 (the ui/ module must exist for ui/auth/ session storage to live in)

## Target

- Add: Alembic migration under `backend/alembic/versions/` creating the `web_session` table; ORM model in `backend/src/meho_backplane/db/models.py`.
- Add: `backend/src/meho_backplane/ui/auth/session_store.py` (create/load/revoke session; encrypt access+refresh tokens at rest; refresh-token rotation per RFC 9700).
- Add: `backend/tests/test_ui_session_store.py`.
- Precedent: web/audit table migration pattern ([`0007_create_topology_graph.py`](../blob/main/backend/alembic/versions/0007_create_topology_graph.py), `AuditLog` at [`db/models.py:331`](../blob/main/backend/src/meho_backplane/db/models.py#L331)); existing Vault/`hvac` chain; `cryptography.fernet` for at-rest encryption.

## Context

Per #337 work-item #4 + decision #11, BFF keeps tokens **server-side**: a Postgres `web_session` row holds encrypted access+refresh tokens; the browser holds only the session cookie. Refresh-token rotation per RFC 9700 — one-time-use refresh tokens; replay revokes the session + surfaces an audit row. This Task ships the storage layer + rotation; the login/callback flow + middleware are T4.

## Current state

`ui/` module exists (#863) but has no session storage; no `web_session` table.

## Desired state

`web_session` migration: `id (uuid PK)`, `operator_id`, `tenant_id`, `created_at`, `expires_at`, `access_token` (encrypted), `refresh_token` (encrypted), `last_seen_at`. ORM model. `session_store.py`: `create_session`, `load_session(cookie_id)`, `revoke_session`, `rotate_refresh` (one-time-use; replay → revoke + audit row). Tokens encrypted at rest (`cryptography.fernet`, key via the Vault-managed chain). Reversible migration.

## Acceptance criteria

- [ ] `alembic upgrade`/`downgrade` round-trips `web_session` (test).
- [ ] create/load/revoke session round-trips; tokens stored encrypted (the DB column never holds plaintext — assert).
- [ ] refresh rotation: a used refresh token cannot be reused; replay revokes the session AND writes an audit row (assert).
- [ ] expired session (`expires_at` past) → `load_session` returns `None`.
[...]

PR #1711, opened 2026-06-12
fix(ui): silent token refresh for UI routes — expired sessions re-auth instead of dead-ending on JSON (#1694)

### LINKED ISSUE #1694: Task: Implement token refresh lifecycle for UI routes — handle access-token expiry with silent re-auth redirect
Parent goal: #221
Parent initiative: #1691

**Severity:** SEV-2 · **Area:** frontend · **Source:** claude-rdc-hetzner-dc v0.14.0 cycle-10 dogfood

## Problem

The BFF session-middleware and `require_ui_admin` dependency validate the session cookie and load the encrypted access token from the `web_session` row, but they never rotate it when it expires. When Keycloak's default 5-minute access-token TTL elapses, any subsequent route handler that calls `require_ui_admin` (or any endpoint that validates the stored token) receives an HTTPException(401, detail="token_expired"). FastAPI converts this to raw JSON `{"detail": "token_expired"}` sent as `application/json`, breaking the HTML UI. The operator sees a blank page with JSON instead of a seamless re-auth redirect or a transparent token refresh.

The BFF holds a refresh token in encrypted storage (`web_session.refresh_token`) and the `rotate_refresh` primitive exists in the session-store module (RFC 9700 § 4.14 compliant), but the refresh pathway is never invoked on the request hot path. This is a design gap: the middleware loads the session row but never checks the access token's expiry or attempts rotation before the token is presented to a downstream handler.

## Current state

- **Session loading** (middleware.py:277 / session_store.py:380): The middleware loads the session row and checks only that the row itself is not revoked and not past `expires_at`. It does NOT check the expiry of the encrypted access token inside the row.
- **Token validation at handler time** (middleware.py:492 / middleware.py:507): `require_ui_admin` loads the session row again and calls `verify_jwt_for_audience(Bearer <access_token>)`. If the token is expired, `verify_jwt_for_audience` raises HTTPException(401, detail="token_expired") at auth/jwt.py:505.
- **No refresh attempt**: There is no code path between the session load and token validation that attempts to refresh the access token using the stored refresh token.
- **Error response shape**: FastAPI's default exception handler converts HTTPException(401) to JSON `{"detail": "token_expired"}` regardless of the request's Accept header or the route's response_class. Operators see raw JSON instead of HTML or a redirect.

Evidence:
- backend/src/meho_backplane/ui/auth/middleware.py:55–65 (comment): "T5 (#866) will register a separate path that performs the RFC-9700-compliant refresh-token rotation... the middleware only loads + validates the existing session row."
- backend/src/meho_backplane/ui/auth/session_store.py:480–640: `rotate_refresh` function exists but is not called anywhere in the UI request path (verified via grep).
- backend/src/meho_backplane/ui/auth/middleware.py:506–517: `require_ui_admin` calls `verify_jwt_for_audience` on a token that may be expired; no refresh attempt precedes it.

## Desired state

When a route handler depends on `require_ui_admin` (or any access-token validation) and the token is expired or near expiry:

1. The middleware or a wrapper around token validation detects the expiry condition.
2. The wrapper attempts a silent refresh: call Keycloak's token endpoint with the stored refresh token.
3. On successful refresh:
   - Call `rotate_refresh(session, session_id, presented_refresh=old_refresh, new_access_token=..., new_refresh_token=...)` to update the session row atomically.
   - Extend `expires_at` so the session remains valid for the full new token lifetime.
   - Continue processing the original request with the fresh token (no re-execution; the handler sees the new token implicitly because the session row was updated).
   - Emit a structlog event `ui_auth_token_refresh_succeeded` with the session_id and new expires_at.
4. On refresh failure (invalid refresh token, Keycloak unreachable, network timeout):
   - Emit a structlog event `ui_auth_token_refresh_failed` with the reason code.
   - Raise HTTPException(401) with a detail code that maps to a redirect to `/ui/auth/login` for HTML routes (not raw JSON).
   - Clear the session cookie so the operator lands back at login.

The refresh is transparent to the route handler: if it succeeds, the handler runs as normal. If it fails, the operator sees a re-auth flow instead of raw JSON.

## Acceptance criteria

- [ ] `require_ui_admin` (or a new wrapper middleware) checks the session's `expires_at` before calling `verify_jwt_for_audience`. When the row is within 60 seconds of expiry, a refresh is attempted.
- [ ] The refresh attempt constructs a token-endpoint request with `grant_type=refresh_token&refresh_token=<stored_token>` (RFC 6749 § 6) and POSTs to Keycloak's token endpoint (resolved via OIDC discovery, cached). Request timeout is 5 seconds (matching `_HTTP_TIMEOUT_SECONDS` in flow.py).
- [ ] On successful refresh, `rotate_refresh` is called to update the session row (one-time-use semantics per RFC 9700 § 4.14). The session row's `expires_at` is extended by the new token's `expires_in` minus 60-second margin (matching the margin at login in routes.py:393).
- [ ] On refresh failure, the reason is logged (`ui_auth_token_refresh_failed` with fields: session_id, reason: "invalid_grant" | "network_error" | "timeout" | "malformed_response"), and HTTPException(401, detail="session_expired") is raised (a new detail code distinct from "token_expired" so the error-handler middleware can map it to a redirect).
- [ ] Error-handler middleware intercepts HTTPException with detail="session_expired" and returns a 302 redirect to `/ui/auth/login?return_to=<current_path>` for HTML requests (Accept: text/html). For JSON requests, return `{"detail": "session_expired"}` as normal.
- [ ] The refresh attempt does NOT retry on failure; a single failed refresh causes the 401. (Fail-closed design: no exponential backoff or jitter tuning knobs.)
- [ ] Concurrent requests on the same session are serialized: if two requests both need a refresh, the first one wins and updates the row; the second sees the new token and skips the refresh attempt. (Use `SELECT ... FOR UPDATE` row locking, mirroring the existing pattern in rotate_refresh:581.)
- [ ] A `ui_auth_token_refresh_succeeded` structlog event is emitted on every successful refresh (for audit / observability). The event includes session_id, old_expires_at, new_expires_at, and time_cost_ms.
- [ ] Refresh does not leak token material to logs: the refresh token, access token, and client secret never appear in any log line. Only structured codes and timestamps.
- [ ] Manual logout (`/ui/auth/logout`) still works and does not trigger a refresh attempt (the logout handler calls `revoke_session`, which sets `revoked_at`; subsequent requests find the session revoked and 302 to login without attempting refresh).

## Out of scope

- Extending the session lifetime beyond the configured absolute cap (`ui_session_absolute_lifetime_seconds`; see session_store.py:692). Refresh extends the session within the existing cap so a 12-hour operator console session is still possible if the cap is set high, but a refresh-alone cannot create an immortal session.
- Refresh-attempt UX: there is no "refreshing..." spinner or client-side notification. The refresh is silent (HTTP-level); if it fails, the operator navigates to login. A future task can add client-side UX (HTMX polling, Alpine message, etc.).
- Refresh for `/api/*` routes: the chassis API already receives a fresh token on every request (passed via Authorization header by the CLI / MCP client / external integrations). Refresh is a BFF-specific concern for browser-based sessions.
- Refresh-token rotation as a background task: single-request refresh is sufficient. If the token expires mid-request, the refresh happens inline. A scheduled refresh job is a v1 optimization (out of scope for this task).
- Configurable refresh margin (the 60-second threshold is hard-coded). Operators cannot tune when a refresh is attempted. (Dumb substrate philosophy.)

## References

- RFC 6749 § 6 (Authorization Code grant: https://www.rfc-editor.org/rfc/rfc6749#section-6)
- RFC 9700 § 4.14 (OAuth 2.0 Security BCP, refresh token rotation: https://datatracker.ietf.org/doc/rfc9700/)
- draft-ietf-oauth-browser-based-apps § 6 (BFF pattern, silent refresh: https://datatracker.ietf.org/doc/draft-ietf-oauth-browser-based-apps/)
- OWASP ASVS v4 §3.3.1 (session timeout: idle vs absolute: https://owasp.org/www-project-application-security-verification-standard/)
- Keycloak Token Endpoint (form-based token exchange with refresh_token grant): https://www.keycloak.org/docs/latest/server_admin/index.html#token-endpoint
- meho backend/src/meho_backplane/ui/auth/session_store.py:480 (rotate_refresh — RFC 9700 § 4.14 primitive already present)
- meho backend/src/meho_backplane/ui/auth/routes.py:393 (session TTL margin calculation at login)
- meho #869 (G10.1-T3, sliding-session extension): precedent for mutating session row `expires_at` within the request path
- Consumer signal: claude-rdc-hetzner-dc v0.14.0 cycle-10 feedback (FE-3)

damir-topic:
## Summary

- **Silent token refresh for UI routes (FE-3, cycle-10 dogfood).** The BFF held a refresh token and the RFC 9700 § 4.14 `rotate_refresh` primitive since T3 (#864), but nothing on the request hot path ever invoked it — when Keycloak's ~5-minute access token expired inside a still-alive (sliding-extended) `web_session` row, `require_ui_admin` dead-ended the browser on raw JSON `{"detail": "token_expired"}`. New `ui/auth/refresh.py` closes the loop: a proactive leg (row within 60 s of `expires_at`, the AC-literal check) and a reactive leg (JWT chain reports `token_expired`) both funnel into one `SELECT ... FOR UPDATE`-serialised chokepoint that POSTs the RFC 6749 § 6 refresh grant (single attempt, 5 s timeout, no retry) and rotates the row one-time-use, re-extending the session lifetime within the absolute cap.
- **Re-auth redirect instead of raw JSON on terminal failure.** New `ui/auth/errors.py` registers an app-level `HTTPException` handler (FastAPI override-and-delegate pattern): exactly `401 session_expired` on `/ui/*` with an HTML `Accept` becomes `302 /ui/auth/login?return_to=<page>` + `meho_session` cookie clear; JSON callers keep `{"detail": "session_expired"}` (cookie cleared too); every other HTTPException delegates byte-for-byte to FastAPI's stock handler, so `/api/*` 401 codes are untouched.
- **Why both legs (design note).** The AC's literal proactive check keys on the *row's* clock, but under the default `ui_session_sliding_extension_seconds=3600` the row deliberately outlives the token by hours — the proactive check alone can never observe token expiry there (it matters for sliding-disabled deploys and sessions near the absolute cap). The reactive catch on the JWT chain's own `token_expired` verdict is what clears the dogfood scenario; the issue's desired state ("the token is expired or near expiry") demands both. Documented in `refresh.py`'s module docstring.
- **No cookie desync (the #1706 class).** A refresh mutates only the encrypted token columns + `expires_at` inside the existing row — zero `Set-Cookie` operations. `meho_session` (the row id) and `meho_csrf` (HMAC-keyed on that same id) are byte-identical before/after rotation, so in-flight pages and already-rendered forms keep working. The only cookie mutation in the lifecycle is the clear on terminal failure, at which point the operator re-authenticates anyway.

## Implementation notes

- `flow.refresh_access_token` uses authlib 1.7.2's `AsyncOAuth2Client.refresh_token` (introspected against the installed wheel: error JSON → `OAuthError(.error="invalid_grant")`; a 200 omitting `refresh_token` backfills the presented one per RFC 6749 § 6's old-token-stays-valid branch — rotation to the same plaintext keeps the one-time-use bookkeeping without special-casing; non-OAuth error pages raise `httpx.HTTPStatusError`).
- Failure reasons map per the AC enum: `invalid_grant` / `network_error` / `timeout` / `malformed_response` (+ one off-contract code `oauth_not_configured` for the client-credentials-vanished-after-login edge, and `session_expired_during_refresh` / `rotation_replay` for two lock-window edges; all documented in `_classify_refresh_failure` / `refresh_session_tokens`). `httpx.TimeoutException` is matched before `httpx.HTTPError` (subclass), `OAuthFlowConfigurationError` before `OAuthFlowError` (subclass).
- `rotate_refresh` gained an optional `new_lifetime=` that extends `expires_at` monotonically (never backwards past a slid expiry — shrinking would log out the wall-monitor sessions #869 protects) and clamped to `created_at + ui_session_absolute_lifetime_seconds`, so chained refreshes cannot mint an immortal session (issue out-of-scope guarantee). Default `None` preserves the pre-#1694 contract; existing tests pass unchanged.
- The wall-clock re-check between the token-endpoint POST and `rotate_refresh` is deliberate deadlock avoidance: `rotate_refresh`'s replay branch commits revoke+audit on a *dedicated* DB session — an UPDATE that would wait on the very row lock the refresh transaction holds. The pre-checks make that branch unreachable from this call site (see adjacent finding below).
- `load_session_for_update` is a new side-effect-free session-store primitive (no `last_seen_at` bump, no sliding) — the refresh decision is made against stable lock-protected values; `rotate_refresh` owns the mutation.
- Seam for #1696 (dashboard feed → session-gated proxy, next wave): `load_fresh_session` + `verify_access_token_with_refresh` / `refresh_session_tokens` are exported from `meho_backplane.ui.auth`; the two-step contract (proactive load, reactive refresh-on-401) is documented in the module docstring and `docs/codebase/ui.md`.

## Test plan

- [x] `uv run mypy src/` (strict) — clean, 470 files
- [x] `python3 .claude/hooks/code-quality.py --diff` — 0 blockers (22 warnings, all pre-existing debt in touched files + two 6-arg PLR0913 warns matching the existing `create_session` precedent)
- [x] `uv run pytest tests/test_ui_token_refresh.py` — 14 passed (new suite)
- [x] `uv run pytest tests/test_ui_auth_flow.py tests/test_ui_session_store.py` — 47 passed (no regressions in the adjacent auth suites)
- [x] Full unit sweep `pytest -n 6 --dist loadscope --ignore=tests/integration` — green (see CI)
- [x] CLI snapshot: `cd cli && make snapshot-openapi` → zero diff (no OpenAPI surface change; handler + refresh path are not schema-visible)

Acceptance-criteria mapping (all proven by the new suite):

| AC | Test |
| --- | --- |
| Proactive check within 60 s of `expires_at` | `test_proactive_refresh_when_row_near_expiry` (sliding pinned to 0) |
| RFC 6749 § 6 grant shape, discovery-resolved endpoint, 5 s timeout | `test_expired_access_token_refreshes_silently_and_serves` asserts `grant_type=refresh_token&refresh_token=<stored>` form body; timeout shares `_HTTP_TIMEOUT_SECONDS=5.0` with the login flow |
| `rotate_refresh` one-time-use + `expires_at` extended by `expires_in − 60` | same two tests + `test_refresh_extension_clamps_at_absolute_lifetime_ceiling`, `test_refresh_never_shrinks_a_slid_expires_at` |
| Failure reasons logged + `401 session_expired` | `test_invalid_grant_redirects_html_to_login_and_clears_cookie`, `test_refresh_failure_reasons_map_per_contract` (timeout / network_error / malformed_response) |
| Error handler: 302 for HTML, JSON for JSON | `test_invalid_grant_redirects_html_to_login_and_clears_cookie`, `test_refresh_failure_keeps_json_shape_for_non_html_callers`, `test_non_session_expired_401_keeps_default_json_shape` |
| No retry on failure | `token_route.call_count == 1` asserted on every failure test |
| Concurrent refreshes serialised, second skips | `SELECT ... FOR UPDATE` in `load_session_for_update` (mirrors `rotate_refresh`); deterministic loser-path test `test_concurrent_loser_skips_token_endpoint_and_returns_stored_pair` (true cross-transaction blocking is not observable on the SQLite test path — same documented caveat as the existing rotation tests) |
| `ui_auth_token_refresh_succeeded` with session_id / old / new expires_at / time_cost_ms | `test_refresh_success_event_fields_and_no_token_leakage` |
| No token material in logs | same test asserts neither access tokens, refresh tokens, nor the client secret appear in any captured event; flow-level refresh logs nothing |
| Logout untouched, no refresh on revoked rows | `test_revoked_session_redirects_via_middleware_without_refresh` |

## Out of scope (per issue)

- Client-side refresh UX (spinner/HTMX polling) — future task; htmx fragment requests without an HTML `Accept` get the JSON 401 shape by design.
- `/api/*` refresh, background refresh job, configurable margin — all explicitly out of scope in #1694.

## Adjacent findings (not addressed here)

- **Latent Postgres deadlock in `rotate_refresh`'s replay branch**: when a caller holds the `FOR UPDATE` lock (as any transactional caller does — the function itself takes it), the replay branch's dedicated-session revoke UPDATE waits on that same lock while the caller awaits the dedicated commit — a circular wait. Unreachable from this PR's call site (pre-checks above) and from the historical no-callers state; latent on SQLite (different locking model). Worth a follow-up ticket.
- `require_ui_admin`'s session-gone race (`401 ui_session_required`) still renders raw JSON for HTML browsers; the middleware normally 302s first, so the window is the revoke-between-middleware-and-dependency gap only. The AC pinned the redirect mapping to `session_expired` exclusively; extending it is a one-line follow-up if wanted.
- `_maybe_extend_expiry` is 63 lines (warn threshold 60) — pre-existing.

Closes #1694

## Release Notes

* **New Features**
  * Operator console sessions now handle mid-session token expiry gracefully by automatically refreshing tokens in the background.
  * If token refresh fails, users are redirected to login instead of encountering an error state.

* **Tests**
  * Added comprehensive test coverage for session token refresh scenarios.

* **Documentation**
  * Updated authentication documentation to describe token refresh behavior.

---

## Follow-up (auto-implement-issue, iteration 1, 2026-06-13)

Addressed both findings from the iteration-1 review (`.claude/auto/review-results/pr-1711-iter-1.json`; verdict comment [here](https://github.com/evoila/meho/pull/1711#issuecomment-4696750465)):

- **B1 (blocker)** `2f088844` — `_refresh_via_idp` now catches `ValueError` (`json.JSONDecodeError`'s base): non-JSON token-endpoint bodies with any sub-500 status (200 included) map to the `malformed_response` failure reason, emit `ui_auth_token_refresh_failed`, and fail closed to the login redirect + cookie clear instead of escaping as an unhandled 500. `non-json-200` / `non-json-400` cases plus a cookie-clear assertion added to `test_refresh_failure_reasons_map_per_contract` (red-proofed against the old catch tuple).
- **M1 (major)** `bc053075` — one clock through the expiry gates: `rotate_refresh` gains optional `now=` (default wall clock; existing callers unaffected), and `_rotate_or_fail` shares its single pre-check reading with the rotation's internal replay gate — the self-deadlocking "expired" replay branch is now provably unreachable from the inline refresh path. New `test_rotate_refresh_honours_caller_supplied_clock` pins the contract in both directions; `docs/codebase/ui.md` documents the `now=` seam.

Disputed: none. Deferred: none.


damir-topic:
@coderabbitai review

damir-topic:
@coderabbitai full review

backend/src/meho_backplane/ui/auth/refresh.py
```diff
@@ -0,0 +1,461 @@
...
+    """Step 3: the single token-endpoint attempt, failure-mapped.
+
+    Every exception class the grant can raise collapses to one
+    ``ui_auth_token_refresh_failed`` event (structured ``reason`` per
+    :func:`_classify_refresh_failure`) and the fail-closed
+    ``session_expired`` 401. No retry.
+    """
+    try:
+        return await refresh_access_token(refresh_token=locked.refresh_token)
+    except (httpx.HTTPError, OAuthError, OAuthFlowError) as exc:
```
damir-topic: **[🛑 B1]** — Non-JSON token-endpoint responses escape the fail-closed mapping: `json.JSONDecodeError` is not in this catch tuple → unhandled 500 instead of `401 session_expired`

**What:** authlib 1.7.2's `parse_response_token` calls `resp.json()` before any error handling for responses with status < 500 (`raise_for_status()` fires only on >= 500). A token-endpoint response of 400/404 — or 200 — with a non-JSON body (ingress default backend, path-rewriting proxy, maintenance page in front of Keycloak) raises `json.JSONDecodeError`, which subclasses `ValueError`, not `httpx.HTTPError` / `OAuthError` / `OAuthFlowError`. It sails past this `except`, out of `refresh_session_tokens`, and surfaces as an unhandled 500 — no `ui_auth_token_refresh_failed` event, no `session_expired` mapping, no login redirect. Verified empirically against the installed wheels (authlib 1.7.2 / httpx 0.28.1): `400 text/html` → `json.decoder.JSONDecodeError`; `200 text/html` → same; `400` + error JSON → `OAuthError`; `502` → `httpx.HTTPStatusError`. The same gap covers `_fetch_discovery_doc`'s `response.json()` on a 200 non-JSON discovery doc, which funnels through this very call path.

**Why it matters:** the AC pins the failure contract — "On refresh failure, the reason is logged (… `malformed_response`) and HTTPException(401, detail=`session_expired`) is raised" — and a non-JSON body is the most literal form of a malformed response. The operator-facing symptom is a raw 500 error body at the browser: exactly the dead-end class FE-3 exists to remove, reintroduced for the misbehaving-proxy case.

**Required change:** add `ValueError` to the tuple — `except (httpx.HTTPError, OAuthError, OAuthFlowError, ValueError) as exc:` (`json.JSONDecodeError` subclasses `ValueError`; `_classify_refresh_failure`'s residual branch already returns `"malformed_response"`, so no classifier change is needed beyond a docstring note). Add one parametrized case to `test_refresh_failure_reasons_map_per_contract`: `httpx.Response(400, text="<html>err</html>", headers={"content-type": "text/html"})` → reason `malformed_response`, 302 + cookie clear, `call_count == 1`.

**Source:** installed `authlib==1.7.2`, `authlib/oauth2/client.py::OAuth2Client.parse_response_token` (`if resp.status_code >= 500: resp.raise_for_status()` then `token = resp.json()`); https://docs.authlib.org/en/latest/client/oauth2.html

backend/src/meho_backplane/ui/auth/refresh.py
```diff
@@ -0,0 +1,461 @@
...
+) -> DecryptedSession:
+    """Steps 4-5: wall-clock re-check, then the RFC 9700 rotation.
+
+    The re-check keeps ``rotate_refresh``'s replay branch unreachable
+    from this call site: that branch commits revoke + audit side
+    effects on a *dedicated* DB session -- an UPDATE that would wait
+    on the very row lock this transaction holds (see the
+    :func:`refresh_session_tokens` docstring, step 4).
+    """
+    if locked.expires_at <= datetime.now(UTC):
```
damir-topic: **[⚠️ M1]** — Expiry pre-check and `rotate_refresh`'s replay gate read two different clocks — the self-deadlocking "expired" replay branch is still reachable through the gap

**What:** this pre-check evaluates `datetime.now(UTC)` here; `rotate_refresh` then captures its own `now = datetime.now(UTC)` and re-evaluates `expires_at <= now` in `_locked_row_or_replay`. If `expires_at` lands in the microseconds between the two reads, the "expired" replay branch fires — and its `_commit_replay_side_effects` revoke UPDATE on a dedicated DB session blocks on the very row lock this transaction holds. Postgres' deadlock detector never fires (this transaction isn't waiting on a DB lock — it's awaiting the dedicated session's commit in Python), so the request hangs indefinitely and the still-locked row blocks every subsequent refresh for that session. The mismatch and revoked branches are genuinely unreachable (both lock-protected); expiry is the one gate whose input is a wall clock the lock can't freeze.

**Why it matters:** the module's deadlock-avoidance argument ("the pre-checks make that branch unreachable from this call site") is load-bearing for running rotation inside a transactional caller — this PR introduces the first such caller — and it currently has a microsecond hole whose consequence is an unbounded hang in the auth hot path, not a clean 401.

**Required change:** make both gates evaluate the same instant: give `rotate_refresh` an optional `now: datetime | None = None` parameter (defaulting to `datetime.now(UTC)`, preserving the existing contract for all current callers), pass it through to `_locked_row_or_replay` / `_apply_rotation`, and have `_rotate_or_fail` capture `now` once, use it for this pre-check, and hand the same value to `rotate_refresh`. Then `expires_at > now` at the pre-check implies `is_expired` is false inside the rotation by construction, and the docstring's unreachability claim becomes exact rather than approximate.

damir-topic:




**Anchor task:** #1694 — token refresh lifecycle for UI routes: silent re-auth instead of raw-JSON dead-end (G0.25 FE-3)
**Surface area:** backend (BFF auth — Python/FastAPI)
**Best-practices packs audited:** python (pack file absent in slim-skill layout; audited against repo-enforced conventions — ruff strict, mypy strict, structlog event discipline, async SQLAlchemy patterns)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `c785cd412831e9b572201a8335922010c4b5a9bf`
**AI reviewer (CodeRabbit):** not received — `@coderabbitai full review` re-triggered 2026-06-13T00:13:54Z after the 2026-06-12 rate-limit decline; bot acked "Full review finished" in 6 s but published zero review objects, zero threads, zero inline findings within the 15-min window (org usage credits exhausted per its own earlier warning)

### Acceptance criteria
- [x] Proactive check: row within 60 s of `expires_at` → refresh before verify — `load_fresh_session` + `test_proactive_refresh_when_row_near_expiry` (ran locally, green)
- [x] RFC 6749 § 6 grant shape, discovery-resolved endpoint, 5 s timeout — form body asserted in tests; `_HTTP_TIMEOUT_SECONDS = 5.0` shared with the login flow (flow.py:135, client timeout flow.py:336)
- [x] `rotate_refresh` one-time-use + `expires_at` extended by `expires_in − 60`, clamped + monotonic — three dedicated tests; verified against the authlib 1.7.2 refresh-token backfill branch (`if "refresh_token" not in token: self.token["refresh_token"] = refresh_token` — old-token-echo handled exactly as the PR claims)
- [❌] Failure reasons logged + `401 session_expired` raised on every refresh failure — holds for JSON-parseable failures (`invalid_grant`/`timeout`/`network_error`/missing-field `malformed_response`, all test-proven), but a non-JSON token-endpoint response with status < 500 raises `json.JSONDecodeError`, escapes `_refresh_via_idp`'s catch tuple, and 500s with no failure event and no redirect — see 🛑 B1
- [x] Error handler: 302 + cookie clear for HTML `Accept`, JSON shape for non-HTML, everything else byte-for-byte FastAPI default — three tests incl. the `token_expired`-still-JSON scoping control; verified `app.add_exception_handler(StarletteHTTPException, …)` replaces the stock registration on fastapi 0.136.3 and that `fastapi.HTTPException` reaches it via MRO lookup
- [x] No retry — `call_count == 1` asserted on every failure test
- [x] Concurrent refreshes serialised, second skips — `load_session_for_update` (`SELECT … FOR UPDATE`) + stored-vs-presented access-token skip check; deterministic loser-path test; SQLite caveat documented, mirroring the existing `rotate_refresh` precedent
- [x] `ui_auth_token_refresh_succeeded` with session_id / old / new expires_at / time_cost_ms — test-proven, emitted post-commit
- [x] No token material in logs — test asserts both access tokens, both refresh tokens, and the client secret are absent from all captured events; failure logs carry only `error_class` + RFC 6749 § 5.2 `idp_error` code
- [x] Logout untouched; revoked rows never reach the refresh path — test-proven (middleware 302s first, zero token-endpoint calls)

### Project-board alignment
|---|---|---|---|---|---|
| #1694 | Open (no project item found) | Closes | ✅ | ⚠️ full minus B1/M1 | fix loop, then close on merge |

### Predicted vs. actual diff
- ✅ `refresh.py` (new orchestrator), `errors.py` (new handler), `flow.py` (`refresh_access_token`), `session_store.py` (`load_session_for_update`, `rotate_refresh(new_lifetime=)`), `middleware.py` (`require_ui_admin` wiring), `routes.py` (margin + `clear_session_cookie` publicised), `main.py` (handler registration), `__init__.py` (exports), tests, `docs/codebase/ui.md`, CHANGELOG
- ⚠️ `refresh.py` — see inline B1 (catch-tuple gap) and M1 (two-clock expiry gate)
- ❓ justified, disclosed: `_assert_tenant_admin_role` extraction and the `rotate_refresh` → `_locked_row_or_replay`/`_apply_rotation` split (function-size budget; behaviour preserved, 47 adjacent auth tests pass unchanged)
- 🚫 none — sibling seams verified untouched: no templates (base.html `component_scripts` #1692), no `csrf.py`/create-modal pairing (#1693/#1706), no datalist (#1695/#1709), no footer (#1698/#1710)
- 🕳️ none

### Integration verification
- [x] `ruff check src/ tests/` + `ruff format --check` — clean (run in review worktree)
- [x] `mypy src/` (strict) — clean, 471 files (run in review worktree)
- [x] `pytest tests/test_ui_token_refresh.py tests/test_ui_auth_flow.py tests/test_ui_session_store.py` — 61 passed (run in review worktree)
- [x] CI rollup on head SHA — all required lanes SUCCESS (Python unit, integration testcontainers, Helm, Go, CodeQL, Semgrep, TruffleHog, DCO, CLI snapshot freshness, backplane image); deploy-smoke SKIPPED as usual for PRs
- [x] Failure-class probe against installed wheels (authlib 1.7.2 / httpx 0.28.1): 400+error-JSON → `OAuthError`, 502 → `HTTPStatusError`, 400/404/200 non-JSON → `json.JSONDecodeError` (the B1 evidence)
- [⚠️ not run] live Keycloak round-trip / browser flow — reasoned about only; token endpoint is respx-mocked in the suite, but the full middleware → dependency → handler → exception-handler path is exercised end-to-end via TestClient
- [x] migrations replay — n/a (no schema change; `web_session` columns pre-exist)

- 🛑 B1 — `refresh.py:307`: non-JSON token-endpoint responses (<500 status, or 200 with a non-JSON body) raise `json.JSONDecodeError`, escape the catch tuple, and 500 instead of the fail-closed `401 session_expired`; required change: add `ValueError` to the except tuple (+ docstring note) and one parametrized `malformed_response` test case for a `400 text/html` response
- ⚠️ M1 — `refresh.py:333`: the wall-clock pre-check and `rotate_refresh`'s internal `now` are two separate clock reads, leaving a microsecond window into the self-deadlocking "expired" replay branch; required change: thread one `now` through (`rotate_refresh(now: datetime | None = None)`) so both gates evaluate the same instant

- During an IdP outage, in-flight requests on one session serialise one 5 s timeout each behind the row lock (no rotation happens, so the skip check never trips). Self-limiting — the first failure clears the cookie and later requests bounce at the middleware — and the AC pins no-retry/no-knobs, so nothing to change here; worth remembering when #1696's polling proxy lands on this seam.
- `require_ui_admin`'s session-gone race still renders raw JSON `401 ui_session_required` for HTML browsers (PR discloses this; the AC pinned the redirect mapping to `session_expired` exclusively). Agree it is a one-line follow-up if wanted, not this PR.
- The latent `rotate_refresh` replay-branch deadlock for transactional callers deserves the follow-up ticket the PR proposes — M1 closes this PR's residual reachability, the primitive-level hazard remains for any future caller that skips the pre-checks.

- Adopted: 0 — no findings were ever published (rate-limited on 2026-06-12; re-trigger acked but produced no review content)
- Disputed: 0
- Out of scope: 0

### Docs
- [x] `docs/codebase/ui.md` — updated (module table rows for `refresh`/`errors`, custody-table lifetimes, new mid-session refresh section); accurate against the diff
- [x] CHANGELOG `[Unreleased]` — one bullet, matches the team format. Note: `mergeable=CONFLICTING` is the known CHANGELOG cascade (main's #1709/#1712 insert at the same hunk; the PR's other overlap, `ui.md`, edits a disjoint region) — a rebase-time matter for the orchestrator, not a defect of this diff

- The two-leg design note (row clock vs token clock under sliding extension) is exactly right, and the cookie-stability argument is structurally airtight: the refresh path constructs no `Response` and performs zero `Set-Cookie` operations, `meho_csrf` is HMAC-keyed on the unchanged session id and re-minted per render — the #1706 desync class genuinely cannot occur here. Verified against `csrf.py` and the dashboard render path, not just the docstring.
- Every authlib behaviour the PR body claims (error-JSON → `OAuthError(.error)`, refresh-token backfill on omission, ≥500 `raise_for_status`) checked out exactly against the installed 1.7.2 wheel — which is also how the one class it missed was found.

backend/src/meho_backplane/ui/auth/refresh.py
damir-topic: Fixed in 2f088844.

- `_refresh_via_idp`'s catch tuple is now `(httpx.HTTPError, OAuthError, OAuthFlowError, ValueError)` — `json.JSONDecodeError` subclasses `ValueError`, so the non-JSON sub-500 shapes authlib's `parse_response_token` lets through (`resp.json()` on everything below 500, 200 included) land in `_classify_refresh_failure`'s residual branch: `malformed_response` reason, `ui_auth_token_refresh_failed` event, fail-closed `session_expired` 401 → login redirect + cookie clear.
- `test_refresh_failure_reasons_map_per_contract` gains `non-json-200` and `non-json-400` cases (text/html bodies) plus a `Max-Age=0` cookie-clear assertion on the redirect path. Red-proofed: with the catch-tuple change stashed, both new cases die on exactly the unhandled `json.decoder.JSONDecodeError` this finding describes.
- Docstring notes added on `_refresh_via_idp` and `_classify_refresh_failure` documenting why `ValueError` is in the tuple.

backend/src/meho_backplane/ui/auth/refresh.py
damir-topic: Fixed in bc053075.

- `rotate_refresh` accepts `now: datetime | None = None` (default `datetime.now(UTC)` — existing callers unaffected, per the #1696 pass-through constraint) and threads it into `_locked_row_or_replay` / `_apply_rotation`.
- `_rotate_or_fail` reads one `now`, uses it for the expiry pre-check, and passes the same value to `rotate_refresh`, so the pre-check and the internal replay gate evaluate the same instant — the "expired" replay branch (whose dedicated-session revoke UPDATE would wait on this transaction's own row lock) is now provably unreachable from the inline refresh path.
- New `test_rotate_refresh_honours_caller_supplied_clock` pins the single-clock contract in both directions: a wall-clock-expired row still rotates under a caller `now` that predates `expires_at` (fails on the old two-clock code), and a caller `now` at `expires_at` fires the expired replay branch (audit reason `expired`).
- `rotate_refresh`'s docstring prose was tightened to absorb the new parameter docs within the 100-line function-size budget — no contract content dropped. `docs/codebase/ui.md` documents the `now=` seam.

damir-topic:
## Fix iteration 1 delta (`/auto-implement-issue --continue-pr`, no-dance)

Both iteration-1 findings addressed on top of `c785cd41` → new head `bc053075`.

| Finding | Commit | Change |
|---|---|---|
| **B1** (blocker) | `2f088844` | `_refresh_via_idp` catches `ValueError` (`json.JSONDecodeError`'s base): non-JSON token-endpoint bodies with any sub-500 status — 200 included — now map to `malformed_response`, emit `ui_auth_token_refresh_failed`, and fail closed to the login redirect + cookie clear instead of escaping as an unhandled 500. New `non-json-200` / `non-json-400` parametrized cases + cookie-clear assertion; red-proofed against the old catch tuple (both cases die on the unhandled `JSONDecodeError` without the fix). |
| **M1** (major) | `bc053075` | One clock through the expiry gates: `rotate_refresh` gains optional `now=` (default `datetime.now(UTC)`, existing callers unaffected), threaded into `_locked_row_or_replay` / `_apply_rotation`; `_rotate_or_fail` shares its single pre-check reading with the rotation's replay gate, making the self-deadlocking "expired" replay branch provably unreachable from the inline refresh path. New `test_rotate_refresh_honours_caller_supplied_clock` pins the contract in both directions. |

Disputed: none. Deferred: none. `docs/codebase/ui.md` updated for the `now=` seam.

**Verification (worktree, full local sweep):**
- `ruff check src/ tests/` + `ruff format --check` — clean
- `mypy src/` (strict) — clean
- `pytest -n 6 --dist loadscope --maxfail=1 --ignore=tests/integration` — **7343 passed, 195 skipped** (standard secret-guard skips)
- Touched modules: `test_ui_token_refresh.py` 16 passed, `test_ui_session_store.py` 14 passed
- code-quality hook (diff mode) — 0 blockers

Re-review (iteration 2) is owned by the orchestrator — CodeRabbit deliberately not re-triggered from this fix run.

damir-topic:
@coderabbitai full review

damir-topic:




**Anchor task:** #1694 — silent token refresh for UI routes
**Surface area:** backend
**Best-practices packs audited:** python (repo-enforced conventions: ruff strict, mypy strict, structlog event discipline, async SQLAlchemy)
**Verdict:** APPROVE
**Head SHA reviewed:** `bc0530752e702a030adb8ed951dd6f679d214645` (= reviewed `c785cd41` + two fix commits, no amend, DCO green)
**AI reviewer (CodeRabbit):** not received — "Full review finished" ack in 6s, zero review objects/threads after 747s (third content-less ack on this PR; org-credit limitation, not treated as a clean bill)

### Iteration-1 findings — both verified resolved at `bc053075`

- **B1 (blocker) — RESOLVED** by `2f088844`. `_refresh_via_idp` now catches `(httpx.HTTPError, OAuthError, OAuthFlowError, ValueError)`; `json.JSONDecodeError` (subclass of `ValueError`, raised by authlib 1.7.2's `parse_response_token` → `resp.json()` on any non-JSON body below 500, 200 included) falls to `_classify_refresh_failure`'s residual branch → `malformed_response` + `ui_auth_token_refresh_failed` + fail-closed 401/302 + cookie clear. Verified three ways: source read, green run of the two new parametrized cases (`non-json-200`, `non-json-400`: 302 to `/ui/auth/login?return_to=…`, `Max-Age=0` cookie clear, `call_count == 1`), and an independent red-proof — with `c785cd41`'s refresh.py restored, both cases fail on unhandled `json.decoder.JSONDecodeError`. The added location/cookie assertions also harden every pre-existing parametrized failure case.
- **M1 (major) — RESOLVED** by `bc053075`. `_rotate_or_fail` reads `datetime.now(UTC)` exactly once; the pre-check (`locked.expires_at <= now`) and `rotate_refresh(..., now=now)` share that instant. `rotate_refresh` gained `now: datetime | None = None` (default wall clock — sole production call site is `refresh.py:373`; tests and any future caller omitting it get pre-fix behavior; mypy strict green across 471 files) and threads it into `_locked_row_or_replay` (gate: `_coerce_utc(row.expires_at) <= now`) and `_apply_rotation`. Same locked row (`FOR UPDATE`, same transaction), same deterministic `_coerce_utc` on both sides (`load_session_for_update` coerces at load), same clock value → identical predicates → the self-deadlocking expired-replay branch is provably unreachable from the inline path. `test_rotate_refresh_honours_caller_supplied_clock` pins both directions: a wall-clock-expired row rotates under a caller `now` before `expires_at` (impossible with an internal wall-clock read), and caller `now == expires_at` fires the expired replay branch with audit reason `expired`.

### Delta regression check (c785cd41..bc053075)

- Only 5 files touched; matches the fix-iteration disclosure exactly. No CHANGELOG/scope drift.
- `rotate_refresh` docstring compression: checked clause-by-clause against the pre-fix text — RFC 9700 rotation semantics, replay/dedicated-session audit survival, missing-row shape, monotonic + absolute-cap extension, pre-#1694 default contract all preserved; prose only tightened. New `now` parameter documented with the M1 rationale.
- `_apply_rotation` consuming the caller clock moves `last_seen_at`/extension math microseconds earlier — strictly conservative, monotonic guard unchanged.
- `docs/codebase/ui.md` session_store row updated accurately for the `now=` seam.

### Integration verification (run locally in the review worktree — required Python lanes did not run on `bc053075` because the merge ref is dirty from the known CHANGELOG cascade)

- [x] `ruff check` + `ruff format --check` — clean (src + tests)
- [x] `mypy src/` — strict, 471 files, 0 issues
- [x] `pytest tests/ -n 6 --dist loadscope --ignore=tests/integration` — **7343 passed, 195 skipped** (full backend unit sweep at `bc053075`, 328s, independent reproduction of the fix run)
- [x] Touched modules verbose: `test_ui_token_refresh.py` + `test_ui_session_store.py` + `test_ui_auth_flow.py` — 64 passed; the three decisive fix tests PASSED individually
- [x] B1 red-proof executed (see above)
- [⚠️ not run] `tests/integration` (testcontainers) — no integration test references `rotate_refresh`/`WebSession`/`load_session_for_update`; the lane passed on `c785cd41` and the delta adds no surface it exercises
- CI on `bc053075`: CodeQL, Analyze, DCO all SUCCESS; deploy-smoke skipped (normal for PRs)

### Project-board alignment
|---|---|---|---|---|---|
| #1694 | Open | Closes | ✅ | ✅ full | close on merge |

### Findings
None. Both iteration-1 inline threads are verified fixed and resolved.

### Adjacent improvements (non-blocking, carried from iteration 1)
- The latent `rotate_refresh` replay-branch deadlock for future transactional callers that skip both pre-checks remains a primitive-level hazard — the follow-up ticket the PR proposes still stands (M1 closed this PR's reachability).
- `require_ui_admin`'s session-gone race still renders raw JSON 401 for HTML browsers (disclosed; AC pinned the redirect mapping to `session_expired` only).
- `mergeable=CONFLICTING` is the known CHANGELOG `[Unreleased]` cascade — rebase-at-merge matter for the orchestrator, not a defect of this PR.

### Docs
- [x] `docs/codebase/ui.md` — updated and accurate at the new head

### What worked well
- Both fixes are minimal-diff, land exactly the iteration-1 required changes, and arrive with tests that cannot pass against the unfixed code (the red-proof reproduced cleanly).
- The single-clock contract is now documented at every layer it crosses (call site, primitive docstring, parameter docs, ui.md).

*Note: GitHub rejects own-author review POSTs, so this verdict is posted as a marker-tagged comment; the binding signal is the result file at `.claude/auto/review-results/pr-1711-iter-2.json`.*

Issue #1655, opened 2026-06-11
Task: work_ref I1-T1 — audit_log.work_ref column + work_ref_var ContextVar + stamp from the 3 primary audit writers


damir-topic:
Parent goal: #1651
Parent initiative: #1652

## Problem
No governed MEHO object can be correlated to an external change ticket
(`rg -i work_ref` → 0 hits). The keystone is a `work_ref` column on `audit_log`
threaded by the same ContextVar mechanism that already carries `run_id` /
`agent_session_id` / `parent_audit_id`. There is NO single audit-write
chokepoint — `rg 'AuditLog\('` finds ~11 construction sites — so the value must
ride a ContextVar read by the writers, not be stamped in one helper.

## Current state (honest — with file:line)
- AuditLog model `db/models.py:360-547`: 18 columns; no external-ref column.
- The four threaded-correlation ContextVars live in `operations/_audit.py:71-138`
  (`parent_audit_id_var`, `agent_session_id_var`, `run_id_var`, `step_id_var`);
  `agent_session_id_var` is the cleanest template (fully filterable; the
  `parent_audit_id` filter is deliberately ungated and is NOT the template).
- Three PRIMARY writers each independently read those vars and stamp columns:
  `audit.py:374` (chassis HTTP), `operations/_audit.py:405` (dispatcher DISPATCH),
  `mcp/audit.py:275` (MCP). ~8 system-internal writers (memory/topology/
  approval/ingest/reaper/ui-session) legitimately leave new soft-FK columns NULL.
- Migration head is `0037` (`0037_create_doc_collections.py`); next is `0038`
  (rebase the number to the true head at implementation time).

## Desired state
- A new migration adds nullable `audit_log.work_ref` (`sa.Text()`) + a btree
  index `audit_log_work_ref_idx`, following `0021_add_audit_log_actor_sub.py`.
- A module-level `work_ref_var: ContextVar[str | None]` is defined beside the
  others in `operations/_audit.py`, read (hasattr-guarded, like the existing
  resolvers) by the three primary writers and written into the column.
- Setting where `work_ref_var` is set FROM is I1-T2 (this task leaves it
  unset → NULL except in a unit test that binds it directly).

## Acceptance criteria
- [ ] `grep -n work_ref backend/src/meho_backplane/db/models.py` shows
  `work_ref: Mapped[str | None] = mapped_column(Text, nullable=True, default=None)`
[...]

Issue #1652, opened 2026-06-11
Initiative: work_ref I1 — audit_log.work_ref capture (ContextVar + 3 writers) + query_audit filter & surfacing


damir-topic:
Parent goal: #1651

## Why
The audit log is the universal join every other surface correlates through, so
the `work_ref` column on `audit_log` is the keystone — without it the headline
"filter every governed write by ticket" use case does not exist. There is a
standing trap to avoid: `actor_sub`, `run_id`, and `step_id` are real indexed
columns that were landed WITHOUT being wired as filters or surfaced on
`AuditEntry` (`audit_query/query.py:314-334`). This Initiative wires the FULL
read surface so `work_ref` does not land half-done like those three.

## Outcome
Any audit row can carry a `work_ref`, the value is surfaced on every returned
`AuditEntry`, and `query_audit` filters exact-match by it — end to end from an
MCP/REST call carrying the ref to a `meho audit query --work-ref …` result.

## Tasks
- [ ] #1655 — I1-T1 migration (audit_log.work_ref + btree index) + `work_ref_var` ContextVar + stamp from the 3 primary audit writers
- [ ] #1657 — I1-T2 bind `work_ref_var` from the request boundary (MCP arg / `Meho-Work-Ref` header / dispatch param) *(Depends-on #1655)*
- [ ] #1658 — I1-T3 read surface: `AuditQueryFilters.work_ref` + where-arm + surface on `AuditEntry` + optional `/by-work-ref/{ref}` route + CLI regen *(Depends-on #1655)*

## Definition of done
- [ ] A new migration adds nullable `audit_log.work_ref` (`Text`) + `audit_log_work_ref_idx` btree index; `uv run alembic upgrade head` then `downgrade` round-trips clean.
- [ ] An MCP tool call (or REST request) carrying a `work_ref` produces an `audit_log` row whose `work_ref` equals it; system-internal writers with no ref leave it NULL.
- [ ] `query_audit` with a `work_ref` filter returns only matching rows (exact match), and every returned `AuditEntry` surfaces the value.
- [ ] `uv run ruff check`, `ruff format --check`, `mypy src/`, `pytest -x` green; CLI OpenAPI snapshot regenerated if any REST schema changed.

## Out of scope
- Approval / runbook-run / agent-run / scheduler columns (I2, I3).
- Enforcement, validation, the ~8 system-internal audit writers (memory/topology/reaper/ui-session — they legitimately leave `work_ref` NULL).

## References
[...]

Issue #1651, opened 2026-06-11
Goal: work_ref — bind every governed operation to its authorizing external change ticket


damir-topic:
## Goal

Add an optional `work_ref` to every governed MEHO operation and durable run
object, so an operator (or an autonomous agent) can answer "which change
ticket authorized this?" with one exact-match query instead of prose
archaeology. `work_ref` is a typed-URI join key — `gh:<org>/<repo>#<number>`
today, `jira:PROJ-123` / `snow:CHG0031337` later — carried on audit rows,
approval requests + decisions, runbook runs, agent runs, and scheduled
triggers, and exposed as a binary exact-match filter on the audit / approval
/ runbook-run / scheduler list surfaces.

This Goal ships the FIELD (capture + filter) only. It is an opaque join key:
no validation, no GitHub/Jira API calls, no lifecycle awareness, no
enforcement.

## Why

MEHO is the one dispatch point in front of every governed write (vCenter,
Vault, NSX, …). GitHub branch protection can gate a PR merge; it cannot gate
a vCenter write. The git half of "every change has a paired ticket" has teeth;
the operations half has none — and scheduler-driven approve→execute autonomy
is already live (`scheduler/loop.py`), so honour-system ticket discipline is
about to stop scaling.

The genuine gap is narrow and worth stating precisely: MEHO already has rich,
indexed, queryable INTERNAL causal correlation — `parent_audit_id` (0006),
`agent_session_id` (0014), `actor_sub` (0021), and `run_id`/`step_id` (0034)
all on `audit_log`. What does NOT exist anywhere (`rg -i work_ref` → 0 hits) is
a binding from any governed object to an EXTERNAL work item / change ticket.
The only external-ref-shaped column (`run_id`) points at an internal runbook
run, not a CR. So this Goal adds the missing external dimension; it does not
re-solve correlation.

The headline asymmetry is real: `approvals.approve` takes only the request id
while `approvals.reject` carries a `reason`
(`mcp/tools/approvals.py:438` vs `:513`) — an approval decision has no durable
"why". `work_ref` gives the structured, filterable answer (a free-text reason
is a separate nicety, not the fix).

Industry precedent for "tracker stays external, the backplane owns the binding
+ evidence": ServiceNow DevOps Change Velocity gates a deployment on an
approved change record and attaches pipeline evidence to it; Kosli positions
ServiceNow as "system of record for change" while supplying immutable evidence
with "one permanent link" per change; FireHydrant federates incident follow-up
[...]

PR #1715, opened 2026-06-13
feat(audit): bind work_ref_var from MCP/header/dispatch boundary (#1657)

### LINKED ISSUE #1657: Task: work_ref I1-T2 — bind work_ref_var from the request boundary (MCP arg / Meho-Work-Ref header / dispatch param)
Parent goal: #1651
Parent initiative: #1652
Depends-on: #1655

## Problem
I1-T1 adds `audit_log.work_ref` and the `work_ref_var` ContextVar, but nothing
SETS it — every row is NULL. This task wires the bind sources so a caller can
supply a `work_ref` that flows into the audit rows of that call/dispatch, scoped
per-request (no leakage to the next call). This is the deliberate Stage-1
replacement for the proposal's refuted "bind once per session" — MEHO holds no
session store (`mcp/server.py:646` "assigns + emits + forgets"), so binding is
per-op/per-request, with durable-run inheritance handled in I3.

## Current state (honest — with file:line)
- `work_ref_var` exists (I1-T1) but is never `.set()`.
- MCP per-call context precedent: `mcp/audit.py` resolves `agent_session_id`
  from the `mcp_session_id` contextvar bound per request from the
  `Mcp-Session-Id` transport header — the template for binding a per-call value.
- HTTP chassis: `AuditMiddleware` (`audit.py:715-1031`) builds the request row;
  it already reads transport headers (request_id) and is the place a
  `Meho-Work-Ref` request header would be read and bound.
- Dispatcher: `run_id_var`/`step_id_var` are bound AROUND `call_operation`
  (`operations/_audit.py`, runbook step bind `run_service.py:1001-1007`) — the
  template for binding `work_ref_var` around a dispatch from an explicit param.

## Desired state
A `work_ref` supplied at any governed entry point binds `work_ref_var` for the
duration of that request/dispatch and is then reset:
- **MCP**: an optional `work_ref` argument accepted on the governed dispatch /
  meta-tool entry, bound around the handler (mirroring the `mcp_session_id`
  resolve).
- **HTTP**: a `Meho-Work-Ref` request header read in `AuditMiddleware` and bound
  for the request scope.
- **Dispatch**: an explicit `work_ref` param on the operations dispatch entry
  that overrides the ambient bind (per-op override the proposal calls for).

## Acceptance criteria
- [ ] New test: an MCP governed dispatch carrying `work_ref="gh:evoila/meho#7"`
  produces an `audit_log` row with `work_ref == "gh:evoila/meho#7"`; a second
  dispatch without it produces a NULL `work_ref` (per-call scoping, no leak).
- [ ] New test: a REST request with header `Meho-Work-Ref: gh:evoila/meho#7`
  produces a chassis `audit_log` row carrying it; absent header → NULL.
- [ ] An explicit per-op `work_ref` overrides the ambient header/arg value
  (asserted in a test where both are present and differ).
- [ ] From `backend/`: `uv run ruff check`, `ruff format --check`, `mypy src/`,
  `pytest -x` green; `cd cli && make snapshot-openapi && make generate` run if
  any REST/MCP schema or documented header changed (commit the regenerated client).

## Out of scope
- The query filter + AuditEntry surfacing (I1-T3).
- Approval / run-object binding (I2, I3) — those inherit from the durable row,
  not from this per-request bind.
- Any session-scoped "bind once" store (explicitly dropped for Stage 1).

## References
- `work_ref_var` (added by I1-T1) in `operations/_audit.py:71-138`.
- `mcp/audit.py:157-299` (mcp_session_id per-call resolve template); `mcp/server.py:646` (no session store).
- `audit.py:715-1031` (`AuditMiddleware`, header-read site); `operations/_audit.py:357-425` (dispatcher writer).
- Runbook step contextvar bind `runbooks/run_service.py:1001-1007` (bind-around-dispatch template).
- Python best practices (this repo): `ContextVar` reset via token, no cross-request leakage.

damir-topic:
## What

Wires the **three governed request-boundary bind sources** for `work_ref_var` (work_ref **I1-T2**, Initiative #1652, Goal #1651) so a caller-supplied external change-ticket reference flows onto the `audit_log.work_ref` column of that call — **scoped per-request, no leak to the next**.

Builds directly on **#1655 (I1-T1)**, which landed the `audit_log.work_ref` column + the `work_ref_var` `ContextVar` + the three primary audit writers that *read* it. Until now nothing **set** the var, so every row was NULL. This PR supplies the binds.

## The three entry points

| # | Source | Where bound |
|---|--------|-------------|
| 1 | **MCP tool arg** (`call_operation` `work_ref`) | `mcp/tools/operations.py` inputSchema → `_call_operation_handler` → `_call_operation_impl` |
| 2 | **`Meho-Work-Ref` HTTP header** | `audit.py` `AuditMiddleware` (`_extract_work_ref` off the raw ASGI scope) |
| 3 | **Dispatch param** (`CallOperationBody.work_ref`) | `meta_tools.py` `_call_operation_impl`, bound around `dispatch(...)` |

- The MCP arg and the REST `CallOperationBody.work_ref` field share the same `arguments` dict (`POST /api/v1/operations/call` calls `call_operation(operator, body.model_dump())`), so both reach the **same** per-op bind in `_call_operation_impl` — the explicit **per-op override** the Goal #1651 design calls for, which takes precedence over an ambient header bind on the inner DISPATCH row.
- The `Meho-Work-Ref` header is read straight off the ASGI scope because `AuditMiddleware` is pure-ASGI and runs **before** route resolution — there is no `Request` to hang a `Header(...)` param on (same posture as `X-Request-Id`). It is therefore **not** in the OpenAPI schema; only the real `CallOperationBody.work_ref` body field moves the documented surface (and thus the regenerated CLI client).

## Set/reset hygiene (the correctness point)

Every bind uses **`token = var.set(...)` + `try/finally` + `var.reset(token)`**, mirroring the runbook engine's `run_id_var` / `step_id_var` bind-around-dispatch in `runbooks/run_service.py`. The `finally` is load-bearing: `AuditMiddleware` is a single instance serving concurrent requests on one event loop, so a leaked binding would mis-stamp the **next** request's audit row. Binds are **gated on a non-empty value** so a bare call never clobbers an ambient header bind, and `reset` restores exactly the prior value (never a hardcoded default).

## Tests (`backend/tests/test_audit_work_ref_bind.py`, new)

Each entry point proves **both halves** of the contract — stamped when supplied AND reset afterward (a subsequent unbound call lands NULL, the per-call scoping the design requires):

- MCP tool arg binds, second bare call → NULL (no leak across calls)
- dispatch param lands on the DISPATCH row
- explicit arg **overrides** a differing ambient bind; ambient restored on return
- bare call does **not** clobber the ambient bind (header flows to the inner row)
- reset fires on the **dispatch-error path** (`try/finally`)
- header stamped on the chassis row (end-to-end via `TestClient`); absent → NULL; case-insensitive match
- in-task `AuditMiddleware` drive: row carries ref, var resets, second header-less call → NULL
- pure-unit `_extract_work_ref` parsing (strip, case-insensitivity, empty → None)

## Verification (all green)

- `uv run ruff check src/ tests/` — clean
- `uv run ruff format --check src/ tests/` — clean
- `uv run mypy src/` — clean (469 files)
- `uv run pytest -x` — **7363 passed, 416 skipped, 0 failed**
- `cd cli && make snapshot-openapi && make generate` — **idempotent** (no further diff); regen committed
- `go build ./...` — clean

## Out of scope (per the issue)

Query filter + `AuditEntry` surfacing (I1-T3); approval / run-object binding (I2, I3, which inherit from the durable row); any session-scoped "bind once" store (explicitly dropped for Stage 1 — MEHO holds no session store).

Closes #1657

## Release Notes

* **New Features**
  * Added support for tracking external work references on operations. The new `work_ref` parameter can be passed directly in operation calls or via the request header to correlate operations with external change tickets.
  * Work references are now recorded in audit logs for improved change tracking and correlation.


damir-topic:
@coderabbitai review

Issue #1643, opened 2026-06-11
Task: enforce application-layer tenant scoping on Vault mount/path


damir-topic:
Parent goal: #221

## Problem

The Vault connector forwards agent-supplied `mount` / `path` to hvac verbatim with no
application-layer tenant binding. If the shared Vault role's policy is over-broad, a
tenant-A caller can read `tenant-b/...` secrets. This task adds a defense-in-depth tenant
scope check on the mount/path before the hvac call.

(Implementation task for the privately-tracked finding `evoila-bosnia/meho-internal#99`.)

## Current state (verified in code, v0.13.0)

`mount`/`path` taken verbatim, no tenant binding:
- [`connectors/vault/ops.py:291-303`](backend/src/meho_backplane/connectors/vault/ops.py#L291-L303) (read)
- [`:388-396`](backend/src/meho_backplane/connectors/vault/ops.py#L388-L396) (list)
- [`:726-734`](backend/src/meho_backplane/connectors/vault/ops.py#L726-L734) (versions)

## Desired state

Before the hvac call, enforce that the requested `mount`/`path` falls within the operator's
tenant namespace (e.g. a `tenant-<id>/` prefix convention or per-tenant allowlist) — a
mismatch returns a structured error. Defense-in-depth behind the Vault policy. The exact
namespace convention is part of this task's design (document it in `docs/codebase/`).

## Acceptance criteria

- [ ] A caller requesting a path outside their tenant namespace is denied (test, all three
      handlers); in-namespace requests unchanged.
- [ ] The tenant-namespace convention is documented.
- [ ] CHANGELOG `[Unreleased]` bullet.

## References

- Private tracker: `meho-internal#99`. Part of the cross-tenant isolation hardening.

PR #1731, opened 2026-06-13
feat(connectors): per-tenant Vault KV path convention + secret_ref migration (#1723)

### LINKED ISSUE #1723: Task: per-tenant Vault KV path convention + migrate target secret_ref off per-sub layout
Parent goal: #221
Parent initiative: #1685

## Problem
Secrets are stored per operator `sub` at `secret/data/targets/<sub>/*`. That
duplicates a target's credential per operator and blocks two operators in one
tenant from sharing it. The #1643 guard already renders a `{tenant_id}` prefix
template but is default-off because flipping it on would break reads against the
per-`sub` layout. This task moves the layout to a **per-tenant shared** path so
the guard *can* go default-on (T3 / the depending task).

## Current state (honest — with file:line)
- Path convention documented per-`sub`: `backend/src/meho_backplane/settings.py:1097`,
  `backend/src/meho_backplane/connectors/vault/tenant_scope.py:25`,
  `docs/cross-repo/connector-vault-policy.md:77-83`.
- Targets store the Vault path in `secret_ref: str | None` —
  `backend/src/meho_backplane/targets/schemas.py:172` (+ `TargetSummary`:112,
  `TargetCreate`:225, `TargetUpdate`:278). Consumed via
  `load_secret_ref(target, operator)` in connector auth loaders
  (e.g. `backend/src/meho_backplane/connectors/vcf_automation/_auth.py:48`).
- KV ops that read/write the path: `backend/src/meho_backplane/connectors/vault/ops.py:232`
  (read), :493 (put).
- `operator.tenant_id: UUID` is always present on authenticated requests —
  `backend/src/meho_backplane/auth/operator.py:145`.

## Desired state
- New canonical convention: `secret/data/tenants/<tenant_id>/<target>`
  (tenant_id = canonical dashed UUID, matching `rendered_tenant_prefix` in
  `backend/src/meho_backplane/connectors/vault/tenant_scope.py:80`).
- A helper that derives the per-tenant `secret_ref` for a target from
  `operator.tenant_id` + target identity, used at target create/update so new
  targets land on the per-tenant path.
- A **migration runbook** (`docs/cross-repo/`) + a relocation helper documenting
  how to move existing `targets/<sub>/*` secrets to `tenants/<id>/<target>`
  (KV-v2 read→write→delete via existing `vault_kv_*` ops), and how to rewrite
  each target's `secret_ref`.
- The system-tenant nil-UUID exemption
  (`backend/src/meho_backplane/connectors/vault/tenant_scope.py:58`) is preserved.

## Acceptance criteria
- [ ] New test: target create/update with `operator.tenant_id=T` produces
  `secret_ref` under `tenants/<T>/…` (not `targets/<sub>/…`).
- [ ] New test: the relocation helper, given an old `targets/<sub>/x` secret,
  writes it to `tenants/<id>/x` and the rewritten `secret_ref` resolves through
  `vault_kv_read`.
- [ ] Migration runbook committed under `docs/cross-repo/`;
  `docs/codebase/connectors-vault-tenant-scope.md` updated to state the
  per-tenant layout is canonical.
- [ ] From `backend/`: `uv run ruff check`, `ruff format --check`, `mypy src/`,
  `pytest -x` green; `cd cli && make snapshot-openapi && make generate` run if
  any REST/MCP schema changed (commit the regenerated client).

## Out of scope
- Templated Vault ACL policies + identity-group mapping (the sibling policy task).
- Flipping the guard default-on (the depending task).
- Auto-running the migration in app code — the runbook is operator-driven (RDC
  owns the Vault deployment).

## References
- #1643 (guard), #1673 (advisory); consumer tracker `evoila-bosnia/meho-internal#99`.
- Vault KV-v2 paths via hvac:
  `backend/src/meho_backplane/connectors/vault/tenant_scope.py`,
  `backend/src/meho_backplane/connectors/vault/ops.py`.
- Python best practices (this repo): UUID canonicalization, pure helper + tests,
  no cross-request leakage.

damir-topic:
## Summary
- Adds the canonical **per-tenant** Vault KV path convention `secret/data/tenants/<tenant_id>/<target>`, replacing the retired per-operator-`sub` layout so a target's credential lives once per tenant and the #1643 tenant-scope guard has a `{tenant_id}` partition it can enforce against.
- New pure helper `tenant_secret_ref(tenant_id, target)` derives the logical `secret_ref`; `POST /api/v1/targets` (no explicit `secret_ref`) and `PATCH /api/v1/targets/{name}` (untouched `secret_ref` on an unset row) now default new targets onto that path. An explicit `secret_ref` (or `{"secret_ref": null}`) is always honoured verbatim.
- New operator-driven `relocate_target_secret(...)` coroutine moves an existing `targets/<sub>/x` secret to `tenants/<id>/x` via the existing `vault_kv_*` ops (read → write → optional reversible soft-delete) and returns the rewritten ref. Not wired into any request path — RDC owns the Vault deployment.
- Migration runbook `docs/cross-repo/vault-per-tenant-migration.md` + updated `docs/codebase/connectors-vault-tenant-scope.md` to state per-tenant is canonical.

## Test plan
- [x] `uv run ruff check` — clean
- [x] `mypy src/` — clean (472 files)
- [x] New unit tests `tests/test_connectors_vault_tenant_paths.py` — 7 passed (pure derivation: canonical dashed-UUID shape, matches guard prefix, strips separators, rejects empty; relocation: read-old/write-per-tenant, new ref resolves through `vault_kv_read`, soft-deletes source on request)
- [x] New route tests in `tests/test_api_v1_targets.py` — 6 passed (create derives `tenants/<T>/<name>`; explicit ref honoured; PATCH homes unset ref; PATCH preserves set ref; explicit null clears)
- [x] No REST/MCP schema change (`secret_ref` field + route signatures unchanged) → no OpenAPI snapshot regen needed.

## Out of scope
- Templated Vault ACL policies + identity-group mapping (sibling policy task).
- Flipping the #1643 guard default-on (depending task).
- Auto-running the migration in app code — operator-driven per the runbook.

## Adjacent findings
- `tests/test_connectors_vault.py::test_preflight_fail_soft_when_probe_raises` errors in teardown on clean `main` (pre-existing, unrelated; test body passes) — deselected in the local full run.
- `code-quality.py --diff` warns `relocate_target_secret` is 76 lines (mostly docstring; body ~12 lines) — warning, not a blocker.

---

## Follow-up (auto-implement-issue, iteration 2, 2026-06-13)

Addressed review finding **B1** from the iteration-1 `REQUEST_CHANGES`
(OpenAPI product-enum truncation; `CLI API snapshot freshness` RED):

- **B1** `73d7a355` — `build_openapi_schema()` now calls
  `_eager_import_connectors()` **unconditionally** instead of behind the
  `if not registered_product_tokens()` guard. The new import chain
  (`targets.py` → `tenant_paths` → `connectors.vault.ops` →
  `connectors.vault.__init__`) self-registers `VaultConnector` at import
  time, so the registry was already `{"vault"}` when the override ran;
  the guard short-circuited the eager import, the other 17 connectors
  never loaded, and `TargetCreate.product` collapsed 18 → `["vault"]`.
  The eager import is idempotent (`sys.modules`-cached), so dropping the
  guard makes the schema build robust to any partially-populated
  registry. Reviewer's preferred option (1).
- **B1 (regression test)** `73d7a355` — added
  `test_openapi_product_enum_full_when_registry_partially_populated` to
  `tests/test_openapi_target_product_enum.py`. It reproduces the bug in
  a fresh subprocess (the only faithful repro — the test process is
  already warm with every connector in `sys.modules`): imports
  `tenant_paths` first, asserts the registry holds exactly `{"vault"}`,
  then asserts the generated enum is the full registered set. Verified
  red against the reintroduced guard, green with the fix.

The committed CLI snapshot already carried the correct full 18-connector
enum (the PR commit never touched `cli/`); the CI freshness job was
failing because it *regenerated* the snapshot in a fresh interpreter and
the regen truncated. A fresh `cd cli && make snapshot-openapi &&
make generate` now leaves the `cli/` tree clean, confirming freshness is
restored. The iteration-1 "no OpenAPI snapshot regen needed" claim in
the test plan above was the original miss.

Closes #1723

## Summary by CodeRabbit

* **New Features**
  * Targets now automatically derive per-tenant secret paths when not explicitly specified during creation and updates
  * Added helper utilities for migrating Vault secrets from legacy per-operator layout to per-tenant structure

* **Documentation**
  * Added comprehensive Vault tenant-scope migration guide
  * Updated Vault KV tenant-scope documentation with canonical per-tenant path conventions

* **Tests**
  * Enhanced test coverage for target secret reference handling and path derivation
  * Added validation tests for secret relocation functionality

### DIFF
--- a/backend/src/meho_backplane/api/v1/targets.py
+++ b/backend/src/meho_backplane/api/v1/targets.py
@@ -153,6 +153,7 @@ from meho_backplane.connectors.registry import (
 )
 from meho_backplane.connectors.resolver import resolve_connector_or_label
 from meho_backplane.connectors.schemas import AuthModel, CandidateHint, FingerprintResult
+from meho_backplane.connectors.vault.tenant_paths import tenant_secret_ref
 from meho_backplane.db.engine import get_session
 from meho_backplane.db.models import GraphNode
 from meho_backplane.db.models import Target as TargetORM
@@ -876,6 +877,15 @@ async def create_target(
     # may have supplied, so the stored row matches the registry /
     # resolver spelling (G0.18-T2 #1355).
     create_fields["product"] = product
+    # #1723: default a fresh target's secret onto the per-tenant shared
+    # path ``tenants/<tenant_id>/<name>`` so new targets land on the
+    # canonical layout the #1643 guard can enforce, instead of the
+    # retired per-``sub`` layout an operator would otherwise type by
+    # hand. An explicitly-supplied ``secret_ref`` is honoured verbatim
+    # (operator override / a non-default mount layout); only the absent
+    # case is derived.
+    if create_fields.get("secret_ref") is None:
+        create_fields["secret_ref"] = tenant_secret_ref(operator.tenant_id, body.name)
     t = TargetORM(
         id=uuid.uuid4(),
         tenant_id=operator.tenant_id,
@@ -989,6 +999,18 @@ async def update_target(
             )
     for k, v in updates.items():
         setattr(t, k, v)
+    # #1723: home an as-yet-unconfigured target onto the per-tenant shared
+    # path. A PATCH that does not touch ``secret_ref`` (the field is absent
+    # from the request body, so it is not in ``updates``) on a row whose
+    # ``secret_ref`` is still unset derives ``tenants/<tenant_id>/<name>``
+    # so the canonical layout is filled in without the operator typing it.
+    # A PATCH that *sends* ``secret_ref`` — including an explicit
+    # ``{"secret_ref": null}`` to clear it — is honoured verbatim and is
+    # never overwritten here (it is in ``updates`` and the branch is
+    # skipped); silent re-homing of an already-configured ref is the
+    # operator-driven migration runbook's job, not the route's.
+    if "secret_ref" not in updates and t.secret_ref is None:
+        t.secret_ref = tenant_secret_ref(operator.tenant_id, t.name)
     t.updated_at = datetime.now(UTC)
     _log.info(
         "target_updated",
--- a/backend/src/meho_backplane/connectors/vault/tenant_paths.py
+++ b/backend/src/meho_backplane/connectors/vault/tenant_paths.py
@@ -1,0 +1,202 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""Per-tenant Vault KV path convention + a runbook-driven relocation helper (#1723).
+
+Goal #221 / Initiative #1685. The shipped deploy convention scoped each
+target's secret per operator ``sub`` (``secret/data/targets/<sub>/*``,
+``docs/cross-repo/connector-vault-policy.md`` §2). That duplicated a
+target's credential per operator and blocked two operators in one tenant
+from sharing it; it also meant the #1643 tenant-scope guard
+(:mod:`meho_backplane.connectors.vault.tenant_scope`) had no universal
+``{tenant_id}`` partition to enforce against, so it had to ship default-off.
+
+This module moves the canonical layout to a **per-tenant shared** path:
+
+.. code-block:: text
+
+    secret/data/tenants/<tenant_id>/<target>
+
+so a target's credential lives once per tenant and every operator in that
+tenant reads the same path. ``<tenant_id>`` is the canonical dashed
+lowercase UUID (``str(UUID)``) — the exact rendering
+:func:`~meho_backplane.connectors.vault.tenant_scope.rendered_tenant_prefix`
+produces, so a deploy can enable the guard with the matching
+``tenants/{tenant_id}/`` prefix once its secrets are relocated.
+
+Two public helpers:
+
+* :func:`tenant_secret_ref` — a pure function deriving the logical
+  ``secret_ref`` (the path relative to the KV-v2 mount, the shape
+  ``targets.secret_ref`` carries) for a target from its tenant + name.
+  Called at target create/update so freshly-created targets default onto
+  the per-tenant path instead of a per-``sub`` path the operator would
+  otherwise have to type by hand.
+* :func:`relocate_target_secret` — an operator-driven coroutine that moves
+  one existing ``targets/<sub>/<x>`` secret to ``tenants/<id>/<x>`` via the
+  existing ``vault_kv_*`` handlers (read → write → optional delete) and
+  returns the rewritten ``secret_ref``. It is **not** wired into any app
+  request path: the migration is operator-run from the runbook
+  (``docs/cross-repo/vault-per-tenant-migration.md``); RDC owns the Vault
+  deployment, so the backplane never auto-relocates secrets.
+
+Both stay clear of the system/shim operator: the Nil-UUID tenant exemption
+that :mod:`meho_backplane.connectors.vault.tenant_scope` preserves applies
+to the *guard*; these helpers only ever run for a real authenticated tenant
+admin (target CRUD) or under an operator the runbook supplies explicitly.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+from uuid import UUID
+
+from meho_backplane.connectors.vault.ops import (
+    vault_kv_delete,
+    vault_kv_put,
+    vault_kv_read,
+)
+
+if TYPE_CHECKING:
+    from meho_backplane.auth.operator import Operator
+
+__all__ = ["TENANT_SECRET_PREFIX", "relocate_target_secret", "tenant_secret_ref"]
+
+#: The logical-path prefix (relative to the KV-v2 mount) under which a
+#: tenant's target secrets live. Combined with the dashed-UUID tenant id
+#: and the target identity this yields ``tenants/<tenant_id>/<target>``;
+#: hvac inserts the mount and the ``/data/`` API segment itself, so the
+#: full KV-v2 wire path becomes ``secret/data/tenants/<tenant_id>/<target>``.
+#: A deploy enables the #1643 guard against this layout by setting
+#: ``VAULT_KV_TENANT_SCOPE_PREFIX="tenants/{tenant_id}/"``.
+TENANT_SECRET_PREFIX = "tenants"
+
+
+def tenant_secret_ref(tenant_id: UUID, target: str) -> str:
+    """Derive the per-tenant logical ``secret_ref`` for *target*.
+
+    Returns ``"tenants/<tenant_id>/<target>"`` — the logical KV-v2 path
+    relative to the mount, which is the shape ``targets.secret_ref``
+    stores (no mount segment, no ``/data/`` API segment; hvac inserts
+    both at read time, see
+    :func:`meho_backplane.connectors._shared.vault_creds._is_api_path_shaped`).
+
+    ``tenant_id`` is rendered in its canonical dashed lowercase form
+    (``str(UUID)``), matching how tenant UUIDs appear everywhere else —
+    audit rows, JWT claims, and
+    :func:`~meho_backplane.connectors.vault.tenant_scope.rendered_tenant_prefix`'s
+    output — so a deploy that enables the guard with
+    ``tenants/{tenant_id}/`` finds the derived ref already inside the
+    rendered prefix.
+
+    Parameters
+    ----------
+    tenant_id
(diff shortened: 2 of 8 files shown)


damir-topic:
@coderabbitai review

backend/src/meho_backplane/api/v1/targets.py
```diff
@@ -153,6 +153,7 @@
 )
 from meho_backplane.connectors.resolver import resolve_connector_or_label
 from meho_backplane.connectors.schemas import AuthModel, CandidateHint, FingerprintResult
+from meho_backplane.connectors.vault.tenant_paths import tenant_secret_ref
```
damir-topic, line 156: **[🛑 B1]** — New import chain truncates the OpenAPI `TargetCreate.product` enum from all 18 connectors to just `vault`

**What:** Importing `tenant_paths` here pulls in `meho_backplane.connectors.vault.ops`, which forces Python to run the `connectors.vault` package `__init__` — and that self-registers `VaultConnector` at import time. By the time `build_openapi_schema()` runs, `registered_product_tokens()` is already non-empty (`{"vault"}`), so its guard `if not registered_product_tokens(): _eager_import_connectors()` (`main.py` ~985) is skipped and the other 17 connectors never load. Result: the generated `TargetCreate.product` enum collapses to `["vault"]`.

**Why it matters:** The `CLI API snapshot freshness` CI check is RED on this head. I reproduced it locally: `cd cli && make snapshot-openapi` on this branch regenerates the product enum as `["vault"]` (18 → 1); the same regen on clean `main` registers all 18 and stays clean — so this PR introduces the regression. A runtime probe of `build_openapi_schema()` on this head confirms `TargetCreate.product.enum == ['vault']`. The PR description's claim "No REST/MCP schema change → no OpenAPI snapshot regen needed" is therefore incorrect — committing the regenerated snapshot would *bake in* the truncated enum, so regenerating is **not** the fix. The `build_openapi_schema` docstring documents the exact invariant this breaks: "Calls `_eager_import_connectors` when the registry is empty so the schema is correct even when the override fires before the FastAPI lifespan." (Runtime serving is unaffected — the lifespan calls `_eager_import_connectors()` at `main.py:333` — but OpenAPI generation and any pre-lifespan enum reader are broken.)

**Required change:** Make the schema build robust to a partially-populated registry. Simplest: in `build_openapi_schema()` drop the emptiness guard and call `_eager_import_connectors()` unconditionally (it is idempotent — modules already in `sys.modules` are a no-op). Alternative: avoid the import-time connector registration triggered from `tenant_paths.py` (e.g. import the `vault_kv_*` ops lazily inside `relocate_target_secret`, and avoid importing the `connectors.vault` package at `targets.py` module load). Either way, add a regression test asserting `build_openapi_schema()['components']['schemas']['TargetCreate']['properties']['product']['enum']` contains the full `registered_product_tokens()` set, then regenerate and commit `cli/api/openapi.json` + `cli/internal/api/client.gen.go`.

**Source:** `backend/src/meho_backplane/main.py` `build_openapi_schema` (~L985) + `connectors/registry.py` `_eager_import_connectors` (~L324); `connectors/vault/__init__.py` (self-registers `VaultConnector` at import).

damir-topic:
@coderabbitai review

damir-topic:




**Anchor task:** #1723 — per-tenant Vault KV path convention + secret_ref migration
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `73d7a355e9bcd959801572bad2513a0dc789d761`
**AI reviewer (CodeRabbit):** received (recovered from iter-1 rate-limit) — 1 Major + 1 nitpick, both outside the fix diff; cross-checked below

### Re-review focus — B1 (iter-1 blocker) resolution

B1 was: the #1723 import chain (`targets.py` → `tenant_paths` → `connectors.vault.ops` → `connectors.vault.__init__`) self-registers `VaultConnector` at import time, leaving the registry partially populated (`{"vault"}`) before `build_openapi_schema()`'s `if not registered_product_tokens()` guard ran. The guard short-circuited, the other 17 connectors never eager-loaded, and `TargetCreate.product` collapsed 18 → `["vault"]` — turning `CLI API snapshot freshness` RED.

The fix (commit `73d7a355`) drops the guard and calls `_eager_import_connectors()` **unconditionally** in `build_openapi_schema()`, plus a subprocess-based regression test.

Verified empirically:
- [x] Eager-import is now unconditional (`main.py` ~L999) and idempotent — `_eager_import_connectors()` uses `importlib.import_module`, `sys.modules`-cached, so re-imports re-run nothing.
- [x] Regression test **fails** when the old guard is reintroduced, **passes** with the fix (proven by temporarily restoring the guard in the worktree).
- [x] `make snapshot-openapi` regenerates the full **18-connector** `TargetCreate.product` enum; `make generate` leaves the `cli/` tree clean — freshness restored, no snapshot commit needed.
- [x] Live CI **`CLI API snapshot freshness` = SUCCESS** on head `73d7a355`.

No regression introduced by the fix: `build_openapi_schema` is the lazy `app.openapi` override, result-cached on `app.openapi_schema`, so the body runs at most once per process. The extra unconditional call is a no-op when modules are warm (post-lifespan, the common case) and does the required import work on the cold snapshot path. No per-request cost, no import-ordering hazard.

### Acceptance criteria
- [x] Path helper / honored-verbatim secret_ref / migration helper / guard-not-flipped — all met in iter-1 (unchanged by fix commit).
- [x] OpenAPI enum no longer truncated — the iter-1 blocker — now resolved and CI-green.

### Project-board alignment
|---|---|---|---|---|---|
| #1723 | In Progress | Closes | ✅ | ✅ full | close on merge |

### Predicted vs. actual diff (fix commit d2f6b9d5..73d7a355)
- ✅ `backend/src/meho_backplane/main.py` — guard removed, eager-import made unconditional; docstring updated to explain why
- ✅ `backend/tests/test_openapi_target_product_enum.py` — subprocess-based regression test pinning the full-enum contract

### Integration verification (in worktree at head SHA)
- [x] `ruff check` — clean
- [x] `mypy` — clean
- [x] `pytest tests/test_openapi_target_product_enum.py` — 5 passed
- [x] affected-area: `test_api_v1_targets.py` + `test_connectors_vault_tenant_paths.py` + `test_connectors_registry_v2.py` + `test_targets_schemas.py` — 132 passed
- [x] `make snapshot-openapi` + `make generate` — full 18-enum, clean tree
- [x] regression-test efficacy — fails with old guard, passes with fix
- ⚠️ CI `Python (ruff + mypy + pytest)` FAILURE = unrelated flake: `tests/test_corpus_client.py::test_forwarded_jwt_never_logged` (JWT-redaction log assertion). It passes deterministically at this head locally (3/3 isolated + full file 18/18) and has no nexus to the fix (corpus-client path, not connector registration/OpenAPI). Not a blocker.

### Findings
None. The iter-1 blocker B1 is resolved; no new blockers or majors from the fix.

CodeRabbit recovered from its iter-1 rate-limit and reviewed all 8 PR files (base→`73d7a355`).
- **Disputed: 1** — 🟠 Major `targets.py:875-888` "explicit `{\"secret_ref\": null}` not honored on create". Real behavior (omitted vs explicit-null are conflated via `.get(...) is None`) but **misattributed as a defect**. #1723's intent is precisely to auto-home unconfigured secrets onto `tenants/<tenant_id>/<name>`; preserving an explicit-null would let a client recreate the stranded null-ref state the task exists to eliminate, and the `update_target` path actively heals exactly that state. No acceptance criterion asks for an explicit-null opt-out on create. Not adopted.
- **Out of scope: 1** — 🧹 nitpick "add explicit-null create test". Would encode the disputed contract above; not adopted.
- This finding is also outside the fix-commit diff (original create-path, untouched by `73d7a355`).

### Docs
- [x] `docs/codebase/connectors-vault-tenant-scope.md` + `docs/cross-repo/vault-per-tenant-migration.md` — verified accurate (updated in original PR commit; unchanged by fix).

### What worked well
- The fix is minimal and root-cause-correct: it makes schema-build robust to *any* partially-populated registry rather than special-casing the vault chain, and the docstring explains "registry non-empty ≠ all connectors loaded" so the guard isn't re-added later.
- The regression test correctly uses a fresh subprocess — the only faithful repro, since the test process is already warm with every connector in `sys.modules` — and pins both the precondition (`partial == ["vault"]`) and the post-fix full-enum count.

Issue #1725, opened 2026-06-13
Task: default VAULT_KV_TENANT_SCOPE_PREFIX to tenants/{tenant_id}/ + run #1643 guard default-on


damir-topic:
Parent goal: #221
Parent initiative: #1685
Depends-on: #1723, #1724

## Problem
The #1643 guard ships default-off (`vault_kv_tenant_scope_prefix = ""`) and #1673
logs an advisory at startup while it's unenforced. Once the layout is per-tenant
(the layout task) and templated policies exist (the policy task), the guard should
default **on** so tenant isolation is enforced without per-deploy opt-in.

## Current state (honest — with file:line)
- Setting default empty (guard off): `backend/src/meho_backplane/settings.py:1102`
  `vault_kv_tenant_scope_prefix: str = ""`; env parse ~:1455.
- Guard enforcement: `backend/src/meho_backplane/connectors/vault/tenant_scope.py:107`
  `enforce_tenant_scope`, raising `VaultTenantScopeError`
  (`backend/src/meho_backplane/connectors/vault/tenant_scope.py:61`); prefix render
  `backend/src/meho_backplane/connectors/vault/tenant_scope.py:80`.
- Startup advisory (fires when empty): `backend/src/meho_backplane/main.py:230`
  `_advise_vault_tenant_scope_unenforced`, event `vault_tenant_scope_unenforced`.
- Tests: `backend/tests/test_connectors_vault_tenant_scope.py`,
  `backend/tests/test_vault_tenant_scope_advisory.py`.

## Desired state
- Default `vault_kv_tenant_scope_prefix = "tenants/{tenant_id}/"` (guard enforced
  out of the box); system nil-UUID tenant stays exempt.
- Advisory inverts: silent when enforced (default), and (optionally) advises only
  if an operator explicitly sets the prefix back to empty.
- Docs state default-on; the v0.14.0 opt-in framing is updated to "default-on as
  of <this release>".

## Acceptance criteria
- [ ] `get_settings()` with no env override yields
  `vault_kv_tenant_scope_prefix == "tenants/{tenant_id}/"`; a KV op on a
  non-matching path raises `VaultTenantScopeError` by default (new/updated test).
- [ ] Advisory test updated: no `vault_tenant_scope_unenforced` warning on default
  config; system-tenant op still exempt.
- [ ] `docs/codebase/connectors-vault-tenant-scope.md` updated to "default-on";
  note the migration prerequisite (the layout task) in the doc.
[...]

PR #1763, opened 2026-06-14
fix(connectors): reject a mount-included vault.kv path with a shape hint

### LINKED ISSUE #1755: Task: detect a mount-included vault.kv path and return a path-shape hint instead of an opaque Forbidden
Parent goal: #221
Parent initiative: #1752

## Summary

`vault.kv.*` with a **mount-included `path`** (e.g. `path="secret/meho/test/federation"`) returns an opaque Forbidden, because the connector prepends the KV-v2 mount again, yielding a doubled `…/data/secret/meho/…`. The mount-relative `path="meho/test/federation"` works. The error reads as an auth/permission failure when it is a path-shape mistake.

## Root cause (grounded on v0.15.0)

- All five KV ops forward the caller's `path` verbatim to hvac with `mount_point=mount` (`connectors/vault/ops.py:295-296, 310-314` for read; same shape for list/put/patch/delete). hvac builds `v1/<mount>/data/<path>`, so a caller `path="secret/meho/…"` becomes `v1/secret/data/secret/meho/…`.
- The doubled path fails the Vault ACL policy → a read-phase 403 that propagates as `connector_error` (`exception_class=Forbidden`) — indistinguishable from a real permission denial.
- The tenant-scope guard also rejects it (candidate `secret/secret/meho/…` fails the prefix), but with a generic tenant-scope error, not a path-shape hint (`connectors/vault/tenant_scope.py:128-140`).
- The `path` schema documents "no mount prefix" (`connectors/vault/ops.py:125-133`) but enforces nothing.

## Proposed fix (lowest-friction)

In the shared mount/path extraction for all five KV ops, detect `path.startswith(mount + "/")` and raise a clear pre-flight error (a new `VaultPathShapeError`) **before** the request: `vault.kv.<op> path includes the mount segment '<mount>/'; pass the mount-relative path instead (e.g. 'meho/test/federation', not 'secret/meho/test/federation')`. (Reject rather than silently strip, so an operator who genuinely has a `secret/`-prefixed logical path isn't surprised.)

## Acceptance criteria

- [ ] `vault.kv.read` (and list/put/patch/delete) with `path="secret/…"` and `mount="secret"` returns a structured error naming the path-shape mistake (`exception_class=VaultPathShapeError`) and **never calls hvac**. Verifiable: new test in `backend/tests/test_connectors_vault.py` asserts the error + that the hvac client was not invoked.
- [ ] The mount-relative path still works unchanged (existing vault read/list/put/patch/delete tests stay green).
- [ ] The error message names the mount-relative form to use (`grep -n "mount-relative\|VaultPathShapeError" backend/src/meho_backplane/connectors/vault/ops.py`).

## Out of scope

- The tenant-scope guard's own messaging (#1725); this task is specifically the mount-double-prefix path-shape hint.

## References

- Signal: #1751 (smaller findings).
- Grounding: `connectors/vault/ops.py:125-133, 295-314`, `connectors/vault/tenant_scope.py:128-140`, `auth/vault.py:127-147`.
- Convention: `docs/codebase/error-message-shape.md` (actionable error over opaque 403).

damir-topic:
## Summary

- `vault.kv.*` returned an opaque `Forbidden` 403 when a caller passed a `path` that re-included the mount segment (`path="secret/meho/…"` with `mount="secret"`). hvac addresses a KV-v2 secret as `v1/<mount>/data/<path>`, so the mount prefix doubled to `v1/secret/data/secret/meho/…` and failed the Vault ACL indistinguishably from a real permission denial.
- All six KV-v2 handlers (read / list / put / patch / versions / delete) now resolve their `(mount, path)` pair through a shared `_extract_mount_and_path(op_id, params)` helper that raises a new `VaultPathShapeError` **before** the Vault round-trip (and before the tenant-scope guard) when `path` begins with `"<mount>/"`. The message names the mount-relative form to use (e.g. `meho/test/federation`) and echoes only the non-secret mount/path the caller supplied — an actionable error over an opaque 403, per `docs/codebase/error-message-shape.md`.
- The guard **rejects** rather than silently strips: an operator whose tenant layout legitimately starts a path with the mount name should not be surprised by a silent rewrite. A bare single-segment `path == mount` is forwarded unchanged.
- Corrected the pre-existing `vault.kv.read` happy-path test, which used a mount-doubled path (`secret/meho/test/federation`) as its stand-in mount-relative value, to the genuinely mount-relative form the production health route uses (`_FEDERATION_PROOF_PATH = "meho/test/federation"`).

## Test plan

- [x] `ruff check` (ops.py + test file) — clean
- [x] `mypy src/meho_backplane/connectors/vault/ops.py` — clean
- [x] `code-quality.py --diff` — 0 blockers (3 pre-existing function-size warnings on untouched bodies)
- [x] **AC (a)** New parametrized test: `vault.kv.read/list/put/patch/versions/delete` with `path="secret/…"`, `mount="secret"` → `status="error"`, `extras.exception_class="VaultPathShapeError"`, and **hvac never invoked** (`login_calls == []` and every `*_calls == []`). Plus a non-default-mount case.
- [x] **AC (b)** Mount-relative path still works — existing vault read/list/put/patch/versions/delete tests stay green; a new `path == mount` boundary test confirms the guard does not over-reject.
- [x] **AC (c)** The error message names the mount-relative form: asserted (`"mount-relative" in message`, `"'team/api-key'" in message`); `grep -n "mount-relative\|VaultPathShapeError" backend/src/meho_backplane/connectors/vault/ops.py` hits.
- [x] `pytest tests/test_connectors_vault.py tests/test_connectors_vault_creds.py tests/test_api_v1_health.py tests/test_api_health_failures.py tests/test_vault_failures.py tests/test_operations_preview.py tests/test_broadcast_events.py tests/test_approval_queue.py` — **282 passed**, 1 pre-existing unrelated teardown error (see Out of scope).

## Out of scope

- The tenant-scope guard's own messaging (#1725) — this PR is specifically the mount-double-prefix path-shape hint. `VaultTenantScopeError` and `VaultPathShapeError` stay distinct; the path-shape check runs first so its more specific message wins.
- `tests/test_connectors_vault.py::test_preflight_fail_soft_when_probe_raises` has a pre-existing **teardown** error (a pytest output-scanner false-positive matching `secret=` in the rich-traceback repr of the deliberately-raised `VaultDown` fake). Reproduces unchanged on `main` (`def9015f`); not touched by this PR.
- Pre-existing function-size warnings on `vault_kv_read` / `vault_kv_write_capability_preflight` / `register_vault_typed_operations` (this PR's edit to `vault_kv_read` is net-shorter; the bodies are otherwise untouched).

Closes #1755

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,20 @@ connector-related release-notes line.
   default-deny. Tenant-scope-aware, idempotent, and audited as a
   single bulk-enable event with the count of ops enabled (#1749).
 
+### Fixed
+
+- `vault.kv.*` now returns an actionable path-shape hint instead of an
+  opaque `Forbidden` 403 when a caller passes a `path` that re-includes
+  the mount segment (`path="secret/meho/…"` with `mount="secret"`).
+  hvac addresses a secret as `v1/<mount>/data/<path>`, so the mount
+  prefix would double to `v1/secret/data/secret/meho/…` and fail the
+  Vault ACL indistinguishably from a real permission denial. All six KV
+  ops (read / list / put / patch / versions / delete) now reject the
+  mount-double-prefix before the Vault round-trip with a
+  `VaultPathShapeError` naming the mount-relative form to use
+  (e.g. `meho/test/federation`). A bare single-segment path equal to the
+  mount name is still forwarded unchanged (#1755).
+
 ## [0.15.0] - 2026-06-13
 
 ### Added
--- a/backend/src/meho_backplane/connectors/vault/ops.py
+++ b/backend/src/meho_backplane/connectors/vault/ops.py
@@ -79,6 +79,7 @@ from meho_backplane.retrieval.embedding import EmbeddingService
 __all__ = [
     "VAULT_KV_READ_PARAMETER_SCHEMA",
     "VAULT_KV_WRITE_CAPABILITIES",
+    "VaultPathShapeError",
     "register_vault_typed_operations",
     "vault_kv_delete",
     "vault_kv_list",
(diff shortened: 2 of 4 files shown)


damir-topic:
@coderabbitai review

Issue #1693, opened 2026-06-12
Task: Inject X-CSRF-Token into create-modal form — enable silent write to succeed


damir-topic:
Parent goal: #221
Parent initiative: #1691

**Severity:** SEV-2 · **Area:** frontend · **Source:** claude-rdc-hetzner-dc v0.14.0 cycle-10 dogfood

## Problem

The Memory create-modal form fails with 403 Forbidden (CSRF token invalid) on every POST submission. The middleware correctly rejects the request because the form does not send the `X-CSRF-Token` header — the CSRF validation chain in `csrf.py:_validate_csrf()` (lines 434-465) requires both the `meho_csrf` cookie AND the `X-CSRF-Token` header to be present and matching.

The code comments claim the form inherits `hx-headers` from the parent `<section>` element on the Memory list page, but HTMX does **not** inherit `hx-headers` from parent elements to child HTMX-driven elements. Each element must explicitly declare its headers.

## Current state

* Form in `/root/repos/evoila/meho/backend/src/meho_backplane/ui/templates/memory/_create_modal.html` lines 56-65 has `hx-post="/ui/memory/create"` but no `hx-headers` attribute
* Parent `<section>` on `memory/index.html` line 36 has `hx-headers='{"X-CSRF-Token": "{{ csrf_token }}"}'` which does not propagate to the modal form
* No global HTMX configuration in `base.html` line 67 (`<body>` tag has no `hx-headers`; no `htmx.config.defaultHeaders` or `htmx:configRequest` listener set up)
* Middleware logs "missing_token" rejection (csrf.py:521-524) when the form posts
* Response body is `{"detail":"csrf_token_invalid"}` with HTTP 403 and `hx-swap="none"` on the form means the operator sees zero visible feedback

## Desired state

The create-modal form POST succeeds with 204 and `HX-Redirect` to `/ui/memory`, returning the operator to the list page with the newly-created memory visible. This requires the `X-CSRF-Token` header to be echoed by the client.

## Acceptance criteria

- [ ] `hx-headers='{"X-CSRF-Token": "{{ csrf_token }}"}'` is present on the `<form id="memory-create-form">` element in `_create_modal.html` line 56
- [ ] Manual POST to `/ui/memory/create` via the modal succeeds with 204 when the header is sent
- [ ] Regression test: existing tests in `test_ui_memory_list.py` that POST to the create endpoint continue to pass
- [ ] No duplicate CSRF headers on other forms (edit detail, KB upload, connector import)

[...]

Issue #1752, opened 2026-06-14
Initiative: G0.25 v0.15.0 closed-loop dogfood hardening — 7 findings from the RDC Hetzner-DC v0.15.0 deploy (CSRF write regression + connector-resolution footgun + 5 gotchas)


damir-topic:
Parent goal: #221

## Why now

The RDC Hetzner-DC operators drove a **live v0.15.0 deploy** (2026-06-14) closed-loop dogfood pass (full feedback: #1751). v0.15.0 is strong on the connector + corpus side — vmware/k8s/vault/argocd/keycloak all dispatch live reads end-to-end, the vmware L2 `NotImplementedError` dead-end is gone, and the corpus round-trip returns cited chunks. This cycle surfaced **one SEV-2 regression the release notes don't reflect (UI writes still 403 on a polling page)**, a **SEV-2 connector-resolution footgun** (#1750), and four smaller hardening gaps.

All findings were **grounded + adversarially verified against the v0.15.0 code** before filing (the RDC discipline: symptoms reliable, root-cause/prior-art often wrong). Notable re-diagnoses:

- **The #1693 CSRF fix is only half-done.** #1693 fixed the modal *render-time* token echo, but the memory page's 60-second background poll re-mints + re-sets the `meho_csrf` cookie while the modal sits open, so the echoed token goes stale by submit time → 403. Not a render-race; a concurrent-poll-rotation race (T3).
- **#1750 is a resolution tie-break bug, not a priority bug.** `NsxConnector` already sets `priority=1`, but the auto-shim's narrower derived version range wins the most-specific-version step *before* priority is consulted — so a stray probe ingest shadows a shipped connector (T1).

## Scope — 8 Tasks

| T# | Sev | Area | Title |
|---|---|---|---|
| T1 · #1750 | SEV-2 | backend/connector | Resolver: a hand-rolled `Connector` must outrank a `GenericRestConnector` auto-shim for the same `(product, version)` label |
| T2 · #1753 | SEV-3 | backend/connector | Ingest near-miss guard for a sibling-impl_id class + name the registered sibling in the auto-shim dispatch error |
| T3 · #1754 | SEV-2 | frontend | Memory-create 403s under a background poll (#1693 partial): stop rotating `meho_csrf` on HTMX polls + surface the silent 403 |
| T4 · #1755 | SEV-3 | connector | `vault.kv.*`: detect a mount-included `path` and return a path-shape hint instead of an opaque Forbidden |
[...]

PR #1764, opened 2026-06-14
fix(memory-ui): stop CSRF cookie rotating under list polls + surface the silent create 403

### LINKED ISSUE #1754: Task: stop the CSRF cookie rotating under background polls (memory create 403) + surface the silent 403
Parent goal: #221
Parent initiative: #1752

## Summary

**Headline regression, not reflected in the v0.15.0 notes.** Operator-console **writes still 403** (`csrf_token_invalid`) on the memory page, despite #1693. The echo from #1693 attaches correctly, but a background list-poll rotates the CSRF cookie while the create modal is open, so the modal's render-time token is stale by submit. Compounding it, the 403 is **swallowed silently** — the modal stays open with no error.

## Root cause (grounded on v0.15.0)

- The memory cards fragment polls `GET /ui/memory` every 60s (`ui/templates/memory/_cards.html`, `hx-trigger="every 60s"`).
- Each `render_index` GET **re-mints and Set-Cookies a fresh `meho_csrf`** (`ui/routes/memory/views.py:363` mint, `:384` `_set_csrf_cookie`) — including HTMX poll fragments, not just full-page renders. The CSRF middleware passes safe GETs through and *expects* the handler to refresh the cookie (`ui/csrf.py:493-500`).
- #1693 fixed the modal **render-time** desync: the create form echoes its own minted token (`ui/routes/memory/create.py:94`, form `hx-headers` in `_create_modal.html`). But a poll firing *after* the modal renders rotates the cookie, so the echoed snapshot no longer matches → 403. On a polling page this reproduces under any non-instant submit.
- The 403 body is JSON (`{"detail":"csrf_token_invalid"}`, `ui/csrf.py:315-328`) and the form uses `hx-swap="none"` with **no `htmx:responseError` handler** → the rejection is silently dropped; the operator clicks Create and nothing happens.

## Proposed fix (lowest-friction)

1. **Stop rotating the cookie on HTMX fragment polls.** In `render_index`, set the `meho_csrf` cookie only on full-page renders, not on HTMX fragment responses (`is_htmx_request(request)` already exists). Full-page loads still mint once; the open modal's echoed token stays valid across polls.
2. **Surface the 403.** Give the create form an `htmx:responseError` (or `hx-on`) path that renders the rejection in the modal (visible message, values intact) instead of failing silently.

## Acceptance criteria

- [ ] A `GET /ui/memory?scope=all` with `HX-Request: true` returns **no** `Set-Cookie: meho_csrf=...` header; a full-page `GET /ui/memory` still sets a fresh cookie. Verifiable: `pytest backend/tests/...memory...` asserts the Set-Cookie presence/absence by request kind.
- [ ] Regression test: render the create modal, fire a simulated `every 60s` poll, then `POST /ui/memory/create` with the modal's paired cookie/header → **204**, not 403.
- [ ] On a genuine 403, the modal renders a visible error (the create form has an `htmx:responseError`/`hx-on` handler; `grep -n "responseError\|hx-on" backend/src/meho_backplane/ui/templates/memory/_create_modal.html`).
- [ ] No CSRF protection regression: a real cross-site/forged submit is still rejected (existing CSRF middleware tests stay green).

## Out of scope

- Reworking the double-submit scheme itself; this preserves the chassis pattern and only stops gratuitous rotation on safe polls.

## References

- Signal: #1751 (headline); prior art #1693 / #1174 (the echo half).
- Grounding: `ui/routes/memory/views.py:363,384`, `ui/csrf.py:315-328,493-500`, `ui/routes/memory/create.py:94`, `ui/templates/memory/_cards.html`, `ui/templates/memory/_create_modal.html`.
- Convention: htmx `htmx:configRequest`/`htmx:responseError` events (htmx 1.x) for the visible-error path.

damir-topic:
## Summary

- **Headline v0.15.0 dogfood regression:** the operator-console memory page still 403'd (`csrf_token_invalid`) on create despite #1693. The `_cards.html` list fragment polls `GET /ui/memory` every 60s through `render_index`, which re-minted and `Set-Cookie`-d a fresh `meho_csrf` on every render — **including the poll**. A poll firing while the create modal is open rotated the cookie out from under the modal's echoed token, so the next create POST failed the chassis double-submit check. The 403 is JSON and the form posts `hx-swap="none"` with no error handler, so it was **dropped silently** — the operator clicked Create and nothing happened.
- **Fix 1 — stop the rotation on safe polls:** `render_index` now sets the cookie on the **full-page render only**. On the HTMX fragment it reuses the token already on the request's cookie (validated via `verify_csrf_token` so a tampered value is never echoed back) and leaves the cookie untouched, so the open modal's token — and the cards fragment's own bulk-action echo — stay valid across polls. A fragment with no prior cookie falls back to a fresh mint + `Set-Cookie`. Mirrors the inline-refresh path's zero-`Set-Cookie` posture (G0.25 #1694).
- **Fix 2 — surface the silent 403:** the create form gains `hx-on::response-error` (htmx 2.x shorthand for `htmx:responseError`) that reveals an `alert alert-error` banner with a status-tailored message (403 = expired token + retry hint, network error = connectivity, other = the server's `detail`). `hx-swap="none"` keeps the field values intact for an immediate retry.
- The double-submit scheme is **unchanged** — a forged / cross-site submit is still rejected; only gratuitous rotation on safe polls stops. Out of scope per the issue: reworking the double-submit chassis itself.

## Test plan

- [x] `ruff format --check src/ tests/` — clean (3 files formatted)
- [x] `mypy src/` (strict) — Success, 474 files
- [x] Full unit suite `pytest -n 6 --dist loadscope --ignore=tests/integration` — **7522 passed, 195 skipped**
- [x] **AC (a)** `GET /ui/memory?scope=all` with `HX-Request: true` + a live cookie returns **no** `Set-Cookie: meho_csrf`; a full-page `GET /ui/memory` still sets a fresh cookie — `test_list_htmx_poll_does_not_rotate_csrf_cookie`, `test_list_full_page_render_sets_fresh_csrf_cookie`, plus the defensive-mint case `test_list_htmx_fragment_mints_cookie_when_none_present`.
- [x] **AC (b)** render modal → simulate the `every 60s` poll → `POST /ui/memory/create` with the modal's paired cookie/header → **204, not 403** — `test_create_after_background_poll_keeps_modal_token_valid_returns_204`.
- [x] **AC (c)** the create form wires a visible-error path (`hx-on::response-error` + the banner element + `window.memoryCreateShowError`) — `test_create_modal_renders_visible_error_handler`.
- [x] **AC (d)** no CSRF regression: a forged header is still **403 `csrf_token_invalid`** after a poll — `test_create_forged_header_still_rejected_after_poll_returns_403`; existing chassis CSRF middleware tests stay green.
- [x] Both rotation tests **fail against the pre-fix handler** (verified by reverting the guard) — they are genuine regression guards, not vacuous.

## Out of scope

- Adjacent finding (recorded, not edited): `render_detail` / `render_edit_form` / `patch_entry` also set the CSRF cookie unconditionally on their HTMX fragment responses (the detail page's `_body_view.html`). That is a latent variant of the same desync class, but the detail page has no `every Ns` poll trigger so it does not reproduce the regression; the issue scopes the fix to `render_index` (the poll source).

Closes #1754

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,18 @@ connector-related release-notes line.
   default-deny. Tenant-scope-aware, idempotent, and audited as a
   single bulk-enable event with the count of ops enabled (#1749).
 
+### Fixed
+
+- Operator-console memory **create no longer 403s** under a background
+  list refresh. The memory list's 60-second card poll re-used the
+  page handler, which re-minted and `Set-Cookie`-d a fresh CSRF token
+  on every render — rotating the cookie out from under an open create
+  modal so the next submit failed the double-submit check with a
+  silent `csrf_token_invalid`. The handler now sets the CSRF cookie on
+  full-page loads only; polls reuse the live cookie token and leave it
+  untouched, and the create modal now renders a visible error banner
+  instead of swallowing a rejected submit (#1754).
+
 ## [0.15.0] - 2026-06-13
 
 ### Added
--- a/backend/src/meho_backplane/ui/routes/memory/views.py
+++ b/backend/src/meho_backplane/ui/routes/memory/views.py
@@ -49,7 +49,11 @@ from meho_backplane.memory import (
     PermissionDeniedError,
 )
 from meho_backplane.ui.auth.middleware import UISessionContext
-from meho_backplane.ui.csrf import CSRF_COOKIE_NAME, mint_csrf_token
+from meho_backplane.ui.csrf import (
+    CSRF_COOKIE_NAME,
+    mint_csrf_token,
+    verify_csrf_token,
+)
 from meho_backplane.ui.routes.memory.bulk import (
     BULK_EXTEND_DURATIONS,
     format_countdown,
@@ -261,6 +265,41 @@ def _set_csrf_cookie(response: HTMLResponse, csrf_token: str) -> None:
     )
 
 
+def _resolve_list_csrf(request: Request, session_id: str, *, is_htmx: bool) -> tuple[str, bool]:
+    """Pick the CSRF token the list render echoes + whether to set the cookie.
+
+    Returns ``(token, set_cookie)``. The list page and its
+    ``hx-trigger="every 60s"`` cards fragment share one handler
+    (:func:`render_index`); the rule (#1754) is:
+
+    * **Full-page render** -- mint a fresh token, echo it, and set the
+      ``meho_csrf`` cookie so the double-submit pair is established for
+      the freshly loaded page.
+    * **HTMX fragment render (the poll)** -- *reuse* the token already
+      carried by the request's ``meho_csrf`` cookie and do **not** set
+      the cookie. A fresh mint + ``Set-Cookie`` on every poll rotates
+      the token out from under any open create modal (whose echoed
+      snapshot from #1693 then fails the middleware's cookie/header
+      match) and out from under the cards fragment's own bulk-action
+      ``hx-headers`` echo. Reusing the live cookie token keeps the
+      rendered echo aligned with the un-rotated cookie, so both the
+      modal create POST and the bulk POST still validate.
+
+    The reuse is gated on :func:`verify_csrf_token` so a tampered or
+    foreign cookie value never gets echoed back as the page's token; a
+    missing-or-invalid cookie on a fragment fetch (defensive -- e.g. a
+    direct fragment request with no prior full-page load) falls back to
+    a fresh mint + ``Set-Cookie`` so the fragment's own forms stay
+    functional.
+    """
+    if not is_htmx:
+        return mint_csrf_token(session_id), True
+    existing = request.cookies.get(CSRF_COOKIE_NAME)
+    if existing and verify_csrf_token(session_id, existing):
+        return existing, False
+    return mint_csrf_token(session_id), True
+
+
 def _common_template_context(
     session_ctx: UISessionContext,
     csrf_token: str,
@@ -355,12 +394,24 @@ async def render_index(
     ``hx-trigger="every 60s"`` on the cards fragment is what refreshes
     the countdown badges; the same handler responds to both the full
     page render and the HTMX poll.
+
+    The ``meho_csrf`` cookie is set on the full-page render **only**,
+    never on the HTMX fragment response (#1754). The cards fragment
+    polls this same handler every 60 seconds; setting the cookie on a
+    poll would rotate the double-submit token out from under any open
+    create modal, whose echoed token snapshot (#1693) would then fail
+    the middleware's cookie/header match and 403 the create POST. The
+    full-page load mints the cookie once; the open modal's token stays
+    valid across polls because the poll no longer touches the cookie.
+    This mirrors the inline-refresh path's zero-``Set-Cookie`` posture
+    (G0.25 #1694) that keeps in-flight pages' CSRF tokens stable.
     """
     scope_filter = resolve_scope_filter(scope)
     operator = build_read_operator(session_ctx)
     entries = await _list_for_render(operator, scope_filter, tag)
     active, recently_expired = partition_expired(entries)
-    csrf_token = mint_csrf_token(str(session_ctx.session_id))
+    is_htmx = is_htmx_request(request)
+    csrf_token, set_csrf = _resolve_list_csrf(request, str(session_ctx.session_id), is_htmx=is_htmx)
     # The refresh URL preserves the active scope + tag so the HTMX
     # poll re-renders with the same filter state. T1's tabs already
     # encode the query string into the request URL; the fragment
@@ -379,9 +430,10 @@ async def render_index(
         "bulk_extend_durations": BULK_EXTEND_DURATIONS,
         "refresh_url": refresh_url,
     }
-    template_name = "memory/_cards.html" if is_htmx_request(request) else "memory/index.html"
+    template_name = "memory/_cards.html" if is_htmx else "memory/index.html"
     response = get_templates().TemplateResponse(request, template_name, context)
-    _set_csrf_cookie(response, csrf_token)
+    if set_csrf:
+        _set_csrf_cookie(response, csrf_token)
     return response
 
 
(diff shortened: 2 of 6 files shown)


damir-topic:
@coderabbitai review

damir-topic:




**Verdict: APPROVE**

> Posted as a PR comment, not a review state: own-author `gh pr review --approve` is API-blocked (reviewer == author). The authoritative verdict is the machine-readable result file referenced above.

**Anchor task:** #1754 — stop the `meho_csrf` cookie rotating under HTMX background polls (memory-create 403) + surface the silent 403
**Parent:** Initiative #1752 (G0.25 v0.15.0 dogfood hardening) → Goal #221
**Surface area:** backend (HTMX/Jinja UI) + small inline JS
**Best-practices packs audited:** python, frontend
**Head SHA reviewed:** `6868bf98b3a617251a01da64c64c4f38123e575f` (re-checked at post time — no drift)
**AI reviewer (CodeRabbit):** not received — rate-limited (posted a "Review limit reached" notice, not a review; no review node on the PR). Recorded `requested=true, received=false, reason="rate-limited"` and proceeded.

### Acceptance criteria
- [x] **(a)** `GET /ui/memory?scope=all` with `HX-Request: true` returns **no** `Set-Cookie: meho_csrf`; a full-page `GET /ui/memory` still sets a fresh cookie — `test_list_htmx_poll_does_not_rotate_csrf_cookie` + `test_list_full_page_render_sets_fresh_csrf_cookie` (+ the defensive-mint case `test_list_htmx_fragment_mints_cookie_when_none_present`). Green.
- [x] **(b)** render modal → simulated `every 60s` poll → `POST /ui/memory/create` with the modal's paired cookie/header → **204, not 403** — `test_create_after_background_poll_keeps_modal_token_valid_returns_204`. **Verified non-vacuous**: reverted `render_index` to the pre-fix handler (unconditional mint + `Set-Cookie`) and this test plus (a)'s poll test both **FAILED** (the #1754 regression reproduced — "the background poll rotated meho_csrf"); they pass post-fix. Worktree restored to PR HEAD afterward.
- [x] **(c)** genuine 403 renders a **visible error** in the modal, field values intact — the form carries `hx-on::response-error="window.memoryCreateShowError(this, event)"` + an `alert alert-error` banner; `hx-swap="none"` preserves the field input for retry; `hx-on::before-request` re-hides the banner so a retry starts clean. `test_create_modal_renders_visible_error_handler`. (htmx **2.0.9** is bundled — the `hx-on::` double-colon shorthand and `event.detail.xhr.status`/`.responseText` reads are version-correct; confirmed against https://htmx.org/attributes/hx-on/ and https://htmx.org/events/.)
- [x] **(d)** **No CSRF regression** (load-bearing) — `test_create_forged_header_still_rejected_after_poll_returns_403` (a forged header through the real modal+poll sequence is still `403 csrf_token_invalid`); this passes **both** pre- and post-fix, confirming the double-submit defense is independent of the rotation behavior. The 5 dedicated chassis CSRF tests are green (rejects no-token / mismatched cookie-header / forged signature; accepts matching; fails-fast on empty key). `ui/csrf.py` is **untouched**, and the poll-path token reuse is gated on `verify_csrf_token`, so a tampered/foreign cookie value is never echoed back as the page token.

### Predicted vs. actual diff
- ✅ `ui/routes/memory/views.py` — `render_index` sets the cookie on full-page renders only; the HTMX poll reuses + validates the request's existing token via the new `_resolve_list_csrf` helper.
- ✅ `ui/templates/memory/_create_modal.html` — `hx-on::response-error` visible-error path + idempotent `window.memoryCreateShowError`.
- ✅ `tests/test_ui_memory_list.py`, `tests/test_ui_memory_create_promote.py` — rotation + regression + forged-header guards.
- ✅ `docs/codebase/ui.md`, `CHANGELOG.md` — updated to match.
- 🕳️ Nothing missing; 🚫 nothing undisclosed.

### Integration verification
- [x] `ruff format --check` — clean (3 files already formatted)
- [x] `mypy` — `views.py` has **0** errors. (The only 2 mypy errors observed — `broadcast/probe.py:74`, `broadcast/stream.py:205` — are in modules **not** touched by this PR; pre-existing and unrelated, not attributable to the diff.)
- [x] `pytest` — **63 passed** across the CSRF / memory-list / memory-create suites; the 5 chassis CSRF tests green.
- [x] **Non-vacuity probe** — pre-fix revert makes the two rotation guards fail (regression reproduced), post-fix they pass.
- [x] Tests run with `PYTHONPATH` pinned to the PR worktree `backend/src` (the editable install pointed at a different worktree — the pin was load-bearing).
- ⚠️ not run: `docker compose up` (no compose-affecting change). UI flow exercised via the TestClient end-to-end create/poll/forge sequence rather than a live browser; the JS handler's branch behavior is render-asserted (no executing-JS harness exists in this backend-only repo) and corroborated by htmx 2.0.9 docs.

### Findings (inline)
None. No blockers, no majors, no minors.

- `render_detail` / `render_edit_form` / `patch_entry` still `Set-Cookie` the CSRF token unconditionally on their HTMX fragment responses. This is a **latent variant** of the same desync class, but it does **not** reproduce today because the detail surface has no `every Ns` poll trigger — so scoping the fix to `render_index` (the poll source) is defensible, not a missed criterion. Worth a follow-up ticket to fold these onto the same safe-GET no-rotation posture (a shared helper) so the rule is uniform across the memory surface. Disclosed by the author as out-of-scope.

- Adopted: 0 · Disputed: 0 · Out of scope: 0 — CodeRabbit was rate-limited this cycle and landed no findings.

### Docs
- [x] `docs/codebase/ui.md` — updated (the full-page-only cookie posture for `render_index`, and the `hx-on::response-error` visible-error path).

### What worked well
- The non-vacuity is real: the two rotation guards genuinely fail against the pre-fix handler, so they're regression guards rather than tautologies.
- The poll-path token **reuse is gated on `verify_csrf_token`**, so the fix stops gratuitous rotation without ever echoing a tampered cookie value — the double-submit verification is left fully intact (and `csrf.py` is untouched).
- Thorough, well-cited inline rationale (htmx 2.x `hx-on::` shorthand, the #1693 echo half, the G0.25 #1694 zero-`Set-Cookie` precedent) and a clean disclosure of the adjacent out-of-scope finding.

Issue #1784, opened 2026-06-15
Task: per-target CA-pin via load_verify_locations(cadata) — secure TLS-trust supersession


damir-topic:
Parent goal: #214
Parent initiative: #1774
Depends-on: #1780

## Problem

`verify_tls=false` (T1/T2, #1774) unblocks self-signed appliances but drops **all** verification (no chain, no hostname), exposing the target credential to MITM. The secure answer for a self-signed/internal-CA endpoint is to **pin its CA**: trust exactly that CA while keeping `CERT_REQUIRED` + hostname checking. This is the govc-thumbprint pattern and the fix the `verify_tls=false` error/docs point operators toward. This Task is the secure supersession; it is a follow-up to the minimum ship (T1–T4).

## Current state (verified in code)

- After T1/T2: `Target` has `verify_tls: bool`; [`adapters/http.py`](backend/src/meho_backplane/connectors/adapters/http.py) builds an insecure `SSLContext` only when `verify_tls=false`, and the pool key carries an `extra_cache_dimensions(target)` hook. There is **no** per-target CA-trust material.
- [`_vcsim.py:331-333`](backend/tests/acceptance/_vcsim.py#L331-L333) shows the `create_default_context()` idiom; the secure variant uses `load_verify_locations(cadata=...)` instead of `CERT_NONE`.
- Research note: stdlib `ssl` has **no native fingerprint pinning** — a CA-PEM pin via `load_verify_locations` is the idiomatic, low-risk path.

## Desired state

- Add a per-target CA-trust field carrying a PEM string (decide column vs `extras` in the PR — lean **first-class column** for typed validation, consistent with T1's `verify_tls`). Mutually exclusive with `verify_tls=false` in spirit (a pin makes insecure unnecessary); document the precedence.
- In `HttpConnector._http_client`, when the CA-pin is set, build `ctx = ssl.create_default_context(); ctx.load_verify_locations(cadata=<pem>)` — which **keeps** `verify_mode=CERT_REQUIRED` and `check_hostname=True`, so chain + hostname are still enforced against the pinned CA. Extend the **same** `extra_cache_dimensions` hook from T2 so the pool key incorporates the CA-material identity (e.g. a digest of the PEM) — a changed pin must not be served by a stale client.
- Update the `connector_tls_verify_failed` message + docs to point at the CA-pin as the preferred fix (closing the loop from T3/T4).

## Acceptance criteria

[...]

Issue #1783, opened 2026-06-15
Task: document verify_tls=false as last resort + CA-bundle as the secure path


damir-topic:
Parent goal: #214
Parent initiative: #1774
Depends-on: #1780

## Problem

`verify_tls=false` (T1/T2, #1774) disables TLS verification for a target, which exposes that target's Vault-resolved credential to MITM on every dispatch. It must be documented as a **per-target, default-secure, never-global last resort**, with the secure alternatives (global trust bundle, and the CA-pin follow-up) framed as the preferred fixes — otherwise operators will reach for it reflexively.

## Current state (verified in code)

- `deploy/values-examples/README.md` documents the global trust-bundle / `SSL_CERT_FILE` story (#209) but has no per-target TLS guidance.
- `docs/architecture/connectors.md` describes connector dispatch but not per-target TLS trust.
- Prior art for *why* operators hit this: #572 (internal-CA mounts clobber public roots) — the footgun that makes a blanket trust bundle awkward and pushes operators toward per-target escape hatches.

## Desired state

Add an operator-facing section (in `deploy/values-examples/README.md` and/or `docs/architecture/connectors.md`) that:
- Describes `verify_tls=false` as **per-target, default-secure (`true`), never-global**, set via `POST`/`PATCH /api/v1/targets`, with an explicit MITM / credential-exposure caveat.
- Names the **secure** fix first: add the appliance's CA to the trust bundle via `SSL_CERT_FILE` / the chart trust-bundle injection (#209); calls out the #572 public-roots-clobber footgun so operators scope the bundle correctly.
- Names **CA-pin (T5, the planned follow-up)** as the secure supersession for self-signed appliances (keeps MITM protection).
- Cites the prior-art shape operators already know: govc `-k`/`GOVC_INSECURE`, kubectl per-cluster `insecure-skip-tls-verify`.
- Notes the **coverage gap**: the k8s reachability probe and the GitHub App token-exchange are out-of-pool and do **not** honor `verify_tls`.

## Acceptance criteria

- [ ] Docs state `verify_tls=false` is per-target, default-secure, never-global, and exposes the target credential to MITM.
- [ ] Docs name `SSL_CERT_FILE`/trust-bundle (#209) as the recommended fix and reference the #572 clobber footgun, and name CA-pin (T5) as the planned secure supersession.
- [ ] Docs note the two out-of-pool connectors (k8s probe, github token-exchange) do not honor `verify_tls`.
[...]

Issue #1782, opened 2026-06-15
Task: add connector_tls_verify_failed dispatch error arm + builder + error-shape doc row


damir-topic:
Parent goal: #214
Parent initiative: #1774

## Problem

When connector dispatch fails TLS verification, an `httpx.ConnectError` (no `.response`) skips the `HTTPStatusError` arm and falls through to the generic `except Exception` → opaque `connector_error: ConnectError`, discarding the SSL cause. The operator sees `[SSL: CERTIFICATE_VERIFY_FAILED]` with no guidance. This Task adds a targeted `connector_tls_verify_failed` result that names the host and both remediations. It is **fully decoupled** from T1/T2 — no schema change, mergeable in parallel.

## Current state (verified in code)

- [`operations/dispatcher.py:828`](backend/src/meho_backplane/operations/dispatcher.py#L828) — `except httpx.HTTPStatusError` arm (403/422 special-cased); [L872](backend/src/meho_backplane/operations/dispatcher.py#L872) — generic `except Exception` → `result_connector_error(op_id, exc, duration_ms)`. There is **no** `except httpx.ConnectError` arm. (Precedent for inserting a specific arm ahead of the catch-all: the docstring at L674.)
- [`operations/_errors.py`](backend/src/meho_backplane/operations/_errors.py) — `result_connector_error` (L432), `result_connector_http_403` (L670), `result_connector_http_422` (L742), `__all__` (L33). New builders slot in alongside.
- [`adapters/http.py`](backend/src/meho_backplane/connectors/adapters/http.py) — `_retryable` retries `ConnectError`/`ConnectTimeout`/`ReadError` on idempotent verbs first, so the new arm fires only after retries exhaust.

## Desired state

- `_errors.py`: add `result_connector_tls_verify_failed(op_id, exc, target, duration_ms)` (+ `__all__` entry) alongside the 403/422 builders. Message names `target.host` + **both** remediations — the secure `SSL_CERT_FILE`/CA-bundle path **and** `verify_tls=false` as the audited last resort (with the MITM/credential-exposure caveat) — plus a doc ref. Structured extras: `{error_code: "connector_tls_verify_failed", host, exception_class, exception_message (raw "[SSL: CERTIFICATE_VERIFY_FAILED]…" capped ~256 chars), remediation_secure, remediation_last_resort}`; `status="error"`.
[...]

Issue #1781, opened 2026-06-15
Task: wire verify_tls into HttpConnector._http_client — insecure SSLContext + cache-key dimension


damir-topic:
Parent goal: #214
Parent initiative: #1774
Depends-on: #1780

## Problem

`HttpConnector._http_client` builds the pooled `httpx.AsyncClient` with **no `verify=` argument**, so every connector dispatch verifies against the global `SSL_CERT_FILE` bundle only — there is no way to reach a self-signed/internal-CA appliance (the #1774 self-signed case). This Task makes the client honor the per-target `verify_tls` flag added in T1: when `verify_tls=False`, build the client with an explicit insecure `SSLContext`; when `True`, leave the global path byte-identical.

## Current state (verified in code)

- [`adapters/http.py:81-95`](backend/src/meho_backplane/connectors/adapters/http.py#L81-L95) — the single production pooled-client builder; passes `base_url`, `timeout`, `follow_redirects=True` and **no** `verify=`. Module docstring (L27-28): *"no extra cert logic is needed here beyond not overriding the default `verify` flag."*
- [`adapters/http.py`](backend/src/meho_backplane/connectors/adapters/http.py) — `self._clients: dict[tuple[str, str], httpx.AsyncClient]` keyed by `target_cache_key(target)` → `(tenant_id, id)` (the #1682/#1642 tenant-isolation key); built lazily under `self._lock`.
- [`_vcsim.py:331-333`](backend/tests/acceptance/_vcsim.py#L331-L333) — the in-tree insecure-context idiom: `ssl.create_default_context()` → `check_hostname = False` → `verify_mode = ssl.CERT_NONE` (this **order** — the reverse raises `ValueError` on Python 3.12).
- `HarborConnector` and `GcloudConnector` delegate to `super()._http_client`, so a base-class change reaches them for free.

## Desired state

- Add `import ssl` and a module-level, **cached** `_insecure_ssl_context()` returning `ctx = ssl.create_default_context(); ctx.check_hostname = False; ctx.verify_mode = ssl.CERT_NONE; ctx` (that order). Build once at module scope and reuse (SSLContext is shareable; never mutate after first use).
[...]

PR #1789, opened 2026-06-15
feat(connectors): add per-target verify_tls column + audit-binding

### LINKED ISSUE #1780: Task: add verify_tls column + migration + thread through Target schemas/projections + audit-binding
Parent goal: #214
Parent initiative: #1774

## Problem

`Target` has no per-target TLS-verification field, so connector dispatch can only verify against the global `SSL_CERT_FILE` bundle. This Task adds the **storage + API surface** for a per-target `verify_tls` flag (default-secure); the dispatch wiring that consumes it lands in T2 (#1774). It also closes a verified never-silent gap: target create/update writes an **empty** audit payload today, so there is no durable record of a security-relevant config change.

- [`db/models.py:875`](backend/src/meho_backplane/db/models.py#L875) — `Target` ORM has `vpn_required: Mapped[bool] = mapped_column(...)` (Boolean, NOT NULL, default) but **no** `verify_tls`.
- [`targets/schemas.py`](backend/src/meho_backplane/targets/schemas.py) — `TargetSummary` (L69), `Target` (L122), `TargetCreate` (L192), `TargetUpdate` (L233) each thread `vpn_required`/`version` (e.g. `vpn_required: bool` L174, `vpn_required: bool | None = None` L280); `project_target_to_summary` projects `vpn_required=t.vpn_required` (L323).
- [`api/v1/targets.py`](backend/src/meho_backplane/api/v1/targets.py) — `_to_full` (L373), `create_target` (L811, `POST ""`), `update_target` (L917, `PATCH /{name}`). `update_target` binds **no** `audit_*` contextvars today → empty `audit_log` payload.
- [`audit.py:235`](backend/src/meho_backplane/audit.py#L235) — `_resolve_audit_payload()` merges `audit_*` contextvars into the request's `audit_log` row.

## Desired state

- ORM: add `verify_tls: Mapped[bool] = mapped_column(sa.Boolean(), nullable=False, default=True)` next to `vpn_required`.
- New Alembic migration adding the column with `server_default=sa.true()` (existing rows backfill to secure). **Verify the live alembic head at implementation time** (tree is at 0043; next is ~0044) and renumber if a parallel wave grabbed it — mirror the `version`-column migration 0032 (#1215) precedent.
- Schemas: `Target` / `TargetCreate` / `TargetSummary` get `verify_tls: bool = True`; `TargetUpdate` gets `verify_tls: bool | None = None` (None = field not sent, matching every PATCH field). `project_target_to_summary` and `_to_full` pass it through like `version`. `verify_tls` is connection-routing → it MUST be listable in `TargetSummary` (per the api-shape-conventions rule the summary docstring cites).
- Audit fold-in: in `create_target` (when `verify_tls is False`) and `update_target` (when `verify_tls` is in the patch), bind `audit_*`-prefixed contextvars (`audit_tls_verification_disabled=True`, `audit_target_id`, `audit_verify_tls_before`/`audit_verify_tls_after`) so `_resolve_audit_payload` merges them; emit a structured WARN log. Use the `audit_*` prefix — **not** a bare `target_id` (that is the soft-FK column key the middleware reads separately).
- Run `cd cli && make snapshot-openapi && make generate` (`verify_tls` is a new `Target` field → the OpenAPI `Target` schema **and** the generated Go struct change) and commit the result.
- Update `tests/integration` `TargetORM` seeders + connector test doubles to carry `verify_tls`, or the testcontainers lane goes red while `uv run pytest` stays green (the known integration-double trap).

## Acceptance criteria

- [ ] `POST /api/v1/targets` accepts `verify_tls` and defaults to `true` when omitted; existing rows read back `verify_tls=true` after `alembic upgrade head`.
- [ ] `PATCH /api/v1/targets/{name}` with `{"verify_tls": false}` persists; `GET` detail **and** `GET` list (`TargetSummary`) both return `verify_tls`.
- [ ] Migration is reversible (`upgrade head` → `downgrade`) and leaves a single alembic head (no multiple-heads).
- [ ] A PATCH setting `verify_tls=false` writes an `audit_log` row whose payload contains `tls_verification_disabled=true` + `target_id` + before/after; a PATCH that does not touch `verify_tls` binds **no** TLS audit keys.
- [ ] OpenAPI snapshot + generated Go `Target` struct regenerated and committed; `uv run pytest` + `cd cli && go build ./...` green; `tests/integration` target doubles updated.

## Out of scope

- Consuming `verify_tls` in client construction (T2, #1774). This Task only stores + exposes + audits the flag.
- CA-pin (T5) and approval-gating of target-config mutations.

## References

- Precedent: [`db/models.py:875`](backend/src/meho_backplane/db/models.py#L875) (`vpn_required`), migration 0032 / #1215 (`version` column), [`audit.py:235`](backend/src/meho_backplane/audit.py#L235) (`_resolve_audit_payload`).
- Alembic add-column + `server_default`: https://alembic.sqlalchemy.org/en/latest/ops.html
- SQLAlchemy 2.0 `mapped_column` / `server_default`: https://docs.sqlalchemy.org/en/20/orm/mapping_columns.html
- Pydantic v2 optional/PATCH field semantics: https://docs.pydantic.dev/latest/concepts/models/

damir-topic:
## Summary

- Add a first-class **`verify_tls`** flag to `Target` (`NOT NULL DEFAULT true`, default-secure), mirroring the proven `vpn_required` / `version` first-class-column precedent rather than an unvalidated `extras` key. This ships the **storage + API + audit** surface for a per-target TLS-verification opt-out; the dispatch path that *consumes* the flag (passing `verify=<insecure SSLContext>` to the pooled httpx client) lands in the follow-up T2 (#1781).
- **Audit-binding** on create (when `verify_tls=false`) and update (when the PATCH sends `verify_tls`): binds `audit_*` contextvars so `AuditMiddleware` folds `tls_verification_disabled` + `target_id` + before/after into the `audit_log` payload, plus a WARN log. This closes the never-silent gap the initiative flagged — a target PATCH wrote an *empty* audit payload today.
- **Alembic migration `0044`** (live head verified as `0043` via `alembic heads`; `down_revision="0043"`) adds the column with `server_default=sa.true()` so pre-existing rows backfill to the secure state and the `NOT NULL` add-column is safe on a populated table. Reversible.

## Test plan

- [x] `ruff check` — clean
- [x] `ruff format --check` — clean (8 files)
- [x] `mypy src/` (strict) — clean (480 files)
- [x] `.claude/hooks/code-quality.py --diff` — 0 blockers (extracted `_validate_preferred_impl_id` + `_apply_product_patch` to keep both handlers within budget; remaining warnings are pre-existing legacy debt)
- [x] **AC1** POST defaults `verify_tls=true` when omitted; existing rows read back `true` after `alembic upgrade head` — `test_target_create_defaults`, `test_migration_0044_backfills_verify_tls_true_and_is_reversible`
- [x] **AC2** PATCH `{"verify_tls": false}` persists; `GET` detail **and** `GET` list (`TargetSummary`) both return `verify_tls` — schema round-trip + `test_patch_verify_tls_false_writes_tls_audit_payload`
- [x] **AC3** migration reversible (`upgrade head` → `downgrade`), single head — verified live (`alembic upgrade head` → `downgrade -1` → `upgrade head`, ends `0044 (head)`) + `test_migration_0044_*`
- [x] **AC4** PATCH `verify_tls=false` writes an `audit_log` payload with `tls_verification_disabled=true` + `target_id` + before/after; a PATCH not touching `verify_tls` binds **no** TLS audit keys — `test_patch_verify_tls_false_writes_tls_audit_payload`, `test_patch_without_verify_tls_binds_no_tls_audit_keys`
- [x] **AC5** OpenAPI snapshot + generated Go `Target` struct regenerated + committed; `go build ./...` green; `git diff --exit-code` on the two CLI files exits 0 — `cd cli && make snapshot-openapi && make generate`
- [x] Targets-area pytest — 175 passed, 1 skipped (sandbox); blast-radius sweep on Target-consuming modules — 143 passed

## Out of scope

- Wiring `verify_tls` into `HttpConnector._http_client` (insecure `SSLContext` + cache-key dimension) — T2 (#1781).
- `connector_tls_verify_failed` dispatch error arm — T3 (#1782); operator docs — T4 (#1783); per-target CA-pin — T5 (#1784).
- The connector dataclass test doubles (`_VcsimTarget`, etc.) deliberately carry only the attributes the connector-under-test reads; no connector reads `verify_tls` in T1, so they grow the attribute in lockstep with T2's client read, not here. The real-ORM integration seeders are safe via the ORM `default=True` + the migration's `server_default` backfill.

Closes #1780

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,15 @@ connector-related release-notes line.
 
 ### Added
 
+- Targets now carry a first-class **`verify_tls`** flag
+  (`NOT NULL DEFAULT true`, default-secure) on create / update / read
+  across `POST` / `PATCH` / `GET /api/v1/targets` and the list
+  projection. It is the storage + API + audit surface for a per-target
+  TLS-verification opt-out (the dispatch wiring that consumes it lands
+  in a follow-up); setting it `false` writes a durable `audit_log` row
+  (`tls_verification_disabled` + before/after) and a WARN log, closing
+  the prior gap where a target PATCH wrote an empty audit payload
+  (#1780).
 - Bulk read-class connector enable path across REST + MCP + CLI:
   `POST /api/v1/connectors/{id}/enable-reads`, the
   `meho.connector.enable_reads` MCP tool, and `meho connector
--- a/backend/alembic/versions/0044_add_targets_verify_tls.py
+++ b/backend/alembic/versions/0044_add_targets_verify_tls.py
@@ -1,0 +1,104 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""Add ``targets.verify_tls`` column for per-target TLS-verification opt-out.
+
+Revision ID: 0044
+Revises: 0043
+Create Date: 2026-06-15
+
+T1 (#1780) under Initiative #1774, Goal #214. Connector dispatch can
+only verify the target endpoint's certificate against the global
+``SSL_CERT_FILE`` bundle today, so a target presenting a self-signed or
+internal-CA cert (a nested lab, a freshly-deployed appliance before cert
+replacement, a Fleet-managed component with its own locker CA) dies on
+``CERTIFICATE_VERIFY_FAILED``. This migration ships the *storage* half
+of a per-target opt-out: a ``verify_tls`` flag, default-secure, mirroring
+the proven ``vpn_required`` first-class column (``db/models.py``) -- not
+an ``extras`` key, because connection-affecting config is the
+first-class-column precedent (``version`` #1215, ``auth_model`` CHECK).
+The dispatch path that *consumes* the flag (passing
+``verify=<insecure SSLContext>`` to the pooled httpx client when the flag
+is ``False``) lands in T2 (#1781); this column only stores + exposes +
+audits the flag.
+
+What this migration adds
+------------------------
+
+* ``targets.verify_tls boolean NOT NULL DEFAULT true`` -- whether
+  connector dispatch verifies the target's TLS certificate chain.
+  ``True`` (the secure default) keeps the dispatch client byte-identical
+  to today; ``False`` is the audited per-target opt-out. No index -- the
+  column is read per-request from the in-memory
+  :class:`~meho_backplane.db.models.Target` row the resolver already
+  loaded; it is never a filter predicate.
+
+Why ``server_default=true`` (not the bare ORM ``default``)
+----------------------------------------------------------
+
+``vpn_required`` (migration 0004) carries no server default because it
+was created with the ``targets`` table, so there were no pre-existing
+rows to backfill. ``verify_tls`` is an ``ADD COLUMN`` on a populated
+table: a ``NOT NULL`` add-column with no default is rejected by both
+PostgreSQL and SQLite, so the migration must supply a ``server_default``
+to backfill every existing row to the secure ``true`` state in one DDL
+statement. The ORM column declares the same ``server_default=sa.true()``
+so ``MetaData``-driven schema creation stays in sync with the migrated
+schema. The acceptance contract ("existing rows read back
+``verify_tls=true`` after ``alembic upgrade head``") is exactly this
+backfill.
+
+Why additive (not amend ``0004``)
+---------------------------------
+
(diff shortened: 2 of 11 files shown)


damir-topic:
@coderabbitai review

Issue #1774, opened 2026-06-15
Initiative: Per-target TLS trust for connector dispatch — reach self-signed appliances without re-issuing certs


damir-topic:
Parent goal: #214

## Why now

Filed from dogfooding meho as a log-sentry — a scheduled agent that queries VCF Operations for Logs and escalates findings. Registering a `vcf-logs` target at a nested-lab vRLI appliance and dispatching `GET:/version` dies on TLS:

```
connector_error: ConnectError
[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain
```

Everything else checks out — credential resolution from the broker, the port, and reachability are all fine; the socket opens and only cert-chain verification fails. A lot of what a VCF governance backplane must reach presents self-signed or internal-CA certs: nested labs, freshly-deployed appliances before cert replacement, anything Fleet-managed with its own locker CA. **The ClaudeVCF wrappers we're replacing all use insecure TLS for exactly this** — so this is direct connector-parity for Goal #214: an operator cannot retire the wrapper until meho can reach the same endpoint. Today the only way in is re-issuing the appliance cert from a CA meho already trusts — a fragile, GUI-breaking dance on a Fleet-managed component, wiped on the next lab rebuild.

## Root cause (verified against HEAD)

Two independent gaps, both confirmed in code:

1. **No per-target TLS escape hatch.** `HttpConnector._http_client` builds the pooled `httpx.AsyncClient` with **no `verify=` argument** (`backend/src/meho_backplane/connectors/adapters/http.py:81-95`) → httpx verifies against the global `SSL_CERT_FILE` bundle only. The module docstring is explicit: *"no extra cert logic is needed here beyond not overriding the default `verify` flag."* Nothing reads `target.extras` for TLS, which is why the reporter's `verify_tls` / `insecure` / `tls_skip_verify` extras keys had **zero** effect (`TargetCreate`/`TargetUpdate` are `extra='forbid'` on top-level fields, but `extras` contents are unvalidated and no connector consults them for TLS).
2. **Opaque failure.** A `ConnectError` has no `.response`, so it skips the `except httpx.HTTPStatusError` arm and falls through to the generic `except Exception` at `operations/dispatcher.py:872` → `connector_error: ConnectError`, discarding the SSL cause.

## Approach — minimum-shippable opt-in, default-secure

[...]

Issue #1682, opened 2026-06-12
Task: key the shared HttpConnector._clients pool (+ ssh._connections) by (tenant_id, target.id) — cross-tenant request misrouting


damir-topic:
Parent goal: #214

## Problem

#1642/#1672 re-keyed every connector **credential/session** cache by `(tenant_id, target.id)`,
but the shared **connection pool** `HttpConnector._clients` (and the SSH analogue
`adapters/ssh.py::_connections`) is still keyed on **`target.name`** — and each pooled
`httpx.AsyncClient` is **bound to a host** via `base_url=self._base_url(target)`. Targets
are unique only on `(tenant_id, name)`, so two tenants can each have a target named
`prod-vcenter` pointing at **different hosts**. First-writer-wins: tenant A dispatches first →
`_clients["prod-vcenter"]` is bound to **A's host**; tenant B's later dispatch against *its
own* `prod-vcenter` **reuses A's client and is sent to A's host** — request misrouting **and**
a credential leak across the tenant boundary (B's now-correctly-keyed creds → A's server).

The #1640/#1641 authz fixes do **not** prevent this: each operator legitimately accesses
their *own* same-named target; the collision is in the process-global pool, below the authz
layer. This is the residual, lower-layer half of the cross-tenant cache bleed.

## Current state (verified in code, main)

- `connectors/adapters/http.py:59` `self._clients: dict[str, httpx.AsyncClient] = {}`;
  created at `:65-67` (`base_url=self._base_url(target)`), read at `:81`, drained at `:195-197`.
- `connectors/adapters/ssh.py` — `_connections` live `SSHClientConnection` pool, same `target.name` keying.
- `vmware_rest/connector.py` reads the `_clients` pool **directly by `target.name`** (e.g. its
  `aclose()` / client-invalidation path) — #1672 added a `_session_names` (tenant-key → name)
  map for exactly this reason; that pattern generalizes.
- Shared helper `target_cache_key(target)` → `(str(tenant_id), str(id))` is on main.

## Desired state

`_clients` and `ssh._connections` key on `target_cache_key(target)`. Any direct by-name pool
access (vmware_rest, NSX cookie-jar invalidation) is migrated to the tenant-unique key (reuse
/ generalize the `_session_names` reverse-map pattern). Same-tenant pooling behaviour
unchanged; two same-named targets in different tenants get **distinct** pooled clients.

## Acceptance criteria

[...]

Issue #214, opened 2026-05-11
Goal: Connector parity with ClaudeVCF wrapper set — RDC operators progressively retire local wrappers in favour of meho <connector> <op>


damir-topic:
## Goal

Ship MEHO connectors that cover the operations the RDC team uses most often in `claude-rdc-hetzner-dc`'s `scripts/*.sh` wrappers, in tiers ordered by usage frequency, so the team can progressively retire local wrappers in favour of `meho <connector> <op>` calls — without losing the existing capabilities, audit trail, or operator UX.

## The connector model: dual-shaped by design (clarifying note added 2026-05-20)

**MEHO connectors are dual-shaped.** Both layers ship through the same operation registry, the same dispatcher, the same audit-tree, and the same agent surface — adopters who treat MEHO as "curated composites only" miss half the architecture and the half that gives them long-tail vendor-API coverage on day 1.

**Layer 1 — Hand-coded composites.** First-class, curated, operator-oriented multi-step governed workflows. Hand-authored Python handlers that orchestrate multiple sub-ops via the dispatcher's recursive `dispatch(...)` call. Each composite carries per-op `safety_level` + `requires_approval` annotations and an LLM-friendly `description`. The v0.3.0 vmware connector ships 13 composite ops (`vmware.composite.host.evacuate`, `vmware.composite.vm.snapshot.revert`, `vmware.composite.cluster.patch`, …) that wrap multi-step flows the raw vendor API can't express in one call. See [`backend/src/meho_backplane/connectors/vmware_rest/composites/_register.py`](backend/src/meho_backplane/connectors/vmware_rest/composites/_register.py) — the `_COMPOSITES` tuple holds all 13 `_CompositeSpec` entries (5 read + 8 write). This is the agent's first-class surface and where governance lives.

[...]

PR #1791, opened 2026-06-15
feat(connectors): wire verify_tls into HttpConnector._http_client (#1781)

### LINKED ISSUE #1781: Task: wire verify_tls into HttpConnector._http_client — insecure SSLContext + cache-key dimension

damir-topic:
## Summary

- Wire the per-target **`verify_tls`** flag (T1 #1780) into `HttpConnector._http_client` so a `verify_tls=false` target can reach a self-signed / internal-CA appliance (Initiative #1774, Goal #214 connector-parity).
- `verify_tls=false` builds the pooled client with a **module-cached insecure `ssl.SSLContext`** (`check_hostname` disabled **before** `CERT_NONE` — reusing the in-tree `_vcsim.py` idiom; reverse order raises `ValueError`) and emits a WARN at construction so the insecure dispatch is loud, not silent.
- `verify_tls=true` (the default) passes **no `verify=` argument at all**, so the global `SSL_CERT_FILE` / chart trust-bundle path (#209) is byte-identical to before — verification stays on. (httpx 0.28.1: `verify=<ssl.SSLContext>` is the forward-safe shape.)
- The pooled-client key gains a `verify_tls` dimension via a new `extra_cache_dimensions` hook → `(tenant_id, id, verify_tls)`. Pure append: the `(tenant_id, id)` cross-tenant isolation prefix (#1682/#1642) is unchanged, and a PATCH that flips the flag is not served the stale client. The two subclass reach-ins into the inherited `_clients` pool (NSX cookie-clear on session re-establish; vSphere session-revoke on close) are updated to build the same full key.

## Test plan

- [x] `ruff format --check src/ tests/` — clean (984 files)
- [x] `mypy src/` (strict) — clean (480 source files)
- [x] `pytest -n 6 --dist loadscope --ignore=tests/integration` — 7625 passed, 195 skipped (sandbox secret guards)
- [x] code-quality `--diff` — 0 blockers, 7 pre-existing warnings
- [x] **AC: `verify_tls=true` builds client with NO `verify=` arg** — asserted via a patched `httpx.AsyncClient` (kwarg-absence) *and* the live transport `SSLContext` (`check_hostname` True, `verify_mode == CERT_REQUIRED`) — byte-identical to today.
- [x] **AC: `verify_tls=false` client uses the insecure context** — live transport `SSLContext` has `check_hostname is False` and `verify_mode == ssl.CERT_NONE`, and is the module-cached instance; a dedicated test asserts the field-ordering.
- [x] **AC: flipping `verify_tls` → distinct key + freshly built client** — stale client not reused; `(tenant_id, id)` prefix shared.
- [x] **AC: same-id different-tenant → distinct clients** — the appended dimension never collapses the tenant prefix.
- [x] **AC: `Harbor`/`Gcloud` inherit via `super()._http_client`** — no extra change; NSX + vSphere session-cleanup tests (`test_aclose_revokes_every_cached_session`, `test_downstream_401_triggers_relogin_and_retry_once`) pass with the new key shape.
- [x] CLI OpenAPI snapshot + generated client byte-unchanged (`git diff --exit-code -- cli/api/openapi.json cli/internal/api/client.gen.go` after `make snapshot-openapi && make generate`) — no route/schema change.

## Out of scope

- The two out-of-pool dispatch clients named out-of-scope in #1774 — the k8s reachability probe (`kubernetes/connector.py`, already hardcodes `verify=False`) and the GitHub App token-exchange (`github/session.py`, public-cert `api.github.com`) — do not honor `verify_tls` and are unchanged.
- CA-pin (T5 #1784) — the secure supersession follow-up; the friendlier error message (T3 #1782, already merged).
- Adjacent findings (recorded, not edited): pre-existing code-quality WARNs on `nsx/connector.py` (529 lines) and `vmware_rest/connector.py` (671 lines, file-size) and a few >60-line functions in those files; `_request_json`'s PLR0913 (6 args) — all pre-existing, untouched by this change.

Closes #1781

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,19 @@ connector-related release-notes line.
   (`tls_verification_disabled` + before/after) and a WARN log, closing
   the prior gap where a target PATCH wrote an empty audit payload
   (#1780).
+- Connector dispatch now **honors the per-target `verify_tls` flag**. A
+  target with `verify_tls=false` reaches a self-signed / internal-CA
+  appliance over an insecure TLS channel (a module-cached `SSLContext`
+  with `check_hostname` off + `CERT_NONE`), emitting a WARN at client
+  construction; a `verify_tls=true` target (the default) is built with
+  **no** `verify=` argument, so the global `SSL_CERT_FILE` / chart
+  trust-bundle path stays byte-identical to before. The pooled-client
+  key gains a `verify_tls` dimension — `(tenant_id, id, verify_tls)` — so
+  a PATCH that flips the flag is not served the stale client, while the
+  `(tenant_id, id)` cross-tenant isolation prefix is unchanged. `Harbor`
+  and `Gcloud` inherit the behaviour through `HttpConnector`; the
+  out-of-pool k8s reachability probe and GitHub App token-exchange are
+  unaffected (#1781).
 - Bulk read-class connector enable path across REST + MCP + CLI:
   `POST /api/v1/connectors/{id}/enable-reads`, the
   `meho.connector.enable_reads` MCP tool, and `meho connector
--- a/backend/src/meho_backplane/connectors/adapters/http.py
+++ b/backend/src/meho_backplane/connectors/adapters/http.py
@@ -24,27 +24,76 @@ client is host-bound via ``base_url``, the collision would route one
 tenant's request to the other tenant's host and leak credentials across
 the tenant boundary (evoila/meho#1682).
 
-**Cert-bundle support:** httpx honours ``SSL_CERT_FILE`` natively; no extra
-cert logic is needed here beyond not overriding the default ``verify`` flag.
+**Cert-bundle support:** httpx honours ``SSL_CERT_FILE`` natively. When a
+target verifies TLS (the default, ``verify_tls=True``) the client is built
+with **no** ``verify=`` argument, so the global ``SSL_CERT_FILE`` /
+chart-trust-bundle path (evoila/meho#209) is in effect unchanged. When a
+target opts out (``verify_tls=False``) the client is built with an explicit
+insecure :class:`ssl.SSLContext` (``check_hostname`` off, ``CERT_NONE``) so
+it can reach a self-signed / internal-CA appliance (evoila/meho#1774). The
+opt-out is per-target, never global, and loud — see
+:func:`_insecure_ssl_context` and the WARN emitted at client construction.
+
+**Client pool key:** the pool is keyed on
+:func:`~meho_backplane.connectors._shared.cache_key.target_cache_key`
+(``(tenant_id, id)``) **plus** :meth:`HttpConnector.extra_cache_dimensions`
+(``(verify_tls,)``), i.e. ``(tenant_id, id, verify_tls)``. Appending
+``verify_tls`` keeps the ``(tenant_id, id)`` tenant-isolation prefix intact
+(evoila/meho#1682/#1642) while ensuring a PATCH that flips ``verify_tls`` is
+not served the stale pooled client built under the previous flag.
 """
 
 from __future__ import annotations
 
 import asyncio
+import ssl
 from typing import Any
 
 import httpx
+import structlog
 from tenacity import retry, retry_if_exception, stop_after_attempt, wait_exponential
 
 from meho_backplane.auth.operator import Operator
 from meho_backplane.connectors._shared.cache_key import target_cache_key
 from meho_backplane.connectors.base import Connector
 
+logger = structlog.get_logger(__name__)
+
 # Forward declaration — replaced with `from meho_backplane.targets import Target`
 # once G0.3 lands the Target model.
 type Target = Any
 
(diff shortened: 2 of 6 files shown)


damir-topic:
@coderabbitai review

PR #1792, opened 2026-06-15
feat(connectors): per-target CA-pin (tls_ca_pin) — secure TLS-trust supersession (#1784)

### LINKED ISSUE #1784: Task: per-target CA-pin via load_verify_locations(cadata) — secure TLS-trust supersession

damir-topic:
## Summary

- **T5 of Initiative #1774** (per-target TLS trust): adds a per-target **CA-pin** — the *secure* supersession of `verify_tls=false`. An operator pins an appliance's CA/cert PEM via the new first-class `tls_ca_pin` field, and connector dispatch trusts that CA **while keeping `CERT_REQUIRED` + `check_hostname` ON** (`ssl.create_default_context()` + `load_verify_locations(cadata=<pem>)`, the govc-thumbprint pattern). A self-signed / internal-CA endpoint becomes reachable **without** weakening verification or touching the global bundle.
- **Precedence + mutual exclusion:** CA-pin (secure) > `verify_tls=false` (insecure) > default (no `verify=` arg, byte-identical to today). Setting both `tls_ca_pin` and `verify_tls=false` is rejected with a `422` (schema-level for one body, route-level for the merged post-PATCH state), so a pin is never silently undone by a stray insecure flag.
- **Cache safety:** the client-pool key gains a CA-pin digest dimension — `(tenant_id, id, verify_tls, ca_pin_digest)` — so rotating a pin builds a fresh client, while the `(tenant_id, id)` cross-tenant isolation prefix (#1682/#1642) is intact. Integrates with T2's `verify_tls` dimension (append-only); vmware_rest / nsx subclasses stay correct (prefix-match / shared `_client_cache_key`).
- **Never silent:** the PEM is validated at the API boundary (malformed → `422`, not an opaque dispatch failure); set/rotate/clear writes a durable `audit_log` row (`tls_ca_pinned` + a digest of the before/after PEM — never the PEM body) + a WARN. The `connector_tls_verify_failed` error (T3) and the operator docs (T4) now name CA-pin as the preferred per-target fix, closing the loop.

## What changed

- **DB:** new nullable `targets.tls_ca_pin` Text column; **migration `0045`** (additive + reversible, chains off `0044`; new head `0045`).
- **Schemas:** `tls_ca_pin` threaded through `Target` / `TargetCreate` / `TargetUpdate` / `TargetSummary` + projections, with PEM validation + mutual-exclusion validators.
- **Connector:** `HttpConnector._http_client` precedence + `_build_ca_pinned_ssl_context` + `extra_cache_dimensions` digest.
- **CLI:** `cli/api/openapi.json` + `cli/internal/api/client.gen.go` regenerated (`make snapshot-openapi && make generate`); `git diff --exit-code` clean, CLI builds.
- **Docs:** `deploy/values-examples/README.md` CA-pin section flipped from *planned* → *shipped* with precedence + worked examples; `docs/architecture/connectors.md` + dispatcher docstring updated; CHANGELOG `[Unreleased]` bullet.

## Test plan

- [x] `ruff check` — clean
- [x] `uv run mypy src/` (strict) — Success, no issues in 480 files
- [x] **CA-pin context trusts the pinned CA with `check_hostname=True` / `CERT_REQUIRED`** (real in-memory TLS handshake) — secure, NOT weakened
- [x] **Fails closed** against a cert that does not chain to the pin, and against a hostname mismatch (chain + hostname still enforced)
- [x] **Precedence** CA-pin > `verify_tls=false` > default — tested (a target carrying both builds the verifying context, never the insecure one)
- [x] **Cache key** distinguishes pinned-CA variants (rotation → fresh client) + preserves tenant isolation (same id + same pin, different tenant → distinct host-bound clients)
- [x] **Migration `0045`** up → backfills NULL on a pre-existing row → down drops cleanly → re-up idempotent (single head)
- [x] Schema PEM validation (malformed → 422) + mutual-exclusion (both → 422); audit binding on create/set/rotate/clear (digest only, no PEM body)
- [x] `uv run pytest` — 110 passed in the touched files; 3780 passed / 255 skipped across the full connector/target/dispatch/audit/schema surface (no regressions)

## Out of scope

- **SHA-256 server-cert fingerprint pinning** — deferred hard per #1784 (stdlib `ssl` has no native fingerprint pinning; needs a custom httpx transport). Separate future task.
- The two out-of-pool dispatch clients (k8s reachability probe, GitHub App token-exchange) do not honor per-target TLS trust — moot for the self-signed case (named in #1774).

Closes #1784


damir-topic:
@coderabbitai review

Issue #1816, opened 2026-06-17
Task: promote the product↔impl_id round-trip check from WARN to hard-fail in register_connector_v2


damir-topic:
Parent goal: #221
Parent initiative: #1810
Depends-on: #1814

## Why now

#1798 added `_warn_if_product_impl_id_diverges` (`connectors/registry.py:183-253`) as an **advisory** check — it could not raise because the 5 `_PRODUCT_SPLITS` connectors would crash `_eager_import_connectors` at boot. Once T1 (#1814) realigns them, nothing violates the round-trip, so the check can become a **hard-fail** — turning a future product↔impl_id mismatch into an immediate, loud deploy-time error instead of a silent dogfood-only namespace shadow.

## Current state (grounded, HEAD)

`_warn_if_product_impl_id_diverges` logs a WARN and returns (never raises) when `parse_connector_id(f"{impl_id}-{version}")[0] != product` (`registry.py:232-253`); called from `register_connector_v2` (`registry.py:172`). The docstring (`:204-213`) explicitly defers the hard-fail to Initiative #1810. The test fixture `_VCF_PRODUCT_SPLITS` (`backend/tests/test_operations_register_ingested.py:1153-1159`) enumerates the 5 connectors that currently warn.

## Desired state

After T1, `register_connector_v2` **raises** on a product↔impl_id divergence (rename `_warn_if_product_impl_id_diverges` → `_assert_product_impl_id_round_trips` or similar), crashing boot with the existing actionable message ("Align product to …"). No allowlist — any violation fails. The `_VCF_PRODUCT_SPLITS` test fixture is removed/emptied (no sanctioned splits remain).

## Acceptance criteria

- [ ] `register_connector_v2` raises (e.g. `ValueError`/`RuntimeError`) on a `(product, version, impl_id)` whose `product` ≠ the `parse_connector_id`-derived product; the wildcard `("", "")` row remains exempt (parser-incompatible by design, per the current docstring).
- [ ] A unit test registers a deliberately-divergent triple and asserts it raises with the actionable message; a second test asserts every real connector boots clean via `_eager_import_connectors`.
- [ ] The `_VCF_PRODUCT_SPLITS` fixture and any "sanctioned split" carve-out are removed; the docstring no longer points at #1810 as pending.
- [ ] Boot (`_eager_import_connectors`) succeeds with all connectors registered.
[...]

Issue #1814, opened 2026-06-17
Task: realign the 5 split connectors' registry product to the short canonical token (sddc/vcfa/fleet/vrops/hetzner)


damir-topic:
Parent goal: #221
Parent initiative: #1810

## Why now

Five connectors register a registry `product` that does **not** round-trip through `parse_connector_id` (the dispatch/discovery surface derives the product from the first hyphen-segment of `impl_id`). This long↔short divergence is the class that shadowed `VcfLogsConnector` in the v0.16.0 dogfood (#1798). vRLI was realigned in #1798; this Task realigns the remaining family so the registration round-trip holds for every connector — the precondition for promoting the check to hard-fail (T2) and retiring the bridges (T3).

## Current state (grounded, post-#1798 / HEAD)

`_warn_if_product_impl_id_diverges` (`connectors/registry.py:183-253`) WARNs on every boot for these 5 because their registered `product` ≠ the parser-derived spelling:

| connector | registry `product` (long) | impl_id | parser-derived (short) | class attr | registration |
|---|---|---|---|---|---|
| sddc-manager | `sddc-manager` | `sddc-rest` | `sddc` | `sddc_manager/connector.py:141` | `sddc_manager/__init__.py:55,67` |
| vcf-automation | `vcf-automation` | `vcfa-rest` | `vcfa` | `vcf_automation/connector.py:155` | `vcf_automation/__init__.py:61,73` |
| vcf-fleet | `vcf-fleet` | `fleet-rest` | `fleet` | `vcf_fleet/connector.py:185` | `vcf_fleet/__init__.py:52,64` |
| vcf-operations | `vcf-operations` | `vrops-rest` | `vrops` | `vcf_operations/connector.py:154` | `vcf_operations/__init__.py:65,77` |
| hetzner-robot | `hetzner-robot` | `hetzner-rest` | `hetzner` | `hetzner_robot/connector.py:148` | `hetzner_robot/__init__.py:46,58` |

[...]

Issue #1810, opened 2026-06-16
Initiative: retire the VCF _PRODUCT_SPLITS long↔short divergence — realign the connector family to one canonical product↔impl_id + hard-fail registration invariant


damir-topic:
Parent goal: #221

## Why now

Filed as the deferred-scope follow-up to **#1798** (G0.26 / Initiative #1800). #1798 fixed the **live vRLI SEV-2** (a `product="vrli"` target resolving the auto-shim instead of `VcfLogsConnector`) by aligning vRLI's canonical identity and adding a registration round-trip **warn** — but it deliberately **scoped out** the rest of the family, because the audit there found the divergence is not a one-off bug: it's a **deliberate, shipped, multi-initiative design** across the whole VCF suite.

The round-trip check `parse_connector_id(f"{impl_id}-{version}")[0] == product` fails for **6 connectors** (5 after #1798 aligns vRLI): the `_PRODUCT_SPLITS` family carries a *long* registry `product` and a *short* parser-/dispatch-derived product, bridged by three sanctioned band-aids. This Initiative retires the split so the registration invariant can become **hard-fail** (early-warning for real violations) without per-spelling allowlists.

## Current state (grounded on v0.16.0 / post-#1798)

| registry product | impl_id | parser-derived | bridged by |
|---|---|---|---|
| `sddc-manager` | `sddc-rest` | `sddc` | `dispatch_product`, mig 0038, `PRODUCT_ALIASES["sddc"]`, `canonical_product_token` |
| `vcf-automation` | `vcfa-rest` | `vcfa` | `dispatch_product` / `_reconciled_row_product`, mig 0038 |
| `vcf-fleet` | `fleet-rest` | `fleet` | same |
| `vcf-operations` | `vrops-rest` | `vrops` | same |
| `hetzner-robot` | `hetzner-rest` | `hetzner` | same |
| (`vcf-logs`/`vrli-rest` — aligned in #1798) | | | — |

The bridges: `operations/_lookup.dispatch_product` + `register_ingested._reconciled_row_product` (#1647/#1677) rewrite ingested rows long→short for dispatch; migration `0038_backfill_endpoint_descriptor_product_splits`; `PRODUCT_ALIASES`/`canonical_product_token` (#1355) accept the short token at `POST /api/v1/targets`.

## Scope — realign the family to one canonical identity (proposed Tasks)

[...]

PR #1851, opened 2026-06-18
feat(connectors)!: reject divergent ingest product at the boundary; retire long↔short bridges (#1817)

### LINKED ISSUE #1817: Task: retire the long↔short product bridges (dispatch_product / _reconciled_row_product / PRODUCT_ALIASES)
Parent goal: #221
Parent initiative: #1810
Depends-on: #1814
Depends-on: #1815

## Why now

The long↔short divergence is bridged by three sanctioned band-aids. Once T1 (#1814) realigns registration and T4 (#1815) migrates existing data, all three become dead code (pure no-ops) and should be retired so the codebase carries one product identity — the end state Initiative #1810 exists to reach, and the anti-pattern (per-spelling reconciliation) that #1798's audit flagged.

## Current state (grounded, HEAD)

- `dispatch_product` (`operations/_lookup.py:94-129`) returns `parse_connector_id(f"{impl_id}-{version}")[0]` — a pure derived-product passthrough (no hardcoded table). Callers: `operations/ingest/register_ingested.py:401` and `operations/ingest/pipeline.py:730`.
- `_reconciled_row_product` (`operations/ingest/register_ingested.py:383-410`) wraps `dispatch_product` + a reconciliation log; rewrites an ingested row's product long→short before persist. After realignment `row_product == product` always → no-op.
- `PRODUCT_ALIASES = {"sddc": "sddc-manager"}` (`connectors/registry.py:80`) + `canonical_product_token` (`registry.py:371-395`) bridge the sddc short→long token at `POST/PATCH /api/v1/targets` (`api/v1/targets.py:364,787`). After sddc-manager→sddc realignment the alias maps a token to itself → dead.

## Desired state

`dispatch_product` and `_reconciled_row_product` and their callers are removed (ingest persists the supplied short product directly); the `PRODUCT_ALIASES["sddc"]` entry is removed so `PRODUCT_ALIASES` is empty and `canonical_product_token` is a pure passthrough — either deleted (and its 2 call sites simplified) or kept as a documented empty no-op for forward-extensibility. No long↔short reconciliation remains anywhere.

## Acceptance criteria

- [ ] `dispatch_product` and `_reconciled_row_product` removed; their callers (`register_ingested.py:401`, `pipeline.py:730`) persist the supplied product directly; grep for both symbols returns nothing.
- [ ] Ingest regression: ingesting a `vcfa`/`sddc`/`fleet`/`vrops`/`hetzner` spec persists short-product `endpoint_descriptor`/`operation_group` rows that dispatch (catalog reports the ops, not "registered, 0 ops").
- [ ] `PRODUCT_ALIASES` is empty (or the dict + `canonical_product_token` removed and `targets.py:364,787` simplified); a `POST /api/v1/targets` with `product="sddc"` validates directly.
- [ ] No remaining long↔short mapping for the 5 (grep the long tokens across `backend/src` excluding the connector packages + migration history returns only intentional historical/migration references).
- [ ] `cd backend && uv run pytest backend/tests/test_operations_register_ingested.py backend/tests/test_targets* backend/tests/test_operations_ingest_*` green; `cd cli && make snapshot-openapi && make generate` if any schema changed.

## Out of scope

- Migration 0038 itself — stays in the chain for historical replays (do not delete a shipped migration).
- The realignment (T1 #1814) and data migration (T4 #1815) — strict prerequisites; removing the bridges before data is migrated would orphan pre-migration long-token rows.

## References

- `operations/_lookup.py:94-129` (`dispatch_product`), `operations/ingest/register_ingested.py:383-410` (`_reconciled_row_product`), `operations/ingest/pipeline.py:730`, `connectors/registry.py:80,371-395` (`PRODUCT_ALIASES`/`canonical_product_token`), `api/v1/targets.py:364,787`.
- Bridge origins: `dispatch_product` #1647/#1677, `PRODUCT_ALIASES`/`canonical_product_token` #1355. Parent: #1810. Depends on T1 #1814 + T4 #1815.

damir-topic:
## Summary

Retires the long↔short product bridges (Initiative #1810, Task #1817) — **re-scoped to "Option B"** per a maintainer decision after a prior worker correctly flagged that the task's original AC1 would regress shipped behavior.

The bridges (`dispatch_product` / `_reconciled_row_product`, and `PRODUCT_ALIASES` / `canonical_product_token`) were **not** pure no-ops post-#1814/#1816: they were still load-bearing for the **generic divergent-ingest path** (`POST /api/v1/connectors/ingest` did not validate the supplied `product` against the connector_id round-trip, so an operator could ingest `--product vcf-automation --impl-id vcfa-rest` and the rows got silently reconciled to `vcfa`). Option B closes that gap at the source, then retires the now-dead bridges.

- **New 422 at the ingest boundary.** `POST /api/v1/connectors/ingest` now rejects a supplied `product` that does not round-trip its connector_id with `422 product_impl_id_mismatch`, before any spec fetch or DB write. Reuses a non-raising predicate (`product_impl_id_round_trips`) factored out of #1816's registration hard-fail, so the parse rule is shared: no-op when `version`/`impl_id` is empty or the parse is lossy; fires only on a lossless parse whose derived product disagrees. The `register_connector_v2` hard-fail (#1816) stays as the backstop.
- **Bridges removed.** With divergent ingests rejected upstream, the supplied product always round-trips, so `dispatch_product` + `_reconciled_row_product` (ingest persists the supplied product verbatim across register / grouping / auto-shim) and `PRODUCT_ALIASES` + `canonical_product_token` (`POST`/`PATCH /api/v1/targets` validate-and-store the supplied token) are deleted, including their `__all__` entries.
- **Docs + CHANGELOG** updated to the single-product-identity end state.

## New 422 contract (operator-facing, breaking)

An ingest of a divergent/long `product` — e.g. `--product drift-test --impl-id drift-impl` (derives `drift`), or any long VCF-family token like `--product vcf-automation --impl-id vcfa-rest` (derives `vcfa`) — **previously returned `200`** and silently reconciled the persisted rows to the derived token. It **now returns `422 product_impl_id_mismatch`**, naming the derived product to use (or rename `impl_id`). Operators ingesting under the short, dispatch-canonical token (what `meho connector list` emits and the realigned family registers under, post-#1814) are unaffected; any tooling that hard-coded a long/divergent `--product` must switch.

## Test plan

- [x] `cd backend && uv run ruff check src/ tests/ && uv run ruff format --check src/ tests/` — clean (CI scope; `dev/seed_demo.py` ruff failures are pre-existing on `main`, outside CI's `src/ tests/` scope).
- [x] `cd backend && uv run mypy src` — clean (480 files).
- [x] Targeted: `pytest tests/test_operations_register_ingested.py tests/test_api_v1_connectors*.py tests/test_targets*.py tests/test_api_v1_targets*.py tests/test_connectors_registry_v2.py tests/test_operations_ingest_*.py -p no:randomly` — **610 passed, 2 skipped** (env-gated).
- [x] Broader xdist: `pytest tests/test_connectors_*.py tests/test_operations_*.py tests/test_api_v1_*.py -n 6 --dist loadscope` — **3601 passed, 4 skipped**.
- [x] `python3 .claude/hooks/code-quality.py --diff` — **0 blockers** (the one over-budget function, `ensure_connector_class_registered`, was a pre-existing 129-line blocker reduced to ~104; carries a `code-quality-allow` with rationale).
- [x] `cd cli && make snapshot-openapi` — **no diff** (the 422 is FastAPI's implicit validation-error schema; the route's declared `responses` map is unchanged, so the OpenAPI doc + CLI client are unaffected).
- [x] Pinned tests audited per supplied-vs-derived product: aligned ingests assert the supplied product persists; divergent service-layer calls assert the hard-fail backstop; the new boundary 422 is pinned in `test_api_v1_connectors_ingest.py::test_ingest_divergent_product_rejected_with_422`; synthetic non-round-tripping ingest fixtures (`vmware-t9`, `tenant-iso`, etc.) realigned to round-tripping triples.

## Out of scope

- Migration `0038`/`0047` and shipped migrations stay (historical replays).
- Broader architecture/cross-repo docs (`docs/architecture/connectors.md`, `docs/cross-repo/*onboarding.md`) mention the long connector tokens as class/name references; they don't reference the deleted symbols and aren't invalidated by this change — flagged for a possible docs follow-up.

Closes #1817

## Summary by CodeRabbit

* **New Features**
  * Product round-trip validation now enforced at all ingest endpoints (REST, MCP, CLI) before database writes or spec fetches.

* **Breaking Changes**
  * Ingest endpoints reject supplied products that don't match the canonical product derived from the connector's implementation ID. REST returns HTTP 422; MCP returns JSON-RPC error; CLI fails early.
  * Product aliasing and canonicalization logic removed. Operators must supply the canonical product token from `meho connector list`.

* **Documentation**
  * Updated guides to reflect canonical product handling and removal of alias reconciliation.


damir-topic:
@coderabbitai review

backend/src/meho_backplane/operations/ingest/connector_registration.py
```diff
@@ -362,66 +366,41 @@ def ensure_connector_class_registered(
...
     from meho_backplane.connectors.registry import all_connectors_v2
-    from meho_backplane.operations._lookup import dispatch_product
 
-    # The shim must live under the dispatch-canonical product the ingested
-    # rows persist under (and that register_connector_v2's round-trip check
-    # now hard-fails on if violated) — not the raw operator-supplied one.
-    canonical_product = dispatch_product(product=product, version=version, impl_id=impl_id)
+    # ``canonical_product`` is a local alias of ``product`` (the boundary
+    # guard guarantees they are equal) kept only for the log fields below.
+    canonical_product = product
```
damir-topic, line 407: **[🛑 B1]** — Divergent product on the MCP ingest path is persisted unreconciled — re-introduces the silent non-dispatchable-shadow bug #1810 exists to kill.

**What:** This line (`canonical_product = product`) and its comment assert "the boundary guard guarantees they are equal." That holds for `POST /api/v1/connectors/ingest`, where `_assert_product_round_trips` runs. It does **not** hold for the MCP tool `meho.connector.ingest`, which builds an `IngestRequest` from arbitrary `product`/`version`/`impl_id` and calls `IngestionPipelineService.ingest(...)` directly (`mcp/tools/connector_ingest.py` → `_run_inline_ingest` / `_spawn_async_ingest`) — it never traverses `_assert_product_round_trips`.

For most divergent products the #1816 registration hard-fail is an adequate backstop (it raises in `register_connector_v2` before any row persists). But when a hand-coded class already serves the `impl_id` under a different product, the deferral branch a few lines below (`if handrolled is not None and handrolled.product != canonical_product: ... return False`) returns **without** calling `register_connector_v2`, so the hard-fail never fires. `check_version_covered_by_registered_class` also passes the case through (it filters candidates by `(product, impl_id)`, so the real class registered under the canonical product is not a candidate). The divergent product then flows to `_run_upsert_loop` and persists.

**Why it matters:** With `_reconciled_row_product` deleted, rows persist under the divergent product (e.g. `vcf-logs`) while every dispatch/discovery probe keys on the parser-derived token (`vrli`). The rows are invisible to dispatch → the connector reports `registered, 0 ops` → the exact silent product-namespace shadow (claude-rdc-hetzner-dc#1136 / the v0.16.0 SEV-2) that Initiative #1810 exists to eliminate. Confirmed empirically: driving `register_ingested_operations(product="vcf-logs", version="9.0", impl_id="vrli-rest")` (with `VcfLogsConnector` registered under `vrli`, the real post-#1814 state) through the service layer inserts 1 row with `connector_registered=False` and **persists it under `product="vcf-logs"`** — no exception.

**Required change:** Enforce the round-trip at a layer every caller traverses. Concretely, call `product_impl_id_round_trips(...)` (raising on `False`) inside `IngestionPipelineService.ingest` (or `_dispatch_real_run` / `_run_register_phase`), **before** `_run_upsert_loop`, so REST, MCP, CLI, and any future caller fail closed identically. `_assert_product_round_trips` at the REST boundary can stay as an early/nicer-envelope 422. Then this `canonical_product = product` alias and the deferral branch can no longer see a divergent product. Add a regression test that drives the divergent + hand-coded-`impl_id` triple (`product="vcf-logs"`, `impl_id="vrli-rest"`) through `service.ingest` / `register_ingested_operations` and asserts it raises and persists nothing — the current `test_divergent_product_ingest_trips_registration_hard_fail` only covers the *no-hand-coded-class* case (`drift-impl`), which is precisely the case where the backstop already works.

**Source:** repro against this PR's head; `register_ingested.py:307-320` (preflight+register precede the upsert), `connector_registration.py:417-436` (deferral returns False without `register_connector_v2`), `connector_registration.py:619+` / `_find_class_covering_version` (pre-flight filters by `(product, impl_id)`), `mcp/tools/connector_ingest.py:181,237,306` (service-layer call, no guard).

damir-topic (commented):




**Anchor task:** #1817 — retire long↔short product bridges; reject divergent ingest product at the boundary
**Surface area:** backend
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `b0bf823f`
**AI reviewer (CodeRabbit):** not received within ~2min poll (trigger present; proceeded per single-turn rule)

The core idea is right and the Option-B re-scope is the correct call: close the divergent-ingest gap at the source, then retire the now-dead reconciliation helpers. The predicate extraction is faithful, the docs rewrite is thorough, and the REST/CLI path is correctly guarded. **But the bridge removal is not safe yet: the 422 guard does not cover every path that reaches the rewired callers, so deleting `_reconciled_row_product` re-introduces the exact silent non-dispatchable-shadow bug Initiative #1810 exists to kill — on the MCP ingest entry point.**

### Acceptance criteria
- [x] New `422 product_impl_id_mismatch` at `POST /api/v1/connectors/ingest` for a non-round-tripping product — verified (`test_ingest_divergent_product_rejected_with_422` passes; correct status, no sensitive leak, message names the fix).
- [x] `dispatch_product` / `_reconciled_row_product` / `PRODUCT_ALIASES` / `canonical_product_token` deleted; 3 callers rewired; `targets.py` validate-as-is preserved — verified.
- [x] `product_impl_id_round_trips` preserves #1816's exact semantics (empty/lossy → True, lossless+disagree → False) — verified; the wildcard / single-segment / vRLI / five-realigned-splits registration tests still pass.
- [❌] **The divergent ingest can no longer reach the rewired callers unguarded** — NOT met. The MCP tool `meho.connector.ingest` calls `IngestionPipelineService.ingest` → `register_ingested_operations` directly, bypassing `_assert_product_round_trips`. See B1.

### Project-board alignment
|---|---|---|---|---|---|
| #1817 | In Review | Closes | manual | ⚠️ partial | Do not close on merge until B1 closes the MCP path |

### Predicted vs. actual diff
- ✅ `registry.py`, `_lookup.py`, `targets.py`, `pipeline.py`, `register_ingested.py`, `list_connectors.py`, the 4 test files, both docs, CHANGELOG — expected and correct.
- ⚠️ `connectors_ingest.py`, `connector_registration.py` — expected, but the guard placement leaves the MCP path uncovered (B1) and a now-false log message (B2).
- ❓ `test_api_v1_connectors_ingest.py` — unexpected-but-justified (new boundary 422 test + fixture realignments).
- 🕳️ **Missing:** `mcp/tools/connector_ingest.py` — the second ingest entry point was not given the guard, and no test covers a divergent product through it.

### Integration verification
- [x] `ruff check` / `ruff format --check` — clean on touched src + tests
- [⚠️ not run] `mypy` — reasoned-only (PR test plan reports clean; not re-run)
- [x] `pytest` — 78 targeted (registry_v2/registry/register_ingested/boundary-422) + 21 MCP ingest + 3 #1816 round-trip, all green
- [x] **Focused repro (then removed):** drove `register_ingested_operations(product="vcf-logs", version="9.0", impl_id="vrli-rest")` (with `VcfLogsConnector` registered under `vrli`, the real post-#1814 state) through the service layer — the path the MCP tool uses. Result: `inserted=1`, `connector_registered=False`, **row persisted under `product='vcf-logs'`**. No hard-fail fired; the row is invisible to every dispatch probe (which keys on the parser-derived `vrli`).

- 🛑 **B1** — divergent product on the MCP ingest path is persisted unreconciled (silent non-dispatchable shadow); the 422 guard only covers the REST route. Required change: enforce the round-trip in `IngestionPipelineService.ingest` (raising before `_run_upsert_loop`) so every entry point fails closed, and add a regression test for the divergent + hand-coded-impl_id triple via the service layer.
- 🛑 **B2** — the `connector_auto_register_deferred_to_handrolled` log message still says "Ingested rows reconcile to the dispatch-canonical product"; reconciliation was deleted, so this is false and documents a removed code path. Required change: drop the "reconcile" claim (once B1 makes a divergent product unreachable here, the branch only handles aligned ingests).

### Why B1 is a blocker, not a "major"

`POST /api/v1/connectors/ingest` is **not** the sole entry point to `register_ingested_operations`. The MCP tool `meho.connector.ingest` (`tenant_admin`, registered at lifespan via `eager_import_mcp_modules`, production-live) builds an `IngestRequest` from arbitrary `product`/`version`/`impl_id` strings and calls `service.ingest(...)` directly — it never invokes `_assert_product_round_trips`. The CLI is safe (it hits the REST route); the MCP tool is not.

For most divergent products the registration hard-fail (#1816) is an acceptable backstop: a fresh divergent ingest with no hand-coded class for the `impl_id` reaches `register_connector_v2` and raises **before** any row persists. **But the hand-coded-deferral branch in `ensure_connector_class_registered` returns `False` without ever calling `register_connector_v2`** when a hand-coded class already serves the `impl_id` under a different product (`--product vcf-logs --impl-id vrli-rest`, with `VcfLogsConnector` under `vrli`). The version-coverage pre-flight also passes that case through (it filters by `(product, impl_id)`, so the `vrli`-registered class is not a candidate for `vcf-logs`). So the divergent product flows straight to `_run_upsert_loop` and persists — exactly what `_reconciled_row_product` used to rewrite, and exactly the `registered, 0 ops` SEV-2 #1810 exists to prevent. Confirmed empirically (see Integration verification).

- Adopted: 0 — not received within the bounded poll (rate-limited; trigger comment present).
- Disputed: 0
- Out of scope: 0

### Docs
- [x] `docs/codebase/spec-ingestion.md` + `docs/codebase/api-shape-conventions.md` — updated, accurate, and well-written for the REST/CLI end state. **Caveat:** the new "Product identity at the ingest boundary" section frames the 422 as *the* upstream guard ("a divergent product supplied to the ingest route is rejected up front"); after B1 this should also note (or the guard should actually be) the service layer so it is true for the MCP surface too.

- The predicate extraction (`product_impl_id_round_trips` + the raising `_assert_*` delegating to it) is clean: one parse rule, two enforcement points, semantics provably preserved by the surviving #1816 tests.
- The CHANGELOG BREAKING bullet is precise about the operator-facing 200→422 change and names the fix (`derived_product` / rename `impl_id`).
- The 422 envelope shape (`kind`, `product`, `derived_product`, `message`) is consistent with the route's other structured 422s and leaks nothing sensitive.
- The `code-quality-allow` on `ensure_connector_class_registered` is justified — the PR reduced the function (~129→104 lines) and the marker rationale is honest about it being pre-existing debt; not masking a new smell.

damir-topic (commented):




**Anchor task:** #1817 — retire long↔short product bridges; reject divergent ingest product at the boundary (Option B)
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `1dd0d488`
**AI reviewer (CodeRabbit):** received (reviewed the iter-2 head `0b534f59..1dd0d488`); 3 actionable comments, all Minor / Quick-win, none blocking — see cross-check below.

Both iter-1 blockers are resolved, and the central claim — that `IngestionPipelineService.ingest` is the single chokepoint every ingest entry point traverses — was independently verified by tracing all callers from scratch and proven load-bearing by a guard-disable repro.

### Acceptance criteria
- [x] **B1 (silent data-corruption on the MCP path) — CLOSED.** Enforcement moved out of the REST handler into `IngestionPipelineService.ingest` (`pipeline.py:619`), which raises the new `ProductImplIdMismatch` (`operations/ingest/exceptions.py`) **first** — before `build_connector_id`, before `_validate_spec_versions` (spec fetch), before both dispatch paths (incl. dry-run), so no spec I/O or DB write precedes it on any path. Verified by exhaustive caller trace (below) + an empirical load-bearing proof.
- [x] **B2 (false reconciliation log) — CLOSED.** The `connector_auto_register_deferred_to_handrolled` message no longer claims rows "reconcile to the dispatch-canonical product"; it now states rows persist under the supplied, round-trip-validated product (no reconciliation — divergent products rejected upstream). All adjacent docstrings corrected from "route boundary" to "service-layer chokepoint, every entry point (REST / MCP / CLI)".
- [x] Predicate `product_impl_id_round_trips` faithfully mirrors #1816 (empty/lossy → True; lossless+disagree → False); `_assert_product_impl_id_round_trips` now delegates to it with the unchanged raise; #1816 hard-fail test still green.
- [x] REST 422 envelope **byte-identical** to pre-fix; MCP surfaces JSON-RPC `-32602` via the shared builder; OpenAPI snapshot unaffected (responses map untouched).
- [x] Bridges (`dispatch_product` / `_reconciled_row_product` / `PRODUCT_ALIASES` / `canonical_product_token`) removed from executable code (only explanatory comments remain); `__all__` updated; `targets.py` validate-as-is preserved.

### The decisive question — caller trace (redone from scratch)
Every path that persists `endpoint_descriptor` / `operation_group` rows from an operator-supplied product was walked:

- `register_ingested_operations` → **exactly one** production caller: `pipeline.py:896` inside `_run_register_phase` ← `_dispatch_real_run` ← `ingest()`.
- `_run_upsert_loop` → **one** caller: inside `register_ingested_operations` itself.
- `.ingest()` → **4** production call sites: REST (`connectors_ingest.py:523,954`) and MCP (`connector_ingest.py:241` inline, `:310` async). CLI hits the REST route.
- `typed_register.py` (the only other `EndpointDescriptor` / `OperationGroup` writer) is the **typed-connector init-time** path — product comes from the connector class, never operator-supplied, and it never used the deleted bridges. Correctly out of scope.
- No scheduler / background / seed / catalog / boot path reaches `register_ingested_operations` or `.ingest()` directly.

→ `IngestionPipelineService.ingest` is genuinely the single chokepoint, and the guard fires before persistence on all of REST + MCP (inline & async) + CLI. Even the async MCP path is fail-closed: the guard raises before the job does any write, so a divergent product flips the job to `failed` and never persists a shadow row.

### Load-bearing proof (executed)
Commented out the service guard (`pipeline.py:619`) and re-ran the two new regression tests: **both FAILED** (no raise / would persist). Restored the guard → both pass. The tests genuinely guard the fix, not just an exception type's existence.

### Regression tests audited
- `test_service_ingest_rejects_divergent_product_with_handrolled_impl_no_rows` — drives the **real** `IngestionPipelineService.ingest` (no stub), `VcfLogsConnector` under `vrli` + operator `product=vcf-logs` (the exact iter-1 repro), asserts `ProductImplIdMismatch` (derived `vrli`) **and** `rows == []` (zero persisted).
- `test_inline_divergent_product_with_handrolled_impl_fails_closed` — drives the **real** `_ingest_handler` (no stub), asserts `McpInvalidParamsError` (`-32602`) with `data.kind=product_impl_id_mismatch`, `product=vcf-logs`, `derived_product=vrli`.

### Predicted vs. actual diff
- ✅ All 22 files expected and correct (service guard + new exception + shared envelope + MCP dispatch-table extension + REST mapping relocation + bridge deletions + 2 new tests + doc/CHANGELOG updates).
- 🚫 None unexpected/undisclosed. 🕳️ None missing (the iter-1 hole — `mcp/tools/connector_ingest.py` + the service chokepoint — is now covered).

### Integration verification
- [x] `ruff check` — clean (all changed src + tests)
- [x] `ruff format --check` — 24 files already formatted
- [x] `mypy` — no issues in 7 changed source files
- [x] `pytest` — **226 passed** across register_ingested + ingest_catalog + api_v1_connectors_ingest + mcp_tools_connector_ingest + connectors_registry + connectors_registry_v2; 2 new tests + #1816 hard-fail + REST boundary tests all green
- [x] Caller trace + envelope byte-identity verified by hand against the deleted REST pre-flight
- [⚠️ not run] migrations replay — no migration in this PR
- [⚠️ not run] `docker compose up` — not exercised (no boot-path change)

- Adopted as blockers: 0
- Disputed: 0
- Out of scope / non-blocking: 3 — all `Minor | Quick win`, recorded as adjacent improvements (none touches B1/B2 or correctness).

- CodeRabbit: `test_api_v1_connectors_ingest.py` ~L2546 — pin the fail-before-fetch contract at the REST boundary by asserting the spec URL is never fetched. The contract is already proven by the service-layer test's `rows == []` assertion and the guard-disable repro, so this is hardening.
- CodeRabbit: `test_operations_ingest_catalog.py` ~L327 — strengthen a structural listing-vs-registry assertion to per-connector equality. Unrelated to this fix.
- CodeRabbit: `CHANGELOG.md` L96 — suggests moving the BREAKING entry to `### Breaking changes`. The `[Unreleased]` block has no such subsection and the sibling PR #1814's BREAKING entry uses the identical inline-prefix-under-`### Changed` pattern, so leaving it matches the cycle's established convention (and avoids a CHANGELOG [Unreleased] rebase hotspot). Non-blocking either way.

### Docs
- [x] `docs/codebase/spec-ingestion.md` + `docs/codebase/api-shape-conventions.md` — the iter-1 caveat (framing the 422 as a REST-only guard) is resolved now that enforcement is genuinely at the service layer for all surfaces; docstrings reworded to match. `verified_unchanged` since iter-1's substantive doc rewrite (no further doc drift introduced by the iter-2 delta).

- The relocation is textbook: one predicate (`product_impl_id_round_trips`), two enforcement points (registration hard-fail + ingest service raise), with the REST 422 reduced to a byte-identical exception→HTTP mapping and the MCP `-32602` sharing the same envelope builder so the two surfaces can't drift.
- The two regression tests drive the real code paths and the service test asserts zero rows persisted — exactly the property that distinguishes a real fix from a cosmetic one. The guard-disable repro confirms they'd catch a regression.
- `ProductImplIdMismatch` lands in the existing `SPEC_ERROR_TYPES` tuple + dispatch table with a comment requiring both to be touched together — the same anti-drift discipline the surrounding sibling errors use.

PR #1044, opened 2026-05-24
feat(ui): G10.1-T3 broadcast wall-monitor + Last-24h replay + long-display session refresh

damir-topic:
## Summary

- **Wall-monitor mode (`/ui/broadcast?wall=1`, work item #5)** — a no-chrome standalone `broadcast/wall.html` (no sidebar / top bar / filter bar) that maximises the feed for a full-screen team-room monitor. It embeds the same `_feed.html` fragment with `wall=True`, so it inherits the live SSE wiring, the EventSource auto-reconnect + `Last-Event-Id` replay durability, and the 1000-row cap unchanged — only the chrome + density differ (taller rows, auto-scroll to newest).
- **Long-display session refresh (work item #5)** — a bounded **sliding-session extension** in `session_store.load_session`. The access-token-TTL-bounded BFF session would otherwise lapse mid-display, and a lapsed session 302s the next SSE reconnect to login — a non-200 that *permanently* kills the browser `EventSource` (WHATWG SSE spec). The fix slides `expires_at` forward on an active `/ui/*` load, bounded by an absolute lifetime cap from `created_at` (OWASP ASVS idle-vs-absolute pairing). No Keycloak round-trip (that refresh handler doesn't exist in v0.2); the real `expires_at` gate is what logs `/ui/*` out, and that is what this extends. Two settings (sliding 1h, absolute cap 12h); `sliding=0` disables.
- **Last-24h replay (`/ui/broadcast/history`, work item #6)** — a "Last 24h" tab that HTMX-loads a **finite** `XRANGE` pull of the tenant's last-24h events (bounded by the 24h window + a 1000-row `COUNT` cap, so the endpoint terminates — no streaming loop), seeded into the **same** `broadcastFeed` controller that renders the live feed, so history rows render via `_event_row.html` and open the **same** T2 drawer.
- **Cross-tenant isolation (DoD)** — the history `XRANGE` keys off `meho:feed:{session.tenant_id}` (never a query param), identical to the live stream bridge; the wall view reuses the live fragment so it inherits the same scoping. Asserted for both.

## Test plan

- [x] `ruff check` — clean
- [x] `ruff format --check` — clean
- [x] `mypy src/meho_backplane/` — clean (288 files)
- [x] `pytest -n 4 tests/test_ui_broadcast_wall_replay.py` — 13 passed (~3s; stream/replay tests proven to terminate fast)
[...]

Issue #100, opened 2026-05-09
Task: Write stack-choice ADR (Python backplane + Go CLI) in meho-internal/.planning/decisions/


damir-topic:
Parent initiative: #99

## Target

New ADR file at `evoila-bosnia/meho-internal/.planning/decisions/0004-stack-choice.md` (or next-available number — verify against existing `0001-license-choice.md` + `0003-spdx-header-convention.md`; the `0002` slot is currently a gap, executor decides whether to fill it or skip it).

- File: `evoila-bosnia/meho-internal/.planning/decisions/0004-stack-choice.md`
- No source-code changes in `evoila/meho` for this Task — ADR only

## Context

Goal #98 locks the stack as **Python backplane + Go CLI**. The decision was made on 2026-05-09 because the team has near-zero Go experience (so backplane preserves MEHO.X familiarity in Python), and the upstream requirement letter mandates a single static binary CLI (so Python CLI shapes — PyOxidizer / shiv / runtime-bundled tarballs — are ruled out as operator-UX-degrading).

This Task is the formal ADR artifact + records the framework-level sub-choices that follow from the top-level stack decision.

## Current state

```
$ gh api repos/evoila-bosnia/meho-internal/contents/.planning/decisions/ --jq '.[].name'
0001-license-choice.md
0003-spdx-header-convention.md
```

No stack ADR. The decision is recorded in `evoila/meho#98` body's "Locked prerequisites" section but has no canonical ADR yet.

## Desired state

A new ADR file using the same MADR-style structure as `0001-license-choice.md` and `0003-spdx-header-convention.md` (Date/Decider/Status header + Decision / Rationale / Alternatives / Implications sections), recording:

**Decision**: Python backplane + Go CLI.

**Framework-level locked sub-choices** (each with one-line rationale):

- Backplane web framework: **FastAPI** (async-native, OpenAPI generation built in, Pydantic v2 integration, MEHO.X carry-over)
- Validation: **Pydantic v2**
- DB: **PostgreSQL** via **SQLAlchemy 2.x** with async support (**asyncpg** driver)
- Migrations: **Alembic** (with backward-compat-only discipline enforced via CI guard — separate Task in G2.3)
- Logging: **structlog** (JSON output to stdout, k8s-native log-collection-friendly)
- HTTP client (where backplane needs to make outbound calls): **httpx**
- Vault client: **hvac** (HashiCorp's official Python client)
[...]

PR #1859, opened 2026-06-18
fix(ui): single-quote runbook editor x-data to close attribute-breakout XSS (#100)

damir-topic:
## Summary

- The runbook authoring editor bound operator-authored fields into a **double-quoted** Alpine `x-data="..."` attribute — the step tree via `| safe` (autoescape disabled) and the scalars via `| tojson` (which does not escape `"`). A field containing a `"` terminated the attribute and grafted attacker markup onto the host element (attribute-breakout XSS on the edit page and the create / 422 re-render path).
- This was the lone data-bearing `x-data` island that never received the PR #1044 hardening already applied to the broadcast feed, filter bar, and recent-ops surfaces.
- Fix applies the established convention: single-quote the attribute (`x-data='runbookEditor({...})'`) and render the step tree with `{{ form_steps | tojson }}` over the raw list — `build_editor_context` no longer pre-serialises it with `json.dumps`. `| tojson` escapes `'` to `&#39;` (never a raw single quote) plus `<`/`>`/`&`, so a single-quoted attribute is breakout-proof while the `"` bytes stay inert inside the value. This removes the last `| safe` in an HTML-attribute context under `ui/templates/`.

## Test plan

- [x] `ruff check` (edited Python files) — clean
- [x] `mypy src/.../runbooks/editor.py` — clean
- [x] `pytest tests/test_ui_runbooks_editor.py tests/test_ui_runbooks_lifecycle.py tests/test_ui_runbooks_acceptance.py -q` — 35 passed
- [x] New regression tests (stored GET-edit + reflected POST-new-422 paths) reuse the parser-grounded `_assert_no_xss_breakout` harness; verified they **fail** against the pre-fix double-quoted `| safe` template and **pass** after the fix.
- [x] `pytest tests/test_ui_broadcast_filters.py -q` (harness source) — 22 passed
- [x] No FastAPI route/schema signature change (context-dict key rename only) → no OpenAPI snapshot regen needed.

## Out of scope

- No change to the runbook field schemas (no character allowlist) — output-context escaping is the correct fix; a schema allowlist is a separate defence-in-depth item.
- CSP `frame-ancestors` / `X-Frame-Options` and the `<script type="application/json">` data-island refactor are tracked separately.

Refs evoila-bosnia/meho-internal#100

### DIFF
--- a/backend/src/meho_backplane/ui/routes/runbooks/editor.py
+++ b/backend/src/meho_backplane/ui/routes/runbooks/editor.py
@@ -438,9 +438,13 @@ def build_editor_context(
 ) -> dict[str, object]:
     """Assemble the Jinja context the ``runbooks/editor.html`` template needs.
 
-    ``form_steps`` is JSON-serialised into ``initial_steps_json`` so the
+    ``form_steps`` is passed through as a raw list and the template renders
+    it with ``| tojson`` inside a single-quoted ``x-data`` attribute, so the
     Alpine model can hydrate from the server-rendered prefill (an edit
-    pre-loads the existing steps; a fresh ``new`` seeds one blank step). The
+    pre-loads the existing steps; a fresh ``new`` seeds one blank step)
+    without any field value being able to break out of the attribute. The
+    list is not pre-serialised here: ``json.dumps`` leaves ``"`` raw, which
+    would have terminated the old double-quoted ``| safe`` attribute. The
     post target + submit verb differ by ``mode`` so a single template serves
     both the draft-create and edit-in-place flows.
     """
@@ -450,7 +454,7 @@ def build_editor_context(
         "title": title,
         "description": description,
         "target_kind": target_kind,
-        "initial_steps_json": json.dumps(form_steps),
+        "form_steps": form_steps,
         "submit_url": "/ui/runbooks/new" if mode == "new" else f"/ui/runbooks/{slug}/edit",
         "submit_label": "Create draft" if mode == "new" else "Save changes",
         "error_message": error_message,
--- a/backend/src/meho_backplane/ui/templates/runbooks/editor.html
+++ b/backend/src/meho_backplane/ui/templates/runbooks/editor.html
@@ -14,8 +14,8 @@
   Alpine model:
     * Top-level fields: slug (new only), title, description, target_kind.
     * steps[]: the discriminated-union step tree, hydrated from
-      ``initial_steps_json`` (one blank step on a fresh new; the existing
-      steps on an edit). Add / remove / reorder mutate this array.
+      ``form_steps`` via ``| tojson`` (one blank step on a fresh new; the
+      existing steps on an edit). Add / remove / reorder mutate this array.
     * On submit, the whole step tree is JSON-serialised into the hidden
       ``steps`` form field so the union shape survives the round-trip; the
       params/expect JSON textareas are passed through as raw strings and
@@ -33,7 +33,8 @@
     mode              -- "new" | "edit"
     slug              -- str (locked on edit; the model's slug on new)
     title/description/target_kind -- str: prefilled top-level fields
-    initial_steps_json -- str: JSON array hydrating the Alpine ``steps`` model
+    form_steps        -- list[dict]: the step tree hydrating the Alpine
+                         ``steps`` model (rendered with ``| tojson``)
     submit_url        -- str: POST target (/ui/runbooks/new | .../{slug}/edit)
     submit_label      -- str: button label
     error_message     -- str | None: inline error banner copy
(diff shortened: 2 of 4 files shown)


damir-topic:
@coderabbitai review

damir-topic (commented):




**Anchor task:** evoila-bosnia/meho-internal#100 — harden runbook editor `x-data` against attribute-breakout XSS
**Surface area:** backend (server-rendered HTMX/Jinja UI)
**Best-practices packs audited:** python, frontend (template output-encoding)
**Verdict:** APPROVE
**Head SHA reviewed:** `1eb9edcb37e1368dacac2938079bd352ead9f4ba`
**AI reviewer (CodeRabbit):** not received within window — CodeRabbit returned a rate-limit notice ("Review limit reached", ~47 min) instead of a substantive review. Proceeded on the independent deep review.

### Acceptance criteria
- [x] AC1 — `x-data='runbookEditor(` present exactly once; `x-data="runbookEditor(` absent (verified by grep).
- [x] AC2 — no `| safe` in `editor.html` (verified).
- [x] AC3 — no `json.dumps` call inside `build_editor_context`; raw `form_steps` list passed through (the only `json.dumps` token in the function is docstring prose).
- [x] AC4 — stored-XSS regression test (`GET .../edit`) reuses `_assert_no_xss_breakout`; independently confirmed it FAILS on the pre-fix double-quoted `| safe` template and PASSES after.
- [x] AC5 — reflected-XSS regression test (`POST /ui/runbooks/new` 422 re-render) added with the same parser-grounded check; also independently confirmed fail-pre / pass-post.
- [x] AC6 — `pytest tests/test_ui_runbooks_editor.py tests/test_ui_runbooks_lifecycle.py tests/test_ui_runbooks_acceptance.py` green (plus the harness suite) — 57 passed in the worktree.
- [x] AC7 — `ruff check` + `ruff format --check` + `mypy` clean for `editor.py`.

### Security correctness (the load-bearing check)
- Rendered jinja2 3.1.6 `tojson` directly: it escapes `'`→`'` (no raw single quote can close the single-quoted attribute), `<`/`>`/`&`→`\uXXXX`, and leaves `"` as `\"` inert inside the value. Every interpolated field on the `<section>` element — `mode`, `initialSteps` (`form_steps` step tree), `slug`, `title`, `description`, `target_kind` — now rides inside the single-quoted `x-data` via `| tojson`. No raw scalar, no `| safe`.
- App env uses `select_autoescape(["html","htm","xml"])`, so `.html` templates autoescape and `tojson` (HTML-safe regardless) is sound in production, not just in test.
- HTML-parser-grounded: reverting the template to the pre-fix double-quoted form makes the parser graft `onfocus`/`autofocus`/`onerror` as live attributes (test fails); the single-quoted fix contains the entire payload inside the one `x-data` value (only `x-data` parsed; test passes). No untrusted field can introduce a new attribute or tag.
- `mode | tojson` is correct: `mode` is `Literal["new","edit"]`, so it emits `"new"`/`"edit"` (a JSON string) — behavior-identical to the prior `'{{ mode }}'`, and the inline `runbookEditor` factory consumes `cfg.mode` as a string (`this.mode !== "new"`), `cfg.initialSteps` as an array. Hydration is unchanged (confirmed by the 35 lifecycle/acceptance tests).
- "Last `| safe` in HTML-attribute context" claim verified: the only remaining `| safe` uses (`_PATHS[name]` static icon paths; Pygments `code_css` inside `<style>`) are element-content, not attribute context, and not untrusted user input.

### Project-board alignment
|---|---|---|---|---|---|
| meho-internal#100 | Open (security task, Parent initiative #91) | `Refs` (private repo — cross-repo close not automatic) | manual | ✅ full | close manually on merge |

The PR uses `Refs evoila-bosnia/meho-internal#100` (correct — cross-org private task can't be auto-closed from a public-repo PR; the security-finding repo-split convention applies). Every line of the task's prescribed scope (template single-quote, `initial_steps_json`→`form_steps` rename, `json.dumps` removal, docstring + context-table updates, both regression tests) is delivered.

### Predicted vs. actual diff
- ✅ `ui/templates/runbooks/editor.html` — single-quote + `| tojson` + rationale comment
- ✅ `ui/routes/runbooks/editor.py` — `form_steps` raw list, docstring updated
- ✅ `tests/test_ui_runbooks_editor.py` — stored + reflected regression tests
- ✅ `docs/codebase/ui.md` — migration recorded, canonical rule named
- 🕳️ none missing — full scope delivered

### Integration verification
- [x] `ruff check` — clean (worktree)
- [x] `ruff format --check` — clean (worktree)
- [x] `mypy editor.py` — clean (worktree)
- [x] `pytest` (named targets + harness suite) — 57 passed (worktree)
- [x] Pre-fix regression proof — both XSS tests independently confirmed fail-pre / pass-post
- [x] Rendered-output proof — jinja2 3.1.6 `tojson` escaping verified directly + HTML-parser breakout check
- ⚠️ CI `Python (ruff + mypy + pytest)` lane is RED on one unrelated test — `test_connectors_argocd_write_e2e.py::test_park_app_set_populates_before_after_preview` (respx route-not-called isolation flake). It PASSES standalone in the PR worktree; the PR touches no connector code. Re-run clears. Not a blocker.

None. No blockers, no majors, no minors requiring an inline change.

- The commit carries no `Signed-off-by` trailer, so the DCO check is `ACTION_REQUIRED`. Not a code-correctness issue; the orchestrator's admin merge bypasses red DCO. Noted for the merge step.
- CI's argocd e2e test is an order-dependent respx flake; a `gh run rerun` on the failed Python lane should green it. Unrelated to this PR.
- CSP `frame-ancestors` / `X-Frame-Options` and the `<script type="application/json">` data-island refactor remain tracked separately (correctly out of scope here).

- Adopted: 0 — CodeRabbit was rate-limited and produced no substantive findings this iteration.
- Disputed: 0
- Out of scope: 0

### Docs
- [x] `docs/codebase/ui.md` — updated; records the migration and names single-quoted `x-data` + `| tojson` as the canonical UI output-encoding rule. Accurate.

- The fix is minimal, precedent-matching (PR #1044), and the inline template comment makes the single-quote rationale non-regressable.
- The regression tests are genuine: parser-grounded via the shared `_assert_no_xss_breakout` harness, covering both stored and reflected paths, and verifiably fail against the pre-fix template (independently reproduced during this review).

PR #1861, opened 2026-06-18
fix(targets): make import --dry-run existence-aware (stop showing CREATE for existing targets)

### LINKED ISSUE #1785: Task: fix meho targets import --dry-run showing CREATE for existing targets (ignores --update)
Parent initiative: #1575
Parent goal: #221
Related: #214 (Connector parity — targets/onboarding domain); #257 (original `meho targets import` feature with `--update`/`--dry-run`/`--json`)

## Problem

`meho targets import --update --dry-run` prints a **CREATE** plan for a target that already exists (`targets describe`/`list` show it fine), while the real apply correctly reports `0 created, 1 updated`. The dry-run plan is misleading for existing targets. Surfaced in the same dogfood session as #1774 but **unrelated** — this is a CLI-only (Go) bug, not a TLS issue.

- [`cli/internal/cmd/targets/import.go:238-239`](cli/internal/cmd/targets/import.go#L238-L239) — `runImport` routes `--dry-run` through `buildOfflinePlan(entries, opts.Update)`.
- [`import.go:409`](cli/internal/cmd/targets/import.go#L409) — `func buildOfflinePlan(entries []map[string]any, _ bool)` **discards** the `--update` flag (`_ bool`) and never queries existing targets, so it unconditionally classifies every entry as CREATE.
- [`import.go:249`](cli/internal/cmd/targets/import.go#L249) — the apply path calls `buildLivePlan(ctx, doer, entries, opts.Update)` → [`listExistingNames` (L651)](cli/internal/cmd/targets/import.go#L651) (`GET /api/v1/targets`) → emits UPDATE for existing names. Same comparison key (target `name`) both sides; the **only** difference is dry-run skips the existence GET.
- [`import_test.go:251`](cli/internal/cmd/targets/import_test.go#L251) `TestBuildOfflinePlanEveryEntryIsCreate` and [`:455`](cli/internal/cmd/targets/import_test.go#L455) `TestRunImportDryRunSkipsBackplaneCalls` codify the buggy "dry-run never touches the network" contract; `fakeDoer` (L273, with `existing: []string{"rdc-vault"}` at L312) is the ready-made harness.

## Desired state

- Route `--dry-run` through the **live** planner: resolve the backplane + authed doer, call `buildLivePlan(...)`, then **return before** `executePlan` (L270). Dry-run becomes existence-accurate **and** read-only (one `GET`, zero `POST`/`PATCH`). Delete the now-dead `buildOfflinePlan`.
- Preserve the sparse-PATCH body contract (existing entries render via `entryToUpdateBody`, not `entryToCreateBody`) so the previewed UPDATE matches what apply will PATCH.
- **Tradeoff to flag in the PR:** this breaks the current "dry-run makes no API call / air-gapped" contract. Rewrite `TestRunImportDryRunSkipsBackplaneCalls` → "dry-run issues GET but no POST/PATCH"; replace `TestBuildOfflinePlanEveryEntryIsCreate` with a dry-run test using `fakeDoer{existing: [...]}` asserting an existing target renders **UPDATE**, and keep a brand-new-target case rendering CREATE.

## Acceptance criteria

- [ ] `meho targets import --update --dry-run targets.yaml` against a tenant where the target exists prints `UPDATE <name>` / "0 to create, 1 to update", matching the real apply's "0 created, 1 updated".
- [ ] Dry-run issues exactly one `GET` and zero `POST`/`PATCH` (assert on `fakeDoer`).
- [ ] A brand-new target still renders CREATE in dry-run.
- [ ] `cd cli && go test ./internal/cmd/targets/` green; the old `buildOfflinePlan` is removed and no longer referenced.

## Out of scope

- Anything TLS / #1774. Server side (the upsert is already correct — it never participates in dry-run).
- Offline-without-login behaviour: if retaining an air-gapped degraded mode is wanted, that's a separate decision; this Task routes dry-run through the live existence query.

## References

- [`cli/internal/cmd/targets/import.go`](cli/internal/cmd/targets/import.go) (`runImport` L219, dry-run branch L238, `buildOfflinePlan` L409, `buildLivePlan` L573, `listExistingNames` L651), [`import_test.go`](cli/internal/cmd/targets/import_test.go) (`fakeDoer` L273).
- Cobra command patterns: https://github.com/spf13/cobra/blob/main/site/content/user_guide.md
- Go `testing` table tests: https://go.dev/wiki/TableDrivenTests

damir-topic:
## Summary
- `meho targets import --update --dry-run` printed `CREATE` for targets that already exist, because the dry-run branch routed through `buildOfflinePlan` — which discarded `--update` and never queried existing targets, so every entry was classified `CREATE` while the real apply correctly reported `0 created, 1 updated`.
- Route `--dry-run` through `buildLivePlan` instead: it issues the single listing `GET /api/v1/targets`, builds the existence-accurate plan (existing names → `UPDATE` with the sparse-PATCH body, new → `CREATE`), and returns before `executePlan`. Dry-run is now read-only (one GET, zero POST/PATCH) and its preview matches what apply would do.
- Deleted the now-dead `buildOfflinePlan`.

## Tradeoff (flagged)
This intentionally retires the old "dry-run makes no API call / air-gapped" contract — dry-run now requires a valid `meho login` session for the existence GET. The previously-buggy "every entry is CREATE" preview is replaced by an accurate one. The `docs/cross-repo/targets-yaml.md` drift-detection section already assumed this live behaviour; Step 1 and the CLI help text are updated to match.

## Test plan
- [x] `gofmt -l` — clean; `go vet ./internal/cmd/targets/` — clean; `go build ./...` — clean
- [x] `cd cli && go test ./internal/cmd/targets/` — green; full `go test ./...` — green
- [x] **AC1** existing target previews UPDATE / "1 to update": `TestRunImportDryRunExistingTargetRendersUpdate`
- [x] **AC2** dry-run issues one GET, zero POST/PATCH: `TestRunImportDryRunIssuesGetButNoWrites`, `TestDryRunPlanExistingRendersUpdateNewRendersCreate`
- [x] **AC3** brand-new target still CREATE in dry-run: `TestRunImportDryRunPrintsPlan` + the new-target arm of the plan test
- [x] **AC4** `buildOfflinePlan` removed and unreferenced (`grep` clean), targets package green
- [x] Rewrote the two tests that codified the old offline contract (`TestBuildOfflinePlanEveryEntryIsCreate` → existence-aware plan test; `TestRunImportDryRunSkipsBackplaneCalls` → "GET but no writes")

## Out of scope
- TLS / #1774 (separate concern, same dogfood session).
- Server side (the upsert is already correct and never participates in dry-run).
- Retaining an air-gapped degraded dry-run mode — a separate decision per the issue.

Closes #1785

### DIFF
--- a/cli/internal/cmd/targets/import.go
+++ b/cli/internal/cmd/targets/import.go
@@ -187,9 +187,11 @@ func newImportCmd() *cobra.Command {
 			"`name` already exists in the tenant (no partial write — the plan " +
 			"is built before any API call fires). `--update` PATCHes existing " +
 			"targets with the fields present in the YAML and POSTs new ones, " +
-			"mixed-mode-safe. `--dry-run` prints the plan and returns without " +
-			"calling the apply path. `--json` formats the plan as a structured " +
-			"object (use with --dry-run).\n\n" +
+			"mixed-mode-safe. `--dry-run` queries existing targets (one " +
+			"read-only GET) and prints the resulting plan — existing names " +
+			"render UPDATE, new ones CREATE — then returns before any " +
+			"POST/PATCH. `--json` formats the plan as a structured object " +
+			"(use with --dry-run).\n\n" +
 			"Authentication. Uses the token `meho login` wrote, with the same " +
 			"401-refresh-retry behaviour as `meho status` and `meho operation " +
 			"call`. The tenant is the operator's JWT-bound tenant — there is no " +
@@ -210,7 +212,7 @@ func newImportCmd() *cobra.Command {
 	cmd.Flags().BoolVar(&updateMode, "update", false,
 		"PATCH existing targets instead of erroring on duplicate names")
 	cmd.Flags().BoolVar(&dryRun, "dry-run", false,
-		"print the plan; no API calls")
+		"print the plan (read-only: one GET, no writes); does not apply")
 	cmd.Flags().BoolVar(&jsonOut, "json", false,
 		"output the plan as JSON (use with --dry-run)")
 	cmd.Flags().StringVar(&backplaneOverride, "backplane", "",
@@ -241,17 +243,6 @@ func runImport(cmd *cobra.Command, opts importOptions) error {
 			output.Unexpected(err.Error()), opts.JSONOut)
 	}
 
-	// Dry-run path: resolve nothing against the backplane. Existence
-	// is unknown, so every entry plans as CREATE. Operators wanting
-	// "what would change" against a live tenant pass --dry-run after
-	// running once with --update, or just look at `meho targets list`
-	// before running. v0.2 keeps the no-API-call contract strict; a
-	// future `--dry-run --against-tenant` could relax it.
-	if opts.DryRun {
-		p := buildOfflinePlan(entries, opts.Update)
-		return renderPlan(cmd, p, opts.JSONOut)
-	}
-
 	backplaneURL, err := backplane.Resolve(opts.BackplaneOverride)
 	if err != nil {
 		return output.RenderError(cmd.ErrOrStderr(), backplane.ClassifyError(err), opts.JSONOut)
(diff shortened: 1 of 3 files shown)


damir-topic:
@coderabbitai review

PR #1869, opened 2026-06-18
feat(knowledge): kb upsert 200-on-overwrite + entry write attribution

### LINKED ISSUE #1845: kb: same-slug cross-principal write is a silent 201 clobber; cross-principal DELETE not principal-scoped; entries attribution-blind
Parent initiative: #1852
Parent goal: #215

## Summary

On **v0.16.0** (git `6bc8e7fc`), the knowledge-base surfaces (`/api/v1/kb` + `POST /api/v1/retrieve {source:"kb"}`) exhibit one architectural property worth documenting/correcting: kb rows are tenant-shared with **no per-row ownership check** and **no entry-level authorship/mutation attribution**. Found via a downstream cross-principal contract test (sibling of the memory-isolation finding #1797), using **two distinct `tenant_admin` principals in one tenant** (A, B) plus a third **service-account** principal (C).

This is **category-2** — *not* a cross-tenant breach (every write stays in-tenant; tenant_admins are mutually trusted by design), and lower severity than the memory per-principal-body leak (#1797). But three facets of the contract are reportable.

### 1. Same-slug different-body cross-principal write → silent in-place clobber, HTTP 201

- **A:** `POST /api/v1/kb {slug:"s", body:"<A body>"}` → `201`, `id=X`, `created_at=T0`, `updated_at=T0`.
- **B:** `POST /api/v1/kb {slug:"s", body:"<B's DIFFERENT body>"}` → **`201`**, `id=X` (unchanged), `created_at=T0` (unchanged), `updated_at=T1` (bumped), **body replaced in place**. No duplicate row; no `403`/`409`/ownership rejection.
- Any subsequent `GET /api/v1/kb/{slug}` returns B's body on A's row.
- Confirmed **bidirectionally** (B→A's slug, and C→B's slug — identical shape each way).

### 2. Cross-principal DELETE → succeeds, row actually gone

- **C:** `DELETE /api/v1/kb/{slug}` on a row written by B → **`204`**; subsequent `GET /api/v1/kb/{slug}` → `404 {"detail":"slug_not_found"}`; `POST /api/v1/retrieve` confirms the slug is gone from the index.
- kb DELETE is **not** principal-scoped (any tenant_admin can delete any other's row). Note: this is **distinct from memory DELETE**, which *is* principal-scoped (cross-principal memory DELETE is a silent `204` no-op — see #1797 cleanup). Same verb shape, different scoping discipline across the two surfaces.

### 3. Attribution-blindness

Every read of a kb entry — `POST /api/v1/kb`, `GET /api/v1/kb/{slug}`, `GET /api/v1/kb`, `POST /api/v1/retrieve {source:"kb"}` — returns:

```
{id, tenant_id, slug, body, metadata: {}, created_at, updated_at}
```

— **no** `user_sub` / `owner_sub` / `created_by` / `author_sub` anywhere (recursive key scan empty; retrieve hits add `bm25_*`/`cosine_*`/`fused_score`/`document_id`/`kind`/`source`/`doc_metadata`, none carrying authorship). There is no on-entry trace of *who* wrote it or *that* it was mutated after creation (`updated_at != created_at` is the only hint). Forensics ("who set this wrong value, and when?") requires audit-log correlation.

## Expected vs observed

The docs document only the **same-body** idempotent body-hash short-circuit. The **different-body cross-principal** case is undocumented and produces a `201`-on-overwrite — REST-incorrect, since `201 Created` should not be returned for an in-place mutation of an existing row.

## Asks (priority order)

1. **HTTP status:** `201` for fresh creates, `200` for overwrites. The response body already differentiates (preserved `id` + `created_at` on overwrite vs fresh on create); the status should match.
2. **Document** the cross-principal upsert + delete semantics — make the wiki-like "any tenant_admin can replace/delete any other tenant_admin's row" explicit, so operators don't assume per-author ownership.
3. **Consider** surfacing `last_updated_by_sub` (and/or `created_by_sub`) on the kb entry shape (even just as a `metadata.last_updated_by_sub` field) so kb is self-describing instead of requiring audit-log correlation.

## Severity

Category-2: documentation + wire-shape correction + entry-shape enrichment. Sibling of #1797 (memory per-principal leak) — same `/api/v1/retrieve` engine, but the **retrieve path is correct for kb** (kb has no per-principal dimension to filter or leak). The issue here is the **write/delete ownership model + attribution**, not retrieval isolation.

damir-topic:
## Summary

- **HTTP status (ask 1):** `POST /api/v1/kb` now returns **201** only for a genuinely-new slug and **200** for an in-place overwrite of an existing slug. `201` on an overwrite is REST-incorrect; the service reports created-vs-overwrite and the route maps it via an injected `Response` (the decorator keeps `201` as the OpenAPI default; `200` is advertised as a documented alternate response).
- **Attribution (ask 3):** every kb write stamps `created_by_sub` (set once, preserved verbatim across overwrites — including cross-principal ones) and `last_updated_by_sub` (rewritten each write) into `documents.doc_metadata`. **No schema migration** — the keys ride the existing JSONB and surface on all four read surfaces named in the issue (`POST /api/v1/kb`, `GET /api/v1/kb/{slug}`, `GET /api/v1/kb` list preview, `POST /api/v1/retrieve {source:"kb"}` hits). Authorship derives from the verified `Operator.sub`; caller-supplied attribution keys are stripped so an operator cannot forge it.
- **Docs (ask 2):** the wiki-like cross-principal upsert/delete model (any `tenant_admin` may replace/delete any other's row in-tenant — intended, no ownership gate) is now explicit in the route module docstring and `docs/codebase/kb.md`, contrasted with `memory`'s principal-scoped DELETE. **No ownership enforcement added** (explicitly out of scope per the issue).

`create_entry` now returns `(entry, created)` and takes an `actor_sub` kwarg; the MCP `add_to_knowledge` tool and both UI write paths were updated. The merge logic lives in a new `kb/attribution.py` to keep `service.py` under the file-size budget.

## Test plan

- [x] `ruff check` (src + tests) — clean
- [x] `mypy src/meho_backplane` — clean (481 files)
- [x] `code-quality.py --diff` — 0 blockers (remaining warnings are pre-existing legacy functions)
- [x] `pytest` kb + retrieve + indexer + db-documents suites — 184 passed
- [x] New unit tests: 201-on-create, 200-on-overwrite, audit `created` flag, `actor_sub` forwarded; service-level attribution (create stamps both subs, cross-principal overwrite preserves creator + bumps mutator, caller-forged attribution stripped, unattributed `actor_sub=None` writes no keys)
- [x] New integration test (`test_kb_routes_pg.py`): cross-principal overwrite returns 200 with preserved `id`/`created_at` and `created_by_sub`/`last_updated_by_sub` surfacing on `GET /{slug}` (Docker-gated; runs in CI)
- [x] CLI OpenAPI snapshot + generated Go client regenerated (`cd cli && make snapshot-openapi && make generate`); `go build ./...` clean

## Out of scope

- Per-row ownership enforcement on write/delete — the wiki-like model is intended (issue explicitly excludes this).
- Tarball ingest, per-entry version history — pre-existing v0.2.next deferrals, untouched.

Closes #1845

### DIFF
--- a/backend/src/meho_backplane/api/v1/kb.py
+++ b/backend/src/meho_backplane/api/v1/kb.py
@@ -18,12 +18,28 @@ Route inventory
 * ``GET /api/v1/kb/{slug}`` -- fetch one entry by slug. Returns
   :class:`KbEntry`. 404 when absent. Role: ``operator``.
 * ``POST /api/v1/kb`` -- create / re-index one entry. Body:
-  :class:`KbEntryCreate`. Returns :class:`KbEntry` with HTTP 201. Role:
-  ``tenant_admin``.
+  :class:`KbEntryCreate`. Returns :class:`KbEntry` with HTTP **201**
+  for a genuinely-new slug and **200** for an in-place overwrite of an
+  existing slug (#1845 -- ``201`` on overwrite is REST-incorrect).
+  Role: ``tenant_admin``.
 * ``DELETE /api/v1/kb/{slug}`` -- delete an entry. Returns 204 on
   success **and** when the row was already absent (idempotent
   semantics per the acceptance criteria; T6 documents the
   "delete-already-missing is 204" contract). Role: ``tenant_admin``.
+
+Cross-principal write & delete model (#1845)
+--------------------------------------------
+
+kb rows are tenant-shared with **no per-row ownership check**: any
+``tenant_admin`` may overwrite or delete **any** other principal's slug
+in the same tenant. This is **wiki-like and intentional** -- there is
+no ``403``/``409`` on a same-slug cross-principal write, and ``DELETE``
+is not principal-scoped. Operators must not assume per-author
+ownership. (This differs from ``memory`` DELETE, which *is*
+principal-scoped.) Writes stamp ``created_by_sub`` /
+``last_updated_by_sub`` into the entry's ``metadata`` so a row is
+self-describing about authorship without an audit-log join; see
+:func:`meho_backplane.kb.attribution.merge_attribution`.
 * ``POST /api/v1/kb/ingest`` -- server-side bulk ingest. Body:
   :class:`IngestKbRequest`. Returns :class:`KbIngestionResult`. Role:
   ``tenant_admin``.
@@ -360,9 +376,20 @@ async def show_kb(
     "",
     response_model=KbEntry,
     status_code=http_status.HTTP_201_CREATED,
+    responses={
+        http_status.HTTP_200_OK: {
+            "model": KbEntry,
+            "description": (
+                "An existing entry with this slug was overwritten in "
+                "place (id + created_at preserved, updated_at bumped). "
+                "201 is returned only for a genuinely-new slug."
+            ),
+        },
+    },
 )
 async def create_kb(
     body: KbEntryCreate,
+    response: Response,
     operator: Operator = _require_admin,
 ) -> KbEntry:
     """Create / re-index one kb entry under the operator's tenant.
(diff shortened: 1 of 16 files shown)


damir-topic:
@coderabbitai review

PR #1870, opened 2026-06-18
docs(targets): align import --dry-run godoc with read-only behavior

damir-topic:
## What

Follow-up to #1785 (PR #1861). The `newImportCmd` godoc still described `--dry-run` as `# print the plan; no API calls` — stale since #1785 routed dry-run through the live existence-aware planner (one GET, no writes). This aligns the comment with the already-updated flag help (`print the plan (read-only: one GET, no writes); does not apply`).

## Scope

- Comment-only. One line in `cli/internal/cmd/targets/import.go`.
- No behavior change; `gofmt`/`go vet`/`go build ./...` clean.

Refs #1785.


damir-topic:
@coderabbitai review

Issue #1804, opened 2026-06-16
Task: surface connector auth/session failures (401 + appliance auth statuses) as a structured dispatch result instead of flat connector_error


damir-topic:
Parent goal: #221
Parent initiative: #1800

## Why now

RDC v0.16.0 dogfood: an operator dispatching against a vRLI appliance saw `connector_error (440)` — an opaque, unactionable result that gave no hint it was an **auth/session** failure. This is the diagnosability gap that made #1798 look like a stub-auth problem. The dispatcher already learned to surface a *specific* failure class structurally for TLS (#1782 `connector_tls_verify_failed`) and credential-deny (`result_connector_http_403`); auth/session expiry never got the same treatment and flattens to the generic catch-all.

## Current state (grounded, v0.16.0)

The dispatcher's `httpx.HTTPStatusError` arm (`operations/dispatcher.py:855-898`) special-cases exactly two statuses:

```python
status_code = http_exc.response.status_code
if status_code == 403:
    return result_connector_http_403(op_id, http_exc, duration_ms)
if status_code == 422:
    return result_connector_http_422(op_id, http_exc, duration_ms)
return result_connector_error(op_id, http_exc, duration_ms)   # everything else: flat + opaque
```

So **401** (the canonical session-expiry / re-login-failed status for the hand-coded session connectors), and any appliance auth status like vRLI's `440`, fall through to `result_connector_error` — which discards the cause and offers no remediation. The `ConnectError` arm below it got the structured `connector_tls_verify_failed` treatment in #1782 (`dispatcher.py:899+`), establishing the pattern; auth-class statuses were never given a builder.

The hand-coded session connectors (NSX, vcf_logs, vmware_rest, …) already retry once on 401 internally (`_get_json_with_session_retry`), so a **401 that reaches the dispatcher means re-login also failed** — i.e. the credential is missing/invalid/expired in Vault, or the `auth_model` is wrong. That is precisely the actionable case an operator needs named.

## Desired state

[...]

PR #1921, opened 2026-06-19
fix(connectors): re-login vcf-logs on vRLI 440 (session expired), not just 401

### LINKED ISSUE #1909: vcf-logs connector re-establishes the vRLI session only on 401, but vRLI expires sessions with 440 — expired sessions never recover
Parent initiative: #1911
Parent goal: #221

> **Triage note (v0.17.0 dogfood verification).** Symptom + root cause **confirmed against source**: the re-login trigger keys strictly on 401 at `backend/src/meho_backplane/connectors/vcf_logs/connector.py:399`, while vRLI signals recoverable session-expiry with **440**. Two corrections for the implementer: (1) the **dispatcher-side 440 handling lives under `operations/`, not `connectors/`** — `operations/_errors.py` (`_AUTH_FAILED_STATUSES`:830, `is_auth_failed_status`:833, misleading message ~:900) and `operations/dispatcher.py:940-941`. (2) **#1804 already added 440 to `_AUTH_FAILED_STATUSES`, but that is only the *diagnosability* half** (flat `connector_error` → structured `connector_auth_failed`); it does **not** add re-login. The actual fix here is in the connector's session-retry (treat 440 like 401), plus fixing the `connector_auth_failed` message that says "retried once on a 401" for the unretried-440 case.

Dispatches to a vRLI (VCF Operations for Logs) target work on the first call after a backplane start, then start failing with `connector_auth_failed http_status=440 "Login Timeout"` and stay failing — until the backplane is restarted, after which the first call works again and the cycle repeats.

### Root cause

`connectors/vcf_logs/connector.py::_get_json_with_session_retry` only re-establishes the session on a 401:

```python
except httpx.HTTPStatusError as exc:
    if exc.response.status_code != 401:   # 440 falls straight through
        raise
    await self._invalidate_session(target)  # + retry once
```

But the vRLI OpenAPI spec defines the two auth responses as:

- **`trait.authenticated.440`** — *"The request was refused because the session ID has **expired**. Obtain a new session ID from `/api/v2/sessions`."* (carried by 117 endpoints)
- `trait.authenticated.401` — *"lacks valid authentication credentials … `Authorization` header missing or invalid session ID."*

So **440 is the recoverable "re-login now" case, and it's exactly the one the connector ignores.** The connector re-logins on 401 (frequently the non-recoverable bad-credentials case) and re-raises 440 (the case a re-login actually fixes).

### Why it bites

The session token is cached in-memory per target with no proactive refresh. vRLI idle-expires the session; the next downstream call gets 440; the connector re-raises instead of re-logging-in; every subsequent call 440s until a backplane restart clears the in-memory cache. This breaks any scheduled / long-running vRLI consumer — e.g. an hourly agent logs in on run 1, the session idles out, and runs 2..N all 440.

Observed live on v0.17.0 against a vRLI 9.0.2 target: `GET:/api/v2/version` returned 200 once (cached session), then `connector_auth_failed http_status=440` on every later call (~15ms — the instant vRLI "expired session" reason-phrase, not a network stall), identical across two separately-ingested connectors (so it's connector-instance-independent).

### Fix

Treat **440 (session expired)** like 401 in the re-login trigger — invalidate the cached token, re-login via `POST /api/v2/sessions`, retry once. The vendor spec literally says 440 means "obtain a new session ID", so re-login on 440 is the documented recovery. (Arguably re-login on 440 *or* 401.)

Minor: the `connector_auth_failed` message says "for the session connectors this is a re-login failure (they already retry once on a 401)" even when the actual failure is an **unretried 440** — misleading while debugging.

### Refs

- `backend/src/meho_backplane/connectors/vcf_logs/connector.py` → `_get_json_with_session_retry`
- vRLI OpenAPI spec → `components.responses.trait.authenticated.{440,401}`

damir-topic:
## Summary

- The vcf-logs connector's session-retry (`_get_json_with_session_retry`) keyed re-login strictly on HTTP **401**, so vRLI's own session-expiry status — **440** (`trait.authenticated.440`, *"the session ID has expired; obtain a new session ID from `/api/v2/sessions`"*) — fell straight through unretried. vRLI idle-expires its in-memory session, so the first call after a backplane start worked off the cached token, then every later call `440`'d until a restart cleared the cache — breaking any scheduled / long-running vRLI consumer (observed live on v0.17.0 against vRLI 9.0.2: `GET:/api/v2/version` 200 once, then `connector_auth_failed http_status=440` on every later call).
- Fix: treat 440 like 401 in the re-login trigger via `_SESSION_EXPIRED_STATUSES = {401, 440}`. A 440 (or 401) now invalidates the cached token, re-logs-in via `POST /api/v2/sessions`, and retries **once**; a second 440/401 still fails fast with `RuntimeError` naming the target + status (no retry loop — the NSX posture).
- #1804 already routed 440 to a structured `connector_auth_failed` at the dispatcher, but that was only the *diagnosability* half (flat `connector_error (440)` → structured); the **recovery** (re-login on 440) belongs here in the connector's session-retry.
- Also fixes the misleading `connector_auth_failed` message (`operations/_errors.py`) that claimed the session connectors *"already retry once on a 401"* even for an unretried 440 — it now names the session-expiry set (401 **or** vRLI's 440) on the `_get_json_with_session_retry` path.

## Test plan

- [x] `ruff check` — clean (connector.py, _errors.py, e2e test)
- [x] `mypy --ignore-missing-imports` — Success, no issues
- [x] **440 triggers re-login + retry-once** — `test_vrli_e2e_440_recovery_via_connector_method` (mirrors the 401-recovery test; respx mocks the session POST + downstream GET): session-create called twice, version GET called twice, cache holds the refreshed token. PASS
- [x] **Second 440 fails fast** — `test_vrli_e2e_second_440_fails_with_runtime_error`: `RuntimeError` naming the target + `"440"`, no third re-login. PASS
- [x] 401 path unchanged — existing `test_vrli_e2e_401_recovery_*` / `*_second_401_*` still PASS
- [x] Message change — `tests/test_operations_connector_auth_failed.py` 28 passed (no stale-message assertion broke; the `"session token invalid or expired"` assertion checks the *upstream-body* tail, preserved)
- [x] `pytest tests/test_connectors_vcf_logs_e2e.py tests/test_operations_connector_auth_failed.py` — 28 passed
- [x] `pytest tests/test_connectors_vcf_logs_auth.py tests/test_connectors_vcf_logs_credread.py tests/test_connectors_vcf_shared_auth.py tests/test_auth_failures.py` — 87 passed

## Docs

- `docs/codebase/connectors-vcf-logs.md` — "401 retry-once" section → "Session-expiry retry-once (440 or 401)", control-flow diagram, Known-issues, references updated.
- `docs/codebase/error-message-shape.md` — `connector_auth_failed` row clause updated to the session-expiry set.

## Out of scope

- No proactive token refresh on TTL — explicitly deferred per the connector's "Session lifecycle" docstring; this PR only adds on-demand recovery from an expired session, which is the documented vendor recovery (440 → "obtain a new session ID").
- The dispatched ingested-op path (e.g. `vrli.event.query`) does **not** go through `_get_json_with_session_retry` — it calls `_request_json` and the dispatcher classifies the resulting 440 as `connector_auth_failed` (unchanged). The session-retry layer covers `fingerprint()` / `probe()` and future typed ops.

Closes #1909


damir-topic:
@coderabbitai review

Issue #1700, opened 2026-06-12
Task: Add connector DELETE surface — clean up zero-op registered stubs left by aborted ingests


damir-topic:
Parent goal: #221
Parent initiative: #1691

**Severity:** SEV-3 · **Area:** connector-ingest · **Source:** claude-rdc-hetzner-dc v0.14.0 cycle-10 dogfood

## Problem

The connector-ingest pipeline auto-registers a GenericRestConnector shim to the v2 registry on the first ingest of a (product, version, impl_id) triple (register_ingested.py:371). If the ingest aborts (spec parses to zero operations, multi-spec batch fails mid-way, etc.) or validation rejects the spec post-registration, the connector class remains registered permanently with no surface to remove it. Over time, operators accumulate stale zero-op stubs in the registry that clutter the connector catalog and consume memory.

## Current state

- No REST DELETE route for connectors exists (grep "@router.delete" across api/v1/connectors*.py finds none)
- No MCP delete tool exists (connector_admin.py + connector_ingest.py register six tools; none delete)
- Connector class registration (ensure_connector_class_registered in connector_registration.py:354-359) is unconditional once pre-flight passes; no rollback on subsequent ingest failure
- Soft-delete pattern exists elsewhere (targets.py uses review_status flag); no equivalent for connectors

## Desired state

Operators can explicitly delete a connector via:
- REST: DELETE /api/v1/connectors/{connector_id} → 204 No Content
- MCP: meho.connector.delete(connector_id, tenant_id=None) → {ok: True}

Both paths:
- Deregister the connector class from the v2 registry (all_connectors_v2() no longer returns it)
- Mark the connector as deleted (operation_groups.review_status='deleted', endpoint_descriptors soft-deleted)
- Return 404 for unknown connector_id or cross-tenant probe (same 404 conflation as other connector routes)
- Emit one audit row (meho.connector.delete)

## Acceptance criteria

- [ ] DELETE /api/v1/connectors/{connector_id} endpoint exists, requires tenant_admin role, returns 204
- [ ] Cross-tenant DELETE returns 404 (same conflation as list/review/enable/disable routes)
- [ ] Deleted connector does not appear in all_connectors_v2() query or list_ingested_connectors() response
- [ ] Deleted connector's endpoint_descriptors are soft-deleted and do not appear in search_operations results
- [ ] meho.connector.delete MCP tool exists with same inputs/outputs as the REST path
[...]

Issue #1534, opened 2026-06-05
MCP ingest strips diagnostic detail to a bare -32603 for the SpecError sibling classes (REST carries it) — extend the #777 envelope pattern


damir-topic:
Parent goal: #214
Parent initiative: #1529

## Problem

On the agent-facing MCP surface, `meho.connector.ingest` degrades every typed ingest error except two to a **bare** `-32603 "internal error: {ClassName}"` with the diagnostic message stripped. This is the single root cause behind the "opaque ingest error" complaints across the campaign (ci-01 `UnsupportedSpecError`, ci-02 `InvalidSpecError` for a `docs:` URI, ci-06 Swagger 2.0, ci-07 `UpstreamNotSpecError`): the exceptions **do** carry informative messages, and the **REST** surface shows them — only the MCP transport throws them away.

Mechanism: the MCP `_ingest_handler` catches only `VersionMismatchError` + `UncoveredVersionLabel` (re-raising as `-32602` with structured `data`); the sibling classes fall through to the dispatcher's generic `except Exception → -32603 "internal error: {type(exc).__name__}"`, which discards `str(exc)`. The fix is the unbuilt remainder of the #777/#785 pattern.

## Current state (verified in code, v0.11.0)

- [`mcp/tools/connector_admin.py:235-254`](backend/src/meho_backplane/mcp/tools/connector_admin.py#L235-L254) — `_ingest_handler` catches only `VersionMismatchError` + `UncoveredVersionLabel`. The siblings `UnsupportedSpecError` / `InvalidSpecError` / `UpstreamNotSpecError` / `InvalidSchemaError` / `OpIdCollision` / `LlmOutputInvalid` are not caught.
- [`mcp/server.py:864-872`](backend/src/meho_backplane/mcp/server.py#L864-L872) — generic `except Exception` → `_error_response(id, INTERNAL_ERROR, f"internal error: {type(exc).__name__}")`; `-32603`, `str(exc)` discarded, no `data`. (`_error_response` accepts a `data` kwarg, so the omission is the gap.)
- [`api/v1/connectors_ingest.py:832-865`](backend/src/meho_backplane/api/v1/connectors_ingest.py#L832-L865) — REST attaches detail for **every** class (400 `str(exc)` / structured 422), so the gap is the MCP↔REST asymmetry, not absent detail.
[...]

Issue #777, opened 2026-05-21
Task: G0.9.1-T5 structured ingest error envelopes on the MCP path — VersionMismatchError/UncoveredVersionLabel detail (Signal #12)


damir-topic:
Parent goal: #221
Parent initiative: #772

## Problem

The Layer 2 connector-ingest validation classes (T8 #740 version-vs-`spec.info.version`, T9 #741 orphan-class) fire correctly, but on the **MCP path** their error envelopes are bare:

> `"internal error: VersionMismatchError"`

— a typed class name with no diagnostic of what was expected vs received. T8's whole purpose is to surface a silent-mismatch class *clearly*; surfacing it as a `-32603` internal error with only the class name is one polish-iteration short. The consumer's suggested shape:

```json
{"error": "VersionMismatchError",
 "detail": "operator label 'version=9.0' does not match spec.info.version '1.0.11'"}
```

## Root cause

The exceptions already carry rich messages. [exceptions.py](backend/src/meho_backplane/operations/ingest/exceptions.py) defines `VersionMismatchError` (L413-493, formats `requested_version` alongside each `(uri, info.version)`) and `UncoveredVersionLabel` (L311-381, lists every registered class's `supported_version_range`). The **REST** route catches them and maps to `422` with structured detail ([connectors_ingest.py:241-315](backend/src/meho_backplane/api/v1/connectors_ingest.py#L241)). The **MCP** path doesn't: the `meho.connector.ingest` handler ([mcp/tools/connector_admin.py](backend/src/meho_backplane/mcp/tools/connector_admin.py)) lets the exception propagate to the dispatcher's generic handler ([mcp/server.py:495-502](backend/src/meho_backplane/mcp/server.py#L495)):

```python
except Exception as exc:
    _log.exception("mcp_handler_error", method=jrpc.method)
    return _error_response(jrpc.id, INTERNAL_ERROR, f"internal error: {type(exc).__name__}")
```

`type(exc).__name__` discards the (already-detailed) message. JSON-RPC 2.0 reserves the error `data` member exactly for structured detail, and these are validation errors (caller's input), not internal faults — they should not be `-32603`.

## Current state

- REST ingest: structured `422` detail (good).
- MCP ingest: generic `-32603 internal error: <ClassName>`, message dropped.
- The detail-formatting logic lives in `connectors_ingest.py` and isn't reused by MCP.

## Desired state

[...]

PR #1923, opened 2026-06-19
fix(connectors): map ambiguous connector scope to structured MCP -32602 (#1910)

### LINKED ISSUE #1910: Product-token reconciliation (#1814) leaves a stale duplicate connector row → curation tools fail AmbiguousConnectorScopeError
Parent initiative: #1911
Parent goal: #221

After the #1814 long→short product-token reconciliation, a connector whose product was renamed (e.g. `vcf-logs` → `vrli`) ends up with **two registry rows sharing the same `(impl_id, version)`** — the new short-product row *and* the stale long-product row. The connector-curation tools then can't resolve which one is meant.

### What happens

`meho_connector_list` shows two rows for the same `impl+version` under different products (the stale one typically has 0 enabled ops):

```
vrli-rest-9.0.2  product=vrli       (… enabled)     ← new, post-#1814
vrli-rest-9.0.2  product=vcf-logs   (0 enabled)     ← stale, pre-#1814
```

The operator-facing connector id is `<impl_id>-<version>` — **no product component**. So the curation tools, which all resolve by `(connector_id, tenant_id)` with no product:

- `meho_connector_review`
- `meho_connector_edit_op`
- `meho_connector_enable`
- `meho_connector_enable_reads`
- `meho_connector_delete`

…match >1 product row and fail with `-32603 internal error: AmbiguousConnectorScopeError` (even when `tenant_id` is passed).

`call_operation` / `preview_operation` are fine — they disambiguate via the **target's** product. So **dispatch works, but curation is blocked**: you can't review / enable / disable / delete the connector, which blocks enabling its staged ops. `meho_connector_delete` is also blocked, so you can't even clean up the stale row through the tool.

### Suggested fixes

- **`#1814` should retire the stale long-product row**, not just add the short one (the reconciliation should be a rename, not an additive insert); and/or
- the curation tools should accept an optional `product` disambiguator, or return a structured `ambiguous: pass product=X|Y` error (with the candidate products) instead of a bare `-32603 internal error`.

### Consumer workaround

Re-ingest the spec under a distinct `version` so the `(impl_id, version)` pair is unique — the curation tools then resolve it. (We did this to enable reads on a vRLI surface; it works, at the cost of an extra connector row.)

damir-topic:
## Summary

- **Closes the last MCP↔REST asymmetry on the connector error surface.** The two scope-resolving connector-curation MCP tools (`meho.connector.review`, `meho.connector.enable_reads`) raised `AmbiguousConnectorScopeError` when a `connector_id` resolved to **both** a tenant-curated row and a built-in (global) row — the state a product-token reconciliation (#1814/#1817) plus a tenant re-ingest leaves behind. The MCP layer had no mapping, so it fell through the dispatcher's generic `except Exception` to a bare `-32603 "internal error: AmbiguousConnectorScopeError"` with the candidate list discarded, **blocking curation** of the affected connector.
- The MCP path now surfaces the **same structured `data` envelope the REST 409 already carried** (`build_connector_scope_ambiguous_detail`: `detail="connector_scope_ambiguous"` + candidate `{product, version, impl_id, tenant_id}` scopes + message) as a JSON-RPC `-32602`, so an operator/agent reads `error.data.candidates` and re-issues with the disambiguating `tenant_id` (`null` for the built-in scope, their own tenant UUID for the tenant row). One builder shared with REST → the two surfaces can't drift. Mirrors the `SpecError`-sibling precedent (#777 / #1534).
- **Re-diagnosis note (RDC report was symptom-right, root-cause-imprecise).** The issue attributed the `-32603` to the visible `vrli`/`vcf-logs` double-list-row. But `parse_connector_id` derives `product` from the `impl_id` prefix unconditionally (`vrli-rest-9.0.2` → `product=vrli`), so the curation tools **never query the stale `vcf-logs` row** — it does not cause the error. The actual trigger is the tenant-vs-built-in ambiguity inside `_resolve_existing_scope` (the only raise site). The stale long-product twin is a deliberately operator-owned cleanup (`meho.connector.delete`; migration `0038` left it on purpose via its `NOT EXISTS` collision-skip), so it is **left untouched** — fixing the error shape is the in-scope, minimal correct fix.

## Test plan

- [x] `ruff check src/ tests/...` — clean
- [x] `ruff format --check` — clean (3 files already formatted)
- [x] `mypy src/` (CI-exact, strict) — Success, 512 source files
- [x] `pytest tests/test_mcp_tools_connector_admin.py tests/test_mcp_tools_connector_ingest.py` — 43 passed
- [x] `pytest tests/test_operations_ingest_review.py` — 55 passed (exercises the `AmbiguousConnectorScopeError` raise path)
- [x] Broad MCP slice (`-k "mcp_tools or connector_shared or ingest_review"`) — 394 passed, 1 skipped
- [x] `code-quality.py --diff` — 0 blockers (1 pre-existing file-size warning on `connector_admin.py`, not introduced here)
- [x] **Acceptance:** new tests assert both handlers (`_review_handler` / `_enable_reads_handler`) raise `McpInvalidParamsError` with the structured `data` envelope, and two end-to-end `tools/call` wire tests assert the response carries `-32602` + `error.data.candidates` (not a bare `-32603`).

## Out of scope / adjacent findings

- **Stale long-product duplicate row cleanup.** Retiring the `vcf-logs` twin would be a data migration; it does not fix this `-32603` and the codebase deliberately delegates twin cleanup to the operator-driven `meho.connector.delete` surface (#1700). Not done here.
- **Sibling unmapped-exception gap.** The other five curation mutators (`edit_op` / `edit_group` / `enable` / `disable` / `delete`) resolve a single literal scope and would surface a `ConnectorNotFoundError` as `-32603` too (also unmapped on MCP). Not the reported bug; noted for a possible follow-up.

Closes #1910


damir-topic:
@coderabbitai review

Issue #1363, opened 2026-05-30
Task: G0.18-T10 Helm chart — first-class secret-managed wiring for the agent-runtime config (ANTHROPIC_API_KEY + KEYCLOAK_ADMIN_*)


damir-topic:
Parent initiative: #1353

## Problem (SEV-3 deploy/devops — chart gaps behind RDC #789's deploy-config note)

The Helm chart provides **no first-class, secret-managed way to set the two config groups the G11 agent runtime requires** — so agents / approvals / grants / agent-principal registration cannot be enabled on a Helm deploy without an operator hand-rolling Secrets + `extraEnv` `valueFrom`.

**(a) `ANTHROPIC_API_KEY`** — the G11.1 agent LLM loop (`backend/src/meho_backplane/settings.py:530`, fail-closed at `agent/run.py:432`; `:678` notes it should be a secret/sealed-secret).

**(b) `KEYCLOAK_ADMIN_*`** — `KEYCLOAK_ADMIN_URL` / `KEYCLOAK_ADMIN_CLIENT_ID` / `KEYCLOAK_ADMIN_CLIENT_SECRET` (`settings.py:208-223,725-727,1134-1136`) configure the `KeycloakAdminClient` (`auth/keycloak_admin.py`) that gates **agent-principal registration** (G11.2 RBAC, `auth/agent_principals.py`). The feature is "configured iff all three non-empty" (`features.py:78-92`); unset → fail-open 503. `KEYCLOAK_ADMIN_CLIENT_SECRET` is a secret (`Field(repr=False)`).

Current state (verified 2026-05-30):
- `deploy/charts/meho/` has **zero** `anthropic` references and **zero** `keycloak_admin` / `KEYCLOAK_ADMIN_*` references.
- The chart wires the **other** operator secrets first-class — Keycloak *client* secret + Postgres creds via ExternalSecret + `secretKeyRef` (`templates/externalsecrets.yaml`; `templates/deployment.yaml:47-57`; `eso.keycloak`, `keycloak.clientSecretSyncFromVault`) — but neither the Anthropic key nor the Keycloak *admin* creds.
- The only path today for either is the generic plaintext `extraEnv: []` (`values.yaml:590`) — the wrong pattern for an API key / admin client secret.

[...]

Issue #956, opened 2026-05-22
Initiative: G0.11 CI + test-infra + release-tooling hardening — collect the flat CI/test/release tasks under one trackable parent


damir-topic:
Parent goal: #221

## Scope

Home for the CI + test-infrastructure + release-tooling hardening tasks that
were filed flat (no parent initiative) and were therefore invisible to every
Goal/version rollup. Surfaced by the 2026-05-22 board audit
([release-plan.md](docs/planning/release-plan.md)) and extended 2026-05-26
with release-tooling fixes from the v0.6.0 release cut (the `/release` skill
exposed 3 latent bugs in Phase 2/3/4; the goreleaser draft default required
a manual `gh release edit --draft=false` to publish). This initiative groups
them so the work is trackable and the
[G0.10 #949](https://github.com/evoila/meho/issues/949) "every task
resolves to a parent" invariant holds.

**Area:** CI workflows ([.github/workflows/](.github/workflows/)), pytest/xdist
config, SonarCloud config, test-runner topology, **release tooling** (the
`/release` skill body at `.claude/skills/release/SKILL.md`,
[docs/RELEASING.md](docs/RELEASING.md), [cli/.goreleaser.yaml](cli/.goreleaser.yaml)).
No product code.

## Why

CI throughput + merge-gating directly gate team velocity (the unit-job
10-min budget, the merge-train inherited-red problem). Release-tooling defects
gate publish quality (the v0.5.0 incident shipped release notes omitting
~24 PRs; the v0.6.0 cut nearly mis-tagged a post-merge race commit). These
tasks were real work hiding outside the hierarchy; collecting them makes the
cost visible and schedulable.

**Status:** All child tasks and Definition-of-done items are shipped — the CI
hardening (#349, #1622, #1983) and the FastAPI-0.137 unpin (#1819) all merged
by 2026-06-24, and the release-tooling fixes (#1126/#1127) landed earlier. This
Initiative appears fully delivered; recommend closing it.

## Definition of done

- [x] Each child task resolves to this initiative (no longer orphaned).
- [x] CI hardening items are either shipped or explicitly deferred with rationale.
- [x] Release-tooling hardening items are either shipped or explicitly deferred with rationale.

## Execution tasks

- [x] #698 — make `Python (integration testcontainers)` a required merge gate
- [x] #761 — dedicated meho-runners-ci-heavy scale set + -n 6 for python-lint
- [x] #769 — adopt GitHub merge queue to end merge-train inherited-red
[...]

PR #2052, opened 2026-06-23
chore(ci): harden helm-test job — remove continue-on-error, promote to required check (#349)

### LINKED ISSUE #349: Harden chart helm-test workflow: remove continue-on-error: true after ~10 green runs (G0.4-T6 follow-up)
Parent goal: #221
Parent initiative: #956

## Background

#432 (G6.1-T8) landed the underlying hardening fixes for the `helm-test` workflow job in `.github/workflows/chart.yml` (kind-load the locally-built `meho:test` image so PR builds stop hitting `ImagePullBackOff`; enable the broadcast/Valkey subchart) but explicitly **deferred its AC#4** — removing `continue-on-error: true` and promoting the job to a required check — behind a "≥10 consecutive green runs" gate. This issue materializes that deferred AC#4.

The `helm-test` job (`.github/workflows/chart.yml:303`) spins up a kind cluster, installs `pgvector/pgvector:pg16` as an in-cluster Deployment+Service, helm-installs the meho chart with `--wait-for-jobs`, and runs `helm test meho-test` to assert the `templates/tests/test-pgvector.yaml` preflight Pod succeeds.

The job ships with `continue-on-error: true` (`.github/workflows/chart.yml:344`) as the v0.2 transitional posture — the in-workflow comment at line 329 records this as "per G6.1-T8 AC". It runs on every chart-affecting PR + push and surfaces failures to operators, but does not block merge while the kind/k3d wiring nuances (image-pull caching, runner-side Docker permissions, helm hook timing) settle now that #432's underlying fixes have landed.

## Scope

- Watch the next ~10 chart-affecting PR + push runs of the `helm-test` job.
- Diagnose and fix any non-trivial flakes (timeouts on `--wait-for-jobs`, kind pull-rate issues, `helm test` Pod startup races).
- Once 10 consecutive runs are green without intervention, flip `continue-on-error: true` to `false` at `.github/workflows/chart.yml:344` so the job becomes a real merge gate, and promote it to a required check on `evoila/meho`'s `main` branch protection (per #432 AC#4).
- Update the `docs/codebase/backend.md` row for `tests/test-pgvector.yaml` (line 812) to remove the transitional-posture note, and update the broadcast-chaos row (line 638) that references the `continue-on-error: true` lane "until the gating-posture flip (#432 follow-up)".

## Acceptance criteria

- [ ] 10 consecutive green `helm-test` runs documented (issue body comments OK).
- [ ] Any flake root causes fixed in-tree (not silenced via retries unless the upstream is genuinely flaky).
- [ ] `continue-on-error: true` removed from the `helm-test` job in `.github/workflows/chart.yml` (line 344) and the job promoted to a required check on `main`.
- [ ] `docs/codebase/backend.md` updated (the `tests/test-pgvector.yaml` row and the broadcast-chaos row's gating-flip note).

## Out of scope

- Extending the helm-test coverage to additional preflight checks (e.g. broadcast/Valkey from #335) — separate issue per chart-touching pillar.
- Switching from kind to k3d / minikube / GH-managed runners — kind is the v0.2 choice.

## References

- Predecessor: #432 (G6.1-T8) — landed the image-load + broadcast-subchart hardening; deferred its AC#4 (the gating-posture flip) to this follow-up. The original chart-preflight task is G0.4-T6 (#263, closing PR #347), which first introduced the `continue-on-error: true` posture.
- Workflow file: `.github/workflows/chart.yml` (the `helm-test` job at line 303; `continue-on-error: true` at line 344).
- Helm test Pod: `deploy/charts/meho/templates/tests/test-pgvector.yaml`.

ddzafic:
## Summary

Hardens the chart `helm-test` job into a **real, fail-closed required merge gate** (#349 AC#3), without blocking unrelated PRs.

What landed (evolved during review):

1. **Remove `continue-on-error: true`** from the `helm-test` job in `.github/workflows/chart.yml` + its transitional-posture comment block.
2. **Drop the `on.pull_request.paths` filter** on `chart.yml`. A workflow skipped by a path filter leaves its required check permanently *Pending* (blocks merge); a job skipped by a job-level `if:` reports *Success*. So the workflow now always triggers and gates at the job level instead.
3. **Add a `changes` job** (plain `gh api`, no third-party action) that detects chart-relevant changes; `validate` + `helm-test` gate on `needs.changes.outputs.chart`.
4. **Add an `always()` `helm-test-gate` aggregator** that owns the required branch-protection context and explicitly evaluates `needs.*.result`, so the gate **fails closed** when `changes`/`validate`/`helm-test` fail or are skipped (a skipped required job otherwise reports Success — GitHub "Troubleshooting required status checks" / actions/runner#2566). The kind-running job is renamed `…[run]`; the required-context string is unchanged → no branch-protection edit needed.
5. **Docs**: `docs/codebase/backend.md` (test-pgvector.yaml + broadcast-chaos rows → hardened posture) and `docs/codebase/devops.md` (the `changes`-job gating rationale).

## Behaviour matrix (PR events)

| Case | `changes` | `validate` | `helm-test` | **gate (required)** |
|---|---|---|---|---|
| Non-chart PR | success (chart=false) | skipped | skipped | **pass** (no kind) |
| Chart PR, all green | success (chart=true) | success | success | **pass** |
| Chart PR, helm-test red | success | success | failure | **fail** |
| Chart PR, validate red | success | failure | skipped | **fail** |
| `changes` job blip | failure | skipped | skipped | **fail** |

## Test plan

- [x] `continue-on-error: true` removed; no stale references in workflow/docs.
- [x] Required context string `helm install + helm test (pgvector preflight)` preserved (now produced by `helm-test-gate`) — branch protection unchanged.
- [x] On this (chart-touching) PR: `Detect chart-relevant changes`, `Lint + template + kubeconform`, `…(pgvector preflight) [run]`, and the `…(pgvector preflight)` gate all run.
- [x] `docs/codebase/backend.md` + `docs/codebase/devops.md` updated.

## Out of scope

- Extending helm-test coverage to additional preflight checks (broadcast/Valkey) — separate issue per chart pillar.
- Switching kind → k3d / minikube.

Closes #349

## Summary by CodeRabbit

* **Chores**
  * Hardened CI chart validation by deterministically running and enforcing the required chart test only when chart-related changes are detected.
  * Removed transitional non-blocking behavior so failed chart tests now fail the CI check normally.
  * Added an explicit gating step to ensure required status checks behave correctly and don’t get incorrectly marked as passing.
* **Documentation**
  * Updated CI/chart-CI guidance to describe the new gating behavior and blocking expectations for chart tests.


ddzafic:
@copilot review

.github/workflows/chart.yml
```diff
@@ -341,7 +334,6 @@ jobs:
     needs: validate
     runs-on: ubuntu-latest
     timeout-minutes: 15
```
ddzafic, line 394: **[🛑 Blocker] — Promoting this path-filtered job to a required check blocks every non-chart PR**

**What:** This `helm-test` job lives in `chart.yml`, which is path-filtered at the **workflow level** (`on.pull_request.paths: ['deploy/charts/meho/**', '.github/workflows/chart.yml']`). The check name `helm install + helm test (pgvector preflight)` was added to `main`'s required status checks (already live via `gh api`). On any PR that doesn't touch those paths, the workflow never triggers, the check never reports, and the PR is blocked.

**Why it matters:** Per [GitHub Docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks): a workflow skipped via path filtering leaves its checks **Pending**, blocking merge — unlike a *job-level* `if:` skip, which reports Success. Sibling PRs #2053 and #2054 (no chart changes) are already affected.

**Suggested fix:** Drop the workflow-level `pull_request.paths` filter and gate the expensive steps with a change-detection job (`dorny/paths-filter`) + a **job-level** `if:` so non-chart PRs report Success; or add an always-running status fan-out job that reports this context as success when chart paths are untouched. Interim: revert the branch-protection promotion until a guard is in place so #2053/#2054 aren't blocked.

**Source:** GitHub Docs — Troubleshooting required status checks; `.github/workflows/chart.yml:52-57` (the `paths:` filter).

ddzafic:
## /review-pr-slim summary

**Anchor task:** #349 — harden chart `helm-test`: remove `continue-on-error`, promote to required check
**Surface area:** devops (CI workflow) + docs
**Verdict:** REQUEST_CHANGES

### Acceptance criteria
- [x] `continue-on-error: true` removed from the `helm-test` job — verified in diff (`.github/workflows/chart.yml`); no stale `continue-on-error` references remain in the workflow or `docs/`.
- [x] `docs/codebase/backend.md` updated — both the `test-pgvector.yaml` row and the broadcast-chaos row reworded to the hardened posture; accurate.
- [⚠️] Promoted to required check on `main` — **done, but as implemented it breaks merge for every non-chart PR.** See blocker below. Verified live: `helm install + helm test (pgvector preflight)` is in `main`'s required contexts now.
- [~] ≥10 consecutive green runs — claimed 20+ in the run; `helm install + helm test (pgvector preflight)` is green on this PR. Not independently re-verified against historical runs.

### The blocker (🛑)
`chart.yml` is **workflow-level path-filtered** (`on.pull_request.paths: ['deploy/charts/meho/**', '.github/workflows/chart.yml']`). Per [GitHub Docs — Troubleshooting required status checks](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks): *"If a workflow is skipped due to path filtering, branch filtering or a commit message, then checks associated with that workflow will remain in a 'Pending' state. A pull request that requires those checks to be successful will be blocked from merging."* (Job-level `if:` skips report Success — workflow-level `paths:` skips do **not**.)

The `helm install + helm test (pgvector preflight)` context is produced **only** by this path-filtered workflow (no status fan-out / always-pass guard exists). Promoting it to a required check therefore blocks every PR that doesn't touch `deploy/charts/meho/**` or `.github/workflows/chart.yml`.

This is not theoretical: the branch-protection change was applied out-of-band via `gh api` and is **already live**, and sibling PRs **#2053** and **#2054** (this same initiative) do not touch chart paths — the required check is absent on both and will block their merge once review clears.

**Suggested fix (any one):**
1. *(preferred)* Drop the workflow-level `pull_request.paths` filter and gate the expensive helm-test steps/job with a change-detection job (`dorny/paths-filter`) + a **job-level** `if:`. Job-level skips report Success, satisfying the required check on non-chart PRs while still actually running on chart PRs.
2. Add a companion "status fan-out" job in an always-running workflow that reports the `helm install + helm test (pgvector preflight)` context as success when chart paths are untouched.
3. **Interim, regardless of which fix lands:** revert the branch-protection promotion now to unblock #2053/#2054 and any other non-chart PR; re-apply once the guard is in place.

### Project-board alignment (initiative #956)
|---|---|---|---|---|---|
| #349 | In Progress | `Closes #349` (body) | ✅ same-repo, base main | ⚠️ code+docs yes; required-check promotion harmful as-is | close on merge **after** blocker resolved |
| #956 (Initiative) | open | parent | — | partial | no change — siblings still open |

### Predicted vs. actual diff
- ✅ `chart.yml`: clean removal of `continue-on-error` + its comment block; job structure intact (`needs`/`runs-on`/`timeout-minutes`/`steps`).
- ✅ `docs/codebase/backend.md`: both rows updated, internally consistent.
- 🕳️ Missing: a path-filter-aware guard for the now-required check (the change that makes "required" safe repo-wide).

### Integration verification
- [x] Live CI on this PR — all checks green, including `helm install + helm test (pgvector preflight)` = SUCCESS (the gate runs and passes because this PR touches `chart.yml`).
- [x] YAML structure — validated by CI (`Helm (lint + template + kubeconform)` + the workflow parsing & running).
- [x] Branch-protection state — queried `gh api .../required_status_checks`; confirmed the check is now required and absent on non-chart sibling PRs (#2053, #2054).
- [⚠️ not run] No local `act`/kind run — relied on live CI, which is authoritative here.

### Findings
- 🛑 1 blocker — required-check promotion blocks non-chart PRs (inline on the `helm-test` job).
- ⚠️ 0 majors
- 💡 0 minors
- ❓ 0 questions

### Docs
- [x] `docs/codebase/backend.md` — updated and accurate for the hardened posture.

### What was done well
- 👏 Clean, minimal removal — no dangling references, comment block removed with the flag.
- 👏 Doc rows updated in lockstep with the behavior change (often forgotten).
- 👏 Required-check promotion was actually applied (AC#3 frequently gets deferred and lost).

.github/workflows/chart.yml
```diff
@@ -326,22 +371,21 @@ jobs:
...
     # needs Docker-in-Docker support that the self-hosted runner
     # pool isn't configured for; the GitHub-hosted runners ship
     # Docker pre-installed and helm/kind-action works out of the box.
-    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
-    needs: validate
+    #
+    # This is the REQUIRED check on `main` (#349). The `needs.changes` clause
+    # skips it (→ reported Success) on PRs that don't touch the chart, so the
+    # required check is satisfied without running kind — see the `on:` note.
+    needs: [changes, validate]
```
zdamir, line 388: **[🛑 Blocker] — Required `helm-test` gate fails open when `changes` or `validate` fails**

**What:** `helm-test` carries the required-check name and has `needs: [changes, validate]` with a custom `if:` that contains no `always()`/status function. When a needed job (`changes` or `validate`) fails, GitHub **skips** `helm-test`, and a skipped job **reports its conclusion as Success to branch protection** — so the required gate is satisfied without the kind install / `helm test` pgvector preflight ever running.

**Why it matters:** #349's deliverable is for this job to "become a real merge gate." As written it isn't one under upstream failure. A chart PR whose `validate` step fails — a `helm template` / kubeconform regression, or the secretKeyRef assertions (#1363) that are unique to `validate` — skips `helm-test` to a green required check, letting the chart change merge un-tested (the unique runtime coverage this job exists to provide is exactly what gets bypassed). The same happens on a transient `changes` `gh api` blip, which has **no** backstop in `ci.yml`. GitHub documents this exact pitfall and prescribes `always()`.

**Required change:** Make the required context a job that *always runs* and explicitly evaluates upstream results, rather than letting a skipped `helm-test` masquerade as success. The standard aggregator pattern:

```yaml
  helm-test-gate:
    name: helm install + helm test (pgvector preflight)   # the required context moves here
    if: always()
    needs: [changes, validate, helm-test]
    runs-on: ubuntu-latest
    timeout-minutes: 5
    steps:
      - name: Evaluate gate
        run: |
          set -euo pipefail
          # Detection must succeed; a flaky `changes` must NOT pass the gate.
          [[ "${{ needs.changes.result }}" == "success" ]] || { echo "changes job failed"; exit 1; }
          if [[ "${{ needs.changes.outputs.chart }}" == "true" ]]; then
            [[ "${{ needs.helm-test.result }}" == "success" ]] \
              || { echo "helm-test did not pass on a chart PR"; exit 1; }
          fi
          echo "gate satisfied"
```

Rename the current job's `name:` to an internal label (e.g. `helm install + helm test (pgvector preflight) [run]`) and repoint the branch-protection required context to `helm-test-gate`'s name. On non-chart PRs the gate runs, sees `chart != 'true'`, and passes without kind; on chart PRs it fails closed when detection or the test itself didn't succeed. (Note: the required-context rename means the branch-protection promotion must be re-applied to the new name.)

**Source:** https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks ("use the `always()` conditional expression in addition to `needs` … to avoid unintended merge allowances"); https://github.com/actions/runner/issues/2566 ("Required status check job is skipped when one of the dependent job fails").

zdamir (changes requested):
## /review-pr-slim summary

**Anchor task:** #349 — harden chart `helm-test`: remove `continue-on-error`, promote to a real required merge gate
**Surface area:** devops (CI workflow) + docs
**Verdict:** REQUEST_CHANGES

The prior review's blocker (path-filtered workflow → required check stuck pending on non-chart PRs) has been **resolved** in the current head: the `on.pull_request.paths` filter is gone, a `changes` detection job was added, and the expensive jobs are gated by a job-level `if:`. That mechanism is correct and doc-verified — a job skipped via a conditional reports **Success**, satisfying the required check, whereas a path-skipped *workflow* stays Pending. One new robustness blocker remains in that same redesign.

### Acceptance criteria
- [x] `continue-on-error: true` removed from `helm-test` — verified in diff; no stale chart-helm-test `continue-on-error` references remain in the workflow or `docs/`.
- [x] `docs/codebase/backend.md` updated — both the `test-pgvector.yaml` row and the broadcast-chaos row reworded to the hardened posture; accurate.
- [⚠️] Promoted to a *required check* on `main` — the context is live in branch protection, but the gate **fails open** under upstream-job failure (see blocker). The issue's stated goal is "a real merge gate"; that is not provably satisfied as written.
- [ ] "≥10 consecutive green `helm-test` runs documented (issue body comments OK)" — **not documented on #349** (the issue has zero comments). The PR body asserts "20+", but the AC asks for it recorded on the issue.

### The blocker (🛑) — fail-open required gate
`helm-test` (the required context) has `needs: [changes, validate]` with a custom `if:` lacking `always()`. If `changes` or `validate` fails, GitHub skips `helm-test`, and a skipped required job reports **Success** to branch protection — the gate passes without the kind install / pgvector preflight running. `validate`'s unique secretKeyRef assertions (#1363) and the runtime install coverage are exactly what gets bypassed; a transient `changes` `gh api` blip has no `ci.yml` backstop at all. Fix = aggregator gate job with `if: always()` that explicitly evaluates `needs.*.result`. Full fix sketch inline on the `helm-test` job. Sources: GitHub "Troubleshooting required status checks"; actions/runner#2566.

### Project-board alignment
|---|---|---|---|---|---|
| #349 | Open | `Closes #349` (body) | ✅ same-repo, base `main` | ⚠️ literal ACs yes; "real merge gate" fails open | resolve blocker, then close on merge |

> Operational note (not a diff defect): the required-check promotion is **already live** on `main`, but this fix is not merged yet — so today every non-chart PR is blocked. Confirmed on sibling **#2054** (touches only `backend/**` + `docs/`): the `helm install + helm test (pgvector preflight)` context is absent from its rollup and it sits `BLOCKED`; **#2053** likewise. Merging this PR is what unblocks them — see handoff for the recommended interim.

### Predicted vs. actual diff
- ✅ `chart.yml`: path filter removed; `changes` detection job added; `validate` + `helm-test` gated by job-level `if:`; `continue-on-error` + its comment block removed; push/tag path (`changes`→`validate`→`publish`→`verify`) intact.
- ✅ `docs/codebase/backend.md`: both rows updated, internally consistent.
- ✅ `docs/codebase/devops.md`: the "why no image build job" paragraph accurately documents the new `changes`-job gating rationale.
- ⚠️ `helm-test` `needs`/`if:` introduces the fail-open gap above — see inline.

### Integration verification
- [x] Live CI on this PR (chart PR) — `Detect chart-relevant changes`, `Lint + template + kubeconform (PR build gate)`, and `helm install + helm test (pgvector preflight)` all **SUCCESS**.
- [x] Branch-protection state — queried `repos/evoila/meho/branches/main/protection`; `helm install + helm test (pgvector preflight)` is in the required contexts now.
- [x] Non-chart sibling state — #2054 confirmed `BLOCKED` with the required context absent (the live impact of promoting before this fix merged).
- [x] GitHub skip semantics — verified against official docs + actions/runner#2566 (workflow-skip → Pending/blocks; job-skip → Success; needs-failure skip → Success = the fail-open).
- [⚠️ not run] No local `act`/kind run, and **the fix's non-chart-PR skip→Success path is not yet empirically exercised** (no non-chart PR exists on this branch's workflow). Reasoned + doc-verified only; recommend confirming on the first non-chart PR after merge.

- 🛑 1 blocker — required gate fails open on `changes`/`validate` failure.
- ⚠️ 0 majors · 💡 0 minors · ❓ 0 questions

- The PR body is now **stale**: it describes only the original simple approach ("remove `continue-on-error`… promote via `gh api` after PR opens") and omits the substantial `changes`-job redesign and the `docs/codebase/devops.md` edit. Worth refreshing so the body matches the diff a reviewer actually sees.

### Docs
- [x] `docs/codebase/backend.md` — updated and accurate.
- [x] `docs/codebase/devops.md` — updated and accurate (the chart-CI gating rationale).
- No stale `transitional posture` / `gating-posture flip` / chart `continue-on-error` references remain.

### What worked well
- The redesign is the textbook fix for the prior path-filter blocker — `on:` note and `changes`-job comments correctly explain *why* workflow-skip blocks but job-skip passes.
- Docs updated in lockstep with the behavior change, including the `devops.md` rationale paragraph (commonly forgotten).
- Push/tag publish DAG left fully intact while restructuring the PR-path gating.

ddzafic:
Thanks @zdamir — the fail-open blocker is fixed in `7ab94cd0`.

**Blocker (fail-open required gate):** Implemented the aggregator pattern you sketched. The required context `helm install + helm test (pgvector preflight)` now lives on a new `helm-test-gate` job (`if: always()`, `needs: [changes, validate, helm-test]`) that explicitly evaluates `needs.*.result`:
- `changes` not `success` → **fail** (a transient `gh api` blip can't pass the gate).
- chart PR & `validate` not `success` → **fail** (the #1363 secretKeyRef + lint coverage is now enforced).
- chart PR & `helm-test` not `success` → **fail**.
- non-chart PR → **pass** without kind.

The kind-running job is renamed `…(pgvector preflight) [run]`; the required-context string is unchanged, so **no branch-protection edit is needed**. Behaviour matrix is in the (now-refreshed) PR body.

**AC#1 (≥10 green runs documented):** recorded on the issue — evoila/meho#349 (comment): 10 completed `helm-test` runs, 10 green, 0 failures (the rest are concurrency-`cancelled`, not flakes).

**Stale PR body:** rewritten to describe the `changes`-job + `helm-test-gate` redesign and the `devops.md` edit.

PR #2020, opened 2026-06-21
fix(connectors): honour RFC6570 reserved expansion in ingested-op path substitution (#2003)

damir-topic:
## Summary

- The ingested-op path substituter percent-encoded **every** reserved char unconditionally (`urllib.parse.quote(safe="")`), so an RFC6570 reserved expansion `{+var}` mangled its slash-delimited value into `%2F`-soup. This blocked vRLI events constraint queries (`/api/v2/events/{+constraints}` with `text/CONTAINS error/...`) — the appliance 400s when the structural slashes arrive encoded.
- `_substitute_path` / `_PATH_VAR_RE` now parse the RFC6570 expression operator: **simple expansion** `{var}` (§3.2.2) keeps `safe=""` (reserved chars encoded, OpenAPI default path-style); **reserved expansion** `{+var}`/`{#var}` (§3.2.3) quotes with the reserved/gen-delim safe set so structural chars pass through literal. Genuinely-unsafe chars (space, control) stay encoded in both arms.
- Also fixes a **latent KeyError**: the regex previously captured the operator into the lookup name, so a literal `{+path}` spec would `KeyError` even with the param supplied. The operator is now stripped from the lookup name.
- Hand-rolled — no `uritemplate` dependency added (templates are single-segment; lower blast radius).

## Test plan

- [x] `ruff check` — clean
- [x] `ruff format --check` — clean
- [x] `mypy src/.../_branches.py` — clean
- [x] `pytest tests/test_operations_branches_substitute_path.py` — 6 passed (unit: `{+var}` keeps `/` literal while `{var}` encodes it; `{+path}` resolves the `path` param with no KeyError; space stays `%20` under reserved expansion; missing param still raises)
- [x] `pytest tests/test_connectors_vcf_logs_e2e.py` — 24 passed, incl. new `test_vrli_e2e_reserved_constraint_keeps_slash_literal_on_wire` (non-empty reserved-char constraint dispatched through the full `call_operation` stack hits the literal-slash respx route)
- [x] `pytest tests/test_operations_request_preview.py tests/test_operations_dispatcher.py tests/test_connectors_vcf_logs_core_ops.py` — no regression
- [x] `code-quality.py --diff` — 0 blockers (2 pre-existing warnings on `_branches.py` size, not introduced here)
- [x] No FastAPI route/schema changed → no OpenAPI snapshot regen needed (verified: diff touches no routers/schemas)

## Out of scope

[...]

Issue #2003, opened 2026-06-20
Task: ingested-op path substitution over-encodes RFC6570 reserved chars in {+var} segments — blocks vRLI events constraint queries


damir-topic:
## Summary

The vRLI events op `GET:/api/v2/events/{…}` carries path-segment constraints (e.g. `text/CONTAINS error/hostname/CONTAINS vcsa`), but meho's ingested-op path substitution **percent-encodes every reserved char unconditionally** (`/` → `%2F`, space → `%20`), so the constraint arrives mangled and vRLI 400s. The empty constraint (recent events) works — which is what a basic sentry needs — but richer path-constraint queries are blocked. The fix is to honour RFC6570 reserved expansion (`{+var}`) for path-segment values instead of blanket-quoting.

## Current state

- `_substitute_path` ([`operations/_branches.py:132-149`](backend/src/meho_backplane/operations/_branches.py#L132-L149)) blanket-quotes every placeholder value with `urllib.parse.quote(str(value), safe="")` (L147), and its regex `_PATH_VAR_RE = re.compile(r"\{([^{}]+)\}")` ([`_branches.py:63`](backend/src/meho_backplane/operations/_branches.py#L63)) makes **no distinction** between RFC6570 simple expansion `{var}` (reserved chars encoded) and reserved expansion `{+var}` (reserved chars literal). Substitution is hand-rolled — no `uritemplate`/RFC6570 library is a dependency.
- Re-diagnosis vs the dogfood note: meho currently models the op as a plain `{constraints}` simple placeholder (not `{+path}`), so there is no `+` operator for the code to be "ignoring" today — the real defect is the unconditional `quote(safe="")`. Two coupled defects share the fix: (a) blanket reserved-char encoding, and (b) a **latent KeyError** — were a literal `{+var}` spec ever ingested, `_PATH_VAR_RE` would capture `+path` (operator included) while the ingest pipeline names the param `path`, so substitution would `KeyError` → `invalid_params` before the wire.

## Desired state

Teach `_substitute_path` (and `_PATH_VAR_RE`) the RFC6570 reserved-expansion operator:

1. Extend the regex to capture an optional leading operator + the bare name, e.g. `re.compile(r"\{([+#./;?&]?)([^{}]+)\}")`; strip the operator from the lookup name so `{+constraints}`/`{+path}` resolves the param keyed `constraints`/`path` (also fixes the latent KeyError).
[...]

Issue #1796, opened 2026-06-16
Connector dispatch drops the OpenAPI servers[].url base path — ops 404 when the spec uses a relative server base


damir-topic:
Parent goal: #221
Parent initiative: #2065

## Summary

Connector dispatch drops the OpenAPI `servers[].url` base path. For any ingested spec whose `servers` declares a **relative base** (e.g. the vRLI / VCF Operations for Logs spec: `servers: [{ "url": "/api/v2" }]`), every op 404s — meho builds the request URL as `target_host:port + op_path` and never prepends the server base. A compliant OpenAPI client resolves an op URL as `server.url + path` (`/api/v2` + `/version` = `/api/v2/version`).

This is the **third (and last) meho-side fix for a durable log-sentry** (#1135/#1139), alongside the two that landed under this initiative — #2066 (RFC6570 `{+path}` ingest) and #2067 (session-expiry dispatch recovery). With those merged, the relative-base drop is the remaining gate to retiring the consumer's `/api/v2`-path-baking overlay.

## Current state (grounded at HEAD)

- **Ingest never reads `servers`.** [`operations/ingest/openapi.py`](backend/src/meho_backplane/operations/ingest/openapi.py) has no `servers[].url` handling — the only mention is a docstring (`openapi.py:547`). The op is ingested with its bare spec path (`GET:/version`), losing the base.
- **No field stores the base.** [`EndpointDescriptor`](backend/src/meho_backplane/db/models.py#L1299) (and `EndpointDescriptorProto`, `operations/ingest/schemas.py:27`) carry no `base_path` / `server_base` column, so a dispatch-time prefix has nowhere to read from today.
- **Dispatch builds `host:port + path`.** So `GET:/version` on `vrli-rest-9.0.2` → `https://10.5.52.21:9543/version` → 404 (real endpoint `…/api/v2/version`). Confirmed by the consumer on v0.19.0.
- The typed `VcfLogsConnector` applies `/api/v2` for its own session/events path via connector logic, which is why those worked despite this bug — but every *generic-ingested* op (e.g. `/version`) 404s.

## Desired state

[...]

Issue #2065, opened 2026-06-24
Initiative: G0.29 · v0.19.0 closed-loop dogfood hardening — 5 findings from the RDC Hetzner-DC v0.19.0 deploy (session-expiry dispatch recovery [vCenter 401 + vRLI 440] + {+path} ingest regression + migration-0049 stub-twin gap + UI datalist URL-rewrite + ingest SSRF on-ramp DX)


damir-topic:
## Summary

Closed-loop dogfood hardening for the **v0.19.0** deploy on the RDC `claude-rdc-hetzner-dc` lab (`meho.evba.lab`, rke2-infra; helm REV 47, chart 0.19.0, git_sha `077740220a`, db.revision `0050`, smoke 6/6). This is the per-cycle G0.x hardening Initiative (series G0.8 → G0.28 #1998), collecting the source-grounded, `evoila/meho`-actionable findings from the v0.19.0 dogfood.

Every finding was **re-diagnosed against the shipped v0.19.0 source** before filing (the RDC reports get symptoms right but frequently mis-attribute root cause and prior art — this cycle was no exception: see the corrections table below). The headline result was a single architectural defect surfacing as two separate "red" regressions: **session-expiry re-login was not wired into the generic-ingested dispatch path**, so both vCenter (401) and vRLI (440) reads hard-failed once their in-process session expired server-side and recovered only on a backplane restart. This is now **FIXED** by #2067 (merged PR #2072), which wires session invalidate + retry-once into the dispatcher's `HTTPStatusError` arm.

## Confirmed FIXED on this deploy (no action — acknowledged)

- **`ask_docs` synthesis leg (#1999 / merged #2017)** — the v0.18.0 headline blocker. `POST /api/v1/ask_docs` now returns HTTP 200 with a grounded, cited answer.
- **Operator console (Goal #336)** — 5 `ui-surface` findings verified fixed via a live Playwright pass (truthful footer version, feed no longer 401-loops, broadcast Alpine registers, `/ui/memory` create CSRF double-submit, scope-default TTL on the UI write path).
- **`catalog ingest vmware/9.0` (#1976)** — profile-backed ingest works; `connector_registered=false` is the expected #1971 staged-by-design property.
- **`SchedulerVaultBrokerError`** is now an actionable message naming the `VAULT_SCHEDULER_TOKEN` policy gap (deploy-config remedy on the consumer side).
[...]

PR #2071, opened 2026-06-24
fix(connectors): key ingested {+path} params on the bare name (#2066)

### LINKED ISSUE #2066: Task: RFC6570 {+path} path params undispatchable — ingest keys the schema property on the raw operator-bearing name while the renderer (PR #2020) looks up the bare name; preview_operation 500s
## Summary

An ingested OpenAPI op whose path-parameter **name carries an RFC6570 expression operator** — e.g. the VCF Operations-for-Logs spec declares `{"name": "+path", "in": "path"}` for the template `/events/{+path}` — is **undispatchable** at v0.19.0. PR #2020 (closing #2003) taught the path *renderer* about RFC6570 operators but left the *ingest* side keying the parameter-schema property on the raw operator-bearing name, so the two halves disagree on the property key and no caller param-set satisfies both. Secondary defect: `preview_operation` raises an uncaught `KeyError` (→ MCP `-32603`) instead of a structured error, because the preview path has no exception mapping. Both vRLI events ops (`/events/{+path}`, `/aggregated-events/{+path}`) — the log-sentry read path (#1135/#1139) — are dead.

This is the **un-landed ingest half of #2003**, not a reopen: the renderer half PR #2020 scoped did merge correctly.

## Current state

- **Ingest keys the schema property on the raw declared name.** [`operations/ingest/openapi.py`](backend/src/meho_backplane/operations/ingest/openapi.py#L789): `name = resolved.get("name")` (no operator normalization), then [`openapi.py:802`](backend/src/meho_backplane/operations/ingest/openapi.py#L802) `properties[name] = prop_schema` and the `required.append(name)` that follows. For a `{"name":"+path"}` param this registers the jsonschema property under **`+path`** (operator included), so validation *requires* `+path`. `descriptor.path` is stored verbatim ([`openapi.py:713`](backend/src/meho_backplane/operations/ingest/openapi.py#L713)), preserving the `{+path}` template operator.
- **The renderer strips the operator and looks up the bare name.** PR #2020 changed `_PATH_VAR_RE` to `\{([+#./;?&]?)([^{}]+)\}` ([`operations/_branches.py:75`](backend/src/meho_backplane/operations/_branches.py#L75)) and made `_substitute_path._replace` resolve `path_params[<bare name>]` ([`_branches.py:180-184`](backend/src/meho_backplane/operations/_branches.py#L180-L184)) — i.e. it looks up `path`, not `+path`. Its comment asserts ingest registers the bare name — **false** for any spec whose param `name` includes the operator.
- **Net dead-op:** schema validation demands `+path`; the caller therefore supplies `+path`; `_split_ingested_params` copies the caller key verbatim into `path_params["+path"]` ([`_branches.py:107-111`](backend/src/meho_backplane/operations/_branches.py#L107-L111)); `_substitute_path` then strips the operator and looks up `path_params["path"]` → `KeyError("path template requires 'path' but it was not supplied")` ([`_branches.py:181-182`](backend/src/meho_backplane/operations/_branches.py#L181-L182)). `{"path": …}` instead fails schema validation (`'+path' is a required property`). No single param-set works.
- **Preview hard-500s.** `resolve_ingested_request` is called with no `try/except` in the preview path ([`operations/_request_preview.py:226`](backend/src/meho_backplane/operations/_request_preview.py#L226)); the file has zero `except` clauses, so the `KeyError` escapes despite the never-raises docstring ([`_request_preview.py:303-306`](backend/src/meho_backplane/operations/_request_preview.py#L303-L306)) → MCP `-32603`. The **execute** path swallows the same `KeyError` via the generic `except Exception` at [`operations/dispatcher.py:1025`](backend/src/meho_backplane/operations/dispatcher.py#L1025), so the two surfaces diverge.
- **Regression vs v0.18:** pre-#2020 the old regex `\{([^{}]+)\}` captured `+path` literally and matched `path_params["+path"]` (it dispatched, merely over-encoding `/`); #2020 turned that into a hard `KeyError`. #2003's own desired-state item 1 anticipated this exact gap ("strip the operator from the lookup name … also fixes the latent KeyError") but the PR implemented only the renderer-lookup strip. The #2020 tests mask it by hand-seeding descriptors keyed on the bare `constraints` name (`backend/tests/acceptance/_vrli_canary_fixtures.py`), never exercising real ingest of an operator-bearing name.

## Desired state

1. **Normalize at ingest (the missing half).** In the parameter-schema builder ([`openapi.py:802`](backend/src/meho_backplane/operations/ingest/openapi.py#L802)/`required`), strip a single leading RFC6570 operator (one of `+#./;?&`) from the property key when the param is `in: path`, so the property is keyed on the bare name (`path`) — matching the renderer. **Keep `descriptor.path` verbatim** (the renderer needs the operator in the template to select the reserved-expansion safe set, so `/` stays literal). Guard against a spec that declares both `path` and `+path` path params — fail that op's ingest loudly rather than silently overwriting one dict key. Optionally emit an operator-review warning (the parser already skips malformed entries quietly at `openapi.py:792-796`) so operators see the vendor spec used a non-conformant param name rather than it being silently rewritten.
2. **Honour the preview never-raises contract.** Wrap the `resolve_ingested_request` call in the **preview-only** callers (`_build_ingested_preview` / `preview_dispatch`, around [`_request_preview.py:226`](backend/src/meho_backplane/operations/_request_preview.py#L226)), catching `KeyError`/`RuntimeError` specifically (the two path-substitution faults it documents) and mapping to a structured `invalid_params` envelope — **not** a blanket `except`. Do **not** change `resolve_ingested_request` itself: it is shared verbatim with `dispatch_ingested` and deliberately raises so the dispatcher's structured-error mapping (`dispatcher.py:1025`) fires; converting it to swallow would change the execute-path contract.

## Acceptance criteria

- Ingesting a spec whose path-param declares an operator-bearing name (`name:"+path"`, `in:"path"`, template `/v1/{+path}`) yields a `parameter_schema` property keyed on the **bare** name `path` (not `+path`), with `x-meho-param-loc: "path"`.
- `call_operation` on such an op dispatches successfully — the path substitutes via reserved expansion (`/` stays literal) with no `KeyError`.
- `preview_operation` on such an op returns a structured `status=error` envelope (`invalid_params`/`dispatch_error`) — never an uncaught exception / MCP `-32603` — including for a genuinely missing path param.
- A spec declaring **both** `path` and `+path` path params fails ingest loudly rather than silently colliding on the dict key.
- A new **ingest-level** test exercises the real parameter-schema builder (not a hand-keyed seed descriptor) over an operator-bearing param name, proving the key normalization; plus an end-to-end `call_operation` + `preview_operation` test.
- `cd cli && make snapshot-openapi && make generate` run if any route/schema changed (expected no delta — internal substitution + ingest keying; verify).

## Out of scope

- The renderer-side reserved-expansion encoding itself (shipped by #2020) — unchanged.
- Adding a `uritemplate` dependency — the hand-rolled single-segment substitution stays (per #2003).
- #1796 (servers base-path) — a separate dispatch defect on the same vRLI ops, tracked under #1796.

## References

- Incomplete-fix follow-up to: #2003 / PR #2020 ({+var} renderer half — desired-state item 1 anticipated this exact ingest gap).
- Log-sentry read path: #1135 / #1139; related still-open dispatch defect on the same ops: #1796.
- Config-driven connectors: Goal #1964 / Initiative #1965.
- Best-practices anchors — *python*: correctness & spec-conformance (key ingest + render on the same normalized name); fail-loud on ambiguity (the `path`/`+path` collision guard); honour documented contracts (the preview never-raises promise); testing — exercise the real ingest path, not a hand-keyed seed that masks the gap.

Parent goal: #221
Parent initiative: #2065

damir-topic:
## Summary

- Lands the **un-landed ingest half of #2003** (the renderer half, PR #2020, already merged). An ingested OpenAPI op whose path parameter declares a leading RFC6570 expression operator in its name — e.g. VCF Operations-for-Logs declares `{"name": "+path", "in": "path"}` for `/events/{+path}` — was undispatchable at v0.19.0 because ingest keyed the JSON-Schema property on the raw `+path` while the renderer looks the value up by the bare `path`. No caller param-set satisfied both halves (both vRLI events ops were dead reads).
- **Ingest now keys an `in: path` param on the bare name** (operator stripped), matching the renderer. `descriptor.path` keeps the operator verbatim so the renderer still selects reserved expansion (the value's `/` stays literal on the wire).
- **The operator set is now a single shared constant** (`operations/_rfc6570.RFC6570_PATH_OPERATORS`) consumed by both the renderer regex and the ingest strip — so the ingest key and the render lookup can't drift again (the root cause of this bug class).
- **A spec declaring both `path` and `+path`** path params fails ingest loudly (`InvalidSchemaError`) instead of silently dropping one.
- **`preview_operation` honours its documented never-raises contract again**: the preview path catches the `KeyError`/`RuntimeError` `resolve_ingested_request` raises and maps them to a structured `dispatch_error` envelope, instead of letting the exception escape as MCP `-32603` / HTTP 500. `resolve_ingested_request` itself is unchanged (the execute path relies on the dispatcher's structured mapping), so the two surfaces stay aligned via their own wrappers.

## Test plan

- [x] `ruff format --check src/ tests/` — clean (1128 files)
- [x] `mypy src/` — clean (563 source files)
- [x] Full CI unit lane (`pytest -n3 --dist loadscope --ignore=tests/integration tests/`) — **8746 passed, 195 skipped**
- [x] Blast-radius sweep (dispatch + ingest + preview + connector e2e + MCP + UI) — 544 passed
- [x] **AC**: real `parse_openapi` over a `{+path}` spec keys the property on bare `path` with `x-meho-param-loc: "path"` (`test_parse_operator_bearing_path_param_keyed_on_bare_name`)
- [x] **AC**: `call_operation` on a `/v1/{+path}` op dispatches; the wire path is `/v1/events/CONTAINS%20error` (slash literal, space encoded) — no `KeyError` (`test_call_operation_dispatches_operator_bearing_path_param`)
- [x] **AC**: `preview_operation` on a missing/unresolvable path var returns a structured `dispatch_error` envelope — never `-32603` — via both `preview_dispatch` and the meta-tool funnel
- [x] **AC**: a `path` + `+path` collision raises `InvalidSchemaError` at ingest
- [x] OpenAPI snapshot: **proven zero delta** — a fresh regen is byte-identical with and without this change; `cli/api/openapi.json` left untouched

## Out of scope

- The renderer-side reserved-expansion encoding itself (shipped by #2020) — unchanged.
- No `uritemplate` dependency — the hand-rolled single-segment substitution stays (per #2003).
- #1796 (servers base-path) — a separate dispatch defect on the same vRLI ops.
- **Pre-existing, unrelated:** `cli/api/openapi.json` on `main` is ~10k lines stale vs a fresh regen (FastAPI 0.137 route-nesting). This PR does **not** touch the snapshot (committing the regen would inject a 10k-line unrelated diff); a snapshot refresh is a separate task.

Closes #2066

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of RFC6570-style path parameters so reserved path values work correctly during dispatch and preview.
  * Preview requests now return a structured error response instead of failing with an uncaught exception when a path value can’t be resolved.
  * Specs that define both bare and operator-prefixed versions of the same path parameter now fail fast with a clear validation error.

* **Documentation**
  * Updated release and API docs to reflect the new preview error format and RFC6570 path-parameter behavior.

### DIFF
--- a/backend/src/meho_backplane/operations/_rfc6570.py
+++ b/backend/src/meho_backplane/operations/_rfc6570.py
@@ -54,9 +54,14 @@ def split_path_operator(name: str) -> tuple[str, str]:
     """Split a single leading RFC6570 operator off a path-variable name.
 
     Mirrors the renderer's ``_PATH_VAR_RE`` capture exactly: at most one
-    leading character is consumed as an operator, and only when it is one
-    of :data:`RFC6570_PATH_OPERATORS`. A second operator char, or an
-    operator anywhere but the first position, stays part of the name.
+    leading character is consumed as an operator, and only when it is one of
+    :data:`RFC6570_PATH_OPERATORS` **and** a non-empty name remains after it.
+    A second operator char, or an operator anywhere but the first position,
+    stays part of the name; an operator-only token (``"+"``) is treated as the
+    whole name, because ``_PATH_VAR_RE``'s ``[^{}]+`` name group requires at
+    least one character -- so for ``{+}`` the regex captures operator ``""`` /
+    name ``"+"``, and this helper must classify it identically (the module's
+    anti-drift invariant).
 
     >>> split_path_operator("+path")
     ('+', 'path')
@@ -66,10 +71,12 @@ def split_path_operator(name: str) -> tuple[str, str]:
     ('+', '+weird')
     >>> split_path_operator("path+")
     ('', 'path+')
+    >>> split_path_operator("+")
+    ('', '+')
 
     Returns ``(operator, bare_name)`` where ``operator`` is ``""`` when the
     name carries none.
     """
-    if name and name[0] in RFC6570_PATH_OPERATORS:
+    if len(name) >= 2 and name[0] in RFC6570_PATH_OPERATORS:
         return name[0], name[1:]
     return "", name
--- a/backend/src/meho_backplane/operations/_rfc6570.py
+++ b/backend/src/meho_backplane/operations/_rfc6570.py
@@ -1,0 +1,75 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""Shared RFC6570 expression-operator vocabulary for ingested-op paths.
+
+A single source of truth for the one RFC6570 (URI Template) detail two
+otherwise-decoupled stages of the ingested-op pipeline must agree on:
+the set of leading **expression operators** a path-template variable may
+carry, and how the *bare* variable name is recovered from an
+operator-bearing one.
+
+* The **ingest parser** (:mod:`meho_backplane.operations.ingest.openapi`)
+  reads a vendor spec's ``parameters`` and keys each one's JSON-Schema
+  property on its name. For an ``in: path`` parameter whose declared
+  ``name`` carries an operator -- e.g. VCF Operations-for-Logs declares
+  ``{"name": "+path", "in": "path"}`` for the template ``/events/{+path}``
+  -- the property must be keyed on the **bare** name (``path``).
+* The **path renderer** (:func:`meho_backplane.operations._branches._substitute_path`)
+  expands the template at dispatch time, stripping the operator to look
+  the value up by the same bare name and choosing the encoding safe-set
+  from the operator (reserved expansion for ``+`` / ``#``).
+
+If those two stages used different operator sets the property key and the
+lookup key would disagree and the op would be undispatchable (#2003 /
+#2066). Sharing this leaf module -- which imports nothing beyond the
+standard library, so neither the heavyweight dispatch branch nor the
+dependency-free parser is coupled to the other -- makes that drift
+impossible by construction.
+
+The operator set is RFC6570 §2.2's expression operators restricted to the
+ones that can sensibly head a single-segment path variable:
+``+`` (reserved expansion, §3.2.3), ``#`` (fragment), ``.`` (label),
+``/`` (path segment), ``;`` (path-style param), ``?`` / ``&`` (form-style
+query). The reserved/query operators are not *expanded* specially by the
+single-segment substituter, but they are recognised so the operator is
+never mistaken for part of the variable name.
+"""
+
+from __future__ import annotations
+
+__all__ = [
+    "RFC6570_PATH_OPERATORS",
+    "split_path_operator",
+]
+
+# The RFC6570 §2.2 expression operators recognised at the head of a
+# single-segment path-template variable. Mirrored verbatim by the
+# renderer's ``_PATH_VAR_RE`` character class so the parser and the
+# renderer can never disagree on what counts as an operator.
+RFC6570_PATH_OPERATORS = "+#./;?&"
+
+
+def split_path_operator(name: str) -> tuple[str, str]:
+    """Split a single leading RFC6570 operator off a path-variable name.
+
+    Mirrors the renderer's ``_PATH_VAR_RE`` capture exactly: at most one
+    leading character is consumed as an operator, and only when it is one
+    of :data:`RFC6570_PATH_OPERATORS`. A second operator char, or an
+    operator anywhere but the first position, stays part of the name.
+
+    >>> split_path_operator("+path")
+    ('+', 'path')
+    >>> split_path_operator("path")
+    ('', 'path')
+    >>> split_path_operator("++weird")
+    ('+', '+weird')
+    >>> split_path_operator("path+")
+    ('', 'path+')
+
+    Returns ``(operator, bare_name)`` where ``operator`` is ``""`` when the
+    name carries none.
+    """
+    if name and name[0] in RFC6570_PATH_OPERATORS:
+        return name[0], name[1:]
+    return "", name
--- a/backend/src/meho_backplane/operations/ingest/openapi.py
+++ b/backend/src/meho_backplane/operations/ingest/openapi.py
@@ -856,30 +856,51 @@ def _populate_param_properties(
     force the caller to pass ``+path`` to satisfy validation, which the renderer
     would then fail to resolve -- the #2003/#2066 dead-op. ``descriptor.path``
     keeps the operator verbatim so the renderer still selects reserved
-    expansion. A spec declaring both ``path`` and ``+path`` collapses onto the
-    same bare key and raises :class:`InvalidSchemaError` rather than silently
-    dropping one.
+    expansion.
+
+    Two **path** parameters that collapse onto the same bare key (a spec
+    declaring both ``path`` and ``+path``) raise :class:`InvalidSchemaError`
+    rather than silently dropping one -- one of the resulting template vars
+    would be unsatisfiable. A same-name collision *across different* ``in``
+    locations (the legal OpenAPI ``id``-in-path + ``id``-in-query) is **not** a
+    fault and never raises: the params flatten to one property (the model
+    carries a single ``x-meho-param-loc`` per name), and the **path** location
+    wins regardless of declaration order so the path template var stays
+    satisfiable -- a query/header twin can't shadow it into an undispatchable
+    op. The guard is scoped to path-vs-path so it never rejects this legal
+    shape.
     """
+    path_bare_names: set[str] = set()
     for (name, location), param in merged.items():
         prop_schema = _build_param_property(param, component_schemas)
         prop_schema["x-meho-param-loc"] = location
         prop_name = name
         if location == "path":
             _operator, prop_name = split_path_operator(name)
-        if prop_name in properties:
-            # Two path params collapse onto the same property key once the
-            # RFC6570 operator is stripped (a spec declaring both ``path`` and
-            # ``+path``, in either order). A raw duplicate ``(name, in)`` can't
-            # reach here -- ``merged`` already dedupes those -- so a collision
-            # is always an operator-normalised path-param clash. Fail loudly
-            # rather than silently drop a parameter the renderer still needs.
-            raise InvalidSchemaError(
-                f"paths.{path}.{method.lower()}: path parameter {name!r} "
-                f"normalises to property key {prop_name!r} (RFC6570 operator "
-                "stripped), which is already declared; a spec must not declare "
-                "both the bare and an operator-prefixed form of one path "
-                "variable."
-            )
+            if prop_name in path_bare_names:
+                # Another path param already claimed this bare key once the
+                # RFC6570 operator was stripped (``path`` + ``+path``, in either
+                # order). Failing loudly beats silently dropping a path var the
+                # renderer still needs. Scoped to path-vs-path: a collision with
+                # a non-path param of the same name is the legal cross-location
+                # shape handled below.
+                raise InvalidSchemaError(
+                    f"paths.{path}.{method.lower()}: path parameter {name!r} "
+                    f"normalises to property key {prop_name!r} (RFC6570 operator "
+                    "stripped), which another path parameter already declares; a "
+                    "spec must not declare both the bare and an operator-prefixed "
+                    "form of one path variable."
+                )
+            path_bare_names.add(prop_name)
+        elif prop_name in path_bare_names:
+            # A non-path param shares its name with a path param already
+            # placed. Don't let it overwrite the path property -- the path var
+            # must stay substitutable or the op is undispatchable. Keep the
+            # path binding; drop this twin (it can't have its own property key
+            # in the flattened model anyway). Order-independent: the symmetric
+            # "path declared after the twin" case is handled by the path branch
+            # overwriting the twin below.
+            continue
         properties[prop_name] = prop_schema
         is_required = param.get("required") is True or location == "path"
         if is_required and prop_name not in required:
--- a/backend/src/meho_backplane/operations/ingest/openapi.py
+++ b/backend/src/meho_backplane/operations/ingest/openapi.py
@@ -62,6 +62,7 @@ from urllib.parse import urljoin, urlparse
 import httpx
 import yaml
 
+from meho_backplane.operations._rfc6570 import split_path_operator
 from meho_backplane.operations.ingest.exceptions import (
     InvalidSchemaError,
     InvalidSpecError,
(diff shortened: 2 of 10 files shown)


damir-topic:
@coderabbitai review

backend/src/meho_backplane/operations/ingest/openapi.py
```diff
@@ -819,6 +831,61 @@ def _build_parameter_schema(
...
+    same bare key and raises :class:`InvalidSchemaError` rather than silently
+    dropping one.
+    """
+    for (name, location), param in merged.items():
+        prop_schema = _build_param_property(param, component_schemas)
+        prop_schema["x-meho-param-loc"] = location
+        prop_name = name
+        if location == "path":
+            _operator, prop_name = split_path_operator(name)
+        if prop_name in properties:
```
damir-topic: **[🛑 B1]** — Collision guard fires on *any* flattened-name collision, not just `path`/`+path`

**What:** This `if prop_name in properties:` branch raises the RFC6570-operator-specific `InvalidSchemaError` for **every** flat-name collision — but the only collision it should claim is a path-vs-path operator clash. A spec that legally declares the same `name` in two different `in` locations — e.g. `{"name": "id", "in": "path"}` plus `{"name": "id", "in": "query"}` — survives `merged[(name, in)]` (distinct keys), then the second insert into `properties["id"]` lands here and raises, even though **no operator was stripped**. Verified empirically: that pair raises `COLLISION on 'id' (param 'id' loc 'query'); existing loc 'path'`.

**Why it matters:** Two problems. (1) **Wrong scope / undisclosed behavior** — the CHANGELOG, the docstring, and the task AC all scope this guard to "a spec declaring both `path` and `+path`", but the implementation fires far more broadly. (2) **Incorrect + misleading on legal input** — the pre-fix code silently last-wins-overwrote that same-name-different-location pair (lossy, but the op still ingested); this PR turns it into a hard ingest failure whose message tells the operator they declared "both the bare and an operator-prefixed form of one path variable" when they did no such thing. For the config-driven-connector goal (#1964), ingesting arbitrary vendor specs means this legal shape is plausible (the repo's own 10 specs don't hit it today, but that's not a guarantee).

**Required change:** Only emit the operator-specific message when an operator was actually stripped on a path-vs-path collision — gate on `location == "path"` **and** the existing property's `x-meho-param-loc == "path"` **and** `prop_name != name`. For any other flattened collision, either let them coexist or raise a separately-worded error that does not claim an RFC6570 operator clash. Add a `{name:"id",in:"path"}` + `{name:"id",in:"query"}` test asserting the correct outcome. Sketch:

```python
if prop_name in properties:
    existing = properties[prop_name]
    existing_loc = existing.get("x-meho-param-loc") if isinstance(existing, dict) else None
    if location == "path" and existing_loc == "path" and prop_name != name:
        raise InvalidSchemaError(
            f"paths.{path}.{method.lower()}: path parameter {name!r} "
            f"normalises to property key {prop_name!r} (RFC6570 operator "
            "stripped), which is already declared; a spec must not declare "
            "both the bare and an operator-prefixed form of one path variable."
        )
    # otherwise: a same-name-different-location collision the flat schema
    # cannot represent — coexist, or raise a generically-worded error.
```

(Also flagged by CodeRabbit, Major.)

backend/src/meho_backplane/operations/_rfc6570.py
```diff
@@ -0,0 +1,75 @@
...
+    ('', 'path')
+    >>> split_path_operator("++weird")
+    ('+', '+weird')
+    >>> split_path_operator("path+")
+    ('', 'path+')
+
+    Returns ``(operator, bare_name)`` where ``operator`` is ``""`` when the
+    name carries none.
+    """
+    if name and name[0] in RFC6570_PATH_OPERATORS:
```
damir-topic: **[💡 m1]** — `split_path_operator` disagrees with `_PATH_VAR_RE` on an operator-only name

**What:** `split_path_operator("+")` returns `("+", "")` — an empty bare name — but the renderer's `_PATH_VAR_RE` (`backend/src/meho_backplane/operations/_branches.py:80`) requires `[^{}]+` after the optional operator, so it parses the template `{+}` as a variable literally named `"+"` (operator `""`, name `"+"`). Verified: for `{+}` the regex yields `('', '+')` while `split_path_operator('+')` yields `('+', '')`. That is the exact ingest/render key disagreement this module was created to make impossible.

**Why it matters:** Pathological in practice — a path variable named literally `"+"` (or a `{+}` template with no varname) is not valid RFC6570 (§2.3 requires a varname), so no real spec reaches it. But this module's stated invariant is "the property key the parser registers can never drift from the bare name the renderer looks up." Leaving an input on which they *do* drift undermines that guarantee and the `test_renderer_regex_uses_the_shared_operator_set` claim.

**Required change:** Require a name character after the operator, matching the regex:

```python
if len(name) > 1 and name[0] in RFC6570_PATH_OPERATORS:
    return name[0], name[1:]
return "", name
```

Add a `("+", ("", "+"))` (and ideally `("#", ("", "#"))`) case to `test_operations_rfc6570.py::test_split_path_operator`. (Also flagged by CodeRabbit, Minor.)

damir-topic (commented):




**Anchor task:** #2066 — RFC6570 `{+path}` ingest keys the schema property on the raw operator-bearing name while the renderer looks up the bare name; `preview_operation` 500s
**Surface area:** backend
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `9bcce0daa7e463d87cf8855fd056adfd62c67e61`
**AI reviewer (CodeRabbit):** received in 213s

The core of this PR is correct and well-built: the ingest/renderer key
agreement is genuinely symmetric (the shared `RFC6570_PATH_OPERATORS`
constant drives both `_PATH_VAR_RE` and `split_path_operator`, proven
identical across a battery of inputs), the preview never-raises wrap is
scoped to exactly `(KeyError, RuntimeError)` and leaves the execute-path
contract untouched, no `uritemplate` dependency was added, the OpenAPI
snapshot is provably neutral (a fresh regen is byte-identical to the
committed file), and the DoD tests assert at the real dispatch level.
One blocker (B1) keeps this from APPROVE: the new collision guard is
materially broader than its stated contract and rejects a legal
OpenAPI shape with a misleading error.

### Acceptance criteria
- [x] Ingest keys a `{name:"+path", in:"path"}` param on the bare `path` with `x-meho-param-loc:"path"` — `test_parse_operator_bearing_path_param_keyed_on_bare_name` (ran, green)
- [x] `call_operation` on a `/v1/{+path}` op dispatches via reserved expansion (`/v1/events/CONTAINS%20error`, slash literal, space encoded, no `KeyError`) — `test_call_operation_dispatches_operator_bearing_path_param` exercises the REAL `dispatcher.dispatch` + a recording HTTP connector (ran, green)
- [x] `preview_operation` returns a structured `dispatch_error` envelope (never `-32603`) for a missing/unresolvable path var — proven through both `preview_dispatch` and the `preview_operation` meta-tool funnel (ran, green)
- [x] A `path` + `+path` collision raises `InvalidSchemaError` — `test_parse_bare_and_operator_path_param_collision_raises` (ran, green)
- [⚠️] A spec declaring **both** `path` and `+path` fails ingest loudly *and nothing else does* — **NOT fully satisfied**: the guard also rejects legal same-name-different-location params (see B1)
- [x] `cli && make snapshot-openapi && make generate` neutral — regen byte-identical (sha256 `9a8972be…`), snapshot correctly left untouched

### Project-board alignment
|---|---|---|---|---|---|
| #2066 | Open (task) | Closes | ✅ | ⚠️ (AC met; impl over-broad) | close on merge once B1 resolved |

Parent goal #221 / Parent initiative #2065 — coherent; this is the
un-landed ingest half of #2003 (renderer half #2020 already merged), as
the task states.

### Predicted vs. actual diff
- ✅ `_rfc6570.py`, `_branches.py`, `_request_preview.py`, the three test files, `CHANGELOG.md`, both `docs/codebase/*.md` — expected and correct
- ⚠️ `ingest/openapi.py` — expected, but the collision guard is wrong (B1, inline)
- ❓ none unexpected-but-justified beyond the expected test additions
- 🚫 none undisclosed
- 🕳️ none missing

### Integration verification
- [x] `ruff check` — clean (4 changed src files)
- [x] `mypy` — clean (4 source files)
- [x] `pytest` — 124 passed (DoD files); 200 passed (ingest+dispatch+preview+lookup); 347 passed / 101 skipped (dispatch/branches/connector-e2e/MCP/UI blast-radius sweep)
- [x] OpenAPI snapshot regenerated — **byte-identical** to committed; snapshot-neutral claim proven; CI `CLI API snapshot freshness` = SUCCESS
- [⚠️ not run] migrations replay — no migration in diff
- [⚠️ not run] `docker compose up` — not needed for this change
- [⚠️ not run] agent trace — no agent-reasoning path touched

- 🛑 B1 — `ingest/openapi.py:869`: collision guard fires on **any** flat-name collision, not just `path`/`+path`. A legal `{name:"id",in:"path"}` + `{name:"id",in:"query"}` (which the pre-PR code silently last-wins-overwrote) now raises `InvalidSchemaError` with a message wrongly claiming an RFC6570 operator clash. Required change: only emit the operator-specific error when an operator was actually stripped on a path-vs-path collision (`location=="path"` AND existing `x-meho-param-loc=="path"` AND `prop_name != name`); otherwise coexist or raise a separately-worded error. Add a same-name-different-location test.
- 💡 m1 — `_rfc6570.py:73`: `split_path_operator("+")` → `("+","")` disagrees with `_PATH_VAR_RE` (which parses `{+}` as a var named `"+"`), reintroducing the exact drift this module exists to prevent. Required change: `if len(name) > 1 and name[0] in RFC6570_PATH_OPERATORS:` + a `"+"` regression case. Pathological in practice, but the module's invariant is "the two stages can never disagree", so close it.

- The flat `properties`-by-bare-name `parameter_schema` model fundamentally can't represent two same-bare-name params from different `in` locations; B1's fix corrects the diagnostic/failure mode, not the underlying representational limit. Namespacing the key by location is a separate, larger change — out of scope for #2066.
- A repo-wide scan found **0 of 10** packaged/test OpenAPI specs hit the B1 collision path today, so B1 has no live-spec blast radius at present — but Goal #1964 (config-driven connectors) means arbitrary vendor specs get ingested, where the legal same-name-multi-location shape is plausible.

- Adopted: 2 — both CodeRabbit findings independently confirmed empirically and adopted (B1 ⇐ CR Major; m1 ⇐ CR Minor), cited inline.
- Disputed: 0
- Out of scope: 0

### Docs
- [x] `docs/codebase/dispatch-request-preview.md` + `docs/codebase/spec-ingestion.md` — updated, accurate, and match the implementation; no `.NET` analogies. `docs_status: verified` (the PR updates the two relevant artifacts).

- The "single shared constant drives both the regex char-class and the strip helper, so they can't drift" design is the right fix for a drift-class bug, and the `test_renderer_regex_uses_the_shared_operator_set` test locks the invariant in.
- The preview never-raises wrap is exemplary: scoped to exactly the two documented exception types, with a comment explaining *why* `resolve_ingested_request` is deliberately left raising (execute path relies on the dispatcher's structured mapping). The `preview_operation` meta-tool funnel is tested end-to-end, so the MCP `-32603` regression is genuinely closed.
- Snapshot discipline: proving byte-identity rather than asserting it, and correctly leaving the snapshot untouched.

damir-topic:
@coderabbitai review

Issue #1974, opened 2026-06-19
Pilot: migrate vcf_logs/vRLI auth+fingerprint to an ExecutionProfile, prove per-method parity (ResultHandle stays typed)


damir-topic:
Parent goal: #1964
Parent initiative: #1965
Depends-on: #1967, #1968, #1969, #1970, #1971, #1972, #1973

## Problem

Validate the whole profiled-connector model against **real bespoke code** by migrating one connector's auth + fingerprint to a profile — proving per-method dispatch parity without regressing the parts that can't be a profile. vRLI is the pilot because it exercises the session-login scheme, a multi-part version parse, and a genuinely bespoke large-result path.

## Current state (verified in code, v0.18.0)

- [`connectors/vcf_logs/connector.py`](backend/src/meho_backplane/connectors/vcf_logs/connector.py) — vRLI session→Bearer auth (`{401, 440}` expiry), a 5-part version parse in `fingerprint`, and a bespoke `ResultHandle` large-result path that is **not** method/path dispatch and is **not** expressible by a profile.

## Desired state

- Author the `vrli_session` profile: named scheme, `login_path`, the provider const via the named extractor (T4), expiry `{401, 440}` (T7), `fingerprint` GET version path + the 5-part splitter (T6).
- Retire **only** `auth_headers` + `fingerprint` from the typed class; keep the `ResultHandle` large-result path hand-coded.
- Keep the connector's existing ingested rows; the connector class shrinks to `ResultHandle` + registration.

## Acceptance criteria

- [ ] vRLI `auth_headers` + `fingerprint` are served by the profile; the typed class shrinks to the `ResultHandle` path + registration.
- [ ] An integration test (testcontainers lane — mind the `tenant_id`-double trap on integration doubles) proves per-method dispatch parity vs the pre-migration connector.
- [ ] The `ResultHandle` large-result path is unchanged.

## Out of scope

- Migrating the other ~5 catalog-fit connectors (harbor/sddc/vcf_fleet/vcf_operations/argocd/keycloak) — a follow-up wave once the pilot proves parity.

## References

- Parent #1964 / #1965. The capstone that exercises T1–T7 against a real connector.

damir-topic:
**Dogfood finding (v0.18.0 RDC Hetzner-DC deploy, G0.28 cycle #1998) — folding this in as an acceptance criterion rather than filing a competing connector patch.**

[...]

PR #2072, opened 2026-06-24
fix(connectors): re-login + retry-once on 401/440 at the ingested dispatch arm

### LINKED ISSUE #2067: Task: wire session-expiry re-login + retry-once (401/440) into the generic-ingested dispatch path — fixes vCenter cold-401 + vRLI 440 (+ nsx) in one dispatcher-arm seam
## Summary

Session-expiry **re-login + retry is not wired into the generic-ingested dispatch path**. A connector's in-process session token can expire server-side while it is still cached; the next dispatched op gets an auth-failure status (`401` for vCenter REST, `440 "Login Timeout"` for vRLI), and the dispatcher *classifies* it as `connector_auth_failed` but **never invalidates the cached token and retries**. The op stays failing until the backplane process restarts (which clears the in-memory cache). This single defect is the root of **two** v0.19.0 RDC "red" findings:

- **vRLI 440** — a soak-confirmed hourly log-sentry blocker (#1135/#1139): a fresh post-restart session returns real logs, then 440s ~36 min later with no restart, while a direct `POST /api/v2/sessions` returns 200 in the same window.
- **vCenter cold-401** — all typed `vmware-rest` reads on `rdc-vcenter` go down once the cached session expires (`~37 ms` fast-fail = cache-hit + immediate 401 on the GET), while the identical credential authenticates via direct REST (`POST /api/session` → 201) and SOAP (`govc` → 200).

Both connectors dispatch as `source_kind=ingested` through the **same** dispatcher `HTTPStatusError` arm, so one seam fixes both (and nsx). This corrects two RDC root-cause misattributions and supersedes the v0.18 fold-in (#1998 folded the 440 fix onto #1974/#1973; those merged but only carried the expiry-status set, never the dispatch-path recovery — which is why the soak still fails).

## Current state — one missing seam, two symptoms

- **Generic-ingested dispatch has no recovery.** vRLI's curated read ops and vCenter's `vcenter.yaml` ops are `EndpointDescriptor` rows dispatched via [`operations/_branches.py::dispatch_ingested`](backend/src/meho_backplane/operations/_branches.py#L419-L427) — a single `connector._request_json` (GET) / `connector._post_json` (write) call. On an auth-failure the `httpx.HTTPStatusError` propagates to the dispatcher's arm ([`operations/dispatcher.py:971-978`](backend/src/meho_backplane/operations/dispatcher.py#L971-L978)) which returns `result_connector_auth_failed` — it diagnoses but never invalidates or retries. `is_auth_failed_status` ([`operations/_errors.py:840-862`](backend/src/meho_backplane/operations/_errors.py#L840-L862)) recognizes both 401 and 440 (module set `_AUTH_FAILED_STATUSES = {401,440}`), so classification is correct; only recovery is missing.
- **vRLI (440): #1921's fix is on a path the failing call never executes.** PR #1921 added 440 to the re-login trigger in `VcfLogsConnector._get_json_with_session_retry` ([`connectors/vcf_logs/connector.py:436-480`](backend/src/meho_backplane/connectors/vcf_logs/connector.py#L436-L480)) — but that helper has **no production caller** for vRLI (grep: only nsx calls its own analog; vRLI `fingerprint()` uses an *unauthenticated* `GET /api/v2/version` at [`connector.py:487-496`](backend/src/meho_backplane/connectors/vcf_logs/connector.py#L487-L496)). #1921's own "out of scope" note admitted the dispatched ingested-op path does not go through it. **Not** a #1973/#1974 migration regression — the profile correctly declares `expiry_statuses={401,440}` ([`connectors/vcf_logs/profile.py:90`](backend/src/meho_backplane/connectors/vcf_logs/profile.py#L90)).
- **vCenter (401): the typed connector has no re-login at all.** `VmwareRestConnector` caches one token per `(tenant_id, target.id)` in `self._session_tokens` with **no TTL / no expiry check** ([`connectors/vmware_rest/connector.py:256`](backend/src/meho_backplane/connectors/vmware_rest/connector.py#L256)); a cache hit returns the (possibly stale) token unconditionally ([`connector.py:380-384`](backend/src/meho_backplane/connectors/vmware_rest/connector.py#L380-L384) — the `~37 ms` fast path), and `_establish_and_cache_session` runs only on a cache **miss** ([`connector.py:386-451`](backend/src/meho_backplane/connectors/vmware_rest/connector.py#L386-L451)). The docstring states 401-driven refresh was "intentionally deferred" since #518 ([`connector.py:49-60`](backend/src/meho_backplane/connectors/vmware_rest/connector.py#L49-L60)). **Not** a v0.19 regression of #2025/#2031/#2047 (those touch the profiled path; the log line `tie_break=hand_rolled_over_shim` confirms the *typed* class dispatched).
- **The dispatcher comment encodes the false assumption.** [`dispatcher.py:939-945`](backend/src/meho_backplane/operations/dispatcher.py#L939-L945) asserts "the hand-coded session connectors already retry once on a 401 internally (`_get_json_with_session_retry`)" — false for the ingested dispatch path (the comment text entered via #1804, which correctly shipped only the *error-shape* surfacing, not the re-login feature).
- **The recovery seam is half-built:** `VcfLogsConnector._invalidate_session` exists ([`connector.py:422`](backend/src/meho_backplane/connectors/vcf_logs/connector.py#L422)); `ProfiledRestConnector._invalidate_session` is defined but has **zero production callers** ([`connectors/profiled.py:531`](backend/src/meho_backplane/connectors/profiled.py#L531)); `VmwareRestConnector` has none.

## Desired state

Wire **invalidate-and-retry-once on an auth-failure status** into the generic-ingested dispatch path — the path both connectors actually traverse:

1. **Duck-typed hook.** Give each session-stateful connector a public/duck-typed `invalidate_session(target)`:
   - `VcfLogsConnector` — expose its existing `_invalidate_session`.
   - `VmwareRestConnector` — new: pop `self._session_tokens[cache_key]` and `self._session_paths[cache_key]` under `self._session_lock`, using `target_cache_key(target)` so per-`(tenant_id, target.id)` isolation (#1642/#1672/#1684) is preserved.
   - Stateless/Basic connectors expose no hook → behaviour unchanged.
2. **Dispatcher-arm recovery.** In the `HTTPStatusError` arm ([`dispatcher.py:971-978`](backend/src/meho_backplane/operations/dispatcher.py#L971-L978)), **before** returning `result_connector_auth_failed`: if the status `is_auth_failed_status(...)`, a retry has not already been attempted, and the connector advertises `invalidate_session` → invalidate the token and **re-dispatch the same ingested request exactly once**. A second auth-failure falls through to `result_connector_auth_failed` (re-login genuinely failed). Retry-once on an auth-failure status is **safe for non-idempotent verbs too**, because 401/440 means the upstream rejected the request *pre-execution* (no side effect occurred) — gate strictly on `status in {401,440}`, **never** on 5xx/timeout (which may have executed). Re-establishing the vCenter session must re-run the modern→legacy `/api/session` 404 fallback (`_establish_and_cache_session` already does).
3. **Audit ordering.** The arm calls `audit_and_broadcast_safe` with `result_status='error'` before the status branch ([`dispatcher.py:961-969`](backend/src/meho_backplane/operations/dispatcher.py#L961-L969)); a retry-and-succeed must not emit a spurious error audit row — emit the auth-failed audit only after the retry is exhausted.
4. **Contracts.** Keep the retried call inside the same `try` so a second-attempt `ConnectError`/etc. is still caught (the dispatcher never-raises contract). Correct the misleading comments at [`dispatcher.py:939-945`](backend/src/meho_backplane/operations/dispatcher.py#L939-L945) and `_errors.py`.

## Acceptance criteria

- A generic-ingested **vRLI GET** that returns `440` once then `200` succeeds via one automatic re-login (`POST /api/v2/sessions`) + single retry — **no backplane restart** — proven by a **dispatch-level** test (seeded `source_kind='ingested'` `EndpointDescriptor`), not a connector-unit test calling the helper directly (the existing `test_connectors_vcf_logs_e2e.py` drives the helper and stays green while the bug ships).
- A typed **vCenter `GET:/vcenter/*`** *and* a non-idempotent **`POST:/vcenter/*`** on an expired session each recover via one re-login (with the modern→legacy 404 fallback) + retry.
- A second consecutive auth-failure (re-login also fails) resolves to `result_connector_auth_failed` with no retry loop.
- Per-`(tenant_id, target.id)` cache isolation is preserved across eviction + re-establish; two same-named targets in different tenants never collapse.
- Recovery is gated on `status in {401,440}` only — a 5xx/timeout is **not** retried.
- A successful retry emits exactly one success audit row (no spurious error row); a failed retry emits the auth-failed row.
- Stateless/Basic connectors (no `invalidate_session` hook) are unaffected.
- The misleading dispatcher/`_errors` comments are corrected.
- `cd cli && make snapshot-openapi && make generate` only if a route/schema changed (expected no delta — transport/dispatch-internal; verify).

## Out of scope

- Proactive token TTL (re-mint before the doomed call) — reasonable defense-in-depth, but the invalidate-on-auth-failure seam is the load-bearing fix; keep optional.
- Migrating vRLI/vCenter to `ProfiledRestConnector` as the runtime class — the duck-typed hook must satisfy both the typed connectors and a future profiled one without an `isinstance` ladder, so this fix forward-fits #1974 but does not depend on it.
- #1796 (servers base-path) — separate dispatch defect.

## References

- Symptom + misdirected prior fix: #1909 / PR #1921 (440 re-login landed in an uncalled helper); v0.18 fold-in: #1998 onto #1973/#1974.
- Where the false "connectors retry on 401" dispatcher comment entered: #1804 (reference only — it correctly shipped the error-shape surfacing, not the re-login).
- Shared expiry-status source: #1973; original typed vSphere connector (deferred-401 origin): #518; recovery seam stubs: `profiled.py:531`.
- Log-sentry: #1135/#1139; config-driven connectors: Goal #1964 / Initiative #1965.
- Best-practices anchors — *python*: recover at the layer the failure occurs (dispatch-time invalidate+retry, not classify-only); idempotency/safety reasoning for the auth-failure retry; observability (no spurious error audit on a recovered call); testing — assert at the path the bug actually traverses (dispatch level), not the helper a passing test exercises.

Parent goal: #221
Parent initiative: #2065

damir-topic:
## Summary

- Wires **session-expiry re-login + retry-once** into the generic-ingested dispatch path — the single seam both vCenter REST (cold-401) and vRLI (idle-440) traverse. Previously the dispatcher's `HTTPStatusError` arm *classified* the auth failure as `connector_auth_failed` but never invalidated the cached session or retried, so both hard-failed until a backplane restart. This is the headline architectural fix of the v0.19.0 dogfood-hardening initiative (#2065): one dispatcher-arm change fixes two RDC "red" findings (and nsx).
- On an auth-class status (`401`/`440`), when the connector advertises a **duck-typed** `invalidate_session(target)` hook, the dispatcher evicts the cached token and re-dispatches the same request **exactly once**. A re-login that succeeds flows up the normal success path (one success audit row, no spurious error row); a second consecutive auth failure falls through to `connector_auth_failed` (re-login genuinely failed).
- The hook satisfies the typed connectors (`VmwareRestConnector` — new; `VcfLogsConnector`/`ProfiledRestConnector` — expose existing `_invalidate_session`) **and** a future `ProfiledRestConnector` runtime class without an `isinstance` ladder — it forward-fits #1974 without depending on it. Stateless/Basic connectors (no hook) are unaffected.
- Retry-once is **safe for non-idempotent verbs** (a `401`/`440` is rejected pre-execution → no side effect) and is gated **strictly** on the auth-class set — a `5xx`/timeout is never retried.

## Why the prior fixes didn't land

`#1921`'s 440-recovery lives in `VcfLogsConnector._get_json_with_session_retry`, which has **no caller** on the dispatch path (vRLI's curated reads dispatch via `dispatch_ingested` → `HttpConnector._request_json`). `VmwareRestConnector` never had 401-driven re-login (deferred since #518) and its `_session_tokens` cache has no TTL. The classification was already correct (profile `expiry_statuses={401,440}`, `_AUTH_FAILED_STATUSES`); only **recovery** was missing.

## Test plan

- [x] `ruff check` / `ruff format --check` — clean (changed files)
- [x] `mypy src/` — `Success: no issues found`
- [x] **Dispatch-level recovery** (`test_operations_connector_auth_failed.py`, seeded `source_kind='ingested'` descriptors — NOT the helper a passing unit test exercises):
  - vRLI GET `440` → `200` recovers via one re-login + single retry (one **success** audit row)
  - vCenter GET **and** non-idempotent POST `401` recover via one re-login + retry
  - second consecutive auth failure → `connector_auth_failed`, no retry loop, one **error** audit row
  - `5xx` with the hook present → **not** retried, never invalidated
  - stateless connector (no hook) → unchanged, single attempt
  - per-`(tenant_id, target.id)` cache isolation preserved across eviction + re-establish
- [x] **Connector-unit** coverage for the real hooks (`test_connectors_vmware_rest_auth.py`, `test_connectors_vcf_logs_auth.py`, `test_connectors_profiled_auth.py`): `invalidate_session` evicts only the target's slot (+ login path), re-login fires on the next `auth_headers`, delegation works
- [x] Full backend unit suite (`pytest -n 3 --dist loadscope --ignore=tests/integration tests/`) — green
- [x] No route/schema/MCP surface touched → no OpenAPI snapshot delta (transport/dispatch-internal)

## Out of scope

- Proactive token TTL (re-mint before the doomed call) — optional defense-in-depth per the issue; the invalidate-on-auth-failure seam is the load-bearing fix.
- Migrating vRLI/vCenter to `ProfiledRestConnector` as the runtime class — the duck-typed hook forward-fits #1974 without it.
- Sibling task #2066 edits a different `_branches.py` function (path-substitution); no hunk overlap.

Closes #2067


damir-topic:
@coderabbitai review

PR #1709, opened 2026-06-12
fix(ui): pin hx-target="this" on the memory tag datalist (#1695)

damir-topic:
## Summary
- The memory list's tag-autocomplete `<datalist>` issues a `hx-trigger="load"` GET to `/ui/memory/tags` but carried no own `hx-target`; htmx 2.0.9 resolves `hx-target` closest-wins up the ancestor chain (verified in the vendored `htmx.min.js`: `getTarget` → `getClosestAttributeValue` ancestor walk), so the fetch inherited the filter form's `hx-target="#memory-cards"` and swapped the `<option>` fragment into the card grid, wiping the memory list on every page load (RDC v0.14.0 cycle-10 finding FE-4).
- Fix pins `hx-target="this"` on the datalist — `getTarget`'s `"this"` branch resolves via `findThisElement` to the declaring element, so the options land in the datalist and the grid stays intact. Same local-override idiom as #1706's form-level `hx-headers`.
- Adds a rendered-page regression test pinning the datalist's self-target, a `docs/codebase/ui.md` HTMX-conventions bullet recording the why, and the CHANGELOG bullet.

## Test plan
- [x] `ruff check src/ tests/` — All checks passed
- [x] `ruff format --check src/ tests/` — 950 files already formatted
- [x] `mypy src/` — no issues in 468 source files
- [x] `pytest tests/test_ui_memory_list.py` — 27 passed (incl. new `test_list_datalist_overrides_inherited_hx_target`)
- [x] `.claude/hooks/code-quality.py --diff` — clean
- [x] AC: grep confirms the datalist in `memory/index.html` now carries `hx-target="this"` (line 130)
- [x] AC (cards intact / swap into `#memory-tag-options` / suggestions / filter): server-rendered contract pinned by the regression test; client swap-location behavior verified against the vendored htmx 2.0.9 source (`Ee`/`Se` = `getTarget`/`findThisElement`); existing `test_list_tag_filter_narrows_results` + `test_tags_endpoint_returns_distinct_sorted_tag_options` cover the filter round-trip. Live-browser confirmation lands with the next RDC dogfood cycle per Initiative #1691 DoD.

## Out of scope
- Tag-filter styling/UX (T1 #877 shipped the MVP), bulk tag management, tag-collection performance — per issue Out of scope.
- Sibling wave-3 tasks #1694 (BFF token refresh) and #1698 (footer version) touch different files.

Closes #1695

PR #2074, opened 2026-06-24
fix(ui): scope tag-datalist hx-push-url/hx-include on /ui/memory load (#2069)

### LINKED ISSUE #2069: Task: /ui/memory tag-datalist still inherits the filter form's hx-push-url/hx-include — rewrites the URL to /ui/memory/tags on every load (residual after #1709)
## Summary

Loading `/ui/memory` silently rewrites the browser URL to `/ui/memory/tags?tag=&scope=all` on every page load. The tag-filter `<datalist>`'s `hx-trigger="load"` GET inherits the parent filter form's `hx-push-url="true"` and `hx-include` (htmx resolves both closest-ancestor-wins, same as `hx-target`), so the options fetch pushes its URL and drags the form's `tag`/`scope` inputs into the query string.

This is the **residual** of #1695/PR #1709, **not** a reopen: #1709 (shipped v0.15.0) fixed the worse half — the grid-clobber — by giving the datalist `hx-target="this"`, and a regression test guards that. But it did not also stop the datalist pushing the URL, so the reporter's "grid still clobbered" is overstated while the URL-rewrite symptom is genuinely live at HEAD. Low severity (cosmetic — the card grid renders correctly), but it dirties browser history and the address bar on every load.

## Current state

- [`backend/src/meho_backplane/ui/templates/memory/index.html:125-131`](backend/src/meho_backplane/ui/templates/memory/index.html#L125-L131) — the `<datalist id="memory-tag-options">` carries `hx-get="/ui/memory/tags"`, `hx-trigger="load"`, `hx-target="this"`, `hx-swap="innerHTML"` — but **no** `hx-push-url` and **no** `hx-include` override.
- [`index.html:101-112`](backend/src/meho_backplane/ui/templates/memory/index.html#L101-L112) — the ancestor filter form carries `hx-target="#memory-cards"`, `hx-push-url="true"`, and `hx-include="closest form"`.
- htmx (vendored 2.0.9) resolves `hx-push-url` and `hx-include` via `getClosestAttributeValue` (walks ancestors; `disableInheritance` defaults false). The datalist has neither attribute and no `hx-disinherit`, so its load-fetch **inherits** `push-url="true"` (→ rewrites the URL) and `hx-include="closest form"` (→ appends `tag=&scope=all`). The `hx-target="this"` override #1709 added correctly keeps the `<option>` fragment ([`memory/_tags_options.html`](backend/src/meho_backplane/ui/templates/memory/_tags_options.html)) inside the datalist (so the grid is safe), but target inheritance and push-url/include inheritance are independent.
- `render_tags` ([`ui/routes/memory/views.py:506-515`](backend/src/meho_backplane/ui/routes/memory/views.py#L506-L515)) sets no `HX-Push-Url: false` response header to suppress it server-side.
- The existing regression test (`backend/tests/.../test_ui_memory_list.py`, `test_list_datalist_overrides_inherited_hx_target`) asserts only `hx-target="this"` — it stays green with the URL-rewrite bug present.

## Desired state

- Add `hx-push-url="false"` to the `<datalist>` at [`index.html:125-131`](backend/src/meho_backplane/ui/templates/memory/index.html#L125-L131) so its load-fetch does not rewrite the URL.
- Add `hx-include="none"` (or `hx-disinherit="hx-include"`) to the datalist so the load-fetch does not drag the form's `tag`/`scope` values into `/ui/memory/tags` (the endpoint ignores them, but it is what produces the exact `?tag=&scope=all` string and is sloppy).
- Extend the existing regression test to also assert the datalist carries `hx-push-url="false"` (and no inherited include) — today it guards only the target.

## Acceptance criteria

- Loading `/ui/memory` leaves the browser URL unchanged (no rewrite to `/ui/memory/tags`), with `#memory-cards` intact.
- The datalist's load-fetch does not append `tag=`/`scope=` to its request.
- The regression test asserts `hx-target="this"` **and** `hx-push-url="false"` on the datalist (extend, don't replace, the #1709 assertion).
- No OpenAPI snapshot regen (pure template attribute change — no route/schema change).

## Out of scope

- Re-litigating the grid-clobber fix (#1709) — it works; this only adds the missing push-url/include scoping.
- Other datalists: the runbook-runs datalist lives inside a modal (not an `hx-push-url` filter form), so it is unaffected — but a one-line grep sweep for any other `<datalist>` inside an `hx-push-url` form is worth doing while filing.

## References

- Adjacent prior art (target half): #1695 / PR #1709 (shipped v0.15.0) — this completes the push-url/include half it left.
- Operator console Goal #336.
- Best-practices anchors — *htmx*: scope inherited attributes (`hx-push-url`, `hx-include`, `hx-target`) on load-triggered child fetches; *testing*: guard every inherited attribute the fix pins, not just the one the prior fix addressed.

Parent goal: #221
Parent initiative: #2065

damir-topic:
## Summary

- The `/ui/memory` tag-autocomplete `<datalist>` fires an `hx-trigger="load"` GET to `/ui/memory/tags` to populate its `<option>` suggestions. It had no `hx-push-url`/`hx-include` of its own, so htmx 2.0.9 resolved both by walking ancestors (closest-wins) and the load-time fetch inherited the filter form's `hx-push-url="true"` + `hx-include="closest form"` — rewriting the browser URL to `/ui/memory/tags?tag=&scope=all` on **every page load** and dragging the form's `tag`/`scope` inputs into a request `render_tags` ignores anyway.
- #1709 (v0.15.0) pinned `hx-target="this"` and fixed the worse grid-clobber half of this same inheritance, but left `hx-push-url`/`hx-include` unscoped. This completes the scoping: add `hx-push-url="false"` + `hx-include="none"` to the datalist (the same surgical child-override pattern), so its load-time fetch leaves the address bar/history untouched and sends no inherited inputs. The options fetch and the card grid (`#memory-cards`) are unchanged.
- Both override values are recognized by the **vendored htmx 2.0.9** source (`ui/static/src/vendor/htmx.min.js`): `"false"` nullifies the push (`if(l==="false")l=null`), `"none"` yields an empty include set (`if(t==="none")…`).
- Template attribute + test + docs only — **no FastAPI route/schema change, OpenAPI snapshot unchanged**.

## Test plan

- [x] `ruff check tests/test_ui_memory_list.py` — clean
- [x] `ruff format --check tests/test_ui_memory_list.py` — already formatted
- [x] `mypy` (repo config: `files=["src"]`, strict) — Success, no issues in 562 source files (the template change has no Python type surface; tests are out of mypy's configured scope)
- [x] `pytest tests/test_ui_memory_list.py` — **30 passed**
- [x] **AC1** browser URL unchanged on load — `hx-push-url="false"` verified in htmx 2.0.9 source + asserted by `test_list_datalist_overrides_inherited_hx_target`
- [x] **AC2** load-fetch sends no `tag=`/`scope=` — `hx-include="none"` verified in htmx 2.0.9 source + asserted by the same test
- [x] **AC3** regression test asserts `hx-target="this"` **and** `hx-push-url="false"` (and `hx-include="none"`) — extended (not replaced) the #1709 assertion
- [x] **AC4** no OpenAPI snapshot regen — confirmed (no route/schema touched)
- [x] code-quality `--diff` gate — pass (exit 0)
- [x] grep sweep — the only other `<datalist>`s (runbooks start-modal, retrieval usage) are server-rendered `{% for %}` option lists with no `hx-trigger="load"` fetch inside an `hx-push-url` form; both unaffected

## Out of scope

- The grid-clobber fix (#1709) is preserved as-is — not re-litigated.
- A server-side `HX-Push-Url: false` response header on `render_tags` was considered and rejected: it would not also stop the `hx-include`, and it splits the fix across client+server when the established pattern (the `hx-target="this"` override) is client-side.
- Adjacent finding (not touched): `tests/test_ui_memory_list.py:256` has a pre-existing `[no-any-return]` mypy note (`return decrypted.id`) in the `_seed_session_sync` helper — out of CI's mypy scope (`files=["src"]`) and unrelated to this change.

Closes #2069


damir-topic:
@coderabbitai review

PR #2075, opened 2026-06-25
fix(connectors): fold relative OpenAPI servers[].url base into ingested op paths

### LINKED ISSUE #1796: Connector dispatch drops the OpenAPI servers[].url base path — ops 404 when the spec uses a relative server base

damir-topic:
## Summary

- Ingested-connector dispatch dropped the OpenAPI `servers[].url` base: it built the request URL as `host:port + op_path`, so an op from a spec with a relative server base (e.g. vRLI's `servers:[{url:"/api/v2"}]`) dispatched to `host:port/version` and 404'd instead of `host:port/api/v2/version`.
- Fix per the reporter's **option A** (lowest blast radius): fold a *relative* server base into each op's `path`/`op_id` **at ingest time** — dispatch is unchanged (no new descriptor column, no dispatch-path edit). `_server_base_path` extracts + normalises the first server's relative url; `_fold_server_base` joins it onto each path, collapsing the double slash a trailing-slash base produces.
- **Absolute** server bases (`https://other-host/...`) are out of scope — meho dispatches against the operator-configured target host, not a spec-declared one (SSRF / topology boundary); only the path of a *relative* base is folded. The RFC6570 `{+path}` template (#2066) rides through untouched (base is a prefix; operator stays on the suffix var).
- Reconciled the two MEHO-authored shipped specs: `vmware_rest_minimal` / `sddc_manager_minimal` already baked their mount (`/api`, `/v1`) into every path key **and** redundantly declared it under `servers` — harmless while `servers` was ignored, but a double-prefix (`/api/api/about`) once it's load-bearing. Dropped the redundant `servers` block from both, so their op_ids are unchanged.

## Test plan

- [x] `mypy` — clean
- [x] code-quality gate — 0 blockers
- [x] **AC: relative base folds** — `servers:[{url:"/api/v2"}]` + `/version` → op_id `GET:/api/v2/version`, `op.path == "/api/v2/version"` (the exact string the dispatcher feeds httpx as the request path).
- [x] **AC: no regression** — absent / `"/"` / empty `servers` ingests `GET:/version` as today (parametrised).
- [x] **AC: regression test** — `servers:[{url:"/base"}]` + `/x` asserts `op.path` contains `/base/x`.
- [x] **AC: double-slash normalised** — trailing-slash base `/api/v2/` + `/version` → `/api/v2/version`.
- [x] absolute-base out-of-scope test — `https://other-host/api/v2` leaves bare `/version`.
- [x] `{+path}` interaction — `/api/v2` + `/events/{+path}` → `/api/v2/events/{+path}`, bare-name param keying intact.
- [x] `pytest` ingest + connector sweep — 1246 passed, 35 skipped, 0 failed; shipped-spec op_ids unchanged after dropping redundant `servers`.
- [x] `cd cli && make snapshot-openapi` — **no delta** (ingest-internal change; no route/schema touched), so `make generate` not needed.

## Out of scope

- Absolute-URL server bases pointing at a different host (SSRF/topology boundary) — only the path component of a relative base is folded.
- Server `variables` templating (`{region}`) — no shipped MEHO spec uses it; file separately if needed.
- The vRLI `GET:/events/{+path}` ingest (#2066) and session-expiry recovery (#2067) — sibling log-sentry fixes, already merged.

Closes #1796

## Summary by CodeRabbit

* **New Features**
  * OpenAPI ingest now respects relative server base URLs when deriving endpoint paths, so operation IDs and dispatched paths include the correct prefix.
* **Bug Fixes**
  * Prevented double-prefixing and double slashes for specs that already include path prefixes.
  * Absolute, protocol-relative, empty, or malformed server URLs are left unchanged.
* **Documentation**
  * Clarified how ingested endpoint paths are built from OpenAPI path templates plus any relative server base.

### DIFF
--- a/backend/src/meho_backplane/operations/ingest/openapi.py
+++ b/backend/src/meho_backplane/operations/ingest/openapi.py
@@ -146,6 +146,88 @@ _INLINE_SPEC_ONRAMP_REMEDIATION = (
 # ``{filter.names}``. Compiled once and reused per operation.
 _PATH_PARAM_RE = re.compile(r"\{([^{}/]+)\}")
 
+
+def _require_spec_mapping(value: Any, label: str) -> dict[str, Any]:
+    """Return ``value`` as a dict, or raise :exc:`InvalidSpecError` naming ``label``.
+
+    The OpenAPI ``components`` container and each of its sub-objects
+    (``schemas`` / ``parameters`` / ``responses`` / ``requestBodies``)
+    must be mappings; a scalar or list there is a malformed spec. Callers
+    pass ``spec.get(key) or {}`` so an absent key arrives as the empty
+    mapping and validates trivially.
+    """
+    if not isinstance(value, dict):
+        raise InvalidSpecError(f"{label!r} must be a mapping, got {type(value).__name__}")
+    return cast(dict[str, Any], value)
+
+
+def _server_base_path(servers: Any) -> str:
+    """Return the relative path prefix declared by ``servers[0].url``.
+
+    OpenAPI resolves an operation's effective URL by *appending* the
+    path-item key to the Server Object's ``url`` (OAS 3.1 §4.8.5: "The
+    path is appended (no relative URL resolution) to the expanded URL
+    from the Server Object's ``url`` field"). meho dispatches against the
+    operator-configured target host, so only the **path** component of a
+    *relative* server base is honoured -- it is folded into each op's
+    path at ingest time (#1796) so the dispatch path stays unchanged.
+
+    Returns the normalised base path (leading ``/``, no trailing ``/``)
+    for a relative server url such as ``/api/v2``; returns ``""`` (no
+    fold, today's behaviour) when:
+
+    * ``servers`` is absent / empty / malformed,
+    * the first server url is ``/`` or empty (a no-op base),
+    * the first server url is **absolute** (has a scheme or authority) --
+      an absolute base points at a spec-declared host, which meho does
+      not dispatch against (SSRF / topology boundary, out of scope per
+      #1796); only its path would be foldable and conflating the two is
+      a footgun, so absolute bases are left entirely to the operator's
+      target configuration.
+
+    Only the first server entry is consulted: OpenAPI allows multiple
+    servers for environment selection, but meho dispatches against one
+    operator-configured target, so the first (primary) base is the one
+    that pairs with it.
+    """
+    if not isinstance(servers, list) or not servers:
+        return ""
+    first = servers[0]
+    if not isinstance(first, dict):
+        return ""
+    url = first.get("url")
+    if not isinstance(url, str) or not url:
+        return ""
+    parsed = urlparse(url)
+    if parsed.scheme or parsed.netloc:
+        # Absolute URL (or protocol-relative ``//host/...``) -- out of
+        # scope; the host is the operator's target, not the spec's.
+        return ""
+    base = parsed.path.rstrip("/")
+    if not base or base == "/":
+        return ""
+    if not base.startswith("/"):
+        base = "/" + base
+    return base
+
+
+def _fold_server_base(base: str, path: str) -> str:
+    """Append an operation ``path`` to the relative server ``base``.
+
+    ``base`` is the normalised output of :func:`_server_base_path`
+    (leading ``/``, no trailing ``/``) or ``""``. Joins them into a
+    single path-template with exactly one separator slash, collapsing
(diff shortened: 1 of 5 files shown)


damir-topic:
@coderabbitai review

Issue #2068, opened 2026-06-24
Task: migration 0051 — retire the registered-stub-twin vcf-logs orphan migration 0049 (#2015) misses (0-op short stub at a divergent version)


damir-topic:
## Summary

Migration `0049` ("retire stale long-product `vcf-logs` orphan rows", #2001 / PR #2015) ran on the RDC v0.19.0 upgrade (db.revision `0050`) but **did not retire the orphan it was built to remove**. `GET /api/v1/connectors` still lists both:

```
product=vcf-logs  connector_id=vrli-rest-9.0.2  state=ingested   136 ops  6 groups   ← the v0.13.0-era orphan
product=vrli      connector_id=vrli-rest-9.0    state=registered   0 ops  0 groups   ← the canonical short stub
```

`0049`'s retire predicate only matches when a **per-op short twin at the same version** exists. The canonical short row here is a **0-op `registered` stub at a divergent version** (`9.0` vs the orphan's `9.0.2`), so it matches zero of the orphan's op rows and nothing is retired. The 136-op orphan survives every upgrade — exactly the undeletable / review-shadow dead weight #2015 set out to clear (the orphan is non-dispatchable because the dispatch surface keys on the short `parse_connector_id` product, and `meho.connector.delete` can't target a divergent-product row → #1910).

A new data-only migration `0051` is the cleanest fix (the head migration is `0050`; no DDL, self-idempotent).

## Current state

Read [`backend/alembic/versions/0049_retire_stale_vcf_logs_orphan_rows.py`](backend/alembic/versions/0049_retire_stale_vcf_logs_orphan_rows.py) — `_retire_table` issues one guarded `DELETE` per product split. The live-twin guard is the correlated `short_twin_exists` `EXISTS` probe:

```python
short_twin_exists = (
    sa.select(sa.literal(1)).select_from(twin).where(
        twin.c.tenant_id.is_(None),
        twin.c.product == short_product,
        twin.c.version == shim.c.version,                       # ← version-equality
        twin.c.impl_id == shim.c.impl_id,
        twin.c[op_discriminator] == shim.c[op_discriminator],   # ← per-op: op_id / group_key
    ).exists()
)
```

[...]

PR #2076, opened 2026-06-25
fix(connectors): fold relative servers[].url base into ingested op paths (#1796)

### LINKED ISSUE #1796: Connector dispatch drops the OpenAPI servers[].url base path — ops 404 when the spec uses a relative server base

damir-topic:
## Summary

- Ingest now honours an OpenAPI spec's **relative** `servers[].url` base. Previously the base was silently dropped, so a spec like vRLI's `servers: [{url: /api/v2}]` ingested `GET:/version` and dispatched to `host/version` → 404 instead of `host/api/v2/version` (#1796).
- `parse_openapi` folds the path component of a relative `servers[0].url` into each op's `op_id` and `path`. **Absolute** or **server-variable-templated** bases are left unfolded — meho dispatches against the operator-configured target host, never a spec-declared one (the SSRF / topology boundary).
- **Scope note (bigger than the ticket implied — please review):** the two shipped MEHO package-data specs had already baked the prefix into their path keys as a workaround for this bug (`vmware_rest_minimal.yaml` under `/api`, `sddc_manager_minimal.yaml` under `/v1`). A naive fold would have double-prefixed them, so this PR restores both to RFC-compliant relative paths. The fold reconstructs their op-ids **byte-for-byte** (`GET:/api/about`, `GET:/v1/releases/system`), so vmware/sddc dispatch is unchanged.

## Test plan

- [x] `mypy` (openapi.py) — clean
- [x] `.claude/hooks/code-quality.py --diff` — 0 blockers (pre-existing file-size on the 1100-line parser carries a scoped `# code-quality-allow`; splitting it is a separate refactor)
- [x] New parse-level tests: relative-base fold, no-servers no-op, root (`/`) no-op, absolute-base-not-folded, and composition with the #2066 `{+path}` bare-name keying
- [x] **Op-id invariance guard** (the load-bearing safety check for the spec edits): `test_connectors_sddc_manager_core_ops`, `test_connectors_sddc_manager_e2e`, `tests/acceptance/test_vmware_rest_dispatch_smoke`, `test_operations_ingest_catalog`, `test_connectors_vmware_rest_composites_l2_ingest_reconcile` — all green; the shipped connectors' op-ids are unchanged
- [x] Broader ingest sweep (register/delete/llm-groups/mcp-ingest/nsx/acceptance) — 137 passed, integration skipped-in-sandbox
- No OpenAPI snapshot delta (ingest-internal; op-ids unchanged; no FastAPI route/schema change)

## Out of scope

- Absolute `servers[].url` bases that name a different host, and server-variable templating — left unfolded by design (documented in `_server_base_path`).
- Splitting the 1100-line `openapi.py` parser — pre-existing debt, recorded as an adjacent finding.
- #2068 (migration 0051) — the other open log-sentry-adjacent task, parked separately.

Closes #1796

### DIFF
--- a/backend/src/meho_backplane/operations/ingest/openapi.py
+++ b/backend/src/meho_backplane/operations/ingest/openapi.py
@@ -1,5 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright (c) 2026 evoila Group
+# code-quality-allow: file-size — pre-existing 1000+-line parser; #1796 only
+# adds the servers-base fold helpers. Splitting this module by responsibility
+# is a separate refactor (tracked as an adjacent finding), out of scope here.
 
 """Pure-function OpenAPI 3.0/3.1 spec parser.
 
@@ -304,6 +307,7 @@ def parse_openapi(
     return list(
         _iter_operations(
             paths=paths,
+            server_base_path=_server_base_path(spec),
             component_schemas=cast(dict[str, Any], component_schemas),
             component_parameters=cast(dict[str, Any], component_parameters),
             component_responses=cast(dict[str, Any], component_responses),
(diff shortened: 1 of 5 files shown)


damir-topic:
@coderabbitai review

damir-topic:
Superseded by #2075 (merged 2026-06-25), which delivered the same `servers[].url` base-path fold and closed #1796. This PR was a near-simultaneous parallel auto-run on the same task; main already carries the fold logic (`_server_base_path` / `_fold_server_base`), so merging here would re-apply a shipped fix and conflicts with #2075's version of `openapi.py`. Closing as duplicate. The one distinct idea here — restoring the two shipped specs to RFC-relative paths + a `servers:` block so they exercise the fold — is left as optional future cleanup, not filed.

PR #2120, opened 2026-07-01
fix(ui): re-auth expired runbooks sessions instead of opacity banner

### LINKED ISSUE #2114: runbooks UI: `/ui/runbooks/<slug>` opacity_floor banner fires non-deterministically per browser for the SAME operator + IDENTICAL engine state — banner persists across hard-refresh, leaks across templates, and locks template authors out of viewing their own drafts
## Summary

A single operator (`tenant_admin`, JWT-confirmed same `sub` across browsers, same Keycloak account) viewing the SAME `/ui/runbooks/<slug>` URL with IDENTICAL engine state gets **DIFFERENT banner outcomes in different browsers** on v0.19.0:

- **Browser A** (regular): banner fires, steps hidden, persists across `Cmd+Shift+R` hard-refresh.
- **Browser A incognito**: initially after publishing → steps + Edit/Deprecate visible; later on re-navigation → banner fires.
- **Browser C** (entirely different browser, same Keycloak credentials, same JWT `sub`): steps visible, no banner.

Plus a **cross-template leak**: in browser A, the banner ALSO fires on an unrelated long-published template (a *different* `slug`, different content, different version, no active run for the operator on it), so the predicate is not per-template-clean either.

Engine state at probe time: **0 `in_progress` runs** for the operator across all templates. So the banner cannot honestly be reading "operator has an active run for this template" — there's no such run anywhere.

The combination breaks the basic UI authoring workflow: an author who creates a draft, navigates away, and returns cannot view their own steps. With the broken Preview button (separate issue) and the missing Edit-on-drafts affordance (separate issue), there is **no UI path for an author to view or modify a draft they just created**.

### Setup

1. Authenticate to the UI as `tenant_admin` (operator role: `tenant_admin`, realm role: `meho-admin`).
2. POST `/ui/runbooks/new` — create a fresh draft, e.g. `rb-test-disposable`, 3 manual steps, ConfirmVerify on each.
3. Engine accepts it as `v1, status: draft, created_by: <operator-sub>`.

### Browser A (regular) — reproduction

1. Submit the new template form → redirected to `/ui/runbooks/rb-test-disposable`.
2. Initially: steps visible (post-create grant).
3. Navigate away to any other page (e.g. `/ui/runbooks` list, `/ui/dashboard`).
4. Navigate back to `/ui/runbooks/rb-test-disposable`.
5. **Banner fires** — steps hidden. Banner text:

   > *"Step details are restricted. Step details are restricted until you complete a run of this template. Once you have a completed (or abandoned) run of `<slug>`, the full steps unlock here for post-run review. A tenant administrator can see them at any time."*

6. `Cmd+Shift+R` (hard refresh): **banner persists.**
7. Navigate to a different, long-published template at `/ui/runbooks/<other-slug>` (one the operator has never started a run on, OR has only an abandoned run on a prior version): **banner also fires there.**

### Browser A incognito — reproduction (state evolves during session)

1. Open incognito window on same browser. Authenticate to Keycloak with the same credentials.
2. Navigate to `/ui/runbooks/<draft-slug>`. Initially: **steps visible, no banner.** Click Publish — succeeds. **Edit + Deprecate buttons appear.** UI behaves normally.
3. Navigate away (any page).
4. Return to `/ui/runbooks/<slug>` (now `status: published`).
5. **Banner now fires** despite the session being only a few minutes old, no run started, no other action taken. The banner's appearance is a state transition that happens between steps 2 and 4.

### Browser C (different browser application) — reproduction (clean)

1. Open a completely different browser application (e.g. Safari if A was Chrome, or vice versa).
2. Authenticate to Keycloak with the same credentials.
3. Navigate to `/ui/runbooks/<slug>` (the same `slug` showing banner in A): **steps visible, no banner.**

### Engine state — probed via REST with the operator's bearer JWT

```
GET /api/v1/runbooks/runs
  → 5 runs total, 0 in state="in_progress"
    (3 abandoned + 2 completed, all from prior drives)

GET /api/v1/runbooks/templates/<slug>
  → status: "published" (or "draft" pre-publish)
    Includes all steps in the response.
```

### JWT introspection — claims confirmed IDENTICAL across all three browsers

- `sub`: the same `<uuid>` in all three.
- `preferred_username`: the same in all three.
- `email`: the same in all three.
- `tenant_role`: `tenant_admin` in all three.
- `realm_access.roles`: includes `meho-admin` + `meho-operator` in all three.

So the variable is **browser-bound state** (cookies / localStorage / sessionStorage / IndexedDB / HTTP cache / service worker), NOT the principal.

## Why this is a load-bearing defect

Three independent reasons the banner should never fire in any of these scenarios:

### 1. The predicate is structurally unsatisfiable for drafts

The banner says *"until you complete (or abandon) a run of this template"*. But drafts cannot be run — `POST /api/v1/runbooks/runs` with `template_slug` of a draft returns HTTP 400 *"every version of template '<slug>' is deprecated; cannot start a new run"* (or analogous draft-specific rejection). A tenant_admin who creates a draft has **zero possible action** to satisfy the predicate → banner is **permanent** on drafts. The opacity property has no meaning on a draft (drafts have no runs in progress; nothing to obscure).

### 2. The promised tenant_admin bypass doesn't exist

The banner explicitly says *"A tenant administrator can see them at any time"*. The tester IS `tenant_admin` (confirmed in JWT). Exhaustive UI walkthrough finds **no admin-bypass affordance**: no "Admin view" button, no URL hint (`?admin=true`, `/admin/runbooks/<slug>`), no toggle, no preference. **Already filed separately as `evoila/meho#2112`**; this finding reinforces that the bypass is critical (not nice-to-have) — without it, the entire authoring loop is broken.

### 3. The author should view their own work

The operator is `created_by` for the template. Denying the author view of the steps they just wrote inverts the natural authoring workflow:

```
write → preview → adjust → publish
```

With this defect (combined with broken Preview + no-Edit-on-drafts):

```
write → submit blind → navigate away → no way back → destroy + restart
```

### Cross-browser non-determinism implicates session-bound state

The bug is **not** a clean function of `(operator, template, engine_state)`. Same triple, different browsers → different banner outcome. The only remaining variable is browser-bound state. Three plausible causes (no claim of which is right):

- **(H1) BFF session cookie carries stale "active run id" state.** A run-start flow may set a cookie or session entry that isn't cleared on run completion / abort. The banner check keys on that cookie, which is true in some browsers (post-prior-run state) and false in others (clean state). Cross-template leak is consistent with this — a cookie set during a run on template X persists and triggers banner on template Y.
- **(H2) BFF reads a cached principal projection separate from the JWT.** The cached projection may carry a different `tenant_role` than the live JWT, failing the bypass check.
- **(H3) BFF check is keyed on browser localStorage/sessionStorage** that JS populates during the run-start flow. Different browser stores → different result.

H1 best matches the symptoms (cross-template leak; persistence across hard-refresh; per-browser variance; "state transition during session" as seen in browser A incognito between steps 2-4 above).

## Why this could affect other features

If H1 is right (BFF session caches that drift from engine state), the same pathology likely affects any UI surface with a "current-state" predicate that involves the operator's recent activity:

- **Approval gating** (operator's recent decisions cached)
- **Broadcast restrictions** (operator's recent broadcast activity)
- **Agent run-state** (similar shape to runbook runs)
- **Memory promotion** (operator's recent writes)

Worth a focused look at the BFF session/caching layer before assuming this is a runbook-local bug.

- **Option A — make opacity_floor server-side-only + read from engine.** The opacity predicate should evaluate using ONLY the engine's authoritative state (runs DB), not session-side caches. Remove any BFF session or cookie-carried run-state caches that influence the banner. Minimum-cost correctness fix.

- **Option B — implement the tenant_admin bypass.** Sibling fix to #2112's aspirational-banner finding. Even if the predicate were perfect, the bypass is the safety valve that prevents authoring lockout. Pair A + B for full coverage.

- **Option C — let the template author always view their own draft/template regardless of opacity state.** Opacity is meant to prevent operators from skipping ahead during a run; it has no meaningful goal against the template's own author. Add a `created_by == operator → permit` short-circuit.

- **Option D — narrow the predicate so it doesn't apply to drafts.** Drafts cannot be run; opacity has no meaning on them. The check should only fire on `status == "published"` templates with an active run for the requesting operator.

Options A + C + D together fix the underlying logic. Option B is the wording-promise fix.

## Verification

Once a fix lands, this reproduces in 10 minutes:

1. Authenticate as `tenant_admin` in two different browsers (same Keycloak account).
2. Create a fresh draft via `/ui/runbooks/new`.
3. Navigate away in browser 1, return. **Expectation**: steps still visible.
4. In browser 2 (different browser application), open the same URL. **Expectation**: identical render to browser 1.
5. Run `GET /api/v1/runbooks/runs` from CLI — confirm 0 `in_progress`.
6. Both browsers should show the steps. If either shows the banner with 0 active runs and the operator is `tenant_admin`, the bug persists.

## Related

- `#2112` — the `tenant_admin` bypass is aspirational; this finding shows the bypass is load-bearing for authoring, not just nice-to-have.
- `#2111` — the REST `GET /api/v1/runbooks/templates/<slug>` opacity_floor violation (REST returns full steps mid-run while UI applies a check at all); this finding shows the UI's check is also broken, just in a different direction.

## Out of scope

- The Preview button's htmx selector defect (separate issue — `Element.closest('.flex-wrap >> .runbook-step-preview')` fails because `>>` is the shadow-DOM piercing combinator, not valid in `closest()`).
- The missing Edit affordance on drafts (separate issue).
- Cross-tenant opacity (the bug reproduces within a single tenant; cross-tenant not tested).
- Other features potentially affected by the same session-cache pathology (call-out only; needs its own focused testing).

zdamir:
## Summary

- Fixes the non-deterministic runbooks opacity banner (#2114). The detail read
  surface recomputes the admin verdict per request by re-verifying the session's
  stored access token (`_resolve_role`); it swallowed **every** exception and
  degraded to a plain operator. So a genuine `tenant_admin` whose session token
  aged out (refresh failed → `session_expired`) silently landed on the
  opacity-restricted banner ("complete a run to unlock") instead of a re-login
  prompt — browser-dependent, since each session's token expires independently,
  and template authors got locked out of viewing their own drafts.
- The fix: re-raise the **terminal** `session_expired` 401 so the app-level
  handler (`ui/auth/errors.py`, G0.25 #1694) redirects the browser to
  re-authenticate. After re-login the fresh token lifts the real role and full
  steps render. **Transient / malformed-token** lifts still fail soft to the
  restricted view (unchanged), so the anti-enumeration posture and "never 5xx
  the read surface" guarantee hold.
- Scope note: no separate "author can view own draft" short-circuit is needed —
  drafting requires `tenant_admin` (the `draft`/`edit` role floor), so every
  author already takes the admin path once their session is valid; the re-auth
  fix restores that path.

## Test plan

- [x] `pytest tests/test_ui_runbooks_lifecycle.py tests/test_ui_runbooks_acceptance.py tests/test_ui_runbook_driver_opacity.py tests/test_runbook_templates_api.py` — **83 passed**
- [x] New: `test_resolve_role_reraises_session_expired` — terminal `session_expired` propagates (not swallowed to `None`)
- [x] New: `test_resolve_role_soft_fails_on_non_session_expired_401` — a non-expiry 401 (bad signature) still degrades soft to operator
- [x] New: `test_detail_session_expired_redirects_admin_to_reauth` — admin detail GET with an expired session → **302 → `/ui/auth/login?return_to=…`**, not the restricted banner
- [x] Regression preserved: the acceptance suite's plaintext-token `operator_soft_session` still degrades to the restricted view (non-`session_expired` path unchanged)

## Out of scope

- `resolve_role_probe` (`ui/routes/connectors/operator.py:183`) has the **same** swallow-and-degrade of `session_expired`, affecting the connectors detail page and the runbooks run-driver POST paths. Left untouched here (different surfaces than the reported read-surface banner); worth a follow-up so those surfaces also re-auth on terminal expiry rather than degrade.
- The broader "does the same session-cache pathology touch approvals / broadcast / agent / memory UI surfaces" call-out from the issue — a grep for the same pattern is warranted separately.

Closes #2114

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runbook access behavior when a session has expired, so users are redirected to sign in again instead of seeing a restricted or partial view.
  * Preserved the existing fallback experience for other temporary authorization failures, avoiding server errors.
  
* **Tests**
  * Added coverage for session-expired redirects and for distinguishing terminal login-expired errors from other 401 responses.

### DIFF
--- a/backend/src/meho_backplane/ui/routes/runbooks/routes.py
+++ b/backend/src/meho_backplane/ui/routes/runbooks/routes.py
@@ -51,10 +51,15 @@ notice. This matches the REST surface's ``403 detail="opacity_floor"``
 posture while keeping the console a navigable page rather than an error.
 
 The role lift fails **soft** to operator-level privileges (no admin) when
-the JWT round-trip can't complete (session row vanished mid-request, JWKS
-transiently unreachable) -- the same posture
+the JWT round-trip can't complete for a *transient* reason (session row
+vanished mid-request, JWKS transiently unreachable, a malformed stored
+token) -- the same posture
 :func:`meho_backplane.ui.routes.connectors.operator.resolve_role_probe`
-adopts. The restricted-detail render is the safe default, never a 5xx.
+adopts. The restricted-detail render is the safe default, never a 5xx. The
+one exception is a **terminal** ``session_expired`` (the token aged out and
+its refresh failed): that is re-raised so the app-level handler redirects to
+re-auth, rather than silently degrading a genuine ``tenant_admin`` onto the
+restricted banner (#2114).
 
 Tenant scoping
 --------------
@@ -90,7 +95,7 @@ from dataclasses import dataclass
 from typing import Annotated, Final, Literal
 
 import structlog
-from fastapi import APIRouter, Depends, HTTPException, Query, Request
+from fastapi import APIRouter, Depends, HTTPException, Query, Request, status
 from fastapi.responses import HTMLResponse
 
 from meho_backplane.auth.operator import Operator, TenantRole
@@ -107,6 +112,7 @@ from meho_backplane.runbooks.service import (
 from meho_backplane.settings import get_settings
 from meho_backplane.ui.auth.middleware import UISessionContext, require_ui_session
 from meho_backplane.ui.auth.refresh import (
+    SESSION_EXPIRED_DETAIL,
     load_fresh_session,
     verify_access_token_with_refresh,
 )
@@ -174,12 +180,23 @@ async def _resolve_role(session_ctx: UISessionContext) -> Operator | None:
     re-verifying -- so an expired-but-refreshable token lifts the real role
     instead of degrading the operator to the restricted view mid-session.
 
-    Fails **soft**: any hiccup (session row vanished between the middleware
-    check and here, JWKS transiently unreachable, refresh unavailable,
-    identity mismatch on the decoded token) returns ``None`` -- the caller
-    then treats the request as a plain operator (no admin privilege). The
-    opacity-floor branch is the safe default; an unavailable role lift must
-    never 5xx the read surface.
+    Fails **soft** for a *recoverable-later* hiccup (session row vanished
+    between the middleware check and here, JWKS transiently unreachable, a
+    malformed / bad-signature stored token, identity mismatch on the decoded
+    token): returns ``None`` so the caller treats the request as a plain
+    operator (no admin privilege). The opacity-floor branch is the safe
+    default; a transiently-unavailable role lift must never 5xx the read
+    surface.
+
+    The one exception is a **terminal** ``session_expired`` -- the BFF
+    session's access token aged out and its refresh failed, so the row is
+    dead and can never mint a fresh token. That must **not** degrade to the
+    restricted view: doing so silently drops a genuine ``tenant_admin`` onto
+    the opacity banner ("complete a run to unlock"), non-deterministically
+    per browser as each session's token expires (#2114). It is re-raised so
+    the app-level ``session_expired`` handler
+    (:mod:`meho_backplane.ui.auth.errors`) redirects the browser to
+    re-authenticate; after re-login the fresh token lifts the real role.
     """
     try:
         decrypted = await load_fresh_session(session_ctx.session_id)
@@ -191,6 +208,12 @@ async def _resolve_role(session_ctx: UISessionContext) -> Operator | None:
             expected_audience=settings.keycloak_audience,
         )
     except Exception as exc:
+        if (
+            isinstance(exc, HTTPException)
+            and exc.status_code == status.HTTP_401_UNAUTHORIZED
+            and exc.detail == SESSION_EXPIRED_DETAIL
+        ):
+            raise
         log.info(
             "ui_runbooks_role_lift_unavailable",
             session_id=str(session_ctx.session_id),
@@ -217,9 +240,11 @@ async def _is_admin(session_ctx: UISessionContext) -> bool:
     Thin wrapper over :func:`_resolve_role` returning just the admin verdict --
     the catalog + list-fragment renders need the boolean (to show / hide the
     lifecycle row actions) but not the full :class:`Operator`. Fails soft to
-    ``False`` (operator privileges) via :func:`_resolve_role`, so the row
-    actions are hidden whenever the role lift can't complete. The POST handlers
-    re-check server-side, so a hidden-but-forged action is still a 403.
+    ``False`` (operator privileges) for a transient lift failure, so the row
+    actions are hidden whenever the role lift can't complete. A terminal
+    ``session_expired`` propagates (via :func:`_resolve_role`) to the re-auth
+    redirect rather than degrading. The POST handlers re-check server-side, so
+    a hidden-but-forged action is still a 403.
     """
     operator = await _resolve_role(session_ctx)
     return operator is not None and operator.tenant_role == TenantRole.TENANT_ADMIN
--- a/backend/tests/test_ui_runbooks_lifecycle.py
+++ b/backend/tests/test_ui_runbooks_lifecycle.py
@@ -44,13 +44,15 @@ import warnings
 from collections.abc import Iterator
 from datetime import UTC, datetime, timedelta
 from typing import Any
+from unittest.mock import AsyncMock, patch
 
 import pytest
 import respx
 from cryptography.fernet import Fernet
-from fastapi import FastAPI
+from fastapi import FastAPI, HTTPException, status
 from fastapi.staticfiles import StaticFiles
 from fastapi.testclient import TestClient
+from starlette.exceptions import HTTPException as StarletteHTTPException
 
 from meho_backplane.auth.jwt import clear_jwks_cache
 from meho_backplane.auth.operator import TenantRole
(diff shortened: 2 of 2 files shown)


zdamir:
@coderabbitai review

zdamir (commented):

Independent review of the diff against issue #2114, plus CodeRabbit (clean — 0 actionable findings, 0 inline comments).

**Correctness (root cause).** `_resolve_role` now re-raises the terminal `session_expired` 401 so the app-level handler (`ui/auth/errors.py`) redirects to re-auth, instead of swallowing it and degrading a genuine `tenant_admin` onto the opacity banner. Verified the production exception type: `ui/auth/refresh.py:_raise_session_expired` raises **fastapi**'s `HTTPException` (import at `refresh.py:102`), which is exactly what the `isinstance(exc, HTTPException)` guard matches — the fix fires in production, not just under the test's patch.

**Scope.** Both `_resolve_role` call sites (`_render_detail`, `_is_admin` → catalog/list) are covered by the single change. Non-`session_expired` failures (bad signature, transient JWKS, malformed token) still fail soft to the restricted view — preserving the anti-enumeration posture and "never 5xx the read surface" guarantee. The acceptance suite's plaintext-token `operator_soft_session` degrade path is unchanged.

**Integration.** `ruff` / `ruff format --check` / `mypy` clean. `pytest` — 83 passed across `test_ui_runbooks_lifecycle`, `test_ui_runbooks_acceptance`, `test_ui_runbook_driver_opacity`, `test_runbook_templates_api`, including 3 new targeted tests (re-raise on expiry, soft-fail on non-expiry 401, detail route → 302 re-auth).

**Docs.** Module + `_resolve_role` + `_is_admin` docstrings updated to describe the soft-vs-terminal boundary; `docs/codebase/ui.md` high-level inventory unaffected.

### Adjacent (out of scope; non-blocking)
- `resolve_role_probe` (`ui/routes/connectors/operator.py:183`) has the **same** `session_expired` swallow-and-degrade, affecting the connectors detail page + the runbooks run-driver POST paths. Worth a follow-up so those surfaces also re-auth on terminal expiry rather than degrade.

PR #2121, opened 2026-07-01
fix(ui): re-auth expired sessions in resolve_role_probe, not degrade

zdamir:
## Summary

- Fixes the sibling of [`evoila/meho#2114`](https://github.com/evoila/meho/issues/2114) / PR #2120 on the **connectors** read surface. `resolve_role_probe` (`ui/routes/connectors/operator.py`) recomputes the admin verdict per request by re-verifying the session's stored access token, wrapped in a broad `except Exception` that returned a **no-privileges probe** on *any* failure — swallowing the terminal `session_expired` 401. So a genuine `tenant_admin` whose session token aged out (refresh failed) was silently degraded to a no-privileges render instead of a re-login prompt, on **two** surfaces: the connectors detail page (`GET /ui/connectors/{name}`) and the runbooks run-driver POSTs (`driver.py`, `driver_render.py`) that read `caller_is_admin` from this probe.
- The fix: re-raise the **terminal** `session_expired` so the app-level handler (`ui/auth/errors.py`) redirects the browser to re-authenticate. **Transient / malformed-token** failures still fail soft to the no-privileges probe — the read surface never 5xxes.
- Contained to `resolve_role_probe`: the strict `resolve_operator_or_403` write gate already propagated `session_expired` (it doesn't catch) and is unchanged; `lift_operator_from_session` is unchanged.

## Test plan

- [x] `pytest tests/test_ui_connectors_view.py` — **38 passed**
- [x] New: `test_resolve_role_probe_reraises_session_expired` — terminal `session_expired` propagates (not swallowed to a no-priv probe)
- [x] New: `test_resolve_role_probe_soft_fails_on_non_session_expired_401` — a non-expiry 401 (bad signature) still degrades soft
- [x] New: `test_detail_session_expired_redirects_to_reauth` — connectors detail GET with an expired session → **302 → `/ui/auth/login?return_to=…`**, not a degraded render

## Out of scope

- Other UI surfaces (approvals / broadcast / agent / memory) that may share the same broad-`except` pattern — a separate audit; this PR fixes the two known `resolve_role_probe` consumers.

Refs evoila-bosnia/meho-internal#143

### DIFF
--- a/backend/src/meho_backplane/ui/routes/connectors/operator.py
+++ b/backend/src/meho_backplane/ui/routes/connectors/operator.py
@@ -52,6 +52,7 @@ from meho_backplane.auth.operator import Operator, TenantRole
 from meho_backplane.settings import get_settings
 from meho_backplane.ui.auth.middleware import UISessionContext, require_ui_session
 from meho_backplane.ui.auth.refresh import (
+    SESSION_EXPIRED_DETAIL,
     load_fresh_session,
     verify_access_token_with_refresh,
 )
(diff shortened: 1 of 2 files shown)


zdamir:
@coderabbitai review

zdamir (commented):

Independent review against issue #143. **CodeRabbit did not complete within the wait window** (`bot_review.received: false`); the identical-pattern sibling PR #2120 received a clean CodeRabbit pass, and this diff is the same shape on `resolve_role_probe`.

**Correctness.** `resolve_role_probe` now re-raises the terminal `session_expired` 401 (→ app-level re-auth redirect) instead of swallowing it into a no-privileges probe. Verified the production exception type: `ui/auth/refresh.py:_raise_session_expired` raises **fastapi**'s `HTTPException` (`refresh.py:102`), which the `isinstance(exc, HTTPException)` guard matches — the fix fires in production. Both consumers benefit: the connectors detail page (`GET /ui/connectors/{name}`) and the runbooks run-driver POSTs (which read `caller_is_admin` from this probe).

**Scope.** Contained to `resolve_role_probe`. The strict `resolve_operator_or_403` write gate already propagated `session_expired` (it doesn't catch) and is unchanged; `lift_operator_from_session` unchanged. Non-`session_expired` failures (bad signature, transient JWKS, malformed token) still fail soft to the no-privileges probe — the read surface never 5xxes.

**Integration.** `ruff` / `ruff format --check` / `mypy` clean. `pytest tests/test_ui_connectors_view.py` — 38 passed, including 3 new: probe re-raises on `session_expired`; probe soft-fails on a non-expiry 401; connectors-detail GET with an expired session → 302 → `/ui/auth/login`.

**Docs.** `resolve_role_probe` docstring updated to describe the soft-vs-terminal boundary; `docs/codebase` inventory unaffected.

### Adjacent (out of scope; non-blocking)
- Other UI surfaces (approvals / broadcast / agent / memory) may share the same broad-`except` session-expiry swallow — a separate audit.

PR #2122, opened 2026-07-01
fix(ui): repair runbooks authoring surface (preview, draft edit, body)

zdamir:
## Summary

Repairs three defects on the `/ui/runbooks` **authoring** surface (evoila/meho#2117 — D1, D2, D3). D4 (slug doubling) and D5 (post-deprecate list refresh empties) from the same report are **not reproducible from the code** and are tracked separately (see Out of scope).

- **D1 — Preview was dead (invalid selector).** The step-body live-preview used `hx-target="closest .flex-wrap >> .runbook-step-preview"`; `>>` is the shadow-DOM piercing combinator, invalid in htmx target resolution / `Element.closest()`, so Preview threw a JS `SyntaxError` and never fired. Retargeted to `next .runbook-step-preview` — the preview div is the next match in document order after each step's textarea (verified against the htmx `hx-target` docs; `find` reaches only descendants, which the cousin preview isn't).
- **D2 — no Edit affordance on drafts.** A draft detail page offered only Publish, forcing authors to publish a possibly-broken first cut to get an Edit button. Added an in-place **Edit** link on the draft branch routing to the shared `/ui/runbooks/<slug>/edit` editor; the engine's `update_or_fork` mutates a draft in place (forks only a published version), so no vestigial version is left behind.
- **D3 — empty step body silently accepted.** `ManualStep.body` / `OperationCallStep.body` were plain `str`; `""` passed. Added `min_length=1` so an empty body is a 422 at request validation (mirrors the existing non-empty title/description checks; covers both REST and MCP entry points via the shared schema).

## Test plan

- [x] `ruff check` / `ruff format --check` / `mypy` — clean
- [x] New: `test_editor_new_step_preview_uses_valid_hx_target` — the rendered editor uses `next .runbook-step-preview`, and the invalid `.flex-wrap >> .runbook-step-preview` selector is gone.
- [x] New: `test_detail_draft_shows_edit_affordance` — a draft detail exposes an Edit link to `/ui/runbooks/<slug>/edit` alongside Publish.
- [x] New: `test_draft_rejects_empty_step_body_422` — a POST with an empty step body is a 422 at request validation.
- [x] `pytest` — **201 passed** across the runbook suites (templates / editor / lifecycle / run-service / template-service / acceptance / MCP-templates / MCP-runs / driver-opacity); D3's `min_length` broke no existing fixture.

## Out of scope

[...]

Issue #2117, opened 2026-06-30
Initiative: runbooks UI authoring-surface cleanup — 5 create/preview/edit/deprecate defects


zdamir:
## Summary

A manual walkthrough of the `/ui/runbooks` **authoring** surface (create new template → preview → save → edit → publish → deprecate, plus validation negative-cases) surfaced **five distinct defects** that together break the basic authoring workflow. Each is independently filable; bundling because they're all on the same UI surface and a fix sweep is naturally one PR.

Observed live on v0.19.0 (chart `0.19.0`).

### 1. Preview button throws JS error — invalid CSS selector in `Element.closest()`

Clicking **Preview** on `/ui/runbooks/new` (or the edit form) does nothing visible. Browser console:

```
Uncaught SyntaxError: Failed to execute 'closest' on 'Element':
  '.flex-wrap >> .runbook-step-preview' is not a valid selector.
    at g (htmx.min.js:1:7279)
    at m (htmx.min.js:1:7930)
    at ue (htmx.min.js:1:9014)
    at Ee (htmx.min.js:1:10070)
    at he (htmx.min.js:1:41369)
```

The `>>` combinator is the **shadow-DOM piercing combinator**, valid in some newer CSS contexts but **not valid in `Element.closest()`** (which accepts only standard selectors). The htmx attribute somewhere on the page (almost certainly an `hx-target=".flex-wrap >> .runbook-step-preview"` on the Preview button or its container) trips `closest()` and the entire preview pipeline fails.

**Fix shape**: replace the selector with a valid alternative — either a descendant selector (` ` space) or a unique ID/class on the target element. Then the preview should fire.

The Preview button being dead is a real UX problem: the new-template form has no other way to inspect what the resulting template will look like before submit. An author writing markdown bodies has no way to verify their formatting without committing to Create Draft.

### 2. UI offers no Edit affordance on drafts

A `status: draft` template viewed at `/ui/runbooks/<slug>` exposes **no Edit button**. The engine cleanly accepts `PATCH /api/v1/runbooks/templates/<slug>` on drafts (we verified: the PATCH mutates in place, no fork — `version` stays the same, content updates). But the UI hides that affordance entirely.

The Edit button only appears AFTER publish — which means editing requires the operator to first publish the draft (committing a possibly-broken first cut), then edit (which forks to v=max+1 draft per the engine contract), then publish v2, leaving v1 as a vestigial published row in history.

**Workflow today**:

```
operator writes draft → submit → wants to fix typo
[...]

PR #2125, opened 2026-07-01
fix(ui): coerce empty runbooks list target_kind filter to no-filter

zdamir:
## Summary

Fixes **D5** of evoila/meho#2117: deprecating a template from the list view blanked the whole catalog.

- **Root cause:** the `/ui/runbooks` + `/ui/runbooks/list` `target_kind` query param lacked the `EMPTY_STR_TO_NONE` coercion the `status` param already has. An empty `?target_kind=` reached `RunbookTemplateService.list_templates` as `""` and applied `WHERE target_kind = ''` (`service.py`: `if filter_.target_kind is not None`), matching nothing.
- **Trigger:** the post-deprecate `runbooks-refresh` re-fetches `/ui/runbooks/list` with `hx-include="[name='status'], [name='target_kind']"` (`index.html`), so an empty target-kind filter input submitted `?target_kind=` and emptied the list after every deprecate. (Same footgun class as the topology empty-kind filter.)
- **Fix:** coerce empty → `None` on both routes (mirroring `status`). The `max_length` guard moves onto the inner `str` branch via `StringConstraints`, so it isn't applied to the `None` the `EMPTY_STR_TO_NONE` `BeforeValidator` yields — a bare `Query(max_length=...)` on the nullable field raises `TypeError` on `None` (verified: that was the initial failure, fixed by the branch-scoped constraint).

## Test plan

- [x] New `test_runbooks_list_empty_target_kind_filter_is_200_unfiltered`: `?target_kind=` **and** the exact refresh shape `?status=&target_kind=` (both empty) return the full catalog (both seeded templates present) — fails on `main` (empty list), passes with the fix.
- [x] Existing `test_runbooks_list_filters_by_target_kind` (non-empty still narrows) + `test_runbooks_list_empty_status_filter_is_200_unfiltered` still green.
- [x] `pytest tests/test_ui_runbooks_list.py` — 21 passed.

## Out of scope

- **D4 (slug doubling):** not reproducible from the code (mode-exclusive `name="slug"` inputs, no auto-slugify/concat handler, single `hx-post` form serialization). Tracked as `evoila-bosnia/meho-internal#145` pending a live DevTools trace.
- **D1/D2/D3** already shipped (#2122).

Refs evoila/meho#2117

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of empty runbook filters so requests with an empty `target_kind` now return the full catalog instead of failing or filtering unexpectedly.
  * Ensured combining empty `status` and `target_kind` also returns the unfiltered runbooks list at HTTP 200.

* **Tests**
  * Added coverage for empty filter values on the runbooks list page to verify the expected results are shown.

### DIFF
--- a/backend/src/meho_backplane/ui/routes/runbooks/routes.py
+++ b/backend/src/meho_backplane/ui/routes/runbooks/routes.py
@@ -97,6 +97,7 @@ from typing import Annotated, Final, Literal
 import structlog
 from fastapi import APIRouter, Depends, HTTPException, Query, Request, status
 from fastapi.responses import HTMLResponse
+from pydantic import StringConstraints
 
 from meho_backplane.auth.operator import Operator, TenantRole
 from meho_backplane.runbooks.run_service import RunbookRunService
@@ -433,7 +434,11 @@ def build_runbooks_router() -> APIRouter:
         request: Request,
         session: UISessionContext = _require_session,
         status: Annotated[_StatusFilter | None, EMPTY_STR_TO_NONE, Query()] = None,
-        target_kind: str | None = Query(default=None, max_length=_MAX_TARGET_KIND_LENGTH),
+        target_kind: Annotated[
+            Annotated[str, StringConstraints(max_length=_MAX_TARGET_KIND_LENGTH)] | None,
+            EMPTY_STR_TO_NONE,
+            Query(),
+        ] = None,
     ) -> HTMLResponse:
         return await _render_index(request, session, status, target_kind)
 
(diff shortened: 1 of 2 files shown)


zdamir:
@coderabbitai review

zdamir (commented):

Independent review of `f3014bba`; CodeRabbit clean (0 actionable findings).

**Correctness.** The `target_kind` list-filter param on both `/ui/runbooks` and `/ui/runbooks/list` now applies `EMPTY_STR_TO_NONE` (mirroring `status`), so an empty `?target_kind=` resolves to the no-filter sentinel instead of `WHERE target_kind = ''`. This is the exact post-deprecate `runbooks-refresh` shape (`hx-include="[name='status'], [name='target_kind']"`) that emptied the catalog (#2117 D5). The `max_length` guard is preserved on the inner `str` branch via `StringConstraints`, so it is not applied to the `None` the `BeforeValidator` yields (a bare `Query(max_length=)` on the nullable field raises `TypeError` on `None` — verified).

**Scope.** Contained to the two route params + a regression test. D4 (slug doubling) is not reproducible from the code and stays tracked as `meho-internal#145`; D1/D2/D3 shipped in #2122. `Refs` (not `Closes`) #2117.

**Integration.** `ruff` / `ruff format --check` / `mypy` clean. `pytest tests/test_ui_runbooks_list.py` — 21 passed, incl. a new test asserting `?target_kind=` and `?status=&target_kind=` (the refresh shape) both return the full catalog, plus the unchanged non-empty-filter narrowing.

Issue #129, opened 2026-05-09
Initiative: G2.6 meho CLI binary — Go module, login/status/version, multi-platform release, signed


damir-topic:
Parent goal: evoila/meho#98
Blocked by: evoila/meho#100 (stack ADR — locks cobra + oapi-codegen + 99designs/keyring); evoila/meho#102 (cosign keyless ADR — release artefacts use same identity claim as image and chart); G2.2 (#108, the JWT validation contract that `meho login` produces tokens against); G2.5-T2 (#125, OpenAPI spec the generated client consumes); evoila/meho#100 also locks the OpenAPI 3.1 seam contract

## Scope

Ship the **`meho` operator-facing CLI binary** in Go: single static binary, multi-platform (linux/macos × amd64/arm64), server-driven subcommand discovery (forward-prepared for v0.2+ operations), structured output (human default, `--json` flag), keychain-stored auth tokens, cosign-signed release artefacts.

**Area**: new `cli/` directory at repo root (Go module). New `.github/workflows/cli-release.yml`. Extension to top-level `README.md`.

This is the operator's **only interface to MEHO in v0.1** (no web UI per Goal #98 scope). Every Goal #98 acceptance criterion that mentions `meho login` or `meho status` ultimately routes through this Initiative.

## Motivation

The CLI is a tightly-bounded surface. Three commands ship in v0.1: `meho login`, `meho status`, `meho version`. Goal #98 adds two structural requirements:

1. **Server-driven `--help`** — adding operations to the backplane (post-Goal-2) shouldn't require a new CLI binary release. Implementation: CLI fetches a subcommand manifest from the backplane and registers dynamic cobra subcommands at runtime. v0.1 backplane returns an empty manifest (no extra ops), so this is scaffold-only — but the scaffold has to be in place from day 1 or the v0.2 operation rollouts force CLI redistribution.
2. **Single static binary** — operators download one tarball, no Python runtime, no dependencies. The Goal #98 stack-locking decision (Python backplane + Go CLI) was made specifically to preserve this property; this Initiative is what redeems it.

Output discipline (per Goal #98 spec section 5): default human-readable summary; `--json` for structured output that agents can parse; clear `auth_expired` / `unreachable` structured error responses with non-zero exit codes; secrets never logged. **Output discipline is mandatory, not optional** — the dogfooding consumer's `install.sh` parses `meho status --json` to verify the federation chain.

## Definition of done

[...]

PR #2126, opened 2026-07-01
fix(bind9): require approval for dangerous config-apply ops (close four-eyes bypass)

zdamir:
## What

`bind9.config.apply_file` and `bind9.config.apply_views` shipped
`safety_level="dangerous"` + `requires_approval=False`. This PR flips both to
`requires_approval=True` so they are four-eyes-gated on **every** principal
path — not just the agent path.

## Why

The non-agent policy gate (`operations/_validate.py:_non_agent_verdict`) keys
**only** on `requires_approval`, not `safety_level`:

```python
if descriptor.requires_approval:
    return PermissionVerdict.NEEDS_APPROVAL
return PermissionVerdict.AUTO_EXECUTE   # safety_level logged, not consulted
```

The agent path floors `dangerous`→`NEEDS_APPROVAL` via `_SAFETY_CEILING`
(`auth/permissions.py`), but a human `tenant_admin` (or any non-agent
principal) executing a `dangerous`-class op registered `requires_approval=False`
got `AUTO_EXECUTE` — no approval gate. Both bind9 ops perform full DNS-config /
view-tree **replacement**, so a single operator could overwrite live DNS with
no four-eyes step. Setting `requires_approval=True` routes them through the
existing park→approve→resume flow for all principals.

## Tests

- New invariant `test_no_dangerous_op_bypasses_approval` — pins that **no**
  `dangerous` bind9 op ships `requires_approval=False`, so a future write op
  can't silently reintroduce the gap. (Scoped to `BIND9_OPS`; see the
  connector-wide sweep below for the other connectors.)
- Updated `test_safety_levels_pin_to_spec` to the per-op contract
  (apply_file/apply_views → approval-required; backup/reload → not).
- The gate mechanism itself is already covered by
  `test_dispatch_human_requires_approval_op_queues` (human + `requires_approval=True`
  → `awaiting_approval`), so the metadata flip + that test together prove the
  bind9 ops now park for a human.
- `test_connectors_bind9_config.py` + `test_broadcast_credential_mint_dispatch.py`:
  **49 passed, 3 skipped** (the 3 skips are the Redis-gated credential-mint tests —
  see deferral). `ruff` / `ruff format --check` / `mypy` clean.

Connector-wide sweep (`grep -rnE "safety_level.*dangerous|requires_approval"
backend/src/meho_backplane/connectors/`): argocd, kubernetes, vmware-rest
composites, and github composites already gate every `dangerous` op with
`requires_approval=True`. bind9 was the only `dangerous`-with-bypass offender.

## Deferred — `harbor.robot.create` (remaining #129 work)

The task also flagged `harbor.robot.create` (credential-mint, `caution`).
Gating it **parks** the op, which breaks the `credential_mint` broadcast-redaction
tests (`test_broadcast_credential_mint_dispatch.py`) — those dispatch
`robot.create` as a human operator and assert `status=="ok"` + a redacted
broadcast, i.e. they require the op to **execute** to verify redaction. Gating
would need them reworked through the approve→resume flow, and they **skip**
without `BROADCAST_REDIS_URL`, so that rework can't be fail-closed-verified in
this sandbox. Deferring keeps this PR green-in-CI. Recommend a follow-up pairing
the harbor flip with a Redis-enabled rework of those three tests.

Refs evoila-bosnia/meho-internal#129

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated BIND9 configuration actions so the more sensitive config changes now require approval before they can run.
  * Added safeguards to ensure risky BIND9 operations can’t bypass the approval step.

* **Tests**
  * Expanded coverage to verify approval requirements for key BIND9 actions.
  * Added checks to prevent dangerous operations from being marked as approval-free.

### DIFF
--- a/backend/src/meho_backplane/connectors/bind9/ops_config.py
+++ b/backend/src/meho_backplane/connectors/bind9/ops_config.py
@@ -1364,7 +1364,7 @@ CONFIG_OPS: tuple[Bind9Op, ...] = (
         group_key="config",
         tags=("write", "config", "atomic-apply"),
         safety_level="dangerous",
-        requires_approval=False,
+        requires_approval=True,
         llm_instructions=BIND9_CONFIG_APPLY_FILE_LLM_INSTRUCTIONS,
     ),
     Bind9Op(
@@ -1384,7 +1384,7 @@ CONFIG_OPS: tuple[Bind9Op, ...] = (
         group_key="config",
         tags=("write", "config", "atomic-apply"),
         safety_level="dangerous",
-        requires_approval=False,
+        requires_approval=True,
         llm_instructions=BIND9_CONFIG_APPLY_VIEWS_LLM_INSTRUCTIONS,
     ),
     Bind9Op(
--- a/backend/tests/integration/test_g3_4_bind9_e2e.py
+++ b/backend/tests/integration/test_g3_4_bind9_e2e.py
@@ -989,6 +989,14 @@ async def test_dispatch_config_apply_file_writes_state_before_after(
 ) -> None:
     """``bind9.config.apply_file`` writes a fragment + audit row carries
     state_before / state_after.
+
+    ``apply_file`` is a dangerous, four-eyes-gated op (#129): a lone-operator
+    dispatch parks awaiting approval. ``_approved=True`` is the flag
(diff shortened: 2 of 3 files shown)


zdamir:
@coderabbitai review

zdamir:
@coderabbitai review

Issue #1908, opened 2026-06-19
datastore.usage composite hard-fails when the optional vm-placement enrichment sub-call errors (filter.datastores 400 on an 8.0 vCenter)


damir-topic:
Parent initiative: #1911
Parent goal: #221

Was using `vmware.composite.datastore.usage` as a quick "which datastores are filling up" report across a fleet (it's a lovely fit for that — one call instead of 1+N). It errored:

```
call_operation vmware.composite.datastore.usage  ->
  connector_error: RuntimeError
  "composite sub-op 'GET:/vcenter/vm' returned status='error': connector_error: HTTPStatusError"
```

Dug in by replaying the sub-ops directly. The datastore listing + capacity/free/type all come back fine — it's the per-datastore VM-placement enrichment that fails:

```
GET:/vcenter/vm  params={"filter.datastores": ["datastore-345"]}  ->
  400 Bad Request for https://<vcenter>/api/vcenter/vm?filter.datastores=datastore-345
```

The bare `GET:/vcenter/vm` (no filter) works fine — 115 VMs. So it's specifically the `filter.datastores` query param this vCenter rejects. My target is a vSphere 8.0 vCenter and the connector is the 9.0 spec, so I'm fairly sure it's a version skew in the VM-list filter param name (the unprefixed `datastores=` form is accepted with a 200 but appears to be ignored — returns the full set — so I don't think that's simply "the right param" either; I didn't chase it further).

Two suggestions, in priority order:

1. **Make the vm-placement enrichment best-effort.** The composite has already successfully read capacity / free / type for every datastore by the time the VM lookup runs — that's the data the "what's about to fill up?" use case actually needs. When the enrichment sub-call fails, it'd be much more useful to still return the datastore rows (capacity/free/type) and just drop or null `vm_count`/`vm_names` (ideally with a per-row note that enrichment was skipped) rather than failing the whole composite. Right now an optional enrichment leg sinks the core report.

2. **Surface the sub-op's structured error at the composite level.** The composite only told me `RuntimeError` / "sub-op returned status='error'"; the actual `400` + the offending URL only showed when I replayed the sub-op myself. Bubbling the sub-op's error (status code + URL) up into the composite's error envelope would've saved the dig.

[...]

PR #2175, opened 2026-07-06
fix(connectors): carry datastore capacity via list-row fallback (#2078)

### LINKED ISSUE #2078: vmware.composite.datastore.usage returns capacity=null for every datastore (only free_space populated)
Picking up where #1908 left off: the best-effort vm-placement fix from that issue landed and works — `vmware.composite.datastore.usage` now returns rows + a per-row `enrichment_note` instead of hard-failing. Nice. But on our current deploy the composite comes back with `capacity` **null** for every row, so you get free bytes but no percent-full — which is the one number a "what's about to fill up?" report exists to produce.

Live call, 31 datastores, every one of them:

```
call_operation vmware.composite.datastore.usage   target=<an 8.0.3 vCenter>  ->
status: ok, 31 rows
{
  "name": "<ds>", "type": "VMFS",
  "capacity": null,             <-- always null
  "free_space": 1753912180736,  <-- fine
  "vm_count": null, "vm_names": null,
  "enrichment_note": "vm-placement enrichment skipped: ... GET:/vcenter/vm?filter.datastores=... 400 ..."
}
```

The JSONFlux handle's reported schema agrees — `capacity: null samples=[null, null, null]` — so it isn't just my row sampling.

A couple of notes:

- #1908 explicitly observed "the datastore listing + capacity/free/type all come back fine", so this looks like a regression (or environment/version-specific) since 2026-06-19. The raw `GET:/vcenter/datastore` detail read returns capacity directly, so the data is there — the composite just isn't carrying it through. My target is an 8.0.3 vCenter against the 9.0 connector spec, so a version-skew in how capacity is mapped off the list response is my first guess (same flavour as the `filter.datastores` skew #1908 chased).
- The vm-placement leg is still 400ing on this 8.0 vCenter exactly as #1908 described — that's the known best-effort skip, not a new complaint. Flagging only because with capacity null *and* placement null, the composite currently returns little beyond name + free bytes.

Without capacity I can't compute %full, so I fell back to the raw datastore detail again. Filing so the curated composite carries capacity through — that's the whole reason to reach for it over the raw 1+N reads.

Context: dogfooding the VMware read surface as the rdc-hetzner-dc deployment's backplane (evoila-bosnia/claude-rdc-hetzner-dc).

The symptom (`capacity=null` on every row, `free_space` fine) is real, but both the reporter's hypothesis (capacity mapped off the LIST response / version-skewed / a regression since #1908) and the first-pass triage guess ("the detail read fetches only `free_space`, not `capacity`") are wrong when checked against the code.

`datastore_usage_composite` sources **both** fields from the same per-datastore DETAIL payload, not the list:

- `backend/src/meho_backplane/connectors/vmware_rest/composites/_read.py:599-608` — the composite dispatches `GET:/vcenter/datastore/{datastore}` (`_OP_GET_DATASTORE`, line 107), unwraps it to `detail_payload`, then does `capacity = detail_payload.get("capacity")` (line 607) **and** `free_space = detail_payload.get("free_space")` (line 608). They are adjacent `.get()` reads off the identical dict — capacity is fetched and merged (line 613). The existing happy-path test proves it: `backend/tests/test_connectors_vmware_rest_composites_read.py:451-492` stubs the detail read with `capacity` present and asserts `"capacity": 100` on the aggregated row.

So this is **not** a missing fetch and **not** a list-vs-detail mapping bug in the sense the reporter framed. Since `free_space` (read one line below `capacity`, from the *same* payload) comes back populated on the live 8.0.3 vCenter while `capacity` is null, the live **detail** (`Datastore.Info`) payload that vCenter returns is not carrying a literal `capacity` key that the `.get("capacity")` can see — the field is absent (or spelled/nested differently) on this build's detail Info, whereas `free_space` is present. (The vcsim-only `capacity` skew, govmomi#1121, is a simulator unit-conversion artifact and does not match a real vCenter returning a correct `free_space` with a null `capacity`.)

The load-bearing fact for the fix: the datastore **list** row already carries capacity. The MEHO-authored `DatastoreSummary` schema exposes both `free_space` and `capacity` (`backend/src/meho_backplane/operations/ingest/specs/vmware_rest_minimal.yaml:241-246`), and the live `GET:/vcenter/datastore` list shape does too. But the composite throws the list row's capacity away: the loop uses `entry` (the list row) only for `name`/`type` (`_read.py:611-612`) and reads capacity/free exclusively from the detail payload. The residual fix is to make capacity resilient — fall back to the list-row capacity (`entry.get("capacity")`) when the detail payload omits it (same for `free_space` for symmetry), so the composite carries capacity through on builds whose detail `Datastore.Info` drops it. This is a real defect: the number is available on a payload the composite already fetched and is silently discarded.

Note: the response schema already permits both shapes — `capacity`/`free_space` are typed `["integer","null"]` (`schemas.py:415-426`), so no schema change is required; the fix is in the row-builder merge.

Parent goal: #221

## Acceptance criteria

- [ ] In `datastore_usage_composite` (`backend/src/meho_backplane/connectors/vmware_rest/composites/_read.py`), `capacity` falls back to the list-row value (`entry.get("capacity")`) when the per-datastore detail payload omits/nulls it, so a detail `Datastore.Info` without a `capacity` key still yields a non-null `capacity` whenever the list row carried one.
- [ ] `free_space` is given the same detail-then-list fallback for symmetry, and still resolves to the detail value when the detail payload provides it (no regression to the existing happy path).
- [ ] A new test asserts that when the detail read returns `free_space` but NO `capacity` key (the live 8.0.3 shape), the aggregated row has a non-null `capacity` sourced from the list row and the correct `free_space`.
- [ ] The existing detail-wins test (`backend/tests/test_connectors_vmware_rest_composites_read.py:451-492`) still passes unchanged — when the detail payload supplies `capacity`, the detail value is used.
- [ ] A test covers the both-missing case: when neither the detail payload nor the list row carries `capacity`, the row's `capacity` is `null` (no crash, schema-valid per `["integer","null"]`).

Parent initiative: #2148

damir-topic:
## Summary
- `vmware.composite.datastore.usage` returned `capacity=null` for every datastore on some vCenter builds (observed on an 8.0.3 vCenter against the 9.0 spec): the row-builder read `capacity`/`free_space` **only** from the per-datastore detail `Datastore.Info`, so a build whose detail omits the `capacity` key (while still populating `free_space`) surfaced a null capacity — leaving `%`-full uncomputable off the composite alone.
- The load-bearing fact from the triage: the `GET:/vcenter/datastore` **listing row** already carries `capacity`/`free_space`, and the composite already fetched it. This change sources both fields from the detail payload with a **fallback to the listing row** (`entry.get(...)`) when the detail value is absent — detail still wins when present; `null` only when neither source carries it.
- No schema change: `capacity`/`free_space` are already typed `["integer","null"]`. Their schema descriptions and `docs/codebase/connectors-vmware-rest.md` are updated to reflect the detail-then-list source.

## Test plan
- [x] `ruff check` — clean (`_read.py`, `schemas.py`, test file)
- [x] `ruff format --check` — clean (3 files formatted)
- [x] `mypy src/.../composites/` — Success, no issues (7 source files)
- [x] `pytest tests/test_connectors_vmware_rest_composites_read.py` — 27 passed
- [x] `pytest -k composite` — 216 passed, 2 skipped
- [x] **AC1/AC3** new-shape: detail returns `free_space` but no `capacity` key → aggregated `capacity` sourced from the list row (non-null), `free_space` from detail — `test_datastore_usage_capacity_falls_back_to_list_row_when_detail_omits_it`
- [x] **AC2/AC4** detail-wins unchanged: existing `test_datastore_usage_three_ops_per_datastore_aggregates_correctly` passes; new `test_datastore_usage_detail_capacity_wins_over_list_row` proves detail value beats a differing list-row value
- [x] **AC5** both-missing: neither detail nor list row carries `capacity` → row `capacity` is `null`, no crash, schema-valid — `test_datastore_usage_capacity_null_when_neither_detail_nor_list_carry_it`

## Out of scope
- The vm-placement leg still best-effort-400s on 8.0 vCenters (`filter.datastores` skew) — that is the known #1908 behavior, not a new complaint.
- No FastAPI route/schema touched → no OpenAPI snapshot regen required (composite result schema is an op-result schema, not a route model).

Closes #2078

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved datastore usage values so capacity and free space are retained when detailed data is incomplete.
  * Prevented valid capacity information from being dropped if it is available in the datastore listing.
  * Added coverage for missing, present, and overriding capacity values to ensure consistent results.

* **Documentation**
  * Clarified how datastore usage values are sourced and when fallback values are used.

### DIFF
--- a/backend/src/meho_backplane/connectors/vmware_rest/composites/_read.py
+++ b/backend/src/meho_backplane/connectors/vmware_rest/composites/_read.py
@@ -604,8 +604,21 @@ async def datastore_usage_composite(
             )
         )
         detail_payload = _unwrap_value(detail)
-        capacity = detail_payload.get("capacity") if isinstance(detail_payload, dict) else None
-        free_space = detail_payload.get("free_space") if isinstance(detail_payload, dict) else None
+        detail_capacity = (
+            detail_payload.get("capacity") if isinstance(detail_payload, dict) else None
+        )
+        detail_free_space = (
+            detail_payload.get("free_space") if isinstance(detail_payload, dict) else None
+        )
+        # The per-datastore detail ``Datastore.Info`` is the primary source,
+        # but some vCenter builds (observed on 8.0.3 against the 9.0 spec,
+        # #2078) return a detail payload that omits/nulls ``capacity`` while
+        # still populating ``free_space``. The ``GET:/vcenter/datastore`` list
+        # row already carries both fields, so fall back to it when the detail
+        # value is absent -- otherwise the composite silently discards a
+        # capacity it already fetched, leaving %-full uncomputable.
+        capacity = detail_capacity if detail_capacity is not None else entry.get("capacity")
+        free_space = detail_free_space if detail_free_space is not None else entry.get("free_space")
         row: dict[str, Any] = {
             "id": ds_id,
             "name": entry.get("name"),
--- a/backend/src/meho_backplane/connectors/vmware_rest/composites/schemas.py
+++ b/backend/src/meho_backplane/connectors/vmware_rest/composites/schemas.py
@@ -415,13 +415,17 @@ DATASTORE_USAGE_RESPONSE_SCHEMA: dict[str, Any] = {
                     "capacity": {
                         "type": ["integer", "null"],
                         "description": (
-                            "Total capacity in bytes; ``null`` when the detail payload omits it."
+                            "Total capacity in bytes; sourced from the per-datastore "
+                            "detail payload, falling back to the listing row when the "
+                            "detail omits it. ``null`` only when neither carries it."
                         ),
                     },
                     "free_space": {
                         "type": ["integer", "null"],
                         "description": (
-                            "Free space in bytes; ``null`` when the detail payload omits it."
+                            "Free space in bytes; sourced from the per-datastore detail "
+                            "payload, falling back to the listing row when the detail "
+                            "omits it. ``null`` only when neither carries it."
                         ),
                     },
                     "vm_count": {
--- a/backend/tests/test_connectors_vmware_rest_composites_read.py
+++ b/backend/tests/test_connectors_vmware_rest_composites_read.py
@@ -508,6 +508,97 @@ async def test_datastore_usage_three_ops_per_datastore_aggregates_correctly() ->
 
 
 @pytest.mark.asyncio
+async def test_datastore_usage_capacity_falls_back_to_list_row_when_detail_omits_it() -> None:
+    """Detail without ``capacity`` (live 8.0.3 shape) -> capacity from the list row (#2078).
+
+    Some vCenter builds return a per-datastore detail ``Datastore.Info``
+    that populates ``free_space`` but omits ``capacity`` entirely. The
+    ``GET:/vcenter/datastore`` list row carries both fields, so the
+    composite must fall back to the list-row ``capacity`` rather than
+    silently emitting ``capacity: null`` (which leaves %-full
+    uncomputable, the whole reason to reach for the composite).
+    """
+    listing = [
+        {"datastore": "datastore-1", "name": "ds-1", "type": "VMFS", "capacity": 1000},
+    ]
+    sequence = [
+        _ok_result("GET:/vcenter/datastore", listing),
+        # Detail carries free_space but NO ``capacity`` key at all.
+        _ok_result("GET:/vcenter/datastore/{datastore}", {"free_space": 400, "type": "VMFS"}),
+        _ok_result("GET:/vcenter/vm", [{"name": "vm-a"}]),
+    ]
+    dispatch = _RecordingDispatchChild(sequence)
+    out = await datastore_usage_composite(
+        operator=_make_operator(),
+        target=object(),
+        params={},
+        dispatch_child=dispatch,
+    )
+    row = out["datastores"][0]
+    # capacity sourced from the list row; free_space from the detail read.
+    assert row["capacity"] == 1000
+    assert row["free_space"] == 400
+
+
+@pytest.mark.asyncio
+async def test_datastore_usage_capacity_null_when_neither_detail_nor_list_carry_it() -> None:
+    """Neither detail nor list row carries ``capacity`` -> row ``capacity`` is null (#2078).
+
+    The fallback is best-effort: with no source for capacity the row must
+    still be schema-valid (``capacity`` typed ``["integer", "null"]``) and
+    the aggregation must not crash.
+    """
+    listing = [
+        {"datastore": "datastore-1", "name": "ds-1", "type": "VMFS"},
+    ]
+    sequence = [
+        _ok_result("GET:/vcenter/datastore", listing),
+        # Detail lacks capacity; list row lacks capacity -> null, no crash.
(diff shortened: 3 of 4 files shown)


damir-topic:
@coderabbitai review

damir-topic (commented):




**Anchor task:** #2078 — `vmware.composite.datastore.usage` returned `capacity=null` when the per-datastore detail omits `capacity`
**Surface area:** backend (VMware REST composite connector)
**Verdict:** APPROVE
**Head SHA reviewed:** `be233eeb`
**AI reviewer (CodeRabbit):** not received within 15min (no review posted)

### Acceptance criteria
- [x] AC1 — `capacity` falls back to `entry.get("capacity")` when detail omits/nulls it (`_read.py:620`); verified by `test_datastore_usage_capacity_falls_back_to_list_row_when_detail_omits_it` (capacity=1000 from list row, free_space=400 from detail).
- [x] AC2 — `free_space` given the same detail-then-list fallback for symmetry (`_read.py:621`); still resolves to detail value when present.
- [x] AC3 — new test asserts detail-returns-free_space-but-no-capacity → non-null capacity from list row. Present and green.
- [x] AC4 — existing detail-wins aggregation test (`test_datastore_usage_three_ops_per_datastore_aggregates_correctly`) unchanged and passing; additionally `test_datastore_usage_detail_capacity_wins_over_list_row` proves detail's 100 beats list's 1000.
- [x] AC5 — both-missing case yields `capacity: null`, no crash, schema-valid — `test_datastore_usage_capacity_null_when_neither_detail_nor_list_carry_it`.

### Predicted vs. actual diff
- ✅ `_read.py` — row-builder merge sources capacity/free_space detail-then-list (`is not None` guard is correct: keeps a legit detail `0`, falls back only on absence/null)
- ✅ `schemas.py` — description-only update to reflect new source; `["integer","null"]` typing unchanged (no schema-shape change, as issue notes)
- ✅ `test_...composites_read.py` — three new tests covering fallback / both-missing / detail-wins
- ✅ `docs/codebase/connectors-vmware-rest.md` — documents the list-row fallback
- 🕳️ Missing: none
- 🚫 Undisclosed: none

### Integration verification
- [x] `ruff check` — clean (3 changed files)
- [x] `mypy composites/` — Success, no issues (7 source files)
- [x] `pytest test_connectors_vmware_rest_composites_read.py` — 27 passed
- [⚠️ not run] migrations — n/a (no DB/migration change)
- [⚠️ not run] OpenAPI snapshot — n/a (op-result schema, not a route model; confirmed no FastAPI route touched)

### Findings
None. No blockers, majors, minors, or questions.

- CodeRabbit posted no review within the 15-minute window (empty `reviews`/`reviewThreads`). Proceeding per skill Hard rule 17; verdict does not depend on the bot.

### Docs
- [x] `docs/codebase/connectors-vmware-rest.md` — updated to document the detail-then-list source and `#2078` rationale.

### What worked well
- Fallback semantics use `is not None` rather than truthiness, so a legitimate detail `capacity: 0` (empty datastore) is preserved rather than being overwritten by a stale list-row value — a subtle correctness win.
- Test matrix maps 1:1 onto the acceptance criteria (fallback, both-missing, detail-wins) and the pre-existing happy-path aggregation test is retained unchanged, proving no regression to the primary path.

Issue #2089, opened 2026-06-26
ui: BFF session lifecycle has three dead-end paths (expired callback state → raw JSON, no token refresh, partial 401 handling) — recoverable states surface as terminal errors


zdamir:
## What we hit

Three legs of one theme — **the BFF never recovers a recoverable auth state**, even when the recovery is mechanical.

### Leg 1: Callback dead-end on expired state

Login started, operator pivoted to 1Password to look up a credential, came back to the browser ~6 minutes later, completed the Keycloak login. The Keycloak redirect carried a valid `code` but the BFF's `state` had expired (`STATE_TTL=300s` ish). The BFF returned:

```
HTTP 400
{"detail":"authorization_failed"}
```

In the browser, this renders as raw JSON. No link. No "your login took too long, try again" affordance. No auto-restart of the flow. The user's options at this point:

- Read the JSON, guess what to do (manually navigate back to `/`).
- Hit back, hit forward, refresh — any of which may restart the flow OR may leave them on the same dead-end page.

A fresh `/ui/auth/login` round-trip succeeded immediately after.

### Leg 2: No refresh on access token expiry

The BFF holds the upstream Keycloak access_token (TTL ~5 min) **and** the refresh_token (TTL much longer) in the encrypted session cookie. After ~5 min of UI use, the access token expires; the BFF does **not** transparently refresh it. Some routes (writes, like `/ui/memory/create` — see sibling signal `ui-memory-create-401-token-expired-no-refresh`) bubble a raw 401. Some routes (reads, like the `/ui/dashboard` poller) just silently stop updating.

A standard BFF would: on upstream 401, attempt a refresh-token round-trip; if that succeeds, retry the original request once; if it fails (refresh_token also expired), redirect to `/ui/auth/login`. The MEHO BFF does none of these consistently.

### Leg 3: Partial 401 handling

Different routes 401 differently:

- Writes (`POST /ui/memory/create`, `POST /ui/library/...`): bubble raw 401 with `{"detail":"token_expired"}`.
- Some reads: silently stop polling.
- Some other reads: return a 401 + Keycloak `WWW-Authenticate: Bearer` header that the browser ignores (no native browser flow for that on htmx-driven fragments).

A consistent posture would be: **every 401 from upstream → try refresh → retry once → redirect on hard-fail**. Today the behavior is per-route and unpredictable.

## Proposed fix

One unified middleware in the BFF that wraps every upstream call:

```python
async def upstream_call(req, target_route):
[...]

PR #2056, opened 2026-06-23
fix(ui): route token-consuming /ui/* lifts through the refresh-on-401 seam

zdamir:
## Summary

- The BFF's RFC-9700 refresh module (`ui/auth/refresh.py`, #1694) refreshes once on a `token_expired` 401 and re-verifies, but its use was inconsistent: nine per-route operator lifts re-verified the stored access token with a **bare** `load_session` + `verify_jwt_for_audience`, so an expired-but-refreshable token 401-ed mid-session even though the session cookie + refresh token were still valid. This is the failure class the consumer signal `ui-memory-create-401-token-expired-no-refresh` observed.
- Every token-consuming `/ui/*` lift now routes through `load_fresh_session` + `verify_access_token_with_refresh`, mirroring the existing `approvals` / `corpus` / `retrieval` precedent — a single refresh chokepoint, fail-soft to the documented `session_expired` terminal state, never a bare 401 mid-session.
- The originally-reported create-modal endpoint is already resolved at HEAD (token-free render); documented below so the signal can be re-scored.

## Route inventory (AC #1)

`grep -rnE "_load_access_token_or_401|access_token|verify_access_token_with_refresh" backend/src/meho_backplane/ui/routes/` — every site that decrypts the session's upstream access token to call the chassis JWT chain. All `access_token` usages were `Bearer {token}` presented to `verify_jwt_for_audience` (no site passes it to an upstream httpx call), so the seam is the exact drop-in.

| Route module | Lift symbol | Before | After |
|---|---|---|---|
| `approvals/routes.py` | `_resolve_operator` | seam ✅ | unchanged (precedent) |
| `corpus/routes.py` | `_resolve_operator` | seam ✅ | unchanged (precedent) |
| `retrieval/routes.py` | `_resolve_operator` | seam ✅ | unchanged (precedent) |
| `agents/operator.py` | `_lift_operator` | bare verify | **seam** |
| `connectors/operator.py` | `lift_operator_from_session` | bare verify | **seam** (also covers topology `batch` / `edges` / `temporal`, which reuse it) |
| `memory/operator.py` | `resolve_ui_operator` | bare verify | **seam** |
| `kb/routes.py` | `_require_tenant_admin` | bare verify | **seam** |
| `runbooks/routes.py` | `_resolve_role` (soft-fail) | bare verify | **seam** (inside `try/except`) |
| `operations/routes.py` | `_resolve_operator` | bare verify | **seam** |
[...]

Issue #2150, opened 2026-07-06
Initiative: Operator-facing surfaces — /ui BFF session resilience + meho CLI DX (v0.19.0)


damir-topic:
Parent goal: #221

## Why

v0.19.0 findings on the **operator-facing surfaces** — the `/ui` BFF (session / CSRF / error-surfacing) and the `meho` CLI (login, secret staging, docs search) — where recoverable states surface as terminal errors or silent failures. Code lands on `evoila/meho`.

## Scope (child tasks)

- [ ] #2088 — `/ui/memory/create` 401 silently swallowed; route the write path through the existing refresh-on-401 seam (#1694) via `resolve_ui_operator`.
- [ ] #2089 — BFF session lifecycle: the expired OAuth-callback-state raw-JSON dead-end (the token-refresh + 401→login legs already shipped via #1694).
- [ ] #2112 — runbooks-UI abort returns HTTP 403 to the run-starter (a CSRF rejection, not RBAC); document the CSRF requirement + wire-or-remove the misleading admin-bypass banner clause.
- [ ] #2091 — `meho targets import` fail-fast when `secret_ref` is outside `secret/meho/*` + cause-naming dispatch error.
- [ ] #2107 — `meho login` DNS-resolver fallback / `--resolve` for split-DNS clients.
- [ ] #2109 — `meho docs search` (CLI) vs `POST /api/v1/search_docs` (REST) apply the same capability gate.

## Out of scope

- The runbooks-UI *authoring* defects (evoila/meho #2117) and the runbook run-state read surface (execution-ledger initiative).

## Definition of done

- [ ] Each recoverable state surfaces gracefully (refresh / redirect / structured error), and each CLI gap has a named remedy; every fix carries a test or manual-repro note.

damir-topic:
Closing — all 6 operator-facing Tasks completed: #2088/#2089/#2091/#2107/#2109/#2112.

PR #2177, opened 2026-07-06
test(memory-ui): pin /ui/memory write-path refresh-on-401 seam (#2088)

### LINKED ISSUE #2088: ui: /ui/memory/create 401 from upstream is silently swallowed — no error banner / no token refresh / no redirect to login
## What we hit

Different shape than the v0.15.0 CSRF-on-poll regression that #1754 fixed — the CSRF double-submit is now working (the `x-csrf-token` header is sent + accepted). But **a new failure mode lands in the same operator-facing surface**: click **Create memory** → **nothing happens**.

## Reproduction (verified live, v0.16.0)

1. Log into `/ui/` via Keycloak SSO (auth_code + PKCE against client `meho-web`). A fresh BFF session cookie + upstream Keycloak access_token are minted.
2. Wait ≥15 minutes (the typical Keycloak access-token TTL on this realm).
3. Navigate to `/ui/memory`, fill in the "Add a memory" form, click **Create memory**.

Observed:

- The button's spinner appears for a fraction of a second.
- The form does NOT clear.
- No error banner.
- No redirect to login.
- The memory does NOT land in the grid.
- Browser DevTools: the `POST /ui/memory/create` request returns `HTTP 401` with body `{"detail": "token_expired"}` (raw, from the upstream API — the BFF didn't refresh, didn't reject, didn't redirect).

Repeating the click silently 401s again. The operator's only recovery: full page reload, which goes through `/ui/auth/login` → fresh tokens → next write succeeds.

## What's expected

For the same upstream 401 case on a token-expired session, the BFF should do **one** of these (in priority order):

1. **Auto-refresh the access token** via the upstream Keycloak refresh-token endpoint (the BFF holds the refresh_token in the encrypted session cookie). On a successful refresh, retry the failed write once. This is the standard BFF pattern and matches the existing oauth2-proxy implementation for the corpus.
2. **Redirect to login on the BFF response.** Return a `303 See Other` to `/ui/auth/login?return_to=<original-path>` so the browser cleanly re-auths and the user comes back to the same page.
3. **Surface a structured error** in the UI's flash-banner slot: "Your session expired — please log in again." with a single click-to-relogin link.

Today the BFF picks **none** of these — the 401 bubbles up as a JSON body that the UI form-handler swallows without rendering.

## Where the gap likely is

The MEHO UI uses HTMX for form submits. The HTMX response handler doesn't have a default "show error for 4xx responses" — it expects the server to return a meaningful HTML fragment OR set a status that maps to a fail-class (`HX-Trigger` etc.). A raw 401 with a JSON body lands in HTMX's "no response swap" path and silently does nothing.

Two fixes (do both):

1. **BFF**: on upstream 401 to a write route, attempt refresh-then-retry. If refresh fails (refresh_token expired too), return `303 → /ui/auth/login?...`. Don't pass raw `{"detail":"token_expired"}` to the browser.
2. **UI**: register a global HTMX `htmx:responseError` handler that renders any 4xx/5xx into the flash slot, so even unhandled BFF errors don't disappear silently.

## Why this is worth filing

The silent-failure UX is the worst kind: the operator believes their change went through (the form clears? no — it doesn't, but the operator's eye is on the success state), and a memory they thought they saved is lost. Every operator on this UI will hit this exactly once after their first ≥15min idle.

## Verification

```bash
# 1. Trigger: wait >15min, click Create memory.
# Expected after fix: either an auto-refreshed token + successful create,
# OR a redirect to /ui/auth/login, OR a visible flash banner.
# Not acceptable: silent no-op.
```

## Related

- `ui-bff-session-expiry-dead-ends` (this batch) — the session-lifecycle siblings (callback dead-end, no-refresh-on-401-from-some-routes). This signal is one specific path of the broader "BFF never recovers a recoverable auth state" pattern.
- `ui-memory-grid-clobbered-by-inherited-htmx-target` (this batch) — different UI bug on the same page, compounding the "memory feature feels broken" impression.

Symptom confirmed live-plausible and correctly described, but the two-part fix the report prescribes is the wrong diagnosis. The refresh-then-retry-else-redirect infrastructure the report asks to *build* **already ships** — it just isn't wired onto this write path.

What already exists (do NOT rebuild):

- `backend/src/meho_backplane/ui/auth/refresh.py` (G0.25 #1694) provides `verify_access_token_with_refresh()` (`refresh.py:435`) — the reactive leg: presents the stored access token to the JWT chain, and on exactly the `token_expired` 401 refreshes once via `refresh_session_tokens()` and re-verifies. On terminal refresh failure it raises `HTTPException(401, detail="session_expired")` (`SESSION_EXPIRED_DETAIL`, `refresh.py:137`), a code distinct from `token_expired`.
- `backend/src/meho_backplane/ui/auth/errors.py` already maps that `session_expired` 401 on a `/ui/*` path to a `302 → /ui/auth/login?return_to=…` for HTML `Accept` (`errors.py:99-133`), and returns the structured `session_expired` JSON (not raw `token_expired`) for non-HTML/fetch callers, clearing the dead session cookie in both cases. Registered app-wide in `meho_backplane.main`.
- The **corpus**, **retrieval**, and **approvals** UI routers already consume this seam (`ui/routes/corpus/routes.py:164-171`, `ui/routes/retrieval/routes.py:232-239`, `ui/routes/approvals/routes.py:211-218`), and the chassis `require_ui_admin` path uses it too (`ui/auth/middleware.py:558-579`). This is the report's own "matches the existing oauth2-proxy implementation for the corpus" — the pattern is present, the memory router just never adopted it.

Actual root cause — the memory write path bypasses the refresh seam:

- `POST /ui/memory/create` (`_create_submit_handler`, `ui/routes/memory/routes.py:181`, registered at `routes.py:363-366`) depends on `_resolve_operator_dep = Depends(resolve_ui_operator)` (`routes.py:98`, `:190`).
- `resolve_ui_operator` (`ui/routes/memory/operator.py:139`) calls `verify_jwt_for_audience(...)` **directly** at `operator.py:174` instead of `verify_access_token_with_refresh(...)`. So an expired access token raises the JWT chain's raw `token_expired` 401 — a code `errors.py` deliberately does NOT intercept (blanket-redirecting all 401s would swallow `signature_verification_failed` etc.) — so it leaks to the browser exactly as observed. No refresh is ever attempted; no redirect/flash is ever produced.
- Same `_resolve_operator_dep` gates every other memory write handler — `_patch_handler` (`routes.py:142`), `_delete_handler` (`:155`), `_promote_submit_handler` (`:270`), `_bulk_handler` (`:304`) — so PATCH/DELETE/promote/bulk carry the **identical** defect, not just create. The fix belongs in `resolve_ui_operator`, which repairs all of them at once.

Scope corrections vs the reporter's hypothesis:

- The report's "BFF fix #1" (build refresh-then-retry) reduces to: route `resolve_ui_operator` through the existing `load_fresh_session` + `verify_access_token_with_refresh` seam, matching the sibling routers. No new refresh primitive.
- The report's "BFF fix #2" (return `303`) is already the shipped contract as a `302` via `errors.py` once the handler emits `session_expired` instead of `token_expired`. No new redirect code.
- The report's "UI fix #2" (global `htmx:responseError` flash handler) is **out of scope for the real residual**: once the write path emits `session_expired`, a full-page HTMX navigation gets the 302 login redirect and a fetch gets the structured code — neither is a silent no-op. A generic client-side error surfacer is a separate hardening idea and belongs on the sibling tracker #2089, not this task; folding it in here would re-introduce speculative substrate.

Sibling: this is the memory-router instance of the broader "BFF never recovers a recoverable auth state" pattern tracked by #2089 (three dead-end paths). Cross-referenced.

Parent goal: #221

## Acceptance criteria

- [ ] `resolve_ui_operator` (`backend/src/meho_backplane/ui/routes/memory/operator.py`) resolves the operator through the existing refresh seam (`load_fresh_session` + `verify_access_token_with_refresh`) instead of the direct `verify_jwt_for_audience` call at `operator.py:174`; no new refresh/redirect primitive is introduced.
- [ ] A test proves `POST /ui/memory/create` on a session whose access token is expired but whose refresh token is still valid refreshes-and-retries once and creates the memory (2xx / grid fragment), with the session row's tokens rotated.
- [ ] A test proves that when refresh fails, `POST /ui/memory/create` yields the `session_expired` contract — a `302 → /ui/auth/login?return_to=/ui/memory…` for an HTML `Accept`, and the structured `session_expired` body (never raw `token_expired`) for a fetch/HTMX `Accept` — and that a raw `token_expired` 401 is asserted to never reach the browser on this path.
- [ ] The same fix is verified to cover the sibling memory write handlers sharing `_resolve_operator_dep` (`_patch_handler`, `_delete_handler`, `_promote_submit_handler`, `_bulk_handler`): at least one of them has a regression test for the expired-token → refresh/redirect behavior.
- [ ] No behavior change on the non-expiry 401 codes (`signature_verification_failed`, `invalid_audience`, `ui_session_identity_mismatch`): a test asserts these still surface their original detail and are not swallowed by a login bounce.

Parent initiative: #2150

damir-topic:
## Summary

- **Root-cause note (important):** the code fix #2088 prescribes as AC #1 — routing the memory router's `resolve_ui_operator` through the shared `load_fresh_session` + `verify_access_token_with_refresh` refresh-on-401 seam (G0.25 #1694) — **already shipped in #2056** (2026-06-23), which swept the seam across every token-consuming `/ui/*` lift including `memory.operator`. The ticket's triage section (dated 2026-07-05) still cites a stale `operator.py:174` direct `verify_jwt_for_audience` call that no longer exists on `main`. So this PR is **tests + a docstring refresh**, not a re-implementation.
- What #2056 left behind is the **memory-router-specific behavioural coverage** the ticket's acceptance criteria pin: #2056 tested only a single agents-console probe as "representative coverage for the whole seam adoption." This PR supplies the missing coverage against the real `POST /ui/memory/create` write path (the exact surface the reporter hit) and one sibling write handler.
- The test app's `_build_app` now registers the #1694 `ui_session_expired_exception_handler` exactly as production `main.py` does, so the HTML login-redirect contract is exercised end to end. It is inert for every other test (only `session_expired` on a `/ui/*` path matches; all else delegates to FastAPI's stock handler).
- Also refreshes the now-stale `resolve_ui_operator` module docstring, which still described the pre-#2056 direct `verify_jwt_for_audience` call.

## Acceptance criteria

- [x] **AC #1** — `resolve_ui_operator` resolves through the refresh seam, not a bare `verify_jwt_for_audience` — already true on `main` via #2056; verified (`operator.py` uses `load_fresh_session` + `verify_access_token_with_refresh`, no new primitive introduced). Docstring corrected to match.
- [x] **AC #2** — `test_create_submit_expired_token_refreshes_and_persists`: expired access token + live refresh → refreshes-and-retries once, creates the memory (204 + `HX-Redirect`), row tokens rotated, exactly one token-endpoint POST.
- [x] **AC #3** — `test_create_submit_refresh_unavailable_redirects_html_to_login` (HTML → 302 `/ui/auth/login?return_to=/ui/memory/create`, cookie cleared) + `test_create_submit_refresh_unavailable_keeps_json_shape_for_fetch` (fetch → 401 `session_expired`, asserts the raw `token_expired` never reaches the browser).
- [x] **AC #4** — `test_delete_handler_expired_token_refreshes_and_serves`: the sibling `DELETE` handler sharing `_resolve_operator_dep` inherits the refresh-and-serve fix.
- [x] **AC #5** — `test_create_submit_wrong_audience_401_is_not_swallowed_by_login_bounce`: a non-expiry 401 (`invalid_audience`) keeps its raw detail, no login bounce, no refresh attempted.
- [x] **Out of scope, per triage:** the reporter's "UI fix #2" (global `htmx:responseError` flash handler) — a separate hardening idea tracked on sibling #2089, not folded in here.

## Test plan

- [x] `mypy src/.../memory/operator.py --ignore-missing-imports` — clean
- [x] `pytest tests/test_ui_memory_create_promote.py` — 38 passed (33 pre-existing + 5 new)
- [x] `pytest tests/test_ui_token_refresh.py tests/test_ui_memory_list.py` — 49 passed (no ripple)
- [x] `.claude/hooks/code-quality.py --diff` — exit 0, no warnings
- [x] `cd cli && make snapshot-openapi` — no OpenAPI drift (only a `Depends`-target docstring changed; no route/schema touched)

## Out of scope

- The global client-side `htmx:responseError` flash surfacer (reporter's UI fix #2) — belongs on #2089.

Closes #2088

### DIFF
--- a/backend/src/meho_backplane/ui/routes/memory/operator.py
+++ b/backend/src/meho_backplane/ui/routes/memory/operator.py
@@ -14,12 +14,17 @@ Initiative #341 (G10.4 Memory UI), Task #877 (T1). The chassis
   server-side; never trusts a client-supplied role hint),
 * the create / promote handlers T2 #878 will add on this same router.
 
-The dependency loads the encrypted session row (via the same
-:func:`~meho_backplane.ui.auth.session_store.load_session` the chassis
-middleware uses), decrypts the access token, and re-runs it through
-:func:`~meho_backplane.auth.jwt.verify_jwt_for_audience` to produce a
-fully-validated :class:`Operator`. The JWT chain caches the JWKS in
-process so the round-trip is local-only after the first hit.
+The dependency loads the encrypted session row through the shared
+refresh seam (:func:`~meho_backplane.ui.auth.refresh.load_fresh_session`),
+decrypts the access token, and re-runs it through
+:func:`~meho_backplane.ui.auth.refresh.verify_access_token_with_refresh`
+(G0.25 #1694, adopted memory-wide by #2056) to produce a
+fully-validated :class:`Operator`. That seam silently refreshes once on
+the ``token_expired`` 401 and re-verifies before failing closed to
+``session_expired`` -- so an expired-but-refreshable access token on a
+memory write no longer leaks the raw ``token_expired`` body to the
+browser (#2088). The JWT chain caches the JWKS in process so the
+round-trip is local-only after the first hit.
 
 Why not store the role on the session row
 -----------------------------------------
--- a/backend/tests/test_ui_memory_create_promote.py
+++ b/backend/tests/test_ui_memory_create_promote.py
@@ -46,7 +46,9 @@ from datetime import UTC, datetime, timedelta
 from typing import Any
 from unittest.mock import AsyncMock
 from unittest.mock import patch as _patch
+from urllib.parse import parse_qs, urlparse
 
+import httpx
 import pytest
 import respx
 from cryptography.fernet import Fernet
@@ -54,6 +56,7 @@ from fastapi import FastAPI
 from fastapi.staticfiles import StaticFiles
 from fastapi.testclient import TestClient
 from sqlalchemy import select
+from starlette.exceptions import HTTPException as StarletteHTTPException
 
 from meho_backplane.audit import AuditMiddleware
 from meho_backplane.auth.jwt import clear_jwks_cache
@@ -67,7 +70,11 @@ from meho_backplane.memory._internal import (
 )
 from meho_backplane.memory.schemas import MemoryScope, kind_for_scope
 from meho_backplane.settings import get_settings
-from meho_backplane.ui.auth import SESSION_COOKIE_NAME, UISessionMiddleware
+from meho_backplane.ui.auth import (
+    SESSION_COOKIE_NAME,
+    UISessionMiddleware,
+    ui_session_expired_exception_handler,
+)
 from meho_backplane.ui.auth import build_router as build_ui_auth_router
 from meho_backplane.ui.auth.flow import (
     clear_discovery_cache,
@@ -75,6 +82,7 @@ from meho_backplane.ui.auth.flow import (
 )
 from meho_backplane.ui.auth.session_store import (
     create_session,
+    load_session,
     reset_fernet_cache_for_testing,
 )
 from meho_backplane.ui.csrf import CSRF_COOKIE_NAME, CSRFMiddleware, mint_csrf_token
@@ -99,6 +107,7 @@ from tests._oidc_jwt_helpers import (
 from tests._oidc_jwt_helpers import (
     public_jwks as _public_jwks,
 )
+from tests.conftest import mint_token as _mint_token_with_ttl
 
 # ---------------------------------------------------------------------------
 # Fixtures
@@ -166,6 +175,16 @@ def _build_app() -> FastAPI:
     app.add_middleware(AuditMiddleware)
     app.add_middleware(CSRFMiddleware)
     app.add_middleware(UISessionMiddleware)
+    # The #1694 app-level handler maps a `401 session_expired` on a
+    # `/ui/*` path to a `302 -> /ui/auth/login` for HTML requests (and
+    # keeps the structured JSON body for fetch callers), exactly as
+    # `meho_backplane.main` registers it. Wiring it here lets the
+    # refresh-failure tests below assert the real login-redirect
+    # contract the memory write path inherits through the seam. It is
+    # inert for every other test: only `session_expired` on `/ui/*`
+    # matches; all other HTTPExceptions delegate to FastAPI's stock
+    # handler byte-for-byte.
+    app.add_exception_handler(StarletteHTTPException, ui_session_expired_exception_handler)
     app.mount(
         "/ui/static",
         StaticFiles(directory=str(static_root_dir()), check_dir=False),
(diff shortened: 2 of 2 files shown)


damir-topic:
@coderabbitai review

damir-topic (commented):




**Anchor task:** #2088 — `/ui/memory/create` 401 silently swallowed; residual = memory-router refresh-on-401 regression coverage
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `f2b78bf1`
**AI reviewer (CodeRabbit):** responded but rate-limited ("Review limit reached", next in 51 min) — produced no findings this iteration

### Acceptance criteria
- [x] **AC #1** — `resolve_ui_operator` resolves through the refresh seam (`load_fresh_session` + `verify_access_token_with_refresh`), not a bare `verify_jwt_for_audience`. Verified this is already true on `main` via #2056 (commit `6184c8ff`); this PR's `operator.py` change is docstring-only (`git diff origin/main...HEAD` for that file touches only doc/comment lines). No code path left unfixed. Docstring corrected to match reality.
- [x] **AC #2** — `test_create_submit_expired_token_refreshes_and_persists`: expired access token (exp −120s, past the 30s leeway) + live refresh → 204 + `HX-Redirect: /ui/memory`, memory persisted, row tokens rotated (`refresh-token-rotated`), exactly one token-endpoint POST. Passes.
- [x] **AC #3 (HTML)** — `test_create_submit_refresh_unavailable_redirects_html_to_login`: refresh `invalid_grant` → 302 `/ui/auth/login?return_to=/ui/memory/create`, dead cookie cleared (`Max-Age=0`), write did not land. Passes.
- [x] **AC #3 (fetch)** — `test_create_submit_refresh_unavailable_keeps_json_shape_for_fetch`: fetch `Accept` → 401 `{"detail":"session_expired"}`, explicitly asserts the raw `token_expired` never reaches the browser. Passes.
- [x] **AC #4** — `test_delete_handler_expired_token_refreshes_and_serves`: sibling `DELETE /ui/memory/user/<slug>` shares `_resolve_operator_dep = Depends(resolve_ui_operator)` (routes.py:160), refreshes once → 204. Genuine regression guard through the same seam. Passes.
- [x] **AC #5** — `test_create_submit_wrong_audience_401_is_not_swallowed_by_login_bounce`: validly-signed token with wrong `aud` → 401 `invalid_audience` verbatim (even for HTML `Accept`), no login bounce, `token_route.call_count == 0`. Discriminator is sound: `verify_access_token_with_refresh` only enters the refresh leg on `token_expired` (`refresh.py:467-472`); every other 401 propagates untouched. Passes.
- [x] **Out of scope, per triage** — reporter's UI fix #2 (global `htmx:responseError` flash handler) correctly deferred to sibling #2089; not folded in.

### Project-board alignment
|---|---|---|---|---|---|
| #2088 | Open (task, parent init #2150, parent goal #221) | Closes | ✅ | ✅ full | close on merge |

### Predicted vs. actual diff
- ✅ `backend/tests/test_ui_memory_create_promote.py` — the 5 AC-pinned regression tests + helpers (`_mint_expired_access_token`, `_mock_token_endpoint`, `_load_session_tokens`, `_refresh_grant_response`); test-app registers `ui_session_expired_exception_handler` mirroring `main.py:622` byte-for-byte
- ✅ `backend/src/meho_backplane/ui/routes/memory/operator.py` — docstring-only refresh (the pre-#2056 `verify_jwt_for_audience` prose was stale); no behavioural change
- 🕳️ none missing — every AC has a passing test
- 🚫 none undisclosed

### Integration verification
- [x] `ruff format --check` (changed files) — 2 files already formatted
- [x] `mypy src/.../memory/operator.py --ignore-missing-imports` — Success, no issues
- [x] `pytest tests/test_ui_memory_create_promote.py` — 38 passed (33 pre-existing + 5 new), ran in the review worktree
- [x] 5 new tests run in isolation — 5 passed
- [x] OpenAPI snapshot — no route/schema touched (docstring on a `Depends` target only); CLI API snapshot freshness check green on the PR
- ⚠️ not run: docker compose / migrations replay — n/a (no migration, no compose surface, no route change)

None. No blockers, no majors, no minors requiring a fix-up.

- None. The test seam (overriding the discovery route to carry `token_endpoint`, relying on respx most-recently-registered-wins) is well-documented inline in `_mock_token_endpoint` and is the correct approach for exercising the real refresh leg rather than mocking it out.

- Adopted: 0
- Disputed: 0
- Out of scope: 0
- Note: CodeRabbit hit its per-developer review limit and produced no substantive review this iteration (the green "CodeRabbit" status is the trigger ack, not a findings-bearing review). Independent review stands on its own.

### Docs
- [x] n/a — `docs/codebase/` doc for this seam is unchanged; the PR corrects the in-code module docstring that #2056 left stale. `docs_status: verified_unchanged`.

- Correct root-cause discipline: the PR verifies AC #1 was already delivered by #2056 rather than re-implementing it, and confines the production change to a stale-docstring fix.
- The tests exercise the *real* refresh leg (stubbing Keycloak's token endpoint + a discovery doc that advertises it) instead of monkeypatching `verify_access_token_with_refresh`, so they'd catch a future resolver reverting to a bare `verify_jwt_for_audience`.
- AC #5's wrong-audience probe stubs the token endpoint to a hard failure and asserts `call_count == 0`, proving no spurious refresh — a genuinely discriminating negative test.

PR #2181, opened 2026-07-06
feat(cli): add `meho login --resolve host:port:ip` split-DNS fallback (#2107)

### LINKED ISSUE #2107: meho login: device-code flow has no DNS-resolver fallback / no --resolve flag — fails on split-DNS clients where the backplane resolves but the Keycloak host doesn't
## Summary

`meho login` (and the `device_authorization` flow it drives) calls Keycloak by hostname directly via the HTTP stack's default resolver. When an operator workstation has a working route to the backplane host but its system resolver returns NXDOMAIN for the Keycloak host — a real and common split-DNS / VPN-DNS-push edge case — the CLI dies with a bare `Could not resolve host: <keycloak-host>` and the operator has no path forward without dropping back to raw curl-with-`--resolve` or hand-editing `/etc/hosts`.

## Observed (consumer-side, against backplane v0.19.0)

In a deploy where:

- Backplane is reachable: `curl https://<backplane-host>/version` returns `200`.
- Keycloak is reachable *at IP* (same lab network): `curl --resolve <keycloak-host>:443:<keycloak-ip> https://<keycloak-host>/realms/<realm>/.well-known/openid-configuration` returns `200`.
- But the workstation's system resolver returns `NXDOMAIN` for `<keycloak-host>` (split-DNS quirk; the VPN-pushed DNS forwarder is configured for the backplane name but not the Keycloak name).

`meho login <backplane>` dies with:

```
Error: backplane auth-config discovery failed (call https://<backplane-host>/api/v1/auth-config: ...)
   ... or further along ...
   Could not resolve host: <keycloak-host>
```

A wrapper script driving the device-code flow with `curl` directly fails identically; only `curl --resolve <keycloak-host>:443:<keycloak-ip> ...` succeeds.

## Why this matters

Split-DNS is the *normal* operational topology for any meho deploy reachable from outside the lab via VPN: the VPN pushes a DNS forwarder that knows the backplane name, but the Keycloak name lives in a different zone or got missed in the forwarder config. Operators hitting this today have to:

1. Diagnose the asymmetry (the error message blames "host not found" without naming which host is fine and which isn't).
2. Find the Keycloak IP manually (lab DNS direct lookup, not the system resolver).
3. Hand-edit `/etc/hosts` or write a one-off `curl --resolve` device-code flow (defeats the point of the official CLI).

`meho login` ought to either degrade gracefully or expose a knob.

### Option A — pass `--resolve <host>:<port>:<ip>` through to the underlying HTTP client

The smallest possible change. An operator who knows the Keycloak IP can:

```bash
meho login https://<backplane> --resolve <keycloak-host>:443:<keycloak-ip>
```

Mirrors `curl --resolve`. No DNS logic in the CLI; the operator is responsible. Solves the immediate problem; doesn't help operators who don't know the workaround exists.

### Option B — alternate-DNS-server flag

```bash
meho login https://<backplane> --dns-server <ip>
```

Sets a `--dns-server` (or `KEYCLOAK_DNS_SERVER` env) that Go's `net.Resolver` uses for Keycloak hostname resolution specifically. More general than `--resolve` (handles the case where the operator doesn't know the IP but knows the lab resolver). Requires plumbing through the HTTP client used for device-code / token endpoints.

### Option C — clearer error + actionable hint

If neither A nor B is appetizing right now, at minimum: when the device-code endpoint resolution fails, emit a structured error naming *which* host failed (`<keycloak-host>` vs `<backplane-host>`) and suggesting the operator pass `--resolve` (or whatever knob lands). Today the error doesn't even tell the operator which host's DNS broke.

## Out of scope

- DNS infrastructure on the deployment side. The right *operational* fix is "make Keycloak resolve via the same forwarder that resolves the backplane" — but in practice this happens, and the CLI being brittle to a common topology quirk is its own DX bug.
- The `meho` CLI's TLS-trust gap (separate issue — if `--insecure-skip-tls-verify` ever lands, it's an orthogonal fix).

## Reproduction is trivial

```bash
# Manufacture the split-DNS state: pin the backplane in /etc/hosts only
echo "<backplane-ip> <backplane-host>" | sudo tee -a /etc/hosts

# Try `meho login` — it'll resolve the backplane fine, then fail on Keycloak
meho login https://<backplane-host>
# → "Could not resolve host: <keycloak-host>"
```

## Verification

Whichever fix shape lands, the verification is:

```bash
# Setup: workstation where system resolver returns NXDOMAIN for <keycloak-host>
# but knows <backplane-host>, and lab DNS @<lab-resolver-ip> can resolve both.

meho login https://<backplane> [--resolve <keycloak-host>:443:<keycloak-ip>]
# expected: device-code flow completes; JWT stored; subsequent `meho status` works
```

## Consumer-side trace

Tracking ticket on the consumer repo: [filed alongside this issue]. The workaround in the consumer's wrapper is an inline `curl --resolve` device-code re-implementation — happy to share if a contributor wants a copy of the working flow.

Parent goal: #221

## Acceptance criteria

- [ ] `meho login` exposes a DNS escape hatch — a `--resolve <host>:<port>:<ip>` flag (Option A) and/or a `--dns-server <ip>` flag (Option B) — verifiable by `grep -n 'resolve\|dns-server' cli/internal/cmd/login.go` finding the flag registration and `meho login --help` listing it.
- [ ] The chosen knob is plumbed into the HTTP client used for BOTH auth-config discovery and the device-code/token endpoints (today all three use `http.DefaultClient` at `cli/internal/cmd/login.go`), not just one call site — verifiable by a unit test that asserts the custom resolver/host-pin is honoured on the device-auth POST, not only on discovery.
- [ ] When the Keycloak host fails to resolve, the emitted error names *which* host failed (`<keycloak-host>` vs `<backplane-host>`) and points the operator at the new knob (Option C) — verifiable by a test asserting the error string contains the Keycloak hostname and the flag name, distinct from the backplane-side discovery failure message.
- [ ] A manual repro on a split-DNS setup (backplane pinned in `/etc/hosts`, Keycloak host NXDOMAIN in the system resolver) succeeds: `meho login https://<backplane> --resolve <keycloak-host>:443:<keycloak-ip>` completes the device-code flow, stores the JWT, and a subsequent `meho status` works.
- [ ] The DNS escape hatch is documented (CLI `--help` long text and/or `docs/`) — verifiable by a grep for the flag name in the login command help text and the docs tree.

> Triaged from raw finding via /review-ticket. CLI DX gap on split-DNS clients.

Parent initiative: #2150

damir-topic:
## Summary
- `meho login` drove auth-config discovery, OIDC discovery, and the device-code/token endpoints through `http.DefaultClient`, so split-DNS operator workstations (backplane host resolves, Keycloak host NXDOMAIN — the normal VPN-DNS-push topology) hit a bare "could not resolve host" with no CLI-native workaround.
- Adds `meho login --resolve <host>:<port>:<ip>` (Option A from the issue), mirroring `curl --resolve`: pins a host to a known IP at connect time only, so TLS SNI/Host still use the real hostname and certificate validation is unchanged.
- The pinned-dial `http.Client` is threaded through **all three** login call sites (auth-config discovery, OIDC discovery, device-code/token), so the knob reaches the device-auth POST — not just discovery.
- When the Keycloak host fails DNS resolution, the error now **names the Keycloak host** and points at `--resolve` (Option C), distinct from the backplane-side auth-config discovery failure (worded around `--issuer`/`--client-id`).

## Implementation
- `cli/internal/auth/resolver.go` — `ParseResolveEntries` (fail-loud validation of `host:port:ip` triples, IPv6-aware), `HTTPClientWithOverrides` (cloned `DefaultTransport.DialContext` pin; no override → `http.DefaultClient` unchanged), `IsHostResolutionError` (`*net.DNSError` classifier).
- `cli/internal/cmd/login.go` — `--resolve` flag + long-help docs; builds the shared client; `hintKeycloakResolution` rewraps DNS failures on the Keycloak-facing calls. Non-resolution errors (expired code, access_denied, timeouts) pass through unchanged.
- `docs/codebase/cli.md` — login-flow section documents the escape hatch and the TLS-safe rationale.

## Test plan
- [x] `go vet ./internal/auth/... ./internal/cmd/` — clean
- [x] `gofmt -l` on all changed files — clean
- [x] `go test ./...` — all pass
- [x] AC #1 flag registration + `meho login --help` lists `--resolve` — `TestLoginCommandHelpListsFlags`, `TestLoginCommandHelpDocumentsResolve`
- [x] AC #2 pin honoured on the device-auth POST (not just discovery) — `TestOverrideHonouredOnDeviceAuthPOST` (phantom Keycloak host, only the `--resolve` pin makes the POSTs land), plus `TestHTTPClientWithOverridesPinsDial`
- [x] AC #3 error names the Keycloak host + `--resolve`, distinct from the backplane failure — `TestHintKeycloakResolutionNamesHostAndFlag`, `TestHintKeycloakResolutionPassesThroughNonDNSErrors`
- [x] AC #4 the manual `--resolve <kc>:443:<ip>` repro path is the code path implemented; automated analog is the pinned-dial test
- [x] AC #5 documented in `--help` long text and `docs/codebase/cli.md`

## Out of scope
- Option B (`--dns-server` alternate-resolver flag) — the issue frames A/B as "and/or"; A fully resolves the reported failure with the smallest blast radius and no resolver plumbing.
- TLS-trust gap (`--insecure-skip-tls-verify` for login) — separate concern per the issue's Out-of-scope.

Closes #2107

### DIFF
--- a/cli/internal/auth/resolver.go
+++ b/cli/internal/auth/resolver.go
@@ -1,0 +1,146 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 evoila Group
+
+package auth
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"net/http"
+	"strings"
+)
+
+// HostOverrides maps a "host:port" key to a literal IP address, mirroring
+// the semantics of `curl --resolve <host>:<port>:<ip>`. It is the escape
+// hatch for split-DNS operator workstations where the system resolver
+// returns NXDOMAIN for the Keycloak host even though that host is
+// reachable at a known IP (VPN-pushed DNS forwarder configured for the
+// backplane name but not the Keycloak name).
+//
+// The map is keyed by the exact "host:port" pair rather than by host
+// alone so an operator can pin a single endpoint (443) without
+// accidentally short-circuiting resolution for a different port on the
+// same host. This matches curl's per-port override model.
+type HostOverrides map[string]string
+
+// ParseResolveEntries parses zero or more `host:port:ip` strings into a
+// HostOverrides map. The format is identical to `curl --resolve`:
+//
+//	kc.example.com:443:10.0.0.5
+//
+// The port is required (and must be numeric) so the override targets a
+// specific endpoint; the IP is validated as a literal IPv4 or IPv6
+// address. IPv6 addresses are accepted in either bare or bracketed form
+// on the IP side (`kc:443:[::1]` and `kc:443:::1` both resolve to `::1`),
+// with the host taken as everything before the last two colon-separated
+// fields so IPv6 literals used as the *host* are not silently mangled.
+//
+// A malformed entry is a hard error rather than a silent skip: an
+// operator who mistypes their override should learn immediately, not
+// discover at connect time that the CLI ignored the flag and fell back
+// to the broken system resolver.
+func ParseResolveEntries(entries []string) (HostOverrides, error) {
+	if len(entries) == 0 {
+		return nil, nil
+	}
+	overrides := make(HostOverrides, len(entries))
+	for _, raw := range entries {
+		host, port, ip, err := splitResolveEntry(raw)
+		if err != nil {
+			return nil, fmt.Errorf("--resolve %q: %w", raw, err)
+		}
+		overrides[net.JoinHostPort(host, port)] = ip
+	}
+	return overrides, nil
+}
+
+// splitResolveEntry parses one `host:port:ip` string. The IP field may
+// itself contain colons (IPv6), so we split off the IP as everything
+// after the last two structural colons: the trailing field is the IP and
+// the field before it is the port. Whatever remains is the host.
+func splitResolveEntry(raw string) (host, port, ip string, err error) {
+	// Find the port:ip boundary by walking from the front: host and port
+	// are the first two colon-delimited fields; the rest is the IP.
+	first := strings.IndexByte(raw, ':')
+	if first < 0 {
+		return "", "", "", errors.New("expected host:port:ip")
+	}
+	rest := raw[first+1:]
+	second := strings.IndexByte(rest, ':')
+	if second < 0 {
+		return "", "", "", errors.New("expected host:port:ip")
+	}
+	host = raw[:first]
+	port = rest[:second]
+	ip = rest[second+1:]
+
+	if host == "" {
+		return "", "", "", errors.New("host is empty")
+	}
+	if _, portErr := net.LookupPort("tcp", port); portErr != nil {
+		return "", "", "", fmt.Errorf("port %q is not a valid TCP port", port)
+	}
+	// Accept a bracketed IPv6 literal on the IP side for symmetry with
+	// how hosts are written elsewhere; strip the brackets before parsing.
+	ip = strings.TrimPrefix(strings.TrimSuffix(ip, "]"), "[")
+	if ip == "" {
+		return "", "", "", errors.New("ip is empty")
+	}
+	if net.ParseIP(ip) == nil {
+		return "", "", "", fmt.Errorf("ip %q is not a valid IP address", ip)
+	}
+	return host, port, ip, nil
+}
+
+// HTTPClientWithOverrides returns an *http.Client whose transport pins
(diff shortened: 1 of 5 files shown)


damir-topic:
@coderabbitai review

Issue #2111, opened 2026-06-30
runbooks: opacity_floor not enforced on runbook_show_template / GET /api/v1/runbooks/templates/<slug> — mid-run operator gets HTTP 200 + full template body (all steps visible) instead of the documented -32602


zdamir:
## Summary

The `runbooks` pillar promises a **step-opacity contract** with two enforcement points: `runbook_next` returns only the current step (no full template visible at run-time), and `runbook_show_template` enforces an `opacity_floor` — *"an operator is denied the full template mid-run, `-32602`; a naive 'I can read the whole procedure while running it' model is wrong by design."*

At v0.19.0, **`runbook_next` honors the contract correctly**, but **`runbook_show_template` does not enforce the `opacity_floor`**. A `tenant_admin` calling `GET /api/v1/runbooks/templates/<slug>` while they have an active run for that template receives **HTTP 200 + the full template body (every step visible)**. The documented `-32602` (or any 4xx) is never returned. Same behaviour at step 1 and step N — consistent across the run's lifecycle, not a corner case.

This is the **load-bearing design property of the runbooks pillar**. An operator wanting to skip ahead in a procedure has a one-call bypass via `show_template`, defeating the structured per-step verification gate that distinguishes a runbook from a raw `kubectl one-liner`.

## Reproduction (live on v0.19.0)

```bash
JWT=$(...)  # tenant_admin device-code or SA-token

# 1. Find any published runbook template the caller can show
SLUG=$(curl -sk -H "Authorization: Bearer $JWT" \
        '<backplane>/api/v1/runbooks/templates' \
       | jq -r '.templates[] | select(.status=="published") | .slug' | head -1)

# 2. Start a run against any registered target (the actual target doesn't
#    matter for this probe; the run just needs to be active)
RUN=$(curl -sk -H "Authorization: Bearer $JWT" -H 'Content-Type: application/json' \
       -X POST '<backplane>/api/v1/runbooks/runs' \
       -d "{\"template_slug\":\"$SLUG\",\"target\":\"<any-registered-target>\"}" \
       | jq -r '.run_id')

# 3. GET show_template while the run is ACTIVE, by the SAME operator who started it
#    Documented contract: -32602 OpacityFloorViolation
#    Live behavior at v0.19.0:
curl -sk -H "Authorization: Bearer $JWT" \
     "<backplane>/api/v1/runbooks/templates/$SLUG" \
     -w '\nHTTP=%{http_code}\n'
# →
# HTTP 200
# {
[...]

PR #2183, opened 2026-07-06
fix(ui): document runbooks-UI abort CSRF gate + surface CSRF-403 and soft-failed admin lift (#2112)

### LINKED ISSUE #2112: runbooks UI: `POST /ui/runbooks/runs/<id>/abort` returns HTTP 403 to the run-starter while `POST /api/v1/runbooks/runs/<id>/abort` permits the identical action — plus a restriction-banner clause promising an admin-bypass that isn't wired
## Summary

Two surface gaps on `/ui/runbooks` surfaced via a manual browser walkthrough as `tenant_admin` on v0.19.0. Filed together because both are about the same UI surface and one (the abort 403) is the harder of the two for an operator to recover from.

### Finding 1 (substantive) — UI cannot abort own run; REST can

The run-starter cannot abort their own active run via the UI:

```
POST /ui/runbooks/runs/<run_id>/abort   → HTTP 403 Forbidden (rendered via htmx)
POST /api/v1/runbooks/runs/<run_id>/abort
   (same run_id, same operator, same browser session window, with the
    operator's bearer JWT)
                                        → HTTP 200
                                        → {"run_id": "...", "state": "abandoned",
                                           "abandoned_at": "..."}
```

The REST endpoint correctly enforces the single-assignee invariant (`run.assigned_to == operator → permit`) and matches the documented role floor for abort. The UI rejects the same action for the same operator through some additional check.

**Net consequence**: an operator who started a run via the UI has **no UI path to abort it**. The run becomes stuck — they can't advance (shouldn't, mid-run) and can't abort. The only recovery is to drop to the REST/CLI/MCP surface, which defeats the UI's value proposition for operators who chose the UI specifically because they prefer not to be on the CLI. In a production case with UI-only access, the run would be permanently stuck.

### Finding 2 (wording defect) — restriction banner promises an admin-bypass that isn't wired

The opacity_floor restriction banner shown on the template view (`/ui/runbooks/<slug>` while the same operator has an active run for that template) reads:

> *"Step details are restricted. Step details are restricted until you complete a run of this template. Once you have a completed (or abandoned) run of `<slug>`, the full steps unlock here for post-run review. **A tenant administrator can see them at any time.**"*

The tester IS `tenant_admin` (verified). Did an exhaustive walkthrough of the page:

- No "Admin view" / "Override" button
- No URL hint (`?admin=true`, `/admin/runbooks/<slug>`, etc.) discoverable
- No toggle / no preference / no alternative path
- The list page filter UI doesn't include an "admin view" toggle either

The banner promise is **aspirational** — either the admin-bypass is unimplemented (banner text written before implementation), or the bypass is restricted to OTHER admins (not run-starter-self) and the wording is wrong.

The second reading is more defensible — it's the *correct* design discipline (the operator driving the runbook can't peek at their own active template, but a supervisor admin can review). But the banner doesn't say that, and a tenant_admin reading the banner will reasonably assume they personally have a bypass available.

### Finding 1 — UI abort 403 vs REST 200

```bash
JWT=$(... tenant_admin JWT ...)

# 1. Find any published template the caller can start
SLUG=$(curl -sk -H "Authorization: Bearer $JWT" \
        '<backplane>/api/v1/runbooks/templates' \
       | jq -r '.templates[] | select(.status=="published") | .slug' | head -1)

# 2. Start a run via REST (matches what the UI Runs tab does internally)
RUN=$(curl -sk -H "Authorization: Bearer $JWT" -H 'Content-Type: application/json' \
       -X POST '<backplane>/api/v1/runbooks/runs' \
       -d "{\"template_slug\":\"$SLUG\",\"target\":\"<any-registered-target>\"}" \
       | jq -r '.run_id')

# 3. UI route abort with the same JWT
curl -sk -H "Authorization: Bearer $JWT" -H 'Content-Type: application/json' \
     -X POST "<backplane>/ui/runbooks/runs/$RUN/abort" -d '{}' \
     -w '\nHTTP=%{http_code}\n'
# → HTTP 403 Forbidden

# 4. REST route abort with the same JWT, same run, same operator
curl -sk -H "Authorization: Bearer $JWT" -H 'Content-Type: application/json' \
     -X POST "<backplane>/api/v1/runbooks/runs/$RUN/abort" \
     -d '{"reason":"clean up"}' \
     -w '\nHTTP=%{http_code}\n'
# → HTTP 200, body {"state": "abandoned", ...}
```

### Finding 2 — banner wording check

Manual: as `tenant_admin`, log in to the UI, start a run for any published template, then in a different tab navigate to `/ui/runbooks/<slug>`. Observe the restriction banner. Search the page (and the surrounding navigation) for any admin-bypass affordance. None found in the walkthrough.

## Probable causes (consumer-side guesses; haven't traced source)

For Finding 1 — three plausible hypotheses, no claim of which is right:

- **UI session principal differs from REST JWT principal** in some claim (e.g. a `tenant_role` or `principal_kind` that the UI's `require_role` machinery reads but the REST's doesn't). The two auth paths might be reading different tokens or applying different mappers.
- **UI has a stricter check than REST** — perhaps the UI's abort action is gated through the same opacity_floor check that's restricting the template view, when abort should always be available to the run-starter regardless of opacity (since the operator who started the run knows they're aborting their own thing — no peek-ahead risk).
- **UI raised the role-floor for abort** independently of REST, perhaps as part of a different surface-protection policy.

Whichever the cause, the user-visible behaviour is: the UI denies an action that REST permits, with no explanation surfaced to the operator beyond a bare htmx-rendered 403.

For Finding 2 — either the admin-bypass is unimplemented or the wording is wrong; can't distinguish without source.

### Ask A — for Finding 1 (UI abort 403)

**Option A1 — UI abort matches REST role-floor**: explicitly permit the run-assignee (regardless of role) to abort their own run via the UI route. The UI is over-restricting an action that needs to remain discoverable to the operator who started the run. This is the obvious user-experience choice.

**Option A2 — surface the gap with a clear error**: if the UI's stricter check is intentional, the 403 response should at minimum carry a user-visible explanation ("Abort restricted to senior `tenant_admin` in this deploy; ask a colleague or use `meho runbook abort <run_id>` via CLI/REST") instead of a bare htmx 403. Minimum-cost stopgap.

### Ask B — for Finding 2 (banner wording)

**Option B1**: implement the admin-bypass — add a UI affordance that actually lifts the opacity restriction for `tenant_admin` (or whichever role the design intends). Banner becomes truthful.

**Option B2**: fix the banner wording to match what's implemented:
- *"A different tenant administrator can see them at any time."* (if the design intent is admin-other-than-self), OR
- Remove the clause entirely (if no bypass is intended for now).

Option B2 is cheap, immediate, and removes the misleading text.

## Related

Filed alongside [`evoila/meho#2111`](https://github.com/evoila/meho/issues/2111) — the REST-side opacity_floor violation. The UI correctly enforces opacity_floor on `/ui/runbooks/<slug>` mid-run via the banner described in Finding 2; the REST endpoint `GET /api/v1/runbooks/templates/<slug>` bypasses the same check. **The engine has all the state needed to enforce opacity_floor** (the UI proves it) — `#2111`'s fix path should mirror what the UI already does.

(A separate evidence-comment on `#2111` from the consumer side will point to this discovery in detail.)

## Out of scope

- The REST-side opacity_floor violation itself (filed at `#2111`).
- The UI's behaviour for OTHER admins viewing the template (would need a second operator with `tenant_admin` claims; not testable from a single session — but worth a paired test if available).
- The UI's behaviour for completed / abandoned runs (per the banner's first clause, opacity should lift post-completion; partially confirmed but not exhaustively probed in this walkthrough).

## Consumer-side trace

Tracking signal: `runbooks-ui-abort-403-self-on-own-run-plus-aspirational-admin-bypass-banner`. Verification probe captures Finding 1 end-to-end (curl-based; auto-cleanup); Finding 2 requires eyeballs on the banner text — manual.

Re-diagnosed against `v0.19.0` code. The reporter got both symptoms right but mis-attributed both root causes; the given RDC correction is itself right on Finding 1 and *wrong* on Finding 2 — the code is the oracle.

### Finding 1 — the 403 is a **CSRF rejection**, not an RBAC / opacity check. Already shipped + tested.

The `/ui` abort route sits on the **same operator floor as REST**, not a stricter gate:

- `POST /ui/runbooks/runs/{run_id}/abort` declares `session: UISessionContext = _require_session` — `require_ui_session`, the operator floor — NOT `_require_admin` (`backend/src/meho_backplane/ui/routes/runbooks/driver.py:554-561`; `_require_session`/`_require_admin` defined at `driver.py:148-149`). Only `/reassign` uses the admin hard-gate (`driver.py:563-570`).
- `_abort` calls `service.abort_run(..., caller_is_admin=probe.is_tenant_admin)` (`driver.py:397-403`) — the **identical** assignee-or-admin allowance REST uses. A failed assignee check raises `NotRunAssigneeError`, which the handler renders as an **inline HTTP-200 fragment** (`_reassigned_away_fragment`, `driver.py:404-405`), never a 403. So a genuine RBAC denial here is a 200, not a 403.
- The only source of a bare `403` on a `/ui/*` POST is the double-submit-cookie CSRF middleware: it returns `403 {"detail":"csrf_token_invalid"}` with an `x-csrf-rejection-reason` header of `missing_token` when the `meho_csrf` cookie is absent (`backend/src/meho_backplane/ui/csrf.py:521-524`, response built at `csrf.py:315-328`; safe-methods/`/api` bypass at `csrf.py:486-500`).

The repro used **raw curl + a Bearer JWT with no `meho_csrf` cookie and no `X-CSRF-Token` header** → `missing_token` → 403. `/api/v1/...` is exempt from CSRF, so the same curl succeeds there — which is exactly the 403-vs-200 differential observed. A real browser driving the run through the driver page carries the cookie/header (minted + refreshed on every fragment render — `driver_render.py:152,213` via `mint_csrf_token`), so the run-starter **can** abort their own run from the UI today.

This is **already covered by tests**, not residual behaviour:
- `backend/tests/test_ui_runbook_driver_opacity.py::test_abort_valid_reason_flips_state_and_audits` (line 751) — assignee self-abort with a valid CSRF token → **HTTP 200**, run flips to `abandoned`, audit row lands.
- `...::test_abort_missing_csrf_rejected_403` (line 778) — the exact repro shape (POST, no CSRF) → **403**, run stays `in_progress`.

So the residual for Finding 1 is **not** a role-floor fix (Option A1 is a no-op — the floor already matches REST). The real residual is Option A2's spirit: the CSRF requirement isn't documented for consumers, and the middleware's bare `{"detail":"csrf_token_invalid"}` JSON is not surfaced as a friendly operator-facing fragment.

### Finding 2 — the admin-bypass **is** wired. The banner is truthful; the real defect is a **silent soft-fail of the role lift**.

Contrary to both the reporter's hypothesis and the RDC correction ("promises an admin-bypass that isn't wired"), the tenant-admin bypass **is** implemented:

- The detail handler resolves `is_admin` and dispatches on it (`backend/src/meho_backplane/ui/routes/runbooks/routes.py:356-357`).
- `_resolve_detail_admin` returns `restricted=False` with the **full** `_render_steps(template)` for a `tenant_admin` (`routes.py:485-501`) — the banner's "A tenant administrator can see them at any time" is accurate and wired.

Why the `tenant_admin` tester still saw the restricted banner: the admin verdict comes from `_resolve_role`, which **fails soft to non-admin** on *any* hiccup — decrypt/reload miss, JWKS transiently unreachable, or a token/session identity mismatch — returning `None` (→ treated as plain operator) with only an info/warning log and **no operator-facing signal** (`routes.py:163-207`, esp. the `except Exception` soft-fail at `routes.py:189-195` and the identity-mismatch branch at `routes.py:200-206`). A `tenant_admin` whose role lift silently degraded lands on the operator path (`_resolve_detail_operator`, `routes.py:504-538`) and sees the restricted banner — which reads as "the promised bypass doesn't exist." The banner wording is fine (Ask B2 is unnecessary; Ask B1 is already done). The residual is the **undiagnosable silent degradation**: when the role lift can't complete for a session that should be admin, the operator gets no hint that the admin view failed to resolve.

Parent goal: #221

## Acceptance criteria

- [ ] Documentation (operator/UI docs or the driver route docstring) states that every `/ui/*` state-changing POST — including `/ui/runbooks/runs/<id>/abort` — requires the double-submit CSRF token (`meho_csrf` cookie + `X-CSRF-Token` header / `csrf_token` form field), and that a raw-curl repro without it yields `403 csrf_token_invalid` by design; `/api/v1/*` is exempt. A `RESTRICTIONS`/known-repro note points reporters at the browser path.
- [ ] A test asserts the intended browser self-abort path succeeds end-to-end for the run-assignee (assignee + valid CSRF → 200 `abandoned` + audit row) AND that the bare no-CSRF POST is a 403 with `x-csrf-rejection-reason: missing_token` (extend/point at `test_abort_valid_reason_flips_state_and_audits` and `test_abort_missing_csrf_rejected_403`), pinning the 403 as CSRF-class not RBAC-class.
- [ ] The htmx CSRF-rejection 403 on `/ui/*` is surfaced to the operator as a readable fragment/message (e.g. "Your session token expired — refresh the page and retry") rather than the bare `{"detail":"csrf_token_invalid"}` JSON body, OR a decision to leave it is recorded with rationale.
- [ ] When the tenant-admin role lift in `_resolve_role` fails soft (JWKS unreachable / session-reload miss / identity mismatch) for a session that should be admin, the restricted detail view surfaces a diagnostic affordance (a note that the admin view could not be resolved + retry), so a real `tenant_admin` is never silently downgraded to the restricted banner with no signal; a test covers the soft-fail-surfacing branch.
- [ ] The banner clause "A tenant administrator can see them at any time" is confirmed truthful by a test exercising the admin path (`_resolve_detail_admin` → `restricted=False` + full steps for a `TENANT_ADMIN`); no wording change is applied (the bypass is already wired).

Parent initiative: #2150

## Decision (recorded 2026-07-06)

**Treat the 403 as the intended CSRF rejection.** The `/ui/*` abort route is correctly CSRF-guarded (double-submit cookie); the raw-curl repro failed only because it sent no CSRF cookie. Fix: (1) document the CSRF-cookie requirement for `/ui/*` POSTs (a legit browser caller with the cookie succeeds — no change to the gate), and (2) reconcile the restriction-banner clause: **remove** the promise of an admin-bypass that isn't wired (or wire it, if that capability is actually wanted). Acceptance: docs + banner reconciled; a test covers the intended browser abort path.

damir-topic:
## Summary

- The reported `POST /ui/runbooks/runs/<id>/abort` **403 vs** `POST /api/v1/.../abort` **200** differential is a **CSRF rejection**, not RBAC/opacity. The `/ui` abort route is on the **same operator floor as REST** (`require_ui_session`, `caller_is_admin=probe.is_tenant_admin`); a genuine assignee denial there renders an inline **HTTP-200** fragment, never a 403. The only 403 source on a `/ui/*` POST is the double-submit-cookie middleware, and `/api/v1/*` is CSRF-exempt — so the raw-curl repro (Bearer JWT, no `meho_csrf` cookie) 403s on `/ui` while succeeding on `/api`. A browser carries the cookie + `X-CSRF-Token` (re-minted every render), so the run-starter **can** abort their own run from the UI today.
- Grounded against v0.19.0: Finding 2's admin-bypass **is** wired (`_resolve_detail_admin` → `restricted=False` + full steps for a `tenant_admin`), so the banner clause *"A tenant administrator can see them at any time"* is **truthful — no wording change applied**. The tester saw the restricted banner only because their admin role lift **silently soft-failed** (`_resolve_role` → `None` on a JWKS/session/identity hiccup). This PR surfaces that degradation instead of removing a correct banner.
- **Docs (AC1):** `docs/codebase/ui.md` gains a *CSRF requirement / RESTRICTIONS* section (cookie + header/form-field required; `403 csrf_token_invalid` + `x-csrf-rejection-reason` by design; `/api/v1` exempt; reproduce against the browser path, not raw curl) + a matching note in the driver-route docstring.
- **CSRF-403 surfacing (AC3):** `session-expiry.js` gains a branch on the same global `htmx:beforeOnLoad` seam that shows a *"session token expired — refresh and retry"* banner, keyed on the `x-csrf-rejection-reason` header so a bare RBAC 403 is untouched. (htmx 2.0.9 does not swap 4xx bodies, so a server-side fragment could not render — the client is the only place this surfaces.)
- **Soft-fail diagnostic (AC4):** `_render_detail` threads `admin_lift_degraded` (`operator is None` uniquely marks a degraded lift; a genuine operator yields a non-`None` `Operator`) and the restricted view renders a diagnostic note + a CSP-safe retry link, so a real `tenant_admin` is never silently dropped onto the opacity banner with no signal.

## Divergence from the ticket's top-line "Decision" (flagged for review)

The ticket's `Decision (recorded 2026-07-06)` says "**remove** the promise of an admin-bypass that isn't wired." The **code oracle contradicts that premise** — the bypass IS wired — and the ticket's own **triage section + acceptance criteria** (AC5: *"no wording change is applied (the bypass is already wired)"*) already resolve the contradiction. This PR follows the acceptance criteria (the contract): it keeps the truthful banner and fixes the **actual** residual (silent soft-fail). Removing the banner would have been a regression.

## Test plan

- [x] `mypy src/.../runbooks/{routes,driver}.py` — clean
- [x] `pytest tests/test_ui_runbooks_lifecycle.py tests/test_ui_runbook_driver_opacity.py` — 45 passed
- [x] Broad CSRF/session/runbook-detail sweep — 132 passed
- [x] **AC2** — `test_abort_missing_csrf_rejected_403` now pins `x-csrf-rejection-reason: missing_token` + `{"detail":"csrf_token_invalid"}` (CSRF-class, not RBAC); `test_abort_valid_reason_flips_state_and_audits` proves assignee self-abort → 200 `abandoned` + audit row.
- [x] **AC5** — `test_detail_admin_sees_full_steps_banner_clause_truthful` (admin path → no restricted banner, full steps).
- [x] **AC4** — `test_detail_soft_failed_admin_lift_surfaces_diagnostic` (degraded lift → diagnostic) + `test_detail_genuine_operator_no_soft_fail_diagnostic` (clean operator → no diagnostic).
- [x] **AC3** — `test_session_expiry_handler_source_carries_csrf_rejection_contract` (JS content contract).
- [x] OpenAPI snapshot regenerated — **byte-identical** (no route signature/response-model change), so no `make generate` needed.

## Out of scope

- The REST-side opacity_floor violation (`#2111`).
- The runbooks-UI *authoring* defects (`#2117`) and the runbook run-state read surface — per Initiative #2150's Out-of-scope.

Closes #2112

### DIFF
--- a/backend/src/meho_backplane/ui/routes/runbooks/driver.py
+++ b/backend/src/meho_backplane/ui/routes/runbooks/driver.py
@@ -81,6 +81,21 @@ on the same response -- refreshes the ``meho_csrf`` cookie (the cookie/
 header desync footgun #1693). The render layer in :mod:`.driver_render`
 owns the mint + cookie refresh.
 
+Every state-changing POST here (``/next`` / ``/abort`` / ``/reassign``)
+REQUIRES the double-submit token: the ``meho_csrf`` cookie AND an
+``X-CSRF-Token`` header (or ``csrf_token`` form field). A request missing
+either half is rejected by the middleware with ``403``
+``{"detail":"csrf_token_invalid"}`` + an ``x-csrf-rejection-reason``
+header -- BY DESIGN, before this handler runs. ``/api/v1/*`` is CSRF-exempt,
+so a raw-curl repro sending a Bearer JWT but no ``meho_csrf`` cookie gets a
+``403`` on ``/ui/.../abort`` while the same call to
+``/api/v1/.../abort`` succeeds: that differential is the CSRF gate, NOT an
+RBAC/opacity difference (this abort route is on the SAME operator floor as
+REST -- see ``_abort`` below -- and a genuine assignee RBAC denial renders
+an inline HTTP-200 fragment, never a ``403``). The run-starter CAN abort
+their own run from the browser, which carries the cookie/header. Consumer
+note + repro guidance: ``docs/codebase/ui.md`` (#2112).
+
 References
 ----------
 
--- a/backend/src/meho_backplane/ui/routes/runbooks/routes.py
+++ b/backend/src/meho_backplane/ui/routes/runbooks/routes.py
@@ -381,9 +381,24 @@ async def _render_detail(
     renders (never a raw 403). A missing slug 404s for an admin and
     collapses to the restricted state for an operator -- the same
     anti-enumeration posture the REST ``show`` route uses.
+
+    Soft-fail visibility (#2112). :func:`_resolve_role` returns the resolved
+    :class:`Operator` on a clean lift (admin or plain operator) and ``None``
+    ONLY on a degraded lift -- a JWKS hiccup, a session-reload miss, a
+    malformed stored token, or a token/session identity mismatch. A genuine
+    plain operator therefore always yields a non-``None`` operator, so
+    ``operator is None`` uniquely marks "the role lift could not complete"
+    -- the exact state that used to drop a genuine ``tenant_admin`` onto the
+    restricted banner with no signal (they read it as "the promised admin
+    bypass doesn't exist"). We thread that as ``admin_lift_degraded`` so the
+    restricted view can surface a diagnostic affordance (a note + a retry)
+    instead of silently downgrading. A terminal ``session_expired`` never
+    reaches here -- :func:`_resolve_role` re-raises it to the app-level
+    re-auth redirect (#2114).
     """
     operator = await _resolve_role(session)
     is_admin = operator is not None and operator.tenant_role == TenantRole.TENANT_ADMIN
+    admin_lift_degraded = operator is None
     detail = await _resolve_detail(session, slug, version, is_admin=is_admin)
     csrf_token = mint_csrf_token(str(session.session_id))
     # The fork-on-edit affordance surfaces how many runs are still pinned to a
@@ -401,6 +416,7 @@ async def _render_detail(
         "steps_rendered": detail.steps_rendered,
         "code_css": pygments_css(),
         "is_admin": is_admin,
+        "admin_lift_degraded": admin_lift_degraded,
         "in_flight_run_count": in_flight_run_count,
         "lifecycle_error": None,
         "swap_badge": False,
--- a/backend/src/meho_backplane/ui/static/src/app/session-expiry.js
+++ b/backend/src/meho_backplane/ui/static/src/app/session-expiry.js
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (c) 2026 evoila Group
 //
-// App-shell session-expiry safety net (#122).
+// App-shell session-expiry + CSRF-rejection safety net (#122, #2112).
 //
 // When a ``/ui/*`` htmx request (``hx-get`` / ``hx-post``) returns ``401``
 // mid-session, htmx 2.0.9 silently swallows it: the default
@@ -15,9 +15,28 @@
 // the ``ui.auth.errors`` handler returns the JSON 401 body rather than the
 // HTML login redirect -- the browser sees a swallowed error, not a page).
 //
-// This module is the client-side last-resort recovery path: any ``401`` on
-// an htmx request surfaces a visible banner ("your session expired -- sign
-// in again") with a button to ``/ui/auth/login?return_to=<current path>``.
+// The same swallow-on-4xx behaviour hides the CSRF middleware's ``403``
+// (#2112). A state-changing ``/ui/*`` POST whose ``meho_csrf`` double-submit
+// cookie was dropped or aged out (the browser rarely hits this because every
+// fragment render re-mints the cookie, but a stale tab or a cookie-cleared
+// session can) gets a bare ``{"detail":"csrf_token_invalid"}`` JSON body at
+// ``403`` that htmx will not swap -- so an operator aborting their own
+// runbook run saw a dead button with no message. Returning an HTML fragment
+// from the pure-ASGI middleware would not help: htmx 2.0.9 does not swap 4xx
+// bodies (``{code:"[45]..", swap:false, error:true}`` in the vendored
+// bundle). The client is the only place that can surface it. The middleware
+// stamps every CSRF rejection with an ``x-csrf-rejection-reason`` response
+// header (``missing_token`` / ``value_mismatch`` / ``signature_invalid`` /
+// ``no_session``); that header is the discriminator -- an RBAC ``403`` (e.g.
+// ``require_ui_admin`` on reassign) carries no such header and flows through
+// untouched.
+//
+// This module is the client-side last-resort recovery path:
+//   * any ``401`` -> a "your session expired -- sign in again" banner with a
+//     button to ``/ui/auth/login?return_to=<current path>``.
+//   * a ``403`` carrying ``x-csrf-rejection-reason`` -> a "your session
+//     token expired -- refresh the page and retry" banner with a reload
+//     button (a fresh render re-mints the ``meho_csrf`` cookie).
 // It is the safety net (Axis B); the server-side refresh seam (#1694, and
 // the sibling #121) reduces how *often* a 401 occurs (Axis A). Both land.
 //
@@ -25,14 +44,15 @@
 // errors) before any swap, redirect-header handling, or ``responseError``,
 // and ``preventDefault()`` on it aborts htmx's entire response processing
 // for that request (vendored htmx 2.0.9: ``Vn()`` returns early when the
-// event helper reports the default was prevented). We act ONLY on status
-// ``401`` -- every other status (a ``2xx`` swap, a ``422`` form-validation
-// re-render, a ``403`` CSRF error a form's ``hx-on::response-error`` handles)
-// flows through untouched, so legitimate non-auth responses are never
-// hijacked. Chosen over ``htmx.config.responseHandling`` (which would
-// reclassify status codes app-wide and risk the finely-tuned 4xx swap
-// behaviour the connectors/agents/kb form modals depend on -- #875, forms.py
-// 422 re-render) precisely because it is surgical: one status, one action.
+// event helper reports the default was prevented). We act ONLY on a ``401``
+// or a CSRF-rejection ``403`` -- every other status (a ``2xx`` swap, a
+// ``422`` form-validation re-render, a bare RBAC ``403`` a form's
+// ``hx-on::response-error`` handles) flows through untouched, so legitimate
+// non-auth responses are never hijacked. Chosen over
+// ``htmx.config.responseHandling`` (which would reclassify status codes
+// app-wide and risk the finely-tuned 4xx swap behaviour the
+// connectors/agents/kb form modals depend on -- #875, forms.py 422 re-render)
+// precisely because it is surgical: two narrow triggers, one action each.
 //
 // Registered via a plain ``htmx:beforeOnLoad`` listener on ``document.body``
 // (htmx events bubble to the body -- the same delegation pattern
@@ -43,7 +63,7 @@
 // first-party JS (no Alpine, no inline script) to preserve the chassis
 // "zero inline script" CSP posture, mirroring ``theme.js`` / ``modal-dialogs.js``.
 //
-// Idempotent banner: the recovery banner is created once and reused -- a
+// Idempotent banners: each recovery banner is created once and reused -- a
 // burst of failing background polls (the approvals badge, an SSE reconnect)
 // all hitting 401 must not stack N banners. ``preventDefault()`` also stops
 // the dead swap, so the operator's current view stays intact behind the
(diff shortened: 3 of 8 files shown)


damir-topic:
@coderabbitai review

damir-topic (commented):




**Anchor task:** #2112 — runbooks-UI abort 403 is a CSRF gate (not RBAC); banner admin-bypass is truthful; real residual is the silent soft-fail of the admin role lift
**Surface area:** backend + frontend (Jinja template + first-party JS)
**Best-practices packs audited:** python, frontend
**Verdict:** APPROVE
**Head SHA reviewed:** `218b81b9`
**AI reviewer (CodeRabbit):** not received — rate-limited (posted only a "review limit reached, next in 13 min" warning; no findings performed for this iteration)

### Adjudication of the flagged divergence (banner kept, not removed)

The ticket's top-line **"Decision (recorded 2026-07-06)"** says to *remove* the admin-bypass banner clause "that isn't wired." The implementer kept it and instead fixed the silent soft-fail. **I independently verified the code oracle and the implementer is correct:**

- `_resolve_detail_admin` (`routes.py:539-555`) returns `restricted=False` with the full `_render_steps(template)` for a `tenant_admin`.
- `_render_detail` (`routes.py:400`) derives `is_admin` from `operator.tenant_role == TENANT_ADMIN`; `_resolve_detail` (`routes.py:532`) dispatches admins to `_resolve_detail_admin`.
- Therefore the banner clause *"A tenant administrator can see them at any time"* is **truthful and wired.** The tester saw the restricted banner only because their role lift silently soft-failed (`_resolve_role` → `None` on a JWKS/session/identity hiccup).

The ticket's **own** "Root-cause correction (triage 2026-07-05)" section and **AC5** (*"no wording change is applied — the bypass is already wired"*) explicitly correct the premise of the top-line Decision. The acceptance criteria are the binding contract; removing the truthful banner would have been a regression. **Divergence justified.**

### Acceptance criteria
- [x] **AC1** CSRF docs — `docs/codebase/ui.md` gains a RESTRICTIONS section (cookie + `X-CSRF-Token`/`csrf_token` required; `403 csrf_token_invalid` + `x-csrf-rejection-reason` by design; `/api/v1` exempt; reproduce against browser path) + matching driver.py docstring. Accurate against `csrf.py`.
- [x] **AC2** abort 403-is-CSRF-class + self-abort 200 — `test_abort_missing_csrf_rejected_403` pins `x-csrf-rejection-reason: missing_token` + `{"detail":"csrf_token_invalid"}`; `test_abort_valid_reason_flips_state_and_audits` proves assignee self-abort → 200 `abandoned` + audit row.
- [x] **AC3** client-side CSRF-403 banner — verified htmx 2.0.9's default `responseHandling` maps `[45]..` to `swap:false` in the vendored bundle, so a server fragment cannot render; client-side is the only surface. JS keys on `x-csrf-rejection-reason`, which every CSRF rejection path stamps (`_forbidden_response`, `csrf.py:326`); a bare RBAC 403 (no header) flows through untouched.
- [x] **AC4** `admin_lift_degraded` diagnostic — `_render_detail` threads `admin_lift_degraded = operator is None`; template renders a CSP-safe note + same-page retry link; `test_detail_soft_failed_admin_lift_surfaces_diagnostic` + `test_detail_genuine_operator_no_soft_fail_diagnostic` cover degraded and the negative (clean operator, no leak).
- [x] **AC5** banner truthful — `test_detail_admin_sees_full_steps_banner_clause_truthful` exercises the real admin path; no wording change.

### Project-board alignment
|---|---|---|---|---|---|
| #2112 | Open | Closes | ✅ | ✅ full | close on merge |

Parent initiative #2150, parent goal #221. No delivered-but-unlinked or linked-but-undelivered mismatches.

### Predicted vs. actual diff (against `origin/main`)
- ✅ `driver.py` — CSRF-requirement docstring on the abort route
- ✅ `routes.py` — `admin_lift_degraded` threaded into `_render_detail`
- ✅ `detail.html` — CSP-safe soft-fail diagnostic block
- ✅ `session-expiry.js` — CSRF-403 branch on the existing `htmx:beforeOnLoad` seam
- ✅ `docs/codebase/ui.md` — RESTRICTIONS/CSRF section
- ✅ 3 test files — AC2/AC3/AC4/AC5 coverage
- No undisclosed/stealth edits; true delta is exactly the 8 reported files (373/36). No route signature/response_model change → OpenAPI snapshot byte-identical.

### Integration verification
- [x] `ruff check` (runbooks routes) — clean
- [x] `ruff format --check` (driver.py, routes.py) — clean
- [x] `mypy` (routes.py, driver.py) — clean
- [x] `pytest` — 80 passed (test_ui_runbooks_lifecycle + test_ui_runbook_driver_opacity + test_ui_templates), run against the PR-head worktree source
- [x] htmx 2.0.9 4xx-no-swap claim verified against the vendored bundle (`version:"2.0.9"`, `[45]..` → `swap:false, error:true`)
- [x] CSRF rejection-reason header stamping verified on all four rejection paths (`csrf.py:315-333, 458-464, 517-540`)
- ⚠️ Note: a naive `pytest` inside `.claude/worktrees/pr-2183-review` reads the editable-installed `meho_backplane` from the **main checkout** (known `.claude`-symlink hazard), producing 3 spurious failures; re-running with the worktree `src` on `PYTHONPATH` (imports proven from the worktree path) → all green.

### Findings (inline)
None. No blockers, no majors, no minors requiring a fix-up.

- Adopted: 0
- Disputed: 0
- Out of scope: 0
- CodeRabbit was rate-limited this iteration and performed no review (only a "review limit reached" notice). Recorded `received: false`.

### Docs
- [x] `docs/codebase/ui.md` — updated (CSRF RESTRICTIONS section + errors-handler row + expired-callback recovery note). Accurate against source.

### What worked well
- The soft-fail discriminator is genuinely sound: `_resolve_role` returns a non-`None` `Operator` for a clean plain-operator lift, so `operator is None` uniquely marks a degraded lift — and the negative test guards it against leaking onto every operator restricted view.
- The CSRF-403 client net keys on the `x-csrf-rejection-reason` header rather than the bare 403 status, so it never hijacks a genuine RBAC denial; the header is stamped on every CSRF rejection path, so it never misses one.
- Docstrings + docs re-diagnose the reporter's misattribution in place (the recurring RDC symptoms-right/root-cause-wrong pattern), which prevents the next reporter from re-filing.

PR #2015, opened 2026-06-21
fix(connectors): migration 0049 — retire stale long-product (vcf-logs) ingest orphan rows (#2001)

damir-topic:
## Summary

- Adds Alembic migration **`0049`** (`down_revision="0048"`) that DELETEs the built-in `tenant_id IS NULL` long-product operation rows (`operation_group` + `endpoint_descriptor`) that migration `0038`'s collision-guard deliberately skipped — the operation-row complement of that backfill and of the `0047` `targets.product` reconciliation (#1814).
- These orphans (`product='vcf-logs'` / `impl_id='vrli-rest'` etc.) are **non-dispatchable** (every probe keys on the short, `parse_connector_id`-derived product) and **unreachable by `meho.connector.delete`** (the resolver derives `product='vrli'` from `vrli-rest-9.0.2`; no product selector exists on the route/MCP tool), so they could only ever return `ConnectorNotFoundError` (#1910). They also surfaced as a redundant `vcf-logs` candidate on `review`/`enable_reads` of the live short twin.
- The retire predicate is the **exact inverse** of `0038`'s collision-skip: `tenant_id IS NULL` AND `product='<long>'` AND the `parse_connector_id` first-hyphen-segment rule on `impl_id` AND a live short-product twin **EXISTS** on the same global natural key. A twin-less long row is preserved (left for `0038`'s write-side rewrite) so no operation loses its only representation.
- Self-contained `sa.table`/`sa.column` shims (no ORM import), idempotent DELETE, documented no-op downgrade (the orphans were never dispatchable). Mirrors the `0038`/`0046`/`0047` precedent.

## Test plan

- [x] `alembic heads` — reports a single head `0049` after the migration
- [x] Full chain `alembic upgrade head` on fresh SQLite — replays clean through `0048 -> 0049`
- [x] `ruff check` / `ruff format --check` — clean
- [x] `mypy --ignore-missing-imports` — clean
- [x] `pytest tests/test_migration_0049_retire_stale_vcf_logs_orphan_rows.py` — 8 passed
- [x] `pytest tests/test_alembic_uuid_param_consistency.py` — 52 passed (drift guard; this migration writes no UUID binds)

### Acceptance criteria

- [x] New `0049_retire_stale_vcf_logs_orphan_rows.py` with `down_revision="0048"`; single head after it.
[...]

Issue #2001, opened 2026-06-20
Task: migration 0049 — retire the stale long-product (vcf-logs) ingest orphan rows #1814 left behind


damir-topic:
## Summary

The long↔short connector product reconciliation (#1814 = migration `0047`) realigned `targets.product` but **deliberately scoped out the ingest-row data migration**, leaving a stale `product=vcf-logs` / `tenant_id=NULL` row for `vrli-rest-9.0.2` that operators **cannot delete** (the resolver derives `product="vrli"` from the connector_id, so `delete` can't target the `vcf-logs` row → `ConnectorNotFoundError`). This task adds the missing data migration to retire those orphaned long-product ingest rows.

## Current state

- `delete_connector(connector_id, tenant_id)` ([`operations/ingest/service.py:734`](backend/src/meho_backplane/operations/ingest/service.py#L734)) resolves via `parse_connector_id`, which derives the product from the connector_id's first hyphen-segment (`vrli-rest-9.0.2` → `product="vrli"`). There is **no `product` selector** on the REST route or the MCP tool, so a row physically stored under `product="vcf-logs"` for the same `(version, impl_id)` is unreachable → `ConnectorNotFoundError` is the only outcome.
- These orphans are exactly what the `endpoint_descriptor` product-split backfill's collision-guard intentionally leaves behind: when a short `vrli` twin already exists under `tenant_id IS NULL`, the long `vcf-logs` row is **skipped, not rewritten**, with that migration's own docstring delegating cleanup to "the DELETE surface (#1700)" — which provably cannot reach a divergent-product row.
- #1814 = `0047_reconcile_split_connector_target_product.py` is **already shipped** (current alembic head = `0048`, chain `…0046→0047→0048`) and only rewrites `targets.product` — it neither did nor can retire the `operation_group`/`endpoint_descriptor` orphan. A **new** migration is required.
- **Migration head verified:** `0048_seed_vmware_collection_manifest.py` is the sole leaf (nothing revises it); `0049`/`0050` are free. Lane A (config-driven connectors #1965/#1966) is migration-free (it ships ExecutionProfiles as package data, no DB schema), so `0049` is uncontested by Lane A — the only same-wave claimant is this Initiative's TLS task (which takes `0050`).

## Desired state

Add migration **`0049`** (`down_revision = "0048"`, the current single head) that DELETEs `operation_group` + `endpoint_descriptor` rows where:

[...]

PR #2185, opened 2026-07-06
feat(connectors): migration 0052 — retire the registered-stub-twin vcf-logs orphan 0049 missed

### LINKED ISSUE #2068: Task: migration 0051 — retire the registered-stub-twin vcf-logs orphan migration 0049 (#2015) misses (0-op short stub at a divergent version)

damir-topic:
## Summary

- Adds data-only Alembic migration `0052` (revises head `0051`) that retires the stale long-product `vcf-logs` ingest orphan that migration `0049` (#2001 / PR #2015) missed — the case where the short (`vrli`) representation is a **0-row v2 class-registry stub at a divergent version** (`9.0` vs the orphan's `9.0.2`), so `0049`'s per-op `EXISTS` twin probe matched nothing and the 136-op orphan survived every upgrade (the #2068 finding; the #1910 review-shadow dead weight).
- The retire predicate keys on the orphan's **OWN** attributes — `tenant_id IS NULL` + a `_PRODUCT_SPLITS` long product + the `parse_connector_id` first-hyphen-segment `impl_id` rule (`operations/_lookup.py:57`) — with **no** DB-visible twin probe. Any `EXISTS` probe against `endpoint_descriptor`/`operation_group` returns false for a 0-row stub and would re-encode the exact `0049` miss; keying on `impl_id` is safe because each split's short product is a registered connector *class* (the live representation), so an in-split long orphan is never the sole live copy.
- Preserves `0049`'s other guards: tenant rows never deleted; `impl_id`-family scoping (a `vcf-logs` row whose `impl_id` derives a non-split product is left alone); `endpoint_descriptor` and `operation_group` retired together; `downgrade()` a documented no-op.

## Test plan

- [x] `ruff check` — clean (migration + both test files)
- [x] `mypy alembic/versions/0052_*.py --ignore-missing-imports` — no issues
- [x] `pytest tests/test_migration_0052_*.py` — 7 passed
- [x] `pytest tests/test_migration_0049_*.py` — 8 passed (regression; `test_twin_less_long_row_preserved` re-pinned to `upgrade("0049")` since `0052` deliberately supersedes that intermediate contract for the in-split subset)
- [x] Load-bearing discriminator proven: temporarily patching `0052` to a naive DB-`EXISTS`-twin implementation (0049's shape) makes `test_retire_with_zero_row_registered_stub_twin_at_divergent_version` **FAIL** — the test genuinely rejects the wrong implementation
- [x] `alembic heads` → single head `0052`; full chain `upgrade head` / `downgrade -1` / re-`upgrade head` clean on fresh SQLite

### Acceptance criteria

- [x] In-split `vcf-logs` orphan retired even with a 0-row `registered` class-registry stub twin at a divergent version — predicate keys on the orphan's own `impl_id`, not a DB twin
- [x] `product='vcf-logs'` row whose `impl_id` derives a non-split product is preserved
- [x] Tenant-scoped (`tenant_id IS NOT NULL`) rows never deleted
- [x] Self-idempotent (stamp-back replay is a no-op); `downgrade()` no-op
- [x] Migration test pinned to its own revision (`stamp("0051") → upgrade("0052")`), covers the missed case + full-twin + non-split-impl_id + tenant + idempotency + dispatch parity

## Out of scope

- A `product` selector on `connector delete` (the #1910 reachability gap) — orthogonal.
- Re-running / amending `0049` (merged #2015) — `0052` is the forward complement.

Closes #2068

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed leftover orphan records from long-form `vcf-logs` entries during upgrade, helping keep related data consistent.
  * Preserved valid rows that belong to other tenants or unrelated identifier families.
  * Ensured the cleanup is safe to run multiple times without changing the final result.

* **Tests**
  * Added coverage for the migration’s cleanup behavior, retention rules, and idempotency.

### DIFF
--- a/backend/alembic/versions/0052_retire_registered_stub_twin_vcf_logs_orphan.py
+++ b/backend/alembic/versions/0052_retire_registered_stub_twin_vcf_logs_orphan.py
@@ -1,0 +1,239 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""Retire the vcf-logs orphan ``0049`` missed (0-row registered stub twin).
+
+Revision ID: 0052
+Revises: 0051
+Create Date: 2026-07-06
+
+Initiative #2065 (G0.29 v0.19.0 closed-loop dogfood hardening), Task #2068.
+The forward complement of migration ``0049`` (#2001 / PR #2015): the
+same long-product ``vcf-logs`` ingest orphan, retired for the case
+``0049``'s live-twin ``EXISTS`` probe could not reach.
+
+Why this migration exists
+-------------------------
+
+Migration ``0049`` retires a long-product orphan
+(``product='vcf-logs'``, ``impl_id='vrli-rest-…'``) only when a
+short-product twin already exists **on the same per-op / per-group
+natural key** -- its guard is a correlated ``EXISTS`` on
+``endpoint_descriptor`` / ``operation_group`` requiring the short twin
+to carry a matching ``op_id`` / ``group_key`` at the **same version**.
+
+On the RDC v0.19.0 upgrade (#2068 finding) the orphan survived anyway.
+The canonical short representation (``vrli``) is registered via the
+**class-side v2 connector registry**, not by an ingest -- so it carries
+**zero** ``endpoint_descriptor`` / ``operation_group`` rows, and it sits
+at a **divergent version** (``9.0`` vs the orphan's ``9.0.2``).
+``0049``'s per-op ``EXISTS`` probe therefore matches nothing and retires
+nothing, leaving the 136-op orphan in place through every upgrade. That
+orphan is exactly the undeletable review-shadow dead weight ``0049`` and
+#1910 set out to clear: non-dispatchable (dispatch keys on the short
+``parse_connector_id``-derived product) and unreachable by
+``meho.connector.delete`` (which resolves a divergent-product row to
+``ConnectorNotFoundError``, #1910).
+
+Fix shape -- key on the orphan's OWN attributes, not a DB twin
+--------------------------------------------------------------
+
+The critical correction over ``0049`` (and over this task's own earlier
+draft, which prescribed a looser connector-grain DB twin probe): the
+retire predicate must **NOT depend on any DB-visible short twin**. A
+class-registry short connector has no ``endpoint_descriptor`` /
+``operation_group`` rows, so *any* ``EXISTS`` probe against those tables
+-- ``0049``'s per-op correlation or a connector-grain one -- returns
+false for a 0-row stub and retires nothing, re-encoding the exact miss.
+
+Instead the predicate keys entirely on the orphan row itself:
+
+* ``tenant_id IS NULL`` -- built-in / global rows only. Tenant-scoped
+  rows are operator-owned (#1699 NULL/UUID-namespace contract) and are
+  **never** deleted. Same boundary ``0049`` / ``0038`` drew.
+* ``product = '<long>'`` -- the row still carries the pre-#1677 registry
+  spelling (one of the six ``_PRODUCT_SPLITS`` long products).
+* ``impl_id = '<short>' OR impl_id LIKE '<short>-%'`` --
+  ``parse_connector_id``'s first-hyphen-segment rule
+  (``operations/_lookup.py:57``) in portable SQL. This scopes the
+  DELETE to rows the dispatcher *would* resolve under the short
+  spelling, so a ``product='vcf-logs'`` row whose ``impl_id`` derives
+  some *other*, non-split product is left alone -- impl_id-family
+  scoping, identical to ``0049``.
+
+There is deliberately **no** ``EXISTS`` clause.
+
+Why retiring without a DB twin is safe
+--------------------------------------
+
+The short product for each split in ``_PRODUCT_SPLITS`` is a **registered
+connector class** in the v2 registry -- the live, dispatchable
+representation of that connector, present regardless of whether it has
+ingested any operation rows yet. The dispatch / query surface
+(:func:`~meho_backplane.operations._lookup.connector_exists`,
+``count_known_ops``, ``list_operation_groups``) keys on the short
+``parse_connector_id``-derived product, so a long-product orphan whose
+``impl_id`` derives a known-split short product is **always**
+non-dispatchable and is **never** the sole *live* representation of that
+connector -- the registered short class is. Retiring it therefore loses
+no reachable data. This class-registry invariant is what ``0049``'s
+per-op DB probe tried and failed to approximate; keying on the orphan's
+own ``impl_id`` states it directly.
+
+Relationship to ``0049``'s twin-less guard
+------------------------------------------
+
+``0049`` deliberately preserved a *twin-less* long row on the theory it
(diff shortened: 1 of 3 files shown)


damir-topic:
@coderabbitai review

Issue #1757, opened 2026-06-14
Task: surface a composite as catalog-unbacked (next_step) instead of enabled-but-broken


damir-topic:
Parent goal: #221
Parent initiative: #1752

## Summary

`gh.composite.pr_status_summary` shows as **enabled** in `GET /api/v1/connectors` but fails at first dispatch — its three L2 sub-ops aren't ingested, so it raises `composite_l2_missing` before any HTTP call. A composite advertised as enabled-but-unbacked reads as broken to an operator.

## Root cause (grounded on v0.15.0)

- The composite's preflight looks up each L2 sub-op descriptor and raises `CompositeL2DependencyMissing` (with the catalog command) when any is absent (`connectors/github/composites/_preflight.py:151-169`, called from `_read.py:331-336`).
- The `gh/3` catalog **entry** exists (`operations/ingest/catalog.yaml:173-221`) but its L2 primitives are **not ingested by default** — the operator must run `meho connector ingest --catalog gh/3` (`connectors/github/_catalog_command.py:41-63`) to populate `endpoint_descriptor`.
- Nothing reflects this gap in the connector listing: the composite shows "enabled" while its dependencies are unbacked.

## Proposed fix (lowest-friction — surfacing, not auto-ingest)

Surface an enabled-but-unbacked composite on the connector listing: when a composite op's L2 sub-ops aren't ingested, attach a `next_step` (`meho connector ingest --catalog gh/3`) / an "unbacked" marker so an operator sees "enabled — run the catalog ingest first" rather than a silent dead-end at dispatch.

## Acceptance criteria

- [ ] `GET /api/v1/connectors` (or the op listing) marks `gh.composite.pr_status_summary` as unbacked with a `next_step` catalog-ingest hint while its L2 sub-ops are absent. Verifiable: `pytest backend/tests/...connectors_listing...` asserts the marker/next_step appears when sub-ops are missing and disappears once ingested.
- [ ] After `meho connector ingest --catalog gh/3`, the composite dispatches cleanly (no `composite_l2_missing`).
- [ ] No false marker on composites whose sub-ops are fully ingested.

## Out of scope

- **Auto-ingesting catalogs at boot** — couples to the catalog-driven ingest line; a separate decision. This task only makes the existing gap visible.

## References

- Signal: #1751 (smaller findings); related #1601 (composite L2 ladder).
[...]

PR #2186, opened 2026-07-06
fix(connectors): assert gh composite L2 backing at connector load (#2050)

### LINKED ISSUE #2050: Task: gh-rest-3 ships gh.composite.pr_status_summary enabled but L2 sub-ops uningested (composite_l2_missing on fresh v0.18.0)
Parent goal: #221
Parent initiative: #2065

> **Migrated from evoila-bosnia/meho-internal#120** — this is connector product-code (not security / not G2.x CI-bootstrap), so it belongs on the public board per the repo-split convention. Privacy-checked: carries no coordinated-disclosure content.

> **Rolled into G0.29 (#2065):** a v0.18.0 finding whose origin cycle (G0.28 #1998) closed before it landed; folded into the current open dogfood-hardening cycle per the series' roll-in convention. Still a log-sentry enabler (#1135).

> **Re-diagnosis note:** the enabled-but-unbacked built-in composite is arguably working-as-designed (the documented path is `meho connector ingest --catalog gh/3` first; cf. the #1757 disposition). The product ask here is independent of that: a fresh deploy should not advertise an *enabled* read op that can't dispatch — ship the L2 deps **curated** in the built-in image, or ship the composite **disabled** with its `next_step`, plus a connector-load assertion that an enabled composite's L2 deps are resolvable. **Confirmed still open at HEAD:** #1757 only added the `unbacked` *listing marker* (`connectors/github/composites/_register.py:162-180`); the composite still ships enabled with its three L2 sub-ops uncurated and no load-time assertion, so the first dispatch still trips `composite_l2_missing`.

## Summary

On a fresh **v0.18.0** deploy, the built-in `gh-rest-3` connector (tenant_id=null) ships its only enabled op — the composite `gh.composite.pr_status_summary` — **enabled but unbacked**: its three GET L2 sub-ops are not curated, so the first dispatch fails before any HTTP call. The meho-native GitHub **read** surface is therefore non-functional out-of-the-box.

## Reproduction (live v0.18.0, git `4d865aca`)

```
call_operation(connector_id=gh-rest-3, target=evoila-bosnia-gh,
  op_id=gh.composite.pr_status_summary,
  params={owner:evoila-bosnia, repo:claude-rdc-hetzner-dc, pull_number:1436})
→ status=error
  error=composite_l2_missing: composite 'gh.composite.pr_status_summary' depends on sub-ops not
        curated for this connector: [GET:/repos/{owner}/{repo}/pulls/{pull_number},
        GET:/repos/{owner}/{repo}/commits/{ref}/check-runs,
        GET:/repos/{owner}/{repo}/pulls/{pull_number}/reviews]
  extras.catalog_command="meho connector ingest --catalog gh/3"
```

`search_operations` already flags the op `unbacked: true` with the same next_step.

## Observed vs expected

- **Observed:** the connector advertises one enabled op that cannot dispatch; an operator following the daily-driver path gets `composite_l2_missing` before any call.
- **Expected:** either (a) the composite's three GET L2 deps ship **curated** in the built-in connector so the daily-driver read works on a fresh deploy, **or** (b) the composite ships **disabled** with its `next_step` until `meho connector ingest --catalog gh/3` is run — rather than enabled-but-unbacked, which advertises a working read it can't perform.

## Suggested fix

Ship `gh.composite.pr_status_summary` with its three GET L2 sub-ops curated in the built-in `gh-rest-3` image (preferred), or ship it disabled. Add a connector-load assertion that an enabled composite's L2 deps are resolvable, so this can't regress silently.

## Acceptance criteria

- On a fresh deploy (no `--catalog gh/3` ingest run), `call_operation(gh-rest-3, gh.composite.pr_status_summary, …)` either dispatches successfully (option a — L2 deps curated in the built-in image) **or** the op is listed/dispatched as **disabled** with its `next_step` (option b) — never enabled-but-`composite_l2_missing`.
- A connector-load assertion fails loudly (at registration / a test) if an **enabled** composite's `sub_op_ids` are not resolvable for the built-in connector — guarding against silent regression. The `sub_op_ids` single-source-of-truth already exists (`_register.py:122-127`), so the assertion reads that tuple.
- `search_operations` continues to surface the op's backing state truthfully (the #1757 `unbacked` marker stays accurate under whichever option is chosen).

## Out of scope

- Auth / wire-shape: the call routes cleanly via connector_id+target (no auth_model error); the gh write-body shape is correct as of v0.18.0 (the prior `gh-l2-dispatch-requestbody-mangled` is FIXED). Purely a composite-curation / packaging-default gap.
- The general "should built-in connectors ship L2 ops curated vs. catalog-on-demand" policy (#1757 disposition) — this task only fixes the **enabled-but-unbacked** inconsistency for the one shipped composite.

## References

- Origin cycle (closed): G0.28 #1998; prior disposition: #1757 (`unbacked` listing marker).
- Found during the RDC consumer **GitHub & Vault test campaign** on v0.18.0 (`claude-rdc-hetzner-dc#1440`, scenario gv-01). Blocks the Log-Sentry workflow (`claude-rdc-hetzner-dc#1135`), which needs the meho-native GitHub surface.
- Code: `connectors/github/composites/_register.py` (composite spec + import-time backing registration), `operations/composite.py:233` (`composite_l2_missing`).
- Best-practices anchors — *devops/packaging*: ship a default that works out-of-the-box or is honestly disabled (no enabled-but-broken op); fail-loud load-time assertion over silent runtime dead-end. *python*: single source of truth for the L2 dep set (already `_register.py:122-127`).

damir-topic:
## Summary

- The built-in `gh-rest-3` connector ships `gh.composite.pr_status_summary` **enabled**, but its three raw-REST L2 sub-ops only land after an operator runs `meho connector ingest --catalog gh/3`. On a fresh deploy the first dispatch trips `composite_l2_missing`.
- #1757 already added the listing `unbacked` marker (the op is surfaced enabled-but-unbacked with a catalog-ingest `next_step`, not as a working read) and the dispatch already returns a structured `composite_l2_missing` carrying the same catalog command — but nothing stopped a **future** composite from shipping enabled with no backing entry at all, silently regressing to the pre-#1757 dead-end.
- Adds `_register_and_assert_composite_backings` at **connector load** (import time): after registering each composite's backing it asserts every enabled composite dispatching raw L2 sub-ops has a backing whose `sub_op_ids` match the spec, raising `UnbackedEnabledCompositeError` otherwise.
- The check reads the spec's own `sub_op_ids` — the same `_register.py` spec ⇄ `_read.py` `_SUB_OPS_*` single source of truth the preflight and the marker already share — so it asserts the **wiring**, not DB state (legitimately empty on a fresh deploy before catalog ingest). This makes the enabled-but-unbacked inconsistency non-regressible **without crashing fresh deploys**, where the raw L2 ops are genuinely absent until ingest.

## Design note (amended ticket)

The ticket was amended 2026-07-05 to ratify a previously-blocked design after a prior auto-run parked it: option (a) "curate L2 in the built-in image" has no substrate (there is no lifespan/import path to seed raw-REST descriptors — that is the rejected Option-C of the preflight design and explicitly out of scope), and option (b) `is_enabled=False` makes the composite vanish from `search_operations` (the `is_enabled` filter). This PR implements the amended design: keep the composite **listed-but-disabled-with-`next_step` via the existing `unbacked` marker machinery** (`is_enabled` stays `True`), plus the connector-load assertion. The assertion is anchored on the backing **wiring** (what drives the `unbacked` safety net), not raw DB resolvability, so it holds on fresh deploys instead of crashing them.

## Test plan

- [x] `mypy src/.../composites/_register.py --ignore-missing-imports` — clean
- [x] `pytest tests/test_connectors_github_composites_register.py tests/test_operations_composite_backing.py` — 24 passed (4 new: load-path passes on shipped specs; raises on missing backing; raises on sub_op_id drift; skips recursion-only composites)
- [x] `pytest tests/test_connectors_github_composites_l2_preflight.py tests/test_connectors_github_composites_read.py` — 26 passed (no collateral)
- [x] **AC#1**: composite listed `unbacked=True` + `next_step` on a fresh deploy; dispatch surfaces structured `composite_l2_missing` with the catalog command — never a silent enabled-but-broken read (existing #1757 machinery preserved; backing tests green)
- [x] **AC#2**: connector-load assertion fails loudly when an enabled composite's `sub_op_ids` lack a resolvable backing — new tests
- [x] **AC#3**: `unbacked` marker stays truthful (backing registration preserved verbatim inside the new helper; `test_operations_composite_backing.py` unchanged and green)

## Out of scope

- Building any lifespan/import mechanism to seed built-in raw-REST L2 descriptors (option a) — no substrate exists and it is the ticket's out-of-scope "built-in connectors ship L2 curated vs. catalog-on-demand" policy.
- The generic composite-registration path (`register_composite_operation`) — vmware-rest composites deliberately do not register backings, so the guard is gh-specific at the gh composite load path.

Closes #2050

### DIFF
--- a/backend/src/meho_backplane/connectors/github/composites/__init__.py
+++ b/backend/src/meho_backplane/connectors/github/composites/__init__.py
@@ -35,6 +35,7 @@ from meho_backplane.connectors.github.composites._read import (
     pr_status_summary_composite,
 )
 from meho_backplane.connectors.github.composites._register import (
+    UnbackedEnabledCompositeError,
     register_github_composite_operations,
 )
 from meho_backplane.operations.typed_register import register_typed_op_registrar
@@ -46,6 +47,7 @@ from meho_backplane.operations.typed_register import register_typed_op_registrar
 register_typed_op_registrar(register_github_composite_operations)
 
 __all__ = [
+    "UnbackedEnabledCompositeError",
     "pr_status_summary_composite",
     "preflight_l2_dependencies",
     "register_github_composite_operations",
--- a/backend/src/meho_backplane/connectors/github/composites/_register.py
+++ b/backend/src/meho_backplane/connectors/github/composites/_register.py
@@ -63,11 +63,60 @@ from meho_backplane.connectors.github.composites.schemas import (
     PR_STATUS_SUMMARY_PARAMETER_SCHEMA,
     PR_STATUS_SUMMARY_RESPONSE_SCHEMA,
 )
-from meho_backplane.operations.composite_backing import register_composite_backing
+from meho_backplane.operations.composite_backing import (
+    register_composite_backing,
+    registered_composite_backing,
+)
 from meho_backplane.operations.typed_register import register_composite_operation
 from meho_backplane.retrieval.embedding import EmbeddingService
 
-__all__ = ["register_github_composite_operations"]
+__all__ = [
+    "UnbackedEnabledCompositeError",
+    "register_github_composite_operations",
+]
+
+
+#: Infix marking a composite-to-composite recursion sub-op
+#: (``gh.composite.*``). Such sub-ops are registered by the lifespan
+#: registrar, not catalog-ingested, so they never need a backing entry --
+#: the assertion below skips a composite whose sub-ops are *all* of this
+#: form. Kept in sync with ``composite_backing._COMPOSITE_OP_INFIX`` (the
+#: presence walk skips the same shape); raw-REST L2 primitives are
+#: ``METHOD:/path`` strings and never contain it.
+_COMPOSITE_OP_INFIX = ".composite."
+
+
+class UnbackedEnabledCompositeError(RuntimeError):
+    """An enabled gh-rest composite dispatches raw L2 sub-ops with no backing.
+
+    Raised at import time (connector load) when a composite in
+    :data:`_COMPOSITES` dispatches into raw-REST L2 sub-ops but has no
+    :func:`~meho_backplane.operations.composite_backing.register_composite_backing`
+    entry, or its registered ``sub_op_ids`` drift from the spec's.
+
+    Every composite in :data:`_COMPOSITES` is an *enabled*
+    ``endpoint_descriptor`` row (``is_enabled=True``), so it surfaces on
+    the op listing as a callable hit. When its raw L2 sub-ops are not yet
+    ingested, the only thing that keeps the listing honest -- flagging the
+    hit ``unbacked`` with the catalog-ingest ``next_step`` instead of
+    advertising a working read it can't perform -- is the backing registry
+    entry the listing consults. A composite that reaches the descriptor
+    upsert without a matching backing would regress silently to the
+    pre-#1757 "enabled-but-``composite_l2_missing``" dead-end (#2050). This
+    guard fails loudly at connector load so that regression can't ship.
+    """
+
+
+def _composite_dispatches_raw_l2(spec: _CompositeSpec) -> bool:
+    """Return ``True`` when *spec* dispatches into at least one raw L2 sub-op.
+
+    A composite whose ``sub_op_ids`` are all composite-to-composite
+    recursion (``gh.composite.*``) needs no backing: its sub-ops are
+    guaranteed by the lifespan registrar, never catalog-ingested, so the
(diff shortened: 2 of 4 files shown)


damir-topic:
@coderabbitai review

PR #2188, opened 2026-07-06
fix(ui): revalidate BFF read-session tokens on a drift-gated cadence

damir-topic:
## Summary

- Hardens the operator console's `/ui/*` read path: the session middleware now re-presents the session's stored access token to the JWKS-cached JWT chain on a drift-gated cadence, instead of authenticating reads off the decrypted `web_session` row alone for the session's full absolute lifetime. The accepted staleness window for read renders drops from the absolute session lifetime (default 12 h) to roughly the access-token TTL plus the new threshold (~10 min at defaults).
- New module `meho_backplane.ui.auth.revalidation`: `revalidate_read_session` runs after every successful row load in `UISessionMiddleware`. Past `ui_session_read_revalidation_seconds` (new setting, default 300; `0` = strict revalidate-every-request) without a validation, the stored token goes through the existing `verify_access_token_with_refresh` seam — an in-memory signature check on a JWKS cache hit, escalating to one reactive refresh round-trip only once the token is past `exp`. Terminal failure collapses to the same redirect-to-login as a missing session. Long-lived SSE streams get the same check on (re)connect via the shared middleware path.
- Drift anchor is a process-local last-validated map (monotonic clock, bounded with fail-safe pruning) rather than `last_seen_at`: active sessions bump `last_seen_at` on every load, so a request-gap gate would never fire for continuously active sessions — exactly the sessions the bound must cover. Anchor loss (restart, prune) only ever causes an extra validation, never a skipped one.
- Docs: lag-window rationale on the module + settings docstrings and `docs/codebase/ui.md`; CHANGELOG entry under `[Unreleased]` → `### Security`.

## Test plan

- [x] `mypy src/meho_backplane/` — clean (565 files)
- [x] `uv run pytest backend/tests/test_ui_auth_flow.py` — 42 passed (5 new revalidation tests)
- [x] A stored token that fails the JWT chain → `require_ui_session`-gated GET returns 302 to `/ui/auth/login`, not a 200 render (`test_read_path_redirects_when_stored_token_fails_validation`)
- [x] IdP no longer honours the grant (`invalid_grant` on refresh once the token passes `exp`) → 302 to login; failure event logged with no token material (`test_read_path_revoked_grant_bounces_to_login_after_token_expiry`)
- [x] Merely-expired token rotates silently through the reactive leg and the read serves 200 (`test_read_path_expired_token_refreshes_silently_and_serves`)
- [x] On a JWKS cache hit the read path makes zero outbound IdP calls per request — one JWKS warm-up fetch across two strict-mode requests, token endpoint never contacted (`test_read_path_revalidation_hits_cached_jwks_no_outbound_calls`)
- [x] Inside the drift window the hot path stays token-free (`test_read_path_within_drift_window_skips_revalidation`)
- [x] Threshold configurable via `UI_SESSION_READ_REVALIDATION_SECONDS`; lag window documented in `revalidation.py`, `settings.py`, `docs/codebase/ui.md`
- [x] Full `/ui` test sweep (`pytest -k ui`) — 1593 passed, 12 skipped
- [x] `cli make snapshot-openapi` — no drift (no route/schema change)

## Out of scope

- The write/admin path (`require_ui_admin`) and the refresh machinery in `refresh.py` — already revalidate; unchanged.
- The sliding-extension mechanism and the 12 h absolute lifetime default — unchanged.
- SSE mid-stream session checks beyond the shared (re)connect path.

Closes evoila-bosnia/meho-internal#149


damir-topic:
@coderabbitai review

damir-topic (commented):




**Anchor task:** evoila-bosnia/meho-internal#149 — bound BFF read-session revocation lag via drift-gated token revalidation
**Surface area:** backend
**Best-practices packs audited:** python (packs not present in checkout — standard MEHO security-hardening conventions applied)
**Verdict:** APPROVE
**Head SHA reviewed:** `8ad503d6f414b5c3a7fb17d1551e1e0d7939ed67`
**AI reviewer (CodeRabbit):** not received — rate-limited ("next review available in 42 minutes" > 15-min wait cap); reviewed without bot input

### Acceptance criteria
- [x] Failing stored token → login redirect, not a 200 render — `test_read_path_redirects_when_stored_token_fails_validation` (302 → `/ui/auth/login?return_to=`) and `test_read_path_revoked_grant_bounces_to_login_after_token_expiry` (IdP `invalid_grant` on the reactive leg → 302; failure event logged token-free). Both executed, green.
- [x] Revalidation call reachable from the read path, drift-gated, configurable — `UISessionMiddleware.__call__` invokes `revalidate_read_session` after every successful row load; the gate keys on `ui_session_read_revalidation_seconds`. Note on the letter of the criterion: the literal `git grep` against `middleware.py` matches only the pre-existing write-path (`require_ui_admin`) lines — the read-path call to `verify_access_token_with_refresh` lives in the new `revalidation.py`, imported and invoked from the middleware load path. Substance delivered; grep target is one file over.
- [x] Lag window documented + threshold in settings — `revalidation.py` "Accepted lag window" docstring (token TTL + threshold ≈ 10 min at defaults), `settings.py` field docstring, `docs/codebase/ui.md`; `UI_SESSION_READ_REVALIDATION_SECONDS` (default 300, `ge=0`, `0` = strict per-request mode).
- [x] Zero extra outbound IdP calls per request on a JWKS cache hit — `test_read_path_revalidation_hits_cached_jwks_no_outbound_calls` pins 1 JWKS warm-up fetch across two strict-mode requests, token endpoint call_count == 0. Executed, green.
- [x] `uv run pytest backend/tests/test_ui_auth_flow.py` — 42 passed (re-run in review worktree).

### Documented deviation — drift anchor (accepted)
The issue's option (a) anchors the gate on `last_seen_at`; the PR anchors on a process-local last-validated map (monotonic clock, bounded at 4096 entries with stale-first pruning). Verified the rationale against the code: `load_session` bumps `last_seen_at` on **every** successful load, so a gap-based gate would never fire for a continuously active session — the exact session the bound must cover. The deviation strengthens the AC's substance and fails safe in every loss mode (restart/prune ⇒ one extra validation, never a skipped one). `time.monotonic()` is the right clock here — "not affected by system clock updates" (docs.python.org/3/library/time.html#time.monotonic). Accepted; not a blocker.

### Project-board alignment
|---|---|---|---|---|---|
| evoila-bosnia/meho-internal#149 | Open | Closes (cross-repo) | ✅ | ✅ full | close on merge |

Out-of-scope boundaries honored: `require_ui_admin`/`refresh.py` untouched, sliding-extension + 12 h absolute lifetime unchanged, SSE covered only at (re)connect via the shared middleware path.

### Predicted vs. actual diff
- ✅ `backend/src/meho_backplane/ui/auth/middleware.py` — one guarded call after row load, placed **after** the `sessionmaker()` transaction block closes, so the reactive refresh's `SELECT ... FOR UPDATE` never contends with a lock held by the calling request (checked against the deadlock shape explicitly).
- ✅ `backend/src/meho_backplane/settings.py` — new field + env wiring + docstring.
- ❓ `backend/src/meho_backplane/ui/auth/revalidation.py` — new module rather than inline middleware edit; justified (keeps the C901-13 handler from growing, isolates the anchor map + test reset seam).
- ✅ `backend/src/meho_backplane/ui/auth/__init__.py`, `backend/tests/test_ui_auth_flow.py`, `docs/codebase/ui.md`, `CHANGELOG.md` (distinct `### Security` header).
- 🕳️ none — no undisclosed changes, no missing surfaces.

### Integration verification
- [x] `mypy backend/src/meho_backplane/ --ignore-missing-imports` — 2 pre-existing errors in files this PR does not touch (`redaction/presidio.py`, `broadcast/probe.py`); zero in changed files
- [x] `pytest backend/tests/test_ui_auth_flow.py` — 42 passed
- [x] `pytest -k ui -n 4` — 1593 passed, 12 skipped (the middleware sits on every `/ui/*` request; full sweep re-run in the review worktree)
- [x] ⚠️ reasoned-only: OpenAPI snapshot — no route/schema surface touched (middleware + settings + new non-route module), so no snapshot drift possible
- n/a migrations (no DB change), docker compose

### Adversarial walk (key paths checked, no blockers)
- `verify_access_token_with_refresh(decrypted, *, expected_audience)` signature/return contract matches the call site; rotated session propagates into `UISessionContext` (id/sub/tenant_id invariant under rotation).
- Catch-all `HTTPException` → `None` → login redirect deliberately widens `refresh.py`'s propagate-non-expiry-failures stance for the read path; fail-closed is the right direction here and observability is retained via the token-free `ui_read_session_revalidation_failed` warning (asserted in tests).
- Failure pops the anchor, so a broken session re-validates (and re-fails) on every subsequent request — no stale-anchor pass-through.
- Concurrent revalidations of one session funnel into the existing `SELECT ... FOR UPDATE` refresh chokepoint (first wins, loser reuses the rotated pair).
- `DecryptedSession` is a frozen dataclass — safe to hand across the closed DB session.
- Overflow prune: stale-first, then full clear; both directions only add validations.

None posted — no finding on this diff carries a required code change.

- Once the stored token is past `exp`, an unreachable IdP now converts to a login redirect on the next drift-gated read (refresh `network_error` → terminal 401). That is the fail-closed semantics the task asked for, and reads inside token validity are unaffected — but ops-facing docs for long-lived wall displays could mention that read surfaces now share the token-presenting paths' IdP-availability dependency once tokens age past `exp`.
- After a process restart, first sight of a session seeds the anchor and defers validation one interval (documented). If a future cycle wants restart-tight behavior, validating on first sight only when the row is older than the threshold would close it without a post-restart validation burst.

- Adopted: 0 / Disputed: 0 / Out of scope: 0 — no CodeRabbit review landed this iteration (org rate limit; next window exceeded the 15-min wait cap).

### Docs
- [x] `docs/codebase/ui.md` — middleware + refresh rows updated and accurate against the diff; sliding-session section verified still correct.

- The anchor-choice rationale (why not `last_seen_at`) is documented at every layer an operator or future implementer would look — module docstring, settings docstring, ui.md — and the trade is stated as an explicit lag formula rather than hand-waving.
- The five new tests pin the contract from both directions: fail-closed (invalid/revoked), fail-open-correctly (silent rotation), and the cost model (zero outbound calls on cache hit, token-free hot path inside the window).

PR #2190, opened 2026-07-06
fix(redaction): widen Tier-1 secret-value capture and redact error-builder free text

damir-topic:
## Summary

- **Tier-1 patterns capture labelled secret values to their natural delimiter.** The `authorization_header`, `bearer_token`, and `api_key` named patterns previously constrained the secret-*value* run to a restrictive character class (alphanumerics plus `. _ - + / =`), which bounded the match at the first byte outside that class. The value run is now a single negated class (`[^\s'"]`, shared `_VALUE_CHARS` constant, bounded by `_VALUE_MAX=4096`), so a value is consumed whole up to whitespace, a closing quote, or end of blob. Linear-scan matching is preserved (one bounded class quantifier, no nested quantifiers); the labelled-context requirement — not the value alphabet — continues to keep opaque IDs (Git SHAs, build hashes, `version 1.2.3`) from matching, and all existing negative-control fixtures pass unchanged.
- **Dispatch error builders route free-text diagnostics through the Tier-1 redactor.** `operations/_errors.py` now funnels every free-text diagnostic it emits from a `str(exc)` or upstream response body — `extras.exception_message` (`result_connector_error`, `result_connector_tls_verify_failed`, `result_no_connector`, `result_ambiguous_connector`), `extras.upstream_message` (403/422/auth-failed via the shared `_http_upstream_message` extractor, which also feeds the `error` summary tails), and `extras.detail` (`result_connector_unsupported`) — through a new `_sanitize_free_text` helper: a Tier-1 pass with the packaged default policy (credential-shaped patterns only, so hosts/statuses stay legible for diagnosis), followed by the existing `_EXC_MESSAGE_CAP=256` truncation. Redaction deliberately runs **before** the cap so truncation can never split a value into a fragment the patterns no longer match. The helper pins `get_default_policy()` (not the resolver override table), so a registered shadow-mode/narrowed policy cannot disable it, and it fails closed — if redaction itself is unavailable, the text is withheld rather than passed through.
- Docs updated (`docs/codebase/redaction.md`, `docs/codebase/error-message-shape.md`) and a `### Security` CHANGELOG entry added.

## Test plan

- [x] `uv run ruff check src/ tests/` — clean; `uv run ruff format --check src/ tests/` — clean
- [x] `uv run mypy src/` — clean (564 files)
[...]

Issue #101, opened 2026-05-09
Task: Write Redis subchart provenance ADR (Bitnami legacy status + alternatives + Valkey)


damir-topic:
Parent initiative: #99

## Target

New ADR file at `evoila-bosnia/meho-internal/.planning/decisions/0005-redis-subchart-provenance.md` (or next-available number — verify at filing time).

- File: `evoila-bosnia/meho-internal/.planning/decisions/0005-redis-subchart-provenance.md`
- No source-code changes in `evoila/meho` for this Task — ADR only; the actual chart work lands in G2.5

## Context

Goal #98's Helm chart Initiative (G2.5) requires a Redis subchart for v0.1. The activity-broadcast pillar (Redis Streams) is provisioned in v0.1 even though full implementation lands in weeks 5–6, so the chart needs Redis available from day 1.

Bitnami's Helm chart suite was the de-facto standard for stateful subcharts. Through 2024–2025 Bitnami progressively moved the original community charts to "Bitnami Legacy" (limited maintenance) and migrated active maintenance to "Bitnami Premium / Bitnami Secure Images" behind a paid offering. The exact current state must be verified at filing time before locking a chart provenance for an OSS project that needs to stay free to consume.

## Current state

No Redis subchart decision exists. G2.5 chart Tasks cannot be drafted concretely until provenance is locked.

## Desired state

A new ADR file using the same MADR-style structure as `0001` and `0003`, recording:

- **Phase 3 research findings** — current Bitnami chart distribution channels (Legacy vs Premium), license of the Legacy charts (still Apache 2.0?), Redis trademark/license situation post-Redis-7.4 (SSPL/RSAL change in 2024), and whether bundling Redis itself in v0.1 has any OSS-friction risk
- **Decision** — chosen approach + rationale
- **Consequences** — what changes downstream in G2.5 chart Tasks (specifically: which subchart values shape, what HA/scaling story v0.1 supports)

**Decision options to evaluate**:

- (a) **Bitnami Legacy** redis chart pinned to a known-permissive version
- (b) **Bitnami Premium / Secure Images** (paid) — almost certainly rejected for OSS posture
- (c) **ot-container-kit/redis-operator** (operator pattern; heavier than v0.1 needs)
- (d) **Custom minimal in-house chart** on top of upstream `redis:7-alpine` OSS image — Deployment + Service + ConfigMap, no PVC for v0.1 since broadcast streams are ephemeral
[...]

PR #2193, opened 2026-07-06
fix(broadcast): Tier-1 defence-in-depth on dispatch broadcasts + classifier-coverage lint

damir-topic:
## Summary

- Hardens the dispatch broadcast feed (`publish_broadcast` in
  `operations/_audit.py`) with a Tier-1 defence-in-depth layer behind
  `classify_op`: every request-params dict now runs through the new
  `scrub_broadcast_params` (`broadcast/events.py`) — a secret-shaped
  key-name scrub (the Tier-1 engine cannot see `{"password": ...}`
  because a dict leaf carries no label) plus a Tier-1 named-pattern
  pass with the packaged default policy (`Bearer ...` / `api_key=...` /
  `Authorization:` shapes inside string values) — **before** the
  payload is built.
- Any secret detection collapses the broadcast to **aggregate-only**
  regardless of `op_class`, so an op missing from the `credential_*`
  allowlists no longer ships secret-bearing params to co-tenant feed
  subscribers; a synthetic unknown `*.write` op with secret-shaped
  params broadcasts no `params` key at all. Fail-closed: a scrub error
  yields aggregate-only, never passthrough. Params with no detected
  secret material keep decision #3's full-detail default, so vetted
  benign write broadcasts (incl. Vault AppRole config writes whose
  `bind_secret_id` / `secret_id_ttl` scalars are configuration, not
  secrets) are unchanged.
- Adds a classifier-coverage lint
  (`tests/test_broadcast_classifier_coverage.py`): it enumerates every
  op the in-tree connectors register (typed + composite, ~150 ops, via
  capture shims over the registrar queue — no DB, no embedding model),
  walks each `parameter_schema` for secret-shaped property names, and
  fails CI when such an op still classifies to a full-detail class
  instead of being pinned to `credential_*`. It bites on a
  deliberately-unpinned fixture op and passes on the current registered
  set; allowlist drift now breaks the build.
- Documents the new floor in `docs/codebase/broadcast.md` and appends a
  `### Security` CHANGELOG entry.

## Design notes

- The runtime gate is *detection-based* rather than a second static
  list: "unrecognised" secret-bearing ops are exactly the ones no
  static allowlist can anticipate, so the collapse keys on detected
  secret material (key names + Tier-1 patterns), and the static sweep
  is the CI-time companion that forces registered ops to get pinned.
  A secret-free unknown write op keeps full detail per decision #3 —
  the feed's mutation signal is not degraded for benign ops.
- The policy is pinned via `get_default_policy()` (not the resolver
  ladder) so an operator-registered shadow-mode or narrowed override
  cannot silently disable the broadcast floor — same posture as the
  error-path sanitizer hardening.
- The MCP publish path (`mcp/handlers.py` via
  `compute_effective_broadcast_detail`) is deliberately out of scope
  per the task; the per-tenant override resolver semantics are
  untouched.

## Test plan

- [x] `uv run mypy src/` — clean (564 files)
- [x] `uv run pytest -n 4 --dist loadscope --ignore=tests/integration tests/` — full unit suite green
- [x] New unit tests: unpinned `write`/`other`/`read` ops with secret-shaped params (`password` key, `Bearer <token>` string, `api_key=...` value) broadcast aggregate-only with by-exclusion whole-event leak assertions; benign write/read ops keep full params; fail-closed scrub test
- [x] Classifier-coverage sweep: ≥100 registered ops enumerated (vacuous-pass guard), zero violations on the current set, fails on a deliberately-unpinned fixture op, boolean/integer config attrs exempt
- [x] No FastAPI route/response-schema change — OpenAPI snapshot unaffected (no `cli` regen needed)

## Out of scope

- Tier-2 Presidio NER on the broadcast path.
- MCP/agent-event broadcast surfaces beyond the dispatch
  `publish_broadcast` path.
- The connector-boundary / audit-row redaction middleware (already
  Tier-1) and the per-tenant broadcast override resolver semantics.

Closes evoila-bosnia/meho-internal#151


damir-topic:
@coderabbitai review

damir-topic (commented):




**Anchor task:** evoila-bosnia/meho-internal#151 — Tier-1 defence-in-depth on dispatch broadcasts + classifier-coverage lint
**Surface area:** backend
**Best-practices packs audited:** python (packs absent from checkout — MEHO security-hardening house conventions applied: fail-closed, defence-in-depth, no tunables)
**Verdict:** APPROVE
**Head SHA reviewed:** `f27c6d3a202162d51b3581231e8736264a7bfd54`
**AI reviewer (CodeRabbit):** not received — rate-limited for this run; the re-trigger returned a contentless "Review finished" no-op (0 reviews, 0 inline comments)

### Acceptance criteria
- [x] Unit test: `write`/`other`-class op with secret-shaped params (`password` key, `Bearer` string, `api_key=` value) broadcasts no cleartext — verified by running `tests/test_broadcast_dispatch_tier1.py` in the review worktree (41 passed); assertions are by-exclusion over the whole serialised event, so a regression that relocated the value would still fail.
- [x] Broadcast params pass through the Tier-1 `redact()` engine before the payload is built — verified by the AC's own `git grep` plus code walk: `scrub_broadcast_params` (events.py) runs the key-name scrub then `redact(..., get_default_policy())`, and `publish_broadcast` (_audit.py) calls it before `redact_payload`.
- [x] Classifier-coverage test enumerates the registered set, bites on a deliberately-unpinned fixture op, passes on the current set — verified by running it (149 ops enumerated via registrar capture shims, ≥100 vacuous-pass guard, fixture `acme.credstore.create` flagged, zero violations today).
- [x] Unrecognised write op with secret-shaped params → aggregate-only, synthetic unknown `*.write` op broadcasts no `params` key — verified by test + code walk. **Interpretation reviewed and accepted** (see design-note assessment below): the collapse is detection-based rather than a second static recognised-op list.
- [x] Suite green — in the review worktree: `ruff check` clean, `ruff format --check` clean, `mypy src/` clean (564 files), the two new modules + 10 adjacent broadcast/redaction suites pass (231 tests); PR CI integration lane green. No route/schema change, so the OpenAPI snapshot is legitimately untouched.

### Design-note assessment (AC-4 detection-based interpretation)
The task's "unrecognised write op → aggregate-only" is implemented as: any detected secret material (secret-shaped key name anywhere in the params tree, or a Tier-1 named-pattern hit inside a string value) collapses the broadcast to aggregate-only for **every** op class, fail-closed on scrub errors; the static classifier-coverage lint is the CI companion that forces registered ops declaring secret-shaped schema properties to be pinned. I probed this rather than taking the design note at face value:

- A literal "all unrecognised write ops → aggregate" would require a second static list of recognised-for-full-detail write ops that the task never defines; absent it, the entire `write` class would collapse, degrading the feed's mutation signal that decision #3 deliberately keeps — a change the task does not ask for.
- The task's own threat definition is "secret-shaped params". Every shape it names is closed: nested secret-named keys (the recursive scrub catches `{"data": {"password": ...}}` — verified live), labelled values (`Bearer ...`, `api_key=...`, `Authorization:` header dumps, `Password=` inside connection strings — verified live), and registered-op allowlist drift (the lint).
- Exemption tuning is sound, verified against leak direction: `token_policies`/`token_ttl` are policy-name references (non-final `token`), `secret_name`/`secret_path` are references; `bind_secret_id`/`secret_id_ttl` are exempt only as **bool/int scalars** — a string-typed `secret_id_ttl` actually trips the (`secret`,`id`) pair rule and collapses. The failure direction of every exemption is over-collapse (signal loss), never passthrough of string secret material.
- The pinned `get_default_policy()` excludes the identifier patterns (UUID/IP/FQDN), so benign params carrying hosts/IDs do not false-collapse — full-detail write broadcasts survive, confirmed by the benign-path tests.
- Shadow-mode interplay is safe: detection (`result.manifest`) fires identically in shadow mode, so even a hypothetical shadow default would still collapse rather than pass through.

Residual (accepted, recorded): an opaque unlabelled string value under a generically-named container key is invisible to any deterministic Tier-1 layer — the test module's docstring states this honestly, and Tier-2 NER is explicitly out of the task's scope. Follow-ups below.

### Project-board alignment
|---|---|---|---|---|---|
| evoila-bosnia/meho-internal#151 | Open (private tracker, Goal #87 / Initiative #101) | Closes | ✅ (cross-repo) | ✅ full | close on merge |

Sibling PR #2190 (approved, unmerged): no source-file overlap, verified from both diffs (#2190: `_errors.py` + `patterns.py`; this PR: `events.py` + `_audit.py`). Only collision is both inserting a `### Security` CHANGELOG block at the same anchor — a trivial union-merge for whichever lands second. Fixture invariance under #2190's value-capture widening holds: widening only redacts more, and this PR's benign fixtures contain no labelled shapes.

### Predicted vs. actual diff
- ✅ `backend/src/meho_backplane/broadcast/events.py` — scrub + predicate, reuses the existing `redact_payload(detail=...)` extension point (G6.3-T2) instead of re-plumbing
- ✅ `backend/src/meho_backplane/operations/_audit.py` — collapse decision at the single publish site, structlog house pattern
- ✅ `backend/tests/test_broadcast_dispatch_tier1.py`, `backend/tests/test_broadcast_classifier_coverage.py`
- ✅ `docs/codebase/broadcast.md`, `CHANGELOG.md` (`### Security` header per convention)
- Nothing missing, nothing undisclosed. The MCP publish path staying classification-only is disclosed in the PR and matches the task's out-of-scope list.

### Integration verification
- [x] `ruff check src/ tests/` + `ruff format --check` — clean (worktree)
- [x] `mypy src/` — clean, 564 files (worktree)
- [x] `pytest` new modules — 41 passed (worktree)
- [x] `pytest` adjacent broadcast/redaction/dispatch suites — 190 passed, 6 skipped (worktree)
- [x] Live probes of the scrub/predicate corners via `uv run python` (worktree)
- [x] CI: integration testcontainers lane, Semgrep SAST/OSS, TruffleHog, CodeQL, DCO — green on the PR
- ⚠️ not run: migrations replay (no migration), docker compose (no runtime wiring change), full 9k-test suite locally (author-run + covered by the PR's Python CI lane)

None — no inline comments posted.

Recorded for future tickets; none block merge or require a response here.

- **URL-userinfo credentials as a Tier-1 named pattern.** A DSN-style value (`scheme://user:pass@host/...`) under a non-secret-shaped key is not caught by the current pattern catalogue (labelled `Password=`/`token=` shapes are). A `url_userinfo` pattern in `redaction/patterns.py` would harden the broadcast floor, the #2190 error path, and the connector boundary in one move.
- **Key-name pair list could grow a few entries** (`signing`/`master`/`tls`/`ssl`/`cert` + `key`) — no currently registered op declares these (grep-verified), so this is future-proofing; pairs avoid the `ssh_public_key`/`label_key` over-match that motivated excluding final `key`.
- **Ingested-op coverage.** Ops registered from ingested OpenAPI specs (`POST:`-prefixed op_ids) are outside the typed-registrar sweep, so the lint cannot force pins for them; the runtime layer is their only cover. A follow-up could sweep the MEHO-shipped package-data specs' schemas, or default ingested mutation-method broadcasts to aggregate-only.
- A cross-link from `docs/codebase/redaction.md`'s consumer overview to the new broadcast floor section would help discoverability (deferred here since #2190 is concurrently editing that file).

- Adopted: 0, Disputed: 0, Out of scope: 0 — no review content landed (rate-limited for the whole run).

### Docs
- [x] `docs/codebase/broadcast.md` — updated, accurate against the diff (including the honest statement of the MCP path's posture)
- [x] `docs/codebase/redaction.md` — verified not invalidated (its "broadcast path unchanged" line is a historical shadow-mode task-scope note, still true)

- The leak assertions are by-exclusion over the whole serialised event, and the lint carries a vacuous-pass guard (≥100 ops) plus a bites-on-purpose fixture test — both are the honest-test postures that make green meaningful.
- Pinning `get_default_policy()` past the resolver override table so no registered policy can lower the broadcast floor mirrors the error-path sanitizer posture — consistent hardening architecture across both surfaces.

PR #2201, opened 2026-07-07
fix(cli): reject named ports in --resolve entries and correct splitResolveEntry docs

damir-topic:
Follow-up to the review of PR #2181 (`meho login --resolve` split-DNS pin, #2107). The auto-review recorded two non-blocking Major findings in `cli/internal/auth/resolver.go` that were deferred past the merge; this PR lands both. Refs #2107.

## M2 — named port passes validation but the pin silently never fires

Port validation used `net.LookupPort("tcp", port)`, which happily accepts named services such as `https`. A `--resolve kc.example.com:https:10.0.0.5` entry therefore *passed* validation but keyed the override map as `kc.example.com:https` — and the transport's dial address always carries a **numeric** port, so the key never matched and the pin was silently accepted yet never applied. That contradicts the module's own fail-loud contract ("a malformed entry is a hard error rather than a silent skip").

Fix: the port must now be strictly numeric via `strconv.Atoi` with a 1–65535 range check. A named port (or `0`, `65536`, `-1`) is rejected loudly at parse time with an error that echoes the offending `--resolve` entry:

```
--resolve "kc.example.com:https:10.0.0.5": port "https" is not a numeric TCP port in the range 1-65535 (named services such as "https" are not supported)
```

## M1 — `splitResolveEntry` docs claimed a back-split parse that doesn't exist

The doc comments claimed the host is "everything before the last two colon-separated fields so IPv6 literals used as the *host* are not silently mangled" — but the implementation front-splits on the first two colons (which is what makes the bare-IPv6 *IP* side work, e.g. `kc:443:::1`). An IPv6-literal host like `[::1]:443:10.0.0.5` errored with a misleading port/IP validation message.

Fix (docs-first, no parser redesign per the review guidance):

- `ParseResolveEntries` / `splitResolveEntry` comments now describe the actual front-split behaviour and state that IPv6-literal hosts are unsupported (the `--resolve` target is a DNS name by definition — an IP-literal host needs no resolution override).
- A cheap honesty check: an entry starting with `[` is rejected up front with `IPv6-literal hosts are not supported; the host must be a DNS name (only the ip field may be IPv6)` instead of the confusing downstream failure.

## Tests / gates

- `TestParseResolveEntriesRejectsMalformed` gains cases: `named-port` (`https`), `port-zero`, `port-out-of-range` (`65536`), `negative-port`, `ipv6-literal-host` — each asserting the loud error and that the bad entry is echoed.
- `TestParseResolveEntries` gains numeric boundary acceptance cases (`1`, `65535`); all pre-existing cases (incl. bare/bracketed IPv6 on the IP side) unchanged and green.
- `docs/codebase/cli.md` split-DNS section and CHANGELOG `[Unreleased]` updated to match.
- Gates: `go build ./...`, `go vet ./...`, `gofmt -l .` (clean), `go test ./...` all green in `cli/`. golangci-lint skipped locally (known go1.24 host-binary vs go1.25 target mismatch; CI runs the pinned version).

No behavioural change for well-formed entries: numeric-port pins parse and fire exactly as before; both fixes only turn silent/ misleading failure modes into loud, actionable parse errors.


damir-topic:
@coderabbitai review

PR #2213, opened 2026-07-07
fix(connectors): async ingest rejects foreign tenant_id with a synchronous 403 (no unpollable job)

### LINKED ISSUE #2208: connectors: async ingest with foreign tenant_id returns 202 + unpollable failed job instead of the sync path's synchronous 403
Parent goal: #221
Parent initiative: #2149

## What we hit

Follow-up from #2085 (PR #2204). The REST ingest route's **async path creates the job row and detaches the pipeline before authorization runs**: `backend/src/meho_backplane/api/v1/connectors_ingest.py` — `registry.create(...)` at ~line 541, then `asyncio.create_task(...)` at ~line 591, returns `202` with a job handle. The tenant-scope authorization (`IngestionPipelineService._authorize`) only runs inside the pipeline, so a caller passing a **foreign `tenant_id`** gets:

- **sync path**: synchronous 403 (correct — the service raises before any write), but
- **async path**: HTTP 202 + a job that flips to `failed`, and because the job row was minted with the *requested* foreign tenant scope, the caller's poll on `GET /api/v1/connectors/ingest/jobs/{id}` is gated cross-tenant — the job is effectively unpollable. The caller can't distinguish "authz denied" from "pipeline crashed".

## Why this is a bug, not a UX gap

The same input yields a synchronous 403 on one branch (`async: false`) and a 202-then-invisible-failure on the other (`async: true`). #2110 just landed the "structured, honest results" contract on the dispatch surface; the ingest surface should fail the same request the same way on both branches.

## Proposed fix

Hoist an eager route-level tenant-scope authorization check to before `registry.create` in the async branch (shared with the sync branch — same predicate the service layer uses, so the service-layer `_authorize` stays as defense-in-depth). Foreign `tenant_id` → synchronous 403 with the existing structured detail on **both** branches. No schema change expected (403 is already a documented response on the sync path — extend to the async branch's OpenAPI response map `_INGEST_RESPONSES`/route decorator if it isn't listed there).

## Acceptance criteria

- [ ] `POST /api/v1/connectors/ingest` with `async: true` and a foreign `tenant_id` returns a synchronous 403 (same body shape as the sync path); no job row is created.
- [ ] Own-tenant and omitted/global `tenant_id` async ingests are unaffected (202 + pollable job).
- [ ] Service-layer `_authorize` remains (defense-in-depth), covered by an existing or new test.
- [ ] Route tests cover the async-403 path; if the OpenAPI response map changes, CLI snapshot + Go client regenerated.

## References

- #2085 / PR #2204 (the tenant-scope semantics this hardens)
- Reviewer finding recorded in the #2149 auto-run (pr-2204-iter-1 review)

damir-topic:
## Summary

- **Async ingest now rejects a foreign `tenant_id` synchronously (403), identically to the sync branch** — previously the async branch minted the job row (`registry.create`) and detached the pipeline (`asyncio.create_task`) *before* the service-layer tenant-scope guard ran, so a foreign `tenant_id` got `202 Accepted` plus a `failed` job scoped to the *foreign* tenant, which the caller's poll on `GET /api/v1/connectors/ingest/jobs/{id}` could never see (the job read gate is tenant-scoped). "Authz denied" was indistinguishable from "pipeline crashed".
- The fix hoists an eager route-level check (`_authorize_write_scope_or_403`) into `ingest_endpoint`, before the sync/async branch split, backed by the same predicate the service layer uses: `IngestionPipelineService._authorize` is renamed to the public `authorize_scope` and stays wired inside `ingest()` as defence-in-depth for the CLI / MCP siblings (covered by a new direct service-level test).
- The route's OpenAPI `responses` map now declares the 403 (plain-string `detail`, the shape the sync path has always returned), so the contract states that **both** branches reject a cross-tenant write scope synchronously and no job row is created. CLI snapshot + Go client regenerated.
- Docs: `docs/codebase/spec-ingestion.md` async-mode + cross-surface write-scope sections updated; CHANGELOG bullet added under `[Unreleased]`.

## Test plan

- [x] `mypy src/` — clean (567 files)
- [x] `.claude/hooks/code-quality.py --diff` — 0 blockers (`ingest_endpoint` refactored back under the 100-line budget via `_fold_compat_band` + `_authorize_write_scope_or_403` helpers)
- [x] AC1: `POST /ingest` with `async: true` + foreign `tenant_id` → synchronous 403, exact sync-path body shape, **no job row created, no background task detached** — `test_ingest_async_cross_tenant_tenant_id_rejected_403_no_job`
- [x] AC2: own-tenant and omitted/global async ingests unaffected (202 + pollable job) — existing `test_ingest_async_job_poll_cross_tenant_is_404_not_403` (explicit own-tenant 202 + poll) and `test_ingest_async_default_returns_202_with_job_handle` (omitted/global 202) still pass
- [x] AC3: service-layer guard remains as defence-in-depth — new `test_pipeline_service_authorize_scope_stays_defence_in_depth` drives `IngestionPipelineService.ingest` directly (no route gate) and proves `PermissionError` before any spec fetch/DB write
- [x] AC4: route tests cover the async-403 path; OpenAPI response map changed → `cd cli && make snapshot-openapi && make generate` run; `go build ./... && go vet ./... && go test ./...` clean
- [x] Full backend unit suite (`pytest -n 4 --dist loadscope --ignore=tests/integration tests/`) — green
- [x] Integration doubles checked: the only ingest fixtures under `tests/integration/` either call `_validate_spec_versions` directly or use the global write scope — unaffected by the eager gate

## Out of scope

- **The MCP async sibling has the same job-row-before-authz gap** (`mcp/tools/connector_ingest.py::_spawn_async_ingest` mints the job before `service.ingest`'s `authorize_scope` runs). #2208 scopes the fix to the REST route; the MCP surface deserves its own task (adjacent finding recorded in the run's result file).

Closes #2208


damir-topic:
@coderabbitai review

PR #2220, opened 2026-07-07
fix(runbooks): surface per-step failed state on the run list read surface

### LINKED ISSUE #2119: runbooks: after a `"no"`/`"escalate"` verify-answer, the per-step `failed` state is persisted internally but invisible from `GET /api/v1/runbooks/runs` — the run's read surface shows `state: in_progress, current_step_id: <step>` with no step-state field, so operators/agents can only discover the failure by making another mutation
## Summary

After a `runbook_next` call with `verify_response.answer = "no"` (or `"escalate"`), the per-step state internally transitions to `"failed"` — the schema's `ConfirmVerifyResponse` docstring says exactly this, and the engine honors it. A subsequent `runbook_next` call surfaces the state via error: HTTP 400 *"previous step '<id>' is in 'failed' state; abort the run"*.

But the run's read endpoints (`GET /api/v1/runbooks/runs`, filtered to the specific `run_id`) still report:

```json
{
  "run_id": "...",
  "state": "in_progress",
  "current_step_id": "<step-that-just-failed>",
  "position": {"n": <N>, "total": <T>},
  ...
}
```

No step-state field appears anywhere in the response. To discover that the step is in `failed` and that the only path forward is `abort`, the caller has to make another mutation attempt and read the error message from the 400.

Observed live on v0.19.0 (chart `0.19.0`).

## Why this matters

- **Observability gap for the run's actual state.** Run-level `state: in_progress` + step-level `state: failed` are an internal divergence — one is exposed, the other isn't. A monitoring system polling `runbooks/runs` to surface runs needing operator attention will see all in-flight runs the same way, even if some are stuck on a failed step that only `abort` can clear.

- **Agent automation has to discover state via mutation.** An agent that just answered `"no"` wants to know "do I need to abort this run, or can I retry the step?" — but the only way to find out is to make another `next` call and parse the resulting error message. There is no `GET /runs/<id>` (or `/runs/<id>/step`) that returns the current step's state.

- **Diverges from the schema docstring's promise.** The `ConfirmVerifyResponse` docstring says:

  > *"Only :attr:`answer` `= "yes"` advances the run. `"no"` transitions the step to `failed`; `"escalate"` transitions to `failed` with `verify_response.escalated=true` semantics so the senior's review path is explicit in the persisted state (per Initiative #1198)."*

  The persisted state IS reached (confirmed via the subsequent mutation's error), but it isn't readable through any documented endpoint. The docstring's promise ("explicit in the persisted state") is honored internally but breaks at the read surface.

## Reproduction

```bash
JWT=...  # tenant_admin JWT
B=https://<backplane>

# 1. Start a run
SLUG="<any-published-template>"
run_resp=$(curl -ksSf -H "Authorization: Bearer $JWT" -H 'Content-Type: application/json' \
  -X POST "$B/api/v1/runbooks/runs" \
  -d "{\"template_slug\":\"$SLUG\",\"target\":\"<any-registered-target>\"}")
RUN=$(echo "$run_resp" | jq -r '.run_id')
# → position: {n:1, total:N}, current_step at step-one

# 2. Advance with answer="no"
curl -ksS -H "Authorization: Bearer $JWT" -H 'Content-Type: application/json' \
  -X POST "$B/api/v1/runbooks/runs/$RUN/next" \
  -d '{"last_verified": false, "verify_response": {"type": "confirm", "answer": "no"}}'
# → HTTP 400 detail="step 'step-one' verify did not pass; abort the run"
# (per docstring: step-one should transition to "failed")

# 3. Read the run's state — is the failure surfaced?
curl -ksS -H "Authorization: Bearer $JWT" "$B/api/v1/runbooks/runs" \
  | jq --arg r "$RUN" '.runs[] | select(.run_id == $r)'
# → {state: "in_progress", current_step_id: "step-one", position: {n:1, total:N}, ...}
# NO indication step-one is "failed". Run-level state unchanged.
# No step-state field anywhere in the response.

# 4. Any subsequent next call surfaces the state as an error
curl -ksS -H "Authorization: Bearer $JWT" -H 'Content-Type: application/json' \
  -X POST "$B/api/v1/runbooks/runs/$RUN/next" \
  -d '{"last_verified": false, "verify_response": {"type": "confirm", "answer": "escalate"}}'
# → HTTP 400 detail="previous step 'step-one' is in 'failed' state; abort the run"
# Confirms step-one IS internally in "failed" state — just invisible from read endpoints.
```

## Probable cause (consumer-side guess; haven't traced source)

The runbook step-state is presumably stored on a step-instance row (or the run row's `steps` JSONB column) but isn't projected through the run-list / run-show serializer. Either:

- The serializer was written before step-level state was added, and never updated to include it.
- The intent was that step-state is an internal detail and only the run-level state matters operationally — in which case the docstring is misleading (it promises the state is "explicit in the persisted state" but the persisted state isn't accessible from outside).

- **Option A — expose step-level state in the run-show endpoint.** Add a `current_step_state` field (e.g. `"running" | "verifying" | "failed"`) on `GET /api/v1/runbooks/runs` and any filtered variant. A reader can then poll and discover failures without making mutations. Minimal delta on the wire.

- **Option B — surface a separate `GET /runs/<id>/step` endpoint.** Returns the current step's full state including verify history. Doesn't bloat the list view but gives a clean read path for "what's happening on this run right now."

- **Option C — transition the run-level state too** (e.g. `in_progress → "failed"` or `"paused"`). Operator/agent reads the run's top-level state alone and knows what to do. Simplest read model, and matches the docstring's "explicit in the persisted state" phrasing best.

Option C is the cleanest if the design intent is "a failed step blocks the run until abort" — having the run say `in_progress` when no forward progress can happen is just a lie. Option A is the minimal change. Option B is the richer read surface.

## Verification

After the fix:

1. Start a run against any published template.
2. Advance with `verify_response.answer = "no"` on step 1 → HTTP 400.
3. `GET /api/v1/runbooks/runs?run_id=<...>` — expect one of:
   - Option A: `current_step_state: "failed"` (or equivalent) in the response body.
   - Option B: a separate step endpoint returns `state: "failed"`.
   - Option C: run's top-level `state` != `in_progress` (e.g. `failed`, `paused`, `needs_operator_attention`).

Any of the three is sufficient to close this issue.

## Out of scope

- Whether the docstring's "step transitions to failed" is itself the right design (retry semantics vs. abort-only is a separate design question).
- The `"escalate"` answer's `escalated=true` semantic (also invisible from the run list; folds into the same fix — if step-state is exposed, escalation state should be too).
- The step-failed audit trail (likely persisted via the audit ledger but not surveyed here).

## Related

- Sibling runbooks-family issues from the same drive week: `#2111` (`show_template` opacity_floor violation), `#2114` (cross-browser non-deterministic banner), `#2115` (UI edit 403+mutation), `#2116` (runbook edits missing from audit ledger), `#2117` (5 bundled UI authoring defects).

Parent goal: #221

## Acceptance criteria

- [ ] After a `runbook_next` call with `verify_response.answer = "no"` (or `"escalate"`) on a step, `GET /api/v1/runbooks/runs` (filtered to that `run_id`) surfaces the per-step `failed` state WITHOUT any further mutation — via one of: a step-state field on the run projection (e.g. `current_step_state: "failed"`), a dedicated read endpoint, or a run-level state that is no longer `in_progress` (design choice open, see below).
- [ ] A new test starts a run, advances step 1 with `answer = "no"`, then reads the run through the chosen read surface and asserts the failed state is readable (no `next`/mutation call in between).
- [ ] The chosen read-surface field is projected in `RunSummary` (`backend/src/meho_backplane/runbooks/runs_schemas.py`) — whose docstring currently states "Run-level state only: no step contents are exposed" — and the docstring is updated to match the new contract (or the run-level `state` Literal is extended if Option C is taken).
- [ ] The `"escalate"` case is covered too: a test asserts the escalated-to-`failed` state (`verify_response.escalated=true` semantics per Initiative #1198) is also readable from the same read surface, not just `"no"`.
- [ ] The OpenAPI snapshot / generated CLI is regenerated if the read schema changes (`cd cli && make snapshot-openapi && make generate`), and the snapshot check passes in CI.

> Triaged from raw finding via /review-ticket. Runbooks read-surface gap; design-choice-open across 3 options (A: step-state field on run projection / B: separate `GET /runs/<id>/step` endpoint / C: transition run-level state) — pick the minimal (surface a step-state field on the run projection, Option A).

Parent initiative: #2151

damir-topic:
## Summary

- After a `runbook_next` with `verify_response.answer = "no"` / `"escalate"`, the step's `failed` state was persisted on `runbook_run_step_states.state` but invisible from `GET /api/v1/runbooks/runs` — operators/agents could only discover it by firing **another** `next` mutation and parsing the resulting 400. This PR surfaces the persisted per-step state on the existing run projection (Option A per the issue's triage note — the minimal shape).
- `RunSummary` gains an additive, optional `current_step_state: Literal["in_progress", "failed"] | None` field, projected in `RunbookRunService.list_runs()` from the step-state rows the projection already loads for `current_step_id`. `None` for terminal runs (no current step). No new DDL — the truth was already persisted; only the read surface changes.
- The projection flows to every consumer of the shape for free: `GET /api/v1/runbooks/runs` (keyed + `?envelope=v2`), the `meho.runbook.list_runs` MCP tool (tool description updated so agents know `failed` ⇒ abort is the only forward path), and the generated Go CLI client (`--json` mode).
- Opacity floor untouched: the step's *state* is not step *content* — no bodies, no adjacent steps, and the load-bearing `test_list_runs_omits_step_contents` regression test now also pins the new field alongside the no-leakage property. The escalate-vs-no distinction stays in the persisted `verify_response` (out of scope per the issue).

## Test plan

- [x] `ruff check src/ tests/` + `ruff format --check` — clean (1143 files)
- [x] `mypy src/` (strict) — clean, 567 files
- [x] `pytest tests/test_runbooks_run_service.py tests/test_runbook_runs_api.py tests/test_runbooks_runs_schemas.py tests/test_mcp_tools_runbook_runs.py` — 116 passed
- [x] `pytest tests/test_runbooks_priming.py tests/test_ui_runbook_runs_list.py tests/test_api_v1_list_envelope_v2.py tests/test_ui_runbook_driver_opacity.py` — 66 passed (consumer suites)
- [x] Full backend unit suite — green (see CI)
- [x] `go build ./... && go vet ./... && go test ./...` in `cli/` — clean after regen
- [x] AC1 — failed state readable without mutation: `test_list_runs_surfaces_failed_step_state_without_mutation[no]` starts a run, answers `"no"`, then reads **only** `list_runs` and asserts `current_step_state == "failed"`
- [x] AC2 — new test with no mutation between the failing `next` and the read (same test; the only calls are `start_run` → `next_step` → `list_runs`)
- [x] AC3 — field projected in `RunSummary` (`backend/src/meho_backplane/runbooks/runs_schemas.py`) and the "Run-level state only: no step contents" docstring updated to the new contract
- [x] AC4 — escalate covered: the same test parametrized with `answer="escalate"` asserts the escalated-to-`failed` state is readable from the same surface
- [x] AC5 — OpenAPI snapshot + generated CLI regenerated (`cd cli && make snapshot-openapi && make generate`); Go client builds and tests pass

## Out of scope

- Rendering a step-state column in the human `meho runbook runs` table (the field rides `--json` today; the 7-column table layout is a CLI render choice) and a failed-step chip in the `/ui` runs list (G10.x UI tasks own that surface).
- Distinguishing escalate-vs-no on the wire — the persisted `verify_response.answer` already carries it; the issue explicitly folds only the *state* into this fix.
- Runbook priming (`assemble_runbook_priming`) still renders a failed-step run with "use meho.runbook.next to advance" guidance — adjacent finding, noted in the result file.

Closes #2119


damir-topic:
@coderabbitai review

PR #2225, opened 2026-07-07
fix(runbooks): htmx-process Alpine-cloned editor steps so step 2+ previews fire

### LINKED ISSUE #2117: Initiative: runbooks UI authoring-surface cleanup — 5 create/preview/edit/deprecate defects

### LINKED ISSUE #2174: runbooks UI: per-step live preview only fires for the first step; steps added via "+ Add step" are never htmx-processed
## Summary

In the runbook authoring editor (`/ui/runbooks/new` and `/ui/runbooks/<slug>/edit`), the per-step live Markdown preview only works for the **first** step. Any step added with **+ Add step** (and, on an edit, steps beyond the first) never renders a preview, no matter what is typed in its Body.

The old #2117 "defect 1" (Preview threw `Element.closest()` on an invalid `.flex-wrap >> …` selector) is fixed — the first step's preview now works. This is a **different** root cause that affects the dynamically-added steps only.

Reproduced on `main` (post-v0.19.0) through the local operator console.

## Reproduction

1. Open `/ui/runbooks/new` as `tenant_admin`.
2. In step 1, type Markdown in **Body**. The preview pane renders. Correct.
3. Click **+ Add step**.
4. In step 2, type Markdown in **Body**.
5. Observed: no preview ever renders for step 2. No `POST /ui/runbooks/preview` is issued for it (confirmed in the server access log — only step 1 produces preview POSTs).

On the same page the browser console also logs:

```
htmx: The selector "find button[type=submit]" on hx-disabled-elt returned no matches!
```

## Observed vs expected

- Observed: the preview fires only for the first step; added steps are inert.
- Expected: each step's Body renders its own live preview, per the `_step_fields.html` design comment ("a debounced `hx-post="/ui/runbooks/preview"` that swaps the sibling `.runbook-step-preview` pane").

## Root cause

Each step's body textarea in `backend/src/meho_backplane/ui/templates/runbooks/_step_fields.html` carries the htmx preview trigger:

```html
<textarea class="… runbook-step-body"
          hx-post="/ui/runbooks/preview"
          hx-trigger="input changed delay:500ms"
          hx-target="next .runbook-step-preview" …></textarea>
```

This markup lives inside an Alpine `<template x-for="(step, idx) in steps">` loop in `editor.html`. When Alpine clones a new step into the DOM, **htmx is never told to process the new node**. The editor script (`editor.html`, `{% block scripts %}`) defines `scanBodies()`, which re-mounts CodeMirror on new `.runbook-step-body` textareas on `htmx:afterSettle` and on click — but it never calls `htmx.process(...)`. So htmx never registers the `hx-post` / `hx-trigger` on any step added after htmx's initial page scan. The first step is present at that initial scan, so only it works.

(htmx does not observe arbitrary JS-inserted DOM; nodes added by Alpine `x-for` must be handed to `htmx.process()` explicitly.)

## Suggested fix

After Alpine adds a step (`addStep()`) and in the existing rescan hooks, also register the new subtree with htmx, for example:

```js
function processHtmx(root) {
  if (window.htmx) window.htmx.process(root || document.getElementById("runbook-editor-form"));
}
```

Call it alongside `scanBodies()` after a step is added and on `htmx:afterSettle`, so each new step's `hx-post` preview trigger is registered.

## Related

- #2117 (runbooks UI authoring-surface cleanup): defect 1 there (the `>>` selector) is fixed; this is the residual per-step case for dynamically-added steps.
- The `hx-disabled-elt "find button[type=submit]" returned no matches` console warning comes from the main form's `x-bind:hx-disabled-elt`; likely benign but worth a look in the same sweep.

Parent goal: #221

## Acceptance criteria (triage 2026-07-06)

Follow-on to the closed runbooks-UI cleanup #2117 (its "defect 1" `closest()` selector is fixed; this is a distinct root cause affecting dynamically-added steps).

- [ ] Each runbook step's Body textarea fires its own debounced `hx-post="/ui/runbooks/preview"` and renders its sibling `.runbook-step-preview` — including steps added via **+ Add step**, not only the first.
- [ ] The newly-inserted step DOM is htmx-processed (`htmx.process()` on the added node, or an equivalent OOB/`hx-swap` approach) so its `hx-trigger` binds.
- [ ] The `htmx: The selector "find button[type=submit]" on hx-disabled-elt returned no matches!` console warning is resolved/scoped so it doesn't fire per added step.
- [ ] A test or manual-repro note confirms preview works for step 2+ on both `/ui/runbooks/new` and `/edit`.

## Out of scope

- The already-fixed #2117 "defect 1" (`closest()` invalid selector on the first step) — this is a distinct root cause on dynamically-added steps.
- The other #2117 D1–D5 defects (all landed) — this is a new post-v0.19.0 multi-step-preview regression.

## References

- Bug site: `backend/src/meho_backplane/ui/templates/runbooks/_step_fields.html:122-124` (per-step `hx-post="/ui/runbooks/preview"` trigger) — verified on `main` 2026-07-06; design comment `:22-23`.
- Follow-on to: #2117 (runbooks-UI cleanup, closed) / PR #2122 (D1 preview fix).
- htmx: dynamically-inserted DOM must be processed with `htmx.process(el)` for its `hx-*` to bind (htmx docs — `htmx.process`). Frontend best-practices pack absent from this checkout.

damir-topic:
## Summary

- The runbook authoring editor's per-step live Markdown preview only fired for the first step; steps added via **+ Add step** (and any step past the first on an edit) never rendered a preview and issued no `POST /ui/runbooks/preview`.
- Root cause: each step's Body textarea carries `hx-post="/ui/runbooks/preview"`, but the step markup lives inside an Alpine `<template x-for>` loop. htmx only binds `hx-*` on DOM it renders itself, so Alpine-cloned steps were never htmx-processed. The editor script re-mounted CodeMirror on new bodies but never called `htmx.process`.
- Fix: a `processSteps()` helper hands the form root to `htmx.process` (idempotent — htmx skips already-initialised nodes), folded into a shared `rescan()` that runs on first render, on every `htmx:afterSettle`, and after each add/remove click.
- Also scoped the form's inherited `hx-disabled-elt`: it bound `find button[type=submit]` and, being inherited, leaked into each descendant preview POST — the textarea resolved `find` against itself, matched nothing, and logged `The selector "find button[type=submit]" on hx-disabled-elt returned no matches!` once per preview. `hx-disinherit="hx-disabled-elt"` keeps the value on the form's own submit while stopping the per-step leak.

## Test plan

- [x] `pytest tests/test_ui_runbooks_editor.py` — 22 passed (3 new: htmx.process wiring on `/new`, `hx-disinherit` on the form, multi-step `/edit` hydration carrying the per-step trigger + processing)
- [x] `pytest tests/test_ui_runbooks_{acceptance,lifecycle,list}.py` — 48 passed (no surface regression)
- [x] `mypy` (src gate) — Success, 567 files
- [x] Acceptance: each step's Body fires its own debounced preview including added steps (htmx.process on new subtrees); the `hx-disabled-elt … no matches` warning scoped away; test confirms wiring for step 2+ on `/new` and `/edit`

## Framework research

- htmx 2.0.9 (vendored, `VENDOR.md`): `htmx.process(elt)` takes an **Element** and enables htmx behavior on content added outside the normal request cycle (htmx.org/api). `hx-disabled-elt` is an inherited attribute; `hx-disinherit="hx-disabled-elt"` on a parent stops descendants from inheriting it while the parent still applies it to itself (htmx.org/attributes/hx-disinherit). `window.htmx` global guard mirrors the existing idiom in `topology-graph.js` / `broadcast-feed.js`.

## Out of scope

- The already-fixed #2117 "defect 1" (`closest()` invalid `>>` selector on the first step) — distinct root cause.

Closes #2174

### DIFF
--- a/backend/src/meho_backplane/ui/templates/runbooks/editor.html
+++ b/backend/src/meho_backplane/ui/templates/runbooks/editor.html
@@ -124,12 +124,23 @@
   </div>
 
   {# ---------- Form ---------- #}
+  {#-
+    ``hx-disabled-elt`` is an inherited htmx attribute: without ``hx-disinherit``
+    every descendant htmx request (each step body's debounced preview POST)
+    would inherit the form's ``find button[type=submit]`` and resolve ``find``
+    from the textarea — which has no submit-button descendant — logging
+    ``The selector "find button[type=submit]" on hx-disabled-elt returned no
+    matches!`` once per preview POST. Disinheriting it here keeps the value on
+    the form's own submit (self, not inherited) while stopping the leak into
+    the per-step triggers.
+  -#}
   <form
     id="runbook-editor-form"
     hx-post="{{ submit_url }}"
     hx-target="#runbook-editor-form"
     hx-swap="none"
     hx-headers='{"X-CSRF-Token": "{{ csrf_token | default('') }}"}'
+    hx-disinherit="hx-disabled-elt"
     x-on:submit="syncSteps()"
     x-bind:hx-disabled-elt="validationErrors().length > 0 ? 'find button[type=submit]' : null"
   >
@@ -396,10 +407,35 @@
     document.querySelectorAll(".runbook-step-body").forEach(enhanceTextarea);
   }
 
+  // htmx only scans the DOM it renders itself. A step cloned by Alpine's
+  // ``x-for`` is inserted outside the htmx request cycle, so its per-step
+  // ``hx-post="/ui/runbooks/preview"`` trigger never binds until the node is
+  // handed to ``htmx.process``. Processing the form root is idempotent —
+  // htmx skips nodes it has already initialised — so re-running it after every
+  // step add/remove and after each HTMX settle registers the new triggers
+  // without double-binding the existing ones.
+  function processSteps() {
+    if (typeof window.htmx === "undefined") {
+      return;
+    }
+    var form = document.getElementById("runbook-editor-form");
+    if (form) {
+      window.htmx.process(form);
+    }
+  }
+
+  // Mount CodeMirror on the step bodies AND register their htmx triggers.
+  // Both must run in every rescan hook: CodeMirror gives the textarea its
+  // editor, htmx.process binds the per-step live-preview POST.
+  function rescan() {
+    scanBodies();
+    processSteps();
+  }
+
   // Initial scan once the DOM (and Alpine's x-for) has rendered, plus a
   // rescan after HTMX settles new content and after Alpine adds a step.
   function init() {
-    scanBodies();
+    rescan();
   }
   if (document.readyState !== "loading") {
     // base.html scripts are deferred, so the DOM is parsed; give Alpine a
@@ -408,11 +444,11 @@
   } else {
     document.addEventListener("DOMContentLoaded", function () { setTimeout(init, 0); });
   }
-  document.body.addEventListener("htmx:afterSettle", scanBodies);
+  document.body.addEventListener("htmx:afterSettle", rescan);
   // Re-scan when a step is added/removed (Alpine re-renders the x-for).
   document.addEventListener("click", function (e) {
     if (e.target && e.target.closest && e.target.closest("#runbook-editor-form, [x-on\\:click]")) {
-      setTimeout(scanBodies, 0);
+      setTimeout(rescan, 0);
     }
   });
 })();
--- a/backend/tests/test_ui_runbooks_editor.py
+++ b/backend/tests/test_ui_runbooks_editor.py
@@ -978,6 +978,129 @@ def test_editor_new_step_preview_uses_valid_hx_target() -> None:
     assert ".flex-wrap >> .runbook-step-preview" not in body
 
 
+def test_editor_processes_htmx_on_dynamically_added_steps() -> None:
+    """The editor hands Alpine-cloned steps to ``htmx.process`` (#2174).
+
+    Each step's Body textarea carries ``hx-post="/ui/runbooks/preview"``, but
+    that trigger lives inside an Alpine ``x-for`` loop. htmx only binds
+    ``hx-*`` on the DOM it renders itself, so a step added via ``+ Add step``
+    (or any step past the first on an edit) is never htmx-processed and its
+    live preview never fires. The regression before #2174 was that the editor
+    script re-mounted CodeMirror on new step bodies but never called
+    ``htmx.process`` — so only the first step (present at htmx's initial page
+    scan) produced a preview.
+
+    This asserts the fix is wired into the rendered editor script: a
+    ``htmx.process`` call plus its use in the shared rescan path that runs on
+    ``htmx:afterSettle`` and after each add/remove click.
+    """
+    _seed_tenant(_TENANT_A, "tenant-a")
+    session_id, jwks = _admin_session()
+
+    mock = respx.mock(assert_all_called=False)
+    mock.start()
+    try:
+        _mock_discovery_and_jwks(mock, jwks)
+        client = _authenticated_client(session_id)
+        response = client.get("/ui/runbooks/new")
+    finally:
+        mock.stop()
+
+    assert response.status_code == 200, response.text
+    body = response.text
+    # The dynamically-added step subtree is registered with htmx.
+    assert "htmx.process" in body
+    # The rescan path (CodeMirror + htmx.process) is bound to the settle event
+    # and the add/remove click so new steps get processed, not just the first.
+    assert "htmx:afterSettle" in body
+
+
+def test_editor_form_disinherits_disabled_elt() -> None:
+    """The editor form disinherits ``hx-disabled-elt`` so step previews are clean (#2174).
+
+    ``hx-disabled-elt`` is an inherited htmx attribute. The form binds
+    ``find button[type=submit]`` (to disable submit while validation fails);
+    without ``hx-disinherit`` every descendant step's debounced preview POST
+    inherits it and resolves ``find button[type=submit]`` against the textarea
+    — matching nothing — logging
+    ``The selector "find button[type=submit]" on hx-disabled-elt returned no
+    matches!`` once per preview POST. The form must carry
+    ``hx-disinherit="hx-disabled-elt"`` to scope the value to its own submit.
+    """
+    _seed_tenant(_TENANT_A, "tenant-a")
+    session_id, jwks = _admin_session()
+
+    mock = respx.mock(assert_all_called=False)
+    mock.start()
+    try:
+        _mock_discovery_and_jwks(mock, jwks)
+        client = _authenticated_client(session_id)
+        response = client.get("/ui/runbooks/new")
+    finally:
+        mock.stop()
+
+    assert response.status_code == 200, response.text
+    body = response.text
+    assert 'hx-disinherit="hx-disabled-elt"' in body
+
+
+def test_editor_edit_multi_step_hydrates_every_step_preview_trigger() -> None:
+    """On edit, every hydrated step carries its own preview trigger (#2174).
+
+    The ``x-for`` renders one ``_step_fields.html`` per step, but only one
+    physical set of markup is in the response — the per-step preview trigger
+    is authored once and cloned per Alpine iteration. The regression is
+    binding, not markup: the trigger and its target must be present so that
+    once the added/hydrated step is htmx-processed the second step's preview
+    fires. This edit path seeds a two-step template and asserts the per-step
+    preview trigger + target + the htmx.process wiring are all present.
+    """
+    _seed_tenant(_TENANT_A, "tenant-a")
+    two_step_body = RunbookTemplateBody(
+        title="Two step",
+        description="A two-step runbook.",
+        target_kind="k8s-node",
+        steps=[
+            ManualStep(
+                id="first",
+                title="First",
+                body="First **body**.",
+                type="manual",
+                verify=ConfirmVerify(type="confirm", prompt="First done?"),
+            ),
+            ManualStep(
+                id="second",
+                title="Second",
+                body="Second **body**.",
+                type="manual",
+                verify=ConfirmVerify(type="confirm", prompt="Second done?"),
+            ),
+        ],
+    )
+    _seed_template(tenant_id=_TENANT_A, slug="two-step", body=two_step_body)
+    session_id, jwks = _admin_session()
+
+    mock = respx.mock(assert_all_called=False)
+    mock.start()
+    try:
+        _mock_discovery_and_jwks(mock, jwks)
+        client = _authenticated_client(session_id)
+        response = client.get("/ui/runbooks/two-step/edit")
+    finally:
+        mock.stop()
+
+    assert response.status_code == 200, response.text
+    body = response.text
+    # Both hydrated steps are in the Alpine model.
+    assert "first" in body
+    assert "second" in body
+    # The per-step preview trigger + target are present, and the editor
+    # processes the cloned step subtrees with htmx so step 2+ binds too.
+    assert 'hx-post="/ui/runbooks/preview"' in body
+    assert 'hx-target="next .runbook-step-preview"' in body
+    assert "htmx.process" in body
+
+
 # ---------------------------------------------------------------------------
 # Audit op-id on UI-path template writes (#2116)
 #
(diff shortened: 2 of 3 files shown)


damir-topic:
@coderabbitai review

damir-topic:




**Anchor task:** #2174 — runbook editor per-step live preview only fired for the first step; Alpine-cloned steps were never htmx-processed
**Surface area:** frontend (server-rendered HTMX/Jinja operator console)
**Verdict:** APPROVE
**Head SHA reviewed:** `f020b03d`
**AI reviewer (CodeRabbit):** not received within window — rate-limited ("Review limit reached; next review available in 40 minutes"). Proceeded without it per skill Hard rule 17 (bot unavailable ≠ blocker).

### Acceptance criteria
- [x] Each step's Body fires its own debounced `hx-post="/ui/runbooks/preview"` including steps added via **+ Add step** — `processSteps()` hands the form root to `htmx.process` (idempotent) in the shared `rescan()`, wired to init / `htmx:afterSettle` / add-remove click.
- [x] Newly-inserted step DOM is htmx-processed — `window.htmx.process(#runbook-editor-form)`, guarded by `typeof window.htmx === "undefined"` (idiom matches `topology-graph.js` / `broadcast-feed.js`).
- [x] The `hx-disabled-elt "find button[type=submit]" returned no matches` warning is scoped away — `hx-disinherit="hx-disabled-elt"` on the form stops the inherited attribute leaking into each descendant preview POST while keeping it on the form's own submit.
- [x] Tests confirm preview wiring for step 2+ on `/new` and `/edit` — 3 new tests (htmx.process on new page, disinherit on the form, multi-step `/edit` hydration carrying the per-step trigger + processing).

### Predicted vs. actual diff
- ✅ `editor.html` — `processSteps()` + `rescan()` + `hx-disinherit` (expected, correct)
- ✅ `test_ui_runbooks_editor.py` — 3 regression tests (expected, correct)
- ✅ `docs/codebase/ui.md` — new "Editor step-body preview (dynamic HTMX processing)" subsection (expected, correct)
- 🕳️ Missing: none. No route/schema touched → no OpenAPI snapshot regen (CLI API snapshot freshness lane confirms green).

### Integration verification
- [x] `pytest tests/test_ui_runbooks_editor.py` — 22 passed (in the review worktree on the PR head)
- [x] `pytest tests/test_ui_runbooks_{acceptance,lifecycle,list}.py` — 48 passed (no surface regression)
- [x] CI `Python (ruff + mypy + pytest)` — pass
- [x] CI `Python (integration testcontainers)` — pass
- [x] CI `CLI API snapshot freshness` — pass (no route/schema drift)
- [x] CI `DCO` — pass
- [x] htmx API cited: `htmx.process(elt)` takes an Element and enables htmx on out-of-cycle DOM (htmx.org/api); `hx-disinherit` on a parent stops descendant inheritance while the parent still applies it to itself (htmx.org/attributes/hx-disinherit); both present in vendored htmx 2.0.9.
- ⚠️ not run: no JS-executing (Playwright/jsdom) test — the operator console has no browser-test harness; server-rendered HTML assertions are the established pattern in this suite, so the tests gate the rendered fix strings that were absent pre-fix.

### Findings (inline)
None — no blockers, no majors, no minors.

- A browser-level test (Playwright/jsdom) would prove the actual `htmx.process` binding rather than the rendered markup/script; out of scope here — the console has no JS test harness and adding one is an initiative-level decision.

- Adopted: 0 (bot rate-limited; no review produced)
- Disputed: 0
- Out of scope: 0

### Docs
- [x] `docs/codebase/ui.md` — updated (new dynamic-HTMX-processing subsection under the Runbooks surface)

### What worked well
- Correct root-cause fix (htmx.process on out-of-cycle Alpine DOM) rather than a workaround; the idempotent form-root processing avoids double-binding.
- The `hx-disinherit` fix for the secondary console warning is the precise htmx-idiomatic scoping, cited to the docs.
- The `x-for`/htmx gotcha is documented in `ui.md` so the next author doesn't re-hit it.

PR #2307, opened 2026-07-09
fix(ingest): strip YAML 1.1 timestamp typing from spec ingest + dry-run parity

### LINKED ISSUE #2272: Task: strip YAML 1.1 timestamp typing from spec ingest + dry-run serialization parity — vendor YAML date examples crash the descriptor INSERT
## Summary

Ingesting a vendor-published **YAML** spec whose schema `example:` fields carry unquoted ISO timestamps (fleet-lcm `fleet-lcm-openapi.yaml` has 102 of them) crashes the descriptor INSERT with `StatementError: Object of type datetime is not JSON serializable` — and `dry_run=true` passes the identical spec clean. Reported from the v0.20.0 RDC test period; **every claim in the original report was confirmed against the code and by runtime probe** (a rare fully-accurate consumer report). Unfixed on main as of 2026-07-09.

## Current state (verified in code, v0.20.0 == main for the ingest tree)

- The YAML leg of `_decode_spec` parses with `yaml.CSafeLoader`/`SafeLoader` ([`openapi.py:97-103`](backend/src/meho_backplane/operations/ingest/openapi.py#L97-L103), consumed at [`:681`](backend/src/meho_backplane/operations/ingest/openapi.py#L681)) — stock YAML 1.1 implicit resolvers, so unquoted `2000-01-23T04:56:07.000+00:00` becomes `datetime.datetime` and date-only `2024-01-15` becomes `datetime.date` (probe-verified in the repo venv). The JSON leg (`json.loads`, [`:679`](backend/src/meho_backplane/operations/ingest/openapi.py#L679)) is immune.
- Schema subtrees ride **verbatim** into `EndpointDescriptorProto` (`_build_param_property` copies the resolved dict including `example`, [`openapi.py:1045`](backend/src/meho_backplane/operations/ingest/openapi.py#L1045); refs are returned uncopied, [`refs.py:80-81`](backend/src/meho_backplane/operations/ingest/refs.py#L80-L81)) and are assigned untouched to the `parameter_schema`/`response_schema` JSON/JSONB columns ([`_upsert.py:202-203`](backend/src/meho_backplane/operations/ingest/_upsert.py#L202-L203), `_PORTABLE_JSON` at [`models.py:307`](backend/src/meho_backplane/db/models.py#L307)). The engine sets no `json_serializer` ([`engine.py:101`](backend/src/meho_backplane/db/engine.py#L101)), so `session.flush()` hits stdlib `json.dumps` and dies with exactly the reported error — dialect-independent.
- The dry-run leg ([`pipeline.py:647` → `_run_dry_run` `:816-863`](backend/src/meho_backplane/operations/ingest/pipeline.py#L816-L863)) is parse-and-count only ("No DB writes, no LLM call") — the datetime-carrying protos are never JSON-encoded, so dry-run green-lights a spec the real run crashes on.
- OAS 3.1 §Format requires YAML tags be "limited to those allowed by the JSON Schema ruleset" — which excludes the timestamp tag. The parser is over-typing per the OpenAPI spec itself; the author meant a string.

## Desired state

1. **Loader fix (primary):** subclass the selected loader at [`openapi.py:97-103`](backend/src/meho_backplane/operations/ingest/openapi.py#L97-L103) and override the YAML 1.1 timestamp constructor to return the raw scalar text (`add_constructor("tag:yaml.org,2002:timestamp", lambda loader, node: node.value)`). Probe-verified: works under `CSafeLoader` (construction is Python-side), preserves the author's exact string, covers both `datetime` and `date` shapes, leaves quoted strings untouched, and PyYAML's copy-on-write `add_constructor` on a subclass leaves the parent loader unmutated. **Do NOT** take the engine-wide route (`json_serializer` with a datetime `default=` at `engine.py:101`) — `_PORTABLE_JSON` backs ~30 columns platform-wide; a permissive default fail-opens all of them and stores a mangled `str(datetime)` instead of the author's text.
2. **Dry-run parity (secondary):** assert JSON-serializability of `parameter_schema`/`response_schema` once at the proto-build boundary (shared by both legs) — so dry-run and the real run fail identically, at parse time, with the structured invalid-spec error rather than a `StatementError` at flush.

## Acceptance criteria

- [ ] A fixture YAML with unquoted full-timestamp AND date-only `example:` values (add under `backend/tests/fixtures/openapi/`) ingests successfully; the persisted `parameter_schema`/`response_schema` carry the original strings verbatim (e.g. `"2000-01-23T04:56:07.000+00:00"`).
- [ ] `dry_run=true` on a spec whose protos are not JSON-serializable fails with the same structured error the real run raises (extend the dry-run cases at `backend/tests/test_api_v1_connectors_ingest.py:2142` / `:3325`) — no more false-green.
- [ ] The JSON spec leg and all existing fixtures (`petstore_30.yaml`, `handauthored_minimal_30.yaml`, shipped catalog specs / boot validation) stay green; quoted timestamps still round-trip as strings.
- [ ] No engine-level `json_serializer` is introduced; the loader change is scoped to the spec-ingest module.
- [ ] No route/schema change expected → no OpenAPI snapshot delta (verify; regen via `cd cli && make snapshot-openapi && make generate` if a schema did move).

## Out of scope

- Normalizing non-crashing YAML 1.1 resolutions in example values (`on`→`True`, sexagesimal `1:30`→`90`) toward YAML 1.2 semantics — silent-fidelity nit, no serialization failure; file separately if a real spec surfaces it.
- Hoisting the timestamp-preserving loader to the other `yaml.safe_load` sites (catalog/profile/kubeconfig/etc.) — those inputs are MEHO-authored and pydantic-validated.
- Fixing the stale docstring claim at `openapi.py:1024-1025` ("example metadata is hoisted" — only description is) may ride along as a comment fix.

## References

- Original report: v0.20.0 RDC test period, consumer ticket evoila-bosnia/claude-rdc-hetzner-dc#1790; repro = the pinned vmware/vcf-api-specs `fleet-lcm-openapi.yaml` URL against any deploy.
- Follow-on defect chain: #2273 (debris from this crash), #2275 (async-path contrast — this failure terminated cleanly through the job boundary).
- OAS 3.1 Format/YAML-tag constraint: https://spec.openapis.org/oas/v3.1.0#format (tags limited to the JSON Schema ruleset).
- Best-practices anchors — *python*: fix input typing at the parse boundary, not with a permissive serializer at the storage layer (fail-closed); parity between validation-only and effectful code paths; regression tests at the layer the bug traverses (descriptor INSERT + dry-run leg).

Parent goal: #221
Parent initiative: #2286

damir-topic:
## Summary

- **Loader fix (primary):** the YAML leg of spec ingest now decodes with
  `_SpecYamlLoader`, a `CSafeLoader`/`SafeLoader` subclass that overrides
  the YAML 1.1 implicit `timestamp` constructor to return the raw scalar
  text. An unquoted `example: 2000-01-23T04:56:07.000+00:00` (fleet-lcm
  has 102 of them) no longer constructs a `datetime` that crashes the
  descriptor INSERT's `json.dumps` — it persists as the verbatim string
  the author wrote. Covers both the full-timestamp and date-only shapes.
- **Dry-run parity (secondary):** `_build_proto` asserts each
  `parameter_schema`/`response_schema` is JSON-serializable at the
  proto-build boundary both the real and `dry_run` legs share, so a
  non-encodable proto fails identically — a structured `invalid_schema`
  400 at parse time — instead of the real run crashing at
  `session.flush()` while `dry_run=true` green-lights the same spec.
- **Scope discipline:** no engine-wide `json_serializer` (it backs ~30
  columns platform-wide; a permissive default would fail-open all of
  them and store a mangled `str(datetime)`). `add_constructor` is
  copy-on-write, so the shared `yaml.safe_load` sites (catalog,
  kubeconfig, topology import) stay on the stock resolver — proven by a
  test.
- Corrects a stale `_build_param_property` docstring (only `description`
  is hoisted into the property schema, not `example`).

## Test plan

- [x] `mypy` (strict) — clean
- [x] New parser unit tests: unquoted date/timestamp `example:` values
  stay verbatim strings; a quoted timestamp still round-trips; a
  `!!binary` example raises `InvalidSchemaError`; the shared
  `yaml.safe_load` loader stays on the stock resolver.
- [x] New REST integration tests: a fixture spec with unquoted
  timestamp/date examples ingests (`200`) and the **persisted**
  `parameter_schema`/`response_schema` carry the exact strings; a
  non-serializable spec returns `invalid_schema` 400 on **both**
  `async=false` and `dry_run=true` (parametrized).
- [x] Regression sweep: `test_operations_ingest_openapi.py`,
  `test_api_v1_connectors_ingest.py` (221 passed) + register / pipeline
  / catalog / MCP-ingest / NSX / Hetzner / topology-batch-parse / UI
  registry-ingest (194 passed) — existing fixtures + JSON leg stay green.
- [x] `code-quality.py --diff` — 0 blockers.
- [x] `make snapshot-openapi` — no `cli/api/openapi.json` delta (no
  route/schema change).

## Out of scope (per the task)

- Normalizing non-crashing YAML 1.1 over-typings in example values
  (`on`→`True`, sexagesimal `1:30`→`90`) — documented as a remaining
  known-issue; file separately if a real spec surfaces it.
- Hoisting the timestamp-preserving loader to the other MEHO-authored,
  pydantic-validated `yaml.safe_load` sites.

## Adjacent findings (deferred, not addressed here)

- `_build_proto` / `_build_parameter_schema` / `_populate_param_properties`
  exceed the 60-line soft warn and carry pre-existing `PLR0913`
  many-argument warnings (0 blockers). Legacy debt in files touched; not
  in scope for this fix.

Closes #2272


damir-topic:
@coderabbitai review

Issue #1975, opened 2026-06-19
Add spec_resource/profile_resource catalog fields → resolve to SpecSource.content; update the catalog class-name validator


damir-topic:
Parent goal: #1964
Parent initiative: #1966
Depends-on: #1967, #1969

## Problem

The catalog row records only the vendor `upstream` URL, so a product whose upstream is an HTML portal or fqdn-templated URL the backend can't dereference (vmware/sddc) forces the operator to fetch + upload the spec by hand. We want to ship MEHO-authored specs + profiles as package data and have the catalog-driven ingest route consume them directly — and the catalog validator must accept a profile-backed row that has no hand-written connector class.

## Current state (verified in code, v0.18.0)

- `ConnectorSpecEntry` + the [`operations/ingest/catalog.py`](backend/src/meho_backplane/operations/ingest/catalog.py) loader resolve the catalog via `importlib.resources`.
- The catalog-driven ingest route builds `SpecSource(uri=...)` from `upstream` and **422s** on an un-fetchable upstream.
- The catalog validator asserts each row's connector class `__name__` exists in `all_connectors_v2` (the `upstream:null==typed` invariant).
- `SpecSource` content-vs-uri at [`operations/ingest/api_schemas.py:92-124`](backend/src/meho_backplane/operations/ingest/api_schemas.py#L92-L124).
- Packaging precedent: [`pyproject.toml:107`](pyproject.toml#L107) already uses `[tool.hatch.build.targets.wheel.force-include]` for alembic.

## Desired state

- Add optional `spec_resource` / `profile_resource` fields to `ConnectorSpecEntry`; the catalog-driven ingest route fills `SpecSource.content` from the shipped resource via `importlib.resources` instead of 422-ing on an un-fetchable upstream.
- Add the specs/profiles dir to the wheel `force-include` (one-line, mirroring the alembic precedent).
- **Update the catalog validator / `requires_connector_class` regression test** so a profile-backed row (whose synthesised `ProfiledRestConnector` name may not pre-exist) does not fail the "class `__name__` must exist" assertion.
- Dry-run-parse every shipped spec/profile at startup with the **same** parser; crash boot on a malformed artifact.

## Acceptance criteria

- [ ] A catalog row may carry a shipped `spec_resource`/`profile_resource`; catalog-driven ingest loads the bytes into `SpecSource.content`, bypassing fetch/upload.
- [ ] The specs/profiles dir is force-included in the built wheel (verified by inspecting the artifact).
[...]

Issue #403, opened 2026-05-14
Task: G0.7-T2 register_ingested_operations() helper — bulk upsert + multi-spec merge + auto-register connector class + body-hash skip


damir-topic:
Parent initiative: #389
Depends-on: #392, #393, #401

## Target

Ship the **`register_ingested_operation()` async helper** that takes parser output from T1 (#401) and writes endpoint_descriptor rows to the database. Multi-spec merge: a single `connector_id` accepts multiple specs ingested (e.g. vSphere's vcenter.yaml + vi-json.yaml), distinguishable via `spec_source` tags on the rows.

Parallel pathway to G0.6-T4 (#395, `register_typed_operation()` for typed connectors). Both write to the same `endpoint_descriptor` table but use different `source_kind` values (`'ingested'` here, `'typed'` in G0.6-T4).

## Files touched

- `backend/src/meho_backplane/operations/ingest/register_ingested.py` — new module.
- `backend/src/meho_backplane/operations/ingest/connector_registration.py` — helper that ensures the connector class is registered via G0.6-T2 (#393) `register_connector_v2()` when ingesting against a previously-unknown `(product, version, impl_id)`.
- Tests at `backend/tests/test_operations_register_ingested.py`.

## API contract

```python
from meho_backplane.operations.ingest.schemas import EndpointDescriptorProto
from typing import Sequence

async def register_ingested_operations(
    *,
    product: str,
    version: str,
    impl_id: str,
    spec_source: str,                    # tag for this spec (e.g. "vcenter.yaml")
    operations: Sequence[EndpointDescriptorProto],
    base_url: str | None = None,         # for the auto-registered HttpConnector class
    tenant_id: UUID | None = None,       # NULL for built-in
) -> IngestionResult:
    """Bulk-upsert parsed operations into endpoint_descriptor.

    Per row:
      - source_kind = 'ingested'
      - tags = proto.tags + [f"spec:{spec_source}"]
      - embedding computed from summary + description + custom_description + tags
        (via G0.4's embedding service; same pattern as G0.6-T4)
      - body-hash skip: unchanged content (per natural key) → no re-embed
      - all rows inserted in 'staged' state (review_status of the parent
        operation_group is 'staged' on first ingest; T4 transitions them later)
      - parameter_schema includes the x-meho-param-loc extension from T1

    Connector class registration:
      - On first ingestion of a (product, version, impl_id) tuple, this helper
        ensures a Connector subclass is registered against G0.6-T2's registry
[...]

PR #2311, opened 2026-07-09
fix(ingest): normalize spec: prefix in spec_source dedup so catalog re-ingest is idempotent (#2274)

### LINKED ISSUE #2274: Task: normalize the spec: prefix in spec_source dedup — catalog re-ingest collides with its own rows (tag-shadowed extractor)
## Summary

Catalog re-ingest is non-idempotent: the second `meho connector ingest --catalog <product>/<version>` on the same deploy 400s with `op_id_collision` against its own rows, and the unbacked-composite remediation loop (which prints exactly that command) is circular. Symptom fully confirmed and **reproduced by executing the code against the real shipped `vmware_rest_minimal.yaml`**; the root cause is subtly different from the original framing: it is **extractor tag-shadowing**, not old-vs-new row labels — even a fresh v0.20.0 deploy hits it on its second catalog ingest, so **no data migration is needed**. This is the bug Goal #2247's body explicitly recommended filing.

## Current state (verified in code + executed repro, v0.20.0 == main for the ingest tree)

The dedup key disagrees with itself about the `spec:` prefix, via a double tag:

1. The catalog shipped-spec route labels the spec `SpecSource(uri=f"spec:{entry.spec_resource}")` ([`connectors_ingest.py:916`](backend/src/meho_backplane/api/v1/connectors_ingest.py#L916), since PR #2022 / Task #1975, v0.19.0+); `pipeline.py:900/906` pass it verbatim as `spec_source`.
2. The parser appends `spec_source` **verbatim** as a row tag → `spec:vmware_rest_minimal.yaml` ([`openapi.py:812-813`](backend/src/meho_backplane/operations/ingest/openapi.py#L812-L813)); the upsert then appends its own marker `f"spec:{spec_source}"` → `spec:spec:vmware_rest_minimal.yaml` ([`_upsert.py:92`](backend/src/meho_backplane/operations/ingest/_upsert.py#L92)).
3. On re-ingest, `_extract_persisted_spec_source` ([`_upsert.py:113-126`](backend/src/meho_backplane/operations/ingest/_upsert.py#L113-L126)) returns the **first** `spec:`-prefixed tag minus one prefix — the verbatim tag shadows the marker — recovering bare `vmware_rest_minimal.yaml`. The guard at [`_upsert.py:243-252`](backend/src/meho_backplane/operations/ingest/_upsert.py#L243-L252) compares it to the incoming `spec:vmware_rest_minimal.yaml` → mismatch → `OpIdCollision` → HTTP 400, aborting the whole batch at the first pre-existing row (`register_ingested.py:399-414` has no per-row handling). The extractor's own docstring assumes "exactly one spec: tag" — written in the #403/#478 era, before `spec:`-prefixed labels existed.
4. `file://`/`https://`/`docs:` labels are self-consistent (verbatim tag doesn't start with `spec:`) — only the catalog shipped-spec path breaks. Consequence: an upgraded deploy can never refresh catalog rows, and every remediation string pointing at `ingest --catalog` (`composite_backing.py:13,109`, `composite.py:180,255`, `meta_tools.py:280`, `api_schemas.py:519,598`) is circular once rows exist. Module docstring `register_ingested.py:49-59` explicitly promises the re-ingest idempotency this breaks.
5. Why CI never caught it: unit tests pass bare `spec_source` values; the REST re-ingest idempotency test uses a file-based uri; the shipped-spec catalog tests run `dry_run=True` only (`test_api_v1_connectors_ingest.py:3478`, dry_run at `:3515`) — nothing `spec:`-labeled is ever persisted-then-re-ingested.

## Desired state

**Compare-time normalization** at `_upsert.py:243-252`: normalize both sides of the `spec_source` equality (strip leading `spec:` prefixes; equivalently fix `_extract_persisted_spec_source` to prefer the **last** `spec:` tag — the marker is always appended last — then normalize the incoming label once). Verified by simulation: existing double-tagged rows and the incoming `spec:<resource>` label normalize equal → falls through to the body-hash skip/re-embed branch (idempotent), while genuine cross-source collisions (`file:///a.yaml` vs `spec:b.yaml`) still fire. Keep the RAW labels in the `OpIdCollision` payload for diagnostics. **Zero data migration** for any historical row shape.

Explicitly rejected shapes: persist-time-only marker suppression at `_upsert.py:92` does **not** fix the comparison (bare extracted vs prefixed incoming still collides) and strands existing double-tagged rows; changing the `connectors_ingest.py:916` label self-heals but breaks the documented `spec:<resource>` audit-label convention (boot guard `catalog.py:784-788`, CLI display `ingest_status.go:258-262`) and forces a fleet-wide re-embed.

## Acceptance criteria

- [ ] End-to-end: POST a catalog-entry shipped-spec ingest **twice without dry_run** (extend the `:3478` fixture in `backend/tests/test_api_v1_connectors_ingest.py`); the second run returns success with `inserted_count == 0` (skip/idempotent), not 400.
- [ ] Unit: re-ingest twice with `spec_source='spec:x.yaml'` where protos come from `parse_openapi` (hand-built protos lack the verbatim uri tag and will NOT reproduce the bug — the test must go through the parser); second call skips/updates.
- [ ] The genuine cross-source collision test (`test_operations_register_ingested.py:483`) stays green — different-source collisions still fire by design.
- [ ] The MCP catalog-ingest leg behaves identically (same `_upsert` seam; assert via `test_mcp_tools_connector_ingest.py` or note the shared-seam coverage).
- [ ] `OpIdCollision` payload still carries the raw persisted/incoming labels.
- [ ] No route/schema change expected → no OpenAPI snapshot delta (verify).

## Out of scope

- Healing the cosmetic `spec:spec:<resource>` double tag on existing rows (baked into embedding text; rewriting forces a re-embed of every catalog row) — the skip path leaves tags untouched; file separately if the review-drawer display bothers operators.
- Same-file-different-source refresh semantics (a `file:///` re-host taking over `spec:`-labeled rows) — deliberate collision per #403; needs a supersede/override story if ever wanted (see #2065's not-filed `--replace` note).

## References

- Original report: v0.20.0 RDC test period, consumer ticket evoila-bosnia/claude-rdc-hetzner-dc#1790; recommended for filing by Goal #2247's verification ("the `spec:` prefix defeats `spec_source` dedup").
- Label origin: Task #1975 / PR #2022 (commit abdd1602, v0.19.0). Guard origin: #403. Circular-remediation context: #2247 failure-mode 5, #2186.
- Best-practices anchors — *python*: normalize at the comparison seam rather than migrating data; keep raw values in diagnostics; close the test gap at the exact path CI missed (persist-then-re-ingest through the parser, non-dry-run catalog leg).

Parent goal: #221
Parent initiative: #2286

damir-topic:
## Summary

- **Fixes non-idempotent catalog re-ingest (#2274).** A second
  `meho connector ingest --catalog <product>/<version>` on the same deploy
  400'd with `op_id_collision` against the rows the *first* ingest just
  wrote, which made the unbacked-composite remediation loop (it prints
  exactly that command) circular.
- **Root cause is extractor tag-shadowing, not old-vs-new row labels.**
  The catalog shipped-spec on-ramp labels its source `spec:<resource>`
  (#1975); the parser persists that verbatim as a row tag and the upsert
  appends its own `spec:<spec_source>` marker, so a catalog row carries
  *two* `spec:` tags. `_extract_persisted_spec_source` returned the
  **first** matching tag — the verbatim one, one `spec:` layer short of
  the real source — so the cross-source collision guard mismatched a row
  against itself.
- **Compare-time fix, zero data migration.** The extractor now reads the
  authoritative **last** (marker) tag, and the guard compares on the
  prefix-normalized *logical* spec identity (`_normalize_spec_source`).
  The guard was extracted into `_check_cross_source_collision` to keep
  `upsert_one_operation` small.
- **Invariants preserved.** Genuine cross-source `op_id` clashes
  (`file:///a.yaml` vs `spec:b.yaml`) still fire; the `OpIdCollision`
  payload still carries the *raw* persisted/incoming labels for
  diagnostics. No route/schema change — the OpenAPI snapshot is unchanged.

## Test plan

- [x] `mypy src/.../ingest/_upsert.py` — clean
- [x] `code-quality.py --diff` — 0 blockers (the one remaining warning,
  `build_upsert_context` 7-args, is pre-existing debt not touched here)
- [x] **AC1 — e2e non-dry-run, twice:** new
  `test_ingest_catalog_entry_shipped_spec_reingest_is_idempotent` POSTs a
  shipped-spec catalog ingest twice without `dry_run`; the second returns
  `200` with `inserted_count == 0` (`skipped_count == 1`), not `400`.
- [x] **AC2 — parser-based unit:** new
  `test_reingest_with_spec_prefixed_source_is_idempotent` parses the
  petstore fixture through `parse_openapi` with a `spec:`-prefixed source
  (a hand-built proto would not reproduce the shadow) and re-registers;
  the second call skips. Asserts both `spec:petstore_30.yaml` (verbatim)
  and `spec:spec:petstore_30.yaml` (marker) tags exist to stay anchored
  to the real bug regime.
- [x] **AC3 — cross-source still collides:**
  `test_op_id_collision_across_calls_with_different_spec_sources_raises`
  stays green (asserts `existing_spec_source == "petstore.yaml"`).
- [x] **AC4 — MCP shared seam:** the `meho.connector.ingest` MCP tool
  calls the same `IngestionPipelineService.ingest` →
  `register_ingested_operations` → `upsert_one_operation` seam the fix
  lives at (`mcp/tools/connector_ingest.py:300`). The MCP test suite
  mocks the pipeline boundary (canned results), so the behavioral proof
  is the REST e2e + register-level unit test; `test_mcp_tools_connector_ingest.py`
  stays green (144 + 1 new).
- [x] **AC5 — raw labels:** `OpIdCollision` still receives the raw
  extracted + raw incoming labels (verified: reverting only `_upsert.py`
  reproduces `existing_spec_source:"_fixture_minimal.yaml"` /
  `incoming_spec_source:"spec:_fixture_minimal.yaml"` in the 400 body).
- [x] **AC6 — no OpenAPI delta:** `make snapshot-openapi` produced no diff
  on `cli/api/openapi.json`.
- [x] Regression sweep: `test_operations_register_ingested.py`,
  `test_api_v1_connectors_ingest.py`, `test_mcp_tools_connector_ingest.py`
  (145 passed) + ingest/catalog/parser/pipeline/registration files
  (309 passed, 1 skipped). Integration ingest tests skip in-sandbox
  (Docker/testcontainers; run in CI).

## Out of scope

- Healing the cosmetic `spec:spec:<resource>` double tag on existing rows
  (baked into embedding text; rewriting forces a re-embed) — the skip
  path leaves tags untouched.
- Same-file-different-source refresh semantics (a `file:///` re-host
  taking over `spec:`-labeled rows) — a deliberate collision per #403;
  needs a supersede/`--replace` story if ever wanted.

Closes #2274

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,26 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Fixed — catalog re-ingest is idempotent instead of colliding with its own rows
+
+- **A second `meho connector ingest --catalog <product>/<version>` on the
+  same deploy now returns an idempotent skip instead of a 400
+  `op_id_collision` against the rows the first ingest just wrote** (#2274).
+  The catalog shipped-spec on-ramp labels its source `spec:<resource>`;
+  the parser then persists that label verbatim as a row tag while the
+  upsert appends its own `spec:<spec_source>` marker, so a catalog row
+  carries two `spec:` tags. The cross-source collision guard recovered the
+  persisted source from the *first* matching tag — the verbatim one, one
+  `spec:` layer short of the real source — so every re-ingest mismatched
+  its own rows and aborted the batch, making the unbacked-composite
+  remediation loop (which prints exactly that command) circular. The guard
+  now reads the authoritative *last* (marker) tag and compares on the
+  prefix-normalized logical spec identity; genuine cross-source `op_id`
+  clashes (`file:///a.yaml` vs `spec:b.yaml`) still fire, and the
+  structured exception still carries the raw persisted/incoming labels for
+  diagnostics. Compare-time fix only — no data migration, no schema
+  change. (#2274)
+
 ### Fixed — vendor YAML date/timestamp `example:` values no longer crash spec ingest
 
 - **Ingesting a YAML OpenAPI spec whose schema `example:` fields carry
--- a/backend/src/meho_backplane/operations/ingest/_upsert.py
+++ b/backend/src/meho_backplane/operations/ingest/_upsert.py
@@ -110,20 +110,81 @@ def build_upsert_context(
     )
 
 
+def _normalize_spec_source(spec_source: str) -> str:
+    """Reduce a ``spec_source`` label to its prefix-free logical identity.
+
+    Cross-call dedup keys on the *logical* spec a row came from, not on
+    the exact rendering of its audit label. The catalog shipped-spec
+    on-ramp labels its source ``spec:<resource>`` (#1975), so the same
+    logical spec can reach the comparison carrying a different number of
+    leading ``spec:`` prefixes on each side -- e.g. a persisted marker
+    that round-trips ``spec:<resource>`` versus an incoming
+    ``spec:<resource>`` re-submission. Stripping the leading ``spec:``
+    layers off both sides makes those compare equal (idempotent
+    re-ingest) while leaving genuinely distinct sources apart:
+    ``file:///a.yaml`` never normalizes onto ``spec:b.yaml``.
+    """
+    while spec_source.startswith(_SPEC_TAG_PREFIX):
+        spec_source = spec_source[len(_SPEC_TAG_PREFIX) :]
+    return spec_source
+
+
 def _extract_persisted_spec_source(existing: EndpointDescriptor) -> str | None:
     """Recover the ``spec_source`` of an already-persisted row.
 
-    The synthetic ``f"spec:{spec_source}"`` marker is appended once
-    per row at first-register (and re-applied on the re-embed path),
-    so a well-formed row carries exactly one ``spec:`` tag. If the
-    tag is missing (older rows, hand-edited fixtures), return
-    ``None`` and let the caller skip the cross-call check rather
-    than raise a spurious collision.
+    ``build_upsert_context`` appends the synthetic ``f"spec:{spec_source}"``
+    marker as the row's **last** tag, so the marker is the authoritative
+    record of the source. The parser also persists the raw ``spec_source``
+    verbatim as a tag (:func:`parse_openapi`); for a catalog shipped-spec
+    source -- already ``spec:<resource>`` -- that verbatim tag is itself
+    ``spec:``-prefixed and sits *before* the marker. Scanning from the
+    end returns the marker instead of letting the verbatim tag shadow it:
+    a first-match scan recovered the verbatim tag's payload, one ``spec:``
+    layer short of the real source, so every catalog re-ingest then
+    collided with its own rows (#2274). If no ``spec:`` tag is present
+    (older rows, hand-edited fixtures) return ``None`` and let the caller
+    skip the cross-call check rather than raise a spurious collision.
     """
-    for tag in existing.tags or []:
+    for tag in reversed(existing.tags or []):
         if tag.startswith(_SPEC_TAG_PREFIX):
             return tag[len(_SPEC_TAG_PREFIX) :]
     return None
+
+
+def _check_cross_source_collision(existing: EndpointDescriptor, ctx: UpsertContext) -> None:
+    """Raise :exc:`OpIdCollision` when a persisted row came from a *different* spec.
+
+    A row already exists under this ``(product, version, impl_id, op_id)``
+    natural key. If a prior ingest wrote it under a different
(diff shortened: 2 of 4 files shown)


damir-topic:
@coderabbitai review

Issue #1117, opened 2026-05-26
Task: G11.1 agent runtime — resume awaiting_approval runs on approval.{approved,rejected} broadcast (operator/agent split)


damir-topic:
Parent goal: #800
Parent initiative: #802

## Context

After G11.2 #803 landed: when an agent calls a `requires_approval=True` op, the dispatcher writes the durable pending row, the synchronous decision audit row's request half, publishes `approval.pending` on broadcast, and returns the agent an "awaiting approval" handle. The agent run is now **suspended** awaiting operator action. When the operator decides, broadcast publishes `approval.{approved,rejected}` with the request's `tenant_id` + `audit_id`.

**The agent runtime ([G11.1 #802](https://github.com/evoila/meho/issues/802)) does not subscribe to these events.** Its run has no resume path on approval.

This works **today only for the REST `/approve` with `params` path** — that endpoint re-dispatches itself with `dispatch(..., _approved=True)` ([dispatcher.py](backend/src/meho_backplane/operations/dispatcher.py)), so the human-as-operator who approved is also the one who executes. The MCP `meho.approvals.approve` path and the `/decide` operator-decision route capture the decision durably but do **not** re-dispatch — documented as the deliberate operator/agent split in the [auto-implement-initiative #803 run summary](https://github.com/evoila/meho/issues/803#issuecomment-4542158277).

Without this Task, an agent run that bridged a `requires_approval` op via any path other than REST `/approve+params` is dead-on-arrival: the operator can approve via CLI/MCP/UI/wall-monitor, but the agent run that started it never finds out.

## Current state

- `operations/dispatcher.py::_handle_needs_approval` writes the pending row + publishes `approval.pending` and returns an "awaiting approval" handle to the caller.
- `api/v1/approvals.py` approve route re-dispatches via `dispatch(..., _approved=True)` — works for the REST+params case only.
- `api/v1/approvals.py::/decide`, `mcp/tools/approvals.py::meho.approvals.{approve,reject}` capture the decision + publish broadcast but do **not** re-dispatch.
- The agent runtime (G11.1 #802) has no broadcast subscription scoped to its own `run_id` / `agent_session_id`.

## Desired state

- The agent runtime subscribes (via the existing broadcast SSE/watch substrate) to events filtered on its own `run_id` / `agent_session_id`.
[...]

PR #2312, opened 2026-07-10
fix(approvals): exactly-one-resumer claim for run-bound approvals

### LINKED ISSUE #2293: Task: exactly-one-resumer claim for run-bound approvals — /decide+MCP silently skip execution when the waiter is gone; /approve+UI can double-dispatch while it lives
## Summary

Run-bound approval requests have **no exactly-one-resumer invariant**, producing two opposite failure modes on current main (v0.20-current, dedupe-clean — surfaced while grounding the stale #2246 re-file):

- **Silent non-execution:** `/decide` and the MCP by-id approval guard on `run_id is not None` and *skip re-dispatch entirely*, assuming the in-process broadcast waiter (#1117) will resume the run. When the waiter is gone — wait timeout exceeded (>1800 s), pod restart, run cancelled — an approval lands, the audit says approved, and **nothing executes**, with no signal.
- **Potential double-execution:** REST `/approve` and the UI approve path re-dispatch `_approved=True` **unconditionally**; if the waiter is still alive it also resumes — one approval can plausibly dispatch an approval-gated **write** twice. The `/decide` 409 already-decided guard prevents double *decisions*, not the waiter+/approve double *dispatch* of one decision. `operations/meta_tools.py:1185-1188` documents the coexistence "by design (#1117)" — but the protective premise is unenforced (REST `/approve` has no run_id discrimination).

Governance-loop integrity issue for approval-gated writes: silent skip on two surfaces, unguarded duplicate dispatch on the other two.

- In-process waiter resume: [`agent/approval_wait.py:528-540`](backend/src/meho_backplane/agent/approval_wait.py#L528-L540) (trigger-agnostic — scheduler-fired runs use the same path; the #2246 hypothesis stays refuted).
- Shared server-side re-dispatch: `resume_dispatch_after_approval` ([`operations/approval_queue.py:634`](backend/src/meho_backplane/operations/approval_queue.py#L634)) called by REST `/approve` ([`api/v1/approvals.py:451`](backend/src/meho_backplane/api/v1/approvals.py#L451)) and UI approve ([`ui/routes/approvals/routes.py:628-630`](backend/src/meho_backplane/ui/routes/approvals/routes.py#L628-L630)) — unconditional.
- Run-bound skip guards: `/decide` at [`api/v1/approvals.py:638-639`](backend/src/meho_backplane/api/v1/approvals.py#L638), MCP at [`mcp/tools/approvals.py:418`](backend/src/meho_backplane/mcp/tools/approvals.py#L418) — no fallback when the waiter is gone.
- No expiry is set at park time (`dispatcher.py` `_handle_needs_approval`), so a decision landing after the waiter died is approvable-into-a-void on `/decide`/MCP.

## Desired state

An **exactly-one-resumer claim**: add a conditional resume-claim column (e.g. `approval_request.resumed_at`; single `UPDATE … WHERE resumed_at IS NULL` as the claim) that every dispatcher of `_approved=True` must win first — the in-process waiter, `resume_dispatch_after_approval`, and any future resumer. Then relax the `run_id`-is-not-None skip guards on `/decide` + MCP to **fall back to server-side re-dispatch when the claim is free** (covers waiter-gone: timeout/restart/cancel), while the same claim prevents the `/approve`/UI double-dispatch when the waiter is alive. Correct the `meta_tools.py:1185-1188` docstring to describe the enforced invariant instead of the unenforced premise.

## Acceptance criteria

- [ ] New Alembic migration (next free number; pin idempotency tests to its own revision) adding the claim column; claim is a single conditional UPDATE (no advisory locks).
- [ ] Waiter-alive + REST `/approve` (or UI approve): the gated op executes **exactly once** — integration test asserting one dispatch/audit row (this is the double-dispatch polarity test the grounding flagged as code-confirmed but runtime-unverified).
- [ ] Waiter-gone (timeout/cancel/simulated restart) + `/decide` or MCP approve: the op executes exactly once via the fallback re-dispatch — re-dispatch the LIVE fixture target with `_approved=True` (NOT `resume_dispatch_after_approval` rehydrating a DB Target → `no_connector`; the #147/#2173 trap), and update the integration-double dispatch sites.
- [ ] A second resumer arriving after the claim is taken no-ops cleanly (no error surfaced to the approver beyond "already executed").
- [ ] Audit trail: exactly one execution audit row per approval, stamped with the requester sub as today; `meta_tools.py` docstring corrected.
- [ ] No route/schema change expected beyond the migration (verify snapshot).

## Out of scope

- Default `expires_at` at park time aligned with the waiter timeout (reasonable follow-up so void-approvals become visibly "expired"; keep optional).
- Re-litigating #2246's filed hypothesis (refuted; see #2286 corrections table) — this task is the residual seam, valid regardless of the v0.20 retest.
- Approval policy/queue semantics beyond resume (unchanged).

## References

- Prior art establishing the split this enforces: #1117 (broadcast resume, operator/agent split), #1401 (policy-gate queue), #147/#2173 (approve→resume integration-test traps).
- Stale re-file that surfaced this: #2246 (v0.10.1 observation; real fixes were #1501/#1502/#1503 in v0.11.0 under G0.20 #1500).
- Best-practices anchors — *python*: exactly-once side effects via a DB claim (conditional UPDATE), not in-process assumptions; fail-open fallback only after winning the claim; integration tests at the surface pair that races (waiter × REST), doubles updated.

Parent goal: #221
Parent initiative: #2286

damir-topic:
## Summary

- Adds an **exactly-one-resumer claim** for run-bound approvals: a new
  `approval_request.resumed_at` column (migration `0055`) written by a single
  conditional `UPDATE ... WHERE resumed_at IS NULL` (`claim_resume`) that every
  resumer of an approved op must win before it re-dispatches `_approved=True`.
- Closes two opposite governance-loop seams: **silent non-execution** (REST
  `/decide` + MCP by-id approve skipped re-dispatch whenever `run_id` was set,
  so a dead in-process waiter meant the approval committed but nothing ran) and
  **double dispatch** (`/approve` + UI re-dispatched unconditionally, so a live
  waiter + `/approve` could run an approval-gated write twice).
- `/decide` + MCP now fall back to a server-side re-dispatch when the claim is
  free (covers waiter-gone: timeout / restart / cancel); the same claim blocks
  the `/approve` / UI double-dispatch when the waiter is alive. A losing resumer
  no-ops cleanly (`already_resumed`, HTTP 200) — no error to the approver.
- Corrects the `call_operation_with_approval` docstring to describe the
  **enforced** claim invariant instead of the old unenforced "coexist by
  design (#1117)" premise.

## How it works

- One claim site covers all four operator surfaces (they route through
  `resume_dispatch_after_approval`); a second claim site guards the in-process
  agent waiter (`approval_wait.resume_or_surface_awaiting_approval`). The claim
  runs in its own committed transaction so it serialises across processes/pods
  via row-locking — no advisory locks. It is a one-way latch (never cleared),
  so a failed dispatch is not silently retried into a possible double write.
- The `run_id`-is-not-None skip guards on `/decide` (REST) and MCP are removed;
  the claim, not the guard, now prevents the waiter-alive double.

## Test plan

- [x] `mypy` (7 changed source modules) — clean
- [x] `.claude/hooks/code-quality.py --diff` — 0 blockers (two large touched
  functions refactored into cohesive helpers to stay under the size budget)
- [x] **Migration `0055`** unit test — pinned to its own revision (`0055` /
  down-rev `0054`), never `head`, per the 0049/0050/0053 stamp-replay footgun
- [x] **Claim primitive** (`test_approval_queue.py`) — wins once, loses on the
  second call; `resume_dispatch_after_approval` returns `already_resumed` +
  never dispatches when the claim is taken; wins + stamps when free
- [x] **Waiter** (`test_agent_approval_resume.py`) — resumes on a free claim;
  no-ops with an `already_resumed` envelope (no re-dispatch) when an operator
  surface already claimed
- [x] **`/decide` + MCP relaxation** (`test_approval_queue.py`,
  `test_secret_move_approval.py`) — a run-bound approval re-dispatches once when
  the claim is free (waiter-gone fallback) and no-ops when it's taken
  (waiter-alive dedup); `write_secret` fires exactly once / zero times
- [x] **Integration (Postgres, real concurrency)** —
  `test_approval_exactly_one_resumer_e2e.py`: N concurrent `claim_resume` yield
  exactly one winner; two concurrent resumers dispatch the approved op exactly
  once (one `ok`, one `already_resumed`). Docker-gated (runs in CI)
- [x] Full approval suite (134 tests across 7 files) green; obsolete
  "`/decide` does NOT re-dispatch an agent-run request" test rewritten to the
  new contract
- [x] OpenAPI snapshot re-generated — delta is **description text only** (the
  `/decide` route + `DecideResponseBody` docstrings); no structural
  route/schema change beyond the migration

## Out of scope

- Default `expires_at` at park time aligned with the waiter timeout (the
  optional follow-up so a void approval becomes visibly "expired").
- Existing live-container approve→resume doubles (harbor/vault/bind9) re-dispatch
  directly via `dispatch(_approved=True)` and bypass the claim sites, so they are
  unaffected; not modified.

Closes #2293


damir-topic:
@coderabbitai review

Issue #1803, opened 2026-06-16
Task: operator-console modals don't dismiss — static modal-open class survives native dialog .close()


damir-topic:
Parent goal: #221
Parent initiative: #1800

## Why now

RDC v0.16.0 operator browser pass: **operator-console modal windows do not close** — the close ("X" / Cancel) buttons are non-functional. The modal opens, but no close button, backdrop click, or ESC dismisses it; the operator must reload the page. This breaks every modal-driven action in the console (approvals approve/deny, memory create/promote, connector create/delete, kb editor, runbook actions). It was surfaced by the expanded modal usage shipped in v0.16.0 (#1778 approvals bell + modal).

## Current state (grounded in code, v0.16.0)

~9 templates inject a `<dialog class="modal modal-open">` fragment via htmx swap and rely on **DaisyUI's `modal-open` class** to show it (no `showModal()` in most paths). Every close control — the X/Cancel buttons **and** the backdrop `<form method="dialog">` — calls `document.getElementById(...).close()`.

The native `HTMLDialogElement.close()` clears the `[open]` attribute but **does not touch CSS classes**. DaisyUI 5's `modal-open` modifier explicitly *"keeps the modal open"* — so after `.close()` the `modal-open` class is still present and the dialog stays visible. ESC and backdrop-click both route through `.close()` (or the `form[method=dialog]` submit, which also only clears `[open]`), so **there is no workaround**.

Confirmed-affected templates (all carry static `class="modal modal-open"` + a `.close()` handler):

- `ui/templates/approvals/_panel.html:20,30,75`, `approvals/_modal.html:40,50,159`, `approvals/_decided.html:19,46,53`
- `ui/templates/memory/_create_modal.html:47,218`, `memory/_promote_modal.html:32,96`
- `ui/templates/connectors/_create_modal.html:26,46`, `connectors/_delete_modal.html:26,69`
- `ui/templates/kb/_editor_modal.html`, `ui/templates/runbooks/_list.html:112`, `runbooks/_detail_actions.html:117`
- Shared container: `ui/templates/base.html:335-353` (the approvals modal swap target).

This is **one shared root cause → one fix applied across the templates**, not a per-modal handler bug.

## Desired state

[...]

PR #2314, opened 2026-07-10
fix(ui): guard shared modal controller against htmx's detached pre-swap target

### LINKED ISSUE #2242: Task: guard the shared modal controller against htmx pre-swap detail.target — outerHTML swaps with descendant dialogs throw InvalidStateError
## Summary

Re-diagnosed from the original filing ("orphaned dialog reference after Abort"): the shared modal controller **re-queries dialogs on every swap** — there is no held reference — but it scans `event.detail.target`, and vendored **htmx 2.0.9 leaves that field pointing at the PRE-swap element**, which an `outerHTML` swap has already detached. The handler then scans the detached old `#runbook-run-step` subtree and calls `showModal()` on a dialog that is not connected → `InvalidStateError`. The thrower on the Abort repro is not the abort dialog (still `open`, skipped by the guard) but the sibling **closed admin-only reassign dialog** — so the filed repro needs tenant_admin, while the same error is code-predicted for the run's **assignee on every Advance click** (steady console noise for anyone driving a run) and on account-sessions revoke with ≥2 sessions. Console-only; abort is fully functional. Fix is a one-line connectivity guard in the shared controller.

- [`ui/static/src/app/modal-dialogs.js:99`](backend/src/meho_backplane/ui/static/src/app/modal-dialogs.js#L99) `dialog.showModal()` inside the delegated `htmx:afterSwap` listener (`:80-101`); identical at the v0.20.0 tag (ops line number exact).
- The abort confirm form swaps `hx-target="#runbook-run-step" hx-swap="outerHTML"` ([`_run_step.html:250-252`](backend/src/meho_backplane/ui/templates/runbooks/_run_step.html#L250-L252)); the run-driver dialogs are descendants of the swap target (container `:36`, abort `:242`, reassign `:287`) — the swap detaches them, and htmx 2.0.9 dispatches `afterSwap` with the pre-swap `detail.target`.
- Event ordering: `afterSwap` fires before `afterRequest`, so the abort dialog is still `open` at scan time and skipped by the `:90` guard — the closed reassign dialog throws. Advance forms render only for the run's assignee (`_run_step.html:174`), hence the assignee-manifestation.
- The named Deprecate/Delete/Publish flows use a different Alpine class-toggle pattern and do **not** share the defect. Controller predates v0.20.0 (#1803-era, v0.17.0; run driver #1930, v0.18.0). Nothing landed post-tag; dedupe clean.

## Desired state

One guard in the shared controller — extend the early-return at `modal-dialogs.js:82` to skip detached scan roots: `if (!root || typeof root.querySelectorAll !== "function" || !root.isConnected) { return; }` (`isConnected` is the modern form of the reporter's own Option A `document.contains`). Because it lives in the shared controller it fixes all manifestations at once (abort/reassign, Advance, account-sessions revoke) and is behavior-preserving for the module's design pattern (innerHTML container swaps always deliver a connected root). **Do not** re-point the scan at the new content/document instead — freshly swapped-in fragments legitimately ship their dialogs closed and must not auto-open. Reporter's Options B/C (hoist dialogs out of the swap zone / innerHTML-only swaps) are per-surface restructures made unnecessary by the guard.

## Acceptance criteria

- [ ] `modal-dialogs.js` carries the `isConnected` guard on the scan root, with a comment explaining htmx 2.0.9's pre-swap `detail.target` semantics for `outerHTML` swaps.
- [ ] Source-text assertion in `backend/tests/test_ui_templates.py` (house pattern, existing showModal wiring assertions ~`:348-384`) pins the guard's presence.
- [ ] Manual/browser verification on the abort flow (admin) — no uncaught `InvalidStateError` on the happy path; abort still succeeds. (Advance/account-sessions manifestations are code-predicted; spot-check Advance as the assignee.)
- [ ] No route/schema change → no OpenAPI snapshot delta.

## Out of scope

- Restructuring the run-driver swap targets or hoisting dialogs (Options B/C).
- The Alpine-pattern dialogs (Deprecate/Delete/Publish) — different mechanism, no defect.
- Upgrading vendored htmx (the pre-swap `detail.target` behavior is upstream semantics this guard accommodates).

## References

- Original report: v0.20.0 RDC test period, evoila-bosnia/claude-rdc-hetzner-dc#1790 — mechanism corrected in #2286's corrections table (no orphaned reference; pre-swap `detail.target` + closed sibling dialog).
- Provenance: shared controller #1803 (v0.17.0), run driver #1930 (v0.18.0); first noticed during the #2112 abort-403 re-verify (that fix confirmed in v0.20.0). Feature home Goal #336 closed — References only.
- Best-practices anchors — *frontend*: guard DOM operations on connectivity after swaps; never auto-open dialogs on fragment insertion; assert load-bearing JS invariants in template tests.

Parent goal: #221
Parent initiative: #2286

damir-topic:
## Summary

- htmx 2.0.9 dispatches `htmx:afterSwap` with `detail.target` still pointing at
  the **pre-swap** element; an `outerHTML` swap has already detached it. The
  shared app-shell modal controller (`app/modal-dialogs.js`) re-queries dialogs
  on every swap and scanned that stale, disconnected subtree, calling
  `showModal()` on the old closed descendant dialog — throwing an uncaught
  `InvalidStateError` into the console.
- The thrower on the abort repro is the admin-only **reassign** dialog (closed
  sibling; the still-`open` abort dialog is skipped by the existing `!dialog.open`
  guard). The same error is code-predicted for the run's **assignee on every
  Advance click**, and on account-sessions revoke with ≥2 sessions.
- Fix: a one-line `isConnected` guard on the `htmx:afterSwap` scan-root early
  return, so a detached root is skipped. Behaviour-preserving for the console's
  `innerHTML`-into-a-stable-container auto-open pattern (always a connected
  root); only the `outerHTML` self-detaching swap zones are newly skipped, and
  their freshly swapped-in replacements ship dialogs closed + button-driven.
- Console noise only — the abort / advance actions were always functional.

## Why `isConnected` (verified against the vendored htmx 2.0.9 source)

- `Me()` (the `outerHTML` swap) filters the old target out of `settleInfo.elts`,
  inserts the new nodes, then `t.remove()` **detaches** the old target.
- `afterSwap` is dispatched on the new elements but carries a shared `detail`
  (`eventInfo`); `responseInfo.target` (`= detail.target`) is pinned to the
  **pre-swap** resolved target and is never reassigned post-swap.
- So `detail.target` is the detached `#runbook-run-step`. A detached element
  still exposes `querySelectorAll`, which is exactly why the existing capability
  check missed it; `isConnected` is the precise still-attached test (the modern
  form of the report's Option A `document.contains`).

## Test plan

- [x] `uv run pytest tests/test_ui_templates.py` — 36 passed (adds
  `test_modal_controller_script_guards_detached_swap_target`, a source-text
  assertion pinning the guard, matching the existing `#1803` showModal wiring
  asserts).
- [x] `uv run mypy src/` — clean (592 files).
- [x] `uv run ruff check src/ tests/` + `ruff format --check` — clean.
- [x] `.claude/hooks/code-quality.py --diff` — pass.
- [x] No route/model touched → **no OpenAPI snapshot delta** (AC #4).
- [ ] **Manual/browser spot-check (AC #3 — needs a running console + tenant_admin + an active run):**
  1. As a tenant_admin, open a runbook run; open the DevTools console.
  2. Click **Abort run**, enter a reason, submit. Expect: the step re-renders
     (`#runbook-run-step` `outerHTML` swap), the abort modal closes, and **no
     `InvalidStateError`** is logged (previously the closed reassign sibling
     threw).
  3. As the run's **assignee**, click **Advance**. Expect the step to advance
     with **no `InvalidStateError`** (the code-predicted steady-state noise).

## Out of scope

- Restructuring run-driver swap targets / hoisting dialogs (report Options B/C).
- The Alpine-pattern dialogs (Deprecate / Delete / Publish) — different
  mechanism, no defect.
- Upgrading vendored htmx (the pre-swap `detail.target` behaviour is upstream
  semantics this guard accommodates).

Closes #2242

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,22 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Fixed — shared modal controller ignores htmx's detached pre-swap target
+
+- **The app-shell modal controller no longer calls ``showModal()`` on a
+  disconnected dialog after an ``outerHTML`` swap** (#2242). htmx 2.0.9
+  dispatches ``htmx:afterSwap`` with ``detail.target`` still pointing at the
+  PRE-swap element, which an ``outerHTML`` swap already detached — the runbook
+  run driver's abort / reassign / advance forms target ``#runbook-run-step``
+  with ``hx-swap="outerHTML"``. The shared controller scanned that stale
+  subtree and reopened the old, closed descendant dialog (the admin-only
+  reassign dialog on the abort repro; the assignee's advance dialog on every
+  Advance click), throwing an uncaught ``InvalidStateError`` into the console.
+  A one-line ``isConnected`` guard on the scan root skips a detached root and
+  is behaviour-preserving for the console's ``innerHTML``-into-a-stable-
+  container auto-open pattern, which always delivers a connected root. Console
+  noise only; the abort / advance actions were always functional.
+
 ### Fixed — exactly-one-resumer claim for run-bound approvals (#2293)
 
 - **A run-bound approval now executes its gated op exactly once — no silent
--- a/backend/src/meho_backplane/ui/static/src/app/modal-dialogs.js
+++ b/backend/src/meho_backplane/ui/static/src/app/modal-dialogs.js
@@ -79,7 +79,24 @@
   // descendant of a container target and we must not double-drive it.
   document.body.addEventListener("htmx:afterSwap", function (event) {
     var root = (event.detail && event.detail.target) || event.target;
-    if (!root || typeof root.querySelectorAll !== "function") {
+    // htmx 2.0.9 dispatches ``htmx:afterSwap`` with ``detail.target`` still
+    // referencing the PRE-swap element. An ``outerHTML`` swap (e.g. the
+    // runbook run driver's abort / reassign / advance forms, which target
+    // ``#runbook-run-step`` with ``hx-swap="outerHTML"``) has already replaced
+    // and DETACHED that element, so ``detail.target`` is now disconnected from
+    // the document. Scanning its stale subtree would find the old, closed
+    // descendant dialogs and call ``showModal()`` on a dialog that is no longer
+    // connected -- which throws ``InvalidStateError``. Bail on a detached root:
+    // the freshly swapped-in replacement ships its own dialogs closed and
+    // button-driven, so there is nothing here to auto-open. This handler's
+    // auto-open pattern always swaps modal fragments via ``innerHTML`` into a
+    // STABLE container, whose ``detail.target`` stays connected, so the
+    // ``isConnected`` check is a no-op for it.
+    if (
+      !root ||
+      typeof root.querySelectorAll !== "function" ||
+      !root.isConnected
+    ) {
       return;
     }
     var dialogs = root.querySelectorAll("dialog.modal");
--- a/backend/tests/test_ui_templates.py
+++ b/backend/tests/test_ui_templates.py
@@ -382,6 +382,31 @@ def test_modal_controller_script_opens_injected_dialog_on_swap() -> None:
     )
 
 
+def test_modal_controller_script_guards_detached_swap_target() -> None:
+    """The controller skips a detached scan root (htmx 2.0.9 pre-swap target).
+
+    ``htmx:afterSwap`` fires with ``detail.target`` still pointing at the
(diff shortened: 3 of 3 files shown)


damir-topic:
@coderabbitai review

Issue #2240, opened 2026-07-09
Task: first-class chart value for MEHO_TARGET_SSRF_ALLOWLIST — values key + schema + ConfigMap render ('action likely required' release note)


zdamir:
## Summary

v0.20.0 ships the default-deny target-destination SSRF guard (PR #2191) whose only opt-out is the `MEHO_TARGET_SSRF_ALLOWLIST` env var, and the release notes say **"Deployment impact (action likely required)"** — but the Helm chart has zero surface for it: no values key, no schema property, no ConfigMap line. Operators must use the untyped `extraEnv` escape hatch, which the schema doesn't validate and `helm show values` doesn't surface. The ops filing is fully accurate (symptoms, root cause, and scope — rare); fix is a one-commit chart PR.

## Current state (verified in code + executed probe)

- Guard: [`targets/ssrf_guard.py:105`](backend/src/meho_backplane/targets/ssrf_guard.py#L105) defines `TARGET_SSRF_ALLOWLIST_ENV`, read **per call** from `os.environ` (`:141`; deliberately NOT in `settings.py` — docstring `:98-104`). Create/update arm → structured 422; connect-time re-screen → `SsrfBlockedError` → structured `connector_error`, non-retryable. In-tag (commit 0785d1ff).
- CHANGELOG v0.20.0 (`:486-514`): *"**Deployment impact (action likely required):** MEHO deployments register on-prem appliances on RFC 1918 space as a matter of course — set the new `MEHO_TARGET_SSRF_ALLOWLIST` env var…"*.
- Chart (`deploy/charts/meho/`): `grep -ri ssrf deploy/` → nothing. `values.schema.json` has `additionalProperties: false` on the `config` block (line 185), so the key **must** be added to the schema to be usable at all; `forwardedAllowIps` (schema `:260`, values `:284`, configmap `:107`) is the natural adjacency. `extraEnv` is free-form array-of-objects (schema `:796-800`) — the reporter's deploy already uses it for `SSL_CERT_FILE`.
- Probe: `_parse_allowlist()` with the env var set to `""` returns empty structures — an unconditional empty ConfigMap render is a genuine no-op (guard stays fully on).
[...]

PR #1323, opened 2026-05-28
fix(connectors): G0.16-T1 — async ingest must not crash pod on large specs (#1303)

damir-topic:
## Summary

- `POST /api/v1/connectors/ingest` defaults to `async=true` and returns `202 Accepted` + a job handle for the non-dry-run path; the pipeline runs off the request thread via `asyncio.create_task` so a real-world vendor spec (7+ MB / 1000+ ops) no longer crashes the backplane pod under the kubelet liveness probe.
- New `GET /api/v1/connectors/ingest/jobs/{job_id}` route polls an in-memory `IngestJobRegistry` for completion; tenant-isolation gate mirrors `ReviewService`'s 404 conflation.
- `dry_run=true` keeps the synchronous shape (parse-only fast path); `async=false` keeps the legacy blocking 200 + `IngestResponse` for small-spec callers (CI tests, ad-hoc scripts, v0.8.x clients).
- Reframes the `composite_l2_missing` envelope per `docs/codebase/api-shape-conventions.md` §1: the catalog command is no longer "the remediation path" (operators followed it into the pod-restart loop in v0.8.0) — the human message names the curation gap first and the L1-wrapper request as the recommended path, with the `catalog_command` retained as the OpenAPI escape-hatch recipe. The structured `extras` payload is unchanged.
- `parse_openapi` calls inside the pipeline wrapped in `asyncio.to_thread` so the synchronous YAML walk doesn't block the event loop while the response is being sent back.
- Docs: `docs/codebase/spec-ingestion.md` gains an "Async ingest mode" section; `docs/codebase/api-shape-conventions.md` §1 references the closed escape-hatch-must-not-crash item; CHANGELOG `[Unreleased]` gains two bullets.

## Test plan

- [x] `ruff check` / `ruff format --check` — clean
- [x] `mypy` on changed paths — clean
- [x] `python3 .claude/hooks/code-quality.py --diff` — 0 blockers
- [x] `pytest backend/tests/test_api_v1_connectors_ingest.py` — 57 passed (3 new G0.16-T1 tests)
- [x] `pytest backend/tests/test_connectors_vmware_rest_composites_l2_preflight.py` — 7 passed (reworded envelope test)
- [x] Wider suite (`pytest backend/tests/ --ignore=integration --ignore=acceptance`) — 5682 passed, 41 skipped, 0 failed
[...]

PR #2317, opened 2026-07-10
fix(ingest): watchdog async ingest jobs to a terminal state + bound LLM client + job-id log binding (#2275)

### LINKED ISSUE #2275: Task: watchdog the async ingest job + explicit LLM-client timeout + job-id log binding — a job must always reach a terminal state
## Summary

An async ingest job can sit at `status=running` forever (observed 10+ min on v0.20.0, surviving until pod restart clears the in-memory registry) while the identical sync request returns a clean 400. The symptom is confirmed structurally — **but the originally-proposed fix ("widen the try/except") is a no-op**: the boundary is already `except BaseException` over the entire pipeline call, and an `OpIdCollision`-raising pipeline provably flips the job to `failed` (probe + existing unit test), exactly like the #2272 datetime crash did. The only mechanism that matches the observation is a **never-completing await** inside the pipeline with **no watchdog anywhere**. Fix: time-box the job body so a job always reaches a terminal state, pin explicit timeouts on the ingest LLM client, and bind `ingest_job_id` into pipeline log events so the next diagnosis isn't blind.

## Current state (verified in code + probe, v0.20.0 == main for the ingest tree)

- Job registry is a process-local `OrderedDict` (docstring: "a pod restart blows the in-memory state away on purpose", [`jobs.py:20-26`](backend/src/meho_backplane/operations/ingest/jobs.py#L20-L26)); **running jobs are exempt from eviction** (`:336-348`); `run_ingest_job` awaits `pipeline_call()` with **no `asyncio.wait_for` anywhere** (`:394-493`).
- The exception boundary is maximal: `except BaseException` → `ingest_job_failed` log → `registry.fail(...)` ([`jobs.py:451-465`](backend/src/meho_backplane/operations/ingest/jobs.py#L451-L465)). Both `OpIdCollision` raise sites run inside it; catalog resolution happens synchronously at the route **before** any job row exists (`connectors_ingest.py:482-483`) and 422s — it cannot strand a job. Probe on this checkout: an `OpIdCollision`-raising pipeline → `status=failed`, `error_class='OpIdCollision'` (see also `test_operations_ingest_jobs.py:213`).
- Wedge candidates, in execution order, with file:line: `asyncio.to_thread(parse_openapi)` on a starved executor ([`pipeline.py:899`](backend/src/meho_backplane/operations/ingest/pipeline.py#L899)); DB session/connection acquisition ([`_upsert.py:144/:262-273`](backend/src/meho_backplane/operations/ingest/_upsert.py#L144)); lazy fastembed model load behind a `threading.Lock` in a to_thread worker ([`retrieval/embedding.py:168/:327`](backend/src/meho_backplane/retrieval/embedding.py#L168)); and the ingest-grouping `AsyncAnthropic(api_key=api_key)` constructed with **no explicit `timeout`/`max_retries`** ([`anthropic_client.py:198`](backend/src/meho_backplane/operations/ingest/anthropic_client.py#L198)) — SDK defaults are a 10-minute request timeout × `max_retries=2` with timeouts retried, so wall-clock can reach ~30 minutes pending, the only candidate that inherently spans "10+ minutes with the event loop responsive". Which one fired on the RDC pod is unrecoverable post-restart.
- "Zero pipeline log events" is largely an observability artifact: pipeline events bind `connector_id` but never `ingest_job_id` ([`pipeline.py:627-633`](backend/src/meho_backplane/operations/ingest/pipeline.py#L627-L633)); `jobs.py:450` binds the job id on jobs.py's **own logger instance** only, not via `structlog.contextvars` (despite `merge_contextvars` being configured, `logging.py:104`) — job-id-filtered greps structurally exclude pipeline events.

## Desired state

1. **Watchdog:** wrap the job body in `asyncio.wait_for(..., timeout=_INGEST_JOB_TIMEOUT_S)` (new module constant near `_MAX_JOBS_RETAINED`, e.g. 30 min) covering the pipeline call **and** the post-`try` dispatchability leg (`jobs.py:539` runs a real DB probe whose hang would still strand the job if only `pipeline_call` were wrapped). `TimeoutError` lands in the existing `except BaseException` → `failed` with `error_class='TimeoutError'`. Guarantee stated precisely: *the job leg always terminates*; note that cancellation cannot interrupt an already-running `to_thread` worker — the job status still flips, the thread may linger.
2. **LLM client bounds:** construct the grouping `AsyncAnthropic` with explicit `timeout=` and `max_retries=` (values ≪ the watchdog; e.g. 120 s × 1 retry) so grouping cannot pend ~30 min on SDK defaults.
3. **Log binding:** bind `ingest_job_id` via `structlog.contextvars.bind_contextvars` at job start so pipeline events carry the job id.

## Acceptance criteria

- [ ] Unit: a never-returning `pipeline_call` (`asyncio.Event().wait()`-shaped) flips the job to `failed` with `error_class='TimeoutError'` within the (test-shrunk) watchdog budget — add beside `test_raising_pipeline_still_fails` in `backend/tests/test_operations_ingest_jobs.py` (no such wedged-pipeline test exists today).
- [ ] Regression pin of the refuted mechanism: an `OpIdCollision` raised through the async path yields `status=failed` with the structured collision error on the job record (this passes today — pin it so the "async swallows collisions" premise stays dead).
- [ ] A hang in the dispatchability leg also terminates the job (covered by wrapping the whole body, or an explicit second test).
- [ ] The ingest-grouping Anthropic client is constructed with explicit `timeout` and `max_retries` (assert constructor kwargs in the existing anthropic_client tests).
- [ ] With the binding in place, a job-id-filtered log capture during a pipeline run includes `ingestion_pipeline_start` (asserts the contextvars binding).
- [ ] No route/schema change expected → no OpenAPI snapshot delta (verify).

## Out of scope

- Persisting the job registry to the DB (in-memory-by-design per #1323; restart-loss is documented behavior).
- Interrupting wedged `to_thread` workers (Python thread cancellation is not available; the watchdog guarantees job-state terminality, not thread reclamation).
- Root-causing which await wedged on the RDC pod — unrecoverable; a future wedge is diagnosable via the new job-id-bound pipeline events and a `py-spy` dump against the named `ingest-job-<uuid>` task.

## References

- Original report: v0.20.0 RDC test period, consumer ticket evoila-bosnia/claude-rdc-hetzner-dc#1790. Contrast case: #2272 (clean async failure through the same boundary). Prior art re-checked: #1647 (async false-success → dispatchability gate), #2208/#2214/PR #2213 (authorize-before-job-row), #1323 (job registry), #1677 (degraded state).
- Anthropic Python SDK client defaults (10-min request timeout, `max_retries=2`, timeouts retried → wall-clock ≈ timeout × 3): https://github.com/anthropics/anthropic-sdk-python (Client Configuration).
- Best-practices anchors — *python*: every background job needs a terminal-state guarantee (watchdog), not just an exception handler; explicit timeouts on outbound clients; bind correlation ids via contextvars so log filtering doesn't lie. *ai*: bound LLM-call latency/retries explicitly in pipelines.

Parent goal: #221
Parent initiative: #2286

damir-topic:
## Summary

- **Watchdog: an async ingest job now always reaches a terminal state.** A wedged pipeline could sit at `status=running` forever (surviving until a pod restart cleared the in-memory registry) while the identical sync request returned a clean 400. Per the issue's re-diagnosis, the `except BaseException` boundary was already maximal — an `OpIdCollision`-raising pipeline provably flips the job to `failed` — so the earlier "widen the try/except" framing was a no-op; the real defect was a **never-completing await with no watchdog**. `run_ingest_job` now time-boxes the whole job body (the pipeline call **and** the post-run dispatchability probe, whose own hang would otherwise strand the job) inside `asyncio.timeout`; at the deadline the job flips to `failed` with `error_class="TimeoutError"`. Budget defaults to 30 min, env-overridable via `INGEST_JOB_TIMEOUT_SECONDS` (parsed defensively — a malformed value falls back with a warning rather than crashing the package at import).
- **Explicit LLM-client bounds.** The grouping `AsyncAnthropic` is now constructed with an explicit `timeout` (120 s) + `max_retries` (1) instead of the SDK defaults, whose 10-min read timeout with retried requests let a hung grouping call pend ~30 min — the one wedge candidate that inherently spans "10+ minutes with the event loop responsive". The fail-closed 503 contract for a missing `ANTHROPIC_API_KEY` (#1386) is unchanged.
- **Job-id log binding via contextvars.** `ingest_job_id` is now bound through `structlog.contextvars` (not just the job coroutine's own logger), so the configured `merge_contextvars` processor stamps it onto pipeline events (`ingestion_pipeline_start` &c.). A job-id-filtered log grep is no longer structurally blind to the pipeline — the "zero pipeline log events" observation was this artifact.

`run_ingest_job`'s succeeded/degraded reconciliation was extracted into `_reconcile_returned_result` to stay within the function-size budget; behavior is unchanged.

## Test plan

- [x] `mypy` — clean
- [x] AC1 — a never-returning `pipeline_call` flips to `failed` / `TimeoutError` within a test-shrunk budget: `test_wedged_pipeline_times_out_to_failed`
- [x] AC2 — an `OpIdCollision` through the async path yields `status=failed` with the structured collision error (regression pin; the "async swallows collisions" premise stays dead): existing `test_op_id_collision_job_error_names_remediation`
- [x] AC3 — a hang in the dispatchability leg also terminates the job (`CancelledError` is `BaseException`, so the probe's fail-open `except Exception` does not swallow it): `test_wedged_dispatchability_probe_times_out_to_failed`
- [x] AC4 — the grouping Anthropic client is constructed with explicit `timeout` + `max_retries`: `test_factory_constructs_client_with_explicit_timeout_and_retries`
- [x] AC5 — a job-id-filtered log capture during a pipeline run includes `ingestion_pipeline_start` carrying `ingest_job_id`: `test_pipeline_log_event_carries_ingest_job_id`
- [x] AC6 — no route/schema change → **OpenAPI snapshot regenerated with zero delta**
- [x] Broader sweep — `pytest -k "ingest or connector_ingest"`: 601 passed, 11 skipped

## Out of scope

- **Helm chart value for `INGEST_JOB_TIMEOUT_SECONDS`.** The watchdog budget is env-configurable and may want a chart surface; sibling **#2240** is doing a chart-value task separately, so chart files are deliberately untouched here.
- Persisting the job registry across restarts (in-memory-by-design, #1323); interrupting wedged `to_thread` workers (Python has no thread cancellation — the guarantee is job-state terminality, not thread reclamation); root-causing which await wedged on the RDC pod (unrecoverable post-restart — a future wedge is now diagnosable via the job-id-bound pipeline events + a `py-spy` dump against the named `ingest-job-<uuid>` task).
- **Adjacent finding (recorded, not addressed):** `ingest/jobs.py` is now 655 lines (soft file-size warning, >600); a responsibility-based module split is a larger refactor beyond this task's scope.

Closes #2275

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,36 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Fixed — async ingest jobs always reach a terminal state (watchdog + bounded LLM client + job-id log binding) (#2275)
+
+- **A wedged async connector ingest can no longer sit at `status=running`
+  forever** (#2275). An ingest job whose pipeline hit a *never-completing
+  await* — a starved `to_thread` executor, a DB connection that never
+  acquires, or a grouping LLM call pending on the Anthropic SDK's default
+  10-min-read × 2-retry ceiling (~30 min wall-clock) — was stranded at
+  `running` until a pod restart cleared the in-memory registry, even
+  though the identical *sync* request returned a clean 400. (The
+  exception boundary was already maximal — an `OpIdCollision`-raising
+  pipeline does flip to `failed` — so the earlier "widen the try/except"
+  framing was a no-op; the real defect was the missing watchdog.)
+  `run_ingest_job` now time-boxes the whole job body — the pipeline call
+  **and** the post-run dispatchability probe — inside
+  `asyncio.timeout`; at the deadline the job flips to `failed` with
+  `error_class="TimeoutError"`. The budget defaults to 30 min and is
+  env-overridable via `INGEST_JOB_TIMEOUT_SECONDS`. Cancellation cannot
+  interrupt an already-running `to_thread` OS thread, so the guarantee is
+  job-state terminality, not thread reclamation.
+- **The grouping LLM client is now constructed with an explicit request
+  timeout + retry ceiling** (120 s × 1 retry) instead of the SDK
+  defaults, so a hung grouping call fails fast rather than consuming the
+  watchdog budget; the fail-closed 503 contract for a missing
+  `ANTHROPIC_API_KEY` (#1386) is unchanged.
+- **Pipeline log events now carry `ingest_job_id`.** The job id is bound
+  via structlog `contextvars` (not just the job coroutine's own logger),
+  so the configured `merge_contextvars` processor stamps it onto pipeline
+  events (`ingestion_pipeline_start` &c.) — a job-id-filtered log grep is
+  no longer blind to the pipeline. No schema or route change. (#2275)
+
 ### Changed — scheduler `fire_at` tick-quantization latency documented on the API schema (#2245)
 
 - **The scheduler's fire-time latency window is now part of the API contract,
--- a/backend/src/meho_backplane/operations/ingest/anthropic_client.py
+++ b/backend/src/meho_backplane/operations/ingest/anthropic_client.py
@@ -40,7 +40,7 @@ tenant + egress story the build-time grouping pass does not have today).
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Final
 
 import structlog
 
@@ -59,6 +59,23 @@ __all__ = [
 
 _log = structlog.get_logger(__name__)
 
+#: Per-request transport timeout (seconds) for the grouping-pass client.
+#: The Anthropic SDK's default is a 10-min *read* timeout, and with the
+#: default ``max_retries`` a timed-out request is retried, so a hung
+#: grouping call can pend on the order of ~30 min wall-clock -- long
+#: enough to outlive the ingest-job watchdog's default budget and the
+#: single wedge candidate that inherently spans "10+ minutes with the
+#: event loop responsive". Pin it well below the watchdog so a stuck LLM
+#: call fails the grouping request (surfaced as the job's failure) rather
+#: than silently consuming the whole watchdog budget.
+_INGEST_LLM_TIMEOUT_SECONDS: Final[float] = 120.0
+
+#: Retry ceiling for the grouping-pass client. One retry (vs the SDK
+#: default of 2) keeps a transient 429/5xx recoverable while bounding the
+#: worst-case wall-clock at ``_INGEST_LLM_TIMEOUT_SECONDS * (1 + retries)``
+#: rather than the SDK default's ~30 min.
+_INGEST_LLM_MAX_RETRIES: Final[int] = 1
+
 
 class AnthropicMessagesLlmClient:
     """:class:`LlmClient` backed by the Anthropic Messages API.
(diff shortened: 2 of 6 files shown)


damir-topic:
@coderabbitai review

PR #2321, opened 2026-07-10
fix(ingest): reject non-finite INGEST_JOB_TIMEOUT_SECONDS + expose watchdog budget as a chart value (#2318)

### LINKED ISSUE #2318: Task: reject non-finite INGEST_JOB_TIMEOUT_SECONDS + expose the ingest-job watchdog budget as a chart value
## Summary

Two small hardening follow-ups to the async ingest-job watchdog shipped in #2275 (PR #2317), both surfaced during that PR's review:

1. **Correctness:** `INGEST_JOB_TIMEOUT_SECONDS=inf` (or `nan`) silently **defeats the watchdog** — the exact "job wedged at `status=running` forever" failure #2275 was built to prevent — because the env loader's only sanity check is `value <= 0`, which a non-finite float passes.
2. **Ops surface:** the watchdog budget has **no first-class Helm chart value** (unlike its sibling `MEHO_TARGET_SSRF_ALLOWLIST`, given one in #2240), so an operator who needs a longer ceiling for a slow shared executor / large spec fleet must reach for the untyped `extraEnv` escape hatch.

## Current state (verified in code on `main`)

- Loader `_load_ingest_job_timeout_seconds` ([`jobs.py:88-118`](backend/src/meho_backplane/operations/ingest/jobs.py#L88-L118)) parses `value = float(raw)` (`:102`) then guards only `if value <= 0:` (`:110`). `float("inf")` / `float("nan")` raise no `ValueError`, and both pass `<= 0` (`inf <= 0` is `False`; every `nan` comparison is `False`), so the non-finite value is returned verbatim as the budget.
- `run_ingest_job` time-boxes the job body in `asyncio.timeout(_INGEST_JOB_TIMEOUT_SECONDS)` — `asyncio.timeout(inf)` never schedules a deadline and `asyncio.timeout(nan)` is ill-defined, so a misconfigured budget re-opens the unbounded-wait hole #2275 closed.
- No chart surface: `git grep INGEST_JOB_TIMEOUT deploy/charts/**` → nothing. The `config.*` → ConfigMap env pattern for numeric `*_SECONDS` knobs already exists (e.g. `VAULT_TIMEOUT_SECONDS` ← `config.vaultTimeoutSeconds`, [`configmap.yaml:62`](deploy/charts/meho/templates/configmap.yaml#L62)).

## Desired state

1. `_load_ingest_job_timeout_seconds` rejects non-finite values — change the guard to `if not math.isfinite(value) or value <= 0:` (add `import math`), log a warning, fall back to the 1800 s default. A misconfigured budget can never disable the watchdog.
2. A first-class `config.ingestJobTimeoutSeconds` chart value that renders `INGEST_JOB_TIMEOUT_SECONDS` on the backplane ConfigMap, mirroring the `config.vaultTimeoutSeconds` precedent (schema-typed, documented). **The default render must not introduce a boot-time warning** — render conditionally (omit the env when the value is empty, so the backend's unset→default path runs silently) OR default to a valid number. Do **not** copy #2240's always-render-`""`: an empty string hits the `float("")` `ValueError` fallback and logs `ingest_job_timeout_env_invalid` on every boot.

## Acceptance criteria

- [ ] `INGEST_JOB_TIMEOUT_SECONDS=inf` and `=nan` both fall back to the 1800 s default with a warning (unit test on `_load_ingest_job_timeout_seconds`); a finite positive value is still honored; `<= 0` and malformed inputs still fall back.
- [ ] A test proves the watchdog still fires (a wedged job flips to `failed`) when the env is set to `inf` — i.e. the sanitized fallback budget is what `asyncio.timeout` receives.
- [ ] `config.ingestJobTimeoutSeconds` exists in `values.yaml` + `values.schema.json`; `helm template` with a set value renders `INGEST_JOB_TIMEOUT_SECONDS` on the ConfigMap and the backplane consumes it via `envFrom`; the **default** render introduces no boot-time `float()` warning (conditional render or numeric default).
- [ ] `helm lint` + kubeconform clean; no OpenAPI snapshot delta (backend-internal + chart only).
- [ ] CHANGELOG `[Unreleased]` entry under a unique header; docs updated (`docs/codebase/spec-ingestion.md` watchdog note + the ConfigMap env catalogue in `docs/codebase/devops.md`).

## Out of scope

- The watchdog cancellation mechanism itself (#2275 is correct — `asyncio.timeout` + `CancelledError`/`BaseException` handling was verified during PR #2317 review).
- Splitting `jobs.py` (soft >600-line warning; disclosed in #2317, separate concern).

## References

- Parent goal: #221 (G0 Foundational substrate). Parent initiative: #2286 (G0.30 v0.20.0 dogfood hardening).
- Hardens #2275 / PR #2317 (async ingest-job watchdog).
- Chart-value precedent: #2240 / PR #2316 (`MEHO_TARGET_SSRF_ALLOWLIST`).

Parent goal: #221
Parent initiative: #2286

damir-topic:
## Summary

Two small hardening follow-ups to the #2275 async ingest-job watchdog, both surfaced during PR #2317's review:

- **Correctness:** `INGEST_JOB_TIMEOUT_SECONDS=inf` (or `nan`) silently defeated the watchdog. The loader's only check was `value <= 0`, which a non-finite float passes (`inf <= 0` is `False`; every `nan` comparison is `False`), so it flowed into `asyncio.timeout(...)` — where `inf` schedules no deadline and `nan` is ill-defined — re-opening the "wedged at `status=running` forever" hole #2275 closed. `_load_ingest_job_timeout_seconds` now rejects non-finite values with `math.isfinite`, logs a warning, and falls back to the 1800 s default. The parsed-but-unusable warning is renamed `ingest_job_timeout_env_out_of_range` (covers both non-finite and non-positive); the un-parseable-string case keeps `ingest_job_timeout_env_invalid`.
- **Ops surface:** the watchdog budget gains a first-class Helm value `config.ingestJobTimeoutSeconds`, rendering `INGEST_JOB_TIMEOUT_SECONDS` on the backplane ConfigMap (consumed via the existing `envFrom.configMapRef`), so operators no longer reach for the untyped `extraEnv` escape hatch.

### Chart render decision — conditional render, not always-render (justification)

I render `INGEST_JOB_TIMEOUT_SECONDS` **only when the value is non-empty** (`{{- if .Values.config.ingestJobTimeoutSeconds }}`) with an empty `""` default — deliberately **not** #2240's always-render-`""` pattern. Rationale:

- The backend's `_DEFAULT_INGEST_JOB_TIMEOUT_SECONDS = 1800.0` stays the **single source of truth** for the default; the chart value is a pure optional override, so a future backend-default change can't be masked by a stale chart literal.
- The default render omits the key entirely → the backend's unset→default path runs silently, with **zero** boot-time `float()` warning. An always-rendered `""` would hit `float("")` → `ValueError` → a spurious `ingest_job_timeout_env_invalid` on every boot — the exact trap the issue calls out. The sibling `targetSsrfAllowlist` gets away with always-render-`""` only because its empty value is a valid no-op string; a numeric env can't.

## Test plan

- [x] `ruff check` + `ruff format --check src/ tests/` — clean
- [x] `mypy src/` — Success, no issues (592 files)
- [x] `pytest tests/test_operations_ingest_jobs.py` — 23 passed (12 new: loader unset / finite-positive-honored / rejected-values ×9 [`inf`, `Infinity`, `-inf`, `nan`, `0`, `0.0`, `-30`, `abc`, `""`] + `inf`-sanitized-so-watchdog-still-fires)
- [x] full unit sweep (`pytest -n 3 --dist loadscope --ignore=integration --ignore=migrations tests/`) — **9351 passed, 189 skipped** (exit 0, 13m12s)
- [x] `helm lint` (CI override set) — 1 chart linted, 0 failed
- [x] `helm template` **default** render → **no** `INGEST_JOB_TIMEOUT_SECONDS` key (no boot-time `float()` warning)
- [x] `helm template --set-string config.ingestJobTimeoutSeconds=3600` → `INGEST_JOB_TIMEOUT_SECONDS: "3600"` on the `<release>-config` ConfigMap the Deployment `envFrom`s (decimal `7200.5` also renders)
- [x] `helm template --set config.ingestJobTimeoutSeconds=abc` → schema rejects: `'abc' does not match pattern '^(|[0-9]+(\.[0-9]+)?)$'`
- [x] `make snapshot-openapi` → **zero** OpenAPI delta (backend-internal + chart only)
- [x] `values.schema.json` valid JSON; `config.ingestJobTimeoutSeconds` typed (optional string, empty-or-decimal), absent from `config.required`

> Note: kubeconform is a CI-only lane (not on the local PATH). The change only adds a string key to an existing ConfigMap `data` map, which is inherently kubeconform-valid; the default render adds no key at all.

## Docs / CHANGELOG

- CHANGELOG `[Unreleased]` — new `### Fixed —` entry scoped to #2318 (two bullets: the non-finite guard + the chart value).
- `docs/codebase/spec-ingestion.md` — watchdog note updated (non-finite guard is load-bearing + chart value).
- `docs/codebase/devops.md` — ConfigMap env-catalogue bullet + operator-overrides table row.

## Out of scope

- The watchdog cancellation mechanism itself (#2275 verified correct during PR #2317 review).
- Splitting `jobs.py` (soft >600-line warning; disclosed in #2317, separate concern).

Closes #2318


damir-topic:
@coderabbitai review

PR #2368, opened 2026-07-11
feat(scheduler): project skip-state onto trigger row + park after N unresolvable skips (#2327)

### LINKED ISSUE #2327: Task: project scheduler skip-state onto the trigger row + park after N unresolvable skips — no more silent every-tick skips
## Summary

When `_prepare_invocation` cannot resolve a due trigger (definition missing / disabled / credentials unresolved), the scheduler loop skips the tick **without advancing state**: the row stays `active`, `next_fire_at` unchanged, `last_fired_at` stays null. Skip-and-retry is the right call for a transient miss — but a **permanent** miss (expired scheduler Vault token, never-persisted agent secret, deleted-but-still-referenced definition) produces an infinite silent loop whose only trace is a WARN pair in the pod log every 30s tick.

Nothing reaches any surface an operator actually reads:

- **No audit row** — despite `in_flight_policy: fail_into_audit`, which only covers fires that *launch*.
- **No field on the trigger row** — `meho scheduler list` shows a healthy-looking `active` trigger with no `last_error`, `last_skip_reason`, `last_skipped_at`, or `skip_count`.
- **No `/ready` degradation** — the pod stays green.
- **No broadcast** — nothing surfaces to the audit stream.

The consumer only finds out by absence.

Observed on `v0.20.0`.

## Reproduction / Observed

Any of the following will reproduce the silent-skip loop:

1. Point a trigger at an `agent_definition_id` that doesn't exist (or delete a definition while a trigger references it).
2. Point a trigger at credentials that resolve via Vault and expire (or revoke) the Vault token the scheduler uses.
3. Point a trigger at an agent whose secret path has never been created.

Then observe the trigger row via MCP:

```bash
curl -ksS -X POST https://<backplane>/mcp \
  -H "Authorization: Bearer $JWT" \
  -H "Content-Type: application/json" \
  -H "Accept: application/json" \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/call",
       "params":{"name":"meho.scheduler.list","arguments":{"limit":5}}}' \
  | jq '.result.content[0].text | fromjson | .triggers[0] | keys'
```

Observed on `v0.20.0`, the trigger row keys are:

```
agent_definition_id, created_at, created_by_sub, cron_expr, event_filter,
fire_at, id, identity_sub, in_flight_policy, inputs, kind, last_fired_at,
next_fire_at, status, tenant_id, timezone, updated_at, work_ref
```

No `last_error`, no `last_skip_reason`, no `last_skipped_at`, no `skip_count`, no `failure_reason`.

Meanwhile the pod log carries, every 30 s:

```
WARN scheduler_vault_read_failed  path=... status_code=403
WARN scheduler_credentials_unresolved trigger_id=...
```

**Lived consequence on a real deploy**: an hourly cron accumulated ~360 missed fires over 15 days before the operator noticed. A second operator hit the same wall on a separately-created trigger, concluded scheduling didn't work, and fell back to direct runs.

## Why this matters

- **Silent scheduler failures compound.** A cron that quietly stops firing is functionally the same as never having existed, but nobody notices for weeks — until the downstream artifact (log summary, backup, cleanup task) is conspicuously stale.
- **Cross-operator confusion.** A second operator hitting the same wall on a separate trigger drew the reasonable conclusion "scheduling doesn't work" and stopped using the feature. The absence of any state on the trigger row makes this indistinguishable from a broken product.
- **`in_flight_policy: fail_into_audit` promises exactly this.** The mental model the flag establishes is "if a fire fails, I'll see it in the audit ledger." Pre-launch failures fall outside that promise silently.
- **Belongs in the same visibility family as `event-kind` triggers accepted-but-never-fires** — two independent code paths reach "trigger looks healthy, does nothing"; both need honest failure modes to be trustworthy.

- **Project skip state onto the trigger row.** Add nullable columns `last_skip_reason` (text), `last_skipped_at` (timestamp), `skip_count` (integer, monotonic). Update from the loop on each skip; surface on `meho.scheduler.list` / `scheduler.show` / the scheduler UI. This mirrors the read-surface pattern used elsewhere to make in-progress state visible.
- **Park after N consecutive unresolvable skips.** Reuse the existing `_park_trigger` path (the corrupted-cron precedent) so the state machine itself communicates "this is broken and I've stopped trying." Combines well with the field above (the parked state comes with a reason).
- **Emit an audit row on first-skip-per-cause.** Extend the `fail_into_audit` promise to pre-launch failures — one row per distinct cause per trigger per some cooldown window, keyed on `(trigger_id, cause_hash)` to prevent per-tick spam. Consumers who already watch the audit stream get a natural surface.
- **Any of the above alone is a big improvement.** Ideally the trigger-row field (immediate visibility) + park after N (state-machine correctness); the audit row is a nice-to-have on top.

## Verification

- Create a deliberately unresolvable trigger (e.g. `agent_definition_id` pointing at nothing).
- After the next tick, `meho scheduler list` / `.show` on that trigger reports at least one of:
  - A non-null `last_skip_reason` / `last_skipped_at`, OR
  - Status transitioned to `parked` (or the deploy's equivalent terminal-error state), OR
  - An audit row exists explaining the skip.
- The pod log WARNs are still fine, but the operator-facing surface now agrees with them.

Suggested source-side probe (post-fix): the `scheduler.list` projection includes at least one of `last_skip_reason`, `last_skipped_at`, `skip_count`, `last_error`.

## Out of scope

- Fixing the underlying causes of skips (that's per-cause work — token renewal, dangling references, etc.). This filing is only about making the state visible.
- Cron-trigger validity checks at create time (a related but distinct filing shape).
- The specific transient-vs-permanent distinction — the loop can continue treating them the same at the tick level as long as the *cumulative* skip state is projected somewhere operator-visible.

---

Parent goal: #221
Parent initiative: #2364

damir-topic:
## Summary

- The scheduler tick loop's precondition gate (`_prepare_invocation`) skips a due trigger without advancing its fire state when the agent definition is missing/disabled or credentials can't be resolved. That skip used to be **invisible on the row** — a permanent misconfiguration (revoked scheduler Vault token, deleted-but-referenced definition, never-persisted agent secret) skipped every 30 s tick with the only trace a pod-log WARN, while `meho scheduler list` showed a healthy-looking `active` trigger. This projects the skip state onto the row and parks the trigger after N consecutive unresolvable skips.
- **Migration 0057** adds three columns to `scheduled_trigger`: `skip_count` (consecutive skips since the last successful fire, NOT NULL default 0), `last_skip_reason` (machine tag), `last_skipped_at`. Additive/reversible; pre-#2327 rows backfill to "never skipped".
- **Loop**: `_prepare_invocation` now returns a typed `_PreconditionSkip(reason=…)` instead of a bare `None`; `_record_skip` increments the count + stamps the reason on every skip and parks the row (`status='paused'`) once it hits `_PARK_AFTER_CONSECUTIVE_SKIPS` (a fixed module constant, 10 — matching `_CLAIM_BATCH_LIMIT`'s "dumb fixed loop-bound" posture, no new tunable). A successful fire clears the skip state (`_clear_skip_state`). The corrupt-cron / unknown-kind park paths (`_park_trigger`) also stamp a reason now, so every paused-by-the-loop row explains itself.
- **Read surfaces**: `ScheduledTriggerRead` carries the three fields, so `GET /api/v1/scheduler/triggers`, `meho.scheduler.list` / `.show` (MCP), `meho scheduler list` (new `SKIPS` column), and the operator console (warning badge on the list row + skip block on the detail page) all agree with the pod-log WARNs. OpenAPI snapshot + typed Go client regenerated.
- The at-most-once fire contract and transient-retry behaviour are unchanged — the columns are additive visibility only.

## Test plan

- [x] `ruff check` / `ruff format --check` — clean (9 files)
- [x] `mypy` — clean (11 source files)
- [x] `code-quality.py --diff` — 0 blockers (pre-existing file-size + sub-100 function-size warnings only)
- [x] `alembic upgrade head` reaches `0057`; `downgrade 0056` + `upgrade head` round-trips on SQLite
- [x] `go build` / `go vet` / `go test ./internal/cmd/scheduler/` — pass; `gofmt` clean
- [x] New loop tests (`tests/test_scheduler.py`): unresolved-credentials skip records reason + count; missing-definition skip records `definition_missing`; N consecutive skips park the trigger; a successful fire clears skip state; corrupt-cron park stamps a reason
- [x] New migration tests (`tests/migrations/test_migration_0057_*`): upgrade adds the columns with correct nullability; `skip_count` backfills 0 on a pre-existing row; downgrade→upgrade round-trip (pinned to own revision `0057`, never `head`, per the stamp-replay footgun)
- [x] New UI-view tests: projection surfaces `skip_count` / `last_skip_reason`
- [x] Full backend unit suite — green

## Out of scope

- The third fix shape in the issue (an audit row on first-skip-per-cause, keyed `(trigger_id, cause_hash)`) — the issue calls it "a nice-to-have on top"; deferred.
- Fixing the underlying skip causes (token renewal, dangling references) — per-cause work, explicitly out of scope in the issue.
- Splitting `scheduler/loop.py` (now ~930 lines, pre-existing over the 600-line advisory; warning only, not agent-introduced) — a behavior-preserving refactor best done as its own task.

Closes #2327

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,29 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Added — scheduler skip-state on the trigger row + park after N unresolvable skips (#2327)
+
+- The scheduler tick loop now projects its precondition **skips** onto the
+  `scheduled_trigger` row (migration `0057`, three columns): `skip_count`
+  (consecutive skips since the last successful fire), `last_skip_reason`
+  (machine tag — `definition_missing` / `definition_disabled` /
+  `credentials_unresolved`; the corrupt-cron / unknown-kind park paths also
+  stamp `invalid_cron_expr` / `unknown_kind`), and `last_skipped_at`.
+  Previously a permanent misconfiguration (revoked scheduler Vault token,
+  deleted-but-referenced definition, never-persisted agent secret) made the
+  loop skip every 30 s tick with the only trace a pod-log WARN, while
+  `meho scheduler list` showed a healthy-looking `active` trigger — a real
+  deploy lost ~360 hourly fires over 15 days before anyone noticed.
+- After a fixed number of **consecutive** unresolvable skips the loop parks
+  the trigger (`status='paused'`) so the state machine itself communicates
+  "broken, stopped trying" instead of re-tripping forever; a successful
+  fire resets the streak and clears the skip fields. The new state is
+  surfaced on `GET /api/v1/scheduler/triggers`, `meho.scheduler.list` /
+  `.show` (MCP), `meho scheduler list` (a `SKIPS` column), and the operator
+  console (a warning badge on the list row + a skip block on the detail
+  page). The at-most-once fire contract and transient-retry behaviour are
+  unchanged — the columns are additive visibility.
+
 ### Added — approval-TTL lifecycle wired end-to-end (parked approvals now expire) (#2322)
 
 - Every parked approval is now stamped `expires_at = created_at +
--- a/backend/alembic/versions/0057_add_scheduled_trigger_skip_state.py
+++ b/backend/alembic/versions/0057_add_scheduled_trigger_skip_state.py
@@ -1,0 +1,91 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""Add ``scheduled_trigger`` skip-state projection columns (#2327).
+
+Revision ID: 0057
+Revises: 0056
+Create Date: 2026-07-11
+
+Initiative #2364, Task #2327. The tick loop's precondition gate
+(:func:`~meho_backplane.scheduler.loop._prepare_invocation`) skips a due
+trigger *without advancing its state* whenever the agent definition is
+missing/disabled or the agent credentials cannot be resolved. Skip-and-
+retry is right for a transient miss, but a **permanent** miss (expired
+scheduler Vault token, never-persisted agent secret, deleted-but-still-
+referenced definition) produced an infinite silent loop whose only trace
+was a WARN pair in the pod log every tick -- nothing on the trigger row,
+so ``scheduler list`` showed a healthy-looking ``active`` trigger while it
+skipped every fire for weeks (a real deploy lost ~360 hourly fires over 15
+days before anyone noticed).
+
+This migration adds three columns that project the cumulative skip state
+onto the row so the read surfaces agree with the pod-log WARNs:
+
+* ``last_skip_reason`` -- ``text`` nullable. The stable machine tag of the
+  most recent skip cause (``definition_missing`` / ``definition_disabled``
+  / ``credentials_unresolved``; the park paths also stamp
+  ``invalid_cron_expr`` / ``unknown_kind``). NULL until the first skip;
+  the loop clears it back to NULL on the next successful fire.
+* ``last_skipped_at`` -- ``timestamptz`` nullable. UTC time of the most
+  recent skip; NULL until the first skip, cleared on the next fire.
+* ``skip_count`` -- ``integer`` NOT NULL, server default ``0``. Consecutive
+  skips since the last successful fire (reset to 0 on the next fire). The
+  loop parks the trigger (``status='paused'``) once this reaches
+  ``_PARK_AFTER_CONSECUTIVE_SKIPS`` so a permanently-unresolvable trigger
+  stops silently re-tripping every tick.
+
+Soft-column discipline mirrors ``0043`` / ``0053`` / ``0055``: the two
+nullable columns take no server default (Python-side ``None`` /
+loop-set); ``skip_count`` takes a ``0`` server default so pre-#2327 rows
+backfill to "never skipped" without a data-migration pass. All three are
+purely additive, so ``helm rollback`` to the pre-0057 image (which never
+mentions the columns) is safe -- the forward-compat regression test
+covers this shape.
+
+Reversibility contract
+----------------------
+
+``downgrade()`` drops the three columns in reverse add order. SQLite's
+ALTER TABLE drop-column has been supported since 3.35.0 (we're on 3.45+),
+so Alembic's batch-mode fallback is not required.
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0057"
+down_revision: str | None = "0056"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    """Add the three skip-state columns to ``scheduled_trigger``."""
+    op.add_column(
+        "scheduled_trigger",
+        sa.Column("last_skip_reason", sa.Text(), nullable=True),
+    )
+    op.add_column(
+        "scheduled_trigger",
+        sa.Column("last_skipped_at", sa.DateTime(timezone=True), nullable=True),
+    )
+    op.add_column(
+        "scheduled_trigger",
+        sa.Column(
+            "skip_count",
+            sa.Integer(),
+            nullable=False,
+            server_default="0",
+        ),
+    )
+
+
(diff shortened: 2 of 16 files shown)


damir-topic:
@coderabbitai review

damir-topic:




**Anchor task:** #2327 — project scheduler skip-state onto the trigger row + park after N unresolvable skips
**Surface area:** backend (+ devops/migration, CLI Go client, operator console)
**Verdict:** APPROVE
**Head SHA reviewed:** `a520c1f2`
**AI reviewer (CodeRabbit):** not received — CodeRabbit is rate-limited ("Review limit reached, next review available in 16 minutes"), outside the 15-min wait window; no bot findings available this iteration.

### Acceptance criteria
- [x] Skip state projected onto the trigger row — `last_skip_reason` / `last_skipped_at` / `skip_count` added (migration 0057) and surfaced on REST `ScheduledTriggerRead`, MCP `scheduler.list`/`.show` (via the same schema), `meho scheduler list` (`SKIPS` column), and the operator console (list warning badge + detail skip block). Verified end-to-end.
- [x] Park after N consecutive unresolvable skips — `_record_skip` increments `skip_count` and parks (`status='paused'`) at `_PARK_AFTER_CONSECUTIVE_SKIPS=10`; verified by `test_consecutive_skips_park_the_trigger`.
- [x] Successful fire clears the streak — `_clear_skip_state` resets count + reason; verified by `test_successful_fire_clears_skip_state`.
- [x] Additive/reversible migration; pre-#2327 rows backfill `skip_count=0` — verified by `test_skip_count_backfills_zero_on_existing_row` and a clean down→up round-trip.

### Predicted vs. actual diff
- ✅ migration 0057, `models.py`, `scheduler/loop.py`, `scheduler/schemas.py`, UI routes/templates, `openapi.json`, `client.gen.go`, `cli/.../list.go`, docs, CHANGELOG, tests — all expected and correct.
- No missing predicted files; no undisclosed scope. Out-of-scope items (audit-row-per-cause, loop.py split) explicitly deferred in the PR body, consistent with the issue.

### Integration verification
- [x] `ruff check` + `ruff format --check` — clean (changed files)
- [x] `mypy src/.../scheduler/loop.py` — clean
- [x] `pytest tests/migrations/test_migration_0057_*` — 3 passed
- [x] `pytest tests/test_scheduler.py tests/test_ui_scheduler_views.py` — 51 passed
- [x] migration replay: down→up round-trip green on SQLite; `skip_count` backfills 0
- [x] OpenAPI snapshot regenerated from source and byte-identical to the committed `cli/api/openapi.json` — **snapshot is fresh**
- [x] `go build ./...` + `go test ./internal/cmd/scheduler/` — pass
- [x] migration head: 0056 confirmed as the sole current head on `origin/main`; no competing 0057

None. No blockers, no majors, no minors posted inline.

- `_clear_skip_state` runs on a successful **prepare** (before the advance / mark-fired step), not on the completed fire. This is deliberate and documented (docs/codebase/scheduler.md) — `skip_count` tracks *precondition* skips only, and a lost advance-race or a swallowed invocation-time failure is a different failure class (visible in audit, at-most-once contract). Semantically sound; noted only for the record.
- In the rare path where `_clear_skip_state` runs (uncommitted) and then the advance loses the replica race, the pending clear can be rolled back if a *sibling* row in the same tick batch later raises. It self-heals on the next tick (and the winning replica clears it), so no durable-state issue — not worth a change.

- Adopted: 0 — CodeRabbit rate-limited, no review produced this iteration.
- Disputed: 0
- Out of scope: 0

### Docs
- [x] `docs/codebase/scheduler.md` — updated with a dedicated "Skip-state projection + park-after-N (#2327)" section; accurate to the code.

- Migration-test idempotency correctly pinned to its own revision `0057`/`0056`, never `head` — dodges the known stamp-replay duplicate-column footgun.
- End-to-end contract is tight: `openapi.json` required-array + Go `SkipCount int` (non-pointer, matching NOT NULL) + CLI `SKIPS` column all agree with the NOT NULL server-default-0 column.
- Thoughtful UI `elif` for the corrupt-cron park case (`skip_count=0` but `last_skip_reason` set via `_park_trigger`) keeps the reason visible.

PR #2369, opened 2026-07-11
fix(scheduler): renew Vault token on use + startup/periodic lookup-self (#2328)

### LINKED ISSUE #2328: Task: renew the scheduler Vault token on use + startup/periodic lookup-self — defuse the ~32-day periodic-token fuse
## Summary

The scheduler authenticates to Vault with the static `VAULT_SCHEDULER_TOKEN`. The documented mint is a **periodic** token (`-period=768h` in the cross-repo onboarding guidance), but at `v0.20.0` the backplane never renews it — there is no `renew-self` call in `scheduler/vault_credentials.py`, and no deploy component (init container, sidecar, external cron) does either.

A Vault periodic token expires `period` after its **last renewal**. With `-period=768h` and never a renewal, that's a built-in ~32-day fuse: when it blows, every Vault-first credential read returns Forbidden and the scheduler enters the silent-skip loop tracked in the sibling filing.

Observed on `v0.20.0`.

## Reproduction / Observed

Source-side confirmation against the pinned tag:

```bash
gh api "repos/evoila/meho/contents/backend/src/meho_backplane/scheduler/vault_credentials.py?ref=v0.20.0" \
  --jq '.content' | base64 -d \
  | grep -nE 'renew_self|renew-self|token/renew|lookup_self|lookup-self'
# → no matches at v0.20.0
```

**Lived reproduction on a real deploy**: `VAULT_SCHEDULER_TOKEN` minted ~2026-06-05 with `-period=768h`, dead by 2026-07-09 (Vault `lookup-self` returned 403), one hourly cron accumulated ~360 missed fires over the intervening 15 days. Remediation required a Vault-admin re-mint, a Kubernetes Secret patch, and a pod restart — anything less does not pick up the new token because the scheduler reads it once at startup.

The immediate downstream failure signature is the WARN pair every 30-second tick:

```
WARN scheduler_vault_read_failed  path=... status_code=403
WARN scheduler_credentials_unresolved trigger_id=...
```

...combined with the total silence on the operator-visible surfaces documented in the sibling filing.

## Why this matters

- **Time-bomb by default.** Any deploy following the documented mint pattern will silently break ~32 days after standup. That includes fresh deploys, CI environments, and every re-mint after this one.
- **Recovery is manual and multi-step.** Re-mint (Vault admin), patch the `Secret` (Kubernetes admin), restart the pod (deploy owner) — no self-healing path exists.
- **Compounds with the silent-skip issue.** With no surface indicating "your scheduler is broken because its Vault token is dead," the ~32-day cliff is functionally invisible until a downstream cron artifact goes stale enough to notice.
- **Applies to every Vault-first credential path the scheduler drives.** Not just one connector — anything the scheduler fetches via Vault fails once the token dies.
- **The periodic-token pattern is the correct Vault primitive.** The gap is that we minted for renewal and then never renewed; a renewed periodic token effectively never expires while the deploy is alive.

## Fix shapes (each independently actionable, roughly in preference order)

- **Renew-on-use.** After every successful agent-secret read/write, fire a best-effort `auth/token/renew-self` (via `hvac.Client.renew_self()` or an equivalent). Cheap, self-healing, no new cron; the token is renewed at scheduler-tick frequency so a periodic token with any sane `period` never dies. Failures on `renew-self` log-and-continue (the underlying read already succeeded), avoiding new failure modes.
- **Startup + slow-cadence validation.** `lookup-self` at scheduler start, plus on a slow cadence (say every hour). On failure, log LOUDLY and surface via `/ready features.scheduler` (or the trigger skip-state field of the sibling filing). Doesn't fix the fuse but at least shortens time-to-notice from ~weeks to ~minutes.
- **Dedicated renewal cron.** A separate scheduler task that calls `renew-self` every N minutes. More moving parts than renew-on-use but cleanly separable from read paths.
- **Documentation-only fallback.** At minimum, spell out the renewal requirement + expiry fuse in the cross-repo scheduler onboarding doc so deployers put the re-mint on the calendar. Least effort; least reliable.

Renew-on-use is the standard Vault-client pattern (documented in Vault's own guidance for long-running clients holding periodic tokens) and is what most other consumers do.

## Verification

- Source-side: `scheduler/vault_credentials.py` (or the client it delegates to) invokes `renew-self` on a definable trigger — either after each read/write, or on a periodic cadence.
- Live: on a deploy, watch the token's `expire_time` via `vault token lookup <accessor>` — it should advance every renewal, not sit frozen at mint time.
- Long-run: leave a scheduler running past the token's original `-period=768h`; fires continue.

## Out of scope

- Non-periodic token modes (batch, service). The renewal fix targets the documented periodic path.
- The surface-the-failure problem — that's the sibling filing about silent scheduler skips. The two are complementary but independently actionable.
- Auth-method changes (Kubernetes auth, AppRole, etc.). Those are alternative approaches to the same end but represent much larger deployment changes than fixing the current token model.

---

Parent goal: #221
Parent initiative: #2364

damir-topic:
## Summary

- The scheduler authenticates to Vault with a static **periodic** service token (`-period=768h`) but never renewed it. A periodic token expires `period` after its *last renewal*, so a token minted per the onboarding guide silently died ~32 days after standup and every Vault-first credential read began returning 403 (silent-skip loop).
- **Renew-on-use**: fire a best-effort `auth/token/renew-self` after every successful agent-secret read/write, renewing the token at scheduler-tick frequency so any sane `period` never ages out. Failures are logged and swallowed (the read/write already succeeded) — never a new failure mode.
- **Startup + hourly `lookup-self`**: `verify_scheduler_token` runs at scheduler start and on a slow cadence from the tick loop, logging a dead/unreachable token loudly (`scheduler_vault_token_dead` / `scheduler_vault_token_unreachable`) so time-to-notice drops from weeks to minutes; the healthy path logs `ttl`/`expire_time`. It returns a `SchedulerTokenStatus` that sibling #2327 can surface via `/ready features.scheduler`.
- **Live token re-read**: the token is resolved from its source per use. New optional `VAULT_SCHEDULER_TOKEN_FILE` names a Vault-Agent sidecar sink re-read on every read/write, so a re-mint is picked up **without a pod restart** (the manual remediation the fuse otherwise forced). Static `VAULT_SCHEDULER_TOKEN` stays the default; an unreadable/empty file falls through to it.

## Test plan

- [x] `ruff check` / `ruff format --check` — clean on all touched files
- [x] `mypy` — no issues (vault_credentials.py, loop.py)
- [x] `pytest tests/test_scheduler_vault_credentials.py` — 32 passed (incl. new renew-on-use, token-file re-read, and `lookup-self` status tests)
- [x] `pytest -k "settings or scheduler_vault or scheduler_credentials"` — 151 passed, 5 skipped
- [x] `pytest tests/test_scheduler.py` + scheduler loop tests — 36 + 2 passed
- [x] Renew-on-use verified best-effort: a failing/non-renewable renew (e.g. dev root token) is swallowed, the read/write still returns

## Out of scope

- Surfacing the dead-token state on `/ready features.scheduler` / trigger skip-state — owned by sibling #2327; this PR emits the `SchedulerTokenStatus` signal it consumes.
- Auth-method changes (Kubernetes auth, AppRole). The fix targets the documented static periodic-token path.

Closes #2328

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduler Vault token reliability by renewing tokens after successful secret operations.
  * Added startup and hourly health checks with clear logging for expired, invalid, or unreachable tokens.
  * Prevented periodic token expiry issues by dynamically using refreshed token values.

* **New Features**
  * Added optional file-based token sourcing through `VAULT_SCHEDULER_TOKEN_FILE`, with fallback behavior for missing or empty files.

* **Documentation**
  * Documented token renewal, health checks, expiration behavior, and token re-minting without restarts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,31 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Fixed — scheduler Vault token renews on use + startup/periodic lookup-self (defuses the ~32-day periodic-token fuse) (#2328)
+
+- The scheduler now fires a best-effort `auth/token/renew-self` after
+  every successful Vault agent-secret read/write, so the documented
+  **periodic** service token (`-period=768h`) is renewed at scheduler-tick
+  frequency and never ages out while the process runs. Previously the
+  backplane never renewed it — a token minted per the onboarding guide
+  silently died ~32 days after standup and every Vault-first credential
+  read started returning 403. Renewal failures are logged and swallowed
+  (the read/write already succeeded), never a new failure mode.
+- The scheduler token is now self-looked-up (`auth/token/lookup-self`) at
+  startup and on a slow cadence (hourly), logging a dead or unreachable
+  token as a loud `scheduler_vault_token_dead` /
+  `scheduler_vault_token_unreachable` ERROR the moment it's observed —
+  cutting time-to-notice from weeks to minutes. The healthy path logs
+  `scheduler_vault_token_verified` with the token's `ttl`/`expire_time`
+  so operators can watch expiry advance across renewals.
+- The token is now resolved from its live source on every use. A new
+  optional `VAULT_SCHEDULER_TOKEN_FILE` names a file (a Vault-Agent
+  sidecar sink) the token is re-read from per read/write, so a re-mint is
+  picked up **without a pod restart** — removing the manual re-mint +
+  Secret-patch + pod-restart remediation the fuse otherwise forced. The
+  static `VAULT_SCHEDULER_TOKEN` remains the default; an unreadable/empty
+  file falls through to it.
+
 ### Added — approval-TTL lifecycle wired end-to-end (parked approvals now expire) (#2322)
 
 - Every parked approval is now stamped `expires_at = created_at +
--- a/backend/src/meho_backplane/scheduler/loop.py
+++ b/backend/src/meho_backplane/scheduler/loop.py
@@ -120,6 +120,7 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import json
+import time
 import uuid
 from dataclasses import dataclass
 from datetime import UTC, datetime
@@ -180,6 +181,14 @@ _SCHEDULER_ADVISORY_LOCK_KEY: int = 0x4D45_484F_5343_4844  # "MEHOSCHD"
 #: one-off workloads (the consumer doc anticipates "dozens" of
 #: triggers per deployment).
 _CLAIM_BATCH_LIMIT: int = 50
+
+#: Slow cadence (seconds) for the scheduler Vault-token ``lookup-self``
+#: health check (#2328). Deliberately a fixed constant, not a tunable:
+#: hourly is frequent enough to cut time-to-notice on a dead token from
+#: weeks to minutes without adding Vault load, and the substrate stays
+#: config-minimal. The renew-on-use path (which actually keeps the token
+#: alive) runs at tick frequency; this is only the visibility backstop.
+_TOKEN_CHECK_INTERVAL_SECONDS: float = 3600.0
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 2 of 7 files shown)


damir-topic:
@coderabbitai review

damir-topic:
@coderabbitai review

Issue #2364, opened 2026-07-10
Initiative: G0.31 v0.20.0 closed-loop dogfood hardening — wave 2 (RDC test period)


damir-topic:
## Summary

Second hardening wave from the v0.20.0 RDC test period (consumer ticket evoila-bosnia/claude-rdc-hetzner-dc#1790): 22 ops-filed issues (#2322–#2340, #2345–#2347, + #2324) triaged 2026-07-10 with code grounding against post-wave-1 main. Three were already answered by shipped code and are closed with grounding comments (#2324, #2330, #2335). The remaining 19 are adopted as child tasks — adopted in place (this wave's ops bodies are high-quality; each carries a grounding-delta comment with file:line evidence), except three that were materially re-diagnosed and rewritten (#2322, #2325, #2337).

## RDC root-cause corrections (recorded so they are not re-filed under a wrong premise)

| Filed claim | Grounded correction |
|---|---|
| #2337 "bulk log retrieval via MCP is functionally unavailable" | REFUTED — the `result_query` meta-tool (G0.20-T7, `mcp/tools/result_query.py`) has been the handle read-back since v0.11: `result_query(handle_id, offset, limit)` walks the spilled set. Real defects: the handle's `fetch_more` never advertises it, and the inline sample is head-first (`jsonflux_reducer.py:696`, `_sample_from_tail` exists but is never set for logs). |
| #2325 "no production code emits events" | STALE on main — `operations/agent_run.py:484` publishes agent-run terminal-transition events onto the outbox. The drain matcher is still the documented T5 no-op; the roadmap issue is #826. |
| #2333 "ingest crashes the pod; the documented remediation crashes the service" | Largely fixed on main: 202 + `create_task` + watchdog job model (#2275/PR #2317), parse + proto-build in `asyncio.to_thread` (`ingest/pipeline.py:874/:1013`), CSafeLoader. The `composite_l2_missing` remediation leg is moot — #2259/PR #2297 deleted that apparatus. Remaining: an end-to-end responsiveness regression pin. |
| #2322 "no sweeper exists" | Half-stale: `expire_stale_requests` EXISTS (`operations/approval_queue.py:907`) but has zero production callers, and nothing stamps `expires_at` at park — the shipped TTL lifecycle is inert (the #2288 `record_profile_stamp` pattern again). |
[...]

Issue #2332, opened 2026-07-10
Task: redaction-safe proposed_effect preview builder for vault.kv.* credential writes — approver sees path + key names, never values


zdamir:
## Summary

Found by exercising the G11.7 approval queue on v0.10.0 (PARTIAL against v0.20.0). A parked `vault.kv.put` approval request's `proposed_effect` field is `{op_id, target_id, connector_id}` — no path, no data, no key names, no before/after diff. An approver calling `approvals_get` sees "some `vault.kv.put` against `<target>` is pending" but cannot see WHICH path or WHAT would be written. The approval decision is made blind to the effect being approved.

Other write ops in the same v0.10.0 release notes ship richer previews:

- `proposed_effect` park-time previews auto-populate a `k8s.apply` dry-run preview (evoila/meho#1454).
- `argocd.app.set` / `argocd.appproject.*` "snapshot the spec before and after the change into the result's proposed_effect block so the reviewer/auditor sees exactly what shifted" (per the op descriptions).

So the preview is op-dependent and simply absent for vault writes.

## Reproduction / Observed

```bash
# Park an approval-gated vault.kv.put
curl -ksS -X POST https://<backplane>/mcp \
  -H "Authorization: Bearer $JWT" \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  --data '{
    "jsonrpc":"2.0","id":1,"method":"tools/call",
    "params":{"name":"call_operation","arguments":{
      "connector_id":"vault-1.x","op_id":"vault.kv.put",
      "target":"<vault-target>",
      "params":{"path":"<tenant>/proposed-effect-probe",
                "data":{"probe":"reject-me"}}
    }}
  }'

# Extract approval_request_id, then read it back:
curl -ksS -X POST https://<backplane>/mcp \
  -H "Authorization: Bearer $JWT" \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  --data '{"jsonrpc":"2.0","id":2,"method":"tools/call",
           "params":{"name":"meho_approvals_get",
                     "arguments":{"approval_request_id":"<uuid>"}}}'
```

Observed `proposed_effect` on the returned parked request:

```json
{
  "op_id": "vault.kv.put",
  "target_id": "<uuid>",
  "connector_id": "vault-1.x"
}
```

No `path`, no `params`, no `preview`, no `data`. The approver has no way to distinguish "write a benign scratch value" from "overwrite a production credential."

[...]

Issue #2091, opened 2026-06-26
targets: fail-fast on import when secret_ref is outside meho's readable subtree, and name the cause when dispatch hits Forbidden


zdamir:
## Summary

A target whose `secret_ref` points outside `secret/meho/*` imports clean via `meho targets import` (no validation against the meho-mcp policy's accessible subtree at import time), but **fails at dispatch** with a bare `connector_error: Forbidden` and no naming of the cause or remedy.

## What we hit

The backplane resolves a target's credential at dispatch by reading the target's `secret_ref` from Vault under its own federated identity — the `meho-mcp` policy, which by design grants read on `secret/meho/*` **only** (the policy's comment + the cross-repo Vault provisioning doc both call this out — "v0.2 delegates per-secret authorization to Keycloak claims, not per-path Vault policy, so a single subtree-read grant is intentional and sufficient").

A target registered with a `secret_ref` **outside** that subtree (e.g. pointing at the path a local CLI wrapper reads, `secret/<project>/.../<role>`) imports with no complaint. The mismatch only surfaces at dispatch:

```json
{
  "status": "error",
  "error": "connector_error: Forbidden",
  "extras": {
    "exception_class": "Forbidden",
    "exception_message": "1 error occurred:\n\t* permission denied\n\non get https://<vault>/v1/secret/metadata/<that-out-of-subtree-path>"
  }
}
```

The error names neither the cause ("the target's `secret_ref` is outside meho's readable subtree `secret/meho/*`") nor the remedy ("stage the cred under `secret/meho/<product>/<name>`"). It reads exactly like a missing Vault grant — and the first instinct (a consumer-side ticket originally asked to **widen** the meho-mcp policy) is the wrong fix:

- The policy is deploy-owned (re-applied on every meho deploy) — any widening is clobbered on the next helm upgrade.
- Widening grows the backplane's blast radius for no benefit; the design says credentials live under `secret/meho/*`.

The correct consumer-side resolution is to **move the secret** to where the backplane already looks. But nothing in the product surfaces that, so the consumer burns several debugging steps to discover it.

### 1. Fail-fast at import

`meho targets import` (and the MCP target-register surface) should reject — or at minimum warn — when `secret_ref` is outside `secret/meho/*`, with a message that states the constraint and the convention:

```
ERROR: target's secret_ref `secret/<project>/.../<role>` is outside meho's
       readable subtree `secret/meho/*`. Stage the credential under
       `secret/meho/<product>/<name>` and re-run import.
[...]

PR #2371, opened 2026-07-11
feat(connectors): structured post-approval vault-write-forbidden error + envelope write-capability warning (#2331)

### LINKED ISSUE #2331: Task: park-time Vault write-capability preflight + structured post-approval auth error + write-identity contract doc
## Summary

Observed on v0.10.1 (still OPEN against v0.20.0). During a break-glass approve→execute proof, a human-initiated agent run resumed after four-eyes approval and correctly re-dispatched its `vault.kv.put`. The approval mechanic worked, and the op reached the connector. But the actual write returned:

```
connector_error: Forbidden — "permission denied" on post .../secret/data/<tenant>/<agent-run-scratch-path>
```

So the op passed the entire governance handshake (park → four-eyes approve → re-dispatch), executed against the target, and then failed at Vault RBAC on the connector's write identity. Two orthogonal problems fall out.

## Reproduction / Observed

1. Configure a Vault target with an OIDC / JWT-federated read path that works for operators via the vault connector.
2. Grant the operator's own Vault token (e.g. userpass-`<operator>` with admin policy) `create`/`update` on `secret/data/<tenant>/*` — verify by hand that the operator CAN read and write under that prefix.
3. Trigger a human-initiated agent run whose plan contains a `vault.kv.put` under that prefix. The op parks for approval as designed.
4. Approve with a distinct approver / break-glass. The run resumes and re-dispatches the write.
5. **Observed**: the connector returns `connector_error: Forbidden` with Vault's `permission denied` on the POST to `secret/data/<tenant>/<path>`.

Additional diagnosis performed on the running deploy:

- The operator's own token (admin policy) can create/update the exact path the agent tried to write. So the connector did **not** federate the operator's identity for the write.
- The MCP OIDC role that federates JWTs → Vault tokens gives federated tokens `token_policies: ["default"]` plus whatever the JWT's `groups` claim maps to. Agent runs / service-account JWTs carry no `groups` claim, so the federated token has no write scope.
- Adding a scratch-scoped write policy to that OIDC role's `token_policies` did NOT unblock the write (the run federated the token after the edit and still got Forbidden). So the connector's write identity for an agent op is **not** the operator-facing OIDC-federated token either — it is a distinct, presently read-only / unprovisioned identity.

The vault connector's daily-driver READ path (`vault.kv.read`) works for operators via JWT federation. The WRITE identity for an agent op is a separate identity that the deploy had not provisioned with write permissions — and this is not visible until after approval.

## Why this matters

Two distinct failure modes stack:

1. **Consumer / deploy-config foot-gun.** To make an approved agent write actually land, whatever identity the connector federates or uses for an agent write path must have create/update on the target Vault paths. There is no documented list of "these Vault paths + policies must exist for approved agent writes to work," and the operator-facing OIDC role is not the identity used. Deploys that follow the read-path documentation provision READ only and discover this only when an approval is executed.

2. **Approval UX / late failure.** The RBAC failure surfaces only AFTER the human approves and the run re-dispatches. An approver can pass the whole governance handshake — spend a human's time reviewing a write, click Approve — and the write still cannot land. From the operator's point of view the four-eyes flow succeeded; from the target's point of view nothing changed. This trains approvers to distrust "approved" as a terminal state.

Even with the approve→execute mechanic working correctly for the human-initiated path, an approved agent write does not land until the connector's write identity has Vault write permissions on the target path. In the meantime, "an agent write lands via approval" is functionally blocked in any deploy that hasn't guessed the right identity + policy pairing.

- **Pre-flight the connector's write authorization at park time.** Before parking the approval, do a Vault capability check (`sys/capabilities-self` or equivalent) using the identity the connector will use to execute the write. If the identity lacks `create`/`update` on the target path, either (a) refuse to park with a structured error naming the missing capability + the identity that lacks it, or (b) park with a `proposed_effect` warning like `connector_identity_may_lack_write` so the approver sees it before clicking Approve.
- **Document the write-identity contract.** In the vault connector's ingest / setup docs, explicitly enumerate: which identity is federated for reads (operator OIDC-federated token), which identity is used for agent writes (currently a distinct, deploy-provisioned identity), and what Vault policies each must carry. Include a curl / CLI probe deploys can run to verify the write identity has scope on their intended target paths.
- **Return a structured error on the post-approval failure**, not the raw connector text. Something like `connector_error.code: "vault_write_identity_forbidden"` with `path`, `identity_hint`, and a link to the setup doc, so the operator sees the actionable problem instead of a bare "permission denied".
- **Consider making the write identity federate the same OIDC role as reads** (with distinct scope), so there is one identity to provision, not two.

## Verification

Heavy / manual — the probe requires an approver + a connector write identity that can be modified.

FIXED when: an approved agent `vault.kv.put` to a Vault path where the connector's write identity has been granted `create`/`update` lands (a follow-up `vault.kv.read` returns the data), AND the connector's write authorization is validated at park time so a mis-provisioned identity fails BEFORE the human approves.

Currently OPEN against v0.20.0: the approve→execute mechanic works, but the connector's write identity remains unprovisioned by default and the failure still surfaces only after approval.

## Out of scope

- Whether the four-eyes / approve→execute mechanic itself works (it does — the write reaches the connector).
- Vault-side policy authoring; that is deploy work, not upstream product work.
- Read-path OIDC federation; the daily-driver read path is unaffected.

---

Parent goal: #221
Parent initiative: #2364

damir-topic:
## Summary

Completes the write-identity leg of the four-eyes vault-write flow for **#2331** (Initiative #2364, G0.31 wave 2). Even with approve→execute working, an approved `vault.kv.*` write does not land until the connector's write identity has Vault write capability on the target path — and the failure was previously indistinguishable from a read-path denial.

Three parts (each independently actionable per the task):

1. **Structured post-approval error** — a typed `vault.kv.put`/`patch`/`delete` denied by Vault at dispatch now surfaces a dedicated `connector_vault_write_forbidden` result (the write-side sibling of `connector_vault_forbidden`, #2091), naming the exact `<mount>/data/<path>` Vault denied, the acting operator identity, and the §6.1 write-policy stanza to add — instead of a bare permission-denied or a misleading read-path diagnosis.
2. **Envelope write-capability warning** — the park-time `sys/capabilities-self` write preflight already shipped; this promotes its `will_be_denied` signal to a top-level `write_capability_warning` on the parked `proposed_effect` envelope (riding #2332's provenance surface) so the approval surface can render a "this write may not land" banner. A warning, not a gate — refuse-to-park stays off by default.
3. **Write-identity contract doc** — `docs/cross-repo/connector-vault-policy.md` documents the per-operator templated write path and the do-NOT-widen-the-shared-policy warning.

## Verification

- ruff / ruff-format / mypy clean on changed src.
- `test_operations_connector_vault_forbidden.py` + `test_operations_dispatcher.py`: 58 passed (incl. new write-forbidden + envelope-warning tests).
- No new alembic migration; no FastAPI route/schema change (so no OpenAPI regen).

Closes #2331

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,22 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Added — structured post-approval vault-write-forbidden error + write-capability warning on the approval envelope (#2331)
+
+- A typed `vault.kv.put` / `patch` / `delete` that Vault denies at
+  dispatch now surfaces a dedicated `connector_vault_write_forbidden`
+  result (the write-side sibling of the read-oriented
+  `connector_vault_forbidden`, #2091) naming the exact `<mount>/data/<path>`
+  Vault denied, the acting operator identity, and the write-policy stanza
+  to add — instead of a bare permission-denied or a misleading read-path
+  diagnosis. The park-time `sys/capabilities-self` write preflight already
+  shipped; #2331 promotes its `will_be_denied` signal to a top-level
+  `write_capability_warning` on the parked `proposed_effect` envelope so
+  the approval surface can render a "this write may not land" banner
+  (a warning, not a gate), and documents the per-operator templated
+  write-identity contract in `docs/cross-repo/connector-vault-policy.md`
+  (#2331).
+
 ### Added — redaction-safe `proposed_effect` preview for `vault.kv.*` credential writes (#2332)
 
 - A parked `vault.kv.put` / `patch` / `delete` approval request now
--- a/backend/src/meho_backplane/operations/_errors.py
+++ b/backend/src/meho_backplane/operations/_errors.py
@@ -50,6 +50,7 @@ __all__ = [
     "result_connector_tls_verify_failed",
     "result_connector_unsupported",
     "result_connector_vault_forbidden",
+    "result_connector_vault_write_forbidden",
     "result_denied",
     "result_handler_unreachable",
     "result_invalid_params",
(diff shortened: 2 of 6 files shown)


damir-topic:
@coderabbitai review

Issue #2259, opened 2026-07-09
Task: Retire the L2 failure-coping apparatus — preflight, composite_l2_* errors, composite_backing/UnbackedEnabledCompositeError


damir-topic:
Parent goal: #2247
Parent initiative: #2248

Depends-on: #2253
Depends-on: #2256
Depends-on: #2255

## Problem

Once no composite dispatches an ingested sub-op (I-B done), the entire failure-coping apparatus that existed only to manage per-deploy catalog state is dead code. This Task deletes it — the failure class is removed, not guarded. This is the payoff the RFC promises.

## Current state (all become unreachable after I-B)

- **vmware dispatch-time preflight + errors:** `preflight_l2_dependencies` + `_classify_sub_ops` (`connectors/vmware_rest/composites/_preflight.py`), `result_composite_l2_missing` / `result_composite_l2_disabled` (`operations/_errors.py:442,523`), their dispatcher catch arms (`operations/dispatcher.py:840-886`), and the `CompositeL2Dependency{Missing,Disabled}` exceptions (`operations/composite.py:172-301`).
- **github import-time backing guard:** `composite_backing` registry (`operations/composite_backing.py`) + `UnbackedEnabledCompositeError` (`connectors/github/composites/_register.py:89`), now superseded by A2's platform invariant.
- **vmware catalog-command escape hatch for composites:** `_catalog_command.py` surfaced via the `composite_l2_missing` remediation.

## Desired state

Delete the above with their tests, leaving no caller. Keep the A2 platform invariant as the sole guard. Verify nothing outside composites uses these symbols (e.g. `_catalog_command` may serve non-composite ingest surfaces — remove only the composite-remediation coupling).

## Acceptance criteria

- [ ] `preflight_l2_dependencies`, `CompositeL2DependencyMissing`, `CompositeL2DependencyDisabled`, `result_composite_l2_missing`, `result_composite_l2_disabled`, `composite_backing`, and `UnbackedEnabledCompositeError` are removed; `grep` shows zero remaining references in `src/`.
- [ ] The dispatcher's L2-error catch arms are gone; the error-code registry / docs no longer list `composite_l2_missing`/`composite_l2_disabled`.
- [ ] The full test suite is green with the corresponding tests deleted (not skipped).
- [ ] CLI OpenAPI snapshot / error-catalog docs regenerated where the removed error codes were surfaced.

## Related

[...]

PR #2372, opened 2026-07-11
test(ingest): pin event-loop responsiveness of a non-dry-run 1275-op ingest (#2333)

### LINKED ISSUE #2333: Task: regression-pin large-spec non-dry-run ingest — event-loop responsiveness proof (7 MB / 1275-op class)
## Summary

At v0.8.0 the catalog ingest pipeline's `dry_run=false` path crashes the meho pod when given any real-world OpenAPI spec. The dry-run path completes cleanly on the same spec, so parsing is fine — the crash is specifically in the commit-to-DB + LLM-grouping pass, which runs synchronously on the request thread and starves the readiness / liveness probes until K8s terminates the pod. Status against v0.20.0: PARTIAL — the endpoint now appears to return 202 (see verification), but the underlying long-running commit pattern still needs to be off the request thread and the crash-on-real-spec regression path is not confirmed fixed end-to-end.

This is a critical robustness issue because the catalog ingest is the **documented remediation path** for every connector that ships L2 composites. When an L2 composite dispatch fails, the `composite_l2_missing` envelope (which is otherwise excellent error UX) literally tells operators:

```
"catalog_command": "meho connector ingest --catalog <product>/<version>"
```

But the documented remediation crashes the pod on any real-world spec.

## Reproduction / Observed

Direct reproduction with a canonical large REST spec (7.5 MB, 1275 typed ops):

```
POST /api/v1/connectors/ingest
Body: {
  "product": "<product>",
  "version": "<version>",
  "impl_id": "<impl>",
  "specs": [{"uri": "<url to a real 7 MB+ OpenAPI spec>"}],
  "dry_run": false
}

Spec size: 7.55 MB
Spec ops:  1275 typed REST operations

Behavior:
  t=0     ingest request accepted
  t=30.1s request returns HTTP 502 "Bad Gateway"
  t=30.1s pod log: "Shutting down" + "Application shutdown complete"
  t=30s+  K8s reports lastState: {terminated: {exitCode: 0,
                                                reason: "Completed"}}
  t=~60s  pod restarts (RESTARTS=1), service back to normal
```

Key signal: `exitCode: 0` + `reason: "Completed"` — **not OOM**. The blocking parse + DB-commit + LLM-grouping pass on the request thread exceeds the readiness / liveness probe timeout, K8s sends SIGTERM, and the pod shuts down cleanly.

The same spec with `dry_run=true` (skips the DB writes + grouping pass) completes in ~30 seconds without crashing:

```
POST /api/v1/connectors/ingest  (dry_run=true)
→ HTTP 200, {"ingestion": {"connector_id": "<impl>-<version>",
                            "inserted_count": 1275,
                            "updated_count": 0,
                            "skipped_count": 0,
                            "connector_registered": false,
                            "operations_grouped": false},
              "grouping": null}
```

So the parse pipeline works. The crash is specifically in the commit-1275-ops-to-DB pass + the LLM grouping pass running synchronously on the request thread.

## Why this matters

- The `composite_l2_missing` error envelope points operators AT this exact command as the remediation. When the remediation crashes the pod, every affected composite becomes a hard dead-end with no documented next step.
- L2 composite dispatch for connectors whose specs exceed the crash threshold is functionally blocked. In the case of the large REST-API family that motivated this signal, ~13 composites in the packaged catalog were unreachable — every one returned `composite_l2_missing`, and running the documented remediation crashed the pod.
- Restarts trip other observers (approval-queue heartbeats, MCP sessions, WebSocket-based UI clients) — not just the ingest caller.
- The dry-run path's existence implies the design intent was to let operators validate specs before committing; the commit-pass then needed to be the LIGHTER operation. Today the inverse is true: dry-run works fine; commit crashes the pod.

- **Run the parse + commit + grouping pass off the request thread** — background worker, async task queue, whatever fits the existing stack. The endpoint returns a `job_id` immediately; the operator polls `GET /api/v1/connectors/ingest/{job_id}` for state (`queued` / `running` / `succeeded` / `failed`) and final counts. Industry-standard shape for long-running operations. This is the durable fix.
- **Stream-yield to the event loop periodically during the commit pass** (e.g. `await asyncio.sleep(0)` after every 50 ops inserted, or between the DB commit and the LLM-grouping pass), so health probes can complete during the ingest. Cheap short-term mitigation while the job-queue path is being built.
- **Bump the readiness / liveness probe timeout on ingest-handling pods.** Less ideal — it blocks all other requests during an ingest and just moves the crash threshold, doesn't remove it.
- **Split the commit pass into smaller transactions** (batched commit + grouping pass moved to a separate worker). Reduces peak time-on-thread; on its own doesn't eliminate the "one big spec" case.
- **Regardless of which shape lands: update the `composite_l2_missing` envelope's `catalog_command` guidance** to point at whatever the new pattern is (`meho connector ingest --catalog <product>/<version> --async` returning a job id, or similar), so operators aren't sent at a pattern that crashes the pod.

## Verification

Auto-runnable — the source signal ships a curl-driven probe.

```bash
SPEC_URL="<url to a real 7 MB+ OpenAPI spec>"
code=$(curl -ksS -o /tmp/sig-ingest.json -w '%{http_code}' --max-time 60 \
  -X POST https://<backplane>/api/v1/connectors/ingest \
  -H "Authorization: Bearer $JWT" -H "Content-Type: application/json" \
  -d "{\"product\":\"<product>\",\"version\":\"<version>\",\"impl_id\":\"<impl>\",\"specs\":[{\"uri\":\"$SPEC_URL\"}],\"dry_run\":false}")
```

- FIXED (exit 0): HTTP 200 / 201 with a real `inserted_count`, or HTTP 202 with a poll-able job id that reaches a completed state without pod restart.
- OPEN (exit 2): HTTP 502 / 503 within ~30s + pod restart observable in `kubectl get pod` `RESTARTS` counter.

Against v0.20.0 the endpoint returned HTTP 202 in a follow-up probe, indicating the endpoint may have moved to an async job model — inconclusive without observing pod stability across the full commit window and confirming the composite_l2_missing remediation guidance was updated to match.

## Out of scope

- The LLM grouping pass's quality / prompt shape (this signal is about it running synchronously on the request thread, not about what it produces).
- The catalog entries pointing at non-spec upstream URLs (filed as a separate signal about the HTML-portal URL pattern).
- Client-side timeouts (some clients would have given up before the 30s crash; the crash is server-side regardless).

---

Parent goal: #221
Parent initiative: #2364

damir-topic:
## Summary

- Regression pin for the v0.8.0 large-spec ingest crash (a `dry_run=false` ingest of a real vendor OpenAPI spec starved the kubelet liveness probe on the request thread and the pod was graceful-killed mid-ingest). The crash mechanics were fixed on main incrementally — `202` + background job (#2275 / #2317) and parse/proto-build in `asyncio.to_thread` — but the **event-loop responsiveness of the non-dry-run 1275-op path** was never pinned end-to-end. This adds that proof.
- **Test-only; no runtime change.** The production fix already shipped.
- The large spec is synthesised in-memory (1275 GET ops) rather than committed as a ~7 MB fixture; the DB-commit responsiveness proof needs op count, not real vendor bytes. Runtime stays ~2 s so the unit perf budget is unaffected.

Three pins (`backend/tests/test_ingest_large_spec_event_loop.py`):

1. **`test_nondryrun_large_spec_ingest_keeps_event_loop_responsive`** — DB-backed non-dry-run ingest of a 1275-op spec driven concurrently with a heartbeat coroutine that measures event-loop lag. Asserts all 1275 ops commit and the loop keeps ticking throughout (≥50 ticks, no single gap past a 0.5 s ceiling). A non-yielding commit pass would collapse the tick count / spike a gap.
2. **`test_blocking_parse_stays_off_the_event_loop`** — deterministic companion: injects a fixed 0.3 s blocking parse and proves `asyncio.to_thread` keeps it off the loop (max gap far below the block, ≥15 responsive ticks during it). CPU-speed-independent; removing the `to_thread` hop fails it (verified: on-loop → `max_gap≈0.33 s`).
3. **`test_catalog_entry_ingest_rides_the_async_job_path`** — pins that the `--catalog` (`catalog_entry`) shape resolves onto the same `202` + job-handle async path (`async_=True` / `dry_run=False`), so the catalog next-step remediation hint never rides the synchronous pass that crashed the pod.

Confirmed no stale `composite_l2_missing` / `catalog_command` remediation strings remain in `src/` (that apparatus was removed in #2259); the surviving `--catalog` references are the legitimate working verb riding the async default.

## Test plan

- [x] `mypy` — clean
- [x] New tests pass in isolation, 6× consecutive runs — no flakiness (~9 s each)
- [x] Heartbeat harness self-check: on-loop block → `max_gap=0.300s`; off-loop `to_thread` → `max_gap=0.011s`
- [x] Teeth verified: forcing the parse on-loop (removing `to_thread`) makes the deterministic pin fail (`max_gap≈0.33s > 0.15s` ceiling)
- [x] Full backend unit suite (`uv run pytest -q`, default xdist) — green

## Out of scope

- The LLM grouping pass quality/prompt shape (grouping is stubbed; it needs an LLM and is unrelated to the responsiveness proof).
- Durable cross-restart ingest jobs (v0.9 follow-up per the jobs registry docstring).

Closes #2333

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when ingesting large OpenAPI specifications.
  * Prevented long-running ingestion and parsing work from blocking application responsiveness.
  * Ensured catalog-entry ingestion follows the correct asynchronous processing path.

* **Tests**
  * Added regression coverage for large, non-dry-run ingests, event-loop responsiveness, and asynchronous catalog-entry processing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,22 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Tested — regression pin: a non-dry-run 1275-op-class spec ingest keeps the event loop responsive (#2333)
+
+- Pinned the v0.8.0 large-spec ingest crash fix end-to-end: a
+  **non-dry-run** ingest of the 1275-op class (the canonical 7.5 MB
+  vmware/9.0 signal) now has a regression test proving the event loop
+  stays responsive while every operation is committed. A heartbeat
+  coroutine measures loop lag concurrently with
+  `IngestionPipelineService.ingest(dry_run=False)`; the test asserts all
+  1275 ops persist and the loop keeps ticking with no starvation. A
+  deterministic companion injects a fixed blocking parse and proves the
+  `asyncio.to_thread` offload keeps it off the loop, and a third pins
+  that the `--catalog` (`catalog_entry`) shape resolves onto the same
+  `202` + job-handle async path rather than the synchronous pass that
+  crashed the pod. Test-only; no runtime change (the fix shipped in
+  #2275 / #2317).
+
 ### Added — structured post-approval vault-write-forbidden error + write-capability warning on the approval envelope (#2331)
 
 - A typed `vault.kv.put` / `patch` / `delete` that Vault denies at
--- a/backend/tests/test_ingest_large_spec_event_loop.py
+++ b/backend/tests/test_ingest_large_spec_event_loop.py
@@ -1,0 +1,409 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""Regression pin: a large-spec **non-dry-run** ingest keeps the event loop responsive.
+
+At v0.8.0 the ``dry_run=false`` ingest of a real-world OpenAPI spec (the
+canonical signal: 7.5 MB / 1275 typed ops) ran the parse + DB-commit +
+LLM-grouping passes synchronously on the request thread. It starved the
+kubelet liveness probe for ~30 s and the pod was graceful-killed
+mid-ingest (exitCode 0 / reason "Completed" — not OOM). The documented
+remediation for a missing L2 composite pointed operators *at* that
+command, so every affected composite became a dead-end.
+
+The crash mechanics were fixed on main incrementally:
+
+* the route hands the pipeline off the request thread via
+  ``asyncio.create_task`` and returns ``202`` + a job handle
+  (#2275 / PR #2317);
+* the synchronous parse + proto-build run inside ``asyncio.to_thread``
+  (``operations/ingest/pipeline.py`` ``_register_one_spec``);
+* YAML uses ``CSafeLoader``.
+
+What was never pinned end-to-end is the *event-loop responsiveness* of a
+**non-dry-run** ingest of the 1275-op class — specifically that the
+DB-commit pass yields (it awaits per-op, one commit per spec) rather than
+blocking the loop for the whole batch. This module is that pin. Two of
+the tests drive :meth:`IngestionPipelineService.ingest` (``dry_run=False``)
+concurrently with a heartbeat coroutine and assert the loop kept making
+progress; the third pins that the ``catalog_entry`` (``--catalog``) shape
+resolves onto the same async job path rather than a separate synchronous
+one.
+
+These are the failure modes each test would catch if the fix regressed:
+
+* :func:`test_nondryrun_large_spec_ingest_keeps_event_loop_responsive` —
+  a register/commit pass that stopped yielding (one big synchronous
+  batch) would collapse the heartbeat's tick count and spike a single
+  gap past the ceiling.
+* :func:`test_blocking_parse_stays_off_the_event_loop` — dropping the
+  ``asyncio.to_thread`` hop (running the parse on the loop) would freeze
+  the loop for the whole parse; the deterministic injected block makes
+  that crisp and CPU-speed-independent.
+* :func:`test_catalog_entry_ingest_rides_the_async_job_path` — a
+  ``--catalog`` request that fell onto the synchronous branch would
+  reintroduce the crash for the exact command the catalog next-step hint
+  prints.
+
+The pipeline is driven directly (not through the HTTP 202 wrapper): the
+blocking risk lives inside ``ingest()``; the ``create_task`` wrapper only
+detaches it. The grouping phase is stubbed to an empty result — it needs
+an LLM and is out of scope for the responsiveness proof — while the real
+parse + register passes run against the autouse-migrated sqlite engine.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import time
+import uuid
+from collections.abc import Awaitable, Iterator
+from typing import Any
+from unittest.mock import AsyncMock
+
+import pytest
+
+from meho_backplane.auth.operator import Operator, TenantRole
+from meho_backplane.connectors.registry import clear_registry
+from meho_backplane.db.engine import get_sessionmaker
+from meho_backplane.operations.ingest import GroupingResult, SpecSource
+from meho_backplane.operations.ingest.api_schemas import IngestRequest
+from meho_backplane.operations.ingest.pipeline import IngestionPipelineService
+
+_PRODUCT = "loadtest"
+_VERSION = "1.0"
+_IMPL_ID = "loadtest-rest"
+_CONNECTOR_ID = f"{_IMPL_ID}-{_VERSION}"
+
+#: The op count the canonical 7.5 MB vmware/9.0 signal carried. Kept at
+#: the real number so this is a proof for the "1275-op class", not a
+#: scaled-down proxy.
+_OP_COUNT = 1275
+
+#: Heartbeat cadence. The loop is polled every 10 ms; a gap materially
+#: larger than this means the loop was starved for that long.
+_HEARTBEAT_INTERVAL_S = 0.01
+
+#: Ceiling on any single observed heartbeat gap during the real
+#: end-to-end ingest. The parse (in ``to_thread``) contends for the GIL
+#: and produces a single ~0.1 s gap locally; the register pass yields
+#: per-op with gaps <= ~0.03 s. 0.5 s is ~5x over the observed worst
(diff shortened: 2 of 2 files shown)


damir-topic:
@coderabbitai review

damir-topic:




**Anchor task:** #2333 — regression-pin large-spec non-dry-run ingest event-loop responsiveness (7 MB / 1275-op class)
**Surface area:** backend (test-only)
**Verdict:** APPROVE
**Head SHA reviewed:** `18b3263f`
**AI reviewer (CodeRabbit):** received — 0 actionable findings (walkthrough + 5 pre-merge checks passed)

### Acceptance criteria
- [x] Non-dry-run 1275-op-class ingest driven end-to-end with a concurrent heartbeat proving event-loop responsiveness — `test_nondryrun_large_spec_ingest_keeps_event_loop_responsive`, verified: all 1275 ops persist, ~165 ticks, max_gap ~0.11s.
- [x] Deterministic pin that the blocking parse stays off-loop via `asyncio.to_thread` — `test_blocking_parse_stays_off_the_event_loop`, verified CPU-independent.
- [x] `--catalog` (`catalog_entry`) shape rides the 202 + job async path — `test_catalog_entry_ingest_rides_the_async_job_path`, mirrors the real route branch `dry_run or not async_`.
- [x] No stale `composite_l2_missing` / `catalog_command` remediation strings reintroduced — grep across `src/` and the test file: none.

### Project-board alignment
|---|---|---|---|---|---|
| #2333 | In Review | Closes | ✅ | ✅ full | close on merge |

### Predicted vs. actual diff
- ✅ `backend/tests/test_ingest_large_spec_event_loop.py` (new pin) — expected and correct
- ✅ `CHANGELOG.md` (`### Tested —` entry, unique header) — expected and correct
- 🕳️ none missing — test-only is the correct shape; the runtime fix already shipped (#2275 / #2317)
- 🚫 none undisclosed

### Integration verification (all run inside the PR worktree)
- [x] `ruff check` — clean
- [x] `pytest tests/test_ingest_large_spec_event_loop.py` — 3 passed, 3 consecutive runs, no flakiness (~2s ingest, ~7s file)
- [x] **Teeth (TEST2)** — forced parse on-loop (removed `to_thread`) → `max_gap=0.304s > 0.15s` ceiling → FAILS as designed
- [x] **Teeth (TEST1)** — simulated a non-yielding register batch → `ticks=38 < 50` AND `max_gap=1.51s > 0.5s` → FAILS on both assertions as designed
- [x] **Teeth (TEST3)** — assertion mirrors the exact route branch `dry_run or not async_`; verified against `connectors_ingest.py:496` and `IngestRequest.async_` default `True`
- [x] Armed-heartbeat check — harness awaits `armed.wait()` + one tick before work; a head-of-coroutine block surfaces as one long gap (confirmed: the near-head non-yielding-register sim still tripped both assertions), not zero ticks

### Findings (inline)
None.

- Adopted: 0
- Disputed: 0
- Out of scope: 0
- CodeRabbit posted a walkthrough only; all 5 pre-merge checks passed; no inline threads, no nitpicks.

### Docs
- [x] n/a — test-only PR pinning already-documented behavior; no `docs/codebase` doc invalidated.

- TEST1's `max_gap` ceiling (0.5s) is the single closest threshold: forcing the parse on-loop produced 0.46s, which squeaks under it. This is not a gap — the parse-offload regression is TEST2's job and it fails that regression decisively (0.30 > 0.15), while TEST1's own target regression (a non-yielding register) blows past 0.5s (1.51s) and also collapses the tick count. Both tests together cover the surface. If a future edit ever wanted TEST1 to independently backstop the offload too, a tighter ceiling (~0.35s) would still leave ~3x headroom over the observed 0.11s.

- Genuinely toothed: I confirmed each of the three regressions the docstring claims to catch actually fails the corresponding assertion. The heartbeat-arming comment is load-bearing, not decoration — the arm-before-work sequence works.
- Synthetic in-memory 1275-op spec (no 7 MB committed fixture) is the right call: op count is what the responsiveness proof needs, and it keeps the file in the unit-lane perf budget (~2s).
- CHANGELOG uses a distinct `### Tested —` header, avoiding the `[Unreleased]` merge-race.

Issue #1976, opened 2026-06-19
Author minimal Apache-2.0 specs + ExecutionProfiles for the spec-only catalog rows (vmware/sddc)


damir-topic:
Parent goal: #1964
Parent initiative: #1966
Depends-on: #1975, #1969

## Problem

With the catalog mechanism in place (T1 · #1975), author the actual MEHO-authored minimal specs + matching `ExecutionProfile`s for the spec-only catalog rows whose upstream the backend can't dereference (vmware/sddc), removing the forced-upload friction for those products.

## Current state (verified in code, v0.18.0)

- The shipped specs are dry-run-parsed by the **same** parser that rejects Swagger 2.0 and cross-doc `$ref` — a spec that trips `_validate_openapi_version` or the cross-doc-`$ref` guard ([`operations/ingest/openapi.py`](backend/src/meho_backplane/operations/ingest/openapi.py)) **crashes boot**.
- The repo already ships Apache-2.0 evoila-authored package data (`catalog.yaml`) and blesses hand-authored minimal OpenAPI as the spec-less on-ramp (Initiative #1529 · #1533).

## Desired state

- Hand-author minimal, self-contained, `$ref`-local OpenAPI 3.x for vmware/sddc — **only the needed read ops**, vendor-neutral descriptions, SPDX `Apache-2.0` + evoila header.
- Author the matching `ExecutionProfile`s (Basic for sddc; the appropriate scheme for vmware).
- Verify each parses under the boot-time validator and ingests via the catalog-driven path without upload.

## Acceptance criteria

- [ ] Minimal Apache-2.0 specs + matching profiles ship for the vmware/sddc spec-only rows.
- [ ] Each spec parses under the boot-time validator (OpenAPI 3.x, self-contained, `$ref`-local).
- [ ] The catalog_ingest forced-upload friction is gone for those rows (catalog-driven ingest succeeds end-to-end).

## Out of scope

- Full vendor specs (wheel bloat + copyright) — keep the `upstream` pointer for provenance + full-surface re-ingest.

## Legal note

A MEHO-authored minimal spec copies vendor path/param/field names verbatim (it must, to dispatch) — arguably derivative. Flag for a SECURITY/legal glance rather than asserting it a pure non-issue.

## References

- Parent #1964 / #1966. Prior art: Initiative #1529 / #1533. Needs the mechanism (T1 · #1975) and the profile schema (T3 · #1969).

Issue #1361, opened 2026-05-30
Task: G0.18-T8 VCF-family --catalog ingest — recheck #1211 upstreams + refine over-promising next_step hints


damir-topic:
Parent initiative: #1353

## Problem (SEV-4, RDC #789 finding N8)

The VCF-family catalog entries can't be ingested via `--catalog`, and the `next_step` hints over-promise that they can.

Current state (consumer-observed on v0.8.1):
- `vmware/9.0` + `sddc-manager/9.0` `--catalog` ingests **fail**: the catalog upstream points at the Broadcom developer-portal HTML page (`developer.broadcom.com/xapis/…`, `Content-Type=text/html`) → "expected OpenAPI YAML or JSON". (The error itself is well-shaped; the catalog *data* is the problem.)
- `nsx/4.2` → structured 422 `catalog_entry_templated_upstream` (`https://<nsx-mgr-fqdn>/api/v1/spec/openapi/nsx_api.yaml` is per-instance, unresolvable server-side; supply a concrete URI).
- #1211 (G0.15-T2) already fixed the vmware/sddc upstream URLs + audited spec_info_version once → this is a **possible regression**.
- The `next_step` hints for `sddc-rest`/`nsx-rest` still say *"spec available in catalog; run ingest"* — but those ingests fail. The hint over-promises; it should point at `--spec <concrete-uri>` for the templated/HTML cases.

(Independently blocked by N9 — even a resolvable spec can't complete grouping without an injected LLM client — but that's a separate Task; this Task is the catalog-data + hint-accuracy half.)

## Desired state

Either the VCF catalog upstreams resolve to real specs again (if #1211 regressed), or — where the upstream is fundamentally HTML/templated — the catalog entry + `next_step` hint honestly direct operators to `--spec <concrete-uri>` instead of promising a working `--catalog` ingest.

## Acceptance criteria
- [ ] Recheck the `vmware/9.0` + `sddc-manager/9.0` catalog upstream URLs against #1211's fix; if regressed, restore a resolvable spec URL (or mark the entry as `--spec`-only with rationale).
- [ ] The `next_step` hint for catalog entries whose upstream is HTML/templated (sddc-rest, nsx-rest, and any other so-classified) points at `meho connector ingest --spec <concrete-uri>`, not "run `--catalog` ingest".
- [ ] Test asserting the `next_step` for a templated/HTML-upstream entry references `--spec` (not `--catalog`).
[...]

Issue #1211, opened 2026-05-27
Task: G0.15-T2 Catalog vmware/9.0 + sddc-manager/9.0 upstream URL fix + spec_info_version:null audit (sub-signal B)


damir-topic:
Parent goal: #221
Parent initiative: #1209

## Problem

`claude-rdc-hetzner-dc#753` sub-signal **B**
(`catalog-vmware-9-0-upstream-points-to-html-portal-not-spec`).
G0.14-T9 #1182 server-resolve works for catalog entries with
directly-resolvable upstreams (`bind9/9.x` → 422
`catalog_entry_typed_connector`; `nsx/4.2` → 422
`catalog_entry_templated_upstream`). But:

```
$ POST /api/v1/connectors/ingest {"catalog_entry":"vmware/9.0","dry_run":true}
→ HTTP 400 {"detail":"could not decode spec: while scanning for the next token
            found character that cannot start any token
            in '<file>', line 33, column 1"}
```

Inspection: the packaged catalog's `vmware/9.0` entry has
`upstream: ["https://developer.broadcom.com/xapis/vsphere-automation-api/latest/",
"https://developer.broadcom.com/xapis/virtual-infrastructure-json-api/latest/"]`
— Broadcom Developer Portal HTML landing pages, not raw OpenAPI
YAML/JSON. The ingest route fetches the URL and tries to YAML-decode
the HTML response; HTML doctype on line 1, opening tags around line
33 — exactly matches the error coordinate.

Likely also affects **`sddc-manager/9.0`** (same `spec_info_version:
null` marker, same Broadcom Developer Portal pattern). Confirm in
implementation.

`harbor/2.x` already carries an equivalent warning in its catalog
`notes`: *"that file is Swagger 2.0, while the G0.7 ingest parser
targets OpenAPI 3.0/3.1 -- convert before ingest"*. `vmware/9.0`
doesn't carry the parallel warning.

## Current state

- Catalog entries live at
  `backend/src/meho_backplane/operations/ingest/catalog/` (per the
  precedent `catalog.py` reads via `importlib.resources` — find via
  `grep -rn "catalog" backend/src/meho_backplane/operations/ingest/`).
- G0.14-T9 #1182 server-resolve at
  `backend/src/meho_backplane/api/v1/connectors_ingest.py` returns
  structured 422 envelopes for known catalog-side failure shapes
  (`catalog_entry_malformed`, `catalog_entry_not_found`,
[...]

PR #2374, opened 2026-07-11
test(connectors): CI trip-wire — every packaged catalog entry dry-run-resolves without a bare-400 (#2334)

### LINKED ISSUE #2334: Task: structured catalog_entry_upstream_not_spec guard + packaged-catalog upstream URL audit + CI dry-run fixture
## Summary

The `POST /api/v1/connectors/ingest` route was extended (G0.14-T9) to accept `{catalog_entry: "<product>/<version>"}` and resolve the upstream spec(s) server-side. The route shape works correctly for other catalog entries — but at least one packaged catalog entry's `upstream` URLs point at HTML developer-portal landing pages rather than raw OpenAPI YAML/JSON, and the server-side resolver falls over with a bare-400 YAML-parser message that gives the operator no actionable remediation. Status against v0.20.0: PARTIAL — the specific catalog entry the signal was filed against now returns a small (9-op) ingestion result on `dry_run=true`, which is progress; the structural gap (HTML-portal URLs + bare-400 error shape for the pattern) is not confirmed fully closed.

## Reproduction / Observed

The route shape works cleanly for other catalog entries:

```
{catalog_entry: "<bind9-family>/<version>"} → 422 catalog_entry_typed_connector
    (clean structured error: typed connector, no ingestable spec)

{catalog_entry: "<nsx-family>/<version>"}   → 422 catalog_entry_templated_upstream
    (clean structured error: fqdn-templated URL cannot be
     resolved server-side; operator supplies concrete spec)
```

But for one packaged entry (a large REST-API family whose vendor publishes docs on a developer portal):

```
{catalog_entry: "<product>/<version>"}
→ HTTP 400 {"detail": "could not decode spec: while scanning for the
                       next token\nfound character that cannot start
                       any token\n in \"<file>\", line 33, column 1"}
```

Inspecting the packaged catalog (`GET /api/v1/connectors/catalog`) shows the entry's upstream URLs are HTML developer-portal pages, not raw OpenAPI YAML/JSON:

```json
{
  "product": "<product>",
  "version": "<version>",
  "impl_id": "<impl>",
  "upstream": [
    "https://<vendor-developer-portal>/xapis/<api-family-1>/latest/",
    "https://<vendor-developer-portal>/xapis/<api-family-2>/latest/"
  ],
  "spec_info_version": null,
  "sha256": null,
  "notes": "Generic-ingested. ... spec_info_version pending the first
            verified ingest."
}
```

The server-side resolver fetches the URL and tries to YAML-decode the response. The response is HTML (the developer-portal landing page). HTML doctype on line 1, opening tags around line 33 — exactly matches the "line 33, column 1" error coordinate. The failure is the catalog-entry CONTENT, not the route shape.

Compare with another packaged entry which calls out a similar sharp-edge in its `notes` field:

```
"that file is Swagger 2.0, while the ingest parser targets
 OpenAPI 3.0/3.1 — convert before ingest"
```

The affected entry doesn't carry an equivalent warning that its HTML-portal upstream URLs aren't directly resolvable.

Likely also affects sibling entries in the same product family that share the `spec_info_version: null` marker and point at the same developer portal — worth an audit of every packaged catalog entry whose `upstream[]` isn't a raw-content URL.

## Why this matters

- The bare-400 with a YAML-parser error is uninformative. An operator hitting it sees "could not decode spec … line 33, column 1" and reasonably concludes the upstream YAML is malformed, then tries to file the bug against the vendor. In reality the catalog entry is pointing at HTML.
- The route already knows how to emit a structured 422 for other unresolvable patterns (`catalog_entry_templated_upstream`, `catalog_entry_typed_connector`). This case is missing from that error taxonomy.
- Because the packaged catalog is the canonical remediation surface referenced by the `composite_l2_missing` error envelope, a broken catalog entry poisons an otherwise well-designed error-UX chain.

- **Catalog content fix.** Replace the affected entry's upstream URLs with directly-resolvable spec URLs. For REST APIs that ship the spec on the appliance itself, point at the on-appliance path; where the vendor publishes a machine-readable mirror on GitHub or a CDN, point there instead. Land the `sha256` + `spec_info_version` after the first verified ingest.
- **Schema extension.** Add a `requires_concrete_spec: true` flag (or extend the existing `catalog_entry_templated_upstream` shape) so the route can return a structured 422 like other unresolvable patterns do, with a remediation step naming the explicit-quadruple shape the operator should use.
- **Route-level guard.** Detect non-YAML/JSON content-type in the upstream fetch (an `HTTP Content-Type: text/html` response is unmistakable) and return a more specific error such as `catalog_entry_upstream_not_spec` rather than the generic "could not decode spec". Trip-wire for any future catalog entry drift into the same pattern.
- **Catalog audit + notes field.** Sweep every packaged catalog entry; for any whose `upstream[]` is an HTML portal / documentation page rather than a raw spec, either fix the URL (preferred) or add an explicit `notes` warning matching the precedent of other sharp-edge entries.
- **Test data.** Add a fixture that dry-run-ingests each packaged catalog entry and asserts it either succeeds or returns a structured 422 — never a bare-400 YAML-parser error. Catches this class of catalog drift at CI time.

## Verification

Auto-runnable — the source signal ships a curl-driven probe.

```bash
resp=$(curl -ksS -X POST https://<backplane>/api/v1/connectors/ingest \
  -H "Authorization: Bearer $JWT" -H "Content-Type: application/json" \
  -d '{"catalog_entry":"<product>/<version>","dry_run":true}')
```

- FIXED (exit 0): either dry-run succeeds cleanly (`inserted_count` present) OR the route returns structured 422 with one of the recognized contract codes (`catalog_entry_templated_upstream`, `catalog_entry_upstream_not_spec`, `requires_concrete_spec_*`).
- OPEN (exit 2): bare-400 with `could not decode spec` / `invalid yaml` / `invalid json`.

Against v0.20.0 the probe returned `HTTP 200` with `inserted_count: 9` for the specific entry originally filed. That is progress — the entry now resolves to *something* — but doesn't confirm (a) that the underlying catalog content has been fixed to point at real specs and not a partial fallback, (b) that other same-pattern entries in the packaged catalog have been swept, or (c) that a non-resolvable HTML-portal URL now returns a structured error rather than the bare-400 shape.

## Out of scope

- The route shape / accepting `{catalog_entry: "..."}` (that works correctly and is not the bug).
- Any specific vendor's documentation strategy — this signal is about how the catalog references upstream specs, not what the vendor publishes.
- The pod-crash-on-large-real-spec issue filed as a separate signal (dry-run cannot trigger that crash even against a correctly-pointed URL).

---

Parent goal: #221
Parent initiative: #2364

damir-topic:
## Summary

- The #2334 signal was a packaged connector-catalog row whose `upstream` pointed at an HTML developer-portal page: the resolver fetched HTML, YAML-decoded it, and surfaced an opaque bare-400 (`could not decode spec … line 33`) with no actionable remediation.
- The structured `catalog_entry_upstream_not_spec` content-type guard (G0.15-T2 #1211) already closes that at request time. This PR adds the **CI guardrails** that keep the whole packaged catalog inside the success-or-structured contract so a future row (or URL edit) can't regress into the bare-400 pattern.
- **Route fixture** (`test_dry_run_every_packaged_catalog_entry_is_success_or_structured_never_bare_400`): parametrized over every shipped catalog entry, dry-run-ingests each with all fetchable upstreams mocked to serve worst-case HTML (deterministic, no live network), and asserts each resolves to a dry-run success or a **structured** error envelope (a `detail`/`kind` snake_case classifier) — never the opaque decode-400 or an unhandled 500.
- **Static audit** (`test_packaged_catalog_fetch_upstreams_are_raw_spec_urls_not_doc_portals`): asserts every fetch-path `upstream` is a raw OpenAPI URL (`.yaml`/`.yml`/`.json`), not a documentation portal — generalizing the prior Broadcom-only sweep.

## Audit finding

The shipped catalog is **already clean**. Only `harbor` (raw.githubusercontent.com `swagger.yaml`) and `gh` (raw.githubusercontent.com `api.github.com.json`) reach the fetch path, and both point at raw-content URLs. The Broadcom portal references survive only as fqdn-templated (`nsx`, gated earlier by `catalog_entry_templated_upstream`) or as `notes` provenance pointers on the profile-backed `vmware`/`sddc` rows (#1976). Parts 1 (structured guard) and 2 (catalog content) of the task landed in prior work (#1211, #1361, #1976); this PR delivers part 3 (the CI dry-run fixture) plus the generalized audit guard, and confirms the guard is complete.

## Test plan

- [x] `ruff check` — clean (both changed test files)
- [x] `mypy src/` — n/a (no `src/` files changed; test-only)
- [x] `pytest tests/test_api_v1_connectors_ingest.py tests/test_operations_ingest_catalog.py` — 176 passed
- [x] New fixture: all 9 packaged entries return success-or-structured (no bare-400 / no 500)
- [x] New audit: current catalog passes (harbor + gh are raw-spec URLs)

Full backend suite runs isolated in CI (authoritative gate).

## Out of scope

- The route shape / accepting `{catalog_entry: "..."}` (works correctly; not the bug).
- The pod-crash-on-large-real-spec signal (separate; dry-run can't trigger it).

Closes #2334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,26 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Added — CI trip-wire: every packaged catalog entry dry-run-resolves without a bare-400, plus a raw-spec-URL upstream audit (#2334)
+
+- A packaged connector-catalog entry whose `upstream` pointed at an HTML
+  developer-portal page used to fetch the HTML, YAML-decode it, and surface
+  an opaque bare-400 (`could not decode spec … line 33`) that told the
+  operator nothing actionable. The structured `catalog_entry_upstream_not_spec`
+  content-type guard (G0.15-T2 #1211) already closes that at request time;
+  #2334 adds the CI guardrails that keep the contract from regressing: a
+  parametrized fixture dry-run-ingests **every** shipped catalog entry (with
+  all fetchable upstreams mocked to serve worst-case HTML — deterministic, no
+  live network) and asserts each resolves to a dry-run success or a
+  **structured** error envelope, never the opaque decode-400 or an unhandled
+  500. A companion static audit asserts every fetch-path `upstream` is a raw
+  OpenAPI URL (`.yaml`/`.yml`/`.json`), not a documentation portal —
+  generalizing the prior Broadcom-only sweep. A future row (or URL edit) that
+  drifts into the portal/HTML pattern now trips at unit-test time instead of
+  on an operator's first POST. Audit finding: the shipped catalog is already
+  clean — only `harbor` and `gh` reach the fetch path and both point at
+  raw-content URLs (#2334).
+
 ### Added — structured post-approval vault-write-forbidden error + write-capability warning on the approval envelope (#2331)
 
 - A typed `vault.kv.put` / `patch` / `delete` that Vault denies at
--- a/backend/tests/test_api_v1_connectors_ingest.py
+++ b/backend/tests/test_api_v1_connectors_ingest.py
@@ -39,6 +39,7 @@ from collections.abc import Callable, Iterator
 from pathlib import Path
 from typing import Any
 from unittest.mock import AsyncMock, patch
+from urllib.parse import urlparse
 from uuid import UUID
 
 import httpx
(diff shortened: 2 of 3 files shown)


damir-topic:
@coderabbitai review

damir-topic:




**Anchor task:** #2334 — structured `catalog_entry_upstream_not_spec` guard + packaged-catalog upstream audit + CI dry-run fixture
**Surface area:** backend (test-only)
**Verdict:** APPROVE
**Head SHA reviewed:** `ae4a9cdf`
**AI reviewer (CodeRabbit):** not received — rate-limited (review limit reached; next available ~55 min). No CodeRabbit findings this iteration.

### Acceptance criteria
- [x] Parametrized route test dry-run-ingests **all 9** packaged catalog entries with fetchable upstreams mocked to worst-case HTML, asserting success OR structured envelope, never a bare decode-400/500 — verified: parametrization is **derived** from `load_catalog().entries` (auto-covers future rows), 9 params + 1 audit = 10 tests pass.
- [x] Static audit that every fetch-path `upstream` is a raw-spec URL (`.yaml/.yml/.json`) — passes; only `harbor` + `gh` reach the fetch path, both raw-content URLs.
- [x] Parts 1+2 (structured guard + clean catalog) genuinely pre-exist on main — test-only scope is correct.

### Verification performed (worktree `pr-2374-review`, head ae4a9cdf)
- `ruff check` / `ruff format --check` — clean (both changed test files).
- `pytest tests/test_api_v1_connectors_ingest.py tests/test_operations_ingest_catalog.py` — **176 passed**.
- Per-entry outcomes instrumented (guard enabled): vmware/9.0 + sddc/9.0 → 200 (shipped spec_resource, fetch=0); harbor/2.x + gh/3 → 422 `catalog_entry_upstream_not_spec` (content-type guard exercised on mocked HTML); nsx/9.0 → 422 `catalog_entry_templated_upstream`; vault/k8s/bind9 → 422 `catalog_entry_typed_connector`; _fixture/1.0 → 422 `product_impl_id_mismatch` (keyed under `kind`).
- **Guard pre-existence confirmed:** `_reject_non_spec_content_type` → `UpstreamNotSpecError` → `build_catalog_entry_upstream_not_spec_detail` all present in `src/` (G0.15-T2 #1211). Not a vacuous pass.
- **Non-vacuity confirmed:** the fixture genuinely exercises the content-type guard for the two fetch-path entries (harbor/gh) — if that mapping regressed to a bare-400/500, the test fails.
- **Determinism confirmed:** `respx.mock` blocks any unmocked HTTP (no passthrough) and `getaddrinfo` is monkeypatched to a resolver that only maps the entry's own fetch hosts to a fixed public test IP. No live network is reachable on a CI runner.

### Predicted vs. actual diff
- ✅ `backend/tests/test_api_v1_connectors_ingest.py` — parametrized route trip-wire (expected, correct)
- ✅ `backend/tests/test_operations_ingest_catalog.py` — static raw-spec-URL audit (expected, correct)
- ✅ `CHANGELOG.md` — unique `### Added — CI trip-wire …` header (merge-race-safe)
- 🕳️ none missing — parts 1+2 already on main, so no `src/` change is expected for this task.

### Integration verification
- [x] `ruff check` — clean
- [x] `pytest` (both modules) — 176 passed
- [x] new fixture + audit — 10 passed; per-entry branch outcomes instrumented
- [⚠️ not run] mypy — n/a, test-only, no `src/` change; CI Python lane authoritative
- [⚠️ not run] migrations / docker / UI — n/a for a test-only PR

### Findings
None blocking. No inline comments — the PR is correct and the two adjacent observations below carry no required change.

- The `_HTML_PORTAL_BODY` docstring claims the body reproduces "the exact `line 33, column 1` coordinate … when the HTML fell through the YAML decoder." In practice this body does **not** raise `yaml.YAMLError`: YAML parses `<!doctype html>…` to a scalar `str`, so if the content-type guard were bypassed it would surface `InvalidSpecError` ("must parse to a mapping"), also structured — and for the real fetch-path entries the content-type guard intercepts *before* decode. So the literal bare-400 `yaml.YAMLError` arm (`connectors_ingest.py:1220`) is never reached by the fixture. This does not weaken the contract assertion (the test still forbids any bare decode-marker string and requires a dict envelope for non-200, and the guard path *is* exercised) — but a future maintainer relying on the comment might over-trust decode-arm coverage. Consider tightening the comment to say the body decodes to a non-mapping scalar / is intercepted by the content-type guard, rather than that it trips the line-33 YAML decoder.
- `product_impl_id_mismatch` keys its classifier under `kind` while `catalog_entry_*` uses `detail`; the fixture's `detail.get("detail") or detail.get("kind")` accepts either. This is acceptable — both are branchable snake_case classifiers and the intent is "structured + branchable" without coupling to one shape. No change required.

- Rate-limited — no findings to adopt/dispute/scope this iteration.

### Docs
- [n/a] No `docs/codebase/` change required — behavior (guard + catalog) unchanged; this PR only adds CI trip-wires. Error-shape docs already reference the classifiers.

### What worked well
- Parametrization derived from the live packaged catalog, so new rows are covered automatically instead of via a hard-coded list — exactly the coverage-honesty property the task needs.
- The static URL-suffix audit is a cheap, parse-time trip-wire that catches the drift class earlier than the route test, with a clear remediation message pointing at the `spec_resource` on-ramp (#1976).

PR #2376, opened 2026-07-11
feat(api)!: converge GET-list endpoints on {items, next_cursor} envelope (#2338)

### LINKED ISSUE #2338: Task: converge GET-list endpoints on the {items, next_cursor} envelope — one breaking pass + platform-wide contract test
## Summary

At v0.8.0 the 5 most-used GET-list endpoints disagreed on envelope shape — and against v0.20.0 the same asymmetry persists: 4 distinct envelope shapes across the 5 endpoints. Writing a generic list-endpoint helper against the surface is impossible; every call site needs its own parsing logic. Not blocking for daily-driver use, but recurring friction whenever anyone writes an SDK, script, or agent against the surface. Status against v0.20.0: OPEN.

## Reproduction / Observed

Enumerating the 5 reference list endpoints against v0.20.0:

| Endpoint                          | Envelope shape                                  |
|-----------------------------------|-------------------------------------------------|
| `GET /api/v1/connectors`          | `{"connectors": [...]}`                         |
| `GET /api/v1/targets`             | `[...]` (bare array)                            |
| `GET /api/v1/conventions`         | `{"budget_status": {...}, "entries": [...]}`   |
| `GET /api/v1/audit/my-recent`     | `{"rows": [...], "next_cursor": "..."}`         |
| `GET /api/v1/broadcast/overrides` | `[...]` (bare array)                            |

Four distinct shapes across five endpoints:

- Named-payload key that echoes the endpoint (`{connectors: [...]}`).
- Bare array (`targets`, `broadcast/overrides`).
- Object with a domain-named list + sidecar (`{entries, budget_status}`).
- Object with a canonical `rows` + pagination cursor (`{rows, next_cursor}`).

The `audit/my-recent` shape (`{rows, next_cursor}`) is the closest to industry-standard cursor-paginated shape; the other four are all divergent.

Auto-probe:

```bash
JWT=$(cat /tmp/meho-jwt)
for ep in connectors targets conventions audit/my-recent broadcast/overrides; do
  curl -ksSf "https://<backplane>/api/v1/${ep}" \
    -H "Authorization: Bearer $JWT" \
    | jq -rc 'if type == "array" then "bare-array" else (keys | join(",")) end'
done
```

Observed against v0.20.0:

```
connectors:connectors
targets:bare-array
conventions:budget_status,entries
audit/my-recent:next_cursor,rows
broadcast/overrides:bare-array
```

Four distinct envelope shapes across five endpoints. OPEN.

## Why this matters

- **Adopter friction.** Anyone writing an SDK, script, or agent that talks to multiple list endpoints has to special-case each one. The generic "list resources of kind X" pattern that every REST-consumer library ships is unusable — you can't write `client.list("connectors")` and `client.list("targets")` through the same code path without a per-endpoint shape table.
- **Pagination inconsistency.** Only one of the five endpoints exposes a cursor (`audit/my-recent`). The bare-array endpoints and the named-payload endpoint have no documented way to page. If the underlying data grows (thousands of connectors, thousands of targets, tens of thousands of broadcast overrides), those endpoints have no forward-compatible way to shrink their response. Adopters relying on "the whole list fits in one response" will silently truncate or hit hard limits.
- **Sidecar metadata pattern is only in one shape.** The `conventions` endpoint proves the surface sometimes needs to return list + sidecar (`entries` + `budget_status`). Bare-array endpoints can never grow a sidecar without a breaking change. Adopting a single `{items, next_cursor, [sidecars]}` shape now means new sidecars can be added additively later.
- **Discoverability.** A caller reading the OpenAPI spec (or trial-and-erroring) has no way to guess which shape a given endpoint will return. Each endpoint is its own tiny puzzle.

- **Pick a single convention and migrate.** The cleanest single shape across all five would be `{items: [...], next_cursor: <opt>, [sidecars]}` — industry-standard cursor-paginated list envelope; matches what `audit/my-recent` half-adopted (rename `rows` → `items`); generalizable to any list endpoint; leaves room for sidecar metadata like `budget_status` without breaking future adopters.
- **Deprecate the current shapes with a version window.** Publish the new shape on the same endpoints under an `Accept: application/vnd.meho.v1+json` (or version-in-URL) discriminator, keep the old shape returned for the default `Accept` for one version, then switch the default. Adopters get time to migrate. Alternative shape: add the new shape as a sibling endpoint (e.g. `GET /api/v1/connectors/list`) and deprecate the old one over a release cycle.
- **Document the convention alongside the error-message-shape convention.** Both are cross-cutting surface conventions; one place, one page — so adopters see "here is how error envelopes look" and "here is how list envelopes look" in the same section of the reference.
- **Contract-test the convention going forward.** Add a CI check that every `GET /api/v1/**` route returning a collection matches the chosen envelope shape schema. Cheap invariant to add; prevents new endpoints from introducing a fifth shape.
- **Even if the migration is deferred, add a `list_envelope_shape` field to the OpenAPI extension** on each of the 5 endpoints so machine consumers can dispatch to the right parser without probing. Bridge until the migration lands.

## Verification

Auto-runnable — the source signal ships a probe that enumerates the five endpoints and counts distinct shapes.

- FIXED (exit 0): all 5 list endpoints agree on a single envelope shape (any shape — the finding is about consistency, not about which shape).
- PARTIAL (exit 1): 2 distinct shapes across the 5 endpoints (partial convergence — e.g. two migrated, three still on legacy).
- OPEN (exit 2): 3+ distinct shapes (today's state).

Current state against v0.20.0: OPEN — 4 distinct envelope shapes across the 5 endpoints.

## Out of scope

- Individual endpoint semantics (this signal is only about envelope shape; whether `conventions` returns `budget_status` at all is a separate design decision).
- Cursor-vs-offset pagination debate (either can fit the `{items, next_cursor}` shape; the signal is about picking one shape at all).
- Non-list endpoints (this signal is scoped to GET-list; single-resource GET / write ops have their own consistency questions).

---

Parent goal: #221
Parent initiative: #2364

damir-topic:
## Summary
- **BREAKING.** Converge the seven §2 reference `GET`-list endpoints on the unified `{items, next_cursor?, ...sidecars}` list envelope as the default **and only** shape, retiring the `?envelope=v2` opt-in (and its `wrap_v2_envelope` builder). Endpoints: `targets`, `connectors`, `conventions`, `audit/my-recent`, `broadcast/overrides`, `runbooks/templates`, `runbooks/runs`.
- Each endpoint keeps a **named, typed `response_model`** (`TargetListResponse`, `ConnectorListResponse`, `ConventionListResponse` (+`budget_status` sidecar), `MyRecentPage`, `BroadcastOverridesListResponse`, `RunbookTemplateListResponse`, `RunbookListRunsResponse`), so the OpenAPI 200 schema — and the generated Go client — stay single-schema (no `anyOf` union).
- New **platform-wide contract test** (`test_api_v1_list_envelope_contract.py`) enumerates the governed endpoints and asserts, against the live app, that both the runtime body and the OpenAPI 200 schema are the envelope (never a bare array, never a resource-named list key) — a new list endpoint can't reintroduce a fifth shape.
- **CLI/OpenAPI snapshot + Go client regenerated**; every CLI list consumer (targets / connectors / conventions / audit / broadcast / runbook verbs) traced and updated to read `items`.

## Migration
Read the list from `response["items"]` instead of the old key (`connectors` / `entries` / `rows` / `templates` / `runs`) or the bare array; read `response["next_cursor"]` for pagination (only `targets` populates it today). Clients still sending `?envelope=v2` are unaffected (the now-unknown param is ignored). The `audit` sibling reads (`/query` / `/who-touched` / `/by-work-ref`) and the topology closure reads (§4 `{kind, nodes}`) keep their existing shapes.

## Test plan
- [x] `mypy` on changed src — clean
- [x] Affected backend suites (targets / connectors / conventions / audit / broadcast / runbooks + new contract test + g81 acceptance) — 374 passed, 14 skipped
- [x] New contract test `test_api_v1_list_envelope_contract.py` — 9 passed
- [x] `make snapshot-openapi && make generate` — idempotent, no uncommitted drift
- [x] `go build ./...`, `go test ./...`, `go vet`, `gofmt -l` — clean
- [x] Docs `api-shape-conventions.md` §2 + `CHANGELOG` Breaking-changes entry updated

## Out of scope
- Audit sibling reads (`/query` / `/who-touched` / `/by-work-ref`) — not §2 reference list endpoints (kept `{rows, next_cursor}`); tracked for a future convergence pass.
- Topology closure reads — converge on the §4 `{kind, nodes}` shape, keep their own `?envelope=v2` opt-in.
- Other legacy bare-array reads (`conventions/{slug}/history`, `topology/edges`, `approvals`, `agent/runs`, `doc-collections`) — out of §2 scope.

Closes #2338

## Summary by CodeRabbit

* **Breaking Changes**
  * Seven list endpoints now consistently return `{items, next_cursor, ...sidecars}` envelopes.
  * Legacy resource-specific fields such as `runs`, `templates`, `connectors`, `entries`, and bare arrays are replaced by `items`.
  * The `?envelope=v2` opt-in is retired; existing parameters are ignored.
  * `targets` supports cursor pagination through `next_cursor`.

* **Documentation**
  * Updated API documentation, OpenAPI specifications, generated Go client, and CLI behavior to reflect the unified response format.
  * Added migration guidance for reading `items` and `next_cursor`.

* **Tests**
  * Added contract coverage validating response envelopes and OpenAPI consistency.


damir-topic:
@coderabbitai review

damir-topic:




**Anchor task:** #2338 — converge the seven §2 GET-list endpoints on the unified `{items, next_cursor}` envelope (BREAKING)
**Surface area:** backend + devops (Go CLI / codegen)
**Verdict:** APPROVE
**Head SHA reviewed:** `edd98b11`
**AI reviewer (CodeRabbit):** not received — rate-limited ("Review limit reached, next review in 12 min" at 16:24Z); the `CodeRabbit: SUCCESS` check only acknowledges the trigger, no review body or inline comments were produced. Known back-to-back auto-run rate-limit. My own Phase 0–6 review stands in.

> Self-review is structurally blocked (reviewer == PR author), so this verdict is posted as a PR comment rather than a formal `APPROVE` review.

### Acceptance criteria
- [x] All 7 endpoints return a **named, single-schema** `{items, next_cursor?}` response_model (`TargetListResponse`, `ConnectorListResponse`, `ConventionListResponse`+`budget_status` sidecar, `MyRecentPage`, `BroadcastOverridesListResponse`, `RunbookTemplateListResponse`, `RunbookListRunsResponse`) — verified in source + OpenAPI.
- [x] OpenAPI 200 for each of the 7 is a single `$ref` — **no `anyOf`/`oneOf`/`allOf`** (scripted check over `cli/api/openapi.json`).
- [x] `?envelope=v2` opt-in + `wrap_v2_envelope` retired for §2 list endpoints; `test_api_v1_list_envelope_v2.py` deleted; `_envelope.py` now scoped to the topology §4 closure reads only.
- [x] Contract test (`test_api_v1_list_envelope_contract.py`) enumerates all 7 governed endpoints and asserts, against the live app, both runtime body (items is a list, next_cursor present, no legacy key, never bare array) AND the OpenAPI 200 schema — 9 passed.
- [x] Every CLI/Go consumer reads `.Items`: targets, connectors (raw `connectorListEnvelope.Items` json:"items"), conventions, audit/my-recent, broadcast/overrides, runbook templates+runs, and `catalog.go registeredTriples` (reads connectors `.Items`). No stale `.Connectors`/`.Templates`/`.Runs`/`.Rows`/`.Entries` decode of these 7.
- [x] Snapshot fresh: `make snapshot-openapi && make generate` → **zero drift** (whole-tree `git status` clean after regen).
- [x] CHANGELOG BREAKING entry + full migration recipe; `docs/codebase/api-shape-conventions.md` §2 updated (converged table, "how it converged", scope exclusions).
- [x] Excluded endpoints defensible: audit `/query` `/who-touched` `/by-work-ref` keep `AuditQueryResult` `{rows}` (not §2 reference lists); topology closure keeps §4 `{kind, nodes}`.

### Integration verification (all run in worktree on head `edd98b11`)
- [x] `go build ./...`, `go vet ./...`, `gofmt -l` — clean
- [x] `go test ./...` — pass
- [x] `make snapshot-openapi` + `make generate` — zero uncommitted drift
- [x] backend `test_api_v1_list_envelope_contract.py` — 9 passed
- [x] affected backend suites (targets/conventions/broadcast/audit/runbook-runs/runbook-templates/connectors-delete/connectors-ingest/g81 acceptance) — 365 passed, 14 skipped
- [x] `ruff check` + `ruff format --check` + `mypy` on changed src — clean

### Predicted vs. actual diff
- ✅ 7 endpoint routers + their schema modules → named response_models
- ✅ `_envelope.py` scoped down to topology-only; v2 opt-in tests removed
- ✅ regenerated `cli/api/openapi.json` + `client.gen.go`; all 7 CLI consumers + tests updated to `.Items`
- ✅ contract test added; CHANGELOG + docs updated
- ❓ (justified) `test_api_v1_connectors_ingest.py` touched — envelope-shape assertions; **also a merge-conflict point, see below**
- 🕳️ none missing

### ⚠️ CONFLICTING — CI did not run against the merge ref
`mergeable_state: dirty`. **Two** files conflict with `main`:
1. `CHANGELOG.md` — the standard `[Unreleased]` merge race (union-resolve).
2. `backend/tests/test_api_v1_connectors_ingest.py` — collides with `main`'s #2334 ("CI dry-run fixture over all packaged catalog entries"). This is a **real test-file conflict**, not just CHANGELOG — resolve it carefully (merge-main-in + reconcile both sides' assertions), not a blind union.

CI green here is against the PR head, **not** the merged result. Orchestrator must merge-main-in, re-run CI, and confirm green before merging.

### Findings (non-blocking — no merge-blocking inline comments posted)
- m1 (Minor, not inline) — `cli/internal/cmd/connector/list.go` `getList()` leading comment (~L208–216) is now **stale/self-contradictory**: it says the endpoint "deliberately returns `dict[str, list[dict[str,object]]]` (no `response_model`), so … `JSON200` is an untyped map." This PR converged the endpoint to `response_model=ConnectorListResponse`, and both the struct-level comment (L44) and the generated client (`JSON200 *ConnectorListResponse`) now contradict it. Runtime is correct (`connectorListEnvelope.Items` reads `json:"items"`). Not posted inline because the stale lines fall outside this PR's diff hunks. **Suggested fix:** update the `getList` comment to state the route now declares `response_model=ConnectorListResponse` and that the raw-`Body` decode into `connectorListEnvelope` is a deliberate choice for a hand-stable field set (matching the `listEntry` doc), not because `JSON200` is untyped.

- The contract test's forward-guard (`_LIST_ENVELOPE_ENDPOINTS`) is a **manually-maintained registry** — a brand-new GET-list endpoint returning a bare array is only caught if a developer adds it to the registry; it is not auto-detected. The test docstring is honest about this ("joins the convention by … being added to `_LIST_ENVELOPE_ENDPOINTS`"), but the PR-body claim "a new list endpoint can't reintroduce a fifth shape" slightly overstates it. Auto-detecting "which GET is a list-of-resources" is genuinely hard, so the registry approach is defensible; noting only that the guarantee is developer-discipline-backed, not automatic.

- Adopted: 0 · Disputed: 0 · Out of scope: 0 — CodeRabbit rate-limited, no findings produced this iteration.

### Docs
- [x] `docs/codebase/api-shape-conventions.md` §2 — updated (converged-endpoints table, migration narrative, scope exclusions).

### What worked well
- Every endpoint keeps a **named** response_model, so the OpenAPI 200 stays single-schema and the Go client is a concrete type — no `anyOf` union to unpack. Exactly what keeps the breaking change clean for typed downstream clients.
- The contract test asserts both the **runtime body** and the **generated OpenAPI schema**, catching client-contract drift a runtime-only test would miss.
- CLI↔backend contract completeness is solid: all 7 consumers + `catalog.go`'s `registeredTriples` trace cleanly to `.Items`; snapshot regen produced zero drift.

Issue #2346, opened 2026-07-10
Task: render backend 4xx/5xx errors in the /ui/agents/create form — runbooks-form error mold


zdamir:
## Summary

The `/ui/agents/create` Create-agent modal's form submit swallows backend HTTP 422 validation responses — the form's htmx swap has no error handler, so validation failures show **nothing** to the operator. The button appears to "do nothing" while DevTools Network shows the actual failure.

Observed live 2026-07-10 on v0.20.0.

## Reproduction

1. Authenticate to `/ui` as `tenant_admin` (any operator).
2. Navigate to `/ui/agents`. Open DevTools → Network tab.
3. Click **+New Agent** to open the Create modal.
4. Fill:
   - name: `ui-probe-1234`
   - system_prompt: some text
   - model_tier: `fast`
   - **identity_ref**: `test` ← intentionally invalid (backend expects `agent:<name>` shape)
   - other fields as defaults
5. Submit.
6. Observe:
   - Network tab: `POST /ui/agents/create` → HTTP 422 with structured JSON detail (e.g. `{"detail": [{"loc": ["body", "identity_ref"], "type": "value_error", ...}]}`).
   - Modal UI: **nothing changes.** No error banner, no field highlight, no toast. The form just sits there with the operator's input still filled in.

The operator has to open DevTools to know why the create failed.

## Why this matters

- **Silent failure on a write.** Same class as the previously-flagged `evoila/meho#2117` D3 (empty step body silently accepted) which was fixed via `#2122`'s `min_length=1` — that fix surfaced the validation error via the form's error path. The `/ui/agents/create` form has no equivalent error-rendering path.
- **Operators without DevTools can't debug.** A first-time operator hitting a mistyped `identity_ref` sees the form "not doing anything" and has no clear path forward. Might retry with the same input, might change random fields to see what "unblocks" the submit.
- **Snowballs into the CSRF-drift defect.** After ~3 retries the operator hits the CSRF 403 (companion filing) — also silent. Now they have TWO invisible failures compounding.

## Fix shapes (each independently actionable)

- **Option A — render backend errors in the modal**: on `POST /ui/agents/create` returning 4xx/5xx, htmx should swap the error response into an alert-error div at the top of the form. Follow the same pattern the runbook Create form already uses (per `#2117` D3's fix pattern) — the alert-error banner reads:

  > `steps.0.manual.body: String should have at least 1 character`

[...]

Issue #2345, opened 2026-07-10
Task: unify the /ui CSRF token pattern — rotate-on-render vs static-reuse drift 403s the 3rd write in a session


zdamir:
## Summary

The double-submit CSRF pattern used on cookie-authenticated `/ui/*` write endpoints (`meho_csrf` cookie → `X-CSRF-Token` header echo) works on the first ~2 writes in a session but **drifts out of sync on the 3rd repeated write**, returning:

```
HTTP 403 Forbidden
body: {"detail":"csrf_token_invalid"}
```

A hard page refresh (Cmd+Shift+R) restores a working token pair, and the next write succeeds. The failure is silent in the UI (see companion filing about the `/ui/agents/create` form swallowing 4xx responses).

Observed live 2026-07-10 on v0.20.0 via `POST /ui/agents/create` in the browser. This defect had been declared FIXED on v0.19.0 (an earlier Playwright pass showed a single `POST /ui/memory/create` returning 204), but the earlier pass didn't stress the repeated-writes pathology.

## Reproduction

1. Authenticate to `/ui` as any operator via Keycloak SSO.
2. Navigate to `/ui/agents`. Click **+New Agent** to open the Create modal.
3. Submit the form 3 times in quick succession (fill dummy values; any invalid input works since we're testing the CSRF layer, not the field validation).
4. Observe:
   - Attempts 1-2: submit hits the backend (may 422 on validation, but reaches the handler).
   - Attempt 3+: `POST /ui/agents/create` → **HTTP 403 `csrf_token_invalid`**.
5. Hard-refresh (Cmd+Shift+R) → next submit succeeds.

## Mechanism (from prior consumer-side investigation)

On the first request the browser sends `meho_csrf` cookie (JS-readable, double-submit pattern). The backend validates against an `X-CSRF-Token` header. The intended pattern:

- Server sets `meho_csrf` cookie on session establish.
- Client-side code (an htmx `configRequest` hook or equivalent) reads the cookie and attaches it as `X-CSRF-Token` on every write.

But something goes wrong on **repeated writes in the same session**:

- Hypothesis A: the server rotates `meho_csrf` on write (single-use tokens), but the client doesn't re-read the cookie for each request — it caches the first value it saw.
- Hypothesis B: the client sends the current cookie value, but the server rotates it *after* validating, and there's a race where the client's cached copy is 1-off from server's expected next value.
[...]

Issue #2244, opened 2026-07-09
Task: reject input-less one_off/cron scheduler triggers at create time (422) — payload-only check; closes the inputs:{} literal-prompt edge


zdamir:
## Summary

Re-diagnosed from the original filing: the structured `scheduled_run_no_input` fire-time error is **not a v0.20.0 partial fix** — it shipped in v0.11.0 (PR #1513, closing #1505 from ops' own v0.10.1 observation), and fail-typed-at-fire was the maintainers' *decided* resolution, with reject-at-create and inject-default-turn both explicitly considered and rejected (PR body, two code docstrings, `docs/codebase/scheduler.md`). What survives grounding: the create-time footgun is genuinely still open, the design rationale protecting it ("a definition that legitimately needs no user turn") defends a shape the code itself documents as not-yet-supported, and a **narrowed** create-time check dodges #1513's layering objection because it needs **no definition load** — it's payload-only. It also closes an unreported edge: `inputs: {}` bypasses the fire-time guard and sends the literal string `"{}"` as the user prompt. A no-inputs cron fails every fire; a no-inputs one_off burns its at-most-once fire.

## Current state (verified in code)

- Fire-time guard + typed error: scheduler tick logs `scheduler_fired_run_failed` at WARN and the run row is typed (not silent, as filed). The rejected alternatives are recorded at [`agent/run.py:411-416`](backend/src/meho_backplane/agent/run.py#L411-L416), [`scheduler/loop.py:306-315`](backend/src/meho_backplane/scheduler/loop.py#L306-L315), `docs/codebase/scheduler.md:358-373`.
- Create-time: `ScheduledTriggerCreate._validate_discriminated_union` ([`scheduler/schemas.py:129-168`](backend/src/meho_backplane/scheduler/schemas.py#L129-L168)) accepts input-less `one_off`/`cron` (201); the create handler runs an id-only FK existence probe — the filed claim that it "already has the system_prompt available" is refuted.
- The `inputs: {}` edge: `_coerce_inputs` ([`loop.py:296-321`](backend/src/meho_backplane/scheduler/loop.py#L296-L321)) + `prompt_is_effectively_empty` (`run.py:315-328`) — an empty dict serializes to `"{}"` and passes the fire-time guard as a non-empty prompt.
[...]

PR #2377, opened 2026-07-11
fix(scheduler): convert one_off fire_at picker to UTC on submit (#2339)

### LINKED ISSUE #2339: Task: convert the scheduler fire_at datetime-local input to UTC at submit — CEST operators schedule 2h off
## Summary

The `/ui/scheduler` Create-trigger modal renders a datetime input for `fire_at`, but the value the form submits is interpreted as **UTC without any timezone conversion** from the operator's local zone. Result: an operator in a non-UTC timezone (e.g. CEST, +02:00) who picks "2 minutes from now" via the picker gets a trigger whose `fire_at` is **2 hours further out** than they intended.

Observed live on v0.20.0.

## Reproduction

1. Authenticate to `/ui` as any operator with an agent-definition available.
2. Set your workstation's timezone to something ≠ UTC (e.g. CEST / Europe/Berlin / `+02:00`).
3. Navigate to `/ui/scheduler`. Click **Create trigger**.
4. Fill:
   - `kind`: `one_off`
   - `agent_definition_id`: any registered agent
   - `fire_at`: pick a datetime approximately 2 minutes from the local clock time on your workstation.
   - `inputs`: `{"instruction": "test"}` (see `#2244` — must be non-empty)
5. Submit.
6. Observe the resulting trigger's `fire_at` field on the detail view.

## Observed

- Operator in CEST (`+02:00`) picks a datetime that reads "2 minutes from now" on their local clock (say `11:00`).
- Engine stores `fire_at: <11:00Z>` (interpreted as UTC 11:00, i.e. 13:00 local CEST).
- Trigger doesn't fire for **~2 hours** — the offset between operator's timezone and UTC.

The user sees the local time on the input picker, but the field is submitted as a UTC string. The form should either:
- Convert the local time to UTC on submit (matching the picker's behavior), OR
- Show the input as UTC with a "(UTC)" label next to the field, so the operator knows to enter UTC time.

Currently it silently mixes the two, leading to hard-to-diagnose delays.

## Why this matters

- **Silent scheduling defect.** The operator sees "next fire" 2 hours later on the detail view but doesn't necessarily connect the timezone offset — the picker's datetime and the stored `fire_at` string look "different but plausibly equivalent." Debugging this is annoying.
- **Every non-UTC operator is affected.** Most European operators sit at CET/CEST (+01:00 / +02:00) — that's a large fraction of the operator base for a European MEHO deploy.
- **Fire-and-check-outcome-at-T workflows misfire.** Any workflow that expects the trigger to fire around a specific wall-clock time (e.g. "at end of business day", "at start of maintenance window") schedules 1-2 hours off. Combined with `#2245`'s ~28-second latency spike after a failed run, autonomous scheduling on the UI is fragile.

- **Option A — convert local → UTC on submit.** Client-side (JS in the form's submit handler): read the datetime picker value as local time, convert to UTC ISO 8601, submit. Match the operator's mental model: "I picked 11:00 local, the trigger fires at 11:00 local."
- **Option B — display the input as UTC.** Label the field "Fire at (UTC)" and set the picker's default to UTC-now. Forces the operator to think in UTC but is explicit.
- **Option C — send timezone offset alongside `fire_at`.** Add a `timezone` field to the trigger schema and let the engine convert. More flexible but higher-scope.

Option A is the cleanest — most operators think in local time, and hiding the UTC-conversion from them is the right ergonomic choice. Option B is the minimum-change alternative.

## Verification

After fix:

1. Same repro steps 1-5.
2. Expected: trigger's `fire_at` (as stored on the engine) equals the operator's local time converted to UTC. The trigger fires within the normal tick cadence of the operator's picked local time.

Quick automation: on submit, check that the trigger's `fire_at` (returned in the POST response) equals `(picker value as local time) → toISOString()`.

## Out of scope

- The scheduler's engine-side handling of `fire_at` — that layer works correctly with any UTC input; the defect is UI-side conversion.
- Cron triggers — those specify time via cron expression and are already interpreted per an explicit `timezone` field on the trigger. The defect is specifically the `one_off` `fire_at` datetime picker.
- Server-side timezone display on the trigger detail view (may or may not already show local time; not verified).

## Related

- Sibling: `#2244` (no-inputs 400 footgun — same Create form)
- Sibling: `#2245` (~28s latency spike after failed run — makes SLA-sensitive scheduling further fragile)

---

Parent goal: #221
Parent initiative: #2364

damir-topic:
## Summary
- The `/ui/scheduler` Create-trigger modal's `fire_at` `datetime-local` picker showed operators their **local** wall-clock time but posted a naive `YYYY-MM-DDTHH:MM` string the engine stored **verbatim as UTC** — a CEST (UTC+2) operator who picked "11:00" got a trigger that fired at `11:00Z`, two hours late. Every non-UTC operator was affected (#2339).
- Fix (Option A, client-side): an inline submit handler converts the picked local value to a UTC ISO-8601 instant on `htmx:configRequest`, using the browser's own **DST-aware** `Date` so CET/CEST both resolve without a hardcoded offset. A live hint next to the field mirrors the resolved UTC so the conversion is **visible**, not silent.
- Engine-side `fire_at` handling is untouched — the server already honours an offset-carrying (`…Z`) value as-is.

## Test plan
- [x] `pytest tests/test_ui_scheduler.py` — 32 passed
- [x] New `test_create_modal_carries_fire_at_utc_conversion` — asserts the modal ships the `htmx:configRequest` handler + `toISOString` conversion + `data-fire-at` hint.
- [x] New `test_create_submit_one_off_utc_fire_at_persists_instant` — a `one_off` submit with the JS-converted `2026-07-01T09:00:00.000Z` persists exactly `09:00Z` (proves the offset is honoured, not re-interpreted).

## Out of scope
- Detail-view local-time rendering of the stored UTC `fire_at` (display-only; the scheduling defect is the submit conversion). Noted as an adjacent follow-up.
- Cron triggers (already carry an explicit `timezone` field) and engine-side `fire_at` handling.
- Sibling Wave-6 tasks: #2345 (CSRF), #2346 (agents), #2244 (no-inputs footgun).

Closes #2339

### DIFF
--- a/backend/tests/test_ui_scheduler.py
+++ b/backend/tests/test_ui_scheduler.py
@@ -233,6 +233,20 @@ def _load_trigger_status(tenant_id: uuid.UUID, trigger_id: uuid.UUID) -> str | N
     return asyncio.run(_do())
 
 
+def _load_trigger_fire_at(tenant_id: uuid.UUID) -> datetime | None:
+    from sqlalchemy import select
+
+    async def _do() -> datetime | None:
+        sessionmaker = get_sessionmaker()
+        async with sessionmaker() as session:
+            stmt = select(ScheduledTrigger.fire_at).where(
+                ScheduledTrigger.tenant_id == tenant_id,
+            )
+            return (await session.execute(stmt)).scalars().first()
+
+    return asyncio.run(_do())
+
+
 def _trigger_count(tenant_id: uuid.UUID) -> int:
     from sqlalchemy import func, select
 
@@ -571,6 +585,31 @@ def test_create_modal_renders_for_tenant_admin() -> None:
     assert 'hx-post="/ui/scheduler/validate-cron"' in body  # live cron validation
 
 
+def test_create_modal_carries_fire_at_utc_conversion() -> None:
+    """The one_off fire_at field ships the client-side local->UTC submit conversion (#2339).
+
+    The datetime-local picker posts a naive local wall-clock string the engine
+    stores verbatim as UTC; the modal must carry the ``htmx:configRequest``
+    handler that rewrites ``fire_at`` to a UTC ISO instant plus the live hint
+    that makes the conversion visible. Asserts the load-bearing markers rather
+    than exercising the JS (covered by the Playwright regression in CI).
+    """
+    _seed_tenant(_TENANT_A, "tenant-a")
+    _seed_agent(tenant_id=_TENANT_A)
+    client, mock, _ = _client_with_role(
+        tenant_id=_TENANT_A, operator_sub=_OP_ADMIN, role=TenantRole.TENANT_ADMIN
+    )
+    try:
+        response = client.get("/ui/scheduler/create")
+    finally:
+        mock.stop()
+    assert response.status_code == 200, response.text
+    body = response.text
+    assert "data-fire-at" in body  # the picker + its live hint are wired
+    assert "htmx:configRequest" in body  # submit-time param rewrite hook
+    assert "toISOString" in body  # DST-aware browser-native UTC conversion
+
+
 def test_create_modal_rejects_operator_with_403() -> None:
     """An operator (non-admin) GET on the create modal is 403'd server-side."""
     _seed_tenant(_TENANT_A, "tenant-a")
@@ -672,6 +711,43 @@ def test_create_submit_persists_cron_trigger_and_redirects() -> None:
     assert response.status_code == 204, response.text
     assert response.headers["HX-Redirect"] == "/ui/scheduler"
     assert _trigger_count(_TENANT_A) == 1
+
+
+def test_create_submit_one_off_utc_fire_at_persists_instant() -> None:
+    """A one_off create with the JS-converted UTC ``fire_at`` stores that instant (#2339).
+
+    The submit handler's client-side conversion posts a UTC ISO-8601 string
+    (``...Z``) rather than the naive datetime-local value; the server honours
+    the offset as-is. A CEST operator picking 11:00 local -> 09:00Z must land
+    exactly 09:00Z, not 11:00Z (the 2h-off defect).
+    """
+    _seed_tenant(_TENANT_A, "tenant-a")
+    _seed_agent(tenant_id=_TENANT_A)
+    client, mock, csrf = _client_with_role(
+        tenant_id=_TENANT_A, operator_sub=_OP_ADMIN, role=TenantRole.TENANT_ADMIN
+    )
+    try:
+        response = client.post(
+            "/ui/scheduler/create",
+            data={
+                "kind": "one_off",
+                "agent_definition_id": str(_AGENT_ID),
+                "fire_at": "2026-07-01T09:00:00.000Z",
+                "inputs": '{"prompt": "scheduled run"}',
+                "in_flight_policy": "fail_into_audit",
+            },
+            headers=_form_headers(csrf),
+        )
+    finally:
+        mock.stop()
+    assert response.status_code == 204, response.text
+    assert _trigger_count(_TENANT_A) == 1
+    stored = _load_trigger_fire_at(_TENANT_A)
+    assert stored is not None
+    # Normalise to an aware UTC instant (the column may hydrate naive-UTC).
+    if stored.tzinfo is None:
+        stored = stored.replace(tzinfo=UTC)
+    assert stored == datetime(2026, 7, 1, 9, 0, tzinfo=UTC)
 
 
 def test_create_submit_invalid_cron_rerenders_modal_with_error() -> None:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,18 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Fixed — scheduler one_off `fire_at` picker now converts local time to UTC on submit (#2339)
+
+- The `/ui/scheduler` Create-trigger modal's `fire_at` `datetime-local`
+  picker showed the operator their own wall-clock time but posted a
+  naive string the engine stored verbatim as UTC, so a CEST (UTC+2)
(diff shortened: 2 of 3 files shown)


damir-topic:
@coderabbitai review

backend/tests/test_ui_scheduler.py
```diff
@@ -571,6 +585,31 @@ def test_create_modal_renders_for_tenant_admin() -> None:
...
 
 
+def test_create_modal_carries_fire_at_utc_conversion() -> None:
+    """The one_off fire_at field ships the client-side local->UTC submit conversion (#2339).
+
+    The datetime-local picker posts a naive local wall-clock string the engine
+    stores verbatim as UTC; the modal must carry the ``htmx:configRequest``
+    handler that rewrites ``fire_at`` to a UTC ISO instant plus the live hint
+    that makes the conversion visible. Asserts the load-bearing markers rather
+    than exercising the JS (covered by the Playwright regression in CI).
```
damir-topic, line 595: **[⚠️ M1]** — Docstring claims Playwright CI coverage that does not exist

**What:** This docstring says the JS conversion is "covered by the Playwright regression in CI," but this repo has no Playwright harness — `test_ui_templates.py:282` states it explicitly: "The console has no browser test harness (Playwright is not wired here)."
**Why it matters:** The claim gives false confidence that the client-side local→UTC conversion is automatically regression-guarded. It isn't — the JS is only verified by inspection. A future edit to the inline script would ship unguarded while readers believe CI covers it.
**Required change:** Reword the parenthetical to reflect reality, e.g. "(the JS conversion has no automated browser coverage in this repo — verified by inspection; see `test_ui_templates.py` note)". Do not cite a Playwright regression that isn't wired.
**Source:** repo `backend/tests/test_ui_templates.py:282`

damir-topic:




**Anchor task:** #2339 — scheduler one_off `fire_at` picker: convert operator local time to UTC on submit (CEST operators scheduled 2h off)
**Surface area:** frontend (HTMX/Jinja operator console) + backend test
**Verdict:** APPROVE
**Head SHA reviewed:** `aa8abbc9`
**AI reviewer (CodeRabbit):** not received within 15 min — **rate-limited** ("Review limit reached" posted 17:19:58; matches the back-to-back auto-run quota pattern)

### Acceptance criteria
- [x] one_off `fire_at` local→UTC conversion on submit — `htmx:configRequest` handler rewrites `params.fire_at` to a DST-aware `Date().toISOString()` instant (verified by inspection; `_create_modal.html:281-316`)
- [x] No hardcoded offset — conversion relies on the browser's own zone via `new Date("YYYY-MM-DDTHH:MM")` (ES2015+ interprets an offset-less date-*time* string as **local**), so summer (CEST +2) and winter (CET +1) both resolve from the actual date's DST rules. `Intl` is used only to *name* the zone in the hint, never to compute the offset.
- [x] Double-conversion guard — `NAIVE_LOCAL = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}(:\d{2})?$/` matches only an offset-less value; a `…Z` / `+hh:mm` value returns `null` and is left untouched, so re-submits and pre-filled offset values aren't shifted twice.
- [x] Server contract consistent — `_parse_fire_at` (`create.py:197`) `datetime.fromisoformat(raw)`: naive→UTC, offset honoured as-is. Client now always sends `…Z`, so no naive ambiguity on the JS path. (Python 3.12 `fromisoformat` parses the `…Z` and `.000` millis — verified by running it.)
- [x] Conversion visible, not silent — `aria-live="polite"` hint mirrors the resolved UTC + local zone name.

### Predicted vs. actual diff
- ✅ `_create_modal.html` — field relabel + `data-fire-at`/`aria-describedby` wiring + inline submit-conversion script (expected, correct)
- ✅ `test_ui_scheduler.py` — marker test (handler ships) + server-persistence test (offset instant stored exactly)
- ✅ `CHANGELOG.md` — unique `### Fixed —` header, no `[Unreleased]` conflict
- 🕳️ None missing. **No route/schema change** (`create.py` untouched) → **no OpenAPI regen needed** (confirmed).

### Integration verification
- [x] `ruff check` (tests + create.py) — clean
- [x] `pytest tests/test_ui_scheduler.py` — **32 passed** (incl. both new tests) via `uv run` in the PR worktree
- [⚠️ not run] browser exercise of the JS conversion — no browser harness exists in this repo (Playwright is explicitly *not wired*, per `test_ui_templates.py:282`); JS correctness asserted by inspection

- ⚠️ M1 — `test_ui_scheduler.py:595` docstring claims the JS is "covered by the Playwright regression in CI," but this repo has **no** Playwright harness (`test_ui_templates.py:282` says so explicitly). Reword to state the JS is verified by inspection / has no automated browser coverage. Non-blocking (does not affect behaviour), but the false coverage claim should be corrected.

### Test teeth (focus-area verdict)
`test_create_submit_one_off_utc_fire_at_persists_instant` posts the *client-converted* `2026-07-01T09:00:00.000Z` and asserts the stored `fire_at == 09:00Z` — this genuinely pins the **server** half of the contract (offset honoured, not re-interpreted as 11:00Z). `test_create_modal_carries_fire_at_utc_conversion` pins the **handler-ships** half (`data-fire-at` + `htmx:configRequest` + `toISOString`). The JS conversion itself (local→UTC arithmetic) is **only assertable by inspection** here — the Python tests cannot exercise it. That gap is acceptable given no browser harness exists; it is flagged so it's on record (see M1 for the docstring overclaim).

- Rate-limited this iteration — 0 findings to cross-check. Recorded `bot_review.received: false`.

### Docs
- [x] n/a — no `docs/codebase/` doc invalidated; `_parse_fire_at` server docstring stays accurate; `CHANGELOG` updated.

- The offset-carrying-value guard + the visible `aria-live` UTC hint make an easy-to-get-wrong tz conversion both correct and auditable to the operator.
- Correctly scoped: engine-side `fire_at` left untouched (it already honoured offsets), no schema/route churn, no snapshot needed.

Issue #574, opened 2026-05-17
v0.2 cannot start out of the box: default RETRIEVAL_EMBEDDING_MODEL + bundled fastembed → onnxruntime NoSuchFile model_optimized.onnx (CrashLoop)


damir-topic:
Continuing the v0.2 dogfood deploy: with Keycloak, Vault, pgvector, the trust bundle, and the fastembed-cache volume all sorted, the backplane now gets all the way to loading the embedding model and **the shipped default doesn't load with the bundled fastembed**:

```
onnxruntime.capi.onnxruntime_pybind11_state.NoSuchFile: [ONNXRuntimeError] : 3 : NO_SUCHFILE :
Load model from /var/cache/fastembed/models--qdrant--bge-small-en-v1.5-onnx-q/snapshots/<sha>/model_optimized.onnx failed.
File doesn't exist
ERROR: Application startup failed. Exiting.
```

i.e. with the chart default `RETRIEVAL_EMBEDDING_MODEL=BAAI/bge-small-en-v1.5`, the fastembed version pinned in `ghcr.io/evoila/meho:v0.2.0` resolves that to the `qdrant/bge-small-en-v1.5-onnx-q` HF repo and then looks for `model_optimized.onnx`, which that repo's current revision doesn't contain. The snapshot directory downloads fine (so it's not egress, not the trust bundle, not the cache volume — all of which we fixed on the way here); fastembed's expected ONNX filename and the model repo's actual contents simply disagree. This is deterministic: every fresh pod hits it identically and CrashLoopBackOffs, so the backplane never reaches Ready and the chart's Deployment fails its progress deadline.

This looks like a fastembed-version ⇔ model-repo pinning drift inside the image rather than anything an operator can fix from `values` (no `RETRIEVAL_EMBEDDING_MODEL` an operator sets changes what's baked into the image's fastembed). Concretely, it means **v0.2 cannot start out of the box** with its own default retrieval model on a clean deploy.

Suggestions, in rough order:

1. Pin fastembed **and** the embedding model to a known-compatible pair in the image build, and add a CI smoke that actually instantiates `TextEmbedding(RETRIEVAL_EMBEDDING_MODEL)` so this can't regress silently (this is exactly the kind of thing green CI on `main` won't catch if the model is fetched at runtime).
2. Bake the default model into the image (or an optional model-bundled image) so first-boot is offline and version-locked — this also closes the air-gap gap from evoila/meho#572.
[...]

PR #2378, opened 2026-07-11
fix(ui): unify /ui CSRF token pattern to stop 3rd-write 403 drift

### LINKED ISSUE #2345: Task: unify the /ui CSRF token pattern — rotate-on-render vs static-reuse drift 403s the 3rd write in a session

damir-topic:
## Summary

- Fixes the operator console's "works twice, 403s the 3rd write" CSRF drift (#2345): `/ui/*` writes intermittently returned `403 csrf_token_invalid` because the codebase mixed **two** double-submit patterns — some renders rotated the `meho_csrf` cookie on every fragment render, others reused the live cookie — so an open modal's baked-in `X-CSRF-Token` echo drifted out of sync with a rotated cookie.
- **Unifies on one session-stable token.** `mint_csrf_token` now derives its nonce deterministically from the session id (domain-separated HMAC) instead of fresh randomness, so re-minting on any render returns the identical value; the token only changes when the session rotates at an auth boundary. Rotation-on-render becomes a value-stable no-op everywhere — no route files touched.
- **Adds a global `htmx:configRequest` hook** (`static/src/app/csrf-token.js`, loaded from `_head_assets.html` after htmx) that re-reads the live `meho_csrf` cookie at request time and overrides the `X-CSRF-Token` header, so the double-submit pair stays in lockstep regardless of render order and any future rotation is picked up without a page reload.
- `verify_csrf_token` is unchanged (recomputes the MAC from the presented nonce), so tokens minted before the deploy still validate across the rollover. Security is unchanged: the HMAC binding to the session id still defeats the sub-domain cookie-injection vector, and the token was never single-use.

Implements the maintainer's adopted fix from the grounding comment (session-stable token + global configRequest hook). The visible-403 error banner is the carrier of the sibling error-mold task (#2346); the existing `session-expiry.js` net already surfaces the CSRF 403 today.

## Test plan

- [x] `ruff check` — clean (csrf.py, both test files)
- [x] `mypy src/meho_backplane/ui/csrf.py --ignore-missing-imports` — clean
- [x] New backend regressions in `test_ui_chassis_smoke.py`: `test_csrf_token_is_session_stable` (two mints equal, both verify; distinct sessions differ) and `test_csrf_repeated_writes_survive_token_rerender` (4 writes with a re-mint-and-reset cookie between each still pass — the drift scenario)
- [x] New contract tests in `test_ui_templates.py`: the hook attaches on `htmx:configRequest`, reads `document.cookie`/`meho_csrf`, sets `X-CSRF-Token`; `_head_assets.html` loads it deferred after htmx
- [x] `pytest tests/test_ui_chassis_smoke.py tests/test_ui_templates.py` — 54 passed, 1 skipped
- [x] Representative CSRF-dependent UI suite (memory list poll-no-rotate, topology annotate, agents, kb upload, account, runbook runs, keycloak write) — 132 passed
- Note: no Playwright/JS test runner exists in this repo (per `session-expiry.js`), so the maintainer's "Playwright repeated-writes" ask is realized as the Python repeated-writes middleware regression above.

## Out of scope

- The `csrf_token_invalid` 403 error banner / error mold — carried by sibling #2346.
- Consolidating the ~28 duplicated `_set_csrf_cookie` helpers and 6 `_resolve_*_csrf` helpers into one shared function — a behavior-preserving refactor now made redundant by the session-stable token (all become value-stable no-ops); left untouched to avoid serial-merge collisions with siblings #2346/#2339. Adjacent finding.

Closes #2345


damir-topic:
@coderabbitai review

damir-topic:




**Anchor task:** #2345 — unify the `/ui` CSRF double-submit token pattern (stop the 3rd-write 403 drift)
**Surface area:** backend (+ first-party app-shell JS)
**Best-practices packs audited:** python, frontend (vanilla JS/CSP)
**Verdict:** APPROVE
**Head SHA reviewed:** `2386bf68e6f120ff51b0b9f9ce6376615a6f090d`
**AI reviewer (CodeRabbit):** not received — rate-limited ("review limit reached, next in 16 min"); the known back-to-back-autorun quota exhaustion. Proceeded on the independent review.

This is a security-sensitive change; the deterministic-token argument was reviewed adversarially rather than rubber-stamped. It **holds**.

### Security analysis — deterministic session-stable token

- **HMAC key is a genuine server-only secret.** `_csrf_secret()` returns `Settings.ui_session_encryption_key` (the Fernet session key, Vault/env-managed). An unset key raises `EncryptionKeyMissingError` (fail-closed) rather than minting off a zero-byte key. The secret is **not** derivable from any cookie-visible material. ✅
- **Domain separation is real.** The nonce is `HMAC(secret, _NONCE_DOMAIN || session_id)` with `_NONCE_DOMAIN = b"meho-csrf-nonce-v1\x00"`, while the token MAC is `HMAC(secret, session_id || nonce)`. The domain prefix appears only in the nonce derivation, so the nonce can never collide with the token MAC under the same key. ✅
- **Knowing a session id alone does not yield the token.** Both the nonce and the MAC require the server secret; `session_id` additionally lives only in the `HttpOnly meho_session` cookie. The token is not computable from anything JS/an attacker can see. ✅
- **Double-submit / cookie-injection still defeated.** `_validate_csrf` keeps both the `value_mismatch` (cookie == header) and `signature_invalid` (`verify_csrf_token` recomputes the MAC against the `session_id` read from the validated `meho_session` cookie) checks. A sub-domain attacker who sets `meho_csrf` + a matching header still cannot produce a MAC that verifies against the victim's session id. The session binding is intact. ✅
- **Rollover safe.** `verify_csrf_token` is byte-for-byte unchanged and stateless: it recomputes the MAC from the *presented* nonce, so pre-deploy random-nonce tokens still validate against the same secret + session id. No mid-session lockout. ✅
- **Never single-use — confirmed against old code.** `origin/main` `mint_csrf_token` used `secrets.token_bytes` but was explicitly stateless ("nothing persists server-side; any in-flight token … validates until logout"). Determinism therefore removes no security property — only the drift. This is the standard OWASP per-session (synchronizer-style) token. ✅
- **Client hook is safe.** `csrf-token.js` reads the same-origin `document.cookie` `meho_csrf` and mutates only `event.detail.headers["X-CSRF-Token"]` on outgoing htmx requests (same-origin app requests). No cross-origin exfiltration path; when no cookie is present it leaves the template's `hx-headers` value untouched (no regression on unauthenticated surfaces). Loaded `defer` after `htmx.min.js`, before the first htmx request resolves. ✅

Net: determinism costs no security property and closes the drift at its source.

### Acceptance criteria
- [x] `/ui/*` writes no longer 403 on the 3rd repeated write — `mint_csrf_token` is now a pure function of the session; `test_csrf_repeated_writes_survive_token_rerender` reproduces the drift and passes.
- [x] Single-file mint change, no route files edited — no serial-merge collision with #2346/#2339.
- [x] Security unchanged — verified above (verify unchanged, session-binding intact, key server-only).
- [x] Docs updated — `docs/codebase/ui.md` describes the session-stable token + the configRequest hook accurately.

### Project-board alignment
|---|---|---|---|---|---|
| #2345 | In Review | Closes | ✅ | ✅ full | close on merge |

### Predicted vs. actual diff
- ✅ `ui/csrf.py` (deterministic nonce + docstrings; verify untouched)
- ✅ `static/src/app/csrf-token.js` + `_head_assets.html` (configRequest hook, deferred after htmx)
- ✅ `test_ui_chassis_smoke.py` (session-stable + repeated-writes regressions), `test_ui_templates.py` (hook content/order contract)
- ✅ `docs/codebase/ui.md`, `CHANGELOG.md` (distinct `### Fixed —` header, no `[Unreleased]` merge conflict)
- 🕳️ None missing — the ~28 `_set_csrf_cookie` helpers are correctly left untouched (now value-stable no-ops); consolidating them is out of scope by design to avoid sibling collisions.

### Integration verification
- [x] `pytest tests/test_ui_chassis_smoke.py tests/test_ui_templates.py` (in worktree) — 54 passed, 1 skipped
- [x] CI unit lane — 2449 passed, 1 failed = `test_real_descriptor_embedding_path` only, the known fastembed-cache flake #574 (task-sanctioned ignore). No CSRF/ui failures.
- [x] `CLI API snapshot freshness` CI check — pass (no route/schema drift, as claimed)
- [x] Semgrep SAST / CodeQL / TruffleHog — pass
- [ ] ⚠️ not run: browser exercise of the htmx hook — no JS test runner exists in-repo (same constraint as `session-expiry.js`); realized as Python content/order contract tests, which is the right call here.

None. No blockers, majors, or minors requiring a diff.

- The `ui_session_encryption_key` keys both Fernet session encryption and the CSRF HMAC. This is pre-existing (old `_csrf_secret()` did the same) and the code documents the distinct-KDF/distinct-domain rationale. If a future hardening pass wants defence-in-depth, deriving a CSRF sub-key via HKDF with a fixed `info` label would fully separate the two consumers — not required for this PR.
- The deferred consolidation of the ~28 `_set_csrf_cookie` / 6 `_resolve_*_csrf` helpers (now value-stable no-ops) is a clean follow-up refactor once #2346/#2339 land.

- Rate-limited this iteration (no findings produced). Adopted: 0 / Disputed: 0 / Out of scope: 0.

### Docs
- [x] `docs/codebase/ui.md` — updated, accurate (session-stable token + configRequest hook).

- The security reasoning is documented end-to-end in the module docstring and the mint docstring, and it matches the code exactly — the "never single-use / determinism costs no property" claim is verifiable against the old source.
- Fail-closed on a missing key (raises rather than minting off an empty HMAC key) is the right posture and is regression-tested.
- CHANGELOG uses a distinct `### Fixed —` header, dodging the `[Unreleased]` serial-merge race.

PR #2381, opened 2026-07-11
fix(ui): disinherit leaking hx-disabled-elt find-selectors across /ui modals (#2340)

### LINKED ISSUE #2340: Task: sweep hx-disabled-elt find-selectors across /ui templates — disable-submit affordance broken in 5+ modals
## Summary

Interacting with the `/ui/scheduler` Create-trigger modal (typing in the cron field, switching kind, etc.) emits an htmx warning to the browser console:

```
htmx.min.js:1  The selector "find button[type=submit]" on hx-disabled-elt returned no matches!
  H  @  htmx.min.js:1
  we @  htmx.min.js:1
  nn @  htmx.min.js:1
  he @  htmx.min.js:1
  (anonymous) @  htmx.min.js:1
  (anonymous) @  htmx.min.js:1
  setTimeout
  s  @  htmx.min.js:1
```

The selector is a `find button[type=submit]` (htmx's own `find` verb inside an `hx-disabled-elt` attribute) — but no element matching that selector exists in the form's DOM subtree at the moment htmx evaluates it. htmx logs a warning; form still submits, but the form is missing the "disable submit while pending" affordance the attribute was supposed to provide.

Observed live on v0.20.0. Same shape as the previously-fixed `#2117` D1 (Preview button `Element.closest` with `>>` selector) — an htmx attribute referencing a selector that doesn't match the DOM.

## Reproduction

1. Authenticate to `/ui` as any operator; open DevTools → Console.
2. Navigate to `/ui/scheduler`. Click **Create trigger**.
3. Interact with the modal — type in the cron field (the real-time validator fires), switch kind, etc.
4. Observe console — the warning appears in response to htmx settle events.

## What's happening

An `hx-disabled-elt="find button[type=submit]"` attribute somewhere in the modal is trying to disable the submit button during pending requests (a common htmx pattern for UX). htmx's `find` verb scopes the selector to descendants of the element carrying `hx-disabled-elt` — but the submit button isn't a descendant of that element, so `find` returns empty and htmx warns.

The warning is technically non-fatal (form submit still works), but:

- The intended UX (submit button disabled during pending) doesn't happen.
- The console warning obscures real defects.
- The pattern is worth fixing project-wide since `#2117` D1's fix already established that "htmx selectors in attributes should match the DOM" is a monitored discipline.

## Fix shapes

- **Option A — use a scoped selector**: replace `find button[type=submit]` with a selector that matches the actual submit button's location relative to the element. For example, if the submit button is a sibling, use `closest form button[type=submit]` or an explicit `#submit-button-id`.
- **Option B — move the attribute**: put `hx-disabled-elt` on an element whose descendants DO include the submit button (e.g. the `<form>` itself).
- **Option C — drop the attribute** if the intended UX isn't important (probably not).

## Verification

After fix:

1. Same repro steps 1-4.
2. Expected: no `The selector "find button[type=submit]" on hx-disabled-elt returned no matches!` warning in the console. Submit button correctly disabled during pending requests.

Add a lightweight check that the console has no htmx warnings during a normal Create-trigger flow — could catch this and similar defects earlier.

## Related

- Same shape as the previously-fixed `#2117` D1 (Preview button `Element.closest('.flex-wrap >> .runbook-step-preview')`) — an htmx attribute pointing at a selector that doesn't match the DOM. Establishing a "htmx selectors must match" discipline would catch this class generically.
- Sibling to `evoila/meho#2242` (modal-dialogs.js `showModal` on orphaned element after successful abort) — same code path pattern (htmx swap + late DOM ref).
- Non-blocking (form submits fine), just noise. Low priority.

## Out of scope

- The Create-trigger form's actual submit logic (works — see sibling `#2244`, `#2245` for the actual scheduler engine defects).
- Any other htmx-attribute misconfigurations elsewhere in `/ui/scheduler` (this filing only covers the one observed on the Create modal).

---

Parent goal: #221
Parent initiative: #2364

damir-topic:
## Summary

- Five `/ui` operator-console forms bind `hx-disabled-elt="find button[type=submit]"` to disable their submit button while a request is in flight. Because `hx-disabled-elt` is an **inherited** htmx attribute, each form's *descendant* htmx requests (debounced cron-validate / token-preview `POST`s and in-form Cancel `hx-get`) inherited the value and resolved `find button[type=submit]` against their own submit-button-less subtree — logging `The selector "find button[type=submit]" on hx-disabled-elt returned no matches!` on every such request while the disable-submit affordance silently never fired.
- Fix: add `hx-disinherit="hx-disabled-elt"` to each of the 5 leaking forms, scoping the value to the form's own submit (the exact pattern established for the runbook editor in #2174). Files: `scheduler/_create_modal.html`, `memory/_create_modal.html`, `memory/_body_edit.html`, `conventions/_create_modal.html`, `conventions/_edit_modal.html`.
- Added a **generic** `test_ui_templates.py` guard that scans every `/ui` template: any form with a `find` `hx-disabled-elt` plus a descendant htmx request must carry `hx-disinherit` — establishing the "htmx selectors must match the DOM" discipline generically (the console has no browser harness to catch this class).

## Grounding correction

The issue's grounding comment named `scheduler/_create_modal.html` (correct — leaks via the cron-validate `POST`) plus `memory/_promote_modal.html` and `runbooks/_run_step.html`. A per-form check showed the latter two carry **no descendant htmx request**, so their form-level `find` resolves cleanly against their own submit button and they warn nowhere — they are correctly **left untouched**. The actual leaking set is exactly 5 modals (matching the title's "5+"): the two `conventions` and two `memory` forms plus `scheduler`, each of which does have a descendant htmx trigger.

## Test plan

- [x] `ruff check` + `ruff format --check` on changed test files — clean
- [x] `pytest tests/test_ui_templates.py tests/test_ui_scheduler.py` — 71 passed (incl. new generic guard + scheduler focused assertion)
- [x] `pytest tests/test_ui_conventions_write.py tests/test_ui_memory_create_promote.py tests/test_ui_runbooks_editor.py` — 76 passed (no regression; the existing #2174 editor-disinherit test still green)
- [x] Confirmed the new guard fails if `hx-disinherit` is removed from any of the 5 forms and passes on the full current template tree

## Out of scope

- `find button[data-action=...]` disabled-elt selectors on keycloak/vault/operations single-button confirm forms — verified to have no descendant htmx requests, so no leak.
- Playwright console-warning assertions (no browser harness wired in this repo; the server-side generic guard is the pytest-level equivalent).

Closes #2340

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,24 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Fixed — disinherited leaking `hx-disabled-elt` find-selectors across `/ui` modals (#2340)
+
+- Five operator-console modals (`scheduler` create-trigger, `memory`
+  create + body-edit, `conventions` create + edit) bound
+  `hx-disabled-elt="find button[type=submit]"` on the `<form>` to disable
+  the submit button while a request is in flight. Because `hx-disabled-elt`
+  is an *inherited* htmx attribute, each form's descendant htmx requests —
+  the debounced cron-validate / token-preview `POST`s and the in-form
+  Cancel `hx-get` — inherited the value and resolved `find button[type=submit]`
+  against their own (submit-button-less) subtree, logging
+  `The selector "find button[type=submit]" on hx-disabled-elt returned no
+  matches!` on every such request while the intended disable-submit
+  affordance silently never fired. Each form now carries
+  `hx-disinherit="hx-disabled-elt"` (the pattern established for the runbook
+  editor in #2174), scoping the value to the form's own submit. A generic
+  `test_ui_templates.py` guard enforces the discipline across every `/ui`
+  template so a new modal cannot reintroduce the leak.
+
 ### Fixed — unified the `/ui` CSRF double-submit token pattern (#2345)
 
 - The operator console's `/ui/*` write surfaces no longer `403
--- a/backend/src/meho_backplane/ui/templates/conventions/_create_modal.html
+++ b/backend/src/meho_backplane/ui/templates/conventions/_create_modal.html
@@ -49,6 +49,7 @@
       hx-target="#conventions-write-result"
       hx-swap="innerHTML"
       hx-disabled-elt="find button[type=submit]"
+      hx-disinherit="hx-disabled-elt"
       class="space-y-3 py-2"
       aria-labelledby="conventions-create-modal-title"
     >
--- a/backend/src/meho_backplane/ui/templates/conventions/_edit_modal.html
+++ b/backend/src/meho_backplane/ui/templates/conventions/_edit_modal.html
@@ -30,6 +30,7 @@
       hx-target="#conventions-write-result"
       hx-swap="innerHTML"
       hx-disabled-elt="find button[type=submit]"
+      hx-disinherit="hx-disabled-elt"
       class="space-y-3 py-2"
       aria-labelledby="conventions-edit-modal-title"
     >
--- a/backend/src/meho_backplane/ui/templates/memory/_body_edit.html
+++ b/backend/src/meho_backplane/ui/templates/memory/_body_edit.html
@@ -26,6 +26,7 @@
   hx-target="this"
   hx-swap="outerHTML"
   hx-disabled-elt="find button[type=submit]"
+  hx-disinherit="hx-disabled-elt"
   aria-label="Edit memory body"
 >
   <label class="form-control">
--- a/backend/src/meho_backplane/ui/templates/memory/_create_modal.html
+++ b/backend/src/meho_backplane/ui/templates/memory/_create_modal.html
@@ -70,6 +70,7 @@
         hx-target="this"
         hx-swap="none"
         hx-disabled-elt="find button[type=submit]"
+        hx-disinherit="hx-disabled-elt"
         data-target-scoped-values="{{ target_scoped_values | join(',') }}"
         class="space-y-3 py-2"
         aria-labelledby="memory-create-modal-title"
--- a/backend/src/meho_backplane/ui/templates/scheduler/_create_modal.html
+++ b/backend/src/meho_backplane/ui/templates/scheduler/_create_modal.html
@@ -58,6 +58,7 @@
       hx-target="#scheduler-modal-container"
       hx-swap="innerHTML"
       hx-disabled-elt="find button[type=submit]"
+      hx-disinherit="hx-disabled-elt"
       hx-headers='{"X-CSRF-Token": "{{ csrf_token }}"}'
       class="space-y-3 py-2"
       aria-labelledby="scheduler-create-modal-title"
--- a/backend/tests/test_ui_scheduler.py
+++ b/backend/tests/test_ui_scheduler.py
@@ -583,6 +583,10 @@ def test_create_modal_renders_for_tenant_admin() -> None:
     assert 'hx-post="/ui/scheduler/create"' in body
     assert str(_AGENT_ID) in body  # agent dropdown option
     assert 'hx-post="/ui/scheduler/validate-cron"' in body  # live cron validation
+    # #2340: the form's `find button[type=submit]` disabled-elt must be
+    # disinherited so the descendant validate-cron POST does not inherit it
+    # and log `... returned no matches!` on every debounced keystroke.
+    assert 'hx-disinherit="hx-disabled-elt"' in body
 
 
 def test_create_modal_carries_fire_at_utc_conversion() -> None:
--- a/backend/tests/test_ui_templates.py
+++ b/backend/tests/test_ui_templates.py
@@ -594,3 +594,81 @@ def test_head_assets_loads_csrf_token_hook_after_htmx() -> None:
     htmx_pos = source.index("/ui/static/src/vendor/htmx.min.js")
     hook_pos = source.index("/ui/static/src/app/csrf-token.js")
     assert htmx_pos < hook_pos, "the CSRF hook must load after htmx.min.js"
+
+
+# ---------------------------------------------------------------------------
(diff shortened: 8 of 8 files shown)


damir-topic:
@coderabbitai review

damir-topic:




**Anchor task:** #2340 — sweep leaking `hx-disabled-elt` find-selectors across `/ui` templates
**Surface area:** frontend (server-rendered HTMX/Jinja console)
**Verdict:** APPROVE
**Head SHA reviewed:** `7fb874bc`
**AI reviewer (CodeRabbit):** not received — rate-limited ("Review limit reached", next in ~56 min). Known back-to-back auto-run cooldown.

### Root-cause verification (the crux)

The re-diagnosis is **correct and verified against the htmx 2.0.9 source**, not just docs:

- `hx-disabled-elt` **is** an inherited htmx attribute (confirmed: htmx.org docs "hx-disabled-elt is inherited and can be placed on a parent element").
- The inheritance walk `o(e,t,n)` in the bundled `htmx.min.js` (v2.0.9) guards the disinherit branch with `if (e !== t)`: the `hx-disinherit` "unset" only applies when the examined ancestor `t` is **not** the requesting element `e`. Therefore:
  - **Form's own submit** — `e === t` → disinherit skipped → the form keeps `hx-disabled-elt="find button[type=submit]"` → submit is disabled while pending. **No regression.**
  - **Descendant request** (validate-cron / preview POST, in-form Cancel `hx-get`) — `e ≠ t` → returns `"unset"` → descendant no longer inherits the `find` selector → no `... returned no matches!` warning. **Fix is correct.**
- Direct runtime evidence in the issue (the reporter observed the warning firing on the descendant validate-cron POST while typing) independently confirms `find` resolves relative to the **issuing** element under inheritance — exactly the failure the fix removes.
- Precedent is real: `runbooks/editor.html` (#2174) carries the identical `hx-disinherit="hx-disabled-elt"` pattern with the same explanatory comment.

### Grounding-comment override — implementer is right

The G0.31 grounding comment named `scheduler` + `memory/_promote_modal` + `runbooks/_run_step`. The implementer overrode this and I independently confirm the override:

- **`memory/_promote_modal.html`** — has `hx-disabled-elt="find ..."` but **no descendant htmx request** (only a vanilla-JS select toggle + an `onclick` Cancel). Its own submit is the only consumer → resolves cleanly → **correctly left untouched.**
- **`runbooks/_run_step.html`** — all four forms carry the `find` selector but **none contains a descendant htmx trigger** (selects/textarea/inputs + plain submit only) → **correctly left untouched.**
- The real leaking set is exactly the 5 fixed forms, each with a genuine descendant htmx trigger: `scheduler/_create_modal` (validate-cron POST), `memory/_create_modal` (preview POST), `memory/_body_edit` (Cancel `hx-get`), `conventions/_create_modal` (2× preview POST), `conventions/_edit_modal` (preview POST). Verified by reading each form body.

This is the "symptoms right, root-cause list partly wrong" pattern — the re-diagnosis is the correct one.

### Acceptance criteria
- [x] Sweep every `/ui` template for leaking `find` disabled-elt selectors — the generic guard `rglob`-scans all templates; its passing means the 5 fixed forms are the complete leaking set.
- [x] Fix each genuinely-leaking modal — 5 forms disinherited; the 2 non-leaking ones correctly excluded.
- [x] Establish the "htmx selectors must match the DOM" discipline generically — `test_forms_with_find_disabled_elt_disinherit_it` enforces it repo-wide (server-side equivalent of the absent browser-console harness, as the issue's follow-up suggested).

### Predicted vs. actual diff
- ✅ 5 template one-line `hx-disinherit` additions — expected and correct
- ✅ `test_ui_templates.py` generic guard + `test_ui_scheduler.py` focused assertion — expected and correct
- ✅ `CHANGELOG.md` `[Unreleased]` entry under a distinct `### Fixed — disinherited leaking ...` header (no merge-race collision with the sibling CSRF entry)
- 🕳️ none missing — the guard's green proves no other leaking form exists
- 🚫 none undisclosed

### Integration verification
- [x] `ruff check` (changed test files) — clean
- [x] `ruff format --check` — 2 files already formatted
- [x] `pytest test_ui_templates.py test_ui_scheduler.py test_ui_conventions_write.py test_ui_memory_create_promote.py test_ui_runbooks_editor.py` — **147 passed** (the `Event loop is closed` teardown warnings are the known pre-existing flaky race, not failures). Includes the new generic guard and the #2174 editor-disinherit regression test.
- [x] Guard teeth reasoned: removing `hx-disinherit` from any fixed form makes its open-tag match (find-selector + descendant trigger + no disinherit) → offender → test fails. Not over-broad: forms without a descendant htmx trigger are skipped (promote_modal / run_step stay green).
- ⚠️ not run: browser-level console-warning assertion — no Playwright/browser harness wired in this repo (correctly out-of-scope; server-side guard is the equivalent).
- n/a: OpenAPI snapshot / Alembic migration — template + test only, no route/schema/DB change.

- Not received (rate-limited, next in ~56 min). Independent review stands on its own; no CodeRabbit findings to adopt/dispute this iteration.

### Docs
- n/a — no `docs/codebase/` doc governs this template-attribute discipline; the in-template comments + the generic guard test are the durable record. `CHANGELOG` updated.

- The guard's `"hx-disinherit" not in open_tag` is a loose substring check — a form carrying an unrelated `hx-disinherit="hx-target"` would pass without actually disinheriting `hx-disabled-elt`. In practice no `/ui` form does this, and the docstring already accepts `hx-disinherit="*"`, so it's a pragmatic guard, not a defect. A future tightening could assert the disinherited value list actually contains `hx-disabled-elt` or `*`.

### What worked well
- Re-diagnosed the root cause (inheritance + issuing-element `find` resolution) rather than cargo-culting the grounding comment's file list, and proved the excluded pair genuinely don't leak.
- The generic `rglob` guard turns a one-off fix into an enforced class-level discipline, which is the correct answer for a console with no browser harness.

Issue #2394, opened 2026-07-12
Task: fail at helm-template time when the MCP resource URI is unresolvable — replace the audience_not_configured runtime crash-loop


damir-topic:
Parent goal: #221
Parent initiative: #2390

## Summary

If neither `MCP_RESOURCE_URI` nor `BACKPLANE_URL` is set, the backplane **fails closed at startup** with a hard crash — not a clear template-time validation error. The fail-closed behaviour itself is defensible (an unset MCP resource URI means every `/mcp` request would fail closed, so refusing to boot is arguably correct), but the *ergonomics* are poor: an **ingress-less** install (`ingress.enabled: false` or empty host) with both `config.backplaneUrl`/`config.mcpResourceUri` blank crash-loops with a runtime stack trace — **grounding note 2026-07-12: with default values the chart derives `BACKPLANE_URL` from `ingress.host` and `MCP_RESOURCE_URI` as `<base>/mcp` (`_helpers.tpl:77-102`, `configmap.yaml:147-154`), so a plain omission does NOT crash; only the ingress-less bring-up below does** — instead of failing at `helm template`/lint time or logging a single clear "you must set X" line.

## Repro

Install without `config.backplaneUrl` / `MCP_RESOURCE_URI`. The pod crash-loops; logs show:

```
File ".../meho_backplane/main.py", line 404, in _run_lifespan_startup
    _assert_mcp_resource_uri_configured()
File ".../meho_backplane/main.py", line 235, in _assert_mcp_resource_uri_configured
    raise RuntimeError(AUDIENCE_NOT_CONFIGURED_REMEDIATION)
RuntimeError: audience_not_configured: the MCP resource URI is unset, so every /mcp request
fails closed. Set MCP_RESOURCE_URI (e.g. https://<host>/mcp, no trailing slash) or BACKPLANE_URL
(the URI is then derived as ${BACKPLANE_URL}/mcp), and add a Keycloak oidc-audience-mapper
carrying that exact value on the client that issues the caller's token.
ERROR:    Application startup failed. Exiting.
```

The remediation *text* is genuinely good (kudos) — the issue is *where* it surfaces.

Note this also bites an intentionally **ingress-less / MCP-less bring-up** (e.g. bootstrapping the backplane to validate the credential backend before wiring ingress + DNS): there's no way to say "I know, bring it up without an MCP resource URI for now."

## Suggested fixes (any one)

[...]

Issue #2392, opened 2026-07-12
Task: document + gate the pgvector superuser prerequisite — cold migration 0003 CREATE EXTENSION fails on a least-privilege app role


damir-topic:
Parent goal: #221
Parent initiative: #2390

## Summary

Migration `0003` runs `CREATE EXTENSION vector` as the **application role** (the DSN's user), which on a stock cluster is a non-superuser. `CREATE EXTENSION` requires superuser (or a role explicitly granted `CREATE` on the DB plus the extension trusted/whitelisted), so a cold migration against a normally-provisioned Postgres fails.

## Repro

1. Provision Postgres the ordinary way — an app database owned by a **non-superuser** app role (e.g. CloudNativePG `initdb` with `owner: meho`), pgvector present in the image but the extension not yet created.
2. Point MEHO's DSN at that app role and run the migration Job.

Observed: the migration aborts on `CREATE EXTENSION vector` with a permission error (`permission denied to create extension "vector"` + `HINT: Must be superuser to create this extension.` — the `must be owner` variant is unreachable here: with `IF NOT EXISTS` a pre-existing extension is a NOTICE+skip, no ownership check), so the release never completes.

## Why it bites cold installs specifically

Managed/warm setups often pre-create the extension or run migrations as a superuser, hiding this. A first-time adopter following the chart with a least-privilege app role hits it immediately.

## Suggested fixes (any one)

- Document a hard prerequisite: the extension must be pre-created by a superuser. **(Grounding note 2026-07-12: migration 0003 already runs `CREATE EXTENSION IF NOT EXISTS vector` — `backend/alembic/versions/0003_create_documents_with_pgvector.py:167` — which is exactly why the pre-create workaround below succeeds. The residual work is documentation + an optional superuser-DSN decision, not migration code.)**
- Or split schema creation from extension enablement — run the `CREATE EXTENSION` step under a separate, clearly-documented superuser DSN (a `migrationSuperuserDsn` chart value), leaving normal migrations on the app role;
- Or, for CNPG adopters, document the `managed.roles` / `postInitSQL` path so the extension is created at cluster-init by the bootstrap superuser.

## Acceptance criteria

- [ ] Chart docs + the `values.yaml` postgres section comment state the hard prerequisite: the role in `postgres.credentialsSecret` must be superuser-capable for the first migration OR the `vector` extension must be pre-created — with the exact CNPG `postInitSQL` / psql one-liner.
[...]

Issue #2391, opened 2026-07-12
Task: fix the fresh-install ServiceAccount ordering deadlock on the migrate pre-install hook — cold helm install must schedule the migration Job


damir-topic:
Parent goal: #221
Parent initiative: #2390

## Summary

On a **fresh install**, the `meho-migrate` Job is a Helm `pre-install` hook, but the chart's `ServiceAccount` is a *non-hook* resource. Helm applies hooks **before** the normal manifests, so at hook time the `meho` ServiceAccount does not exist yet. The migration pod can never be scheduled — the kubelet/apiserver returns `serviceaccount "meho" not found` on every admission attempt — and after ~14 `FailedCreate` retries the release times out.

## Repro

1. Fresh namespace, `serviceAccount.create: true` (the default).
2. `helm install meho oci://ghcr.io/evoila/meho-chart --version 0.21.0 -n meho -f values.yaml --timeout 10m`

Observed:

```
Error: INSTALLATION FAILED: failed pre-install: resource Job/meho/meho-migrate not ready.
status: Failed, message: Job Failed. failed: 1/1
```

with the migrate Job's pod events showing repeated `FailedCreate` / `serviceaccount "meho" not found`.

## Root cause

Helm's apply order is: CRDs → `pre-install` hooks → normal resources → `post-install` hooks. A `pre-install` hook that references a ServiceAccount created as a normal (non-hook) resource is a classic ordering deadlock. The migration Job depends on a resource that Helm hasn't created yet.

## Suggested fixes (any one)

- Annotate the ServiceAccount (and its RBAC, if any) as a `pre-install`/`pre-upgrade` hook with a `hook-weight` lower than the migrate Job so it lands first; or
- Give the migrate Job its **own** dedicated hook ServiceAccount created in the same hook phase at a lower weight; or
- Drop the migrate Job's hook annotation and order it via an init-container / `helm.sh/hook-weight` on a normal resource so it shares the release's normal apply ordering.

[...]

PR #2418, opened 2026-07-12
fix(chart): add backplane startupProbe to survive slow first boots (#2393)

### LINKED ISSUE #2393: Task: add a startupProbe to the backplane Deployment — stop liveness crash-looping slow first boots
Parent goal: #221
Parent initiative: #2390

## Summary

On first boot the backplane registers its full typed-op catalog (hundreds of ops) before it opens `:8000` — this takes ~2-3 minutes. The chart ships a liveness probe with a short `initialDelaySeconds` (default `30`) and **no `startupProbe`**, so the kubelet starts liveness-checking mid-registration, the check fails, the container is killed, and the pod enters a CrashLoopBackOff before it ever finishes its first successful boot.

**Grounding note (2026-07-12):** confirmed — `values.yaml:132-148` defines only liveness/readiness, `deployment.yaml:127-130` renders no startupProbe, and no `probes.startup` knob exists. Attribution is broader than registration alone: the pre-bind lifespan also awaits the fastembed embedding-model preload (`main.py:415`), which on a cold install with an empty cache PVC downloads the model weights — size the default startup budget for both contributors.

## Repro

Install on a cluster where first-boot registration takes longer than `liveness.initialDelaySeconds` (any cold install, in our experience). Observed pod history:

```
0/1  Running   1 (3s ago)     21s
0/1  CrashLoopBackOff  1
0/1  Running   2
0/1  CrashLoopBackOff  2
... (liveness kills it repeatedly during registration)
```

Eventually — with a long-enough delay — it boots and goes `1/1 Running`, but out of the box the default probe timing turns a slow-but-healthy first boot into a crash loop.

## Suggested fix

Add a `startupProbe` to the chart (with configurable `failureThreshold` × `periodSeconds` giving a generous first-boot budget, e.g. 5 min). A `startupProbe` is exactly the right primitive here: it gates liveness/readiness until first boot succeeds, then hands off to the normal (short-delay) liveness probe. Expose the knobs under `probes.startup.*`.

Right now the chart has **no startupProbe knob at all**, so the only lever is inflating `probes.liveness.initialDelaySeconds`, which is a worse tool (it also delays detection of a genuinely-hung process for the whole life of the pod).

## Acceptance criteria

- [ ] `deployment.yaml` renders a `startupProbe` on `/healthz` from `probes.startup.*` with defaults giving a ≥5-minute first-boot budget (e.g. `failureThreshold: 30` x `periodSeconds: 10`).
- [ ] `values.schema.json`'s probes object accepts the new `probes.startup` subtree (it currently knows only liveness/readiness).
- [ ] Existing liveness/readiness defaults are unchanged; chart template tests + kubeconform pass.

## Out of scope

- Changing `/healthz`//`ready` semantics; moving registration/model-preload out of the lifespan (backend change — separate ticket if wanted).

## Workaround we used

```yaml
probes:
  liveness:
    initialDelaySeconds: 300
```

damir-topic:
## Summary

- The backplane's FastAPI lifespan registers the full typed-op catalog and preloads the fastembed embedding model **before** it binds `:8000` — ~2-3 min, longer on a cold install that downloads model weights into an empty cache PVC. The chart shipped only a short-delay `livenessProbe` and **no** `startupProbe`, so the kubelet started liveness-checking mid-registration and CrashLoopBackOff'd a slow-but-healthy pod before its first successful boot.
- Add a `startupProbe` on `/healthz` rendered from a new `probes.startup.*` values subtree, guarded by `{{- with .Values.probes.startup }}` so a fast cluster can opt out by clearing `probes.startup`. Default budget: `failureThreshold: 30` × `periodSeconds: 10` = 300s (5 min). The kubelet disables liveness/readiness until it first passes, then the short-delay liveness probe takes over for fast hang detection.
- `values.schema.json` accepts and validates the new subtree via the existing `httpProbe` definition. Liveness/readiness defaults are **unchanged**. `docs/codebase/devops.md` Probes section updated.

## Test plan

- [x] `helm lint` (with the CI override set) — clean
- [x] `helm template` renders `startupProbe` on the backplane Deployment: `/healthz`, `failureThreshold: 30`, `periodSeconds: 10`, `timeoutSeconds: 1` (= 300s / 5-min budget)
- [x] `kubeconform -strict -kubernetes-version 1.28.0` — 11/11 resources valid
- [x] **AC1** — `deployment.yaml` renders `startupProbe` on `/healthz` from `probes.startup.*` with a ≥5-min default budget
- [x] **AC2** — schema accepts `probes.startup`; meaningfully validated (`--set probes.startup.bogusField=1` → `additional properties 'bogusField' not allowed`; valid `failureThreshold` override passes)
- [x] **AC3** — liveness (`initialDelaySeconds: 30`, `failureThreshold: 3`) and readiness (`/ready`, `initialDelaySeconds: 5`) defaults unchanged; template tests + kubeconform pass
- [x] Opt-out verified — `--set probes.startup=null` renders no `startupProbe`
- [x] G0.8-T4 regression guards (`MCP_RESOURCE_URI` / `BACKPLANE_URL` derivation) still pass

## Out of scope

- Moving registration / model-preload out of the lifespan (backend change — separate ticket if wanted), per the issue's Out-of-scope.
- Sibling chart tickets #2391 / #2392 / #2394 touch the same chart; this PR is scoped to the `probes.startup` subtree to minimize serial-merge conflicts.

Closes #2393

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,20 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Fixed — Helm chart `startupProbe` stops liveness crash-looping slow first boots (#2393)
+
+- The backplane Deployment now renders a `startupProbe` on `/healthz` from a
+  new `probes.startup.*` values subtree. The kubelet disables the liveness and
+  readiness probes until the startup probe first passes, so a slow-but-healthy
+  first boot — full typed-op catalog registration plus the fastembed
+  embedding-model preload before the app binds `:8000`, ~2-3 min and longer on
+  a cold install that downloads model weights into an empty cache PVC — no
+  longer trips the short-delay liveness probe into a CrashLoopBackOff.
+- The default budget is `failureThreshold: 30` × `periodSeconds: 10` = 300s
+  (5 min), operator-tunable. `values.schema.json` accepts the new
+  `probes.startup` subtree; liveness/readiness defaults are unchanged. Opt out
+  on a fast cluster by clearing `probes.startup` (e.g. `--set probes.startup=null`).
+
 ### Changed — stage-aware `connector_auth_failed` causes + truthful remediation (#2400)
 
 - The `connector_auth_failed` envelope's `extras.cause` is now **stage-aware**
--- a/deploy/charts/meho/templates/deployment.yaml
+++ b/deploy/charts/meho/templates/deployment.yaml
@@ -124,6 +124,23 @@ spec:
           # would mask startup deadlocks and let an unready Pod accept
           # traffic. Every probe field is operator-tunable via
           # `.Values.probes.<kind>.<field>`.
+          #
+          # `startupProbe` gates liveness/readiness until first boot succeeds
+          # (Issue #2393). First boot runs the full typed-op catalog
+          # registration plus the fastembed model preload before the app binds
+          # :8000 — ~2-3 min, longer on a cold install that downloads model
+          # weights into an empty cache PVC. Without this probe the kubelet
+          # starts liveness-checking mid-registration and CrashLoopBackOffs a
+          # slow-but-healthy pod. The default budget is
+          # `failureThreshold × periodSeconds` = 30 × 10s = 300s (5 min); the
+          # kubelet disables liveness/readiness until it passes, then the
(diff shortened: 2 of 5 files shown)


damir-topic:
@coderabbitai review

PR #2473, opened 2026-07-14
fix(types): annotate _collect_2xx_response_codes candidates as list[str]

zdamir:
## What

Add an explicit `list[str]` annotation to `candidates` in `_collect_2xx_response_codes` (`backend/src/meho_backplane/operations/ingest/openapi.py`).

## Why

mypy **2.3+** infers the list comprehension over the string-literal tuple `("200", "201", "202", "203", "204")` as `list[Literal["200", ...]]` instead of `list[str]`. Because `list` is invariant, that narrower type then rejects everything the function does next:

- `candidates.append("2XX")` — incompatible literal
- `candidates.extend(key for key in responses ...)` — `str` not assignable
- `return candidates` — `list[Literal[...]]` vs declared `list[str]`

= 3 errors under strict mypy. The explicit `list[str]` annotation widens the element type and clears all three. **Behavior is unchanged**, and the annotation is a no-op under the currently pinned mypy 2.1.0.

## Context

This unblocks the Dependabot mypy `2.1.0 → 2.3.0` bump (#2441), which otherwise fails the strict mypy CI gate on these exact three errors. After this lands, `@dependabot rebase` on #2441 will make its mypy gate pass.

## Validation

Ran locally in isolated worktrees, replicating CI (`uv sync --locked --all-groups` → `uv run mypy src/`):

- **mypy 2.1.0 (current main):** no openapi.py errors before or after this change (no regression)
- **mypy 2.3.0 (the bump):** 3 errors before → 0 after
- `ruff check` + `ruff format --check`: clean

## Summary by CodeRabbit

* **Refactor**
  * Improved internal type annotations and static type-checking support.
  * No changes to runtime behavior or API response-code handling.

### DIFF
--- a/backend/src/meho_backplane/operations/ingest/openapi.py
+++ b/backend/src/meho_backplane/operations/ingest/openapi.py
@@ -1329,7 +1329,12 @@ def _collect_2xx_response_codes(responses: dict[str, Any]) -> list[str]:
     then any other key starting with ``"2"`` that wasn't already
     picked.
     """
-    candidates = [c for c in ("200", "201", "202", "203", "204") if c in responses]
+    # Annotate as list[str] so mypy widens the element type: the
+    # comprehension over a tuple of string literals otherwise infers
+    # list[Literal["200", ...]], which then rejects the "2XX" append,
+    # the str-keyed extend, and the list[str] return (list is
+    # invariant). mypy 2.3+ makes this inference; 2.1 did not.
+    candidates: list[str] = [c for c in ("200", "201", "202", "203", "204") if c in responses]
     if "2XX" in responses:
         candidates.append("2XX")
     candidates.extend(


zdamir (commented):
> _Posted as a comment, not a formal Approve: GitHub blocks approving your own PR. Verdict below is **APPROVE** — no blockers, no changes requested._

## /review-pr-slim summary

**Anchor task:** none — maintenance/typing fix unblocking Dependabot #2441 (mypy `2.1.0 → 2.3.0`). No project-19 task; filing a Goal→Initiative→Task for a one-line annotation would be over-structuring.
**Surface area:** backend (Python)
**Verdict:** APPROVE

### What this changes
Adds an explicit `list[str]` annotation to `candidates` in `_collect_2xx_response_codes` (`backend/src/meho_backplane/operations/ingest/openapi.py`). Runtime behavior identical; return type (`list[str]`) unchanged.

### Correctness
- The annotation matches the function's declared return type and the value it builds. mypy 2.3's narrower `list[Literal[...]]` inference (which rejects the `"2XX"` append, the `str`-keyed `extend`, and the `list[str]` return under list invariance) is resolved by widening the element type — the canonical fix, no runtime effect.
- Verified empirically (not doc-recall): under mypy **2.1.0** the file is clean before and after (no regression); under mypy **2.3.0** the 3 errors go from present → absent with this change.

### Blast radius
- Single internal caller `_extract_response_schema` (openapi.py:1365); consumes the result as `list[str]` — no contract change, no other call sites, no consumers affected. 🕳️ nothing missing, 🚫 nothing undisclosed.

### Integration verification (ran in `pr-2473-review` worktree)
- [x] `ruff check` — clean
- [x] `mypy src/` — clean (only pre-existing macOS-only `icmp.py MSG_ERRQUEUE` false positive; Linux CI has the constant, so green there)
- [x] `pytest tests/test_operations_ingest_openapi.py` — **130 passed**, incl. `test_parse_petstore_31_yaml_wildcard_2xx_response` (the `2XX` append path), `test_parse_petstore_30_204_response_yields_none`, and response-schema-extraction tests — the exact paths this function drives
- n/a migrations / docker / UI / agent — none touched

### Bot review
- CodeRabbit reviewed at head `cc65ad92`: **"No actionable comments were generated."** All 5 pre-merge checks passed.

### Findings
- 🛑 0 blockers · ⚠️ 0 majors · 💡 0 minors · ❓ 0 questions

- The 5-line explanatory comment is verbose relative to the one-line change, but it's justified: it records a non-obvious mypy-version gotcha and discourages a future "simplification" that would re-break the strict gate. Fine as-is; could be trimmed to 2 lines if the file's comment density is later tightened.

### Docs
- n/a — no `docs/codebase/` area covers this internal helper; a one-line typing annotation doesn't warrant one.

### What worked well
- Minimal, surgical fix with the rationale captured in-code so the annotation survives future edits.
- Correctly widens rather than reaching for `cast`/`# type: ignore`, keeping the value genuinely type-checked.

Issue #2506, opened 2026-07-15
Task: add the Dashboard entity + five-state worst-of rollup (hysteresis, severity caps, UNKNOWN→degraded) and the /ui/checks console surface — one answer to "is everything OK?"


damir-topic:
Parent goal: #221
Parent initiative: #2416
Depends-on: #2503

## Summary

Add the **Dashboard** entity (named, tenant-scoped composition of Sensors, many-to-many) and the **five-state worst-of rollup** (`ok / degraded / critical / unknown / skip`) with `UNKNOWN→degraded`, per-Sensor severity caps, and `for:` hold-time hysteresis — then surface it as a **`/ui/checks`** console page (list + detail) in the existing HTMX/Jinja chassis plus a minimal REST CRUD so an operator can compose Sensors into a Dashboard and see one rolled-up answer to "is everything OK?". Scope item 4 of Initiative #2416.

## Current state

- **No Dashboard/rollup/checks surface exists; this task creates all three.** `grep -ri "dashboard" backend/src/meho_backplane/db/models.py` returns nothing — no dashboard table; `ls backend/src/meho_backplane/ui/routes/` has no `checks/`; `ui/templates/base.html:36-54` `nav_surfaces` registry has no checks entry.
- Naming caution: `ui/routes/dashboard.py` already exists and is the **`GET /ui/` landing page** (`build_dashboard_router`, included at `ui/routes/__init__.py:176`) — the new entity must not collide with it, hence table `check_dashboards` and surface slug `checks` (not `dashboard`).
- Console surface mould (list + detail, the ONE precedent to copy): the scheduler surface, Task #1826 —
  - `ui/routes/scheduler/__init__.py:74-94` — `build_scheduler_router()` factory aggregating sub-routers; registration order literal-prefix-before-parametrised is documented as load-bearing at `:83-88`.
  - `ui/routes/scheduler/list_view.py:143-145` — one handler serving full page (`scheduler/list.html`) vs HTMX rows fragment (`scheduler/_table_rows.html`) branched on `HX-Request`; `:72` `_LIST_LIMIT = 200` posture; `:105-118` in-process service call (session-cookie BFF cannot auth Bearer REST routes — rationale at `ui/routes/scheduler/__init__.py:14-25`).
  - `ui/routes/scheduler/detail.py:4` — `GET /ui/scheduler/{trigger_id}` detail with cross-tenant/absent-id collapsed to 404.
[...]

Issue #2505, opened 2026-07-15
Task: build the deterministic sensor check-runner — replica-safe no-LLM evaluation loop (interval + cron cadence) dispatching safe ops under a system identity


damir-topic:
Parent goal: #221
Parent initiative: #2416
Depends-on: #2503
Depends-on: #2504

## Summary

Build the deterministic, no-LLM check-runner for Initiative #2416: a lifespan-owned asyncio loop that claims due Sensor rows (#2503), dispatches each sensor's `safe` read-only op through `dispatch()` under a synthetic per-tenant identity, feeds the payload to the typed assertion evaluator (#2504), and persists `{state, value, evidence, evaluated_at}` via #2503's `record_sensor_result`. One runner handles both cadence kinds: interval seconds (sub-minute) and 5-field cron (≥1 min), both riding the same durable claim/advance discipline the #804 scheduler proved out. No agent, no model call, no LLM anywhere on this path.

## Current state

- **No runner or sensor runtime exists.** At Initiative start `grep -ril "sensor" backend/src/meho_backplane/` matched nothing; #2504 landed the pure `checks/` package (evaluator + `CheckState`) and #2503 extended it with the Sensor entity + registry (`checks/schemas.py`, `checks/repository.py`, `checks/service.py`). This task adds the runner beside them: `checks/runner.py`.
- **Interval-tick loop mould (sub-minute cadence):** the memory-expiry sweeper `_sweeper_loop` at `backend/src/meho_backplane/memory/expiry.py:222-253` — sleep-then-tick, per-tick `try/except`, `CancelledError` propagates; `start_memory_expiry_sweeper`/`stop_memory_expiry_sweeper` at `expiry.py:256-283` return/cancel a strongly-referenced `asyncio.Task`. Same loop shape in `topology/scheduler.py:255-277`.
[...]

Issue #2503, opened 2026-07-15
Task: add Sensor entity + registry — scheduled-trigger-moulded table, safe-only 422 create guard, REST/MCP/CLI CRUD for the deterministic check layer


damir-topic:
Parent goal: #221
Parent initiative: #2416
Depends-on: #2504

## Summary

Create the **Sensor** entity — the deterministic check layer's registration substrate under Initiative #2416. A Sensor row carries `(op + args + assertion + cadence + severity)`, modelled on the `scheduled_trigger` row but deliberately a **new table**: `ScheduledTrigger.agent_definition_id` is `NOT NULL` with a real FK (`db/models.py:4000-4004`, `scheduler/repository.py:84-86`), so the trigger row structurally cannot carry an op-based check. This task ships the table + migration, the tenant-scoped CRUD service, the safe-only 422 create guard, and the REST/MCP/CLI surfaces; the assertion payload is validated at the wire by #2504's already-landed `AssertionSpec` (hard dependency). The runner (#2505), dashboard/rollup/UI (#2506), and investigator (#2507) build on this storage shape.

## Current state

- **No sensor entity or registry exists** — `grep -ril "sensor" backend/src/meho_backplane/ --include="*.py"` returns nothing at Initiative start. **#2504 (a dependency of this task) lands the pure `backend/src/meho_backplane/checks/` package** — `checks/assertions.py` (frozen `AssertionSpec` + the five-state `CheckState` vocabulary `ok|degraded|critical|unknown|skip`) and `checks/evaluate.py`. This task adds the entity, storage, and surfaces beside them: it EXTENDS that package, never re-creates it.
[...]

Issue #1301, opened 2026-05-28
Task: G12.3-T2 Runbook step-execution engine + runtime substitution helper


damir-topic:
Parent initiative: #1198
Parent goal: #1195

Depends-on: #1295, #1300

## Problem

The engine is the load-bearing isolation. **It's the place where step opacity, verify gating, and state-machine transitions live.** Keeping it as a pure-function module (no DB, no HTTP, no MCP) lets the property-style opacity tests (`advance(...)` never returns a step body that isn't the current step) be reviewed in isolation, without HTTP/MCP transport noise.

Also lands the runtime substitution helper — distinct from the publish-time `validate_substitutions()` shipped by #1295 (which rejects disallowed patterns at template publish). T2's helper **resolves** the allowed patterns at advance time using the run's `target` and `params`.

## Current state

- After #1295 lands, `runbooks/schemas.py` carries `Step`, `Verify`, `OperationCallStep`, `ManualStep`, `ConfirmVerify`, `OperationCallVerify`, `RunbookTemplateBody`, plus the publish-time `validate_substitutions()` helper.
- After #1300 lands, `runbooks/runs_schemas.py` carries `StepBody`, `StepBodyVerify`, `VerifyResponse`, `CurrentStepResponse`, `RunCompletedResponse`, `NextStepResponse`, etc.
- The state machine (per #1292 / #1296):
  - `RunbookRun.state` ∈ {`in_progress`, `completed`, `abandoned`}.
  - `RunbookRunStepState.state` ∈ {`pending`, `in_progress`, `verified`, `failed`}.
- No engine module exists. T2 introduces `runbooks/engine.py` and `runbooks/substitution.py`.

### New module — `runbooks/substitution.py`

Pure-function helpers for resolving `${run.target}` and `${run.params.X}` at advance time. No DB, no contextvars.

```python
__all__ = ["RUN_TARGET_PATTERN", "RUN_PARAMS_PATTERN", "resolve_substitutions"]


#: Matches the literal ``${run.target}`` substitution token.
RUN_TARGET_PATTERN: Final[re.Pattern[str]] = re.compile(r"\$\{run\.target\}")

#: Matches ``${run.params.X}`` where X is ``[a-z_][a-z0-9_]*``.
RUN_PARAMS_PATTERN: Final[re.Pattern[str]] = re.compile(r"\$\{run\.params\.([a-z_][a-z0-9_]*)\}")


def resolve_substitutions(
    value: object,
    *,
    target: str,
[...]

PR #2554, opened 2026-07-16
feat(checks): bounded typed assertion evaluator — pure select→compare (#2504)

### LINKED ISSUE #2504: Task: add the bounded typed assertion evaluator — pure select→compare (dotted-path + one aggregate → typed comparator) emitting {state, value, evidence}
Parent goal: #221
Parent initiative: #2416

## Summary

Build the deterministic heart of Initiative #2416's check layer: a **pure, no-I/O assertion evaluator** that maps `(assertion spec, op-result payload, now)` to a typed verdict `{state, value, evidence}`. Assertions are exactly two bounded stages — **select** (dotted-path + at most one bounded aggregate: `max`/`min`/`sum`/`count`/`any`/`all`) → **typed comparator** (`threshold` / `equals`|`in` / `bool` / `freshness`) — with no free-form code. This task creates the `checks/` package and lands the spec schemas, the evaluator, the set-wide five-state vocabulary, and exhaustive unit tests, nothing else. Sibling tasks build on it: #2503 (which depends on this task) stores Sensors and validates their `assertion` payload by importing `AssertionSpec`; #2505 feeds the evaluator from the check-runner; #2506 rolls verdicts up into Dashboards.

## Current state

- **No checks/sensors package or assertion evaluator exists** — `ls backend/src/meho_backplane/ | grep -i 'check\|sensor'` is empty. This task creates `backend/src/meho_backplane/checks/`; #2503 (downstream, depends on this task) extends the same package with the Sensor entity's schemas/repository/service.
- **No JSONPath library is in the dependency tree**: `grep -c jsonpath backend/uv.lock` returns 0; `cd backend && uv run python -c "import jsonpath_ng"` → `ModuleNotFoundError` (verified 2026-07-15); `backend/pyproject.toml` `[project].dependencies` contains no path-query lib.
- **Op results arrive in the dispatch envelope**: `dispatch()` (`backend/src/meho_backplane/operations/dispatcher.py:1893`) never raises and returns a frozen `OperationResult` (`backend/src/meho_backplane/connectors/schemas.py:616`) whose `result: dict[str, Any] | list[Any] | None` (`connectors/schemas.py:632`) is the JSON the assertion runs over. The envelope carries `status` (`:630`, `"ok"|"pending"|"error"|"denied"`) and `duration_ms` (`:634`) but **no timestamp field** — relevant to the `freshness` design below.
- **Reducer seam**: after a handler returns, the dispatcher runs the JSONFlux reducer (`backend/src/meho_backplane/operations/reducer.py:4-19`); the production `JsonFluxReducer` (`operations/jsonflux_reducer.py`) replaces large set-shaped payloads (>50 rows / >4 KB) with a markdown summary + `ResultHandle`. So `OperationResult.result` is not always the raw op output; sourcing the pre-reduction payload is the runner's problem (#2505 — `AuditLog.raw_payload`, `db/models.py:507`, retains the raw result as one candidate source). The evaluator assumes nothing about payload shape.
- **NOT-this precedents (binding per #2416)**: the in-house jsonflux SQL engine `query()` (`backend/src/meho_backplane/jsonflux/query/engine.py:541`) is free-form SQL and explicitly not the evaluator. The same file's `_extract_path` (`engine.py:513-539`) is an in-house dotted-path navigator (`$.` prefix, dict keys, numeric index, `items[]`) proving a subset walker is ~25 lines of stdlib.
- **Substrate-minimalism precedent**: the runbook verify `OperationCallVerify` (`backend/src/meho_backplane/runbooks/schemas.py:182-198`, #1301) deliberately shipped "no operators, no JSONPath, and no boolean composition"; its `type`-discriminated union `Verify = Annotated[..., Field(discriminator="type")]` (`runbooks/schemas.py:205-208`) is the schema mould this task copies.
- **Pure-module moulds**: `backend/src/meho_backplane/scheduler/cron.py` (pure functions; aware-datetime discipline at `cron.py:22-31` — naive bases are rejected because they make results host-dependent) and `backend/src/meho_backplane/operations/composite_invariant.py` with its dedicated test file `backend/tests/test_operations_composite_invariant.py`.

## Desired state

New package `backend/src/meho_backplane/checks/` containing:

1. `assertions.py` — frozen Pydantic spec models:
   - `SelectSpec { path: str, aggregate: Literal["max","min","sum","count","any","all"] | None = None }`, with the path compiled/validated at model-parse time.
   - Comparator union discriminated on `type` (exactly five members): `ThresholdCompare`, `EqualsCompare`, `InCompare`, `BoolCompare`, `FreshnessCompare`; `Compare = Annotated[..., Field(discriminator="type")]`.
   - `AssertionSpec { select: SelectSpec, compare: Compare }`.
   - `CheckState = Literal["ok","degraded","critical","unknown","skip"]` — **the set-wide five-state vocabulary, declared once here**. `degraded` (not `warn`) is the middle state: the Initiative's rollup rule is `UNKNOWN→degraded` and #2503's per-Sensor `severity` values are `degraded|critical`, so one word serves state, severity, and cap. #2503 pins these five values into the sensor row's DB CHECK with a drift-guard equality test; #2505/#2506 import this type — nobody re-declares it.
   - `AssertionOutcome { state, value: str|int|float|bool|None, evidence: dict }` (frozen; `evidence` JSON-serializable — persisted by #2505, rendered by #2506).
2. `evaluate.py` — `evaluate_assertion(spec: AssertionSpec, payload: object, *, now: datetime) -> AssertionOutcome`. Pure function: no I/O, no async, no imports beyond stdlib/pydantic/own package. **Never raises on payload data** — every payload/spec mismatch becomes `state="unknown"` with `evidence["reason"]` (mirrors the dispatcher's never-raises contract, `dispatcher.py:1904-1906`, and the Reducer Protocol's tolerate-every-shape rule, `operations/reducer.py:62-66`). The single permitted exception: `ValueError` on a naive `now` (caller misuse, not data — same posture as `scheduler/cron.py:22-31`).

Design decisions bound here (each grounded):

- **Dotted-path subset instead of a JSONPath dependency.** Grammar: `path := "$" segment*`; `segment := "." name | "[" int "]" | "[*]"`; `name := [A-Za-z0-9_-]+`; at most one `[*]` (a second is a validation error). `$` alone selects the root. Navigation is strict: missing key / out-of-range index / type mismatch (including inside a `[*]` projection) → `unknown`, evidence names the failing segment. Justification: no JSONPath lib is installed (grounded above); `jsonpath-ng` ships recursive descent, filter expressions and an extended parser with arithmetic — an open surface #2416's Out-of-scope bans ("free-form assertion languages"), so we would have to deny-list most of the library we just pinned; the in-house `_extract_path` (`engine.py:513-539`) already proves the subset walker pattern; and the runbook verify (`runbooks/schemas.py:186-190`) is the repo's explicit precedent for refusing JSONPath expressivity per substrate minimalism. `cron.py`'s header documents the bar for adding a dep ("no stdlib alternative") — here the stdlib alternative is ~60 lines.
- **Select/aggregate semantics.** No aggregate → selection must resolve to a scalar (`str|int|float|bool|None`), else `unknown`. With aggregate → selection must resolve to a list, else `unknown`. `count`: any list → int. `sum`/`max`/`min`: all elements must be `int|float` (bool excluded), else `unknown`. `any`/`all`: all elements must be strictly `bool` (no truthiness coercion), else `unknown`. Empty list: `count=0`, `sum=0`, `any=False`, `all=True` (Python identity semantics, documented in the module docstring); `max`/`min` on empty → `unknown`.
- **Comparators.**
  - `threshold { op: "gt"|"gte"|"lt"|"lte", degraded: float|None, critical: float|None }` — at least one bound required (model validator); when both are set the validator enforces degraded-less-severe ordering (`gt`/`gte`: degraded ≤ critical; `lt`/`lte`: degraded ≥ critical). Violation test is `value <op> bound`, critical checked first → `critical`/`degraded`/`ok`. Non-numeric (or bool) input → `unknown`.
  - `equals { value: scalar }` / `in { values: [scalar, ...] (min 1) }` — strict equality: `bool` never equals `int` (Python's `True == 1` footgun explicitly neutralized). Mismatch → `critical`.
  - `bool { expect: bool = True }` — selected value must be strictly `bool`, else `unknown`; mismatch → `critical`.
  - `freshness { max_age_seconds: positive number, degraded_age_seconds: positive number < max_age_seconds | None }` — **the timestamp source is the selected value from the op payload**: an RFC3339 string (parsed with `datetime.fromisoformat`; verified on the installed Python 3.12.3 to accept both `Z` and numeric offsets) or an `int|float` epoch-seconds (interpreted as UTC). Naive strings (no offset) → `unknown` — host-TZ-dependent comparison is banned, same reasoning as `cron.py:22-31`. `age = (now - ts).total_seconds()`; `age > max_age_seconds` → `critical`, `> degraded_age_seconds` → `degraded`, else `ok` (future timestamps → `ok`). `now` is a required tz-aware keyword the runner injects (`datetime.now(UTC)` in #2505): the envelope has no timestamp field to read instead (`connectors/schemas.py:616-636`), and "the check itself is stale / didn't run" is #2506's rollup-level UNKNOWN, not an assertion.
- **Emitted states are a subset**: the evaluator emits only `ok|degraded|critical|unknown`. `skip` (unreachable-by-design) is a scheduling-time fact a pure function cannot observe — #2506 derives it (paused sensors). Per-Sensor severity caps are applied at rollup (#2506); the evaluator's `critical` is pre-cap.
- **Evidence minimum keys**: `path`, `aggregate`, `comparator` (the type tag), `observed` (post-select value or failure description), the comparator's expected bounds/values, and `reason` when `state="unknown"`.

## Implementation sketch

- Add `backend/src/meho_backplane/checks/__init__.py` (re-export `AssertionSpec`, `SelectSpec`, `Compare`, `CheckState`, `AssertionOutcome`, `evaluate_assertion`), `checks/assertions.py`, `checks/evaluate.py`. (#2503 later adds DB-facing siblings — `schemas.py`/`repository.py`/`service.py` — to this package; keep these two modules dependency-pure.)
- `assertions.py`: copy the runbooks discriminated-union mould (`runbooks/schemas.py:168-208`) — `model_config = ConfigDict(frozen=True)`, `type: Literal[...]` tags, `Annotated[..., Field(discriminator="type")]`. Compile `SelectSpec.path` into segments in a field validator so an invalid path (bad grammar, >1 `[*]`) fails at model-parse time — #2503 reuses exactly this model to 422 bad Sensor payloads at create.
- `evaluate.py`: private `_select` / `_aggregate` / `_compare` helpers + public `evaluate_assertion`. Port the navigation approach of jsonflux `_extract_path` (`engine.py:513-539`) but strict and pre-validated; do NOT import or call the jsonflux engine.
- Tests `backend/tests/test_checks_assertions.py` (mould: `backend/tests/test_operations_composite_invariant.py` — pure module, no DB/app fixtures). Table-driven: grammar accept/reject, each aggregate incl. empty-list semantics, each comparator's state matrix, bool-vs-int strictness, freshness against a fixed injected `now` (Z, offset, epoch, naive, garbage), and a degenerate-payload sweep (`None`, scalars, deep nesting, a reducer summary dict) asserting no exception and no `skip` ever emitted.
- `CHANGELOG.md` `[Unreleased]` entry under the unique header `### Checks — bounded assertion evaluator (#2416)`.
- No FastAPI routes, no DB models, no migration, no CLI/OpenAPI impact.

## Acceptance criteria

- [ ] `cd backend && uv run pytest tests/test_checks_assertions.py -q` passes; the file contains (at least) tests named `test_path_grammar_rejects_second_wildcard`, `test_path_missing_key_yields_unknown`, `test_aggregate_empty_list_semantics`, `test_threshold_degraded_then_critical_ordering`, `test_equals_bool_is_not_int`, `test_freshness_naive_timestamp_unknown`, `test_freshness_epoch_seconds`, `test_evaluator_never_raises_or_emits_skip` (verify: `uv run pytest tests/test_checks_assertions.py -q --collect-only | grep -c test_` ≥ 8).
- [ ] `grep -n 'discriminator="type"' backend/src/meho_backplane/checks/assertions.py` hits, and `grep -oE 'Literal\["(threshold|equals|in|bool|freshness)"\]' backend/src/meho_backplane/checks/assertions.py | sort -u | wc -l` returns 5 — exactly the five comparator tags, no extras.
- [ ] Vocabulary guard: `grep -rn '"warn"\|"warning"' backend/src/meho_backplane/checks/` returns nothing — the middle state and all field names say `degraded`, matching the Initiative's `UNKNOWN→degraded` and #2503's severity values.
- [ ] Purity (per-file — #2503 later adds DB-facing siblings to the package): `grep -nE '^(import|from) +(httpx|sqlalchemy|asyncpg|redis|fastapi|meho_backplane\.(db|operations|connectors|jsonflux))' backend/src/meho_backplane/checks/assertions.py backend/src/meho_backplane/checks/evaluate.py` returns nothing; `grep -n 'async def' backend/src/meho_backplane/checks/evaluate.py` returns nothing.
- [ ] No new dependency: the PR diff (`git diff --name-only origin/main`) contains neither `backend/pyproject.toml` nor `backend/uv.lock`; `grep -c jsonpath backend/uv.lock` returns 0.
- [ ] End-to-end spot check prints `critical 3.1`: `cd backend && uv run python -c "from datetime import UTC, datetime; from meho_backplane.checks import AssertionSpec, evaluate_assertion; s=AssertionSpec.model_validate({'select':{'path':'$.disks[*].free_pct','aggregate':'min'},'compare':{'type':'threshold','op':'lt','degraded':20,'critical':5}}); o=evaluate_assertion(s, {'disks':[{'free_pct':3.1},{'free_pct':42.0}]}, now=datetime.now(UTC)); print(o.state, o.value)"`.
- [ ] Unknown-not-raise spot check prints `unknown`: same command shape with payload `None` prints `unknown` and a non-empty `evidence['reason']`.
- [ ] `CheckState` exports all five states while the evaluator emits four: `grep -n '"skip"' backend/src/meho_backplane/checks/assertions.py` hits; `grep -n '"skip"' backend/src/meho_backplane/checks/evaluate.py` returns nothing.
- [ ] No migration: `git diff --name-only origin/main -- backend/alembic/` is empty (`has_migration: false`).
- [ ] No route/OpenAPI/CLI impact: `git diff --name-only origin/main -- backend/src/meho_backplane/api/ cli/` is empty (hence no `make snapshot-openapi && make generate` needed for this task).
- [ ] `grep -n '### Checks — bounded assertion evaluator' CHANGELOG.md` hits inside `[Unreleased]`.

## Out of scope

- Sensor entity, registry, and the safe-op-only 422 create gate — #2503 (a downstream task that depends on this one and imports `AssertionSpec` from `checks/assertions.py` to validate Sensor payloads at the wire).
- Dispatching ops, sourcing the raw pre-reducer payload (candidate: `AuditLog.raw_payload`, `db/models.py:507`), cadence, and persisting outcomes — #2505.
- Worst-of rollup, per-Sensor severity caps, `for:` hold-time hysteresis, deriving `skip` (paused) and stale-`unknown`, Dashboard entity, `/ui/checks` — #2506. Investigator — #2507.
- Any free-form assertion language (SQL/CEL/embedded code); reusing the jsonflux SQL engine (`engine.py:541`); pinning `jsonpath-ng` or any path-query dependency; recursive descent, filters, multiple `[*]`, aggregate composition (exactly one aggregate max); cross-op assertions (an assertion sees exactly one op result).
- Remote/isolated-network Sensors — #2415 gateway workload later (soft link, per #2416).

## References

- Initiative #2416, Scope item 2 (binding: two bounded stages, `{state, value, evidence}`, no free-form code); parent goal #221. Soft link: #2415.
- Prior art: #1301 runbook verify — `backend/src/meho_backplane/runbooks/schemas.py:182-198` (deliberate no-JSONPath structural match) and `:205-208` (the `Field(discriminator="type")` union mould).
- `backend/src/meho_backplane/jsonflux/query/engine.py:513-539` (`_extract_path`, in-house dotted-path precedent) vs `:541` (`query()`, the free-form SQL NOT-evaluator).
- `backend/src/meho_backplane/connectors/schemas.py:616-646` — `OperationResult` envelope (`result` at `:632`; no timestamp field).
- `backend/src/meho_backplane/operations/dispatcher.py:1893` — `dispatch()` (never-raises contract at `:1904-1906`).
- `backend/src/meho_backplane/operations/reducer.py:4-19,62-66` — reducer seam + tolerate-every-shape rule; `db/models.py:507` — `AuditLog.raw_payload` (raw-result pointer for #2505).
- `backend/src/meho_backplane/scheduler/cron.py:22-31` — aware-datetime discipline; module header — dependency-addition bar.
- `backend/src/meho_backplane/operations/composite_invariant.py` + `backend/tests/test_operations_composite_invariant.py` — pure-module + dedicated-test mould.
- Installed-runtime verification (2026-07-15, Python 3.12.3): `cd backend && uv run python -c "from datetime import datetime; print(datetime.fromisoformat('2026-07-14T10:00:00Z'), datetime.fromisoformat('2026-07-14T10:00:00+02:00'))"` — RFC3339 `Z` and offset both parse; `uv run python -c "import jsonpath_ng"` → `ModuleNotFoundError`.

damir-topic:
## Summary

- Adds `backend/src/meho_backplane/checks/` — the deterministic heart of Initiative #2416's check layer: a **pure, no-I/O** assertion evaluator that maps `(AssertionSpec, op-result payload, now)` to a typed `{state, value, evidence}` verdict.
- Assertions are exactly two bounded stages: **select** (a strict dotted-path subset + at most one bounded aggregate `max`/`min`/`sum`/`count`/`any`/`all`) → **typed comparator** (`threshold` / `equals` / `in` / `bool` / `freshness`). No free-form assertion language, and **no new dependency** — the dotted-path subset is ~60 lines of stdlib rather than pinning `jsonpath-ng` (whose recursive descent / filters / arithmetic #2416's out-of-scope bans).
- Declares the set-wide five-state vocabulary `CheckState = {ok, degraded, critical, unknown, skip}` **once**, here (imported by #2503/#2505/#2506). The evaluator emits only the four observable states; `skip` is a scheduling-time fact a pure function cannot observe (#2506 derives it).
- **Never raises on payload data** — every payload/spec mismatch becomes `state="unknown"` with an `evidence["reason"]` (mirroring the dispatcher's never-raises contract and the reducer's tolerate-every-shape rule). The single permitted exception is `ValueError` on a naive `now` (caller misuse, same posture as `scheduler/cron.py`).

## Design notes (grounded in the issue's cited prior art)

- Discriminated-union spec mould copied from the runbook verify gate (`runbooks/schemas.py`, #1301): frozen models, `type` Literal tags, `Field(discriminator="type")`.
- Boolean-vs-int strictness: Pydantic 2.13 smart-union already preserves the JSON scalar type of an expected value (`True` stays `bool`, `1` stays `int`); the evaluator's `_strict_eq` neutralizes Python's `True == 1` footgun at compare time (both `equals` and `in`).
- Empty-list aggregate identity is fixed by contract: `count`→0, `sum`→0, `any`→False, `all`→True; `max`/`min` on empty → `unknown`.
- `freshness` reads the timestamp from the selected value (RFC3339 string **with** a required offset, or epoch seconds as UTC); a naive string is `unknown` (host-TZ-dependent comparison is banned). `now` is a required tz-aware keyword the runner injects — the op-result envelope carries no timestamp field.

## Test plan

- [x] `uv run pytest tests/test_checks_assertions.py -q` — 81 passed (table-driven; includes all eight acceptance-named tests; `--collect-only | grep -c test_` = 81 ≥ 8)
- [x] `uv run ruff check` — clean
- [x] `uv run ruff format --check` — clean
- [x] `uv run mypy src/meho_backplane` — Success, no issues (646 files)
- [x] Broad unit sweep (`tests`, excluding the testcontainers integration lane) — green
- [x] Spot check prints `critical 3.1`; `None`-payload spot check prints `unknown` with a non-empty `evidence['reason']`
- [x] Purity greps clean (no httpx/sqlalchemy/db/operations/connectors/jsonflux imports; no `async def` in `evaluate.py`); vocabulary guard clean (no `warn`/`warning`)
- [x] No new dependency (`pyproject.toml` / `uv.lock` untouched; `grep -c jsonpath uv.lock` = 0); no migration; no `api/` or `cli/` impact (no OpenAPI snapshot regen needed)

## Note on the CHANGELOG header

The CHANGELOG entry uses `### Added — Sensor assertion evaluator (#2504)` per the initiative-run's distinct-per-PR-header convention (avoids the `[Unreleased]` serial-merge collision). This is a deliberate, task-number-scoped wording of the issue's suggested `### Checks — bounded assertion evaluator (#2416)` line — same content, unique header.

## Out of scope (sibling tasks in Initiative #2416)

- Sensor entity + safe-op-only 422 create gate — #2503 (imports `AssertionSpec` from this package)
- Sourcing the raw pre-reducer payload, cadence, persisting outcomes — #2505
- Worst-of rollup, per-Sensor severity caps, `for:` hysteresis, deriving `skip`, Dashboard entity, `/ui/checks` — #2506

Closes #2504

## Summary by CodeRabbit

* **New Features**
  * Added a bounded assertion evaluator for inspecting operation results.
  * Supports `$`-prefixed path selection (with optional `[*]` wildcard), optional bounded aggregation, and comparators for thresholds, equality/membership, boolean, and freshness.
  * Produces typed outcomes (`ok`, `degraded`, `critical`, `unknown`) with observed values and evidence; mismatches are reported as `unknown` without raising (except for naive timestamps).
* **Documentation**
  * Updated the changelog with the new evaluator details.
* **Tests**
  * Added extensive tests covering path parsing, aggregation semantics, comparator behavior, validation rules, and evidence content.


damir-topic:
@coderabbitai review

backend/src/meho_backplane/checks/evaluate.py
```diff
@@ -0,0 +1,352 @@
...
+            )
+        bool_state: CheckState = "ok" if value == compare.expect else "critical"
+        return AssertionOutcome(state=bool_state, value=value, evidence=evidence)
+    return _compare_freshness(compare, value, now, evidence)
+
+
+def _compare_threshold(
+    compare: ThresholdCompare, value: Any, evidence: dict[str, Any]
+) -> AssertionOutcome:
+    if isinstance(value, bool) or not isinstance(value, (int, float)):
```
damir-topic, line 243: **[🛑 B1] — Non-finite (`NaN`/`Infinity`) observed values return `ok` instead of `unknown`** (also flagged by CodeRabbit)

**What:** A `NaN` value reaching a threshold compares `ok`: `nan > bound`, `nan < bound`, etc. are all `False`, so `_violates()` is `False` for every band and `state` stays `"ok"`. The aggregate path propagates it — `sum([nan, 1])` / `max([nan, 1])` → `nan` → `ok` — and `degraded`/`critical` bounds accept `inf`/`nan` at parse time.

**Why it matters:** This is reachable. Stdlib `json.loads('{"v": NaN}')` returns `{'v': nan}` by default, and op-result payloads are exactly this JSON (some real APIs emit `NaN`/`Infinity`; a `0/0` computed metric is `nan`). A `NaN` metric is a definitively-not-OK signal, so reporting `ok` is a silent false-negative — the most dangerous verdict for an alerting evaluator — and it contradicts this module's own "non-numeric → unknown" contract (a degenerate numeric should follow the same path).

**Required change:** Gate on `math.isfinite`:
- `_compare_threshold` (here, ~L229): a non-finite observed `value` → `unknown` (reason e.g. `"non-finite value"`), before the `_violates` checks.
- `_aggregate` sum/max/min (evaluate.py ~L181–189): reject a non-finite element or result → `unknown`.
- `ThresholdCompare` (assertions.py ~L179–180): reject non-finite `degraded`/`critical` at model-parse (field/model validator → 422).

Add tests: `NaN` value → `unknown`, `NaN` aggregate element → `unknown`, non-finite bound → `ValidationError`.

backend/src/meho_backplane/checks/assertions.py
```diff
@@ -0,0 +1,312 @@
...
+class SelectSpec(BaseModel):
+    """The select stage: a dotted path and at most one bounded aggregate.
+
+    With no aggregate the path must resolve to a scalar; with an aggregate it
+    must resolve to a list (either a ``[*]`` projection or a directly selected
+    array). The path is compiled at model-parse time so an invalid grammar is
+    a 422, not an evaluation-time surprise.
+    """
+
+    model_config = ConfigDict(frozen=True)
```
damir-topic: **[⚠️ M1] — Spec models accept unknown fields; a typo silently changes the assertion** (also flagged by CodeRabbit)

**What:** None of the seven spec models (`SelectSpec`, the five comparators, `AssertionSpec`) set `extra="forbid"`, so Pydantic's default `extra="ignore"` drops unknown keys. `{"type": "bool", "expectt": false}` validates cleanly with `expect` defaulted to `True` — the assertion is inverted vs. the operator's intent, with no error. A misspelled bound (`criticl`) likewise passes with the bound left unset.

**Why it matters:** This module is the wire-validation surface #2503 imports to 422 a bad Sensor `assertion` payload at create (per #2504's spec). Silently accepting a typo'd field defeats that 422 guarantee and can invert a safety-relevant check.

**Required change:** Add `extra="forbid"` to each model's `ConfigDict(frozen=True)` (all seven here). Add a test asserting an unknown field raises `ValidationError`.

damir-topic (commented):



**Anchor task:** #2504 — bounded typed assertion evaluator (pure `select → compare`)
**Surface area:** backend
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `930a73a5ed331aa9ee47acbc77ee873b72bab3af`
**AI reviewer (CodeRabbit):** received in ~40s (3 inline findings — all confirmed)

> Note on review state: GitHub forbids setting a formal `REQUEST_CHANGES`/`APPROVE` state on one's own PR (author == reviewer), so this review is transported as a `COMMENT`-state review. The **binding verdict is REQUEST_CHANGES**, recorded authoritatively in the machine-readable result file the orchestrator consumes. The one blocker (B1) below must be resolved before merge.

The package is clean, genuinely pure, well-documented, and its 81 table-driven tests exercise real branches (grammar accept/reject, strict navigation failures, every aggregate incl. empty-list identity, every comparator's state matrix, bool-vs-int strictness, freshness forms, and a never-raises/never-`skip` sweep). Every written acceptance criterion is met and lint/format/mypy/pytest are green. The blocker below is a reachable correctness gap the prose spec did not anticipate, not an AC miss.

### Acceptance criteria
- [x] `pytest tests/test_checks_assertions.py -q` — 81 passed; all 8 named tests present
- [x] Exactly 5 comparator `Literal` tags + `Field(discriminator="type")`
- [x] Vocabulary guard: no `warn`/`warning`; five-state `CheckState` declared once
- [x] Purity: no I/O imports, no `async def` in `evaluate.py`
- [x] `skip` in `assertions.py`, never in `evaluate.py`; evaluator emits only 4 states
- [x] No new dep / migration / api / cli touched; `jsonpath` absent from `uv.lock`
- [x] Spot checks: `critical 3.1` and `None`-payload → `unknown` (verified in-tree)
- [x] CHANGELOG entry under `[Unreleased]` — header intentionally `### Added — Sensor assertion evaluator (#2504)` (distinct-per-PR header to dodge the `[Unreleased]` serial-merge collision; a justified wording change from the issue's suggested `### Checks —` line, same content)

### Project-board alignment
|---|---|---|---|---|---|
| #2504 | Open | Closes | ✅ | ⚠️ full-minus-B1 | close on merge (after B1) |

No sibling PR touches `checks/`; #2503/#2505/#2506 correctly not yet open (they depend on this task). Anchor and hierarchy (#2504 → #2416 → #221) are coherent.

### Predicted vs. actual diff
- ✅ `checks/__init__.py`, `checks/assertions.py`, `checks/evaluate.py`, `tests/test_checks_assertions.py`, `CHANGELOG.md` — all expected and correct in shape
- 🚫 none undisclosed
- 🕳️ none missing

### Integration verification
- [x] `ruff check` — clean (checks pkg + test)
- [x] `ruff format --check` — 4 files already formatted
- [x] `mypy src/meho_backplane/checks/` — Success, no issues (3 files)
- [x] `pytest tests/test_checks_assertions.py -q` — 81 passed
- [x] Adversarial probes (in-tree `uv run`) — confirmed B1/M1/m1 empirically
- ⚠️ n/a — no migration / route / UI / agent path in this PR

### Findings (inline — each a merge blocker per team policy)

- 🛑 **B1** — Non-finite (`NaN`/`Infinity`) observed values return `ok` instead of `unknown`. Required change: gate on `math.isfinite` in `_compare_threshold` (non-finite value → `unknown`), in `_aggregate` sum/max/min (non-finite element/result → `unknown`), and reject non-finite `degraded`/`critical` bounds at model-parse in `ThresholdCompare`; add NaN/inf tests.
- ⚠️ **M1** — Spec models omit `extra="forbid"`, so a typo'd field (`expectt`, `criticl`) is silently dropped and the model validates with defaults — inverting the assertion and defeating the 422-at-create wire gate #2503 imports these models for. Required change: add `extra="forbid"` to all seven models' `ConfigDict`; add an unknown-field `ValidationError` test.

- **m1 (CodeRabbit finding 3, severity disputed):** `datetime.fromisoformat` accepts space-separated, compact, and week-date forms in addition to strict RFC3339. CodeRabbit rated this Major; I rate it non-blocking because every accepted form denotes the *same* instant, so the freshness verdict is still correct — it is a documentation/strictness gap, not a wrong-verdict bug. Tightening to a bounded RFC3339 shape check is a reasonable hardening for a later commit but is not required for merge.
- The validated path is re-compiled by `parse_path()` on every `evaluate_assertion` call (the field validator discards the compiled segments). Caching the compiled tuple on the model would avoid the re-parse; micro-optimization, no correctness impact.

- Adopted: 2 — non-finite handling (B1) and `extra="forbid"` (M1); both independently confirmed by in-tree probes.
- Disputed (severity, not validity): 1 — the RFC3339-strictness finding is real but non-blocking (no wrong verdict); recorded as adjacent m1.
- Out of scope: 0.

### Docs
- [x] n/a — no `docs/codebase/checks.md` exists yet and none is required by this task; the modules are thoroughly docstring'd and the CHANGELOG documents behavior. (A `docs/codebase/` artifact is a reasonable follow-up when #2503–#2506 land the surrounding surface.)

### What worked well (review-body only — never inline)
- The never-raises contract is honored precisely: every payload/spec mismatch routes to `unknown` with a named `reason`, and the sole `ValueError` is on a naive `now` (caller misuse), matching the scheduler's aware-datetime discipline.
- Boolean-vs-int strictness (`_strict_eq`) is correct in both directions and well-tested, neutralizing Python's `True == 1` footgun.
- The discriminated-union mould, frozen models, and single-source `CheckState` are exactly the substrate-minimal shape the Initiative called for.

damir-topic:
@coderabbitai review

backend/src/meho_backplane/checks/evaluate.py
```diff
@@ -0,0 +1,364 @@
...
+    if aggregate in ("sum", "max", "min"):
+        for element in values:
+            if isinstance(element, bool) or not isinstance(element, (int, float)):
+                return None, f"aggregate {aggregate!r} requires numbers, found {_typename(element)}"
+            # Only a float can be non-finite; an int is always finite (and passing
+            # a huge int to math.isfinite would itself raise OverflowError).
+            if isinstance(element, float) and not math.isfinite(element):
+                return None, f"aggregate {aggregate!r} requires finite numbers, found {element!r}"
+        if aggregate == "sum":
+            total = sum(values)
```
damir-topic: **[🛑 B2] — `sum` over a huge `int` + a `float` raises `OverflowError`, breaking the never-raises contract**

**What:** The element loop just above correctly rejects non-finite *floats* while letting any `int` through (calling `math.isfinite` on a huge int would itself raise). But `sum(values)` on a list that mixes an `int` beyond the float range (`> ~1.8e308`) and a `float` raises `OverflowError: int too large to convert to float` — Python must coerce the int to float to add it. Nothing catches it, so it escapes `evaluate_assertion`.

**Why it matters:** "Never raises on payload data" is an explicit acceptance criterion of #2504 and the module's defining contract — #2505's runner is built around it (mirrors the dispatcher's never-raises posture) and will not wrap the call, so an escaping `OverflowError` crashes check evaluation for that tenant rather than yielding `unknown`. The payload is reachable: `json.loads` maps arbitrary-precision JSON integer literals to Python `int`, so an op-result array like `{"xs": [<309-digit integer>, 1.5]}` under a `sum` aggregate triggers it. The fix commit states "a huge int can't itself raise OverflowError" — true for the two `math.isfinite` call sites it guards, but not for the `sum()` on this line. `test_evaluator_never_raises_or_emits_skip` passes only because its payload sweep contains no such array, so the AC test gives false confidence.

Reproduced on head `cc4d4370` (Python 3.12.3): evaluating a `sum` aggregate over `{"xs": [10**400, 1.5]}` raises `OverflowError` instead of returning `unknown`.

**Required change:** Guard the summation the same way `_parse_timestamp` already guards `datetime.fromtimestamp` (try/except → `unknown`):

```python
if aggregate == "sum":
    try:
        total = sum(values)
    except OverflowError:
        return None, "aggregate 'sum' overflows the float range"
    if isinstance(total, float) and not math.isfinite(total):
        return None, f"aggregate 'sum' is non-finite ({total!r})"
    return total, None
```

Add a regression test: a `sum` over `[10**400, 1.5]` yields `unknown` (extending `test_evaluator_never_raises_or_emits_skip`'s sweep to include a huge-int + float array closes the false-green gap).

damir-topic (commented):




**Anchor task:** #2504 — bounded typed assertion evaluator (pure select→compare)
**Surface area:** backend (pure module)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `cc4d4370`
**AI reviewer (CodeRabbit):** re-triggered 17:09Z, not received within 15 min (proceeded per skill Hard rule 17); iteration-1 CodeRabbit findings cross-checked below

Re-review of the fix push `cc4d4370` addressing iteration-1 findings B1 and M1. Scope: confirm resolution + no regressions, not a full re-derivation.

### Resolution of iteration-1 findings
- ✅ **B1 (non-finite observed value read as `ok`) — RESOLVED.** `_compare_threshold` now routes `NaN`/`±Inf` to `unknown` (evaluate.py L243–244); `_aggregate` rejects non-finite elements for sum/max/min (L188–189) and a non-finite `sum` result (L192–193); `ThresholdCompare._check_bounds` rejects non-finite `degraded`/`critical` bounds at parse (assertions.py L186–187 → `ValidationError`/422). The `isinstance(x, float)` guard correctly avoids `math.isfinite(huge_int)` raising. Verified: `NaN`/`+Inf`/`-Inf` observed → `unknown`; `json.loads('{"v": NaN}')` → `unknown`; non-finite aggregate element/result → `unknown`; non-finite bound (incl. `json.loads(... Infinity)`) → `ValidationError`. Well-tested (7 new parametrized cases).
- ✅ **M1 (spec models accept unknown fields) — RESOLVED.** `extra="forbid"` added to all seven frozen spec models (`SelectSpec`, `ThresholdCompare`, `EqualsCompare`, `InCompare`, `BoolCompare`, `FreshnessCompare`, `AssertionSpec`). Verified: a typo'd field on each of the seven raises `ValidationError`. Covered by `test_spec_models_forbid_unknown_fields` (6 members) + `test_assertion_spec_forbids_unknown_top_level_field`.
- ➖ **CodeRabbit RFC3339-strictness — left as DEFER (non-blocking), correctly.** Every form `datetime.fromisoformat` accepts denotes the same instant and the offset requirement is still enforced, so the freshness verdict is unchanged. Not re-raised.

### New blocker surfaced by the fix's own huge-int reasoning
- 🛑 **B2 — `sum` over a huge `int` + a `float` raises `OverflowError`, violating the never-raises contract.** See inline on evaluate.py. The fix guarded the `math.isfinite` call sites against huge ints but not the `sum()` itself: `sum([10**400, 1.5])` raises `OverflowError: int too large to convert to float`, which escapes `evaluate_assertion`. "Never raises on payload data" is an explicit acceptance criterion of #2504; the payload is reachable via `json.loads` (arbitrary-precision int literal + a float under a `sum` aggregate). Reproduced on `cc4d4370`. One-line fix (try/except → `unknown`, mirroring the existing `datetime.fromtimestamp` guard) + a regression test.

### Acceptance criteria
- [❌] **Never raises on payload data** — `sum` over `[huge_int, float]` raises `OverflowError` (B2). The AC test `test_evaluator_never_raises_or_emits_skip` passes but does not exercise this payload.
- [x] Non-finite observed value → `unknown` (B1) — verified
- [x] `extra="forbid"` on the seven spec models (M1) — verified
- [x] `uv run pytest tests/test_checks_assertions.py -q` — 108 passed
- [x] Purity preserved — still no I/O imports, no `async def`; fix adds only `import math` (stdlib) to evaluate.py; models unchanged in shape

### Integration verification
- [x] `ruff check` (4 changed files) — clean (8 tree-wide errors are in an unrelated dev-seed script, not PR files, pre-existing)
- [x] `ruff format --check` (changed files) — clean
- [x] `mypy src/meho_backplane/checks` — Success, no issues
- [x] `pytest tests/test_checks_assertions.py` — 108 passed
- [x] Reproduction probe (huge-int / float-mix / NaN / bound-rejection) — B2 confirmed; all B1/M1 paths confirmed
- [x] Blast radius — new top-level `meho_backplane.checks` package has no importers yet (siblings #2503/#2505/#2506 unlanded); regression surface is the package's own test

- 🛑 B2 — `sum` over huge-int + float raises `OverflowError`; required change: wrap `sum(values)` in `try/except OverflowError` → `unknown`, add a `[10**400, 1.5]` regression case.

- Iteration-1 CodeRabbit findings: extra="forbid" (→ M1, resolved), non-finite handling (→ B1, resolved), RFC3339 strictness (DEFER, not re-raised).
- Re-review of `cc4d4370` did not land within the 15-min window; B2 is an independent finding CodeRabbit did not raise.

### Docs
- [n/a] No `docs/codebase/` doc for this new package; issue #2504 scopes docs to the CHANGELOG entry only (present).

- B1/M1 fixes are minimal, correctly targeted, and each landed with focused parametrized tests (non-finite value/element/result/bound; per-model typo rejection). The `isinstance(x, float)` guard around `math.isfinite` shows the author was reasoning about the huge-int hazard — B2 is the one remaining sum() site that reasoning didn't reach.

damir-topic:
@coderabbitai review

damir-topic (commented):
## /auto-review-pr summary (iteration 3 — final)




**Verdict: APPROVE**

> GitHub 422s a formal APPROVE/REQUEST_CHANGES state on a self-authored PR
> (author == authenticated reviewer), so this is posted as a `COMMENTED`
> review; the binding verdict is **APPROVE**, carried here and in the result
> file. Mirrors iterations 1 and 2.

**Anchor task:** #2504 — bounded typed assertion evaluator (pure select → compare)
**Surface area:** backend
**Head SHA reviewed:** `32763baf`
**AI reviewer (CodeRabbit):** no new review for the fix commits — its manual
re-trigger (comment 4994826778) returned *"Review finished … does not
re-review already reviewed commits"*. Its only substantive review (iter-1,
13:03:46Z) raised extra=forbid (→ M1, resolved), non-finite handling (→ B1,
resolved) and RFC3339 strictness (correctly deferred — every accepted form
denotes the same instant). It never raised B2.

### Scope of this iteration
Focused re-review of the **B2 fix** (`32763baf`) plus a regression guard that
B1/M1 stayed resolved, and an adversarial hunt for any *other* reachable raise
path in `evaluate_assertion` (the crux, since iter-1 missed B2).

### B2 — RESOLVED ✅
`_aggregate` now wraps `total = sum(values)` in `try/except OverflowError`
→ returns `(None, "aggregate 'sum' overflows the float range")` (i.e.
`unknown`). Verified:
- **Fix behaves**: `sum([10**400, 1.5])` → `unknown`, reason "aggregate 'sum'
  overflows the float range" — no raise. `sum([1.5e308, 1.5e308])` → `unknown`
  "is non-finite (inf)" (the `math.isfinite(total)` arm, unchanged).
- **Scoping is tight**: the `try` wraps *only* `sum(values)` and catches *only*
  `OverflowError`. Elements are pre-validated int / finite-float in the loop
  above, so `sum` can raise nothing else — no unrelated exception is swallowed.
- **Genuine regression guard**: on the pre-fix path `sum([10**400, 1.5])` raises
  `OverflowError: int too large to convert to float`; the new test
  `test_sum_aggregate_bignum_plus_float_overflow_is_unknown` asserts `unknown`
  + "overflow" in the reason, and the `test_evaluator_never_raises_or_emits_skip`
  sweep gained a `{"disks": [{"free_pct": 10**400}, {"free_pct": 1.5}]}` payload.

### B1 + M1 — still resolved ✅ (regression guard)
- Non-finite observed (`json.loads('{"v": NaN}')`) → `unknown` ("requires a
  finite number").
- `extra="forbid"` on all seven spec models — a typo'd field → `ValidationError`.

### Adversarial never-raises hunt (the crux)
Brute-forced **81,270 combinations** (86 hostile payloads × 945 valid specs) —
bignum / -bignum / NaN / ±inf / 1.5e308 scalars and arrays, epoch edges, naive /
garbage / dict / list / bool timestamps, year-9999 & year-0001 datetimes, huge
`max_age_seconds`, huge-int equality operands — through every comparator:
- **RAISES = 0.** No reachable raise on payload data.
- **NON_SCALAR outcome values = 0.** `AssertionOutcome.value` is always a JSON
  scalar (dict/list selections are caught as `unknown` before construction).
- **Freshness** (`_compare_freshness` / `_parse_timestamp`) specifically probed:
  every malformed / absent / wrong-type / timezone-naive / huge / negative /
  NaN / inf timestamp → `unknown`; year-9999 & year-0001 aware datetimes subtract
  cleanly (bounded to `datetime.min…max`, `total_seconds()` in float range); a
  future timestamp → `ok`. `fromtimestamp` / `fromisoformat` raises are all
  caught (`ValueError`/`OverflowError`/`OSError`).
- **max / min** over mixed bignum + float → safe (exact int↔float comparison, no
  arithmetic); non-finite floats are rejected before `max`/`min`; `math.isfinite`
  is only ever called behind an `isinstance(_, float)` guard, so a huge int never
  reaches it.
- **equals / in** (`_strict_eq`) use `==` (not hashing) over scalar operands —
  no raise on exotic/huge values.
- Spec-time hardening confirmed (out of the payload contract but clean): a
  huge-int `critical` bound → Pydantic `ValidationError` (422-shaped), not an
  uncaught `OverflowError`.

### Integration verification (ran inside the review worktree at `32763baf`)
- [x] `ruff check` (checks pkg + test) — All checks passed
- [x] `ruff format --check` — 4 files already formatted
- [x] `mypy src/meho_backplane/checks/` — Success, no issues (3 source files)
- [x] `pytest tests/test_checks_assertions.py` — **109 passed** (was 108; +1 B2 test)
- [x] adversarial probe — 81,270 combos, 0 raises, 0 non-scalar values
- [x] red-before check — pre-fix `sum([10**400, 1.5])` raises `OverflowError`

### Purity
Intact: no I/O, no async; the fix adds no imports (`math` already present).
`evaluate_assertion` remains a pure sync `(spec, payload, now) → outcome`.

### Predicted vs. actual diff
- ✅ `checks/evaluate.py` (`_aggregate` sum guard), `tests/test_checks_assertions.py`
  (+1 B2 test, +sweep payload), `CHANGELOG.md` — all expected and correct.
- No unexpected or missing files. (The large `git diff origin/main` is only main
  advancing on a parallel topology workstream — the PR's own diff is the five
  checks-package files; the branch is CONFLICTING and the orchestrator will union
  merge-main-in at merge, which is not a review concern.)

### Docs
- n/a — `docs/codebase/` has no checks doc yet; #2505/#2506 land the DB/rollup
  siblings that would anchor one. Unchanged from iter-1/iter-2 assessment.

### What worked well (review-body only — not posted inline)
- The fix is minimal and correctly scoped: it guards exactly the one arithmetic
  site that can overflow (int→float coercion in `sum`), mirroring the existing
  `fromtimestamp` guard, and leaves the already-correct `isfinite(total)` arm for
  the float-saturation case.
- The never-raises contract is now backed by a real degenerate-payload sweep that
  includes the bignum+float case that slipped past iterations 1–2.

PR #2573, opened 2026-07-17
chore(backend): migrate to python 3.14

ddzafic:
## Description

Migrate the backplane from Python 3.12 to 3.14. **Supersedes #2527**, the
Dependabot PR that bumped only the Dockerfile tag and could not build — the
image build fails deterministically without the changes here.

Two blockers, neither visible to Dependabot:

1. **`uv.lock` pinned `py-rust-stemmers 0.1.5`** (uploaded 2025-02-19), whose
   newest wheels are cp313. On 3.14 uv falls back to a source build via
   maturin/cargo, and `python:3.14-slim` has no Rust or C compiler, so
   `uv sync --locked` exits 1 on ``linker `cc` not found``. Bumped to 0.1.8,
   which ships cp314 wheels and is already inside `fastembed`'s
   `>=0.1.0,<0.2.0` range. Dependabot can't fix this — it bumps only the
   package it targets, and `py-rust-stemmers` is a transitive no direct
   constraint names.
2. **The Dockerfile hardcoded `/app/.venv/lib/python3.12/site-packages`** for
   the Tailwind build step, baking the interpreter minor version into a path.
   Now resolved at build time from `sysconfig` purelib, which fixes 3.14 and
   stops the next base-image bump breaking the same way.

### Operator-visible TLS behavior change (the load-bearing part)

CPython 3.13 added `VERIFY_X509_STRICT` (and `VERIFY_X509_PARTIAL_CHAIN`) to
`ssl.create_default_context()`'s default `verify_flags`
([python/cpython#107361](https://github.com/python/cpython/issues/107361)), so
the 3.12 → 3.14 jump crosses it. Strict mode enforces RFC 5280 to the letter —
notably requiring an Authority Key Identifier on non-self-signed certs — which
is exactly the conformance the self-signed / internal-CA appliance certs
`_build_ca_pinned_ssl_context` exists to trust tend to lack. Left on, pinned
targets that verify fine today would start failing with an opaque
`Missing Authority Key Identifier`.

This was caught by an existing test (`test_ca_pinned_context_trusts_chaining_cert`),
which asserts the product requirement and is a true positive — so the
production context is fixed rather than the test. `STRICT` is cleared; chain
and hostname verification are untouched, restoring pre-3.13 pinning semantics
rather than weakening them. `PARTIAL_CHAIN` is deliberately kept: terminating
path validation at the pinned cert is the intent of a pin.

Verified empirically that the sibling `validate_ca_pin_pem`
(`targets/schemas.py`) does **not** regress — it only `load_verify_locations()`,
which `VERIFY_X509_STRICT` does not gate.

### ruff `target-version` pinned to py313 (deliberate)

Set explicitly *below* the `requires-python` floor, because ruff otherwise
infers `py314` from `requires-python`
([docs](https://docs.astral.sh/ruff/settings/#target-version)). At py314 the
formatter applies PEP 758 and rewrites every `except (A, B):` into
`except A, B:` across 34 files — valid only on 3.14+
([PEP 758](https://peps.python.org/pep-0758/)), and a meaning-inverting
lookalike for Python 2's `except E, name:`. Nothing here needs 3.14-only
syntax; that belongs in a separate style decision, not an interpreter bump.

## Related issue

Supersedes Dependabot PR #2527 (close on merge — closing keywords don't apply
across PRs). No project-board task; Dependabot-origin.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] CI / build / chore

## Verification

- Image builds for `linux/amd64`; runs 3.14.6 with `py-rust-stemmers 0.1.8`
  from a wheel; `tailwind.css` lands at the resolved `python3.14` path; app
  imports clean.
- `ruff check` + `ruff format --check` clean; `mypy src/` strict matches the
  `main` baseline (only the Linux-only `MSG_ERRQUEUE` constant, which resolves
  on CI).
- Full unit suite on 3.14: 10619 passed, 66 skipped. The 3 failures are
  pre-existing macOS-only environment artifacts (raw-socket ICMP, DNS
  resolver) that also fail on `main` under 3.12 and don't apply on Linux CI.
- `uv.lock` diff is large but inert: identical 179-package set, only
  `py-rust-stemmers` changes version. The ~1200 removed lines are the
  3.12/3.13 resolution-marker branches the universal lock no longer needs once
  the floor is 3.14.

## Outstanding from self-review (docs drift — non-blocking, fixable on this branch before merge)

- **Dockerfile internal comments** still reference `3.12-slim` (7 spots incl. a
  now-false `Source: … Python 3.12.x` provenance line) — the FROM lines moved
  to 3.14 but the surrounding comments didn't.
- **`backend/README.md`** and **`ui/.../vendor/VENDOR.md`** still say
  `python:3.12-slim` — docs this change invalidates.
- **`CHANGELOG.md`** `[Unreleased]` has no entry for this migration; the
  operator-visible TLS change in particular belongs under `Changed` / `Security`.
- Follow-up (not in scope here): `pyopenssl` is now retireable — its dependency
  rationale (`get_unverified_chain` being 3.13+) no longer binds on a 3.14
  floor; and add a `dependabot.yml` ignore rule to return the docker block to
  digest-only refreshes.

## Checklist

- [x] Conventional Commits format on every commit
- [x] Every commit has a `Signed-off-by:` line (DCO; `git commit -s`)
- [x] Tests added/updated where applicable (existing TLS test now guards the 3.14 behavior)
- [ ] Documentation updated where applicable (see *Outstanding* above)
- [ ] CI green (first honest CI run — Dependabot PRs skip the Python jobs; see #2527 context)

## Summary by CodeRabbit

* **New Features**
  * Upgraded the application’s supported Python runtime to version 3.14.
  * Updated container environments to use Python 3.14 for more consistent builds and deployments.
  * Preserved secure certificate-chain and hostname verification while maintaining compatibility with newer Python and OpenSSL behavior.

* **Documentation**
  * Updated backend deployment documentation and image refresh guidance for Python 3.14.

* **Tests**
  * Updated compatibility checks and test configuration for the new Python runtime.


backend/pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "meho-backplane"
 version = "0.1.0-dev"
 description = "MEHO governance-layer backplane (HTTP server, observability, audit, federation)"
 readme = "README.md"
-requires-python = ">=3.12"
+requires-python = ">=3.14"
```
ddzafic, line 6: **🛑 Blocker** — raising the floor to `>=3.14` breaks a CI workflow still pinned to Python 3.13

**What:** `.github/workflows/dependency-license-check.yml:75` pins `actions/setup-python` to `"3.13"`, then runs `pip install -e ./backend` (line 89). With this bump, pip refuses: `ERROR: Package 'meho-backplane' requires a different Python: 3.13.14 not in '>=3.14'`.
**Why it matters:** `Python License Check` is a **required** status check — the PR can't merge, and it's already failing on this PR. `ci.yml` / `eval-gate.yml` are unaffected (they use `setup-uv` + `uv sync`, which resolves 3.14 from `requires-python`); this workflow is the only one on `setup-python` + bare `pip install`.
**Suggested fix:** in `dependency-license-check.yml`, bump `python-version: "3.13"` → `"3.14"` (setup-python v6 → 3.14 already works — `migration-compat.yml` uses it and passes).

ddzafic (commented):
## /review-pr-slim summary

**Anchor:** supersedes Dependabot PR #2527 (no project-board task; Dependabot-origin).
**Surface area:** devops + python (Dockerfile, CI workflow, pyproject, ssl code, lockfile).
**Best-practices packs audited:** none loadable — only the *slim* skills are installed in this repo; `implement-issue/`/`explore-and-refactor/` packs don't exist. Audited against standard Dockerfile/Python practice instead of claiming a pack run.
**Reviewer note:** this is a **self-review** — I authored the PR. GitHub blocks author self-`REQUEST_CHANGES`, so this is posted as a `COMMENT`; the effective verdict is below.

**Verdict: REQUEST_CHANGES** — one CI-proven blocker (a required check fails).

### 🛑 Blocker (1)

**`Python License Check` fails — `requires-python = ">=3.14"` breaks a workflow still pinned to Python 3.13.**

`.github/workflows/dependency-license-check.yml:75` pins `actions/setup-python` to `"3.13"`, then line 89 runs `pip install -e ./backend`. With the floor raised to `>=3.14`, pip refuses:

```
ERROR: Package 'meho-backplane' requires a different Python: 3.13.14 not in '>=3.14'
```

`Python License Check` is a **required** status check, so the PR cannot merge as-is. This is the migration's own blast radius — the `requires-python` bump (this PR) invalidated an unchanged workflow.

**Blast radius (verified):** the only affected workflow. `ci.yml` and `eval-gate.yml` install the backend via `astral-sh/setup-uv` + `uv sync --locked`, which resolves 3.14 from `requires-python`; `dependency-license-check.yml` is the sole workflow using `setup-python` + bare `pip install -e`.

**Fix:** bump `dependency-license-check.yml:75` `"3.13"` → `"3.14"` (setup-python v6 → 3.14 already works — `migration-compat.yml` uses it and passes).

### ⚠️ Majors (3) — documentation drift (already itemized in the PR body's "Outstanding" section)

1. **Dockerfile self-contradicts** — 7 comment refs to `3.12-slim` remain (incl. a now-false `Source: … Python 3.12.x` provenance line at ~26–27) while the `FROM` lines moved to 3.14.
2. **`backend/README.md:61` + `ui/.../vendor/VENDOR.md:158`** still say `python:3.12-slim` — docs this PR invalidates.
3. **`CHANGELOG.md` `[Unreleased]`** has no entry; the operator-visible TLS behavior change belongs under `Changed`/`Security` (the CHANGELOG is the authoritative Release-notes source).

### 💡 Minors (2)

- `.github/dependabot.yml:59` and `http.py:122` carry stale `3.12` references (illustrative / the `check_hostname`-before-`CERT_NONE` ordering note still holds, just dated).

### Verified clean (non-findings)

- **`validate_ca_pin_pem` (`targets/schemas.py:99`) does not regress** — tested empirically on 3.14: a no-AKI CA loads identically with `VERIFY_X509_STRICT` on/off. STRICT gates handshake verification, not `load_verify_locations()`. The STRICT fix was correctly scoped to `_build_ca_pinned_ssl_context` only.
- `_build_insecure_ssl_context` (CERT_NONE) and `net/tls.py` inspection (unverified pyOpenSSL) — STRICT irrelevant, no regression.
- No `sys.version_info` branches in `backend/src`.
- `uv.lock` diff inert: identical 179-package set, only `py-rust-stemmers` moved; the ~1200 removed lines are collapsed 3.12/3.13 resolution markers.

### Framework research (cited)

- `VERIFY_X509_STRICT` added to `create_default_context` defaults in **3.13** — [python/cpython#107361](https://github.com/python/cpython/issues/107361).
- Ruff infers `target-version` from `requires-python`; explicit wins — [ruff docs](https://docs.astral.sh/ruff/settings/#target-version). The py313 pin + its comment are accurate.
- `except A, B:` (no `as`) is **3.14-only** — [PEP 758](https://peps.python.org/pep-0758/). Pinning ruff to py313 to avoid that 34-file churn is justified.

### Integration verification

- [x] `uv sync --locked` (3.14) — clean, in review worktree on head SHA
- [x] `ruff check` / `ruff format --check` (3.14) — clean, review worktree
- [x] `mypy src/` strict — matches `main` baseline (only Linux-only `MSG_ERRQUEUE`, resolves on CI)
- [x] Full unit suite (3.14) — 10619 passed, 66 skipped (byte-identical SHA); 3 fails are pre-existing macOS-only artifacts, not present on Linux
- [x] Docker image build + runtime (linux/amd64) — 3.14.6, wheel install, app imports, tailwind.css at resolved path
- [x] **Real Linux CI:** DCO / CodeQL / Semgrep SAST / Helm / **Backward-compat migration guard (on 3.14)** — pass; **Python License Check — FAIL (the blocker)**; ruff+mypy+pytest, integration testcontainers, image build — still running at review time
- CodeRabbit: **no actionable comments** (file-diff-scoped; would not catch the cross-file license-check blocker)

### Docs

- [x] `docs/codebase/backend.md` — updated (base-image section; also corrected two pre-existing inaccuracies)
- [ ] `README.md` / `VENDOR.md` / `CHANGELOG.md` — not updated (⚠️ majors above)

### What was done well

- 👏 TLS regression caught by an existing test and fixed at the production context (not the fixture), with the fix scoped to the handshake path only — verified the sibling validator doesn't regress.
- 👏 The `sysconfig.purelib` Dockerfile fix is version-agnostic rather than a 3.12→3.14 find-replace, so the next base bump won't break the same way.
- 👏 ruff `target-version` pinned below the floor with a documented rationale to keep PEP 758 churn out of an interpreter bump.

damir-topic:
Pushed a merge of current main onto this branch plus the two fixes that were keeping it red. Authorship preserved via `Co-Authored-By` — the design decisions here are yours, this is just unblocking.

**What was wrong, and what changed:**

1. **`backend/uv.lock` was the only merge conflict.** Main moved a lot today — nine dependency bumps landed in an open-PR sweep. Resolved by taking main's lock and re-running `uv lock`, so the 3.14 resolution is regenerated against current pins rather than hand-merged. `pyproject.toml` auto-merged cleanly and correctly: your `requires-python = ">=3.14"` alongside main's `fastapi>=0.140.0,<0.141`.

2. **`py-rust-stemmers` 0.1.5 → 0.1.8 — this is what actually broke the image build.** 0.1.5 publishes no cp314 wheel, so on `python:3.14-slim` uv fell back to the sdist, tried to build the Rust extension, and died:

   ```
   × Failed to build `py-rust-stemmers==0.1.5`
     error: linker `cc` not found
   ```

   The slim image has no toolchain, by design. 0.1.7+ ship cp314 wheels — the lock now carries `manylinux_2_17_x86_64` and `manylinux_2_28_x86_64` among them, so the image installs a prebuilt wheel. Worth knowing: it is an **unconstrained transitive of fastembed**, so a plain `uv lock` happily kept 0.1.5 (it satisfies the requirement); it needed an explicit `uv lock --upgrade-package py-rust-stemmers`. That is also why the same failure hits #2527, which only bumps the Dockerfile.

3. **`dependency-license-check.yml` still pinned `python-version: "3.13"`.** That job runs a real `pip install -e ./backend`, so raising the floor to 3.14 failed it outright with `Package 'meho-backplane' requires a different Python: 3.13.14 not in '>=3.14'`. Bumped to 3.14, with a comment tying the pin to the `requires-python` floor so the next bump does not repeat this. You had already handled `migration-compat.yml`.

**Not changed:** your `target-version = "py313"` decision on ruff. The reasoning in the comment — that py314 makes the formatter apply PEP 758 and rewrite `except (A, B):` into a form that reads like Python 2's meaning-inverting `except E, name:` — is sound, and keeping the runtime bump separate from a syntax-style decision is the right call.

**Once CI is green this still needs a human decision, not just a merge button:** it changes the runtime interpreter for the whole backend, and #2527 (Dockerfile `3.12-slim` → `3.14-slim`) overlaps with it. My read is that #2527 is subsumed by this PR and should close once this lands, rather than racing it — but that is your call as the author.

Issue #2172, opened 2026-07-06
Initiative: SSH cluster-node OS-lifecycle write ops — governed RKE2/node maintenance


damir-topic:
Parent goal: #221

## Why

Field evidence (evoila-bosnia/claude-rdc-hetzner-dc#615): a P0 RKE2 join-token rotation ran almost entirely on the local SSH-wrapper fallback because the `kubernetes` connector is Kubernetes-API-only (~25 ops, verified on `main`) and no node/linux/rke2-ssh connector exists. Only the health gates (`k8s.node.list` / `k8s.pod.list`) rode the backplane. This Initiative adds a **governed, typed, path-bounded SSH node-OS-lifecycle op surface** — the same approval-gated write-op mold as holodeck-ssh (G3.18 #2145) — so cluster-node maintenance runs through one identity + audit plane instead of an untracked local wrapper.

## Depends-on

- **#2155** — the SSH adapter mis-consumes `secret_ref` (the "bind9 anti-shape"). **RESOLVED — merged 2026-07-07**; the adapter now resolves creds via `load_vault_secret_data` (either/or key-or-password). The Initiative is unblocked.

## Scope (proposed ops — typed/narrow/path-bounded, NOT a generic shell op)

- `rke2.token.rotate` — old/new server-side, never echoed (approval-gated)
- `node.service.restart` — unit allow-list: `rke2-server`, `rke2-agent` (approval-gated)
- `node.config.update` — path-bounded to `/etc/rancher/rke2/*`, key-level edit + mode (0600) enforcement (approval-gated)
- `rke2.etcd-snapshot.save` — safe-tier
- read-only posture tier first: file modes + token-key presence (values redacted)

## Design decision needed (before the write-op Tasks are filed)

**RATIFIED 2026-07-13.** Op surface + tiers below (grounded against bind9/holodeck/proxmox molds + RKE2 mechanics). Six sub-decisions settled: product-prefixed op_ids; etcd-snapshot=safe/non-gated; multi-node rotation=operator runbook (no composite op); config.update=backplane-owned key merge; new token written to Vault + pointer returned; OLD token read on-disk from `/var/lib/rancher/rke2/server/token`. **Load-bearing:** the audit `raw_payload` stores the raw result, so `token.rotate`'s handler must never return the token. Child Tasks #2429/#2430/#2431 filed.

## Out of scope

- A generic shell/exec op — deliberately excluded; every op is typed + path/unit-bounded.
[...]

Issue #2494, opened 2026-07-14
Initiative: G0.32 v0.21/v0.22 closed-loop dogfood hardening — MCP wire contract, governance, connector-session & deploy-docs findings


damir-topic:
Parent goal: #221

## Summary

Per-cycle hardening Initiative for the **v0.21.0/v0.22.0 consumer probe** (series G0.8 → G0.31 #2364). An agent-consumer walked the MCP tool surface and live connectors on 2026-07-13/14 and filed 20 findings back (#2471–#2489); the maintainer filed two more from the same period (#2398 TLS-SNI, #2468 deploy docs), and an RDC operator dogfooding the shipped v0.22.0 CLI filed one more on 2026-07-17 (#2577, targets-import list envelope). Every finding was re-diagnosed against the shipped v0.22.0 source (tag `1be8cac3`) and current `main` before adoption; corrections are recorded below so wrong premises are not re-filed. Sibling Initiative #2495 carries the operator-console (`/ui/*`) findings from the same walkthrough.

## Confirmed refuted / duplicates (no task — close-candidates, left open for the reporter)

- **#2476** (dual `slug` + `template_slug` on template responses) — the response mirror is the **canonical, deliberately retained** #1612 behaviour (`_mirror_template_slug`, `mcp/tools/runbooks.py:212-225`; pinned in `docs/codebase/mcp.md`); #1625 removed *input/tool-name* aliases only. **Folded into the #2471 family-wide pass** (2026-07-18): the only actionable residue is a one-line `template_slug` inputSchema description clarification acknowledging the retained `slug` mirror — carried in #2471; #2476 closes when #2471 merges.
- **#2477 / #2478 / #2482** — accurate observations, folded into the #2471 family-wide pass (same pin-test file, same docs sections, one CHANGELOG header — separate PRs would churn the same shared files).

## Root-cause corrections (recorded so they are not re-filed under a wrong premise)

| Reporter framing | Ground truth at v0.22.0 / main HEAD |
|---|---|
[...]

PR #2605, opened 2026-07-18
feat(operations): scrub credential_read dispatch responses by default (#2467)

### LINKED ISSUE #2467: Task: default key-name scrub for credential_read dispatch responses — params.reveal_secret=true opt-in, stamped on the audit row
Parent goal: #221
Parent initiative: #2494

## Observed on

Chart `0.21.0` (git `bdf56146`).

## Summary

`call_operation` with `connector_id=vault-1.x, op_id=vault.kv.read, target=rdc-vault, params={path: "meho/holodeck/holorouter-hd5fs"}` returned the raw credential value (`{"password": "...", "username": "root"}`) in the response's `content[0].text` field. That response text was ingested by the agent session that made the call, and consequently ended up in the agent's conversation transcript / model API context.

MEHO's audit ledger correctly tagged the operation as `op_class: credential_read` (verified via `query_audit` after the incident — the audit row is on the `meho.broadcast.recent` stream with `op_class: credential_read`). So the substrate *knows* this is a credential read. The design question: **should `credential_read`-class ops via the MCP path surface the raw value to the caller by default, or should they redact / summarize?**

## The trade-off (as I see it)

**Arguments for status-quo (raw value):**
- `vault.kv.read` is a documented Vault read; hiding the value would break the tool's contract.
- Agents genuinely need to *pass* credential values into subsequent tool calls (e.g. authenticate a downstream API). Redaction would break those flows.
- The audit trail already tags the class; consumers who care can filter downstream.

**Arguments for redaction-by-default:**
- The most common consumer of `call_operation` is a human-agent (Claude / Cursor / etc.), and a `credential_read` response's raw value lands directly in that agent's conversation transcript, which is sent to the model API. That's a real-world exposure surface that no amount of downstream audit filtering can retract.
- CLAUDE.md-style consumer discipline explicitly says "never paste a secret into a chat with Claude" — but here the *tool result* pastes it *for* the agent, so the agent can't avoid it by discipline alone.
- Programmatic consumers (non-agent) that legitimately need the value can opt into a `reveal=true` flag or use a different op / connector.

## Concrete proposal

Add an opt-out redaction to `credential_read`-class ops when the calling principal is `agent_kind: human_agent`:

- Default: `credential_read` responses include a `redacted: true` marker and a `value_sha256` or `value_length` summary in place of the raw value.
- Explicit opt-in: `params.reveal_secret = true` (or a per-op flag) returns the raw value AND is stamped in the audit row as `reveal_secret: true` so it's queryable.
- Programmatic consumers set `reveal_secret=true` when they know they need to pipe the value onward; agents that don't know they need it get redaction by default.

Alternative shape: a broker pattern where MEHO stores a short-lived (30s) reference token in the response, and the caller can only get the raw value by presenting the reference token to a separate reveal endpoint — deliberately breaking the "value ends up in agent transcript" pattern.

## Real-world incident

Filed a rotation ticket downstream ([claude-rdc-hetzner-dc issue](https://github.com/evoila-bosnia/claude-rdc-hetzner-dc/issues/)) because this exact call, made as a test during the MCP surface walk, dumped a live HoloRouter root password into my agent session. Consumer-side discipline says "don't call `vault.kv.read` from an agent session against credential paths" — but the tool's *shape* invites that call (returns a JSON payload with `data.password`), and the leak surface is severe (agent transcript → model API).

This isn't a bug in `vault.kv.read` per se — it's doing what it's documented to do. It's a design question at the MCP boundary about whether `credential_read`-class ops should default to summary-mode for agent callers.

## Related

- The `op_class: credential_read` machinery already exists (visible in `query_audit`, `meho.broadcast.recent`, retrieval scores, etc.) — the classification is there; the redaction hook isn't.
- Consumer-side workaround (documented downstream): agents that don't need the raw value should call `vault.kv.versions` (metadata-only) instead of `vault.kv.read`. But that's a discipline-only mitigation, not a substrate guarantee.

---

## Triage grounding note (2026-07-14 v0.21/v0.22 batch triage)

Reporter's symptom and trade-off framing are accurate (live root password reached an agent transcript; rotation ticket filed), but 'the classification is there; the redaction hook isn't' is doubly wrong: the connector-boundary redaction hook runs on every dispatch response (dispatcher.py:673, #1071) and a key-name-aware scrub already exists (_scrub_secret_named_keys, broadcast/events.py:673) — the boundary policy is label-in-string-leaf regex that structurally cannot fire on {"password": "..."} dicts, and the key-name scrub runs only on broadcast params, never the caller-bound response. The proposed 'agent_kind: human_agent' discriminator doesn't exist (kinds: user/service/agent) and would have missed this incident, since the caller was a service-account token driven by an agent. The sibling credential_read op sddc.credential.list already ships exactly the scrub-by-default treatment; secret.move is the existing broker alternative. Correctly filed as a design question — the op's documented contract (return the value) is what makes the default a product call.

Evidence:
- `backend/src/meho_backplane/operations/dispatcher.py:673` — Connector-boundary redaction middleware runs on every dispatch response — the 'missing hook' exists (G11.4-T2 #1071)
- `backend/src/meho_backplane/redaction/engine.py:206` — _walk_and_apply matches rules only inside string leaves; dict keys feed the manifest path, never matching
- `backend/src/meho_backplane/redaction/patterns.py:145` — _API_KEY requires the label AND [=:] inside one string; {"password": "<value>"} structurally never fires
- `backend/src/meho_backplane/redaction/policies/default.yaml:47` — Default policy's only password-covering rule is the labelled api_key pattern
- `backend/src/meho_backplane/redaction/resolver.py:33` — No production policy-authoring path; zero register_policy callers outside tests
- `backend/src/meho_backplane/broadcast/events.py:673` — _scrub_secret_named_keys documents this exact structured-payload gap; applied only to broadcast params (_audit.py:557), not the caller-bound response
- `backend/src/meho_backplane/connectors/vault/ops.py:391` — vault_kv_read returns raw secret_data; ops.py:270-298 documents returning the value as the op's purpose
- `backend/src/meho_backplane/broadcast/events.py:84` — _CREDENTIAL_READ_OPS = {vault.kv.read, vault.kv.list, sddc.credential.list} — the existing taxonomy key for a redaction gate

## Product decision (default adopted at filing — veto before implementation)

Question: should credential_read-class dispatch responses (allowlist broadcast/events.py:84-95 — in practice vault.kv.read; vault.kv.list and sddc.credential.list are already value-free) return raw secret values to call_operation callers by default, or return a key-name-scrubbed summary (key names + value_length/value_sha256) with raw values gated behind an explicit params.reveal_secret=true that is stamped queryably on the audit row and broadcast? Recommended default: scrub-by-default keyed on classify_op(op_id)=="credential_read" at the dispatcher's existing redaction seam (dispatcher.py:673), reusing the _scrub_secret_named_keys walk / the SDDC _redact_secrets treatment — NOT keyed on principal kind (the incident caller was the backplane service-account token driven by Claude, so caller-kind gating misses the actual exposure; no human_agent kind exists). reveal_secret=true returns the raw value and stamps the audit row so reveals are queryable. Approved human flows keep working by opting in at their two known call sites: the /ui Vault console BFF (ui/routes/vault/routes.py, reveal-on-click stays) and CLI meho secret read (#2146, pipe-only). secret.move (#1577) remains the recommended no-transit path for piping a credential onward; vault.kv.versions remains the metadata-only alternative. Rejected alternative: the reporter's 30s reference-token broker endpoint — heavier new infrastructure duplicating secret.move, against the substrate-minimalism posture.

damir-topic:
## Summary
- **Default key-name scrub for `credential_read` dispatch responses.** A `credential_read` op (`vault.kv.read` + its `_CREDENTIAL_READ_OPS` siblings) returns the raw secret in its handler payload — its documented contract. The connector-boundary redaction engine matches only *labelled* secret shapes inside string leaves, so a `{"password": "<value>"}` dict value passed through verbatim into the `call_operation` caller's response (and, for an agent caller, its model-API transcript). The dispatcher now key-name scrubs the **caller-bound** response: secret-named values become `[REDACTED:param_name]`, non-secret siblings (`username`, `version`) survive.
- **`params.reveal_secret=true` opt-in, stamped on the audit row.** A caller that must pipe the raw value onward passes the reserved flag; the raw value is returned AND the choice is recorded queryably as `payload['reveal_secret']`. The flag is stripped before op-schema validation (`additionalProperties: false`) and before `params_hash`, so a scrubbed read and its reveal share one hash.
- **Audit `raw_payload` keeps the raw result** in both cases — only the transport response is scrubbed (mirrors the #2172 secret-handler posture). The scrub keys on the **op class, not the caller kind** (the incident caller was a service-account token driven by an agent, so principal-kind gating would have missed it).
- **Approved human flows opt in at their two known call sites:** the `/ui` Vault console read (reveal-on-click) and CLI `meho secret read` (pipe-only) both pass `reveal_secret=true`; their audit rows honestly record the reveal. `secret.move` stays the recommended no-transit path; `vault.kv.versions` is the metadata-only browse.

## Test plan
- [x] `ruff check` / `ruff format --check` — clean (all changed files)
- [x] `mypy` — clean (all changed source files)
- [x] `code-quality.py --diff` — 0 blockers (extracted `_apply_contextvar_lineage_mirrors` to keep `_build_audit_payload` under the size/complexity caps)
- [x] New `tests/test_credential_read_response_scrub.py` (10 cases): scrubbed-by-default, `reveal_secret=true` reveals + stamps audit, `raw_payload` keeps raw, reveal flag stripped from schema + hash (one shared `params_hash`), non-`credential_read` op untouched, `scrub_secret_named_values` unit contract — all pass
- [x] `test_operations_dispatcher.py`, `test_mcp_tools_operations.py`, `test_broadcast_credential_mint_dispatch.py`, `test_agent_audit_invocation.py`, `test_audit_work_ref.py`, `test_ui_vault.py`, broadcast/runbook lineage sweep — all green (refactored audit-lineage mirrors verified)
- [x] `go build ./...` + `go test ./internal/cmd/secret/...` + `go vet` — pass (CLI opt-in + updated assertion)
- [x] No OpenAPI drift (MCP tool description is JSON-RPC, not a FastAPI route; no REST model touched → no `snapshot-openapi` regen)

## Out of scope / notes
- `meho vault kv read` is **not** an opt-in site (per the product decision's two named sites) — its browse envelope now shows the scrubbed value; use `meho secret read` for the raw bytes. Flagged for maintainer awareness in case a `--reveal` flag is later wanted.
- Pre-existing (not caused here): `tests/test_connectors_vault.py::test_preflight_fail_soft_when_probe_raises` emits a teardown ERROR (test itself passes) on both this branch and pristine `main`.

Closes #2467

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,29 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Added — credential_read response scrub + reveal_secret opt-in (#2467)
+
+- `credential_read`-classified dispatch responses (`vault.kv.read` and
+  its `_CREDENTIAL_READ_OPS` siblings) are now key-name scrubbed before
+  they reach the `call_operation` caller: secret-named values in the
+  transport response are replaced with `[REDACTED:param_name]` while
+  non-secret siblings (`username`, `version`) survive. This closes the
+  gap where the connector-boundary redaction engine — which matches only
+  labelled secret shapes inside string leaves — passed a
+  `{"password": "<value>"}` dict through verbatim into an agent caller's
+  model-API transcript. A caller that must pipe the raw value onward
+  passes a reserved `params.reveal_secret=true`, which returns the raw
+  value and is stamped queryably on the audit row
+  (`payload['reveal_secret']`). The audit `raw_payload` keeps the raw
+  result in both cases — only the transport response is scrubbed
+  (mirrors the #2172 secret-handler posture). The scrub keys on the op
+  class, not the caller kind. Approved human flows keep working by
+  opting in: the `/ui` Vault console read (reveal-on-click) and CLI
+  `meho secret read` (pipe-only) both pass `reveal_secret=true`;
+  `secret.move` stays the recommended no-transit path. The reserved
+  `reveal_secret` flag is stripped before op-schema validation and
+  `params_hash`, so a scrubbed read and its reveal share one hash.
+
 ### Added — operator-console OAuth chart values (#2594)
 
 - The Helm chart now renders the operator-console (`/ui/*`) browser
--- a/backend/src/meho_backplane/broadcast/events.py
+++ b/backend/src/meho_backplane/broadcast/events.py
@@ -72,6 +72,7 @@ __all__ = [
     "classify_op",
     "redact_payload",
     "scrub_broadcast_params",
+    "scrub_secret_named_values",
 ]
 
 
@@ -731,6 +732,32 @@ def _scrub_secret_named_keys(node: Any) -> tuple[Any, bool]:
             found = found or child_found
         return items, found
     return node, False
+
+
+def scrub_secret_named_values(node: Any) -> tuple[Any, bool]:
+    """Key-name-aware secret scrub for a caller-bound dispatch response.
+
+    Public entry to :func:`_scrub_secret_named_keys` (the same walk the
+    broadcast params scrub uses). Returns ``(scrubbed, secret_detected)``:
+    every mapping entry whose key is secret-shaped (``password``,
+    ``client_secret``, ``sessionToken``, ...) has its value subtree
+    replaced with :data:`_REDACTED_PARAM`.
+
+    The connector-boundary redaction engine (``dispatcher.py`` step 7a)
+    matches only *labelled* secret shapes inside string leaves, so a
+    ``credential_read`` response like ``{"data": {"password": "..."}}``
+    passes through it verbatim. The dispatcher runs this scrub on the
+    caller-bound response of a ``credential_read``-classified op by
+    default (#2467) so the raw value never reaches the ``call_operation``
+    caller / agent transcript; ``params.reveal_secret=true`` opts out.
+    The audit row keeps the raw result regardless (the raw payload is
+    captured before this scrub runs), mirroring the #2172 secret-handler
+    posture. Unlike :func:`scrub_broadcast_params` this does **not** run
+    the Tier-1 engine pass -- the boundary middleware already applied it
+    to the response, so this layer is only the structured-dict
+    key-name complement.
+    """
+    return _scrub_secret_named_keys(node)
 
 
 def scrub_broadcast_params(params: dict[str, Any]) -> tuple[dict[str, Any], bool]:
--- a/backend/src/meho_backplane/mcp/tools/operations.py
+++ b/backend/src/meho_backplane/mcp/tools/operations.py
@@ -476,7 +476,17 @@ register_mcp_tool(
                         "Operation-specific parameters. The dispatcher "
                         "validates against the operation's parameter_schema "
                         "before invoking the handler; unknown fields are "
-                        "rejected at the schema layer."
+                        "rejected at the schema layer. One reserved key is "
+                        "handled by the dispatcher itself and never reaches "
+                        "the op schema: `reveal_secret` (bool). A "
+                        "credential_read op (e.g. `vault.kv.read`) returns a "
+                        "key-name-scrubbed response by default -- secret "
+                        "values are replaced so they never enter this agent "
+                        "transcript. Pass `reveal_secret: true` ONLY when you "
+                        "must pipe the raw value onward; the raw value is "
+                        "then returned AND the reveal is stamped on the audit "
+                        "row. Prefer `secret.move` to relocate a credential "
+                        "without it transiting your context."
                     ),
                 },
                 "work_ref": {
--- a/backend/src/meho_backplane/operations/_audit.py
+++ b/backend/src/meho_backplane/operations/_audit.py
@@ -57,6 +57,7 @@ __all__ = [
     "publish_broadcast",
     "resolve_agent_session_id",
     "resolve_broadcast_lineage",
+    "reveal_secret_var",
     "run_id_var",
     "step_id_var",
     "work_ref_var",
@@ -182,6 +183,24 @@ work_ref_var: ContextVar[str | None] = ContextVar(
 #: for the nullable ``audit_log.policy_decision`` column.
 policy_decision_var: ContextVar[str | None] = ContextVar(
     "policy_decision",
+    default=None,
+)
+
+
+#: ContextVar carrying the ``reveal_secret`` opt-in for a ``credential_read``
+#: dispatch (#2467). Bound in
+#: :func:`~meho_backplane.operations.dispatcher.dispatch` with a
+#: token/finally reset scoping it to the dispatch -- the same
+#: bound-at-a-boundary / read-at-the-write-call-site discipline as
+#: :data:`policy_decision_var`. Read by :func:`_build_audit_payload`, which
+#: stamps ``payload["reveal_secret"]`` so "which credential reads returned a
+#: raw value vs. the default key-name-scrubbed response" is queryable on the
+#: audit row. ``True`` = caller opted into the raw value; ``False`` = the
+#: default scrubbed response was returned; ``None`` (the value for every
+#: non-credential_read dispatch, and for pre-gate usage errors) leaves the
+#: key absent.
+reveal_secret_var: ContextVar[bool | None] = ContextVar(
+    "reveal_secret",
     default=None,
 )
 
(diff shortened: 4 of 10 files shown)


damir-topic (commented):




**Anchor task:** #2467 — default key-name scrub for `credential_read` dispatch responses + `reveal_secret=true` opt-in, stamped on the audit row
**Surface area:** backend, devops (CLI)
**Verdict:** APPROVE
**Head SHA reviewed:** `689019cf`

### Acceptance criteria
- [x] `credential_read` responses are scrubbed by default — verified in code (`dispatch()` sets `scrub_response = credential_read and not reveal_secret`; `_reduce_and_audit_success` runs `scrub_secret_named_values` over `redaction.redacted`) and by test (`test_credential_read_response_scrubbed_by_default`: `data.password == [REDACTED:param_name]`, `username`/`version` survive, raw absent from `result.model_dump_json()`).
- [x] `reveal_secret=true` reveals AND stamps the audit row — `test_reveal_secret_true_returns_raw_and_stamps_audit` (raw returned, `payload['reveal_secret'] is True`).
- [x] Audit `raw_payload`/`redaction.raw` ALWAYS keeps the raw value — the scrub only rewrites `redaction.redacted` via `model_copy(update={"redacted": ...})`; both audit writes (`_reduce_or_error(raw_payload_for_audit=redaction.raw)` and `audit_and_broadcast_safe(raw_payload=redaction.raw)`) use `redaction.raw`. `RedactionMiddlewareResult` is a frozen pydantic model; `scrub_secret_named_values` builds a new dict (no in-place mutation), so `raw` cannot alias-mutate. Test asserts `row.raw_payload["data"]["password"] == secret` in BOTH the scrubbed and revealed cases. Mirrors the #2172 posture.
- [x] `reveal_secret` stripped before hashing/schema-validation/handler-forward — stripped at the top of `dispatch()` before `compute_params_hash`; handler asserts `"reveal_secret" not in params`; `test_reveal_secret_does_not_reach_op_schema_or_hash` proves a scrubbed read and its reveal share one `params_hash` and neither 422s on the `additionalProperties: false` op schema.
- [x] Two opt-in sites pass `reveal_secret=true` — `/ui` Vault console read (`_render_read` → `_dispatch(..., reveal_secret=True)`) and CLI `meho secret read` (`params["reveal_secret"] = true`, asserted in `read_test.go`). `meho vault kv read` intentionally not opted in (scrubbed-by-design, flagged for maintainer) — acceptable per the product decision's two named sites.
- [x] No OpenAPI drift — only internal helpers touched in the `/ui` route (`_render_read`, `_dispatch`); no route decorator/signature changed. The MCP `operations.py` edit is a JSON-RPC tool-description string, not a FastAPI model. No `snapshot-openapi` regen needed.
- [x] Docs — `docs/codebase/redaction.md` gains a full `#2467` section; CHANGELOG `[Unreleased]` entry present with a unique header.

### Project-board alignment
|---|---|---|---|---|---|
| #2467 | In Review | Closes | ✅ | ✅ full | close on merge |

Parent Initiative #2494 / Goal #221. No delivered-but-unlinked or linked-but-undelivered mismatches.

### Predicted vs. actual diff
- ✅ `operations/dispatcher.py` — strip flag, classify, `scrub_response`, bind `reveal_secret_var`
- ✅ `operations/_audit.py` — `reveal_secret_var` + `payload['reveal_secret']` stamp (clean refactor of the contextvar-mirror block into `_apply_contextvar_lineage_mirrors`, under the size/complexity caps)
- ✅ `broadcast/events.py` — public `scrub_secret_named_values` entry to the existing `_scrub_secret_named_keys` walk
- ✅ `mcp/tools/operations.py` — documents the reserved `reveal_secret` param
- ✅ `ui/routes/vault/routes.py` + `cli/.../read.go` (+ test) — the two opt-in sites
- ✅ `CHANGELOG.md`, `docs/codebase/redaction.md`, new `tests/test_credential_read_response_scrub.py`
- 🕳️ Missing: none. No undisclosed scope.

### Integration verification
- [x] `ruff check` — clean (6 changed files)
- [x] `mypy` (5 changed source files) — no issues
- [x] `pytest tests/test_credential_read_response_scrub.py` — 7 passed
- [x] `pytest` siblings (`test_operations_dispatcher`, `test_ui_vault`, `test_agent_audit_invocation`, `test_audit_work_ref`) — 73 passed
- [x] `go build ./...` + `go test ./internal/cmd/secret/...` — pass
- [x] Pre-existing teardown ERROR in `test_connectors_vault.py::test_preflight_fail_soft_when_probe_raises` reproduced on **both** this branch and pristine `main` — NOT introduced here (the PR touches no vault-connector preflight code)

None.

- PR body's test plan says "10 cases" for the new file; there are 7 test functions. The coverage is complete for the contract — this is a description/count nit only, no code change required.
- `reveal_secret=true` passed to a non-`credential_read` op is silently stripped and ignored (no stamp, no scrub). This is the correct fail-safe (no leak, no schema break); noting only for awareness.

### Docs
- [x] `docs/codebase/redaction.md` — updated with the #2467 section (accurate to the implementation; no .NET analogies).

- The raw-vs-redacted separation is exactly right: scrub is confined to `redaction.redacted`, `redaction.raw` (the audit ledger copy) is provably untouched, and the test asserts the raw secret survives in `raw_payload` in both the scrubbed and revealed paths.
- Stripping `reveal_secret` before hash/schema/handler and stamping it on a ContextVar with the same token/finally discipline as `policy_decision_var` is a clean, composite-safe design; the shared-`params_hash` test locks the contract.

Issue #2489, opened 2026-07-14
Task: reject grants for unregistered agent principals + surface the principal_kind=agent-only enforcement scope — grant.create/elevate descriptions + permission-model doc


zdamir:
Parent goal: #221
Parent initiative: #2494

## Summary

A `meho.agents.grant.create` with `verdict=needs-approval` silently never gates dispatches when the granted `principal_sub` authenticates without `principal_kind=agent`: the policy gate consults `AgentPermission` rows only for agent-kind tokens, and the grant surface accepts any string sub without checking it names a registered agent principal. The dogfood-observed "needs-approval grant on the caller's own sub does not park the caller's `call_operation`" is neither a self-service exemption nor a precedence bug — it is the documented principal-kind carve-out (G11.2-T3), plus a grant-creation surface that silently accepts unenforceable grants and never says so.

## Current state

- `policy_gate` branches on principal kind before grants are ever loaded: non-`AGENT` principals short-circuit to `_non_agent_verdict` (`backend/src/meho_backplane/operations/_validate.py:207-209`), which is default-allow unless `descriptor.requires_approval` (`_validate.py:100-138`, needs-approval routing since #1401). `AgentPermission` rows are loaded only on the agent branch via `resolve_verdict` (`_validate.py:211-221`, `backend/src/meho_backplane/auth/permissions.py:316-333`).
- A JWT with no `principal_kind` claim resolves to `PrincipalKind.USER` (`backend/src/meho_backplane/auth/jwt.py:685-689`); only clients registered via `meho agent-principal register` carry the `principal_kind=agent` hardcoded mapper (`backend/src/meho_backplane/auth/keycloak_admin.py:177`). A backplane service-account token therefore never reaches the grant resolver — `k8s.about` is not `requires_approval`, so the dispatch auto-executes with zero parked approvals, exactly as observed on v0.21.0.
- "Self" is a red herring: a `needs-approval` grant on any distinct sub is equally unenforced when that sub's tokens lack `principal_kind=agent`.
[...]

PR #2613, opened 2026-07-19
test(memory): prove service-account search_memory round-trip (#2484)

### LINKED ISSUE #2484: mcp: search_memory 0 hits for the backplane service-account — REFUTED at triage; open pending clean same-principal re-probe
Parent goal: #221
Parent initiative: #2494

> **Adopted 2026-07-18 (re-framed).** This was initially triaged as *refuted* in #2494 ("no principal-type filter exists on the path; the only per-principal predicate is `user_sub` string equality") pending a clean same-principal re-probe. Adopted as a task to **close that re-probe**: reproduce the write→search round-trip under the *exact* backplane service-account (`client_credentials`) principal and the exact MCP surface the finding used, then either fix the divergence it surfaces or land the missing regression coverage. The scope is the same-principal MCP `add_to_memory` → `search_memory` round-trip; the reporter's "principal-type filter" hypothesis is a lead, not the assumed root cause.

## Problem (as reported)

Under the backplane service-account JWT (`sub` = `6a695b29-a691-4ed9-a1f9-e4f58a4d86a3`, minted via `scripts/meho-mint-jwt.sh`), every `search_memory` call on v0.21.0 (git `bdf56146`) returns **0 hits** — regardless of query specificity, scope filter, or elapsed time since a write. REST `GET /api/v1/memory?scope=user` under the same JWT returns the rows. See the original reproduction rounds below.

## Reproduction (all live-verified on v0.21.0)

**Round 1 — fresh write + immediate search:**
- `add_to_memory scope=user slug=<unique> ttl=PT1M body="<UNIQUE_TOKEN>"` → 201 with `id` and correct `expires_at`.
- `search_memory query="<UNIQUE_TOKEN>"` polled 5×1s → 0 hits every time.

**Round 2 — 30-second buffer** (in case indexing is async): same create, `sleep 30`, `search_memory` → **0 hits**.

**Round 3 — common terms:** `search_memory query="the"`/`"a"`/`"<body_fragment>"` → 0 hits.

**Round 4 — REST cross-check:** `GET /api/v1/memory?scope=user&limit=5` under the SAME JWT → returns real entries with populated `body` fields Round 3 should have matched.

So the rows EXIST, are OWNED BY the caller's `user_sub`, and are visible via REST — but `search_memory` returns 0 hits.

## Acceptance criteria

- [ ] A regression test reproduces the finding's exact path: mint a `client_credentials` service-account token, `add_to_memory scope=user`, then `search_memory` for a token in the just-written body — over the **MCP** surface, single principal, against Postgres (not a unit fake). The test currently fails (or, if it passes, the divergence is proven to be surface/principal-mixing in the probe and this task closes with that evidence recorded — matching the #2494 DoD "re-probed clean before closing").
- [ ] If the round-trip diverges, the fix makes MCP `search_memory` return the SA's own writes under `user_sub`, matching the REST list rule (`rbac.py:165-168`); the `memory/retriever.py:449-455` predicate and any embedding/index path are inspected for a `client_credentials`-specific gap (empty query vector, missing embedding on the SA-written row, scope coercion).
- [ ] REST list and MCP search return the same rows for the same principal on the same scope — asserted in the test.
- [ ] No behaviour change for interactive-OIDC principals; existing memory tests stay green. CHANGELOG `[Unreleased]` note if a wire/behaviour fix lands.

damir-topic:
## Summary
- Closes the #2494 re-probe of the v0.21.0 field finding that `search_memory` returned **0 hits** under the backplane service-account (`client_credentials`) JWT while the same row was visible via the REST list verb.
- The finding was triaged **refuted**, and the code confirms it: `principal_kind` (`user`/`service`/`agent`/`runner`) is **never** consulted on the memory or retrieval path — there is no principal-type filter. The only per-principal predicate is `metadata ->> 'user_sub' = :principal_sub` string equality (`retriever.py` `_PRINCIPAL_PREDICATE_SQL`), mirroring the REST list rule (`MemoryRbacResolver.can_read`, `rbac.py` L165-168). `remember` stamps `user_sub = operator.sub` on write; `search_memories` filters on the same `operator.sub`.
- This adds a Docker-gated integration suite that reproduces the finding's **exact path** against a **real pgvector cluster** under a single, consistent `client_credentials` principal (reusing the finding's `sub`): mint a `principal_kind=service` operator → `add_to_memory scope=user` over the **MCP tool surface** → `search_memory` for a token in the just-written body. The round-trip is **clean**; the service account retrieves its own write.
- **The finding did not reproduce** for one consistent principal. The reported 0-hits was a surface/principal-mixing artefact of the probe (a search `sub` that differed from the write `sub`), exactly as the #2494 triage predicted. Per the #2494 DoD ("re-probed clean before closing"), the proof-of-round-trip test is the deliverable. **No behaviour change** — interactive-OIDC principals are scoped by the same `sub` equality and are unaffected.

## Test plan
- [x] `mypy --ignore-missing-imports` — clean
- [x] New suite collects (4 tests, imports resolve); **skips** in-sandbox (no Docker) and **runs in CI** where the pgvector testcontainer is provisioned — same gate as the rest of `tests/integration/`.
- [x] Existing always-on memory/isolation unit tests green: `test_retrieve_isolation.py` + `test_mcp_tools_memory.py` + `test_memory_rbac.py` — 104 passed.
- [x] AC coverage:
  - Regression test reproduces the finding's exact path (MCP surface, single `client_credentials` principal, real Postgres) — `test_service_account_search_finds_own_user_scoped_write` (+ `scope=user` and stopword/cosine-only variants).
  - REST list and MCP search return the **same** row for the same principal/scope — `test_rest_list_and_mcp_search_return_the_same_row`.
  - No behaviour change for interactive-OIDC principals (test-only diff).

## Findings (pushback-lite)
The reporter's "principal-type filter" hypothesis has **no basis in code** — grep of `memory/` + `retrieval/` shows zero `principal_kind`/`PrincipalKind` references. The divergence cannot arise for one consistent principal, so this ships as regression coverage (`### Tested` CHANGELOG entry), not a wire/behaviour fix.

## Out of scope
- Related #2489 (grant principal guard) — same principal-kind family at enforcement time (token-sub ↔ keycloak_client_id linkage), tracked separately.

Closes #2484


damir-topic:




**Anchor task:** #2484 — re-framed #2494 re-probe of the `search_memory` service-account 0-hits field finding
**Surface area:** backend (test-only) + ai (memory/retrieval path)
**Best-practices packs audited:** python, ai
**Verdict:** APPROVE
**Head SHA reviewed:** `dfeb7996`

### Acceptance criteria
- [x] **No principal-type filter exists (code claim sound)** — `grep principal_kind|PrincipalKind` across `backend/src/meho_backplane/memory/` + `retrieval/` returns **zero** hits. The only per-principal predicate is `metadata ->> 'user_sub' = :principal_sub` string equality (`retriever.py:449-455`, inlined verbatim into both BM25 and cosine candidate queries), mirroring `MemoryRbacResolver.can_read`'s `operator.sub == user_sub` for USER_SCOPED (`rbac.py:165-171`). A `client_credentials` service account is therefore scoped by exactly the same `sub` equality as a human operator — the "no divergence" conclusion is correct.
- [x] **Test exercises the exact path** — mints an `Operator(principal_kind=SERVICE, sub=<finding's sub>)`, writes via the real MCP `_add_to_memory_handler` (`scope=user`), searches via the real MCP `_search_memory_handler`, over the real `pg_engine` pgvector testcontainer. `search_memories` pushes `principal_sub=operator.sub` (`service.py:778,847`); `remember` → `index_document` computes and persists the embedding (`indexer.py:273,308`). Same `sub` writes and searches — the round-trip is the proof.
- [x] **Stopword/cosine-only guard is genuinely anti-vacuous** — `search_memory("the")` yields an empty `plainto_tsquery`, so BM25 contributes nothing and the row must arrive purely on cosine. If the SA write had left `embedding` NULL this assertion would fail while the token-overlap probes still passed — a real NULL-embedding regression on the SA-written row is caught.
- [x] **Docker-gate honest, not vacuous green** — `@pytest.mark.skipif(not DOCKER_AVAILABLE)` on every test; the predicate is PG-only (pgvector cosine + jsonb `->>`), unrunnable on SQLite. Verified locally: no Docker socket → 4 skipped, imports resolve, no collection error. CI's "Python (integration testcontainers)" lane provisions Docker so these run for real there.
- [x] **No behaviour change for interactive-OIDC principals** — diff is CHANGELOG + one new test file; zero production code touched.
- [x] **CHANGELOG `### Tested` entry present** — `### Tested — search_memory service-account round-trip (#2484)`, distinct header, accurate.

### Project-board alignment
|---|---|---|---|---|---|
| #2484 | In Review | Closes | ✅ | ✅ full (re-framed AC: regression proof, not a fix) | close on merge |

### Predicted vs. actual diff
- ✅ `backend/tests/integration/test_memory_service_account_roundtrip_e2e.py` — expected and correct
- ✅ `CHANGELOG.md` — expected and correct
- 🕳️ No production-code change was *predicted or required*: the finding is refuted (no principal-kind filter to remove), so the deliverable is regression coverage. Correctly absent.

### Integration verification
- [x] New suite collects + imports clean; 4 skipped without Docker (honest gate) — ran locally
- [x] `pytest tests/test_retrieve_isolation.py tests/test_mcp_tools_memory.py tests/test_memory_rbac.py` — 104 passed (matches PR claim)
- [x] Code claim verified from source: `principal_kind` absent from memory/retrieval; `_PRINCIPAL_PREDICATE_SQL` is `user_sub` equality only
- [⚠️ not run] Docker-gated round-trip itself — no Docker socket in review sandbox; runs in CI's integration-testcontainers lane (IN_PROGRESS at review time). Reasoned end-to-end: write persists embedding + `user_sub=sub`, search filters on same `sub`, assertions fail on 0-hits — the test would catch a genuine principal-kind divergence, so it is not vacuous.

None.

- The three round-trip probes (scope-omitted, `scope=user`, stopword-only) share one `written_memory` fixture; if a future change makes the write path lazy about embeddings, only the stopword test guards it. That guard is present and correct — noting only that it is the single line of defence for the NULL-embedding case.

### Docs
- [n/a] No `docs/codebase/` doc invalidated — test-only PR closing a refuted re-probe; the memory/retrieval contract is unchanged.

- The stopword-only cosine probe is a deliberate anti-vacuousness guard — the test author explicitly reasoned that a NULL embedding could otherwise let a token-overlap probe mask a real bug, and pinned it. This is exactly the honesty bar a "clean re-probe closes the ticket" deliverable needs.
- Reproduction drives the real MCP handler surface the finding named, not the service shortcut, so the "exact path" claim holds.

PR #2619, opened 2026-07-19
fix(kb): attribute-selector form for dot-slug preview targets on /ui/kb

### LINKED ISSUE #2451: Task: fix /ui/kb preview selectors for dot-containing slugs — attribute-selector form at both the hx-target and the after-swap handler
Parent goal: #221
Parent initiative: #2495

## Summary

Both selector sites in the KB result-card preview wiring inline the entry slug into a `#id` CSS selector unescaped. The KB slug charset explicitly allows `.` (dots), and `.` is the class delimiter in CSS selector syntax, so for every dot-slug card (a) the Alpine after-swap handler throws `SyntaxError` on every htmx swap in the document, and (b) the `hx-target` cannot be resolved, so the hover-preview swap itself fails. Fix both sites with the attribute-selector form.

## Current state (file:line evidence)

- `backend/src/meho_backplane/ui/templates/kb/_results.html:96` — `hx-target="#kb-preview-{{ hit.slug }}"`. htmx resolves `hx-target` via `querySelector`; a dot slug makes the selector invalid (digit-leading class part, e.g. `meho-0.2-deployment` → `SyntaxError`) or silently wrong (letter-leading part → matches nothing) — the preview fetch/swap is broken either way.
- `backend/src/meho_backplane/ui/templates/kb/_results.html:101` — `x-on:htmx:after-swap.document="$el.closest('.card').querySelector('#kb-preview-{{ hit.slug }}').classList.remove('hidden')"`. The `.document` modifier runs this handler on EVERY htmx swap anywhere on the page — including the search-results swap itself — which is why a single search response emits one `SyntaxError` per dot-slug card (61 observed).
- `backend/src/meho_backplane/ui/templates/kb/_results.html:45` — `id="kb-preview-{{ hit.slug }}"` is a valid HTML id even with dots; the ids are fine, only the selectors built from them are broken.
- Slug charset: `backend/src/meho_backplane/kb/schemas.py:99` — `SLUG_PATTERN = ^[a-z](?:[a-z0-9.\-]*[a-z0-9])?$`. Dots are legal; quotes, backslashes, and whitespace are not. Therefore `[id="kb-preview-{{ hit.slug }}"]` is a valid CSS selector for every legal slug with no client-side escaping needed (Jinja autoescape covers the HTML-attribute context).
- No JS under `ui/static/src/` and no test references `kb-preview-` ids (grep), so the change is contained to this template plus a new test.

## Desired state

- Line 96 (`hx-target`) and line 101 (after-swap `querySelector`) both use the attribute-selector form `[id="kb-preview-{{ hit.slug }}"]` (or an equivalent that never places the raw slug in id/class-selector position).
- Optional hardening, not required for correctness: guard the after-swap handler to act only when the swap targeted its own preview div (inspect `$event.detail.target`), so document-wide swaps stop invoking N no-op handlers per keystroke.

## Acceptance criteria

1. `grep -n "querySelector('#kb-preview" backend/src/meho_backplane/ui/templates/kb/_results.html` and `grep -n 'hx-target="#kb-preview' backend/src/meho_backplane/ui/templates/kb/_results.html` both return nothing.
2. A test renders the results partial with a hit whose slug contains a dot (e.g. `meho-0.2-deployment`) and asserts the fragment contains `[id="kb-preview-meho-0.2-deployment"]` in both the `hx-target` attribute and the Alpine handler expression.
3. Browser check: a search returning dot-slug cards produces zero console errors, and the Preview button swaps the preview partial into the correct card and unhides it (screenshot of DevTools console + a working dot-slug preview).

## Out of scope

- Changing `SLUG_PATTERN` or the slug-derived element ids.
- The hover-preview UX itself (trigger timing, partial content).

## References

- htmx `hx-target` (standard CSS selector, resolved via querySelector): https://htmx.org/attributes/hx-target/
- CSS attribute selectors (quoted value form accepts dots): https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors

damir-topic:
## Summary

- On `/ui/kb`, the hover-preview affordance broke for any KB slug containing a dot. Both the `hx-target` and the Alpine `after-swap` handler on each result card interpolated the slug into a `#kb-preview-<slug>` CSS selector, where `.` parses as a class delimiter — making the `hx-target` unresolvable and throwing a `SyntaxError` (one per dot-slug card) in the document-wide after-swap handler on every htmx swap.
- Both selector sites now use the attribute-selector form `[id="kb-preview-<slug>"]`, a valid selector for every legal slug. The dotted element ids themselves are unchanged (valid HTML ids).

## Test plan

- [x] `pytest tests/test_ui_kb_search.py` — 25 passed (adds `test_kb_search_dot_slug_uses_attribute_selectors`)
- [x] AC #1 — `grep "querySelector('#kb-preview"` and `grep 'hx-target="#kb-preview'` against `_results.html` both return nothing
- [x] AC #2 — the new test renders the results partial with a dot slug (`meho-0.2-deployment`) and asserts `[id="kb-preview-meho-0.2-deployment"]` at both the `hx-target` (literal, single-quoted attribute) and the after-swap handler (HTML-escaped `&quot;` on the wire, decoded to `"` by the browser / `html.unescape`)
- [ ] AC #3 — browser check (zero console errors + working dot-slug preview) runs where a live console is available; the render-level contract is proven by the unit test above

## Out of scope

- The optional `$event.detail.target` guard (issue "Optional hardening, not required for correctness") — the after-swap handler still runs document-wide, unchanged pre-existing behavior; not required to fix the dot-slug break.
- `SLUG_PATTERN` and the slug-derived element ids (unchanged).

Closes #2451


damir-topic (commented):




**Anchor task:** #2451 — fix `/ui/kb` preview selectors for dot-containing slugs (attribute-selector form at both sites)
**Surface area:** frontend (HTMX/Jinja operator console, G10.x)
**Verdict:** APPROVE
**Head SHA reviewed:** `01671ced`

### Acceptance criteria
- [x] AC #1 — no raw `#kb-preview-<slug>` selector survives at either site. Verified in diff (both `hx-target` and after-swap `querySelector` switched) and asserted by the new test (`querySelector('#kb-preview` and `hx-target="#kb-preview` both absent from the rendered body).
- [x] AC #2 — new test renders a dot slug (`meho-0.2-deployment`) and asserts the attribute-selector form at both sites: `hx-target='[id="kb-preview-…"]'` (literal, single-quoted HTML attr) and `querySelector('[id="kb-preview-…"]')` after `html.unescape` (the after-swap attr is double-quoted, so its selector quotes ship as `&quot;` and the browser/`html.unescape` decode them). Both forms verified correct.
- [~] AC #3 — live-browser zero-console-error + working dot-slug preview screenshot is inherently non-CI-automatable; the render-level contract (both selectors valid for a dotted slug, dotted element id unchanged) is proven by the unit test. Acceptable — reasoned, not a blocker.

### Project-board alignment
|---|---|---|---|---|---|
| #2451 | In Progress | Closes | ✅ | ✅ full | close on merge |

### Predicted vs. actual diff
- ✅ `backend/src/meho_backplane/ui/templates/kb/_results.html` — both selector sites (lines 96, 101) converted to attribute-selector form; dotted element id (line 45) correctly left unchanged.
- ✅ `backend/tests/test_ui_kb_search.py` — new dot-slug render test asserting both sites + unchanged id.
- ✅ `CHANGELOG.md` — `[Unreleased] > Fixed` entry with a distinct `#2451` header.
- 🕳️ Missing: none. No OpenAPI regen expected (template/test only, no route/schema change) — correctly absent.

### Integration verification
- [x] `ruff check` (test file) — clean
- [x] `ruff format --check` (test file) — already formatted
- [x] `pytest tests/test_ui_kb_search.py` — 25 passed (was 24; +1 new)
- [x] Selector correctness reasoned against CSS/htmx semantics — a dotted attr value is not a valid unquoted CSS identifier, so quoting is required; `[id="…"]` resolves via `querySelector` for every legal slug. Both quote-context choices render the intended selector.
- [x] XSS/escaping — slug is server-validated (`validate_slug` / `SLUG_PATTERN` = `^[a-z](?:[a-z0-9.\-]*[a-z0-9])?$`, no quotes/backslashes/`<`/`>`/`&`) with Jinja autoescape as defense-in-depth; no breakout of either the single-quoted or double-quoted attribute. No injection surface introduced by the quote handling.
- [⚠️ not run] live browser DevTools console check — non-CI; see AC #3.

None.

- The optional `$event.detail.target` guard on the document-wide after-swap handler is explicitly out of scope per the issue ("optional hardening, not required for correctness"). Left correctly deferred; not requested here.
- The PR is currently `mergeable:false` on a `CHANGELOG.md [Unreleased]` conflict (expected serial-merge race). The distinct `#2451` header makes a merge-main-in union trivial; orchestrator handles at merge time. Not a review blocker.

### Docs
- [n/a] No `docs/codebase/` artifact covers this template selector detail; the fix is a two-line template correction fully captured by the CHANGELOG and the new test. No doc invalidated.

- Both quote contexts handled precisely and documented in the test with inline rationale (`&quot;` on the wire vs. literal double-quotes in the single-quoted attr), and the test verifies the decoded form via `html.unescape` rather than asserting the raw escaped string — a faithful check of what the browser actually sees.
- Tight blast radius: element ids left unchanged, only the selectors built from them fixed.

Issue #2507, opened 2026-07-15
Task: wire the tiered-triage investigator to Dashboard non-green transitions — topology-correlated, noise-suppressed, budget-gated, diagnose-only runs via AgentInvoker.run_scheduled


damir-topic:
Parent goal: #221
Parent initiative: #2416
Depends-on: #2503
Depends-on: #2505
Depends-on: #2506

## Summary

Productize the `examples/r1-tiered-triage` investigator mould against the Dashboard check layer: when a Dashboard's five-state rollup transitions from green to a non-green state — detected at #2505's runner result-persist seam against the `last_rollup_state` memo #2506 ships — correlate the affected non-green Sensors through the topology blast-radius graph so one underlying cause produces exactly ONE investigation, check tenant memory for a known-noise policy that suppresses re-escalation, and — only for novel, non-suppressed groups — fire the deep-tier investigator as a real, durable, budget-gated agent run via `AgentInvoker.run_scheduled`. The investigator is diagnose-only: its structured finding (verdict, summary, evidence, suggested action) lands in the durable `agent_run` row and is written back to memory as the noise-suppression policy; any change op the agent attempts parks in the existing approval queue and is never executed by this wiring.

## Current state

- **The mould is an examples/ harness, not a deployed service.** `examples/r1-tiered-triage/workflow.py:613-731` (`run_closed_loop`) drives an in-process `PydanticAgentRun` with harness-owned `recorder`/`finalizer` hooks; the cheap tier is an LLM classifier that escalates via `invoke_agent`. Its closed loop: the deep tier returns a structured `PolicyDecision` (`workflow.py:190-210` — `verdict ∈ {benign, acknowledged, actionable}`, `re_escalate: bool`, `summary`, `evidence`, `recommended_action`), and the **harness** (not the agent) persists it to memory under `r1-policy-<alert_class>` (`workflow.py:544-576`, slug prefix constant at `:141`, rationale for harness-side persistence at `:26-41`, render shape at `:579-610`, skip-on-unparseable-output arm at `:515-541`).
[...]

Issue #826, opened 2026-05-21
Task: G11.3-T5 scheduler admin surface (create/list/cancel triggers on CLI+MCP+REST) + durability test


damir-topic:
Parent goal: #800
Parent initiative: #804

## Context

Operators need to create, list, and cancel triggers, and verify the scheduler's durability. This Task ships the admin surface over the trigger model + a durability test that proves the floor of 24/7 operation.

## Current state

- No trigger admin surface (the model is 11.3-T1). CLI/MCP/REST surface-wiring precedents throughout `cli/internal/cmd/`, `mcp/tools/`, `api/v1/`. Migration/durability test precedents in `backend/tests/`.

## Desired state

- Verbs to create / list / show / enable-disable / cancel triggers (cron, one-off, event) on CLI + MCP + REST, tenant-scoped + RBAC-gated (`tenant_admin` to write; operator to list). Creating a trigger references an existing agent definition + identity.
- A durability/integration test: create triggers → restart the scheduler/process → assert no trigger is lost and none double-fires; a one-off in the past fires once on startup (catch-up) per policy.

## Acceptance criteria

- [ ] Create/list/show/enable-disable/cancel triggers across CLI + MCP + REST; tenant-scoped + RBAC-gated.
- [ ] A durability test restarts the scheduler and asserts triggers survive, fire correctly, and don't double-fire.
- [ ] `Python (ruff + mypy + pytest)` + integration green (+ `Go` for CLI verbs).

## Out of scope

- The firing/resume internals (11.3-T2/T3/T4) — this is the management surface + the end-to-end durability proof over them.

## References

- Grounding: surface-wiring precedents [cli/internal/cmd/](cli/internal/cmd/), [mcp/tools/](backend/src/meho_backplane/mcp/tools/), [api/v1/](backend/src/meho_backplane/api/v1/); durability-test patterns in `backend/tests/`.
- Best-practices: **devops_best_practices** pack (operable scheduling, durability tests) + **python_best_practices** pack.

PR #2637, opened 2026-07-19
feat(checks): atomically coalesce concurrent Dashboard investigations (#2575)

### LINKED ISSUE #2575: Task: atomically coalesce concurrent Dashboard investigations (per-(tenant,dashboard) claim) — #2507 follow-up
Parent goal: #221
Parent initiative: #2416

## Summary

Make the check-layer investigator's transition-detect → correlate → in-flight-claim → schedule-run sequence **atomic per (tenant, dashboard)**, so concurrent Sensor outcomes for the same Dashboard coalesce into exactly one investigation over the fully-settled member set. Deferred follow-up from #2507's review (CodeRabbit "Heavy lift", PR #2569 review comment on `checks/investigate.py`; deferred in the #2574 fast-follow).

## Problem

The per-Sensor-result hook `investigate_on_transition` (`backend/src/meho_backplane/checks/investigate.py:298`), the `last_rollup_state` memo compare-and-set, and the in-flight gate `_has_in_flight_run` (`investigate.py:871`) form a **non-atomic** scheduling protocol. Two Sensor outcomes for the same Dashboard that land close together can interleave such that:

- **Duplicate runs** — both outcomes pass the memo transition and the `_has_in_flight_run` check before either persists its `AgentRun`, so two investigator runs fire for one transition; or
- **Partial correlation** — the correlation snapshot (`run_investigation` → `_correlate`, `investigate.py:505`) is taken before all concurrently-failing members have persisted, so the investigation covers only the first failure and misses correlated siblings.

The in-flight dedup keys on `(tenant_id, work_ref)` and the run is idempotent, so the blast radius is **bounded**: worst case a duplicate *budget-gated, diagnose-only* run, or a first-pass investigation that self-corrects on the next tick/transition. This is a hardening, not an active correctness or safety breach — hence it was split out of the #2574 fast-follow rather than folded in.

## Approach (sketch — implementer decides)

Serialize the transition + claim per `(tenant, dashboard)` using the same durability primitive the check-runner already relies on (advisory lock + `FOR UPDATE SKIP LOCKED`, the #804 / #2505 `scheduler/loop.py` pattern), OR make the in-flight claim a DB-level atomic upsert on `(tenant_id, work_ref)` so the claim-and-create is a single statement. Whichever is chosen, the memo transition, the member-set finalization, and the run creation must observe a consistent, settled view. Reuse existing seams — do not add a new bespoke locking scheme.

## Acceptance criteria

- Two concurrent Sensor outcomes for the same Dashboard produce **exactly one** investigator `AgentRun` (regression test with two interleaved `investigate_on_transition` calls).
- The single investigation's correlated member set reflects **all** members that had failed at claim time, not just the first-persisted one.
- No change to the diagnose-only guarantee, the budget gate, or the tenant-scoping of the synthetic investigator operator.
- Replica-safe: the claim holds across processes (not just within one event loop).

## References

- `backend/src/meho_backplane/checks/investigate.py:298` (`investigate_on_transition`), `:505` (`run_investigation`/`_correlate`), `:871` (`_has_in_flight_run`)
- `backend/src/meho_backplane/checks/runner.py` + `scheduler/loop.py` — the #804/#2505 advisory-lock + `FOR UPDATE SKIP LOCKED` durability pattern to reuse
- PR #2569 (the #2507 investigator wiring) CodeRabbit review comment "Atomically coalesce and claim Dashboard investigations"; deferred in PR #2574

damir-topic:
## Summary
- Concurrent check-Dashboard investigations for the same `(tenant, dashboard)` now **coalesce onto exactly one investigator run** instead of racing the memo-transition → in-flight-check → schedule sequence into duplicate runs or partial correlation (deferred #2507 review follow-up).
- Each green→non-green edge is claimed under a **transaction-scoped Postgres advisory lock** (`pg_advisory_xact_lock`, keyed on the `(tenant, dashboard)` blake2b digest — the topology-scheduler key mould), with the member set read **under the lock** so the winning investigation correlates over the settled membership.
- The `check_dashboards.last_rollup_state` memo is finalised with an **atomic compare-and-swap** (`IS NOT DISTINCT FROM`-guarded conditional `UPDATE`) whose row count is the claim token — only the caller that wins the swap fires. DB-enforced, so the guarantee holds across replicas, not just within one event loop.
- **No schema migration** — reuses #2506's `last_rollup_state` column and the existing `(tenant_id, work_ref)` in-flight index. No FastAPI route/schema change (internal module), so no OpenAPI snapshot regen.

## Test plan
- [x] `mypy src/meho_backplane/checks/investigate.py` — clean
- [x] `pytest tests/test_checks_investigate.py` — 28 passed (3 new)
- [x] **AC1 exactly one AgentRun:** `test_concurrent_transitions_fire_one_agent_run` (two `asyncio.gather`ed `investigate_on_transition` calls sharing a topology anchor → one `run_scheduled`) + `test_concurrent_transitions_coalesce_to_one_investigation` (one scheduled)
- [x] **AC2 full member set at claim time:** the single investigation's members == `{sensor-a, sensor-b}`
- [x] **AC3 diagnose-only / budget / tenant-scope unchanged:** containment + diagnose-only + suppression suites still green (`test_module_never_imports_operations_dispatcher`, budget/unconfigured/credentials)
- [x] **AC4 replica-safe:** claim is DB-enforced (advisory lock + atomic CAS), not a per-process guard
- [x] Adjacent suites: `test_checks_rollup / test_checks_dashboards / test_checks_admin / test_db_dashboard / test_sensor_runner / test_ui_checks` — 114 passed

## Out of scope
- The residual persist-hook timing gap (a member not yet committed when the winning claim reads membership re-escalates on its own next worsening edge — all runs diagnose-only) is documented in `docs/codebase/checks-investigator.md`; migrating off the runner-persist seam to an event trigger is separate (#826).
- Pre-existing `investigate.py` file-size warning (>600 lines; now 1221) — untouched legacy debt, diff-only code-quality gate reports 0 blockers.

Closes #2575


damir-topic (commented):




**Anchor task:** #2575 — atomically coalesce concurrent Dashboard investigations (per-`(tenant, dashboard)` claim)
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `fddeb73f`

The concurrency design is sound. Concurrent check-Dashboard investigations for the same `(tenant, dashboard)` now coalesce onto exactly one investigator run, and the claim is genuinely atomic and race-free on both the Postgres and the lockless SQLite paths.

### Acceptance criteria
- [x] **AC1 — exactly one AgentRun for two concurrent outcomes** — `test_concurrent_transitions_fire_one_agent_run` (two `asyncio.gather`ed `investigate_on_transition` calls, shared topology anchor → `stub.calls == 1`). Verified passing locally. Genuinely discriminating: the stub persists no `agent_run` row, so the `_has_in_flight_run` gate cannot dedupe — without the CAS claim both outcomes would fire.
- [x] **AC2 — member set reflects all failed members at claim time** — `test_concurrent_transitions_coalesce_to_one_investigation` asserts the single fire's members == `{sensor-a, sensor-b}`. Members are read *under the lock* (`members_by_dashboard` after `_lock_dashboard_transition`), so the winner correlates over the settled set.
- [x] **AC3 — diagnose-only / budget / tenant-scope unchanged** — no change to those paths; `_investigate_group`/budget/containment suites remain green (part of the 28 passed). Cross-tenant guard (`dashboard.tenant_id != tenant_id`) preserved in the re-fetch under lock.
- [x] **AC4 — replica-safe** — the claim is DB-enforced (`pg_advisory_xact_lock` + `IS NOT DISTINCT FROM`-guarded conditional `UPDATE`), not a per-process guard.

### Concurrency correctness (the core ask)
- **Transaction-scoped lock** — `_lock_dashboard_transition` issues `SELECT pg_advisory_xact_lock(:k)`, no manual unlock. Auto-released at commit; cannot strand on crash. Correctly distinguished from the topology/check-runner session-level `pg_advisory_lock` + `pg_advisory_unlock` pair.
- **Key derivation** — `blake2b(tenant.bytes + dashboard.bytes, digest_size=8) & 0x7FFF_FFFF_FFFF_FFFF`. Deterministic, well-distributed, non-negative signed-63-bit `bigint`. Byte-for-byte mirrors `topology.scheduler._advisory_lock_key`. Collision-safe for realistic cardinalities.
- **Members under lock** — read after the lock is acquired via a fresh `session.get(CheckDashboard, …)` re-fetch; no partial-member-set snapshot.
- **Atomic CAS** — `UPDATE … WHERE id=? AND last_rollup_state IS NOT DISTINCT FROM expected`; `rowcount == 1` is the claim token. NULL memo compares cleanly.
- **PG interleaving** — advisory lock serialises the critical section: winner CAS `NULL→critical` (rowcount 1), commits, releases; loser re-reads `critical`, `current == previous_memo` short-circuits the CAS → `None`. Single-fire, no double-win, no lost-update.
- **SQLite interleaving** — lock is a no-op; SQLite write-serialisation means the second `UPDATE` re-evaluates its `WHERE` against the first's committed write → rowcount 0. Single-fire holds. Empirically confirmed: the `asyncio.gather` race test coalesces to exactly one fire.
- **Deadlock** — each claim takes exactly one lock for one dashboard then commits; no nested/ordered locks → no deadlock. Per-dashboard transactions (was one batch transaction) are an acceptable trade for the coalescing guarantee.
- **Both-win / both-lose** — impossible to both-win (serialisation + committed-read CAS); "both-lose" only in the genuinely-unchanged-memo case, which correctly schedules nothing.

### Project-board alignment
|---|---|---|---|---|---|
| #2575 | Open (task) | Closes | ✅ | ✅ full | close on merge |

Parent Initiative #2416 (Sensors & Dashboards) / Goal #221. No sibling-PR collision.

### Predicted vs. actual diff
- ✅ `backend/src/meho_backplane/checks/investigate.py` — claim helpers (`_claim_dashboard_transition`, `_dashboard_transition_lock_key`, `_lock_dashboard_transition`, `_claim_rollup_transition`)
- ✅ `backend/tests/test_checks_investigate.py` — 3 new race/lost-swap tests
- ✅ `docs/codebase/checks-investigator.md` — Concurrency note flipped from "single-fire not guaranteed" to the DB-enforced coalescing guarantee
- ✅ `CHANGELOG.md` — `[Unreleased]` entry
- 🕳️ none missing — no migration expected (reuses 0065 `last_rollup_state`; head stays 0066), no route touched so no OpenAPI regen

### Integration verification
- [x] `ruff check` src/meho_backplane/checks/investigate.py — clean
- [x] `mypy src/meho_backplane/checks/investigate.py` — clean
- [x] `pytest tests/test_checks_investigate.py` — 28 passed (3 new)
- [x] `pytest tests/test_checks_rollup.py tests/test_checks_dashboards.py` — 26 passed
- [x] migration head — 0066 unchanged, no migration in diff
- ⚠️ not run — full 114-suite adjacent set (ran the checks subset; PG-dialect advisory-lock path reasoned-only — the SQLite CAS path is the one under test)

### Findings

- `test_lost_swap_caller_does_not_reschedule` exercises the *sequential unchanged-memo* path and would also pass under the pre-PR code, so on its own it is not a regression discriminator for the new CAS — the two `asyncio.gather` tests are the ones that isolate the claim. Fine to keep as a companion; a future tightening could assert on the CAS `rowcount` path directly.
- CHANGELOG/PR body says the claim "reuses the existing `(tenant_id, work_ref)` in-flight index" — the CAS actually keys on the primary key (`id`); that index still belongs to `_has_in_flight_run`. The statement is true as "no new index needed" but slightly conflates the two mechanisms.
- PR is `mergeable_state: dirty` — the only true conflict is `CHANGELOG.md` `[Unreleased]` (the known merge race). Orchestrator resolves via merge-main-in union at merge; not a code issue.
- The status-check rollup shows CodeQL / DCO / pr-smoke / CodeRabbit-status but no backend unit lane; the orchestrator should confirm the unit lane is green in CI before merge (verified green locally).

### Docs
- [x] `docs/codebase/checks-investigator.md` — updated: Concurrency section now documents the coalescing guarantee and the residual persist-hook timing gap (diagnose-only, acceptable boundary; event-trigger migration tracked at #826).

### What worked well
- Transaction-scoped lock choice (`pg_advisory_xact_lock`) over the runner's session-level lock is the right call and the crash-safety rationale is documented at the call site.
- The CAS-rowcount-as-claim-token pattern gives a single-fire guarantee that holds on the lockless SQLite test path *and* across Postgres replicas — the design cleanly separates the lock (correlation-settling) from the CAS (single-fire), and the tests target the CAS in isolation from the in-flight gate.

Issue #2398, opened 2026-07-12
Task: thread tls_server_name SNI into every session-establish request — vmware /api/session, vcf_session_login (vROps/vRLI/SDDC), NSX, VCFA dual-plane, proxmox ticket, keycloak token grant, profiled logins


damir-topic:
Parent goal: #221
Parent initiative: #2494

## Summary

`tls_server_name` (the per-target TLS SNI + cert-verify name from #2002 / PR #2021) is honored only on the shared dispatch request seams. Every **session-establish** login request across the connector fleet is issued directly on the pooled httpx client with no `extensions=`, so httpcore falls back to `server_hostname = origin.host` and verifies the appliance's FQDN-SAN cert against the registered IP: `connector_tls_verify_failed` → `[SSL: CERTIFICATE_VERIFY_FAILED] IP address mismatch` **before** auth. This blocks secure (`verify_tls=true` + `tls_ca_pin`) registration of every nested vCenter reached by a DNAT-alias IP. #2395 (PR #2404, `cc142dfb`, in v0.22.0) silently widened the blast radius: vROps moved from stateless Basic — whose credential rode the SNI-carrying dispatch seam — to a `token/acquire` login that bypasses it, so the OpsToken fix is unreachable on any by-IP `verify_tls=true` target (it traded a 401 for a TLS-establish failure). sddc looks unaffected only because the exercised targets run `verify_tls=false`.

## Current state (file:line evidence)

- The correct plumbing exists and dispatch uses it: `adapters/http.py:503-531` (`_request_extensions` → `{"sni_hostname": tls_server_name}`), applied at `http.py:594` (`_request_json`) and `http.py:661` (`_post_json`).
- Session-establish sites bypassing it (direct `client.post` / `client.request`, no `extensions=`):
  - `vmware_rest/connector.py:457` + `:463` — `POST /api/session` and the legacy-fallback POST.
  - `_shared/vcf_auth.py:597` — `vcf_session_login()` login POST; consumers: `vcf_operations/connector.py:317` (`POST /suite-api/api/auth/token/acquire`), `vcf_logs/connector.py:400` (vRLI sessions), `sddc_manager/connector.py:290` (`POST /v1/tokens`).
  - `nsx/connector.py:253` — form-encoded `POST /api/session/create`.
  - `vcf_automation/_auth.py:128` + `:180` — provider- and tenant-plane logins.
[...]

Issue #2002, opened 2026-06-20
Task: decouple connector TLS SNI/cert-verification host from the HTTP Host header (target tls_server_name)


damir-topic:
## Summary

After a target is hardened to `host=<cert-CN FQDN>` + `verify_tls=true` + CA-pin, the vRLI appliance **403s** because meho sends that FQDN as the HTTP `Host` header, but the appliance only accepts `Host: <IP>` (and the FQDN is NXDOMAIN off-cluster). meho derives the TCP connect address, the wire `Host` header, **and** the TLS SNI/cert-verification name all from the single `host` value, with no way to decouple them — so verified TLS is structurally unusable for an appliance that pins its cert to an FQDN but enforces `Host: <IP>`, and operators are pushed back to the insecure `verify_tls=false`.

## Current state

- `HttpConnector._base_url` ([`connectors/adapters/http.py:336-339`](backend/src/meho_backplane/connectors/adapters/http.py#L336-L339)) builds one `https://{target.host}{port}` URL; `httpx.AsyncClient(base_url=…)` derives all three of (a) connect address, (b) `Host:` header, and (c) TLS SNI + cert-CN/SAN verify name from that single host (httpcore: `server_hostname = sni_hostname or origin.host`). The connector passes no `Host` override and no `extensions`.
- The only existing schema escape valve, `fqdn`, moves the **entire** URL host (connect + Host + SNI together) — it does **not** decouple `Host` from the verification host.
- The decoupling primitive already exists in the pinned stack: an explicit `Host` header overrides the URL-derived one, and `request(..., extensions={"sni_hostname": …})` independently drives `server_hostname` (verified live against `httpx==0.28.1` / `httpcore==1.0.9`).

## Desired state

Add an optional per-target **`tls_server_name: str | None`** (the SNI + cert-CN/SAN verification name), decoupled from `host` (connect address) and the wire `Host` header. When set, dispatch keeps `base_url=https://{target.host}` (connect=IP, `Host: <IP>` → appliance 200) and threads `extensions={"sni_hostname": target.tls_server_name}` into every `client.request(...)`/`client.get(...)` in `HttpConnector` (the two request helpers + the vcf_logs unauthenticated fingerprint `client.get`). This lets an operator keep `verify_tls=true` + CA-pin (cert-CN FQDN as the SNI/verify name) while sending `Host: <IP>` — the config the appliance accepts — instead of the `verify_tls=false` downgrade.

[...]

PR #2657, opened 2026-07-21
fix(cli): map tls_server_name to the targets-import column set (#2643)

### LINKED ISSUE #2643: Task: add tls_server_name to the meho targets import column map — CLI parity with TargetCreate/TargetUpdate
Parent initiative: #2656 (G0.37 v0.25.0 ops-feedback hardening)
Parent goal: #221

## Summary

`meho targets import` cannot set `tls_server_name`: the CLI's recognised-column set is strictly narrower than the REST `TargetCreate`/`TargetUpdate` models, so a manifest's `tls_server_name` spills to `extras` instead of the real column. The import reports "1 updated", the manifest looks complete, and the target stays misconfigured — dispatch then fails at TLS (`connector_tls_verify_failed`, "IP address mismatch") for exactly the FQDN-SAN-appliance-reached-by-IP class the #2398/#2604 SNI fix just unblocked. Add `tls_server_name` to the column map with the same treatment `verify_tls`/`tls_ca_pin` got in #1780/#1784.

## Problem

- `cli/internal/cmd/targets/import.go:71-86` — `knownTopLevel` maps `aliases, auth_model, extras, fqdn, host, name, notes, port, preferred_impl_id, product, secret_ref, tls_ca_pin, verify_tls, vpn_required`. **No `tls_server_name`.**
- The generated REST client (`cli/internal/api/client.gen.go`) carries `TlsServerName *string \`json:"tls_server_name"\`` on the target models — the field is right there in the OpenAPI; the CLI mapping is the only gap.
- Observed in the field (SDDC Manager reached by IP, FQDN-only SAN): `verify_tls` + `tls_ca_pin` applied, `tls_server_name: null` after "1 updated"; `PATCH /api/v1/targets/{name}` with the same value works.
- Failure is silent: no warning, no error — the worst kind of gap for the *documented, repo-tracked* way to register targets.

## Approach

Add `"tls_server_name"` to `knownTopLevel` in `import.go` so the manifest key goes top-level on both CREATE and UPDATE bodies — mirroring the Initiative-#1774-amendment treatment of `verify_tls` (#1780) and `tls_ca_pin` (#1784), including the existing test patterns in `cli/internal/cmd/targets/import_test.go:183-265` (top-level on create, top-level on update, never leaked into `extras`).

## Acceptance criteria

- [ ] A `targets.yaml` entry with `tls_server_name` produces a CREATE body carrying it top-level (test mirroring `import_test.go:183-207`).
- [ ] `meho targets import --update` on an existing target sends `tls_server_name` top-level (test mirroring `import_test.go:219-233`).
- [ ] `tls_server_name` never appears in `extras` (test mirroring `import_test.go:245-265`).
- [ ] CHANGELOG `[Unreleased]` entry under a unique `###` header.

## Out of scope

- Any backend/OpenAPI change — the REST models already expose the field.
- Import-plan diff display polish or other missing-column audits (none known; this is the only `TargetUpdate` field absent from the map that a manifest can meaningfully carry).

## References

- Original ops report: this issue's pre-rewrite body (edit history) — VCF SDDC Manager, FQDN-only SAN leaf cert, reached by IP.
- `cli/internal/cmd/targets/import.go:71-86` (`knownTopLevel`); `cli/internal/cmd/targets/import_test.go:183-265` (the #1780/#1784 test mold to mirror).
- SNI-on-session-establish fix this completes the tooling story for: #2398 / PR #2604 (shipped v0.25.0).

damir-topic:
## Summary

- `meho targets import` silently dropped `tls_server_name`: the CLI's `knownTopLevel` set never grew the key when #2002 added the column to `TargetCreate` / `TargetUpdate`, so a manifest key spilled into the `extras` JSONB blob. The import printed "1 updated", the descriptor looked complete, and the typed column stayed `NULL`.
- Dispatch then kept deriving the SNI / cert-verification hostname from `host`, so an appliance reached by IP with an FQDN-only SAN failed with `connector_tls_verify_failed` ("IP address mismatch") — the exact class the #2398 SNI fix shipped to unblock. Operators had to work around it with a manual `PATCH /api/v1/targets/{name}`, which the repo-tracked `targets.yaml` then couldn't express.
- Fix is one map entry, given the same treatment `verify_tls` (#1780) and `tls_ca_pin` (#1784) got: the key reaches the body root on both the POST (create) and the sparse PATCH (`--update`) shapes, and an explicit `tls_server_name: null` still clears the override.
- No client regeneration and no backend change: the generated REST client already carries `TlsServerName *string` on `TargetCreate` and `TargetUpdate`, and `backend/src/meho_backplane/targets/schemas.py` has had the column since #2002. The CLI map was the only gap.

## Test plan

- [x] `go vet ./...` — clean
- [x] `gofmt -l ./internal/` — no output
- [x] `go test ./...` (whole CLI module) — all packages ok
- [x] AC 1 — CREATE body carries it top-level: `TestMapEntryTLSServerNameIsTopLevel` (mirrors `import_test.go` `TestMapEntryTLSTrustKeysAreTopLevel`)
- [x] AC 2 — `--update` sends it top-level: `TestEntryToUpdateBodyTLSServerNameIsTopLevel` (mirrors `TestEntryToUpdateBodyTLSTrustKeysAreTopLevel`; also pins the explicit-null clear)
- [x] AC 3 — never leaks into `extras`, unknown keys still spill: `TestMapEntryTLSServerNameStillSpillsUnknown` (mirrors `TestMapEntryTLSTrustTopLevelStillSpillsUnknown`)
- [x] AC 4 — CHANGELOG `[Unreleased]` entry under the unique header `### Fixed — CLI targets import tls_server_name (#2643)`

Local `golangci-lint` could not run (`the Go language version (go1.24) used to build golangci-lint is lower than the targeted Go version (1.25)` — a stale local binary, not a code finding); the `Go (golangci-lint + go test)` CI lane runs the pinned version. The diff adds no production code paths, and `_test.go` files are already excluded from `revive` / `errcheck` by `cli/.golangci.yml`.

## Docs

- `cli/internal/cmd/targets/import.go` — the verb's `--help` mapping line now lists `tls_server_name`.
- `docs/codebase/cli.md` — mapping-rules list + a bullet explaining why all three TLS columns must go top-level.
- `docs/cross-repo/targets-yaml.md` — the "Mapping rules at a glance" table had never recorded `verify_tls` / `tls_ca_pin` either; one row now covers all three.

## Out of scope

- **`version` is a second gap of the same class.** `TargetCreate` / `TargetUpdate` both carry the operator-asserted `version` column (G0.15-T6 #1215 — it flips the connector resolver from the wildcard fallback to the versioned connector), and it is *also* absent from `knownTopLevel`, so a manifest's `version:` spills into `extras` just as silently. This contradicts the issue body's "this is the only `TargetUpdate` field absent from the map" line. Left untouched here to keep the fix matching its ticket; worth a one-line follow-up issue.
- **The Python-side mirror lags.** `backend/tests/test_api_v1_targets_import.py:_KNOWN_TOP_LEVEL` is documented as the mirror of `knownTopLevel` but never grew the #1780/#1784 columns either. It is inert today (the pinned consumer fixture carries none of the TLS columns, verified by grep), and it is not an asserted equality check — so it does not detect this drift class. Untouched here per the "no backend change" scope; a real drift guard would be a better fix than adding a fourth key by hand.
- Import-plan diff display polish; any other missing-column audit.

Closes #2643

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `meho targets import` so `tls_server_name` is correctly applied to targets instead of being stored as extra data.
  - Prevented TLS verification failures caused by using an IP address when the certificate requires a hostname.
  - Preserved support for clearing `tls_server_name` and storing unrelated unknown fields as extras.

- **Documentation**
  - Clarified YAML import behavior for TLS settings, including certificate hostname and SNI configuration.

### DIFF
--- a/docs/codebase/cli.md
+++ b/docs/codebase/cli.md
@@ -2027,11 +2027,14 @@ output), `--backplane` (override the configured backplane URL).
 - **Known top-level columns** map 1:1 to the API's `TargetCreate` /
   `TargetUpdate` body fields: `name`, `aliases`, `product`, `host`,
   `port`, `fqdn`, `secret_ref`, `auth_model`, `vpn_required`,
-  `notes`, `preferred_impl_id`. The list in
+  `notes`, `preferred_impl_id`, `verify_tls`, `tls_ca_pin`,
+  `tls_server_name`. The list in
   `knownTopLevel` is the canonical reference; the
   Python-side mirror lives in
-  `backend/tests/test_api_v1_targets_import.py:_KNOWN_TOP_LEVEL` and
-  keeps drift detectable in CI.
+  `backend/tests/test_api_v1_targets_import.py:_KNOWN_TOP_LEVEL`,
+  which pins the pre-#1774 core set the consumer fixture exercises
+  (it carries none of the TLS columns, so the two sets agree on
+  every key that fixture can produce).
 - **`fingerprint`** is dropped silently with a warning log line.
   Server-managed per the G0.3-T1.5 (#477) amendment — the probe
   verb is the only legitimate writer, and the API rejects
@@ -2042,6 +2045,15 @@ output), `--backplane` (override the configured backplane URL).
 - **`preferred_impl_id`** is a real top-level column post-#477.
   Sent at the body root, not spilled into extras — the G0.6 #388
   resolver's tie-break ladder reads it.
+- **`verify_tls` / `tls_ca_pin` / `tls_server_name`** are the
+  per-target TLS columns (Initiative #1774's #1780 / #1784, plus
+  #2002's SNI override). All three go at the body root. Spilling any
+  of them into `extras` is silent misconfiguration: the typed column
+  keeps its default, the import still prints "1 updated", and the
+  operator only finds out at dispatch time — `tls_server_name`
+  specifically surfaces as `connector_tls_verify_failed` /
+  "IP address mismatch" on an appliance reached by IP whose leaf cert
+  carries an FQDN-only SAN (#2643).
 - **Every other key** spills into the `extras` JSONB column.
   Explicit `extras:` blocks in the YAML merge with spilled keys
   rather than overwriting them.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,20 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Fixed — CLI targets import tls_server_name (#2643)
+
+- `meho targets import` now maps a manifest's `tls_server_name` to the
+  real top-level column on `TargetCreate` / `TargetUpdate` instead of
+  spilling it into the `extras` JSONB blob. Before this fix the import
+  reported "1 updated", the descriptor looked complete, and the typed
+  column stayed `NULL` — so dispatch kept deriving the SNI /
+  cert-verification hostname from `host` and an appliance reached by IP
+  with an FQDN-only SAN failed with `connector_tls_verify_failed`
+  ("IP address mismatch"), the exact class the #2398 SNI fix shipped to
+  unblock. Operators who worked around it with `PATCH
+  /api/v1/targets/{name}` can drop the manual step and let the
+  repo-tracked `targets.yaml` be the source of truth again.
+
 ## [0.25.0] - 2026-07-19
 
 This release lands the complete **v0.21/v0.22 dogfood-hardening cycle** —
--- a/cli/internal/cmd/targets/import.go
+++ b/cli/internal/cmd/targets/import.go
@@ -68,6 +68,17 @@ type importDoc struct {
 // the descriptor. The server enforces their mutual exclusivity
 // (`verify_tls=false` ⊕ `tls_ca_pin`) and PEM validation with a 422,
 // which the import path surfaces via the existing error handling.
+//
+// #2002 amendment: `tls_server_name` is the third first-class TLS
+// column (per-target SNI / cert-verification hostname, decoupled from
+// `host`). It belongs in the same bucket for the same reason —
+// spilling it into `extras` leaves the typed column NULL, so dispatch
+// keeps deriving the SNI from `host` and an appliance reached by IP
+// with an FQDN-only SAN fails verification with
+// `connector_tls_verify_failed` even though the descriptor named the
+// right hostname. The column composes with `verify_tls` / `tls_ca_pin`
+// (no mutual-exclusion validator); the server caps it at 512 chars and
+// returns 422 on overflow, which the import path already surfaces.
 var knownTopLevel = map[string]struct{}{
 	"aliases":           {},
 	"auth_model":        {},
@@ -81,6 +92,7 @@ var knownTopLevel = map[string]struct{}{
 	"product":           {},
 	"secret_ref":        {},
 	"tls_ca_pin":        {},
+	"tls_server_name":   {},
 	"verify_tls":        {},
 	"vpn_required":      {},
 }
@@ -179,7 +191,8 @@ func newImportCmd() *cobra.Command {
 			"Mapping rules. Top-level columns recognised on the API's " +
 			"TargetCreate / TargetUpdate models are mapped 1:1: name, aliases, " +
 			"product, host, port, fqdn, secret_ref, auth_model, vpn_required, " +
-			"notes, preferred_impl_id, verify_tls, tls_ca_pin. Any other field " +
+			"notes, preferred_impl_id, verify_tls, tls_ca_pin, tls_server_name. " +
+			"Any other field " +
 			"is spilled into the `extras` JSONB column. `fingerprint` is " +
 			"server-managed and skipped with a warning if present in the YAML " +
 			"(the probe verb is the only legitimate writer).\n\n" +
--- a/cli/internal/cmd/targets/import_test.go
+++ b/cli/internal/cmd/targets/import_test.go
@@ -266,6 +266,99 @@ func TestMapEntryTLSTrustTopLevelStillSpillsUnknown(t *testing.T) {
 	}
 }
 
+func TestMapEntryTLSServerNameIsTopLevel(t *testing.T) {
+	t.Parallel()
+	// #2002 shipped tls_server_name as a first-class TargetCreate /
+	// TargetUpdate column: the per-target SNI / cert-verification
+	// hostname, decoupled from `host`. It must land in the body root,
+	// NOT in extras — spilling leaves the typed column NULL, dispatch
+	// keeps deriving the SNI from `host`, and an appliance reached by
+	// IP with an FQDN-only SAN fails with connector_tls_verify_failed
+	// despite the descriptor naming the right hostname (#2643).
+	body, warnings := mapEntry(map[string]any{
+		"name":            "vcf-sddc-manager",
+		"product":         "vmware-rest",
+		"host":            "10.20.30.40",
+		"tls_server_name": "sddc-manager.evba.lab",
+	})
+	if len(warnings) != 0 {
+		t.Errorf("warnings: got %v; want none", warnings)
+	}
+	if v, ok := body["tls_server_name"]; !ok || v != "sddc-manager.evba.lab" {
+		t.Errorf("tls_server_name top-level: got %v (present=%t); want sddc-manager.evba.lab", v, ok)
+	}
+	if extras, ok := body["extras"].(map[string]any); ok {
+		if _, leaked := extras["tls_server_name"]; leaked {
+			t.Errorf("tls_server_name leaked into extras: %v", extras)
+		}
+	}
+}
+
+func TestEntryToUpdateBodyTLSServerNameIsTopLevel(t *testing.T) {
+	t.Parallel()
+	// `meho targets import --update` routes through entryToUpdateBody.
+	// tls_server_name must reach the sparse PATCH body's top level so a
+	// descriptor re-import flips the typed column; an explicit null
+	// clears the override and returns the target to deriving the SNI
+	// from `host` (TargetUpdate's documented clear semantics).
+	body, _ := entryToUpdateBody(map[string]any{
+		"name":            "vcf-sddc-manager",
+		"tls_server_name": "sddc-manager.evba.lab",
+	})
+	if v, ok := body["tls_server_name"]; !ok || v != "sddc-manager.evba.lab" {
+		t.Errorf("tls_server_name top-level on update: got %v (present=%t); want sddc-manager.evba.lab", v, ok)
+	}
+	if extras, ok := body["extras"].(map[string]any); ok {
+		if _, leaked := extras["tls_server_name"]; leaked {
+			t.Errorf("tls_server_name leaked into extras on update: %v", extras)
+		}
+	}
+
+	cleared, _ := entryToUpdateBody(map[string]any{
+		"name":            "vcf-sddc-manager",
+		"tls_server_name": nil, // clearing the override is an explicit-null PATCH
+	})
+	v, ok := cleared["tls_server_name"]
+	if !ok {
+		t.Errorf("tls_server_name missing from update body; want top-level null: %v", cleared)
+	}
+	if v != nil {
+		t.Errorf("tls_server_name cleared value: got %v; want nil", v)
+	}
+}
+
+func TestMapEntryTLSServerNameStillSpillsUnknown(t *testing.T) {
+	t.Parallel()
+	// No-regression guard: adding tls_server_name to knownTopLevel must
+	// not change the extras-spill behaviour for genuinely-unknown keys,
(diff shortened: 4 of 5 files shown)


docs/codebase/cli.md
```diff
@@ -2027,11 +2027,14 @@ output), `--backplane` (override the configured backplane URL).
...
+  `notes`, `preferred_impl_id`, `verify_tls`, `tls_ca_pin`,
+  `tls_server_name`. The list in
   `knownTopLevel` is the canonical reference; the
   Python-side mirror lives in
-  `backend/tests/test_api_v1_targets_import.py:_KNOWN_TOP_LEVEL` and
-  keeps drift detectable in CI.
+  `backend/tests/test_api_v1_targets_import.py:_KNOWN_TOP_LEVEL`,
+  which pins the pre-#1774 core set the consumer fixture exercises
+  (it carries none of the TLS columns, so the two sets agree on
+  every key that fixture can produce).
```
damir-topic, line 2037: **[⚠️ M1]** — this names the inert *test* set as "the Python-side mirror"; the production Python port of `knownTopLevel` is somewhere else and is still out of parity

**What:** The rewritten bullet points the reader at `backend/tests/test_api_v1_targets_import.py:_KNOWN_TOP_LEVEL` as *the* Python-side mirror. The claim it makes about that file is true (I verified the set and the fixture), but it is not the only — or the important — Python copy. The production port is `backend/src/meho_backplane/ui/routes/connectors/import_view.py:_KNOWN_TOP_LEVEL`, whose own docstring says "Verbatim parity with ``knownTopLevel`` in ``cli/internal/cmd/targets/import.go`` so the web import and the CLI import spill the same keys into ``extras``". That set stops at `preferred_impl_id`: it never grew `verify_tls` / `tls_ca_pin` at #1780/#1784, and it does not have `tls_server_name` after this PR.

**Why it matters:** `docs/codebase/cli.md` is the doc the next person hits when they ask "where else does this key set live?" — which is exactly the question #2643 exists because someone answered wrong. As written, the bullet answers it with the one copy that is inert (not an asserted equality check, and the pinned fixture carries no TLS keys) and stays silent about the copy that serves live traffic. Concretely: after this PR, an operator who pastes the same `targets.yaml` into the `/ui` bulk importer still gets `tls_server_name` spilled into `extras` and the typed column left `NULL` — the identical silent misconfiguration, on the surface `docs/codebase/ui.md` advertises as producing "byte-identical writes for the same YAML".

**Required change:** Docs-only, in the paragraph this PR already rewrites — do **not** change `import_view.py` here, that is out of scope per the ticket. Extend the bullet along these lines:

```suggestion
  every key that fixture can produce). The *production* Python port
  lives in
  `backend/src/meho_backplane/ui/routes/connectors/import_view.py:_KNOWN_TOP_LEVEL`
  (the `/ui` bulk importer, docs/codebase/ui.md "Bulk import"), which
  claims verbatim parity with `knownTopLevel` but has drifted: it
  carries none of `verify_tls` / `tls_ca_pin` / `tls_server_name`, so
  the web import still spills all three into `extras`. Closing that
  gap is follow-up work, not part of #2643.
```

A follow-up issue for the code fix (plus the stale mapping-rules list in `docs/codebase/ui.md`) would be the right companion — I've recorded it under "Adjacent improvements" in the review summary so it isn't lost.

**Source:** `backend/src/meho_backplane/ui/routes/connectors/import_view.py:92-110` (the parity docstring and the set) and `docs/codebase/ui.md:2354` ("Mapping rules (parity with `import.go`)").

damir-topic (commented):
> **Note:** GitHub refuses `APPROVE` on a self-authored PR (`Can not approve your own pull request`), so this is posted as a `COMMENTED` review. The binding verdict below is **APPROVE** — the machine-readable result file records it as such.

**Anchor task:** #2643 — add `tls_server_name` to the `meho targets import` column map (CLI parity with `TargetCreate`/`TargetUpdate`)
**Surface area:** cli (Go) + docs
**Best-practices packs audited:** devops (CLI/Go slice), plus repo-local Go lint config `cli/.golangci.yml`
**Verdict:** APPROVE
**Head SHA reviewed:** `fe369a49`

### Acceptance criteria

- [x] **AC1 — CREATE body carries `tls_server_name` top-level.** `TestMapEntryTLSServerNameIsTopLevel` asserts body-root presence and non-leak. Verified it is a *real* guard: with the one-line map entry reverted the test fails (`got <nil> (present=false)` + `leaked into extras`).
- [x] **AC2 — `--update` sends it top-level.** `TestEntryToUpdateBodyTLSServerNameIsTopLevel` covers both the set case and the explicit-null clear. Also fails against pre-fix code (`missing from update body ... map[extras:map[tls_server_name:<nil>]]`).
- [x] **AC3 — never appears in `extras`.** `TestMapEntryTLSServerNameStillSpillsUnknown` pins the composition with `verify_tls` and keeps `sso_realm` spilling. Fails pre-fix.
- [x] **AC4 — CHANGELOG `[Unreleased]` entry under a unique `###` header.** `### Fixed — CLI targets import tls_server_name (#2643)`; no collision with any other open PR's `[Unreleased]` header.

All three tests were executed against a locally reverted `knownTopLevel` to confirm none of them passes both before and after the fix.

### Project-board alignment

|---|---|---|---|---|---|
| #2643 | open (`bug`,`task`,`go`,`priority:medium`,`effort:small`) | `Closes #2643` | ✅ | ✅ full | close on merge |
| #2656 (Initiative G0.37) | open | parent-of | — | partial (1 of 4 children) | stays open |
| #221 (Goal) | open | grandparent | — | partial | stays open |

No other open PR touches `cli/internal/cmd/targets/`, `docs/codebase/cli.md`, or `docs/cross-repo/targets-yaml.md`. No delivered-but-unlinked issue.

### Predicted vs. actual diff

Predicted before re-reading the diff: `knownTopLevel` entry; the `Long` help mapping line (it enumerates the same set); 3 tests mirroring the `#1780/#1784` mold; CHANGELOG; `docs/codebase/cli.md` mapping-rules list; `docs/cross-repo/targets-yaml.md` mapping table. No client regeneration (field already generated), no OpenAPI snapshot change, no migration.

- ✅ `cli/internal/cmd/targets/import.go` — map entry + `--help` mapping line + rationale comment
- ✅ `cli/internal/cmd/targets/import_test.go` — 3 tests, faithful to the `import_test.go:183-265` mold
- ✅ `CHANGELOG.md`, `docs/codebase/cli.md`, `docs/cross-repo/targets-yaml.md`
- ❓ Unexpected but justified: none — the diff is exactly the predicted set
- 🚫 Unexpected and undisclosed: none
- 🕳️ Missing: none in the CLI surface. One adjacent production surface *is* out of parity — see M1 and "Adjacent improvements"; it is scoped out by the ticket's "no backend change" line, so it is not counted as missing here.

### Integration verification

- [x] `go vet ./...` — clean
- [x] `gofmt -l ./internal/` — no output
- [x] `go test ./...` (whole CLI module) — every package ok
- [x] **Negative control** — reverted the `"tls_server_name": {}` entry locally, re-ran the 3 new tests: all 3 FAIL. Restored; tree clean.
- [x] **End-to-end wire-body check (temporary test, removed after the run).** Drove `runImport --update` against an `httptest` backplane with a two-entry YAML. Observed on the wire:
  - `PATCH /api/v1/targets/rdc-vault` → `{"host":"10.20.30.40","tls_server_name":"vault.evba.lab","verify_tls":true}` — top-level, and `name`/`product` still stripped.
  - `POST /api/v1/targets` (explicit-null entry) → `{"extras":{"sso_realm":"evba.lab"},"host":"10.20.30.41","name":"new-appliance","product":"vault","tls_server_name":null}` — the clear reaches the body root, the unknown key still spills.
- [x] `meho targets import --help` rendered — the concatenated mapping sentence reads correctly ("… `verify_tls`, `tls_ca_pin`, `tls_server_name`. Any other field is spilled into the `extras` JSONB column.").
- [x] `golangci-lint` — the author could not run it locally (stale go1.24 binary vs `run.go: "1.25"`), but the required **`Go (golangci-lint + go test)`** lane is green on this head. Reviewed the diff against `cli/.golangci.yml` independently: the enabled set is `errcheck / govet / ineffassign / staticcheck / unused / revive` + `gofmt / goimports`; `_test.go` is excluded from `errcheck` and `revive`; nothing in the added code introduces an unchecked error, an ineffectual assignment, an unused symbol, or a formatting deviation.
- [x] `CLI API snapshot freshness` lane — green; no route/schema change, so no regen was owed.
- ⚠️ not run — migrations replay (no migration), `docker compose up`, UI flow, agent eval. Not applicable to this diff.
- Backend `Python (ruff + mypy + pytest)` / `Python (integration testcontainers)` lanes were still IN_PROGRESS at review time; the diff contains **zero** Python changes, so neither is a correctness gate for this PR.

### Cross-checks against the real contracts

- Generated client: `cli/internal/api/client.gen.go` carries `TlsServerName *string \`json:"tls_server_name"\`` on the target create/update/read models — the YAML key name the map now recognises matches the wire name exactly. No regeneration owed.
- Backend schema: `backend/src/meho_backplane/targets/schemas.py` — `TargetCreate.tls_server_name: str | None = Field(default=None, max_length=512)` and the identical field on `TargetUpdate`, with the documented "explicit null clears the override" PATCH semantics. `update_target` uses `model_dump(exclude_unset=True)`, so the sparse body the CLI emits touches only the listed columns — which is what makes the explicit-null clear work rather than being swallowed.
- The `--help` mapping list, `docs/codebase/cli.md`'s known-column list, and `docs/cross-repo/targets-yaml.md`'s mapping table now all enumerate the same 3 TLS columns. The `targets-yaml.md` row also back-fills `verify_tls` / `tls_ca_pin`, which that table had never recorded — accurate, and in-scope as part of the same row.
- The `cli.md` claim about `backend/tests/test_api_v1_targets_import.py:_KNOWN_TOP_LEVEL` was verified: that set really does stop at `preferred_impl_id`, and the pinned consumer fixture contains no `verify_tls` / `tls_ca_pin` / `tls_server_name` key, so the two sets do agree on every key that fixture can produce. The claim is true as written.

- ⚠️ M1 — `docs/codebase/cli.md` names the inert *test* set as "the Python-side mirror" while the production Python port of `knownTopLevel` lives elsewhere and is still out of parity; required change: correct that bullet to name `backend/src/meho_backplane/ui/routes/connectors/import_view.py:_KNOWN_TOP_LEVEL` and record the drift.

No blockers. No minors. No open questions.

Items for follow-up issues — none of these block this PR or require a response here.

1. **`/ui` bulk import has the same silent-spill bug, for all three TLS columns.** `backend/src/meho_backplane/ui/routes/connectors/import_view.py:_KNOWN_TOP_LEVEL` is documented as "Verbatim parity with `knownTopLevel`" and `docs/codebase/ui.md`'s "Bulk import (Task #875)" section states the web import and the CLI import "produce byte-identical writes for the same YAML". That set stops at `preferred_impl_id` — it never grew `verify_tls` / `tls_ca_pin` (drifted at #1780/#1784) and does not have `tls_server_name` now. An operator who pastes the *same* `targets.yaml` into the web importer still gets the silent `extras` spill this PR fixes for the CLI. Correctly out of scope here (the ticket's "Out of scope" pins "any backend change"), but it is the highest-value follow-up in this class and deserves its own issue together with the `docs/codebase/ui.md` mapping-rules list.
2. **`version` is a second missing key of the same class**, as the PR description already calls out — absent from `knownTopLevel` despite being on `TargetCreate` / `TargetUpdate`. Worth the one-line follow-up the author suggests.
3. **A real drift guard beats hand-maintained mirrors.** Three copies of this key set now exist (Go production, Python production port, Python test set) and two have already drifted independently. A test that reads the authoritative set out of the OpenAPI schema (or asserts the three sets equal) would have caught #2643, the `version` gap, and the `import_view.py` gap in one go — a better fix than adding a fourth hand-written key.

### Docs

- [x] `docs/codebase/cli.md` — updated (mapping-rules list + a dedicated TLS-columns bullet explaining why spilling is silent misconfiguration). Accurate, with the one wording correction in M1.
- [x] `docs/cross-repo/targets-yaml.md` — updated ("Mapping rules at a glance" now covers all three TLS columns).
- [ ] `docs/codebase/ui.md` — not updated, and not required by this diff: its "parity with `import.go`" list was already stale before this PR. Tracked as Adjacent improvement 1.
- No .NET analogies introduced.

- The three tests are genuine guards, not decoration. I verified each one fails against the pre-fix `knownTopLevel`, which is the property that actually matters for a one-line map fix and the one that is easiest to get wrong.
- The explicit-null clear case was tested even though the ticket didn't ask for it — and it is the case that would have silently regressed if the key had been routed through a typed body instead of the sparse map.
- The PR description does the honest thing: it names the `version` gap it deliberately did not fix and explicitly contradicts the issue body's "this is the only field absent from the map" claim, rather than quietly widening scope or quietly leaving it unsaid.

Issue #2656, opened 2026-07-21
Initiative: G0.37 v0.25.0 ops-feedback hardening — agent-run schema publish, GSM-WIF background dispatch, CLI TLS import parity, scheduler-token diagnostics


damir-topic:
Parent goal: #221

## Outcome

The four ops reports from the first days on the v0.25.0 per-operator-WIF GSM deploy (2026-07-19/20) are closed out:

- Agent runs start again on the Anthropic provider — the top-level JSON-Schema union on the published `meho.broadcast.watch` tool schema (which the hosted-agent bridge forwards verbatim and the Anthropic Messages API rejects at model-init) is stripped, with a sweep test so no published tool schema can regress.
- Target-bound Sensors evaluate on per-operator-WIF GSM deploys — background dispatch gets a real credential identity instead of fail-closing on the empty-JWT guard.
- `meho targets import` can fully secure an FQDN-SAN appliance reached by IP — `tls_server_name` reaches the real column instead of spilling to extras.
- A dead scheduler Vault token is reported as what it is ("re-mint the token") instead of masquerading as a policy-scope error that sends the operator to widen a policy that is already correct.

## Why it matters

This is the v0.25.0 iteration of the per-cycle ops-feedback hardening series (G0.30 #2286 → G0.32 #2494 / G0.33 #2495 → G0.36 #2592). Severity is front-loaded: #2644 is a total outage of the agent layer (every run dies at `turns: 0` before the first model turn, including scheduled agents), and #2642 makes the entire Sensors feature (#2416) non-functional for credentialed targets on the reporting deploy class — the scheduled path is precisely the one with no operator JWT. The other two are small, sharp diagnosability/parity gaps that cost operators real time.

## Children

- [ ] #2644 — strip top-level JSON-Schema unions from published tool schemas (agent layer down; `tools.5` = broadcast.watch)
- [ ] #2642 — give background dispatch a credential identity on per-operator-WIF GSM deploys (Sensors non-functional)
- [ ] #2643 — add `tls_server_name` to the `meho targets import` column map (CLI narrower than the REST model)
- [ ] #2652 — disambiguate a dead scheduler Vault token from an under-scoped policy (lookup-self before the policy message)
- [ ] #2702 — surface safety-class changes to Sensor-backed ops at connector re-ingest (2026-07-28 test-campaign finding; the platform-level half of the create-time-only trade-off documented in `checks/runner.py`)

## Definition of done

[...]

PR #2658, opened 2026-07-21
fix(scheduler): split a dead Vault token from an under-scoped policy (#2652)

### LINKED ISSUE #2652: Task: disambiguate a dead scheduler Vault token from an under-scoped policy — lookup-self on the write-failure path
Parent initiative: #2656 (G0.37 v0.25.0 ops-feedback hardening)
Parent goal: #221

## Summary

When the scheduler's Vault token is dead (revoked / lost lease), agent-credential writes fail with a message that names the **wrong remediation**: "scheduler Vault write failed — VAULT_SCHEDULER_TOKEN policy must grant create/update on the agent-credentials path". A dead token and an under-scoped policy both surface from Vault as a 403, but the two fixes are completely different (re-mint the token vs. widen the policy) — and the broker **already has** the machinery to tell them apart: `verify_scheduler_token` / `lookup-self` shipped in #2328. Wire a validity check into the write-failure path and emit a distinct dead-token message on every surface.

## Problem

- `backend/src/meho_backplane/scheduler/vault_credentials.py:113` — `SchedulerVaultBrokerError` is raised for any failed write; the broker's `lookup-self` capability (`:400-402` `_lookup_self_blocking`, plus `verify_scheduler_token` per the module docstring `:50-67`, #2328) is **not consulted** on the failure path.
- Four surfaces map that one exception to the policy-only remediation:
  - MCP: `mcp/tools/agent_principals.py:173-176` ("policy must grant create/update…").
  - REST: `api/v1/agent_principals.py:211` and `api/v1/runner_principals.py:204` (`scheduler_vault_write_error`, 502).
  - UI: `ui/routes/agents/principals_forms.py:272` (banner `scheduler_vault_write_error`).
- `docs/cross-repo/vault-provisioning.md` reinforces the policy framing ("Under-scoped surface 6 … widen the meho-scheduler policy") with no dead-token entry.
- Field report: policy, path pattern, and mount were all verified correct before anyone thought to `vault token lookup` the token itself — the message actively cost the operator the diagnosis.

## Approach

On a failed scheduler Vault write, run the existing `lookup-self` check on the broker's token and split the failure into two distinguishable cases carried by the exception (e.g. a `token_invalid` flag or subclass):

- **Token invalid/expired** → "scheduler token invalid or expired — re-mint VAULT_SCHEDULER_TOKEN and update the deployment secret".
- **Token live, write denied** → keep today's policy-scope message.

Fix at the broker level so all four surfaces (MCP / 2× REST / UI banner) inherit the distinction from the exception rather than re-implementing it. Add the dead-token row next to the under-scoped-policy row in `docs/cross-repo/vault-provisioning.md`.

## Acceptance criteria

- [ ] With a stubbed Vault where `lookup-self` fails and the write 403s, the raised error carries the token-invalid disposition and the MCP tool's `-32602` message says "invalid or expired … re-mint" (test).
- [ ] With a stubbed Vault where `lookup-self` succeeds and the write 403s, the existing policy-scope message is preserved verbatim (test).
- [ ] REST (`agent_principals`, `runner_principals`) and the UI banner surface the same two-way distinction (tests or shared-message assertion).
- [ ] `docs/cross-repo/vault-provisioning.md` documents the dead-token failure mode and its remediation alongside the under-scoped-policy row.
- [ ] CHANGELOG `[Unreleased]` entry under a unique `###` header.

## Out of scope

- The consumer-side periodic-token lifecycle that keeps killing the token (RDC-side operational issue, separate).
- Token auto-renewal changes — renew-on-use (#2328) is untouched; this is diagnosis-only.
- Retrying the write after a failed lookup-self (fail fast with the right message; no new retry semantics).

## References

- Original ops report: this issue's pre-rewrite body (edit history) — `meho_agent_principals_register` on a deploy whose policy/path/mount were all verified correct; `vault token lookup` on the stored token errored.
- `backend/src/meho_backplane/scheduler/vault_credentials.py:113` (`SchedulerVaultBrokerError`), `:400-402` (`_lookup_self_blocking`), docstring `:50-67` (#2328 renew-on-use + self-lookup).
- Surfaces: `mcp/tools/agent_principals.py:173-176`, `api/v1/agent_principals.py:211`, `api/v1/runner_principals.py:204`, `ui/routes/agents/principals_forms.py:272`.
- Vault API: `auth/token/lookup-self` (hvac `client.auth.token.lookup_self()` — already in use at `:402`).

damir-topic:
## Summary

- A dead scheduler Vault token (revoked / expired / lost periodic lease) and an
  under-scoped `meho-scheduler` policy both make Vault answer the
  agent-credential write with a **403**, but every register surface named only
  the policy remediation. The field report in #2652 had the operator verify
  policy, path pattern, and mount — all correct — before anyone thought to
  `vault token lookup` the token itself.
- `write_agent_secret` now fires the `auth/token/lookup-self` probe the broker
  already shipped for #2328 when Vault rejects the write, on the **same** hvac
  client that was denied, and stamps the outcome as `token_invalid` on
  `SchedulerVaultBrokerError`. Vault answers `lookup-self` for any live token
  regardless of policy and 403s an invalid one, so the two 403s become
  distinguishable.
- Fixed **at the broker**, so all four surfaces inherit one diagnosis instead of
  re-implementing it: the MCP `meho.agent_principals.register` `-32602` message,
  both REST register routes' 502 `detail`
  (`api/v1/agent_principals.py`, `api/v1/runner_principals.py`), and the
  `/ui/agents/principals` register banner. A dead token gets the three-clause
  `scheduler_vault_token_invalid: … invalid or expired … Re-mint
  VAULT_SCHEDULER_TOKEN …` detail; a live-token-denied write keeps today's
  policy-scope wording **verbatim**.
- Diagnosis only, per the issue's out-of-scope list: no retry after a failed
  lookup, renew-on-use (#2328) untouched, consumer-side token lifecycle
  untouched. A transport failure on the probe is not treated as evidence of a
  dead token (stays conservative → policy message), and the
  already-unambiguous unreachable-Vault arm does not probe at all.
- `docs/cross-repo/vault-provisioning.md` gains the dead-token row alongside the
  under-scoped-policy row (with a `vault token lookup` confirmation step and the
  re-mint command), plus a "when the fuse has already blown" note under the
  periodic-token section; `docs/codebase/scheduler.md` and `docs/codebase/ui.md`
  record the new split.

## Design note

The probe reuses `_lookup_self_blocking` against the client that just failed
rather than calling `verify_scheduler_token()`, which re-resolves the token from
`VAULT_SCHEDULER_TOKEN` / `VAULT_SCHEDULER_TOKEN_FILE`. The answer must describe
the identity Vault actually denied, and re-resolution could pick up a token
re-minted between the two calls.

## Test plan

- [x] `ruff check` / `ruff format --check` on `backend/src` + `backend/tests` — clean
- [x] `mypy src/meho_backplane --ignore-missing-imports` — 680 files, clean
- [x] Full backend unit sweep (`pytest -n 4 --dist loadscope --ignore=tests/integration --ignore=tests/migrations tests/`)
- [x] **AC1** — stubbed Vault, `lookup-self` fails + write 403s → `token_invalid is True`
  (`test_write_denied_with_dead_token_sets_token_invalid`) and the MCP `-32602`
  message is the re-mint detail, asserting `"invalid or expired"`, `"re-mint"`,
  and that `"policy must grant"` is **gone**
  (`test_register_dead_token_names_the_remint_not_the_policy`)
- [x] **AC2** — stubbed Vault, `lookup-self` succeeds + write 403s →
  `token_invalid is False` (`test_write_denied_with_live_token_keeps_policy_disposition`)
  and the MCP message equals the pre-#2652 string character for character
  (`test_write_denied_detail_is_unchanged` pins the constant literally;
  `test_register_vault_write_failure_maps_to_invalid_params` asserts equality)
- [x] **AC3** — REST `agent_principals` + `runner_principals` parametrized over both
  dispositions asserting the 502 `detail`
  (`test_register_vault_write_failure_detail_splits_on_token_validity` in both files);
  UI banner asserted in `test_register_dead_vault_token_banner_names_the_remint`
  (dead) and the pre-existing `test_register_vault_error_renders_502` (live)
- [x] **AC4** — `docs/cross-repo/vault-provisioning.md` dead-token row added
- [x] **AC5** — CHANGELOG `[Unreleased]` under `### Fixed — scheduler Vault dead-token diagnostics (#2652)`
- [x] Also covered: the unreachable-write arm does **not** spend a `lookup-self`,
  an unreachable probe stays conservative, and every pre-existing raise site keeps
  `token_invalid=False`
- [x] `cd cli && make snapshot-openapi` — `cli/api/openapi.json` **unchanged**
  (no route signature or docstring touched; `detail` is already a free-form string)

## Out of scope

- The consumer-side periodic-token lifecycle that keeps killing the token
  (RDC-side operational issue).
- Token auto-renewal / renew-on-use (#2328) — untouched.
- Retrying the write after a failed `lookup-self`.
- The `read_agent_secret` failure path: it does not reach an operator-facing
  surface (the scheduler logs a skip), so it keeps its single message.
- `docs/codebase/error-message-shape.md`'s audit table is an explicit **v0.6.0**
  point-in-time snapshot and does not list this site; left unedited rather than
  back-dating a new row into it.

Closes #2652

---

## Follow-up (auto-implement-issue, iteration 1, 2026-07-21)

Addressed review findings from `.claude/auto/review-results/pr-2658-iter-1.json`:

- **B1** `d97d9009` — `_scheduler_token_rejected` now condemns the token only
  on `hvac.exceptions.Forbidden` (403). Every other `VaultError` — hvac maps
  429/500/502/503 onto their own `VaultError` subclasses — is inconclusive:
  logged as `scheduler_vault_token_lookup_inconclusive` with `reason=<class>`
  and returns `False`, so a sealed or failing-over Vault keeps the
  policy-scope remediation instead of ordering a re-mint mid-outage.
- **M1** `d97d9009` — the probe now fires from an `except
  hvac.exceptions.Forbidden` write arm; a following `except VaultError` raises
  with `token_invalid=False` and spends no round-trip. The
  `scheduler_vault_token_dead` event gained `reason=` to match
  `verify_scheduler_token`'s shape.

Tests: `test_only_a_403_lookup_proves_the_token_is_dead` (5 cases) and
`test_non_403_write_rejection_skips_the_probe` (5 cases). The nine non-403
cases fail against the pre-fix code (verified by reverting the source file and
re-running). `SCHEDULER_VAULT_WRITE_DENIED_DETAIL` is byte-for-byte unchanged
(AC2), and the doc / `docs/codebase/scheduler.md` / CHANGELOG wording is
aligned with the narrowed 403-only split.

Disputed: none.

Deferred: none. The reviewer's adjacent note about
`scheduler/vault_credentials.py` nearing the 600-line file threshold still
stands — the file is 599 lines after this iteration (prose trimmed to absorb
the fix); the lookup-self / token-lifetime split seam remains the natural
follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduler Vault error diagnostics by distinguishing expired or invalid tokens from insufficient permissions.
  * Registration failures now provide targeted remediation, including re-minting dead tokens or widening policy permissions when appropriate.
  * Consistent error details are shown across REST APIs, MCP registration, and the agent registration UI.

* **Documentation**
  * Updated scheduler and Vault provisioning guidance for token diagnosis, lookup permissions, renewal, and recovery procedures.

### DIFF
--- a/backend/src/meho_backplane/scheduler/vault_credentials.py
+++ b/backend/src/meho_backplane/scheduler/vault_credentials.py
@@ -26,10 +26,9 @@ records AppRole as the alternative service identity but flags its
 ``secret_id`` bootstrap (secret-zero) cost. A static token bound to a
 narrow read/write policy on the agent-credentials path is the
 lowest-friction shippable identity — it reuses hvac's
-``Client(token=…)`` primitive (the same one the live-Vault test harness
-uses) with no ``secret_id`` exchange. An operator who prefers AppRole
-runs a Vault Agent sidecar that renews a token into
-``VAULT_SCHEDULER_TOKEN``: additive, no code change here.
+``Client(token=…)`` primitive with no ``secret_id`` exchange. An
+operator who prefers AppRole runs a Vault Agent sidecar that renews a
+token into ``VAULT_SCHEDULER_TOKEN``: additive, no code change here.
 
 Path convention
 ===============
@@ -43,9 +42,9 @@ mount (they insert ``data/`` themselves — verified against hvac 2.4.0).
 setting keeps working unchanged (#823 promised "a code swap, not an
 env-var rename").
 
-The secret payload shape is ``{"client_secret": "<value>"}``; the read
-returns the ``client_secret`` field. No secret value ever enters a log
-event or an error message — only the path and the field *name* do.
+The secret payload shape is ``{"client_secret": "<value>"}``. No secret
+value ever enters a log event or an error message — only the path and
+the field *name* do.
 
 Token lifetime — renew-on-use + self-lookup (#2328)
 ===================================================
@@ -58,27 +57,24 @@ Vault-first read returns 403 and the scheduler silently skips. Two
 mechanisms defuse it here:
 
 * :func:`_maybe_renew_scheduler_token` fires a best-effort
-  ``auth/token/renew-self`` after every successful read/write. The
-  token is renewed at scheduler-tick frequency, so a periodic token
-  with any sane ``period`` never expires while the process runs. The
-  renewal is best-effort: the read/write already succeeded, so a failed
-  renewal is logged and swallowed rather than turned into a new failure
-  mode.
+  ``auth/token/renew-self`` after every successful read/write, at
+  scheduler-tick frequency, so a periodic token with any sane
+  ``period`` never expires while the process runs. A failed renewal is
+  logged and swallowed — the read/write it follows already succeeded.
 * :func:`verify_scheduler_token` runs ``auth/token/lookup-self`` at
   scheduler startup and on a slow cadence from the tick loop. It does
   not fix the fuse — it shortens time-to-notice from weeks to minutes
-  by logging a dead/unreachable token as a loud ``ERROR`` the moment
-  it's observed. Sibling #2327 consumes the same signal for its
-  ``/ready features.scheduler`` skip-state surface.
+  by logging a dead/unreachable token as a loud ``ERROR``. Sibling
+  #2327 consumes the same signal for its ``/ready features.scheduler``
+  skip-state surface.
 
 The same ``lookup-self`` primitive also disambiguates the *write*
-failure path (#2652): Vault answers a dead token and a live token on an
-under-scoped policy with the same 403, but the fixes are opposites
-(re-mint vs. widen the policy). :func:`write_agent_secret` probes
-:func:`_scheduler_token_rejected` on rejection and stamps
-``token_invalid`` on the raised :class:`SchedulerVaultBrokerError`, so
-every register surface picks the right remediation off one flag.
-Diagnosis only: the write is not retried.
+failure path (#2652): a dead token and a live token on an under-scoped
+policy both draw a **403**, but the fixes are opposites (re-mint vs.
+widen the policy). :func:`write_agent_secret` probes
+:func:`_scheduler_token_rejected` on that 403 — and only that 403 — and
+stamps ``token_invalid`` on the raised
+:class:`SchedulerVaultBrokerError`. Diagnosis only: no retry.
 
 The token is resolved from its live source on **every** use
 (:func:`_current_scheduler_token`) rather than frozen at process start,
@@ -152,9 +148,9 @@ class SchedulerVaultBrokerError(Exception):
     (#2652) so every consuming surface — MCP, the two REST register
     routes, the UI banner — picks the right remediation from one shared
     diagnosis instead of re-probing Vault itself. ``True``: the token was
-    rejected by ``auth/token/lookup-self`` and must be re-minted.
-    ``False``: it is live (or its liveness could not be established) and
-    the pre-existing policy-scope remediation stands.
+    403'd by ``auth/token/lookup-self`` and must be re-minted. ``False``:
+    it is live, or its liveness could not be established — either way the
+    pre-existing policy-scope remediation stands.
     """
 
     def __init__(self, *args: object, token_invalid: bool = False) -> None:
@@ -336,11 +332,12 @@ async def write_agent_secret(identity_ref: str, client_secret: str) -> str:
         with a warning (env-var fallback remains); the read caller treats
         it as "fall back to the env var".
     SchedulerVaultBrokerError
-        Vault is unreachable or rejected the write. On a Vault-level
-        rejection the error carries ``token_invalid`` — ``True`` when a
-        follow-up ``lookup-self`` shows the scheduler token itself is
-        dead (re-mint it), ``False`` when the token is live and the
-        policy scope is the fault (#2652).
+        Vault is unreachable or rejected the write. On a **403** the
+        error carries ``token_invalid=True`` when a follow-up
+        ``lookup-self`` also 403s (the token is dead — re-mint it) and
+        ``False`` when the token is live, leaving the policy scope at
+        fault (#2652). No other status is evidence about the token, so
+        each keeps ``False``.
     """
     settings = get_settings()
     api_path = vault_path_for_client_id(identity_ref, settings=settings)
@@ -357,19 +354,30 @@ async def write_agent_secret(identity_ref: str, client_secret: str) -> str:
         raise SchedulerVaultBrokerError(
             f"vault unreachable writing agent secret at {api_path!r}: {type(exc).__name__}"
         ) from exc
-    except hvac.exceptions.VaultError as exc:
-        # Vault answers a dead token and a live-but-under-scoped policy
-        # with the same 403, yet the remediations are opposites (re-mint
-        # vs. widen the policy). Split them with one ``lookup-self`` on
-        # the *same* client that just failed, and carry the disposition
-        # on the exception so all four register surfaces inherit it
+    except hvac.exceptions.Forbidden as exc:
+        # 403 is the one ambiguous rejection: a dead token and a live-but-
+        # under-scoped policy look identical, yet the remediations are
+        # opposites (re-mint vs. widen the policy). Split them with one
+        # ``lookup-self`` on the *same* client that just failed and carry
+        # the disposition so all four register surfaces inherit it
         # (#2652). Diagnosis only — the write is never retried.
         token_invalid = await _scheduler_token_rejected(client)
         if token_invalid:
-            _log.error("scheduler_vault_token_dead", check="agent_secret_write")
+            _log.error(
+                "scheduler_vault_token_dead",
+                reason=type(exc).__name__,
+                check="agent_secret_write",
+            )
         raise SchedulerVaultBrokerError(
             f"vault rejected agent-secret write at {api_path!r}: {type(exc).__name__}",
             token_invalid=token_invalid,
+        ) from exc
+    except hvac.exceptions.VaultError as exc:
+        # Every other status hvac maps onto ``VaultError`` (429, 500, 502,
+        # 503 sealed/down …) describes Vault, not the token — nothing to
+        # disambiguate, so no probe; the policy-scope remediation stands.
+        raise SchedulerVaultBrokerError(
+            f"vault rejected agent-secret write at {api_path!r}: {type(exc).__name__}"
         ) from exc
     # The token just authenticated a write — renew it so the periodic
     # token never ages out (#2328). Best-effort; never raises.
@@ -462,10 +470,10 @@ def _lookup_self_blocking(client: hvac.Client) -> object:
 async def _scheduler_token_rejected(client: hvac.Client) -> bool:
     """Is *client*'s token itself dead, rather than merely under-scoped?
 
-    Called from the write-failure path (#2652) after Vault rejected a
-    write. A revoked / expired / lost-lease token and a live token on a
-    policy without ``create``+``update`` both produce a 403, so the write
-    response alone cannot name the remediation.
+    Called from the write-failure path (#2652) after Vault answered a
+    write with 403. A revoked / expired / lost-lease token and a live
+    token on a policy without ``create``+``update`` both produce that
+    403, so the write response alone cannot name the remediation.
     ``auth/token/lookup-self`` can: Vault answers it for *any* live token
     regardless of policy (it is self-scoped) and 403s an invalid one.
 
@@ -473,14 +481,24 @@ async def _scheduler_token_rejected(client: hvac.Client) -> bool:
     describe the identity that was actually denied, and re-resolving
     could pick up a token re-minted between the two calls.
 
-    Returns ``True`` only on a Vault-level rejection. A transport failure
-    is not evidence of a dead token, so it returns ``False`` and the
-    caller keeps the policy-scope remediation.
+    Returns ``True`` only on a 403. hvac maps *every* Vault status onto a
+    :class:`~hvac.exceptions.VaultError` subclass, so a sealed (503),
+    overloaded (429) or broken (500/502) Vault would otherwise read as a
+    dead token and have an operator re-mint a healthy one mid-outage.
+    Those, like a transport failure, are inconclusive — ``False``, which
+    keeps the caller on the policy-scope remediation.
     """
     try:
         await asyncio.to_thread(_lookup_self_blocking, client)
-    except hvac.exceptions.VaultError:
+    except hvac.exceptions.Forbidden:
         return True
+    except hvac.exceptions.VaultError as exc:
+        _log.warning(
+            "scheduler_vault_token_lookup_inconclusive",
+            reason=type(exc).__name__,
+            check="agent_secret_write",
+        )
+        return False
     except requests.exceptions.RequestException:
         _log.warning("scheduler_vault_token_lookup_unreachable", check="agent_secret_write")
         return False
@@ -510,10 +528,7 @@ async def verify_scheduler_token(*, reason: str = "check") -> SchedulerTokenStat
 
     Called at scheduler startup and on a slow cadence from the tick loop
     (:mod:`meho_backplane.scheduler.loop`). It does **not** fix the fuse
-    (that is :func:`_maybe_renew_scheduler_token`) — it shortens
-    time-to-notice from weeks to minutes by surfacing a dead or
-    unreachable token as a loud ``ERROR`` the moment it's observed,
-    rather than as silent credential-unresolved skips downstream.
+    (that is :func:`_maybe_renew_scheduler_token`; see the module docstring).
 
     Never raises. An unconfigured token returns ``configured=False``
     (the documented env-var-fallback opt-out) and any Vault failure is
--- a/backend/src/meho_backplane/scheduler/vault_credentials.py
+++ b/backend/src/meho_backplane/scheduler/vault_credentials.py
@@ -71,6 +71,15 @@ mechanisms defuse it here:
   it's observed. Sibling #2327 consumes the same signal for its
   ``/ready features.scheduler`` skip-state surface.
 
+The same ``lookup-self`` primitive also disambiguates the *write*
+failure path (#2652): Vault answers a dead token and a live token on an
+under-scoped policy with the same 403, but the fixes are opposites
+(re-mint vs. widen the policy). :func:`write_agent_secret` probes
+:func:`_scheduler_token_rejected` on rejection and stamps
+``token_invalid`` on the raised :class:`SchedulerVaultBrokerError`, so
+every register surface picks the right remediation off one flag.
+Diagnosis only: the write is not retried.
+
 The token is resolved from its live source on **every** use
 (:func:`_current_scheduler_token`) rather than frozen at process start,
 so a Vault-Agent sidecar (or an operator) that re-mints the token into
@@ -92,6 +101,8 @@ from meho_backplane.auth.vault import _build_client
 from meho_backplane.settings import Settings, get_settings
 
 __all__ = [
+    "SCHEDULER_VAULT_TOKEN_INVALID_DETAIL",
+    "SCHEDULER_VAULT_WRITE_DENIED_DETAIL",
     "SECRET_FIELD",
     "SchedulerTokenStatus",
     "SchedulerVaultBrokerError",
@@ -109,6 +120,26 @@ _log = structlog.get_logger(__name__)
 #: Both the write (register) and the read (scheduler) agree on this key.
 SECRET_FIELD: str = "client_secret"
 
+#: Remediation for a write Vault denied while the scheduler token is
+#: **live** — the bound policy lacks ``create``/``update`` on the
+#: agent-credentials path. Kept verbatim from the pre-#2652 MCP message.
+SCHEDULER_VAULT_WRITE_DENIED_DETAIL: str = (
+    "scheduler Vault write failed — VAULT_SCHEDULER_TOKEN policy must "
+    "grant create/update on the agent-credentials path"
+)
+
+#: Remediation for a write Vault denied **because the token is dead**
+#: (revoked / expired / lost lease) — indistinguishable from the case
+#: above by status code alone, hence the ``lookup-self`` probe (#2652).
+#: Three-clause shape per ``docs/codebase/error-message-shape.md``.
+SCHEDULER_VAULT_TOKEN_INVALID_DETAIL: str = (
+    "scheduler_vault_token_invalid: the scheduler Vault token is invalid "
+    "or expired — Vault rejected auth/token/lookup-self for it, so the "
+    "policy scope is not the fault. Re-mint VAULT_SCHEDULER_TOKEN (or "
+    "refresh the file VAULT_SCHEDULER_TOKEN_FILE points at) and update "
+    "the deployment secret per docs/cross-repo/vault-provisioning.md."
+)
+
 
 class SchedulerVaultBrokerError(Exception):
     """Base class for scheduler-service-token Vault broker failures.
@@ -116,7 +147,19 @@ class SchedulerVaultBrokerError(Exception):
     Raised on read/write failures that are *not* the
     not-configured case (which has its own subclass so callers can choose
     to fall back to the env-var path rather than fail).
+
+    ``token_invalid`` carries the broker's ``lookup-self`` disposition
+    (#2652) so every consuming surface — MCP, the two REST register
+    routes, the UI banner — picks the right remediation from one shared
+    diagnosis instead of re-probing Vault itself. ``True``: the token was
+    rejected by ``auth/token/lookup-self`` and must be re-minted.
+    ``False``: it is live (or its liveness could not be established) and
+    the pre-existing policy-scope remediation stands.
     """
+
+    def __init__(self, *args: object, token_invalid: bool = False) -> None:
+        super().__init__(*args)
+        self.token_invalid = token_invalid
 
 
 class SchedulerVaultNotConfiguredError(SchedulerVaultBrokerError):
@@ -293,7 +336,11 @@ async def write_agent_secret(identity_ref: str, client_secret: str) -> str:
         with a warning (env-var fallback remains); the read caller treats
         it as "fall back to the env var".
     SchedulerVaultBrokerError
-        Vault is unreachable or rejected the write.
+        Vault is unreachable or rejected the write. On a Vault-level
+        rejection the error carries ``token_invalid`` — ``True`` when a
+        follow-up ``lookup-self`` shows the scheduler token itself is
+        dead (re-mint it), ``False`` when the token is live and the
+        policy scope is the fault (#2652).
     """
     settings = get_settings()
     api_path = vault_path_for_client_id(identity_ref, settings=settings)
@@ -311,8 +358,18 @@ async def write_agent_secret(identity_ref: str, client_secret: str) -> str:
             f"vault unreachable writing agent secret at {api_path!r}: {type(exc).__name__}"
         ) from exc
     except hvac.exceptions.VaultError as exc:
+        # Vault answers a dead token and a live-but-under-scoped policy
+        # with the same 403, yet the remediations are opposites (re-mint
+        # vs. widen the policy). Split them with one ``lookup-self`` on
+        # the *same* client that just failed, and carry the disposition
+        # on the exception so all four register surfaces inherit it
+        # (#2652). Diagnosis only — the write is never retried.
+        token_invalid = await _scheduler_token_rejected(client)
+        if token_invalid:
+            _log.error("scheduler_vault_token_dead", check="agent_secret_write")
         raise SchedulerVaultBrokerError(
-            f"vault rejected agent-secret write at {api_path!r}: {type(exc).__name__}"
+            f"vault rejected agent-secret write at {api_path!r}: {type(exc).__name__}",
+            token_invalid=token_invalid,
         ) from exc
     # The token just authenticated a write — renew it so the periodic
     # token never ages out (#2328). Best-effort; never raises.
@@ -400,6 +457,34 @@ class SchedulerTokenStatus:
 def _lookup_self_blocking(client: hvac.Client) -> object:
     """Synchronously call ``auth/token/lookup-self`` for *client*'s token."""
     return client.auth.token.lookup_self()
+
+
+async def _scheduler_token_rejected(client: hvac.Client) -> bool:
+    """Is *client*'s token itself dead, rather than merely under-scoped?
+
+    Called from the write-failure path (#2652) after Vault rejected a
+    write. A revoked / expired / lost-lease token and a live token on a
+    policy without ``create``+``update`` both produce a 403, so the write
+    response alone cannot name the remediation.
+    ``auth/token/lookup-self`` can: Vault answers it for *any* live token
+    regardless of policy (it is self-scoped) and 403s an invalid one.
+
+    Probes the **same** client that failed the write — the answer must
+    describe the identity that was actually denied, and re-resolving
+    could pick up a token re-minted between the two calls.
+
+    Returns ``True`` only on a Vault-level rejection. A transport failure
+    is not evidence of a dead token, so it returns ``False`` and the
+    caller keeps the policy-scope remediation.
+    """
+    try:
+        await asyncio.to_thread(_lookup_self_blocking, client)
+    except hvac.exceptions.VaultError:
+        return True
+    except requests.exceptions.RequestException:
+        _log.warning("scheduler_vault_token_lookup_unreachable", check="agent_secret_write")
+        return False
+    return False
 
 
 def _unwrap_token_lifetime(payload: object) -> tuple[int | None, str | None]:
--- a/backend/src/meho_backplane/scheduler/vault_credentials.py
+++ b/backend/src/meho_backplane/scheduler/vault_credentials.py
@@ -474,8 +474,9 @@ async def _scheduler_token_rejected(client: hvac.Client) -> bool:
     write with 403. A revoked / expired / lost-lease token and a live
     token on a policy without ``create``+``update`` both produce that
     403, so the write response alone cannot name the remediation.
-    ``auth/token/lookup-self`` can: Vault answers it for *any* live token
-    regardless of policy (it is self-scoped) and 403s an invalid one.
+    ``auth/token/lookup-self`` can: Vault answers it for a live token
+    granted ``read`` there (``meho-scheduler`` grants it — load-bearing,
+    see ``docs/cross-repo/vault-provisioning.md``) and 403s an invalid one.
 
     Probes the **same** client that failed the write — the answer must
     describe the identity that was actually denied, and re-resolving
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,15 +100,19 @@ connector-related release-notes line.
   the operator verify policy, path pattern, and mount (all correct)
   before anyone thought to `vault token lookup` the token itself. The
   broker now fires the `auth/token/lookup-self` probe it already shipped
-  for #2328 on the write-failure path and carries the disposition on
+  for #2328 whenever Vault answers the write with a 403 — the one
+  ambiguous status — and carries the disposition on
   `SchedulerVaultBrokerError`; a dead token surfaces as
   `scheduler_vault_token_invalid: the scheduler Vault token is invalid or
   expired …` with the re-mint remediation on all four surfaces (the MCP
   `meho.agent_principals.register` `-32602` message, both REST register
   routes' 502 detail, and the `/ui/agents/principals` register banner),
   while a live-token-denied write keeps the existing policy-scope wording
-  unchanged. Diagnosis only — the write is not retried, and the
-  renew-on-use loop from #2328 is untouched.
+  unchanged. Only a 403 on the probe condemns the token: a sealed (503),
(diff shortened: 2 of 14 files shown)


backend/src/meho_backplane/scheduler/vault_credentials.py
```diff
@@ -402,6 +459,34 @@ def _lookup_self_blocking(client: hvac.Client) -> object:
...
+    could pick up a token re-minted between the two calls.
+
+    Returns ``True`` only on a Vault-level rejection. A transport failure
+    is not evidence of a dead token, so it returns ``False`` and the
+    caller keeps the policy-scope remediation.
+    """
+    try:
+        await asyncio.to_thread(_lookup_self_blocking, client)
+    except hvac.exceptions.VaultError:
+        return True
```
damir-topic, line 495: **[🛑 B1]** — a sealed / down / rate-limited Vault is reported to the operator as a dead token

**What:** `_scheduler_token_rejected` treats **every** `hvac.exceptions.VaultError` from `lookup-self` as proof the token is dead. In hvac 2.4.0 that class is the base of the whole status-code family — `VaultDown` (503), `InternalServerError` (500), `BadGateway` (502), `RateLimitExceeded` (429), `VaultNotInitialized` (501), `InvalidRequest` (400) all subclass it, not just `Forbidden` (403). Verified by running the write path against this branch with `lookup_raises` set to each of `VaultDown` / `InternalServerError` / `BadGateway` / `RateLimitExceeded`: all four yield `token_invalid is True`.

**Why it matters:** the arm is reachable with a perfectly healthy token. A sealed Vault (or one mid-HA-failover, or overloaded) answers the KV write with 503 → that lands in the `except hvac.exceptions.VaultError` arm at line 360 → the probe also gets 503 → the operator is told, verbatim: *"the scheduler Vault token is invalid or expired — Vault rejected auth/token/lookup-self for it, so the policy scope is not the fault. Re-mint VAULT_SCHEDULER_TOKEN …"*. That is a confident, false, and actively expensive diagnosis — it sends the operator to rotate a working credential during a Vault outage, which is the exact defect class #2652 exists to remove. It also contradicts this function's own docstring ("Returns `True` only on a Vault-level *rejection*"), the conservatism the PR applies to the transport arm and to the unreachable-write arm, and the doc this PR ships: `docs/cross-repo/vault-provisioning.md` states the backplane "told them apart by getting a **403** from `auth/token/lookup-self`" — the code accepts any status.

**Required change:** make 403 the only evidence of a dead token and treat every other Vault-side status as inconclusive (→ `False`, keep the policy-scope remediation), mirroring the transport arm:

```suggestion
    try:
        await asyncio.to_thread(_lookup_self_blocking, client)
    except hvac.exceptions.Forbidden:
        # 403 on a self-scoped endpoint is the dead/revoked/expired-token
        # signature — Vault answers lookup-self for any live token
        # regardless of policy.
        return True
    except hvac.exceptions.VaultError as exc:
        # Any other Vault status (503 sealed/failing-over, 500, 502, 429)
        # says nothing about *this token's* validity. Inconclusive is not
        # evidence — stay on the policy-scope remediation.
        _log.warning(
            "scheduler_vault_token_lookup_inconclusive",
            reason=type(exc).__name__,
            check="agent_secret_write",
        )
        return False
    except requests.exceptions.RequestException:
        _log.warning("scheduler_vault_token_lookup_unreachable", check="agent_secret_write")
        return False
    return False
```

Add a parametrized regression test next to `test_write_denied_with_unreachable_lookup_stays_conservative` over `hvac.exceptions.VaultDown` / `InternalServerError` / `BadGateway` / `RateLimitExceeded` asserting `token_invalid is False`, so the narrowing can't silently regress.

**Source:** Vault HTTP API status codes — 403 "your authentication details are either incorrect…" vs. 503 "Vault is down for maintenance, currently sealed, or temporarily overloaded" (https://developer.hashicorp.com/vault/api-docs); hvac 2.4.0 `VaultError.from_status` status→class map, every class subclassing `VaultError` (https://github.com/hvac/hvac/blob/main/hvac/exceptions.py). Ships-in-this-PR doc claim: `docs/cross-repo/vault-provisioning.md` failure-modes dead-token row.

backend/src/meho_backplane/scheduler/vault_credentials.py
```diff
@@ -311,8 +358,18 @@ async def write_agent_secret(identity_ref: str, client_secret: str) -> str:
             f"vault unreachable writing agent secret at {api_path!r}: {type(exc).__name__}"
         ) from exc
     except hvac.exceptions.VaultError as exc:
+        # Vault answers a dead token and a live-but-under-scoped policy
+        # with the same 403, yet the remediations are opposites (re-mint
+        # vs. widen the policy). Split them with one ``lookup-self`` on
+        # the *same* client that just failed, and carry the disposition
+        # on the exception so all four register surfaces inherit it
+        # (#2652). Diagnosis only — the write is never retried.
+        token_invalid = await _scheduler_token_rejected(client)
```
damir-topic, line 364: **[⚠️ M1]** — the probe fires for every Vault status, not just the ambiguous 403

**What:** the probe is triggered from the broad `except hvac.exceptions.VaultError` arm, so a write that failed with 400 / 404 / 429 / 5xx also spends a `lookup-self` round-trip. The ambiguity this feature exists to resolve is specific to **403**: only there do "dead token" and "under-scoped policy" collide.

**Why it matters:** it buys nothing on the non-403 statuses (the write's own status already names the fault) and it is the second half of B1's misdiagnosis — a 503 write plus a 503 probe currently produces the re-mint message. Gating the probe on 403 also makes the register path spend zero extra Vault calls while Vault is unhealthy, which is when you least want a second request.

**Required change:** split the rejection arm so only `Forbidden` probes; every other `VaultError` raises with the default `token_invalid=False` and no extra request:

```suggestion
    except hvac.exceptions.Forbidden as exc:
        # 403 is the ambiguous case: a dead token and a live token on an
        # under-scoped policy are indistinguishable by status alone. One
        # ``lookup-self`` on the *same* client that just failed splits
        # them, and the disposition rides on the exception so all four
        # register surfaces inherit it (#2652). Diagnosis only — the
        # write is never retried.
        token_invalid = await _scheduler_token_rejected(client)
        if token_invalid:
            _log.error(
                "scheduler_vault_token_dead",
                reason=type(exc).__name__,
                check="agent_secret_write",
            )
        raise SchedulerVaultBrokerError(
            f"vault rejected agent-secret write at {api_path!r}: {type(exc).__name__}",
            token_invalid=token_invalid,
        ) from exc
    except hvac.exceptions.VaultError as exc:
        # Any other Vault status already names its own fault; there is no
        # 403 ambiguity to split, so don't spend a probe on it.
        raise SchedulerVaultBrokerError(
            f"vault rejected agent-secret write at {api_path!r}: {type(exc).__name__}"
        ) from exc
```

Extend `test_scheduler_vault_credentials.py` with a case asserting a non-403 write rejection leaves `token_invalid is False` **and** `fake_kv.token_api.lookup_calls == 0` — the same shape `test_write_unreachable_maps_to_broker_error` already uses for the transport arm.

damir-topic (commented):




**Anchor task:** #2652 — disambiguate a dead scheduler Vault token from an under-scoped policy (lookup-self on the write-failure path)
**Surface area:** backend (+ devops-adjacent docs)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `a0d33348`

The shape of this change is right — the fix lands at the broker, all four
surfaces read one flag, the verbatim-preservation AC is genuinely met, and the
docs are the best part of the PR. One blocker: the probe's evidence is too
broad, so the feature can hand an operator a *new* wrong remediation on a
Vault outage — the same defect class #2652 was filed against.

### Acceptance criteria

- [x] **AC1** — stubbed Vault, `lookup-self` fails + write 403s → error carries the token-invalid disposition and the MCP `-32602` says "invalid or expired … re-mint". Verified by running `test_write_denied_with_dead_token_sets_token_invalid` and `test_register_dead_token_names_the_remint_not_the_policy` (pass), and by reading `mcp/tools/agent_principals.py:174-186`.
- [x] **AC2** — stubbed Vault, `lookup-self` succeeds + write 403s → policy-scope message preserved **verbatim**. Confirmed by byte-diffing the new `SCHEDULER_VAULT_WRITE_DENIED_DETAIL` against the pre-PR literal at `mcp/tools/agent_principals.py:174-176` on `origin/main` — identical, including the em dash and the line-continuation split. `test_write_denied_detail_is_unchanged` pins the literal independently of the constant, which is the right way to write that test.
- [x] **AC3** — REST + UI carry the same distinction. Both register routes and the UI banner read `exc.token_invalid`; each has its own test rather than an assertion on the shared constant in isolation (`test_register_vault_write_failure_detail_splits_on_token_validity` parametrized over both dispositions in *both* REST files; `test_register_dead_vault_token_banner_names_the_remint` + the pre-existing `test_register_vault_error_renders_502` for the UI). Also verified the UI template renders `banner` as free text (`_register_modal.html:45-47`, `break-words`), so a long detail string renders rather than falling through a code lookup.
- [x] **AC4** — `docs/cross-repo/vault-provisioning.md` dead-token row added next to the under-scoped-policy row, with the `vault token lookup` confirmation step and the re-mint command.
- [x] **AC5** — CHANGELOG `[Unreleased]` under the unique header `### Fixed — scheduler Vault dead-token diagnostics (#2652)`.

### `token_invalid` default — every pre-existing raise site enumerated

All five construction sites in `scheduler/vault_credentials.py` were checked; none
passes the new keyword, so all keep `token_invalid=False`:

| Site | Line | Disposition |
|---|---|---|
| `SchedulerVaultNotConfiguredError` (no token wired) | `:284` | inherits `__init__`, `False` ✅ |
| write — Vault unreachable | `:357` | `False` ✅ |
| write — Vault rejected | `:370` | explicit kwarg (the new behaviour) ✅ |
| read — Vault unreachable | `:420` | `False` ✅ |
| read — Vault rejected | `:424` | `False` ✅ |

Propagation was traced too: `auth/agent_principals.py:373` and
`auth/runner_principals.py:388` catch with a bare `raise`, so the *same instance*
reaches the surfaces and the flag survives the Keycloak rollback.
`test_broker_error_defaults_to_token_valid` covers the base class *and* the
`NotConfigured` subclass.

### Design deviation — probing the same hvac client instead of `verify_scheduler_token()`

Correct call, and for the reason the author gives. `verify_scheduler_token()`
re-enters `_current_scheduler_token()`, which re-reads `VAULT_SCHEDULER_TOKEN_FILE`
on every call (`:263-271`) precisely so a sidecar re-mint is picked up without a
restart — exactly the property that would make it the wrong probe here. A diagnosis
of "which identity did Vault deny" must interrogate the identity that was denied,
and a re-mint landing between the write and the probe would otherwise flip a real
dead-token failure into a policy-scope message. Reusing `_lookup_self_blocking` (the
seam, not the wrapper) keeps the single-sourcing without the re-resolution.

### Out-of-scope discipline

Clean. No retry after a failed lookup (`test_write_denied_with_dead_token_sets_token_invalid`
asserts `fake_kv.writes == []`), `_maybe_renew_scheduler_token` / #2328 untouched,
`read_agent_secret` deliberately left on its single message, consumer-side token
lifecycle untouched. `git diff origin/main -- deploy/` is empty. No stealth edits:
every one of the 14 files maps to a stated AC.

### Project-board alignment

|---|---|---|---|---|---|
| #2652 | open (task, G0.37) | `Closes` | ✅ | ⚠️ partial — AC1/AC2 hold, but the diagnosis is unsound on non-403 Vault statuses (B1) | close on merge, after B1 |
| #2656 | open (initiative) | none | — | 1 of 4 children | leave open |

No sibling open PR touches `vault_credentials.py`, either register route, or
`principals_forms.py` — no collision.

### Predicted vs. actual diff

- ✅ `scheduler/vault_credentials.py` — probe + flag + two constants (predicted)
- ✅ `mcp/tools/agent_principals.py`, `api/v1/agent_principals.py`, `api/v1/runner_principals.py`, `ui/routes/agents/principals_forms.py` — all four surfaces (predicted)
- ✅ five test files, one per surface plus the broker (predicted)
- ✅ `docs/cross-repo/vault-provisioning.md`, `CHANGELOG.md` (predicted, AC4/AC5)
- ❓ `docs/codebase/scheduler.md`, `docs/codebase/ui.md` — unexpected but justified: both documented the old single-message behaviour and would have gone stale. Correct call under the docs-freshness rule.
- ⚠️ `scheduler/vault_credentials.py:360-373` and `:462-487` — expected, but wrong (see B1/M1 inline)
- 🚫 none
- 🕳️ none. `cli/api/openapi.json` was predicted-then-ruled-out: the only docstring touched under a route module is the `submit_register` *helper* (`principals_forms.py`), not the decorated handler (`ui/routes/agents/routes.py:336-347`), so the snapshot is genuinely unaffected — and the "CLI API snapshot freshness" lane is green, which confirms it.

### Integration verification

Every command run inside `.claude/worktrees/pr-2658-review` at `a0d33348`.

- [x] `ruff check backend/src backend/tests` — clean
- [x] `ruff format --check` — 1357 files already formatted
- [x] `mypy src/meho_backplane --ignore-missing-imports` — 680 files, no issues
- [x] `pytest` on all five touched test files — **106 passed**
- [x] `pytest tests/test_scheduler_credentials.py` (the other broker consumer) — 23 passed
- [x] **Adversarial probe (the finding):** temporarily parametrized the broker's write-denial test over `hvac.exceptions.VaultDown` / `InternalServerError` / `BadGateway` / `RateLimitExceeded` on the `lookup-self` stub — **all four return `token_invalid is True`**. Worktree restored afterwards; nothing pushed.
- [x] Pre-existing-red claim checked: `test_chart_mcp_resource_uri.py` — `git diff origin/main -- deploy/` is empty, the module is `pytest.mark.skipif(shutil.which("helm") is None)`, and the required `helm install + helm test (pgvector preflight)` lane is **SUCCESS** on this PR. Claim holds; not attributable here and not a blocker.
- [⚠️ not run] live-Vault round trip (`tests/integration/test_scheduler_vault_credentials_dev_e2e.py`) — the testcontainers lane was still IN_PROGRESS on CI at review time; the disposition split is fake-client-covered and the probe is a pure exception-classification change, so the fake seam is the right level. Reasoned-only for the real-Vault status codes; the status→exception mapping is cited from hvac 2.4.0 source instead.

### Framework research

- Vault HTTP API status codes — https://developer.hashicorp.com/vault/api-docs — 403 = "your authentication details are either incorrect, you don't have access to this feature"; 503 = "Vault is down for maintenance, currently sealed, or temporarily overloaded"; 502 = upstream third-party error; 429 = standby / too many requests. These are *different faults*, and only 403 is token-validity evidence.
- hvac 2.4.0 (pinned `hvac>=2.4.0`) — https://github.com/hvac/hvac/blob/main/hvac/exceptions.py — `VaultError.from_status` maps 400→`InvalidRequest`, 401→`Unauthorized`, 403→`Forbidden`, 404→`InvalidPath`, 429→`RateLimitExceeded`, 500→`InternalServerError`, 501→`VaultNotInitialized`, 502→`BadGateway`, 503→`VaultDown`; **all** subclass `VaultError`. This is what makes B1 real rather than theoretical.
- `requests` — `JSONDecodeError` / `InvalidJSONError` subclass `RequestException`, so a proxy-mangled probe response lands in the conservative arm. Verified as correct.

- 🛑 **B1** — `_scheduler_token_rejected` accepts any `hvac.exceptions.VaultError` as proof the token is dead, so a sealed / down / rate-limited Vault yields the "re-mint VAULT_SCHEDULER_TOKEN" message for a healthy token. Required change: catch `hvac.exceptions.Forbidden` → `True`; every other `VaultError` → log inconclusive + `False`; add a parametrized regression test over `VaultDown` / `InternalServerError` / `BadGateway` / `RateLimitExceeded`.
- ⚠️ **M1** — the probe is triggered from the broad `VaultError` write arm, spending a `lookup-self` on statuses that carry no 403 ambiguity. Required change: split the arm so only `Forbidden` probes and every other `VaultError` raises with the default disposition and no extra request; assert `lookup_calls == 0` for a non-403 rejection.

Both were independently flagged by `copilot-pull-request-reviewer` (two unresolved
threads at the same two lines) — the inline comments here carry the concrete fix
sketches, so resolving these resolves those too.

- `_log.error("scheduler_vault_token_dead", check="agent_secret_write")` reuses the event name `verify_scheduler_token` emits at `:543` but drops the `reason=` field that version carries, so a log consumer branching on `reason` sees it absent on the write-path emission. The B1/M1 suggestions fold `reason=type(exc).__name__` back in; worth keeping when applying them.
- `scheduler/vault_credentials.py` is now 584 lines against the repo's 600-line threshold. Nothing to do in this PR, but the module has room for roughly one more feature before it needs a split — the `lookup-self` / token-lifetime helpers (`_lookup_self_blocking`, `_scheduler_token_rejected`, `_unwrap_token_lifetime`, `verify_scheduler_token`, `SchedulerTokenStatus`) are the natural seam.
- The REST 502 detail is now asymmetric — a bare code (`scheduler_vault_write_error`) for the live-token case, a full three-clause string for the dead-token case. That is the correct outcome under `docs/codebase/error-message-shape.md` (name the remediation only when it is knowable) and the same shape `KEYCLOAK_ADMIN_NOT_CONFIGURED_DETAIL` uses, so it is deliberate rather than a defect. Noted only so a later reader does not "fix" it into symmetry.

### Docs

- [x] `docs/codebase/scheduler.md` — updated (new subsection under the token-lifetime section)
- [x] `docs/codebase/ui.md` — updated (banner contract at `:2633-2640`)
- [x] `docs/cross-repo/vault-provisioning.md` — updated (dead-token row + "when the fuse has already blown")
- [x] no .NET analogies introduced
- Note: the shipped `vault-provisioning.md` row already describes the *narrower* behaviour B1 asks for ("told them apart by getting a **403** from `auth/token/lookup-self`"). Fixing B1 makes the code match the doc — no doc edit needed with the fix.

- `test_write_denied_detail_is_unchanged` asserts the constant against a re-typed literal rather than against itself. That is the only version of a "preserved verbatim" test that can actually fail, and it is what let AC2 be verified rather than trusted.
- `assert fake_kv.token_api.lookup_calls == 0` on the unreachable-write path is a test for an *absence* — the unambiguous arm not spending a probe. Most PRs would have left that behaviour untested and let it rot.
- The `vault-provisioning.md` failure-modes table now reads as a genuine decision table: the under-scoped row was amended to say the backplane already proved the token live, so the two rows are mutually exclusive instead of merely adjacent. That is the artefact the operator in the field report needed.

backend/src/meho_backplane/scheduler/vault_credentials.py
damir-topic: Fixed in d97d9009. `_scheduler_token_rejected` now returns `True` only for `hvac.exceptions.Forbidden`; every other `VaultError` is logged as `scheduler_vault_token_lookup_inconclusive` (with `reason=<class>`) and returns `False`, so the caller keeps the verbatim policy-scope remediation. Pinned by `test_only_a_403_lookup_proves_the_token_is_dead`, parametrized over `Forbidden` / `VaultDown` / `InternalServerError` / `BadGateway` / `RateLimitExceeded` — the four non-403 cases fail against the previous code.

backend/src/meho_backplane/scheduler/vault_credentials.py
```diff
@@ -402,6 +459,34 @@ def _lookup_self_blocking(client: hvac.Client) -> object:
...
+    caller keeps the policy-scope remediation.
+    """
+    try:
+        await asyncio.to_thread(_lookup_self_blocking, client)
+    except hvac.exceptions.VaultError:
+        return True
+    except requests.exceptions.RequestException:
+        _log.warning("scheduler_vault_token_lookup_unreachable", check="agent_secret_write")
+        return False
+    return False
```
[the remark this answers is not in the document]
damir-topic: Fixed in d97d9009. `_scheduler_token_rejected` now returns `True` only for `hvac.exceptions.Forbidden`; every other `VaultError` is logged as `scheduler_vault_token_lookup_inconclusive` (with `reason=<class>`) and returns `False`, so the caller keeps the verbatim policy-scope remediation. Pinned by `test_only_a_403_lookup_proves_the_token_is_dead`, parametrized over `Forbidden` / `VaultDown` / `InternalServerError` / `BadGateway` / `RateLimitExceeded` — the four non-403 cases fail against the previous code.

backend/src/meho_backplane/scheduler/vault_credentials.py
damir-topic: Fixed in d97d9009. The probe now fires from an `except hvac.exceptions.Forbidden` arm; a following `except hvac.exceptions.VaultError` raises `SchedulerVaultBrokerError` with `token_invalid=False` and no round-trip. The dead-token log event gained `reason=type(exc).__name__` to match `verify_scheduler_token`'s shape. Pinned by `test_non_403_write_rejection_skips_the_probe` (asserts `token_invalid is False` and `lookup_calls == 0` over 503/500/502/429/400).

backend/src/meho_backplane/scheduler/vault_credentials.py
```diff
@@ -311,8 +358,18 @@ async def write_agent_secret(identity_ref: str, client_secret: str) -> str:
...
     except hvac.exceptions.VaultError as exc:
+        # Vault answers a dead token and a live-but-under-scoped policy
+        # with the same 403, yet the remediations are opposites (re-mint
+        # vs. widen the policy). Split them with one ``lookup-self`` on
+        # the *same* client that just failed, and carry the disposition
+        # on the exception so all four register surfaces inherit it
+        # (#2652). Diagnosis only — the write is never retried.
+        token_invalid = await _scheduler_token_rejected(client)
+        if token_invalid:
+            _log.error("scheduler_vault_token_dead", check="agent_secret_write")
```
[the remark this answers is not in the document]
damir-topic: Fixed in d97d9009. The probe now fires from an `except hvac.exceptions.Forbidden` arm; a following `except hvac.exceptions.VaultError` raises `SchedulerVaultBrokerError` with `token_invalid=False` and no round-trip. The dead-token log event gained `reason=type(exc).__name__` to match `verify_scheduler_token`'s shape. Pinned by `test_non_403_write_rejection_skips_the_probe` (asserts `token_invalid is False` and `lookup_calls == 0` over 503/500/502/429/400).

damir-topic (commented):




**Anchor task:** #2652 — split a dead scheduler Vault token from an under-scoped policy
**Surface area:** backend, devops
**Verdict:** APPROVE
**Head SHA reviewed:** `3d3ef06b4f9890948029e037fd9adcc309a0506b`

This iteration verifies the fix commit `d97d9009` (+ merge `3d3ef06b`) against the
iteration-1 punch list rather than re-deriving the whole PR. Both iteration-1
findings are **closed**, each confirmed independently rather than on the author's
report.

### Iteration-1 punch list

| ID | Status | Independent verification |
|---|---|---|
| 🛑 B1 — any `VaultError` from `lookup-self` read as a dead token | ✅ **closed** | `_scheduler_token_rejected` now returns `True` only under `except hvac.exceptions.Forbidden`; a following `except VaultError` logs `scheduler_vault_token_lookup_inconclusive(reason=…)` and returns `False`. Introspected the **installed** hvac wheel (2.4.0, `backend/.venv/.../hvac/exceptions.py`): `Forbidden` is a direct subclass of `VaultError`, `STATUS_EXCEPTION_MAP` is `{400:InvalidRequest, 401:Unauthorized, 403:Forbidden, 404:InvalidPath, 429:RateLimitExceeded, 500:InternalServerError, 501:VaultNotInitialized, 502:BadGateway, 503:VaultDown}`, all flat siblings, and `utils.raise_for_error` → `VaultError.from_status` is the only raise path. So `VaultDown`/`InternalServerError`/`BadGateway`/`RateLimitExceeded` now land in the inconclusive arm. |
| ⚠️ M1 — probe fired from the broad write arm | ✅ **closed** | Write arms are now, in order: `(ConnectionError, Timeout)` → `Forbidden` (probes) → `VaultError` (no probe, default `token_invalid=False`). `Forbidden` precedes the broader `VaultError`, so the narrow arm is **not** dead code — Python matches `except` clauses top-down and `Forbidden` is the more specific type. Verified by execution, not just by reading: `test_non_403_write_rejection_skips_the_probe` asserts `lookup_calls == 0` over 503/500/502/429/400 and passes. |

### Acceptance criteria (re-verified)
- [x] **AC1** — dead token → `token_invalid is True` + re-mint detail on the MCP `-32602`. Now correctly reachable **only** on a 403/403 pair.
- [x] **AC2** — `SCHEDULER_VAULT_WRITE_DENIED_DETAIL` byte-for-byte unchanged. Verified by AST-extracting the pre-PR implicit-concat literal from `origin/main:backend/src/meho_backplane/mcp/tools/agent_principals.py` and comparing against the imported constant: `IDENTICAL: True`, `bytes equal: True` (em dash included).
- [x] **AC3** — all four surfaces still read the one flag: `mcp/tools/agent_principals.py:184`, `api/v1/agent_principals.py:220`, `api/v1/runner_principals.py:213`, `ui/routes/agents/principals_forms.py:284`. Untouched by `d97d9009`, so iteration 1's per-surface verification still holds.
- [x] **AC4** — `docs/cross-repo/vault-provisioning.md` dead-token row present and now 403-scoped.
- [x] **AC5** — CHANGELOG `[Unreleased]` entry present, updated to describe the 403-only narrowing.

### Raise-site regression check
`SchedulerVaultBrokerError` raise sites went 4 → 5 in this file (`:354` write-unreachable, `:371` the `Forbidden` arm — the only one passing `token_invalid=`, `:379` the new non-403 arm, `:428` read-unreachable, `:432` read-rejected). The new `:379` arm uses the default `token_invalid=False`, and `test_broker_error_defaults_to_token_valid` still pins the default. No pre-existing site changed disposition.

### New tests genuinely fail pre-fix (spot-checked, not trusted)
Reverted **only** `scheduler/vault_credentials.py` to `a0d33348` in the review worktree, kept the new tests, re-ran:

```
9 failed, 1 passed, 38 deselected
```

The 9 are exactly the non-403 parametrisations across both new tests; the 1 pass is
`test_only_a_403_lookup_proves_the_token_is_dead[forbidden]`, which was correct
pre-fix too. Matches the author's claim exactly. Worktree restored to head (`git status` clean).

### Docs no longer over-claim
The under-scoped-policy row previously asserted the backplane "already confirmed via
`auth/token/lookup-self` that the token itself is **live**". It now reads: "probed
`auth/token/lookup-self` after that 403 and did **not** get a 403 back, so the token
is live (or Vault could not answer the probe — check for an outage first)". That is
the honest statement of what a non-403 probe result proves. `docs/codebase/scheduler.md`
carries the matching narrowing. Over-claim resolved.

### Integration verification
- [x] `ruff check src tests` — All checks passed
- [x] `ruff format --check` — 1357 files already formatted
- [x] `mypy src/meho_backplane --ignore-missing-imports` — no issues in 680 source files
- [x] `pytest` over the five touched test files + `test_scheduler_credentials.py` — **139 passed**
- [x] Adversarial pre-fix revert probe — 9 failed / 1 passed, as above
- [x] CI `Python (ruff + mypy + pytest)` — **pass** (16m30s)
- [x] CI `Python (integration testcontainers)` — **pass** (12m2s). This closes iteration 1's `live_vault_integration: not_run` gap.
- [x] CI `Semgrep SAST`, `TruffleHog`, `CodeQL`, `Go`, `Helm`, `helm install + helm test`, `CLI API snapshot freshness` — all pass
- [x] `migrations_replay` — n/a, no migration
- ⚠️ `CodeRabbit` red = **"Review limit reached"** rate limit, zero findings produced. Not a required check.

### Predicted vs. actual diff
- ✅ `scheduler/vault_credentials.py` — now correct; this was iteration 1's ⚠️ file
- ✅ `test_scheduler_vault_credentials.py`, `docs/codebase/scheduler.md`, `docs/cross-repo/vault-provisioning.md`, `CHANGELOG.md`
- ✅ The four surfaces + their tests — unchanged since iteration 1, still correct
- 🚫 none
- 🕳️ none

The merge commit `3d3ef06b` brings `cli/internal/cmd/targets/import.go` and
`docs/codebase/cli.md` into the branch, but those are `4b259a46` (#2657) arriving from
`main` — they are not in this PR's diff against base (still 14 files).

### Findings

- ⚠️ **M1 (new, non-blocking)** — `auth/token/lookup-self` is **not** unconditionally
  answered for a live token, so a 403 there is a very good but not airtight
  discriminator. Vault's `lookup-self` is ACL-checked and reachable via the built-in
  `default` policy; a token minted with `-no-default-policy` (a real hardening choice
  for machine identities, and the module docstring explicitly contemplates an
  AppRole/Vault-Agent operator) gets a 403 on `lookup-self` while being perfectly
  live — which would produce the same wrong "re-mint" remediation #2652 exists to
  remove, just via a much narrower door.

  **Why this is not a blocker.** The code is correct for every token minted by this
  repo's own documented procedure — `docs/cross-repo/vault-provisioning.md:232` is
  `vault token create -policy=meho-scheduler -period=768h -field=token`, which attaches
  `default`; `-no-default-policy` appears nowhere in the repo. The property is also
  inherited from the `lookup-self` probe primitive already shipped for #2328
  (`verify_scheduler_token`), not introduced here. And the failure is self-limiting:
  the doc's remediation column tells the operator to confirm with
  `VAULT_TOKEN=… vault token lookup` before re-minting.

  **Required change (cheap, doc-only):** drop the unqualified "regardless of policy"
  from the `_scheduler_token_rejected` docstring and from
  `docs/cross-repo/vault-provisioning.md`, replacing it with the capability the signal
  actually depends on; and add `path "auth/token/lookup-self" { capabilities = ["read"] }`
  to the `meho-scheduler` policy stanza in the provisioning doc so the probe is sound
  even for a `-no-default-policy` token.

  **Not posted inline** — `copilot-pull-request-reviewer` already has two open,
  non-outdated threads naming exactly this at
  `docs/cross-repo/vault-provisioning.md:201` and
  `backend/src/meho_backplane/scheduler/vault_credentials.py:478`. Duplicating them
  would create a redundant merge obligation under the repo's
  `required_conversation_resolution` setting. Resolve those two threads (or fold the
  doc fix in) rather than treating this as a separate item.

  **Source:** Vault HTTP API status codes — 403 is "your authentication details are
  either incorrect, **you don't have access to this feature**, …"
  (https://developer.hashicorp.com/vault/api-docs); the `default` policy "may be
  explicitly excluded at token creation time"
  (https://developer.hashicorp.com/vault/docs/concepts/policies).

No blockers. Nothing posted inline this iteration.

### Merge prerequisite (not a review finding)
`main` has `required_conversation_resolution: true` and there are **four** unresolved
threads: the two iteration-1 threads at `:494`/`:364` (both answered by the author with
the `d97d9009` fix — safe to resolve) and the two Copilot threads above. They must be
resolved or admin-bypassed before merge; `enforce_admins` is `false`.

- `scheduler/vault_credentials.py` is now **599** lines against the repo's 600-line
  threshold — one line of headroom. `write_agent_secret` also crossed the 60-line
  function warning by gaining a third `except` arm. The split seam named in iteration 1
  still stands: the lookup-self / token-lifetime group (`_lookup_self_blocking`,
  `_scheduler_token_rejected`, `_unwrap_token_lifetime`, `verify_scheduler_token`,
  `SchedulerTokenStatus`). Deliberately **not** required in this PR — the fix commit
  absorbed its own cost by trimming prose, which is the right call for a hotfix but
  means the next change to this file will trip the gate.
- The new `scheduler_vault_token_lookup_inconclusive` warning has no test asserting it
  is emitted; the tests pin behaviour (`token_invalid is False`) rather than the log
  event. Fine as-is, but the event is the operator's only breadcrumb that a probe was
  attempted and came back ambiguous.
- `Unauthorized` (401) is unreachable here — Vault's documented status list has no 401,
  so `Forbidden` is the complete discriminator on the token-validity axis. Recorded so a
  later reader doesn't "fix" a nonexistent gap by widening the narrow arm back out.

### Docs
- [x] `docs/codebase/scheduler.md` — updated, 403-only split recorded accurately
- [x] `docs/cross-repo/vault-provisioning.md` — updated, over-claim removed
- [x] `docs/codebase/ui.md` — unchanged this iteration, still accurate

- The fix is narrower than the punch list required and the commit message explains
  *why* each arm is where it is, including why the non-403 arm skips the probe.
- Reverting the source file to prove the new tests fail pre-fix is the right way to
  demonstrate a regression test has teeth — and it reproduced exactly (9/1) when
  re-run independently here.
- The `reason=type(exc).__name__` addition aligns `scheduler_vault_token_dead` with
  `verify_scheduler_token`'s existing event shape, which was iteration 1's adjacent
  note rather than a required change.

damir-topic:
Follow-up (orchestrator, post-iteration-2): the `lookup-self`-is-ACL-checked finding is correct and is now fixed in two places.

- `b8ad1980` — `docs/cross-repo/vault-provisioning.md`: dropped the "regardless of policy" claim, and the `meho-scheduler` policy stanza now grants `read` on `auth/token/lookup-self` **explicitly** rather than relying on the built-in `default` policy.
- `739091e0` — `_scheduler_token_rejected`'s docstring: same correction, pointing at that grant.

Why it matters: a live token minted with `-no-default-policy` and no explicit grant 403s the probe and would be reported dead — the same wrong-remediation class #2652 exists to fix, through a narrower door. Code behaviour is unchanged (403-only condemnation, from `d97d9009`); this closes the deployment-shape hole around it.

PR #172, opened 2026-05-11
feat(deploy): values-rdc-example.yaml + ESO sync patterns (G2.5-T4)

damir-topic:
## Summary

Ships the sanitized chart-values template and documents the External Secrets Operator (ESO) sync patterns the chart expects the consumer to provision. Wave 4 of G2.5 (Helm chart).

- `deploy/values-examples/values-rdc-example.yaml` — sanitized RDC-lab template; every site-specific field uses a `<REPLACE: ...>` placeholder that fails `values.schema.json` at `helm install` time (CIDRs fail the IPv4 pattern; `image.tag` fails the new OCI-tag pattern). A forgotten substitution surfaces immediately instead of CrashLoopBackOff'ing against the wrong system.
- `deploy/values-examples/README.md` — substitution recipe, full `ClusterSecretStore` + `ExternalSecret` manifests, Vault KV path mapping (`secret/meho/postgres` → `DATABASE_URL`), and the end-to-end install ordering (ESO → ClusterSecretStore → ExternalSecret → wait-for-Secret → `helm install`).
- `deploy/charts/meho/templates/externalsecrets.yaml` — optional chart-side rendering gated on `eso.enabled: false` (default). When `eso.enabled: true`, renders an `ExternalSecret` for `postgres.credentialsSecret`; with `eso.keycloak.enabled: true`, a second one for the Keycloak client secret. Chart never renders the SecretStore — that's consumer-owned (it outlives releases and embeds cluster Vault creds).
- `values.schema.json` — adds the `eso` block with conditional `allOf` requiring `secretStore.{name,kind}` only when `eso.enabled: true`; tightens `image.tag` to the OCI tag grammar so placeholders fail-loud.
- `docs/codebase/devops.md` — adds an "Example values + ESO patterns" section and removes the now-filled known-gap entry. Top-level `README.md` gets a "Deploy" section linking to the example.

Goal #11 cross-repo convention: the *actual* `values-rdc.yaml` is environment-private and lives in `evoila-bosnia/claude-rdc-hetzner-dc`. This PR ships the public, sanitized template.

## Test plan

- [x] `helm lint deploy/charts/meho/` clean with default values + required overrides.
- [x] `helm install --dry-run=client meho deploy/charts/meho/ --values /tmp/values-rdc-real.yaml` → exit 0 after substituting placeholders.
- [x] `helm template ... -f deploy/values-examples/values-rdc-example.yaml` → fails-loud on `<REPLACE: ...>` for `image.tag` and all three CIDRs.
[...]

PR #2683, opened 2026-07-24
fix(ui): make connector registry Product/Kind filters actually fire (#174)

smujkic94:
## Summary

Fixes **task evoila-bosnia/meho-internal#174** (child of Initiative #172). What was reported as a cosmetic issue turned out to be a **functional bug**: selecting **Product** or **Kind** in the registry filter did nothing to the table.

**Root cause (functional).** The filter form triggered on `hx-trigger="change from:find select"`. Per the htmx docs, `find <selector>` resolves to the **first** matching descendant — so the listener bound only the **Status** select. `change` events from **Product** and **Kind** were never wired to a request, so the table silently didn't narrow. (The topology filter reuses this pattern but has a single select, so it worked there; the registry has three.)

**Fix.** Trigger on `change` at the **form** level — a `change` bubbles from any of the three selects to the form, and `hx-include="closest form"` already sends all three values, so every filter now narrows the table. Server-side filtering (status / product / kind, incl. the `all` sentinel) was already correct and is unchanged.

**Also (the visual half).** Migrated the three filter labels off the daisyUI **v4** dead classes (`form-control` / `label-text`) to `flex flex-col gap-1` + `text-sm font-medium` with `w-full` selects, so the controls render as aligned label-over-select.

## Test plan
- [x] `pytest tests/test_ui_connectors_registry_list.py` — **17 passed** (incl. a new guard: the rendered filter form triggers on `change` and no longer carries `from:find select`)
- [x] Server-side status/product/kind narrowing already covered by existing tests — unchanged.
- [ ] **Manual:** on `/ui/connectors/registry`, changing Product or Kind now re-renders the table (the trigger is client-side htmx; verified by the source-anchored guard + htmx docs, best confirmed with a browser click).

## Notes
Scope grew from the ticket's "presentation" framing to include the functional trigger fix — that's the actual defect the reporter hit. htmx `find` semantics confirmed against https://htmx.org/attributes/hx-trigger/.

Closes evoila-bosnia/meho-internal#174

### DIFF
--- a/backend/src/meho_backplane/ui/templates/connectors/registry.html
+++ b/backend/src/meho_backplane/ui/templates/connectors/registry.html
@@ -75,7 +75,7 @@
     hx-get="/ui/connectors/registry"
     hx-target="#registry-table-body"
     hx-swap="outerHTML"
-    hx-trigger="change from:find select"
+    hx-trigger="change"
     hx-include="closest form"
     hx-push-url="true"
     role="search"
@@ -83,14 +83,14 @@
     onsubmit="return false;"
   >
     <div class="card-body grid gap-3 md:grid-cols-3">
-      <label class="form-control">
-        <span class="label-text">Status</span>
+      <label class="flex flex-col gap-1">
+        <span class="text-sm font-medium">Status</span>
         {# The ``all`` option is a REAL sentinel value, NOT ``value=""`` --
            the backend enum 422s on an out-of-range value and an empty
            string would make the swap silently no-op. #}
         <select
           name="status"
-          class="select select-bordered select-sm"
+          class="select select-bordered select-sm w-full"
           aria-label="Filter by review status"
         >
           {%- set status_options = [
@@ -106,8 +106,8 @@
           {% endfor %}
         </select>
       </label>
-      <label class="form-control">
-        <span class="label-text">Product</span>
+      <label class="flex flex-col gap-1">
+        <span class="text-sm font-medium">Product</span>
         {# Same footgun-safe shape: the "all products" option omits the
            ``product`` filter by submitting an empty string, which the
            handler treats as "no narrowing" -- ``product`` is a plain
@@ -115,7 +115,7 @@
            safe here (it maps to ``None``), unlike the status enum. #}
         <select
           name="product"
-          class="select select-bordered select-sm"
+          class="select select-bordered select-sm w-full"
           aria-label="Filter by product"
         >
           <option value="">All products</option>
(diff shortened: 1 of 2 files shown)


smujkic94:
@coderabbitai review

smujkic94:
> ℹ️ Posted as a comment (self-PR — GitHub blocks a formal *Approve*). Verdict: **APPROVE**.

## /auto-review-pr summary

**Anchor task:** evoila-bosnia/meho-internal#174 — registry Product/Kind filters not firing (child of Initiative #172)
**Surface area:** frontend (Jinja + htmx + Tailwind/daisyUI v5)
**Verdict:** APPROVE
**Head SHA reviewed:** `fd17faa8`
**AI reviewer (CodeRabbit):** received — walkthrough only, no actionable comments / no inline findings
**DCO:** ✅ green

### Root-cause + fix (functional — the real defect)
- **Cause:** the filter form used `hx-trigger="change from:find select"`; htmx's `find <sel>` resolves to the **first** matching descendant (confirmed against https://htmx.org/attributes/hx-trigger/), so only the Status select was wired — Product/Kind `change` events fired nothing.
- **Fix:** `hx-trigger="change"` at the form level. `change` bubbles from any of the three selects to the form; `hx-include="closest form"` already sends all three values. Verified the server handler already filters status/product/kind (incl. the `all` sentinel) and returns the `#registry-table-body` fragment on `HX-Request` — no server change needed.
- **Why it regressed only here:** the topology filter reuses the same `from:find select` pattern but has a single select, so `find`=first happened to be correct there.

### Acceptance criteria
- [x] Status/Product/Kind now each trigger the filter (form-level `change`); server narrowing unchanged.
- [x] Controls render label-over-select — migrated off dead `form-control`/`label-text` to `flex flex-col gap-1` + `text-sm font-medium`, `w-full` selects; no dead form classes remain.
- [x] No functional regression — `pytest tests/test_ui_connectors_registry_list.py` → **17 passed**, incl. a new guard asserting the form triggers on `change` and no longer carries `from:find select`.

### Integration verification
- [x] `pytest` (worktree @ `fd17faa8`) — 17 passed
- [x] Tailwind rebuilt; migration classes present
- ⚠️ not run: real-browser click — the trigger fix is client-side htmx; verified via the cited docs + the source-anchored guard test + the (unchanged, already-tested) server filter path. Recommend a quick manual click before/after merge.

### Findings
None. No blockers, no majors.

### Note
Scope grew from the ticket's "presentation" framing to include the functional trigger fix — that is the defect the reporter actually hit; the ticket's "functional bug = separate follow-up" note is superseded. `Closes evoila-bosnia/meho-internal#174` is cross-repo — close #174 manually on merge.

damir-topic (approved):
Reviewed as part of the open-PR sweep.

Verified the premise rather than assuming it: the repo is on daisyUI 5, and `styles.css` defines no shim for `form-control` / `label-text` / `label-text-alt`, so those v4 classes do compile to zero rules. The migration to explicit `flex flex-col gap-1` + `w-full` is the right fix, and the added regression test asserting the dead classes are absent from rendered output is what keeps it from regressing.

All required checks green. APPROVE.

PR #2696, opened 2026-07-29
fix(ui): tidy runbooks Start-a-Run dialog + format Params JSON (#176)

smujkic94:
## What

Two fixes to the runbooks **Start-a-Run** modal (`_start_modal.html`).

### Layout
The form used daisyUI v4 `form-control` + `label-text`, both removed in v5 (they compile to zero rules), so labels didn't stack over their inputs and helper text sat misaligned. Migrated every field — Template slug, Target, Params, Work ref — to:

- `flex flex-col gap-1` label wrappers
- `text-sm font-medium` labels
- helper notes as `text-xs opacity-60` blocks
- `w-full` on the controls (daisyUI v5 `.input`/`.textarea` cap their own width, so `w-full` is needed to fill the modal)

### Params JSON formatting
The **Params (optional, JSON object)** textarea now pretty-prints its JSON on blur via a small Alpine `formatParams` handler (`JSON.stringify(JSON.parse(v), null, 2)`), so the operator sees a readable indented object instead of a one-line blob. Invalid / in-progress JSON is left exactly as typed — and the handler only reformats a genuine JSON object (arrays, scalars, and `null` are left alone) — so the server's `_parse_json_object` still validates and surfaces the typed error, and formatting never eats an edit.

## Why
Reported in **evoila-bosnia/meho-internal#176** (subtask of the runbooks-view polish initiative **evoila-bosnia/meho-internal#175**).

## Scope / notes
Presentation + client-side formatting only — the start POST, params substitution, CSRF double-submit, and the published-slug datalist are unchanged. The start route echoes no field values on the typed-error re-render, so there is no server-side value to pretty-print today; that input-preservation gap is a separate concern noted for follow-up.

## Test
- `pytest tests/test_ui_runbook_runs_list.py` — 22 passed (adds a guard asserting no dead form classes render and the Params format-on-blur handler is present)
- `pytest tests/test_ui_runbooks_acceptance.py` — 2 passed

## Issue
Closing keywords don't cross repositories, so the reference below is a traceability link — the issue is closed manually on merge.

Closes evoila-bosnia/meho-internal#176

### DIFF
--- a/backend/src/meho_backplane/ui/templates/runbooks/_start_modal.html
+++ b/backend/src/meho_backplane/ui/templates/runbooks/_start_modal.html
@@ -62,15 +62,15 @@
         </div>
       {% endif %}
 
-      <label class="form-control">
-        <span class="label-text">Template slug</span>
+      <label class="flex flex-col gap-1">
+        <span class="text-sm font-medium">Template slug</span>
         <input
           type="text"
           name="template_slug"
           list="runbook-run-published-slugs"
           required
           maxlength="256"
-          class="input input-bordered input-sm font-mono"
+          class="input input-bordered input-sm font-mono w-full"
           placeholder="e.g. drain-node"
           autocomplete="off"
           aria-describedby="runbook-run-start-slug-hint"
@@ -83,7 +83,7 @@
             <option value="{{ slug }}"></option>
           {% endfor %}
         </datalist>
-        <span id="runbook-run-start-slug-hint" class="text-xs opacity-60 mt-1">
+        <span id="runbook-run-start-slug-hint" class="text-xs opacity-60">
           {% if published_slugs %}
             Pick a published template or type its slug.
           {% else %}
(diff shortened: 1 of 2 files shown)


smujkic94:
@coderabbitai review

smujkic94:
Addressed in 239a8ec9 — `formatParams` now reformats only when the parsed value is a non-array, non-null object; arrays, scalars, and `null` are left exactly as typed so the server surfaces the error on submit. Thanks @coderabbitai.

smujkic94:



**Anchor task:** evoila-bosnia/meho-internal#176 — runbooks Start-a-Run dialog: migrate dead daisyUI-v4 classes + format Params JSON
**Surface area:** frontend (Jinja/daisyUI/Alpine template) + backend test
**Verdict:** APPROVE
**Head SHA reviewed:** `239a8ec9`
**AI reviewer (CodeRabbit):** received — 1 minor finding, now resolved + acknowledged on the fix commit

### Acceptance criteria
- [x] Form migrated off dead daisyUI-v4 classes — grep guard in `test_start_modal_renders_with_published_picker` asserts `form-control`/`label-text` absent from the render
- [x] Params textarea formats content as properly-formatted JSON — Alpine `formatParams` pretty-prints on blur; guard asserts `x-on:blur="formatParams($event.target)"` present

### Predicted vs. actual diff
- ✅ `_start_modal.html` — 4 fields migrated to `flex flex-col gap-1` + `text-sm font-medium` + `w-full`; Params format-on-blur added
- ✅ `test_ui_runbook_runs_list.py` — guard assertions added
- 🕳️ Missing: none — no route/JS-module change needed (handler is inline Alpine)

### Integration verification
- [x] `ruff check tests/test_ui_runbook_runs_list.py` — clean
- [x] `pytest tests/test_ui_runbook_runs_list.py tests/test_ui_runbooks_acceptance.py` — 24 passed
- [x] Alpine `formatParams` logic reasoned through: `typeof null === 'object'` handled by explicit `=== null`; `Array.isArray` excludes arrays; scalars fail `typeof !== 'object'` → only genuine JSON objects reformatted, invalid input left for the server to reject
- n/a migrations / docker / agent-eval — presentation-only template change

### Findings
None blocking. CodeRabbit's sole finding (💡 minor — "format only JSON objects") was addressed in `239a8ec9` before this review; the thread is resolved + outdated.

- Adopted: 1 (JSON-object-only formatting — fixed in `239a8ec9`)
- Disputed: 0
- Out of scope: 0

### Docs
- n/a — no `docs/codebase/` artifact covers this modal's markup; presentation-only change

### What worked well
- Format-on-blur deliberately leaves invalid/in-progress JSON untouched so the server (`_parse_json_object`) stays the single source of validation truth — formatting never eats an edit.
- Source-anchored guard test locks in the dead-class migration so a future daisyUI regression is caught in CI.

- The start route echoes no field values on the typed-error re-render, so a rejected submit loses the operator's typed Params/Target. Out of scope here; worth a follow-up ticket for input preservation.

damir-topic (approved):
Reviewed as part of the open-PR sweep.

Verified the premise rather than assuming it: the repo is on daisyUI 5, and `styles.css` defines no shim for `form-control` / `label-text` / `label-text-alt`, so those v4 classes do compile to zero rules. The migration to explicit `flex flex-col gap-1` + `w-full` is the right fix, and the added regression test asserting the dead classes are absent from rendered output is what keeps it from regressing.

All required checks green. APPROVE.

damir-topic:
Non-blocking note on `formatParams`, for whenever this area is next touched:

The textarea carries `maxlength="16384"`, but `maxlength` only constrains user typing — setting `.value` programmatically bypasses it. So a near-limit minified paste can pretty-print past the cap and submit oversized. The server still rejects it via `_parse_json_object`, so this is a worse-error-message issue rather than a correctness one. Not worth holding the PR for.

The handler's bail-out logic (invalid JSON, arrays, scalars, `null` all left exactly as typed) is the right call — it keeps formatting from tidying invalid input into something that looks valid.

Issue #2140, opened 2026-07-05
Task: Run database-migration tests only when migration files change (relocate to tests/migrations/ + conditional CI job)


damir-topic:
Related: evoila-bosnia/meho-internal#26 (closed — G2.3 Audit log + DB migrations, the initiative that created these tests), evoila-bosnia/meho-internal#30 (closed — forward-compat testcontainers test = `backend/tests/test_migration_rollback.py`, one of the 3 container tests this Task relocates)

Parent goal: #221

> Moved from `evoila-bosnia/meho-internal#127` (CI-infra work lands on `evoila/meho`; re-homed under Goal #221 per the board-cleanup review, 2026-07-05).

## Target

Stop running the database-migration test suite on every PR. Run it only when a PR actually changes migration-related files. Two coupled changes, landed together in one PR on `evoila/meho`:

- **A — Relocate migration tests into their own folder and exclude it from the unit lane.** Move the 30 migration/alembic test files out of the flat `backend/tests/` into a new `backend/tests/migrations/`, mirroring how `backend/tests/integration/` is already excluded. Add `--ignore=tests/migrations` to the required unit lane (`python-lint-test`) in `.github/workflows/ci.yml`.
- **B — Add a conditionally-run, required `python-migration-tests` job.** Using a `changes`-job + job-level `if:` pattern (added net-new — see Context for why it beats `on.paths` for a *required* check), run `pytest tests/migrations/` (including the 3 Postgres-container tests) only when migration files changed. Make it a required check.

## Context

The CI test suite is slow and expensive on the self-hosted runners and sometimes does not finish. The migration tests run on **every** PR inside the required unit lane, even though ~95% of PRs touch no migration files.

A migration **squash** (collapsing the 50 `backend/alembic/versions/*.py` into one) was evaluated and **rejected**:
- It saves ~0 test time. `backend/tests/conftest.py` (the session-scoped `_schema_template_db` fixture) already runs `alembic upgrade head` **once per xdist worker** and byte-copies the result into every other test, so the chain is not replayed per test today.
- It deletes coverage. The per-migration tests (e.g. `0011` backfill, `0038` product-split reconcile, `0048` seed, `0049` retire) assert migration-specific data-transformation logic; squashing removes their subject.
- It risks the production rollback contract (#1607): customer DBs sit at the current head revision; collapsing history makes those revisions unknown to the image and undermines the DB-ahead-tolerant readiness probe.

[...]

PR #2703, opened 2026-07-29
fix(ci): unblock merge gates — path-filtered license checks + Dependabot login

damir-topic:
Found while sweeping the 27 open PRs: **26 of them are unmergeable, for two unrelated CI-config reasons.** Neither is about the code in those PRs.

## 1. Required checks gated by `on.pull_request.paths` — blocks nearly every PR

`Python License Check` and `NPM License Check` are required status checks in branch protection. But `dependency-license-check.yml` gated them with `on.pull_request.paths` limited to dependency manifests.

GitHub leaves a required context **Pending forever** when its whole workflow is skipped by a path filter. So any PR that doesn't touch a `pyproject.toml` / `package.json` / `package-lock.json` can never satisfy branch protection. Confirmed against the 6 green UI PRs — those two contexts are simply absent from their check rollups, while everything else is green:

```
$ gh pr view 2697 --json statusCheckRollup   # → no "License" context at all
SUCCESS  DCO, Python (ruff + mypy + pytest), Go, Helm, Semgrep SAST, TruffleHog, ...
(absent) Python License Check
(absent) NPM License Check
```

The repo already solved this exact problem and wrote down why — `ci.yml`'s `changes` job, added in #2140:

> **WHY a `changes`-job + job-level `if:` and NOT `on.paths` (#2140):** GitHub treats a job skipped by its own `if:` as **Success** (it satisfies a required status check), whereas a whole workflow skipped by a path filter stays **Pending** and blocks merge.

`dependency-license-check.yml` was never converted. This PR converts it, mirroring that pattern exactly. **Which manifests trigger a real license run is unchanged** — the same file set moved from the trigger's `paths:` into the `changes` job's diff check. The regex is anchored with `(^|/)` so `notes-package.json` doesn't false-positive; verified against both positive and negative paths.

There's a merge-queue-shaped assumption in the old comments (`merge_group:` was added so these contexts report against the queue's merge commit), but `required_merge_queue` is `null` on `main` and `allow_auto_merge` is `false` — no queue was ever enabled, so nothing was satisfying these contexts on PRs.

## 2. Dependabot runs can't reach the Actions secret store — blocks all 18 Dependabot PRs

Every Dependabot PR fails `Python (ruff + mypy + pytest)` and `Python (integration testcontainers)` at the same step:

```
##[error]Username and password required
```

That's `docker/login-action`. The job log shows `Secret source: Dependabot` — Dependabot-triggered runs read the *Dependabot* secret store, not the Actions one, so `DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN` resolve empty, and the action **hard-fails rather than no-opping**. Nothing to do with the dependency bumps themselves.

All four login steps now carry `if: ${{ secrets.DOCKERHUB_USERNAME != '' }}`.

Skipping the login is safe: every testcontainers image resolves through the `harbor.evba.lab` proxy cache (`MEHO_TEST_PGVECTOR_IMAGE` and friends), so unauthenticated runs still pull. What's forfeited is the Docker Hub rate-limit headroom the login buys — relevant mainly to the integration job, which pulls the most images.

**Follow-up worth doing separately:** mirror `DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN` into the repo's Dependabot secrets (Settings → Secrets and variables → Dependabot). That restores the headroom and makes this guard a no-op. The guard is still the right thing to keep — it's what stops a missing secret from hard-failing a whole job.

## Verification

- Both workflows parse (`YAML.load_file` round-trip); job graphs are `changes → python-licenses / npm-licenses` and unchanged for `ci.yml`.
- Manifest-detection regex tested against positive paths (`backend/pyproject.toml`, `pyproject.toml`, `frontend/package.json`, `backend/package-lock.json`, the workflow itself) and negatives (`docs/pyproject.toml.md`, `notes-package.json`, `.github/workflows/ci.yml`, template/test files).
- This PR touches `dependency-license-check.yml`, which is in the detected set — so the license jobs run for real here rather than skipping. That makes it self-validating: if the `changes` job is wrong, this PR's own checks show it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CI reliability when Docker Hub credentials are unavailable.
  * Prevented required license checks from remaining pending when dependency manifests are unchanged.
  * Reduced unnecessary license scans for pull requests without dependency changes.

* **Documentation**
  * Added unreleased changelog entries describing the CI workflow fixes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -247,6 +247,20 @@ jobs:
         # vault custody — see meho-internal#74 for the regression
         # context this addresses. Long-term, the in-cluster proxy
         # cache from claude-rdc-hetzner-dc#463 retires this step.
+        #
+        # The `if:` guard is what keeps Dependabot PRs green. A
+        # Dependabot-triggered run reads the *Dependabot* secret store,
+        # not the Actions one, so `secrets.DOCKERHUB_USERNAME` is empty
+        # there — and `docker/login-action` hard-fails with "Username and
+        # password required" instead of no-opping, which failed this job
+        # (and every other job carrying this step) on all ~18 open
+        # Dependabot PRs. Skipping the login is safe: every image above
+        # resolves through the harbor.evba.lab proxy cache, so an
+        # unauthenticated run still pulls — it only gives up the Docker
+        # Hub rate-limit headroom the login buys. Mirroring
+        # DOCKERHUB_USERNAME / DOCKERHUB_TOKEN into the repo's Dependabot
+        # secrets restores that headroom and makes this guard a no-op.
+        if: ${{ secrets.DOCKERHUB_USERNAME != '' }}
         uses: docker/login-action@af1e73f918a031802d376d3c8bbc3fe56130a9b0  # v4.4.0
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
@@ -564,6 +578,9 @@ jobs:
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
 
       - name: Docker Hub login (for testcontainers pulls)
+        # Empty-secret guard — see the python-tests job's login step for
+        # why Dependabot runs need it.
+        if: ${{ secrets.DOCKERHUB_USERNAME != '' }}
         uses: docker/login-action@af1e73f918a031802d376d3c8bbc3fe56130a9b0  # v4.4.0
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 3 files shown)


damir-topic:
Superseded by #2704.

Two things were wrong here. The DCO failure was mechanical — the commit was authored with a machine-local email (`damir-topic@Damirs-MacBook-Air.local`) and carried no `Signed-off-by`.

The substantive one: the Dependabot guard was written as `if: ${{ secrets.DOCKERHUB_USERNAME != '' }}`, and the `secrets` context is **not available in a step-level `if:`** — GitHub allows it in `with:` / `env:` / `run:` only. So rather than guarding the step, it made the whole workflow invalid: the CI run failed outright with *"This run likely failed because of a workflow file issue"*, which is why no `ci.yml` check reported on this PR at all. #2704 reads the secret through a job-level `env` var instead, which is evaluatable there.

Worth recording that part 1 of the fix was already proven on this PR before that bug surfaced: `Detect manifest changes`, `Python License Check`, and `NPM License Check` all reported SUCCESS here, where previously they never reported at all.

PR #2723, opened 2026-07-31
docs(site): first Do-real-work guides — targets & secrets, first operations, sensors quickstart

damir-topic:
## Summary

- Lands the first three *Do real work* task guides into the #2670 docs-site scaffold: **Register targets and secrets** (`docs-site/guides/targets-and-secrets.md`), **Run your first operations** (`docs-site/guides/first-operations.md`), and **Watch your estate with sensors** (`docs-site/guides/sensors-quickstart.md`), plus a rewritten guides landing page that sequences them and reserves the approvals & break-glass slot (#2669).
- Every command, tool name, flag, error string, and role requirement was verified against the shipped surfaces (Go CLI command definitions, MCP tool registrations, dispatcher/error-builder strings, checks/rollup code) — not the aspirational spec. The guides say plainly what does *not* exist: no MCP tools for target registration, no CLI preview verb, `result_query` as the only shipped read-back tool, no dashboard MCP tools, no sensor update/pause verbs.
- Ports the load-bearing operation-ladder discipline from the dogfood skill contracts into the product docs (discover → groups → search → inspect `safety_level`/`requires_approval` → preview → call), and carries the honest per-backend caveats: the sensors background-credential matrix (Vault `checkRunner.*` privilege warning, GSM WIF cases), the per-operator-WIF probe limitation, and the open #2668 tracker for long-unattended operation.
- Each guide ends with a "What can go wrong here" table sourced from the real error strings (`secret_ref_outside_tenant_scope`, the API-path-shaped-ref rejection, `connector_vault_forbidden`, `connector_auth_failed`, `connector_tls_verify_failed`, `unknown_connector` vs `connector_not_ingested`, `handle_not_found`, the sensor guard codes) rather than invented FAQs.

## Test plan

- [x] `uv run --project backend --no-sync mkdocs build --strict` — exit 0 (the same gate `docs-site.yml` runs on PRs; strict turns broken links / missing-nav pages into failures)
- [x] `python3 .claude/hooks/code-quality.py --diff` — exit 0
- [x] AC 1 — three guides live under *Do real work*, each written to be executable start-to-finish by someone who has only read *Install & operate* + *Connect clients*: prerequisites and role requirements (`tenant_admin` vs `operator`) are stated up front on every guide; all commands verified against `cli/internal/cmd/*` definitions and MCP tool schemas. (Execution against a live fresh deploy is what eval S2/S3/S6 will measure — that program is the acceptance instrument, per #2665.)
[...]

Issue #2721, opened 2026-07-31
Task: per-dashboard investigator prompt + emailed findings


damir-topic:
Parent goal: #221
Parent initiative: #2716
Depends-on: #2717, #2719

## Summary

Let an operator attach a **custom investigation prompt to a Dashboard** and get the **investigator's finding by email**: a new `investigator_prompt` column threaded into the briefing the diagnose-only investigator (#2507) already receives, and — after the finding is awaited — a deterministic email of the `ChecksFinding` to the Dashboard's `notify_email`. **Binding decision:** the *wrapper* sends the email, not the agent — the investigator keeps its diagnose-only, no-dispatcher contract and gets no mail tool; an agent that should send ad-hoc mail does so in its own runs via `call_operation` on the mail connector, which is what makes the transport reusable. Scope item 5 of Initiative #2716.

## Current state

- The investigator fire path: `investigate_on_transition` → `_fire_investigation` with `briefing=_build_briefing(dashboard, group)` (`checks/investigate.py:875-883`), invoked via `AgentInvoker.run_scheduled(name, briefing, ...)` (`:927-933`); the briefing is entirely server-built from the transition snapshot; there is no operator-supplied content.
- The finding is awaited (`_await_finding`, `:887-889`) and persisted only to `agent_run` + tenant memory as noise-suppression (`_persist_finding`, `:890`); it reaches no operator channel.
- The diagnose-only contract is load-bearing and documented: the module never imports the operations dispatcher; agent write attempts park as `ApprovalRequest` (`investigate.py:4-79` module docstring).
- After #2719, `check_dashboards` carries `notify_email`/`notify_min_state` and the notify hook exists; after #2717, `send_email()` exists.
- `check_dashboards` is set-at-create-only (posture from #2506).

## Desired state

1. **Migration** (one file): `check_dashboards.investigator_prompt` (Text NULL, bounded — CHECK or app-level cap at 4 KiB). NULL ⇒ today's behavior bit-for-bit.
2. **Briefing threading**: `_build_briefing` appends, when set, a clearly-delimited operator-context section (e.g. `## Operator instructions for this dashboard`) after the server-built snapshot — appended, never replacing, so the deterministic transition facts always lead and prompt content cannot suppress them.
[...]

Issue #2720, opened 2026-07-31
Task: checks.transition broadcast events + checks op-class on the tenant feed


damir-topic:
Parent goal: #221
Parent initiative: #2716
Depends-on: #2719

## Summary

Publish a **`checks.transition` broadcast event** on every claimed Dashboard state transition (both directions) and teach `classify_op` a `checks.` op-class, so feed watchers (`broadcast_recent` / `broadcast_watch` / `/ui/broadcast/stream?op_class=checks`) see check state changes the way they already see `approval.*` lifecycle events. Scope item 3 of Initiative #2716.

## Current state

- `checks/` imports nothing from `broadcast/` (verified by grep). Sensor evaluations reach the feed only as generic per-dispatch `operation` events attributed to the sensor's `identity_sub` — there is no state-transition event.
- The precedent to copy: `publish_approval_event` (`operations/approval_queue.py:1228`) publishes approval lifecycle `BroadcastEvent`s whose op-ids (`approval.pending` etc.) `classify_op` maps to op_class `approval` via a prefix branch (`broadcast/events.py:478-482`); the approvals console bell subscribes to `/ui/broadcast/stream?op_class=approval`.
- After #2719, `checks/notify.py` receives every claimed transition (worsening and recovery) with the dashboard snapshot — the natural single publish point, exactly-once per edge via the existing CAS.
- Publish posture reference: audit-derived broadcast publishing is fail-open (`operations/_audit.py` `audit_and_broadcast_safe`); a Valkey outage must never break the persist seam.

## Desired state

1. `publish_check_transition_event(...)` (new, in `checks/notify.py` or a small `checks/broadcast.py`): builds a `BroadcastEvent` (kind `operation`) with op_id **`checks.transition`** and a structured payload — `dashboard_id`, `dashboard_name`, `previous_state`, `new_state` — attributed to the tenant; fire-and-forget with a broad fail-open guard.
2. `classify_op` gains a `checks.` prefix branch returning op_class **`checks`** (three lines beside the `approval.` branch, `broadcast/events.py:478-482`), docstring + doctest updated in the same style.
3. Called from the #2719 transition hook for every claimed edge — recovery visibility is deliberate (a watcher who saw `degraded` must see the clear).
4. No new consumer in this task: existing `broadcast_recent`/`broadcast_watch` filters and the SSE stream pick the class up immediately.

## Implementation sketch

[...]

Issue #2719, opened 2026-07-31
Task: dashboard email notifications on check state transitions, recovery included


damir-topic:
Parent goal: #221
Parent initiative: #2716
Depends-on: #2717

## Summary

Wire **email notification onto Dashboard state transitions**: two new columns on `check_dashboards` (`notify_email`, `notify_min_state`), a `checks/notify.py` hook on the existing CAS claim-win branch that fires on **any** transition crossing the configured threshold — **recovery/all-clear included** — and delivers via the mail connector's shared `send_email()` transport. The investigator's worsening-only gate is untouched. Scope item 2 of Initiative #2716.

## Current state

- The transition detector already observes both edge directions: `_evaluate_and_claim` (`checks/investigate.py:395-428`) folds the membership under a per-(tenant,dashboard) `pg_advisory_xact_lock` and CASes `last_rollup_state` on **any** `current != previous_memo` (`:417-423`); `rowcount==1` is the replica-safe exactly-once claim (`_claim_rollup_transition`, `:461-486`). But the function returns a snapshot only for `worsening and claimed and non_green` (`:426-428`) — recovery claims are won and then discarded.
- The only consumer today is the diagnose-only investigator fire (`investigate.py:875-898`); no operator channel exists.
- `check_dashboards` columns today: id/tenant_id/name/description/`last_rollup_state` memo/created_by_sub/timestamps (`db/models.py:6343-6418`); memberships and posture are **set-at-create-only** (edit = delete + recreate).
- Dashboard create surface: REST `POST /api/v1/checks/dashboards` (`api/v1/checks_dashboards.py`), CLI `meho dashboard create` (`cli/internal/cmd/dashboard/`), no MCP dashboard tools (verified — MCP has only `meho.sensor.*`).
- Mail transport (dependency): `connectors/mail/transport.py` `send_email()` with the return-failures contract, from task #2717.
- Migration head at time of writing: `0067` (`backend/alembic/versions/`); take the next free number at merge time.

## Desired state

1. **Migration** (one file): `check_dashboards` gains `notify_email` (Text NULL — one recipient address; NULL ⇒ notifications off for this Dashboard) and `notify_min_state` (Text NOT NULL default `'critical'`, CHECK `IN ('degraded','critical')`). Set at create only, posture parity with membership.
[...]

PR #2724, opened 2026-07-31
feat(connectors): mail.* typed connector — SMTP send as a reusable, agent-callable operation

### LINKED ISSUE #2717: Task: mail.* typed connector — SMTP send as a reusable, agent-callable operation
Parent goal: #221
Parent initiative: #2716

## Summary

Add a **`mail.*` synthetic typed-op connector** with one operation, **`mail.send`** — a reusable SMTP send that (a) agents and operators reach through the normal `call_operation` dispatch path, and (b) the checks notifier (follow-up task) imports directly as a shared transport function. Modelled 1:1 on the `net.*` synthetic product. Scope item 1 of Initiative #2716.

## Current state

- **No email capability exists anywhere in `backend/src`**: greps for `smtp|sendmail|mailer` hit only Keycloak-connector redaction of realm `smtpServer` config (`connectors/keycloak/redaction.py:70-87`) and incidental text. `backend/pyproject.toml` has no SMTP library; `pydantic[email]` (v2, `EmailStr`) is already a dependency (`backend/pyproject.toml:11`).
- **The mould to copy** — `net.*` (`connectors/net/ops.py:4-43`): a synthetic typed-op product with no vendor connector behind it; ops registered via `register_typed_operation` (`ops.py:602-673`) under natural key `(product="net", version="1.x", impl_id="net-probe")` → wire `connector_id` `net-probe-1.x`; module-level handlers so the dispatcher resolves `connector_instance=None`, `target=None`; an env-allowlist floor where an empty allowlist makes the connector inert (`connectors/net/allowlist.py`, `MEHO_NETDIAG_PROBE_ALLOWLIST`); the return-failures-as-product contract (a refused/failed probe returns `{"connected": false, "reason": ...}` with dispatch `status="ok"`).
- Settings conventions: secret fields use `Field(default="", repr=False)` (`settings.py:1312` `check_runner_client_secret`); feature blocks carry a `#:`-comment with the owning issue.

## Desired state

1. **Package** `backend/src/meho_backplane/connectors/mail/` with:
   - `transport.py` — `async def send_email(*, to: list[str], subject: str, body: str) -> MailSendResult`: stdlib `smtplib.SMTP`/`SMTP_SSL` + `email.message.EmailMessage`, run via `asyncio.to_thread` (smtplib is blocking; no new dependency). STARTTLS when `mail_smtp_starttls` is set (`SMTP.starttls()` then re-`ehlo()`), `SMTP.login()` only when a username is configured, `SMTP.send_message()` for delivery. This function is the single implementation both the dispatch handler and the checks notifier call.
   - `allowlist.py` — recipient-domain/address allowlist on the `net` allowlist mould; **empty allowlist ⇒ every send refused (connector inert)**. Checked before any SMTP connection.
   - `ops.py` — `mail.send` handler + `register_mail_typed_operations` registrar; natural key `(product="mail", version="1.x", impl_id="mail-smtp")` → `connector_id` `mail-smtp-1.x`; parameter schema `{to: [email], subject: str, body: str}` (bounded lengths); registered with `safety_level="caution"` + `requires_approval=False` — outward-facing side effect, so operators auto-run while agent calls stay policy-gated; the recipient allowlist is the hard floor either way.
2. **Settings**: `mail_smtp_host` (default `""` ⇒ transport unconfigured, send refused with `reason="smtp_unconfigured"`), `mail_smtp_port` (int, default 587), `mail_smtp_starttls` (bool, default True), `mail_smtp_username` (default `""`), `mail_smtp_password` (`repr=False`), `mail_from` (default `""`), `mail_recipient_allowlist` (default `""` ⇒ inert).
3. **Return-failures contract** (mould parity): allowlist-refused, unconfigured, connect-failed, auth-failed, recipients-refused each return `{"sent": false, "reason": <code>}` with dispatch `status="ok"`; only unexpected bugs raise. `SMTPException` subclasses map to stable reason codes (`smtp_connect_error`, `smtp_auth_error`, `smtp_recipients_refused`, `smtp_error`).
4. Registrar wired the same way `net`'s is (package `__init__` via `register_typed_op_registrar`), so the op lands in `endpoint_descriptor` and is discoverable via `search_operations` / callable via `call_operation`.

## Implementation sketch

- Copy the `net` package layout; keep handlers module-level. No `Target` involvement — recipients are params.
- Password never logged: structlog events carry host/port/reason only; the audit `raw_payload` carries `to`/`subject` (recipient addresses are the audit answer to "who was mailed"), never the body of credentials.
- Unit tests with a stubbed SMTP (monkeypatched `smtplib.SMTP`): allowlist refusal, unconfigured refusal, STARTTLS + login called when configured, `send_message` invoked with the built `EmailMessage`, each `SMTPException` subclass → its reason code, success path `{"sent": true}`.
- CHANGELOG `[Unreleased]` bullet under `### Mail connector — SMTP send as a typed operation`.

## Acceptance criteria

- [ ] `uv run pytest backend/tests/test_connectors_mail.py -q` green, covering by name: empty-allowlist refusal; unconfigured-host refusal; STARTTLS+login sequencing; per-`SMTPException`-subclass reason codes; success.
- [ ] `grep -n 'safety_level="caution"' backend/src/meho_backplane/connectors/mail/ops.py` hits; `grep -rn "smtplib" backend/src/meho_backplane/connectors/mail/transport.py` hits and `grep -n "aiosmtplib\|smtp" backend/pyproject.toml` shows **no new dependency**.
- [ ] A route/integration test proves `call_operation` dispatch of `("mail-smtp-1.x", "mail.send")` with a stubbed transport returns `status="ok"` and an audit row is written (shared-dispatcher parity).
- [ ] `mail_smtp_password` excluded from repr: `grep -n 'mail_smtp_password' backend/src/meho_backplane/settings.py` shows `repr=False`.
- [ ] No migration: `git diff --stat` touches nothing under `backend/alembic/versions/`.
- [ ] `CHANGELOG.md` `[Unreleased]` gains the unique header above; full backend suite green (`uv run pytest`, backend/).

## Out of scope

- The checks notifier and its `check_dashboards` columns (follow-up task in this Initiative).
- Templated/HTML email, attachments, per-tenant SMTP config — one deployment-level SMTP block, plain-text body.
- Slack/PagerDuty/webhook transports; MCP/CLI convenience verbs for sending mail (dispatch + generated client suffice).
- Inbound mail of any kind.

## References

- Initiative #2716 (parent). Mould: `backend/src/meho_backplane/connectors/net/ops.py:4-43,602-673`; `connectors/net/allowlist.py`; registration seam `operations/typed_register.py`.
- Settings conventions: `backend/src/meho_backplane/settings.py:1311-1312` (`repr=False` secret), `:1063-1068` (feature-knob comment style).
- Python 3.14 `smtplib` — `SMTP`, `SMTP_SSL`, `starttls()`, `login()`, `send_message()`, `SMTPException` hierarchy: https://docs.python.org/3.14/library/smtplib.html (verified 2026-07-31; no 3.14 breaking changes; `keyfile`/`certfile` gone since 3.12 — use `ssl.SSLContext`).
- `email.message.EmailMessage`: https://docs.python.org/3.14/library/email.message.html
- Empirical anchor: Prometheus Alertmanager's `email_config` is plain SMTP with STARTTLS/auth — the industry-standard alert-email shape: https://prometheus.io/docs/alerting/latest/configuration/#email_config

damir-topic:
## Summary

- Add the synthetic `mail.*` typed connector on the `net.*` mould: one op, `mail.send` (`connector_id` `mail-smtp-1.x`), registered via `register_typed_operation` with a module-level handler (`connector_instance=None`, `target=None` — recipients are params, not a `Target`).
- `transport.send_email()` is the single shared implementation (stdlib `smtplib` + `email.message.EmailMessage` via `asyncio.to_thread`, **no new dependency**) that both the dispatch handler and the checks notifier (#2719) call — the recipient-allowlist floor and the return-failures contract hold on every send path.
- `MAIL_RECIPIENT_ALLOWLIST` (full addresses and/or domains) is the inverted-default hard floor: empty ⇒ every send refused, connector inert. Unparseable recipients are refused, which also makes SMTP envelope injection structurally impossible; a CR/LF subject is rejected by `EmailMessage`'s own header guard.
- Deployment-level SMTP settings block (`MAIL_SMTP_HOST/PORT/STARTTLS/USERNAME/PASSWORD`, `MAIL_FROM`) with `repr=False` on the password; port 465 selects implicit TLS (`SMTP_SSL`, RFC 8314), other ports upgrade via `starttls()` + re-`ehlo()` (RFC 3207); `login()` only when a username is configured.
- Registered `safety_level="caution"` + `requires_approval=False`: operators (non-agent principals) auto-run; agent principals land on the policy gate's caution ⇒ needs-approval default with a needs-approval ceiling. Refused/unconfigured/failed sends return `{sent: false, reason}` with dispatch `status="ok"`; stable reason codes `not_in_recipient_allowlist` / `smtp_unconfigured` / `smtp_connect_error` / `smtp_auth_error` / `smtp_recipients_refused` / `smtp_error`. Audit `raw_payload` carries `to`/`subject`, never the body.
- New `docs/codebase/connectors-mail.md`; CHANGELOG `[Unreleased]` entry under `### Added — Mail connector — SMTP send as a typed operation`.

## Test plan

- [x] `uv run mypy src/` (strict) — clean except pre-existing `net/icmp.py` `MSG_ERRQUEUE` (Linux-only constant flagged on macOS; file untouched, green in Linux CI)
- [x] `uv run pytest tests/test_connectors_mail.py -q` — 29 passed, covering by name: empty-allowlist refusal, unconfigured-host refusal, STARTTLS+login sequencing, per-`SMTPException`-subclass reason codes, success
- [x] Dispatch parity: `test_dispatch_mail_send_returns_ok_and_writes_audit_row` proves `("mail-smtp-1.x", "mail.send")` through the real dispatcher returns `status="ok"` and writes an audit row whose `raw_payload` has `to`/`subject` and no body
- [x] `grep -n 'safety_level="caution"' .../mail/ops.py` hits; `grep -rn smtplib .../mail/transport.py` hits; `grep -n "aiosmtplib\|smtp" backend/pyproject.toml` — no new dependency
- [x] `mail_smtp_password` carries `repr=False` in `settings.py`
- [x] No migration: nothing under `backend/alembic/versions/` in the diff
- [x] Full backend suite (`pytest -n 3 --dist loadscope --ignore=tests/integration --ignore=tests/migrations`) green apart from two pre-existing env-dependent `net.*` failures (`test_trace_reaches_loopback`, `test_chosen_resolver_queries_that_nameserver`) — both fail identically on untouched `main` on this macOS host (real-socket tests; green in Linux CI)
- [x] `python3 .claude/hooks/code-quality.py --diff` — 0 blockers (2 warnings: pre-existing `settings.py` size; `send_email` at 71 lines, warn-level)

## Out of scope

- Checks notifier + `check_dashboards` columns (#2719), `checks.transition` broadcast events (#2720), in-band advisory (#2718), investigator prompt (#2721).
- HTML/templated mail, attachments, per-tenant SMTP config, Slack/PagerDuty/webhook transports, inbound mail, MCP/CLI convenience verbs.
- Adjacent finding (not changed): `classify_op("mail.send")` falls through to `other` (full-detail broadcast — payload is to/subject, audit-visible by design); a dedicated classification branch was deliberately not added here.

Closes #2717

## Summary by CodeRabbit

* **New Features**
  * Added a typed `mail.send` connector for sending plain-text email through configurable SMTP.
  * Added recipient and domain allowlisting, optional authentication, TLS support, and approval safeguards.
  * Added structured failure responses and privacy-conscious auditing of recipients and subjects.
  * Added deployment settings for SMTP configuration and sender details.

* **Documentation**
  * Added documentation covering configuration, safety behavior, error handling, and known limitations.

* **Tests**
  * Added comprehensive coverage for sending, validation, security controls, and SMTP failures.

---

Addressed the review findings from https://github.com/evoila/meho/pull/2724#pullrequestreview-4832689478:

- **B1** `df71e2f2` — DCO remediation commit for `21c57ff4` (retroactive `Signed-off-by`, no history rewrite, no force-push). Every commit added in this iteration is signed off.
- **B2** `9b386f99` — `subject` gains a schema `pattern`, so a line-break-bearing subject is rejected as `invalid_params` by the dispatcher instead of raising `ValueError` out of `send_email` as a `connector_error`. The transport docstring and `docs/codebase/connectors-mail.md` no longer document the hole as intended behaviour.
- **M1** `6cca4894` — `_send_sync` tracks channel encryption and refuses `login()` on a cleartext socket with a new stable reason code `smtp_auth_requires_tls`, documented in the transport module docstring, the op's response schema + `llm_instructions`, and the docs reason/config tables.
- **m1** `e722c441` — `parse_recipient_allowlist` now rejects `foo@`, `@`, and dots-only entries instead of silently keeping them.
- `26d4fc07` — merged `main` in after #2723 landed and conflicted on the `## [Unreleased]` CHANGELOG section. Both entries kept; merged rather than rebased because the branch is published and carries a posted review.

**B1 could not be closed — see [this comment](https://github.com/evoila/meho/pull/2724#issuecomment-5148340889).** The remediation commit is correct and pushed, but `dcoapp/app` ignores remediation commits unless `.github/dco.yml` on the *default branch* sets `allowRemediationCommits.individual: true`, and no such config exists in this repo. Closing it needs either that one-file PR on `main` or a rebase + force-push, neither of which an autonomous run may decide.

### Deviation from the punch list (one, deliberate)

**B2's pattern is wider than the `^[^\r\n]+$` the review specified.** `EmailMessage` guards header assignment with `len(value.splitlines()) > 1`, and `str.splitlines()` breaks on VT, FF, the three information separators, NEL, and U+2028/U+2029 in addition to CR/LF — verified empirically on the worktree's CPython 3.14.6. `^[^\r\n]+$` would have closed the reported case while leaving nine other characters able to raise `ValueError` out of the handler. The shipped pattern is the exact complement of the stdlib guard, and `test_subject_pattern_is_the_exact_complement_of_the_stdlib_guard` sweeps the whole Unicode range to assert no admitted subject can make `EmailMessage` raise — so a future interpreter widening `splitlines()` is a red build rather than a silently reopened contract.

### Test plan

- [x] `ruff format --check src/ tests/` — 1375 files already formatted
- [x] `mypy src/meho_backplane/ --ignore-missing-imports` — only the pre-existing macOS-only `net/icmp.py:208 MSG_ERRQUEUE` error (Linux CI lane is green)
- [x] `pytest tests/test_connectors_mail.py` — 53 passed (was 29; +24 covering the four findings)
- [x] `pytest -n 3 --dist loadscope --ignore=tests/integration --ignore=tests/migrations tests/` — 10737 passed, 193 skipped, exit 0
- [x] `code-quality.py` — 0 blockers

### DIFF
--- a/backend/src/meho_backplane/connectors/mail/allowlist.py
+++ b/backend/src/meho_backplane/connectors/mail/allowlist.py
@@ -1,0 +1,157 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""Recipient allowlist for the ``mail.*`` connector.
+
+``mail.send`` gives the backplane (and, through the normal policy path,
+an agent) an outward channel: it delivers operator-authored text to an
+arbitrary mailbox. Unbounded, that is an exfiltration and spam primitive.
+:attr:`~meho_backplane.settings.Settings.mail_recipient_allowlist`
+(``MAIL_RECIPIENT_ALLOWLIST``) is the single floor that scopes it — the
+same **inverted-default** shape as the ``net.*`` probe allowlist
+(``connectors/net/allowlist.py``): the parsed set is the **whole
+permitted recipient space**, and an **empty** value means **deny
+everything**, so the connector is inert until an operator deliberately
+opts recipients in. The floor applies to every caller of
+:func:`~meho_backplane.connectors.mail.transport.send_email` — the
+dispatch handler *and* the checks notifier (#2719) — so no code path
+can mail an unlisted recipient.
+
+Two entry shapes, comma-separated:
+
+* a full address (``oncall@example.com``) — matches that mailbox only;
+* a domain (``example.com``, leading ``@`` accepted) — matches every
+  mailbox at that domain.
+
+Matching is case-insensitive on the whole address. That folds the
+local part too — RFC 5321 technically leaves local-part case to the
+receiving host, but no practical mail system distinguishes
+``OnCall@`` from ``oncall@``, and a case-sensitive floor would turn
+an operator's capitalization habit into a silent refusal.
+
+There is deliberately no pattern/glob dimension (#1177: one closed-set
+config, no DSL) and no per-tenant allowlist — the SMTP block is
+deployment-level (one MTA, one floor), per task #2717's scope.
+"""
+
+from __future__ import annotations
+
+from meho_backplane.settings import get_settings
+
+__all__ = [
+    "RecipientNotAllowedError",
+    "assert_recipient_allowed",
+    "parse_recipient_allowlist",
+]
+
+
+class RecipientNotAllowedError(ValueError):
+    """A recipient address is not inside ``MAIL_RECIPIENT_ALLOWLIST``.
+
+    Subclasses :class:`ValueError`. The transport catches it and converts
+    it into the connector's structured refusal (``{"sent": false,
+    "reason": "not_in_recipient_allowlist"}``, dispatch ``status="ok"``)
+    — a refused send is a normal result, never a ``connector_*`` error
+    (the return-failures contract shared with ``net.*``).
+    """
+
+
+def _malformed_entry(entry: str) -> ValueError:
+    """Build the fail-loud error for an allowlist token that cannot be used."""
+    return ValueError(
+        f"MAIL_RECIPIENT_ALLOWLIST entry {entry!r} is not a valid "
+        "address or domain; fix the deployment configuration"
+    )
+
+
+def parse_recipient_allowlist() -> tuple[frozenset[str], frozenset[str]]:
+    """Parse ``mail_recipient_allowlist`` into ``(addresses, domains)``.
+
+    A token containing ``@`` in an interior position is a full address;
+    a token without one (or with only a leading ``@``) is a domain. Both
+    are lower-cased, and a trailing dot on a domain is stripped, so
+    matching is case-insensitive and FQDN-dot-tolerant (mirrors the
+    ``net.*`` hostname fold).
+
+    A token that is not a well-formed address or domain raises
+    :class:`ValueError` naming it — loud, because silently keeping an
+    unusable entry misreports the permitted recipient space the operator
+    explicitly opted in (same fail-fast posture as
+    ``net.allowlist.parse_probe_allowlist``). Rejected: whitespace,
+    more than one ``@``, an empty local part or domain part
+    (``foo@``, ``@``), and a domain that folds away to nothing
+    (``...``). The bare ``@`` case is the sharpest: it used to parse to
+    ``domains={""}``, which matches no address yet makes the allowlist
+    look non-empty — suppressing the "connector is inert" refusal
+    message that would otherwise tell the operator exactly what is
+    wrong.
+
+    Reads the live :class:`~meho_backplane.settings.Settings` singleton
+    per call; tests swap the value by mutating the env and clearing
+    ``get_settings``'s cache, like every other settings-backed surface.
+    """
+    addresses: set[str] = set()
+    domains: set[str] = set()
+    for token in get_settings().mail_recipient_allowlist.split(","):
+        entry = token.strip()
+        if not entry:
+            continue
+        if any(ch.isspace() for ch in entry) or entry.count("@") > 1:
+            raise _malformed_entry(entry)
+        folded = entry.lower()
+        if "@" in folded[1:]:
+            local, _, domain = folded.partition("@")
+            if not local or not domain:
+                raise _malformed_entry(entry)
+            addresses.add(folded)
+            continue
+        domain = folded.removeprefix("@").rstrip(".")
+        if not domain:
+            raise _malformed_entry(entry)
+        domains.add(domain)
+    return frozenset(addresses), frozenset(domains)
+
+
+def assert_recipient_allowed(address: str) -> None:
+    """Raise :class:`RecipientNotAllowedError` unless *address* is allowlisted.
+
+    Called by the transport on **every** recipient of a send, *before*
+    any SMTP connection opens. The decision is absolute membership in
+    :func:`parse_recipient_allowlist`'s parsed set:
+
+    * empty allowlist → always refuse (the connector is inert);
+    * a listed full address → allowed verbatim (case-insensitive);
+    * an address whose domain is listed → allowed.
+
+    An address the floor cannot positively parse — empty, containing
+    whitespace or control characters, or not exactly ``local@domain`` —
+    is **refused**, not guessed at: only a parseable address can match
+    the allowlist, which also keeps SMTP envelope injection (CR/LF in a
+    recipient) structurally impossible past this gate. The message never
+    echoes the parsed set (no recipient-space oracle).
+
+    Raises:
+        RecipientNotAllowedError: *address* is empty, malformed, not
+            covered by the allowlist, or the allowlist is empty.
+    """
+    addresses, domains = parse_recipient_allowlist()
+    if not addresses and not domains:
+        raise RecipientNotAllowedError(
+            "recipient refused: MAIL_RECIPIENT_ALLOWLIST is empty, so the "
+            "mail.* connector is inert; add the address or domain to mail"
+        )
+    candidate = address.strip().lower()
+    local, sep, domain = candidate.partition("@")
+    if (
+        not local
+        or not sep
+        or not domain
+        or "@" in domain
+        or any(ch.isspace() or not ch.isprintable() for ch in candidate)
+    ):
+        raise RecipientNotAllowedError(f"recipient refused: {address!r} is not a valid address")
+    if candidate in addresses or domain.rstrip(".") in domains:
+        return
+    raise RecipientNotAllowedError(
+        "recipient refused: address is not listed in MAIL_RECIPIENT_ALLOWLIST"
+    )
--- a/backend/src/meho_backplane/connectors/mail/ops.py
+++ b/backend/src/meho_backplane/connectors/mail/ops.py
@@ -1,0 +1,249 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""``mail.send`` typed op + registrar for the synthetic ``mail.*`` product.
+
+``mail.*`` is a **synthetic** typed-op product on the ``net.*`` mould
+(#2717): no vendor connector backs it, so the package calls neither
+``register_connector`` nor ``register_connector_v2``. The op is
+registered under the natural key
+``(product="mail", version="1.x", impl_id="mail-smtp")``, so the wire
+``connector_id`` is ``mail-smtp-1.x`` — which round-trips through
+:func:`~meho_backplane.operations._lookup.parse_connector_id` back to
+``("mail", "1.x", "mail-smtp")``. The handler is a **module-level**
+function, so the dispatcher resolves it with ``connector_instance=None``
+and ``target=None`` — recipients are params, not a registered
+``Target``.
+
+The handler is a thin adapter over
+:func:`~meho_backplane.connectors.mail.transport.send_email` — the
+single shared transport the checks notifier (#2719) also imports — so
+the recipient-allowlist floor, the unconfigured refusal, and the
+return-failures contract live in exactly one place. The handler's
+return dict carries the literal ``to``/``subject`` (never the body), so
+the durable audit row's ``raw_payload`` answers "who was mailed about
+what" without persisting message content.
+
+``safety_level="caution"`` + ``requires_approval=False``: sending mail
+is an outward-facing side effect, so operators (human/service
+principals, default-allow) auto-run it while agent principals land on
+the ``caution`` needs-approval default of the policy gate
+(:mod:`meho_backplane.auth.permissions`) — and the recipient allowlist
+is the hard floor either way.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Final
+
+from meho_backplane.connectors.mail.transport import send_email
+from meho_backplane.operations.typed_register import register_typed_operation
+
+if TYPE_CHECKING:
+    from meho_backplane.auth.operator import Operator
+    from meho_backplane.retrieval.embedding import EmbeddingService
+
+__all__ = [
+    "MAIL_SEND_PARAMETER_SCHEMA",
+    "mail_send",
+    "register_mail_typed_operations",
+]
+
+#: Every character :meth:`str.splitlines` treats as a line boundary,
+#: excluded from ``subject`` so the schema layer — not the transport —
+#: owns header-injection rejection.
+#:
+#: :class:`~email.message.EmailMessage` guards header assignment with
+#: ``len(value.splitlines()) > 1`` and raises :class:`ValueError`, which
+#: would escape :func:`~meho_backplane.connectors.mail.transport.send_email`
+#: as a ``connector_error`` and break the return-failures contract.
+#: ``splitlines`` breaks on far more than CR/LF (VT, FF, the three
+#: information separators, NEL, and the Unicode line/paragraph
+#: separators), so a ``[^\r\n]`` class would leave the hole half-open;
+#: this class is the exact complement of the stdlib guard, so a
+#: rejected subject is always ``invalid_params``.
+_SINGLE_LINE_PATTERN: Final[str] = r"^[^\r\n\v\f\x1c\x1d\x1e\x85\u2028\u2029]+$"
+
+MAIL_SEND_PARAMETER_SCHEMA: dict[str, Any] = {
+    "type": "object",
+    "properties": {
+        "to": {
+            "type": "array",
+            "minItems": 1,
+            "maxItems": 32,
+            "items": {
+                "type": "string",
+                "minLength": 3,
+                "maxLength": 254,
+                "pattern": r"^[^@\s]+@[^@\s]+$",
+                "description": (
+                    "Recipient email address. Must be covered by "
+                    "MAIL_RECIPIENT_ALLOWLIST (full address or its "
+                    "domain) or the whole send is refused before any "
+                    "SMTP connection opens."
+                ),
+            },
+            "description": "Recipient addresses; one refused recipient refuses the whole send.",
+        },
+        "subject": {
+            "type": "string",
+            "minLength": 1,
+            "maxLength": 200,
+            "pattern": _SINGLE_LINE_PATTERN,
+            "description": (
+                "Subject line (plain text, single line). A line break of "
+                "any kind is rejected as invalid_params — it would be a "
+                "header-injection attempt."
+            ),
+        },
+        "body": {
+            "type": "string",
+            "maxLength": 100_000,
+            "description": "Plain-text message body (no HTML, no attachments).",
+        },
+    },
+    "required": ["to", "subject", "body"],
+    "additionalProperties": False,
+}
+
+_MAIL_SEND_RESPONSE_SCHEMA: dict[str, Any] = {
+    "type": "object",
+    "properties": {
+        "sent": {
+            "type": "boolean",
+            "description": "True iff the MTA accepted the message for delivery.",
+        },
+        "reason": {
+            "type": ["string", "null"],
+            "description": (
+                "Null on success; otherwise a failure code: "
+                "not_in_recipient_allowlist, smtp_unconfigured, "
+                "smtp_connect_error, smtp_auth_requires_tls, "
+                "smtp_auth_error, smtp_recipients_refused, smtp_error."
+            ),
+        },
+        "to": {
+            "type": "array",
+            "items": {"type": "string"},
+            "description": "The requested recipients (audit-visible).",
+        },
+        "subject": {
+            "type": "string",
+            "description": "The subject line (audit-visible; the body is never persisted).",
+        },
+    },
+    "required": ["sent", "reason", "to", "subject"],
+    "additionalProperties": False,
+}
+
+_MAIL_SEND_WHEN_TO_USE = (
+    "Send a plain-text email to operator-allowlisted recipients from the "
+    "backplane's deployment-level SMTP block — e.g. to deliver an "
+    "investigation finding, a maintenance notice, or an alert summary to "
+    "an on-call mailbox. An outward-facing side effect: recipients must "
+    "be inside MAIL_RECIPIENT_ALLOWLIST, and a refused or failed send is "
+    "a normal result (sent=false with a reason), not an error."
+)
+
+_MAIL_SEND_LLM_INSTRUCTIONS: dict[str, Any] = {
+    "when_to_use": (
+        "Use to deliver a short plain-text message to a human mailbox "
+        "when the operator has asked for email delivery. Recipients must "
+        "be allowlisted at the deployment level; do not retry a "
+        "not_in_recipient_allowlist, smtp_unconfigured, or "
+        "smtp_auth_requires_tls refusal — all three are deployment "
+        "configuration, not transient state."
+    ),
+    "parameter_hints": {
+        "to": "Required. 1-32 recipient addresses; every one must be allowlisted.",
+        "subject": (
+            "Required. Single-line subject (max 200 chars); a line break "
+            "of any kind is rejected as invalid_params."
+        ),
+        "body": "Required. Plain-text body; no HTML or attachments.",
+    },
+    "output_shape": (
+        "On success: {'sent': true, 'reason': null, 'to': [<str>], "
+        "'subject': <str>}. On a refused or failed send: {'sent': false, "
+        "'reason': '<not_in_recipient_allowlist|smtp_unconfigured|"
+        "smtp_connect_error|smtp_auth_requires_tls|smtp_auth_error|"
+        "smtp_recipients_refused|smtp_error>', 'to': [<str>], "
+        "'subject': <str>} — still a "
+        "successful (status=ok) op."
+    ),
+}
+
+
+async def mail_send(operator: Operator, target: Any, params: dict[str, Any]) -> dict[str, Any]:
+    """Send one plain-text email through the shared transport.
+
+    Op-id: ``mail.send``. Synthetic typed op (no vendor connector,
+    ``target`` is always ``None``). The dispatcher has validated the
+    param schema, so ``to`` / ``subject`` / ``body`` are present and
+    well-typed.
+
+    Delegates to
+    :func:`~meho_backplane.connectors.mail.transport.send_email` (the
+    allowlist floor and refusal mapping live there) and adapts the
+    :class:`~meho_backplane.connectors.mail.transport.MailSendResult`
+    into the audit-visible payload: the returned dict carries the
+    literal ``to``/``subject`` — the durable audit row's ``raw_payload``
+    answer to "who was mailed" — and never the body.
+    """
+    to = [str(address) for address in params["to"]]
+    subject = str(params["subject"])
+    result = await send_email(to=to, subject=subject, body=str(params["body"]))
+    return {
+        "sent": result.sent,
+        "reason": result.reason,
+        "to": to,
+        "subject": subject,
+    }
+
+
+async def register_mail_typed_operations(
+    *,
+    embedding_service: EmbeddingService | None = None,
+) -> None:
+    """Upsert the ``mail.*`` typed ops into ``endpoint_descriptor``.
+
+    Queued onto the lifespan-driven registrar list by the package
+    ``__init__`` (via ``register_typed_op_registrar``) and run by
+    :func:`~meho_backplane.operations.typed_register.run_typed_op_registrars`
+    after the connector eager-import pass. Idempotent: a re-run against
+    unchanged text is a no-op for the embedding pipeline. Registers
+    ``mail.send`` (#2717) as ``caution`` + ``requires_approval=False``
+    — operators auto-run, agents ride the policy gate's needs-approval
+    default; the recipient allowlist is the hard floor either way.
+    """
+    await register_typed_operation(
+        product="mail",
+        version="1.x",
+        impl_id="mail-smtp",
+        op_id="mail.send",
+        handler=mail_send,
+        group_key="send",
+        when_to_use=_MAIL_SEND_WHEN_TO_USE,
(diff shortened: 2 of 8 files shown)


backend/src/meho_backplane/connectors/mail/ops.py
```diff
@@ -0,0 +1,223 @@
...
+                ),
+            },
+            "description": "Recipient addresses; one refused recipient refuses the whole send.",
+        },
+        "subject": {
+            "type": "string",
+            "minLength": 1,
+            "maxLength": 200,
+            "description": "Subject line (plain text, single line).",
+        },
```
damir-topic, line 98: **[🛑 B2]** — a CR/LF `subject` escapes the return-failures contract as a `connector_error`

**What:** the `subject` schema constrains type and length but not shape, so `"alert\r\nBcc: evil@x.test"` passes `validate_params` unchanged. `transport.send_email` then assigns it with `msg["Subject"] = subject`, which raises `ValueError` — reproduced live in the review worktree on CPython 3.14.6:

```
CRLF subject -> RAISED ValueError : Header values may not contain linefeed or carriage return characters
```

`send_email` does not catch it, so it lands on the dispatcher's generic arm (`operations/dispatcher.py:1118`) and becomes `result_connector_error` with an audit row written as `result_status="error"`.

**Why it matters:** `subject` is agent-supplied on an agent-callable op, so this is caller-controlled input, not "an unexpected bug". Task #2717's Desired state §3 pins the contract as "only unexpected bugs raise", and `transport.py`'s own module docstring plus `docs/codebase/connectors-mail.md` both assert the contract holds — while the docstring simultaneously documents this exact escape as intended. The dispatcher already has the correct channel for a malformed parameter (`result_invalid_params`, ahead of the policy gate at `dispatcher.py:1978`); routing it through `connector_error` instead misfiles an input error as a transport fault in the audit trail, and an agent reading the error has no stable reason code to act on. No header-injection risk — `EmailMessage` does block the write — but the failure mode is wrong.

**Required change:** constrain the schema so the dispatcher rejects it before the handler runs, and stop documenting the hole as a design choice.

```suggestion
        "subject": {
            "type": "string",
            "minLength": 1,
            "maxLength": 200,
            "pattern": "^[^\\r\\n]+$",
            "description": "Subject line (plain text, single line — no CR/LF).",
        },
```

Then drop the "Only an unexpected bug propagates — e.g. a CR/LF-bearing *subject*, which `EmailMessage` rejects with `ValueError`" clause from `transport.send_email`'s docstring, and adjust the matching sentence in `docs/codebase/connectors-mail.md` ("a CR/LF-bearing *subject* is rejected by `EmailMessage` itself") to say the parameter schema rejects it. A test asserting `dispatch(...)` with a CR/LF subject returns `status="invalid_params"` pins it.

**Source:** <https://docs.python.org/3.14/library/email.message.html> (header assignment guard); `backend/src/meho_backplane/operations/dispatcher.py:1118` (generic `except Exception` → `result_connector_error`) and `:1978` (`validate_params` → `result_invalid_params`).

backend/src/meho_backplane/connectors/mail/transport.py
```diff
@@ -0,0 +1,223 @@
...
+            client = smtplib.SMTP(host, port, timeout=_SMTP_TIMEOUT_SECONDS)
+        with client:
+            client.ehlo()
+            if starttls and port != _IMPLICIT_TLS_PORT:
+                client.starttls()
+                # RFC 3207: the pre-TLS EHLO response is void; rediscover
+                # the server's extensions on the encrypted channel.
+                client.ehlo()
+            if username:
+                client.login(username, password)
```
damir-topic, line 156: **[⚠️ M1]** — SMTP AUTH runs on an unencrypted channel when `MAIL_SMTP_STARTTLS=false`

**What:** the `if username:` branch is unconditional on channel state. With `MAIL_SMTP_STARTTLS=false` (or a server that never got the upgrade) on any port other than 465, `client.login(username, password)` executes on the plaintext socket. `smtplib` offers `PLAIN` and `LOGIN` whenever the server advertises them (`preferred_auths = ['CRAM-MD5', 'PLAIN', 'LOGIN']`), both of which put the base64-encoded password on the wire.

**Why it matters:** relay credentials leak to any on-path observer, and `MAIL_SMTP_PASSWORD` is precisely the value the settings block goes out of its way to protect with `repr=False`. The `starttls=True` default does not prevent it — a single operator env flip on an otherwise-working config silently downgrades from encrypted-auth to cleartext-auth with no signal. `smtplib` imposes no precondition of its own here (the 3.14 docs for `login()` state none), so the guard has to live in this function. The empirical anchor task #2717 cites for the whole design, Alertmanager's `email_config`, defaults `require_tls` to `true` rather than authenticating over an unencrypted channel.

Note the adjacent path is already safe: with `starttls=True` against a server that does not advertise STARTTLS, `starttls()` raises `SMTPNotSupportedError` → `smtp_error`, so no silent downgrade happens there. This is the one hole.

**Required change:** track whether the channel is encrypted and fail closed with a new stable reason code rather than authenticating in the clear.

```suggestion
    try:
        if port == _IMPLICIT_TLS_PORT:
            client: smtplib.SMTP = smtplib.SMTP_SSL(host, port, timeout=_SMTP_TIMEOUT_SECONDS)
            encrypted = True
        else:
            client = smtplib.SMTP(host, port, timeout=_SMTP_TIMEOUT_SECONDS)
            encrypted = False
        with client:
            client.ehlo()
            if starttls and port != _IMPLICIT_TLS_PORT:
                client.starttls()
                # RFC 3207: the pre-TLS EHLO response is void; rediscover
                # the server's extensions on the encrypted channel.
                client.ehlo()
                encrypted = True
            if username:
                if not encrypted:
                    return "smtp_auth_requires_tls"
                client.login(username, password)
```

Document `smtp_auth_requires_tls` in the three places the other reason codes already appear: the `transport.py` module docstring's reason-code list, `_MAIL_SEND_RESPONSE_SCHEMA` / `_MAIL_SEND_LLM_INSTRUCTIONS` in `ops.py`, and the reason table in `docs/codebase/connectors-mail.md`. A test in the existing refusal block (`MAIL_SMTP_STARTTLS=false` + `MAIL_SMTP_USERNAME` set ⇒ no `login` in `client.calls`) pins it.

**Source:** <https://docs.python.org/3.14/library/smtplib.html> — `login()` documents no TLS precondition; `starttls()` is what makes "all SMTP commands that follow" encrypted. <https://prometheus.io/docs/alerting/latest/configuration/> — `email_config.require_tls` defaults to `true`.

damir-topic (commented):




**Anchor task:** #2717 — `mail.*` typed connector, SMTP send as a reusable, agent-callable operation
**Surface area:** backend
**Verdict:** REQUEST_CHANGES (posted as COMMENT-state review because GitHub rejects a blocking review from the PR author's identity; the result file is the binding verdict)
**Head SHA reviewed:** `21c57ff4`

The architecture is right and the mould parity is genuine — this is a faithful `net.*` clone with a shared transport seam, an inverted-default allowlist, and the return-failures contract. Two things block: the PR cannot merge (DCO), and the return-failures contract has one agent-reachable hole (a CR/LF subject escapes as a `connector_error`).

### Acceptance criteria

- [x] `pytest backend/tests/test_connectors_mail.py -q` green, covering by name: empty-allowlist refusal, unconfigured-host refusal, STARTTLS+login sequencing, per-`SMTPException`-subclass reason codes, success — **verified**: ran in the review worktree, 29 passed. All five named cases present (`test_empty_allowlist_refuses_before_any_smtp_connection`, `test_unconfigured_host_refuses_with_smtp_unconfigured`, `test_starttls_and_login_run_in_sequence_when_configured`, the `smtp_exception_subclasses` parametrize, `test_send_success_returns_sent_true`).
- [x] `grep 'safety_level="caution"' .../mail/ops.py` hits; `smtplib` in `transport.py`; **no new dependency** — verified: `ops.py:220` registers `safety_level="caution"`, `requires_approval=False`; `transport.py` imports stdlib `smtplib`; `backend/pyproject.toml` untouched in the diff.
- [x] Route/integration test proves `("mail-smtp-1.x", "mail.send")` dispatch returns `status="ok"` and writes an audit row — verified: `test_dispatch_mail_send_returns_ok_and_writes_audit_row` drives the real `dispatch()` and asserts `raw_payload` has `to`/`subject` and no `body`.
- [x] `mail_smtp_password` carries `repr=False` — verified: `settings.py` `mail_smtp_password: str = Field(default="", repr=False)`.
- [x] No migration — verified: nothing under `backend/alembic/versions/` in the diff; the "Detect migration-related changes" check confirms.
- [x] `CHANGELOG.md` `[Unreleased]` gains the unique header; full backend suite green — verified: `### Added — Mail connector — SMTP send as a typed operation (#2717)`; the `Python (ruff + mypy + pytest)` and `Python (integration testcontainers)` lanes are SUCCESS on this SHA.
- [ ] ❌ **Return-failures contract complete** (task Desired state §3: "only unexpected bugs raise") — an agent-supplied `subject` containing `\r` or `\n` passes the parameter schema and raises `ValueError` out of `send_email`, which the dispatcher maps to `connector_error` with `result_status="error"`. See 🛑 B2.

### Project-board alignment

|---|---|---|---|---|---|
| #2717 | Open (task, `priority:high`) | `Closes #2717` | ✅ | ✅ full (bar B2) | close on merge |
| #2716 | Open (parent initiative) | none | — | partial (scope item 1 of 5) | stays open — #2718–#2721 outstanding |

No stealth closures, no delivered-but-unlinked work. Siblings #2718–#2721 are untouched and correctly out of scope; sibling PRs #2722/#2723 are docs-site work on unrelated paths — no collision.

### Predicted vs. actual diff

- ✅ `connectors/mail/{__init__,allowlist,ops,transport}.py` — the predicted four-file package, on the `net.*` layout, module-level handler, registrar queued from `__init__` via `register_typed_op_registrar`.
- ✅ `settings.py` — the seven predicted fields + the `get_settings()` env mapping, with the `#:`-comment carrying the owning issue per convention.
- ✅ `backend/tests/test_connectors_mail.py`, `CHANGELOG.md`, `docs/codebase/connectors-mail.md`.
- ⚠️ `connectors/mail/ops.py` — predicted-and-present but the `subject` schema is missing the single-line constraint the contract needs (🛑 B2).
- ⚠️ `connectors/mail/transport.py` — predicted-and-present but authenticates on an unencrypted channel when an operator sets `MAIL_SMTP_STARTTLS=false` with a username (⚠️ M1).
- 🚫 None. No undisclosed edits: the diff is 8 files, all named in the PR body, zero deletions.
- 🕳️ None missing that the task asked for. `deploy/charts/meho/` plumbing for the `MAIL_*` vars is *not* a gap: the comparable `CHECKS_*` settings block shipped without chart plumbing too, `extraEnv` covers it, and the connector is inert-by-default — recorded under Adjacent improvements for #2719 instead.

### Integration verification

Every command below ran in `.claude/worktrees/pr-2724-review/` on `21c57ff4`.

- [x] `uv run ruff format --check src/ tests/` — clean (1375 files)
- [x] `uv run mypy src/` — 1 error, `net/icmp.py:208 MSG_ERRQUEUE`, pre-existing and macOS-only (untouched file; the Linux CI mypy lane is SUCCESS on this SHA)
- [x] `uv run pytest tests/test_connectors_mail.py -q` — 29 passed
- [x] `uv run pytest tests/ -k "typed_register or maturity_surface"` — 47 passed (registration seam + the #2678 surface-drift guard both unaffected)
- [x] Dispatch path exercised by the PR's own `test_dispatch_mail_send_returns_ok_and_writes_audit_row` (real `dispatch()`, real audit row) — re-run green
- [x] Contract deviation reproduced live, not reasoned about: `send_email(subject="alert\r\nBcc: …")` raises `ValueError: Header values may not contain linefeed or carriage return characters`; `dispatcher.py:1118`'s generic `except Exception` maps that to `result_connector_error`
- [x] Allowlist edge parses reproduced live: `MAIL_RECIPIENT_ALLOWLIST="foo@"` → `({'foo@'}, {})`; `="@"` → `({}, {''})`
- [x] `smtplib` / `EmailMessage` idioms cross-checked against the Python 3.14 docs (see Sources) — the STARTTLS + re-`ehlo()` sequence, the explicit `from_addr`/`to_addrs` envelope, and `SMTP_SSL` on 465 are all correct as written
- ⚠️ not run — migrations replay (no migration in the diff, `Detect migration-related changes` SKIPPED downstream jobs)
- ⚠️ not run — `docker compose up` (no runtime-surface change; lifespan registrar path is covered by the typed_register suite)
- ⚠️ not run — real SMTP session against a live MTA. Every SMTP interaction in the suite is a monkeypatched stub, so the wire-level session (real STARTTLS negotiation, real AUTH, real `.` -stuffing of the body) is unproven. Acceptable for merge given the connector is inert by default, but the first deployment enabling `MAIL_RECIPIENT_ALLOWLIST` is the real integration test — worth a smoke note on #2719.

- 🛑 **B1** — the PR cannot merge: the required `DCO` check is `action_required` because commit `21c57ff4` has no `Signed-off-by` trailer. Required change: add the sign-off without rewriting history — commit a DCO remediation commit on the branch whose message is exactly `I, Damir Topic <89133183+damir-topic@users.noreply.github.com>, hereby add my Signed-off-by to this commit: 21c57ff414a2820b7417d84af705d0ab06ea6b36`, itself signed off with `Signed-off-by: Damir Topic <89133183+damir-topic@users.noreply.github.com>`. (Sibling PRs #2722/#2723 pass DCO, so this is a one-off miss on this branch, not a repo-wide config change.) **Not posted inline** — it is a commit-level, not a file-level, defect.
- 🛑 **B2** — a CR/LF-bearing `subject` escapes the return-failures contract as a `connector_error`; required change: add `"pattern": "^[^\\r\\n]+$"` to the `subject` schema in `MAIL_SEND_PARAMETER_SCHEMA` so the dispatcher's own `validate_params` rejects it as `invalid_params`, and drop the "only an unexpected bug propagates — e.g. a CR/LF-bearing *subject*" clause from `transport.send_email`'s docstring, which currently documents the hole as intended behaviour.
- ⚠️ **M1** — `login()` runs on a plaintext socket when `MAIL_SMTP_STARTTLS=false` and a username is set, leaking relay credentials to any on-path observer; required change: track whether the channel is encrypted and return a new `smtp_auth_requires_tls` reason instead of authenticating in the clear, documented in the transport module docstring, `_MAIL_SEND_RESPONSE_SCHEMA`, and `docs/codebase/connectors-mail.md`.
- 💡 **m1** — `parse_recipient_allowlist` silently accepts `foo@` and `@`, contradicting its own "fails loud, never silently narrows the permitted space" docstring; required change: reject a token that is not a well-formed `local@domain` or bare domain with the existing `ValueError`.

- `send_email(to=[])` on the **direct-import** path (the #2719 notifier, which has no parameter schema in front of it) skips the allowlist loop entirely, opens the SMTP session, and only then fails as `smtp_recipients_refused` from `sendmail`'s all-recipients-refused branch. Structurally safe and contract-shaped, but it dials the MTA for a guaranteed no-op — an early `if not to: return MailSendResult(sent=False, reason=...)` would be cheaper. Worth folding into #2719 when the notifier lands, since that is the caller that can produce an empty list.
- A recipient like `a,b@example.com` passes the allowlist (comma is printable and not whitespace), so `msg["To"] = ", ".join(to)` renders a header that parses as two addresses while the envelope stays correctly single-recipient. Harmless today because the envelope is explicit; a comma exclusion in the allowlist parse would close the cosmetic gap.
- `MAIL_*` is not plumbed into `deploy/charts/meho/` the way `MEHO_NETDIAG_PROBE_ALLOWLIST` is (`values.yaml` + `values.schema.json` + `configmap.yaml` + `docs/deploying.md`). Deliberately fine here — `CHECKS_*` shipped the same way and `extraEnv` covers it — but #2719 is the natural place to add it, since that is when a deployment actually needs SMTP configured.
- `classify_op("mail.send")` falls through to `other` (full-detail broadcast). Called out in the PR body as a deliberate non-change and correct as-is (the payload is `to`/`subject`, audit-visible by design) — #2720 is the ticket that touches `classify_op`.

### Docs

- [x] `docs/codebase/connectors-mail.md` — **added**, and it is a genuinely good area doc: control flow, the reason-code table, the load-bearing `except`-ordering note, the config table, and a Known-issues section. Two lines need to follow B2/M1 (the CR/LF-subject claim under Allowlist semantics, and a new `smtp_auth_requires_tls` row in the reason table). No `docs/codebase/` index file exists, so nothing else needed updating. No .NET analogies present.

- The mould parity is real rather than nominal: natural key → `mail-smtp-1.x`, module-level handler, registrar queued from `__init__`, inverted-default allowlist, return-failures contract. Someone reading `net/` next week will recognise every seam.
- The `except`-ordering comment in `transport._send_sync` (SMTP arms before the `OSError` arm because `SMTPException` subclasses `OSError`) is exactly the kind of non-obvious invariant that deserves a comment — and it is pinned by a parametrized test rather than left to prose.
- `test_empty_allowlist_refuses_before_any_smtp_connection` proves the floor with an `_ExplodingSMTP` stub instead of asserting on a return value. That is the right way to test "refused *before* any connection opens".

### Sources

- Python 3.14 `smtplib` — <https://docs.python.org/3.14/library/smtplib.html>: `starttls()` "Put the SMTP connection in TLS mode. All SMTP commands that follow will be encrypted. You should then call `ehlo()` again" (confirms the RFC 3207 re-`ehlo()` in `_send_sync`); `send_message` fills `from_addr`/`to_addrs` from headers only when they are `None`, so the explicit envelope here is correct; `login()` carries **no** TLS precondition of its own — the caller must impose it (basis for M1).
- Python 3.14 `email.message` — <https://docs.python.org/3.14/library/email.message.html>; header assignment raises `ValueError` on CR/LF, verified empirically on CPython 3.14.6 (basis for B2).
- Prometheus Alertmanager `email_config` — <https://prometheus.io/docs/alerting/latest/configuration/>: `require_tls` defaults to `true`. The empirical anchor cited by task #2717 itself refuses unencrypted SMTP by default rather than authenticating over it (basis for M1).

backend/src/meho_backplane/connectors/mail/allowlist.py
```diff
@@ -0,0 +1,137 @@
...
+        if any(ch.isspace() for ch in entry) or entry.count("@") > 1:
+            raise ValueError(
+                f"MAIL_RECIPIENT_ALLOWLIST entry {entry!r} is not a valid "
+                "address or domain; fix the deployment configuration"
+            )
+        folded = entry.lower()
+        if "@" in folded[1:]:
+            addresses.add(folded)
+        else:
+            domains.add(folded.removeprefix("@").rstrip("."))
```
damir-topic: **[💡 m1]** — the parser silently accepts entries it documents as fail-loud

**What:** the address/domain split keys off `"@" in folded[1:]` with no well-formedness check on either branch. Reproduced live in the review worktree:

```
allowlist='foo@' -> (frozenset({'foo@'}), frozenset())
allowlist='@'    -> (frozenset(), frozenset({''}))
```

`foo@` becomes an address entry no real recipient can ever match (`assert_recipient_allowed` requires a non-empty domain). `@` is worse: it yields `domains={''}`, which is a *non-empty* parsed set, so `assert_recipient_allowed` skips the inert branch — the connector reports as configured, nothing matches, and the operator gets "address is not listed" instead of the "MAIL_RECIPIENT_ALLOWLIST is empty, so the mail.* connector is inert" message that would tell them what is actually wrong.

**Why it matters:** the docstring three lines up commits to the opposite behaviour — "A token containing whitespace or more than one `@` raises `ValueError` naming the offending token — loud, because silently dropping an entry would narrow the permitted recipient space the operator explicitly opted in". These entries do exactly the narrowing the comment promises to prevent, and the `@` case additionally defeats the inert-state diagnostic. Either the code or the docstring is wrong; the code is the one worth fixing, since the whole point of the fail-fast posture is that a typo in a deployment's allowlist surfaces at parse time rather than as a mail that never arrives.

**Required change:** extend the existing validation to reject a token that is not a well-formed `local@domain` or bare domain — i.e. after the whitespace / multi-`@` check, refuse an entry whose local part or domain part is empty, reusing the same `ValueError` message shape. `test_parse_recipient_allowlist_rejects_malformed_entry` already exists; adding `"foo@"` and `"@"` to its parametrize list covers it.

damir-topic:
## B1 is not fixable as specified — DCO remediation commits are disabled on this repo

Iteration 1 pushed the remediation commit the review asked for
(`df71e2f2`), in the exact format the DCO app matches:

```
I, Damir Topic <89133183+damir-topic@users.noreply.github.com>, hereby add my Signed-off-by to this commit: 21c57ff414a2820b7417d84af705d0ab06ea6b36
```

DCO re-ran against the new head (`37bc297f`) and **still fails**. The
cause is not the commit's wording — it is repository configuration.

From [`dcoapp/app/lib/dco.js`](https://github.com/dcoapp/app/blob/main/lib/dco.js#L136),
the entire remediation-matching block is gated:

```js
if (allowRemediationCommits.individual) {
  // if individual remediation commits are allowed, look for one in the PR with a matching sha
```

and [`index.js`](https://github.com/dcoapp/app/blob/main/index.js#L36-L44)
defaults the flag to **false**:

```js
const config = await context.config("dco.yml", {
  require: { members: true },
  allowRemediationCommits: { individual: false, thirdParty: false },
});
```

`context.config()` reads `.github/dco.yml` from the repository's
**default branch**. There is no `.github/dco.yml` in `evoila/meho` and
none in an `evoila/.github` org repo (both return 404), so
`allowRemediationCommits.individual` is `false` and the remediation
commit is ignored outright. That is also why the check's own failure
text offers only the rebase remedy and never mentions remediation
commits — that guidance is rendered only when the flag is on.

`DCO` is in `main`'s required status checks, so this genuinely blocks
merge.

### Two ways forward — both need a maintainer decision

1. **Enable remediation commits** (preferred; no history rewrite).
   Land a one-file PR on `main`:

   ```yaml
   # .github/dco.yml
   allowRemediationCommits:
     individual: true
   ```

   Once that is on the default branch, re-run DCO here and `df71e2f2`
   satisfies it immediately — nothing else on this branch changes. This
   also makes every future missed sign-off recoverable without a
   force-push.

2. **Rebase with `--signoff` and force-push.** `git rebase HEAD~6
   --signoff && git push --force-with-lease`. This rewrites published
   history, which the autonomous run is explicitly forbidden from
   doing, and it would drop `df71e2f2` (an empty commit) along the way.

The remediation commit is left in place: it is correct, it is inert
while the flag is off, and it becomes sufficient the moment option 1
lands.

The other three findings (**B2**, **M1**, **m1**) are fixed, tested,
and pushed — see the updated PR description.

Issue #2716, opened 2026-07-31
Initiative: checks alerting delivery — mail connector, dashboard email notifications, transition broadcast, in-band advisory


damir-topic:
Parent goal: #221

## Summary

Alert **delivery** for the shipped checks layer (#2416): a reusable **`mail.*` typed connector** (SMTP send, agent-callable through `call_operation`), **per-Dashboard email notification** on state transitions (warning/alert = the existing `degraded`/`critical` vocabulary), a **`checks.transition` broadcast event**, an **in-band `checks_alert_advisory`** on dispatch responses so anyone actively working through the backplane learns of a non-green Dashboard on their next call, and **operator-prompted investigations whose findings are emailed**.

## Why

#2416 shipped detection end-to-end — per-value `degraded`/`critical` threshold bounds (`checks/assertions.py:166-201`), five-state rollup, and a CAS-guarded transition detector — but a transition today reaches no human: no email/SMTP transport exists anywhere in `backend/src` (verified by grep: the only SMTP mentions are Keycloak-connector redaction), `checks/` publishes no broadcast event, and a dispatching agent/operator sees nothing. The driving consumer (RDC lab, `claude-rdc-hetzner-dc`) wants datacenter monitoring (physical hosts, base virtualization, nested labs — CPU/memory/storage) where a dangerous level sends a configurable email and ambient callers get a nudge. Every needed metric op already exists and is `safe` (`vmware.host.usage`, `vmware.composite.datastore.usage`, `net.*` probes, `holodeck.disk.usage`) — delivery is the only missing layer. Comparable: Prometheus Alertmanager treats notification (SMTP receivers) as a separate layer on top of alert-state detection; this Initiative is that layer for MEHO.

## Current state

- Transition seam (the one every child hooks): `checks/investigate.py:395-428` — per-(tenant,dashboard) `pg_advisory_xact_lock`, membership fold, then `_claim_rollup_transition` compare-and-swaps `check_dashboards.last_rollup_state` on **any** state change (`investigate.py:417-423`); exactly one caller per edge wins the claim, replica-safe. Only the worsening+claimed+non_green path fires anything today (`:426-428` → the diagnose-only investigator).
- The investigator's finding lands only in `agent_run` + tenant memory (`investigate.py:887-898`) — never in an operator channel.
[...]

Issue #2550, opened 2026-07-16
Task: Dispatch-time target-activity advisory — write-op responses carry recent peer activity on the same target via OperationResult.extras


damir-topic:
Parent goal: #2542
Parent initiative: #2543
Depends-on: #2544
Depends-on: #2545

## Summary

Close the crossfire loop with awareness, not locks: when a write-class operation dispatches against a target that has recent peer activity (another principal's operations or an active announcement claim), the dispatch response carries a compact advisory so the caller — human or agent — learns about the overlap at the moment it matters. Pre-op checking remains the discipline's `recent` step; this is the safety net for the caller who didn't check.

## Current state (verified in code, 2026-07-16)

- No dispatch-time consult of the feed exists; crossfire avoidance is 100% voluntary reader-side polling (review finding 23).
- The response seam: OK envelope built by `wrap_ok_result` (`operations/dispatcher.py:756`; builder `operations/_errors.py:1478`); `OperationResult.extras: Mapping[str, Any]` is the established extension slot (`connectors/schemas.py:616-636`, frozen model).
- Target identity in scope at the seam: `target.name` / `_resolve_target_id(target)` (`operations/_audit.py:474-475,575`).
- Recent-activity query primitives: `_list_recent_events_core` + `event_matches` target filter (`broadcast/history.py:248-493`); post-T1 `active_only` claims; post-T3 `actor_sub` distinguishes the acting agent.
- Sibling-collision: topology #2537 edits the same dispatcher region (MCP handlers dispatching through `policy_gate`) — serialize merges.

## Desired state

- In the dispatch success path for write-class ops (`op_class` already resolved for audit/broadcast), a bounded read of the tenant stream (last N minutes, target-filtered, excluding the caller's own principal/actor) yields `extras["target_activity_advisory"]`: up to 5 entries of `{principal_sub, actor_sub?, kind (operation|announcement), op_id?/phase?, ts}` — structured fields only, **no announcement prose** (the untrusted envelope never enters an op response).
- Fail-open and bounded: advisory lookup failure or timeout (single capped stream read) never fails or delays the dispatch beyond a small budget; a settings knob (`dispatch_activity_advisory_window_minutes`, default 30, `0` = off) gates the feature.
- Read-class ops skip the lookup entirely (no cost on the hot read path).

## Acceptance criteria

[...]

PR #2725, opened 2026-07-31
docs(site): the five named deployment shapes, with rendered diagrams

damir-topic:
## Summary

- Adds **Deployment shapes** (`docs-site/deployment-shapes.md`, under *Start here*) — the five materially different ways a real estate deploys MEHO: **flat LAN**, **segmented network with satellite gateways**, **air-gapped**, **cloud-native without Vault (Google Secret Manager)**, and **multi-tenant MSP**. Each shape gets a rendered mermaid component diagram, a plain-language "pick this shape when", and the constraints that actually decide the install.
- Framing is **one chart, five topologies**, so the shared parts — components and who provides them, the private-address guard, the broadcast posture, render-time schema validation — are stated once up front instead of five times. A reader picks a shape *before* the install trail, because that decision fixes which address ranges get allowlisted, whether a credential store they don't already run is needed, and whether any part of the estate needs a gateway rather than a route.
- **Leads with the edges rather than burying them:** satellite gateways are read-only by design, so a satellite-only enclave gets observation and not change automation; air-gapped estates cannot use cloud-hosted MCP clients at all, and no configuration changes that; per-operator Workload Identity Federation obligates a background-dispatch decision or credentialed Sensors silently read nothing; and **multi-tenant MSP is labelled a target shape, not a field-proven one** — consistent with the feature-maturity honesty discipline, rendered as a `!!! warning` admonition and linked to the maturity index.
- **Every operational claim links a public file in this repo**, collected in the page's own *Sources* section: `docs/deploying.md`, `deploy/charts/meho/values.yaml` + `values.schema.json`, `deploy/values-examples/`, `docs/codebase/satellite-runner.md`, `docs/codebase/connectors-shared-vault-creds.md`, and the README's signature-verification section.
- **Mermaid rendering is enabled site-wide** via the Material-documented `pymdownx.superfences` custom fence. That config uses the `!!python/name:` YAML tag, which `yaml.safe_load` rejects, so `mkdocs.yml` joins the chart templates in the pre-commit `check-yaml` exclude — the strict mkdocs build validates the file far harder anyway.

## Coordination with #2722 (task #2671)

Deliberate scope split, agreed on #2671 before this PR was written:

| Owner | Slice |
|---|---|
| #2671 / #2722 | Landing page, the **single-system** reference-architecture page (`architecture.md`), the continuous install trail, and its side-quest pages (credential backends, Keycloak realm, TLS, upgrades, kind quickstart) |
[...]

PR #2726, opened 2026-07-31
feat(checks): in-band checks_alert_advisory on dispatch responses, deduped per caller and state

### LINKED ISSUE #2718: Task: in-band checks_alert_advisory on dispatch responses, deduped per caller and state
Parent goal: #221
Parent initiative: #2716

## Summary

Add the **in-band `checks_alert_advisory`**: when the caller's tenant has a Dashboard whose `last_rollup_state` memo is `degraded`/`critical`, the next successful dispatch response — agent `call_operation` and operator CLI alike, read ops included — carries one compact `extras` entry naming it, **once per (caller, dashboard, state)** via a Valkey `SET NX EX` dedupe. Second extras fragment beside the shipped #2550 `target_activity_advisory`. Scope item 4 of Initiative #2716.

## Current state

- Precedent (#2550): `build_target_activity_advisory` (`broadcast/history.py:1112-1188`) — awaited at the dispatcher success path and merged into `wrap_ok_result(..., extras=advisory)` (`operations/dispatcher.py:775-780`); keyed `ADVISORY_EXTRAS_KEY = "target_activity_advisory"` (`history.py:1039`); fail-open broad guard (`:1165-1173`); `0`-disables settings knob short-circuiting before any I/O (`dispatch_activity_advisory_window_minutes`, `settings.py:1063-1068`); gated to write-class ops.
- The state to surface already exists and is maintained exactly-once per transition: `check_dashboards.last_rollup_state` (`db/models.py:6380-6383`), CAS-maintained at the persist seam (`checks/investigate.py:461-486`), riding the `check_dashboard_tenant_idx` index (`db/models.py:6398-6403`).
- Valkey client + key conventions: broadcast uses `meho:feed:{tenant_id}` streams (`broadcast/publisher.py`); redis-py `>=8.0.1,<9` is pinned (`backend/pyproject.toml:30`).
- Caller identity fields for the dedupe key: `operator.sub` + `resolve_actor_sub()` — the same pair the #2550 advisory uses to drop self-activity (`history.py:1156-1157`).

## Desired state

1. `build_checks_alert_advisory(operator) -> dict` (new, `backend/src/meho_backplane/checks/advisory.py` — checks domain, not `broadcast/`): 
   - Short-circuit when `checks_alert_advisory_window_minutes == 0` (new setting, default 30, `ge=0`; `#:` comment naming this task).
   - One indexed query: `SELECT id, name, last_rollup_state FROM check_dashboards WHERE tenant_id = :t AND last_rollup_state IN ('degraded','critical')`. NULL memo ⇒ no advisory. (Deliberate, documented limitation: staleness-derived `unknown` is not reflected — the memo is transition-maintained; acceptable for an advisory.)
   - Per hit, Valkey `SET key 1 NX EX <window_seconds>` on `meho:checks:advisory:{tenant}:{principal_sub}:{actor_sub}:{dashboard_id}:{state}`; include the dashboard in the fragment **only when the SET succeeded** (returned OK, i.e. key was absent). A state change mints a new key ⇒ automatic re-announce on escalation/recovery-to-worse; window expiry re-reminds.
   - Returns `{"checks_alert_advisory": [{"dashboard_id", "name", "state"}, ...]}` or `{}`. Server-derived fields only — no free-form text lifted from sensor evidence.
   - **Applies to all op classes** — deliberate divergence from the write-class-gated precedent, documented in the docstring: ambient awareness is the feature; the #2550 gate exists because that advisory is target-overlap-specific.
   - Fail-open: a broad guard mirrors `history.py:1165-1173` — DB or Valkey failure logs `checks_alert_advisory_failed` and returns `{}`; never converts a successful dispatch into a failure.
2. Dispatcher merge: beside `dispatcher.py:775-780`, `extras = {**activity_advisory, **checks_advisory}` into the same `wrap_ok_result` call.

## Implementation sketch

- Keep the builder dependency-light: sessionmaker + the shared Valkey client; no imports from `investigate.py`.
- Tests: settings=0 short-circuit (no DB/Valkey touched — assert via stubs); green tenant ⇒ `{}`; non-green ⇒ fragment once, second call same state ⇒ `{}` (NX dedupe), state change ⇒ new entry; per-caller isolation (different `principal_sub` each get one); Valkey down ⇒ `{}` + warning; dispatcher merge test asserting both fragments coexist on one response.
- CLI note: the advisory rides `OperationResult.extras` which the CLI already renders/passes through — verify with one CLI-path test or snapshot; no Go changes expected.
- CHANGELOG bullet under `### Checks advisory — non-green Dashboards surface on dispatch responses`.

## Acceptance criteria

- [ ] Unit tests (`backend/tests/test_checks_advisory.py`) cover, by name: disable short-circuit; green ⇒ empty; dedupe NX (same caller+state once per window); re-announce on state change; per-caller isolation; fail-open on Valkey and on DB errors.
- [ ] Dispatch integration test: with a seeded `critical` memo, a successful `call_operation` response carries `extras.checks_alert_advisory` with `{dashboard_id, name, state}`; an immediate second dispatch by the same caller does not.
- [ ] `grep -n "checks_alert_advisory" backend/src/meho_backplane/operations/dispatcher.py` hits at the success path beside the #2550 merge; the #2550 fragment still present in the same response (coexistence test).
- [ ] `grep -n "checks_alert_advisory_window_minutes" backend/src/meho_backplane/settings.py` shows `Field(default=30, ge=0)` with the `#:` comment convention.
- [ ] No migration; CHANGELOG bullet present; full backend suite green.

## Out of scope

- Advising on raw Sensor states or staleness-derived `unknown` — memo-backed Dashboard states only (documented limitation).
- Persisting advisory delivery state anywhere durable — the Valkey TTL key IS the state; a Valkey flush simply re-reminds once.
- MCP `_meta`/`structuredContent` surfacing — the fragment rides `extras` inside the tool-result JSON like #2550 does.
- Any blocking/locking semantics — advisory only, never gates dispatch.

## References

- Initiative #2716 (parent). No build dependency on the other tasks (reads the memo #2506 shipped; coexists with whatever T2/T3 add).
- Precedent: `backend/src/meho_backplane/broadcast/history.py:1039,1112-1188` (builder, fail-open guard `:1165-1173`, self-drop identity pair `:1156-1157`); `operations/dispatcher.py:775-780` (merge site); `settings.py:1063-1068` (knob mould).
- Memo + index: `backend/src/meho_backplane/db/models.py:6380-6383,6398-6403`; CAS maintainer `checks/investigate.py:461-486`.
- Redis `SET` with `NX` + `EX` — atomic set-if-absent-with-TTL, returns nil when the key exists: https://redis.io/docs/latest/commands/set/ (verified 2026-07-31); redis-py pinned `>=8.0.1,<9` (`backend/pyproject.toml:30`).

damir-topic:
## Summary

- Add the in-band **`checks_alert_advisory`** (#2718, Initiative #2716 scope item 4): when the caller's tenant has a Dashboard whose `last_rollup_state` memo is `degraded`/`critical`, the next successful dispatch response — agent `call_operation` and operator CLI alike, **read ops included** — carries one compact `extras` entry naming it (`{dashboard_id, name, state}`, server-derived fields only).
- Deduped **once per (caller, dashboard, state)** via an atomic Valkey `SET NX EX` on `meho:checks:advisory:{tenant}:{principal_sub}:{actor_sub}:{dashboard_id}:{state}` — a state change mints a new key (instant re-announce on escalation), window expiry re-reminds; the TTL key IS the delivery state (nothing durable persisted).
- Second extras fragment beside the shipped #2550 `target_activity_advisory`, merged at the same dispatcher success path (`extras={**activity, **checks}`). Deliberate divergence from the write-class gate, documented in the builder: ambient awareness is the feature; the #2550 gate exists because that advisory is target-overlap-specific.
- Fail-open end to end: a DB or Valkey failure warn-logs `checks_alert_advisory_failed` and returns `{}` — never gates, blocks, or fails a dispatch. `CHECKS_ALERT_ADVISORY_WINDOW_MINUTES` (default 30, `ge=0`) short-circuits before any I/O when `0`.
- No migration. New doc `docs/codebase/checks-advisory.md`; cross-ref in `docs/codebase/broadcast.md`; CHANGELOG bullet.

## Test plan

- [x] `ruff check` + `ruff format --check` — clean on `src/` + `tests/`
- [x] `mypy` — clean on all touched files (the sole repo error is pre-existing `connectors/net/icmp.py` `MSG_ERRQUEUE`, macOS-environmental; CI runs Linux)
- [x] `pytest tests/test_checks_advisory.py` — 10 passed, covering by name: disable short-circuit (no DB/Valkey I/O, stub-asserted); green tenant ⇒ `{}` (NULL + `ok` memos); NX dedupe (same caller+state once per window, TTL = window); re-announce on state change (new key per state); per-caller isolation (distinct principals each reminded once); fail-open on Valkey and on DB errors (warn-logged, `{}`)
- [x] Dispatch integration: with a seeded `critical` memo, a successful `dispatch()` of a **read-class** op carries `extras.checks_alert_advisory` with `{dashboard_id, name, state}`; an immediate second dispatch by the same caller does not; a dead advisory path leaves the dispatch `ok`
- [x] Coexistence: one write-class response carries **both** `target_activity_advisory` (#2550) and `checks_alert_advisory`
- [x] `grep -n "checks_alert_advisory" backend/src/meho_backplane/operations/dispatcher.py` — hits at the success path beside the #2550 merge
- [x] `grep -n "checks_alert_advisory_window_minutes" backend/src/meho_backplane/settings.py` — `Field(default=30, ge=0)` with the `#:` comment convention (+ the explicit `os.environ.get` mapping `get_settings()` requires — the class is a plain `BaseModel`, not `BaseSettings`)
- [x] Full backend suite green (one deselect: `test_connectors_net_dns_lookup.py::test_chosen_resolver_queries_that_nameserver` fails identically on pristine `main` — live-DNS environmental, unrelated)
- [x] `code-quality.py --diff` — 0 blockers (6 pre-existing legacy warnings in brushed files)

## Out of scope

- Raw Sensor states / staleness-derived `unknown` — memo-backed Dashboard states only (documented limitation in the builder + doc).
- Durable advisory delivery state — the Valkey TTL key is the state; a flush re-reminds once.
- MCP `_meta`/`structuredContent` surfacing — the fragment rides `extras` inside the tool-result JSON like #2550.
- Sibling #2717 (`mail.*` connector), #2719–#2721 — the rest of Initiative #2716.

Closes #2718

---

Addressed the review findings from [pr-2726-iter-1](https://github.com/evoila/meho/pull/2726#pullrequestreview):

- **B1** `944e99dd` — the doc claimed the operator CLI renders the fragment; it does not. `printCallResult` returns inside its `status == "ok"` branch before the `extras` block, which lives only on the non-ok path, and this advisory only ever rides a *successful* dispatch. Scoped honestly instead of patched: new docs § "Operator reach (`--json`-only today)" with the per-front table and the reason it was scoped out (printing extras on success is a CLI-wide UX change that would also surface #2550's fragment, and the vendor verbs render through `dispatch.Render`), matching corrections on the module docstring and the CHANGELOG bullet, plus the CLI-path test #2718 asked for — as a **pair**: `TestCallJSONCarriesChecksAlertAdvisoryOnOK` (the reachable path) and `TestCallHumanRenderOmitsExtrasOnOK` (the gap, failing loudly with a pointer to the doc if anyone closes it).
- **M1** `333f20cd` — added `_ADVISORY_MAX_DASHBOARDS = 100` and applied it as `.limit()` on the non-green SELECT, mirroring #2550's `_ADVISORY_SCAN_LIMIT` (same value, same rationale); the `**Bounded.**` docstring bullet now names the cap instead of asserting the property. New test asserts on the count of Valkey round-trips issued, not just the fragment length — a Python-side cap after the loop would satisfy the entry count and still leave the fan-out unbounded.

Also in this iteration:

- **Merge conflict** `37caa7c3` — merged `origin/main` (advanced to c5fa843f / #2723 after the branch was cut). One conflicting file, `CHANGELOG.md`, both sides inserting under `## [Unreleased]`; both bullets kept. Merge commit, no rebase, no force-push.
- **Copilot's dedupe-key `:`-collision thread** — [replied with the grammar citation](https://github.com/evoila/meho/pull/2726#discussion_r3695373299) and closed out. `agent:<name>` is the Keycloak *clientId*, not the `sub`; subs are Keycloak UUIDs and agent names are pinned to `^[A-Za-z0-9_\-\.]+$` (`auth/agent_principals.py:95,98`), so both free-form segments are colonless and no reachable pair aliases. Recorded under "Known issues / boundaries" rather than re-litigated.

Verification: `ruff` / `ruff format` / `mypy` clean on the touched files; `pytest tests/test_checks_advisory.py` 11 passed; blast-radius neighbours 151 passed; the checks/advisory/dispatcher/broadcast/settings slice 1037 passed, 37 skipped; `go build ./...` + full `go test ./...` on the CLI green; `code-quality.py --diff` exit 0.

---

## Follow-up (auto-implement-issue, iteration 2, 2026-08-01)

Addressed the one blocker from [pr-2726-iter-2](https://github.com/evoila/meho/pull/2726#discussion_r3695398767):

- **B1** `466ddcbc` — `_ADVISORY_MAX_DASHBOARDS = 100` bounded the SELECT but not the cost M1 was filed about. The review is right that the dedupe cannot help: the `SET` is issued before anyone knows whether it claims, so the fan-out recurred on **every** successful dispatch, read ops included, for as long as a tenant stayed red. Fixed on both axes — the cap is now **10** (re-anchored to `_ADVISORY_MAX_ENTRIES = 5`'s "awareness nudge, not an audit" rationale, since `_ADVISORY_SCAN_LIMIT = 100` bounds a one-round-trip `XREVRANGE` and so carries no argument for an agent-facing cap), **and** the claims are staged into one pipelined `MULTI`/`EXEC` batch, so the Valkey cost per dispatch is **one** awaited round-trip instead of up to 100.

  Option 2 was taken, but deliberately **not verbatim**. Keeping a 100-row read with a separate nudge-sized *fragment* cap breaks the module's load-bearing invariant — the Valkey key IS the delivery record, so claiming 100 keys while emitting 5 entries marks 95 Dashboards "announced" without announcing them, silently suppressed for a full window. That is strictly worse than the tail-invisibility the doc already accepts, and repairing it needs a compensating `DEL` or a claim-until-K loop that defeats the batching. One nudge-sized cap keeps claims and entries equal, so a claimed key still means "this caller was told", and the pipeline is pure win on top.

  `transaction=True` (the `broadcast/rate_limit.py` idiom) additionally means a mid-batch teardown cannot leave keys claimed while the fail-open guard discards the entries they stood for. redis-py 8.0.1's `Pipeline` shares the client's response-callback table, so `parse_set_result` still yields `True`/`None` per staged `SET` — verified against the pinned wheel. The claim step moved into `_claim_unannounced` to stay under the function-size gate.

- **Non-blocking note taken** — `extras` is set after `_reduce_and_audit_success`, so the fragment bypasses JSONFlux reduction entirely. Named in the `build_checks_alert_advisory` docstring (and in the docs' design contract) as the reason the cap is nudge-sized; not re-architected.

Tests keep the round-trips-issued style rather than length-only, and now pin both halves of the bound: `len(store.ex_values)` the `SET`s actually staged, `store.round_trips` the batching. Both mutations were run and observed to fail the right assertion, then reverted — post-claim Python cap → `assert 13 == 10`; unpipelined loop → `assert 10 == 1`. A third test pins the cap's magnitude (`<= 10`) so widening it back stays an explicit decision.

Verification: `ruff` / `ruff format` clean on `src/` + `tests/`; `mypy` clean on the touched module (the one repo-wide error is a pre-existing macOS-only `socket.MSG_ERRQUEUE` false positive in untouched `connectors/net/icmp.py`); `pytest tests/test_checks_advisory.py` 12 passed; checks + broadcast neighbours 269 passed / 3 skipped; the dispatch-path slice 476 passed / 87 skipped; `code-quality.py --diff` exit 0, no warnings.

## Summary by CodeRabbit

- **New Features**
  - Successful dispatch responses can include a `checks_alert_advisory` identifying degraded or critical Dashboards.
  - Advisories are limited, ordered by Dashboard name, and avoid repeated alerts while still reporting state changes.
  - The advisory applies across operation types and does not prevent dispatches from completing.
  - JSON output displays advisories for successful operations.

- **Documentation**
  - Added guidance covering advisory behavior, configuration, response formatting, and delivery limits.

### DIFF
--- a/backend/src/meho_backplane/checks/advisory.py
+++ b/backend/src/meho_backplane/checks/advisory.py
@@ -1,0 +1,256 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""Dispatch-time checks-alert advisory (#2718, Initiative #2716).
+
+When the caller's tenant has a Dashboard whose ``last_rollup_state``
+memo (#2506's column, maintained exactly-once-per-transition by #2507's
+compare-and-swap at the persist seam) is ``degraded`` or ``critical``,
+the next successful dispatch response carries one compact
+``extras["checks_alert_advisory"]`` entry naming it -- once per
+``(caller, dashboard, state)`` window, deduped by an atomic Valkey
+``SET NX EX``. Ambient awareness for anyone actively working through
+the backplane: agent ``call_operation`` and operator CLI dispatches ride
+the same path, so the fragment is on the wire for both without anyone
+polling anything. Reaching the *operator's terminal* is a separate
+question -- ``meho operation call`` prints ``extras`` on its human
+render only for non-ok statuses, so the advisory surfaces there through
+``--json`` today (see ``docs/codebase/checks-advisory.md``).
+
+This is the second ``extras`` fragment beside the #2550
+``target_activity_advisory`` (:mod:`meho_backplane.broadcast.history`),
+and it deliberately diverges from that precedent in one way: it applies
+to **all** op classes, read dispatches included. The #2550 gate to
+write-class ops exists because that advisory is target-overlap-specific
+-- it warns about concurrent mutation crossfire, which only matters when
+the caller is about to mutate. A non-green Dashboard concerns every
+active caller in the tenant regardless of what they are dispatching, so
+gating here would defeat the feature. The per-caller NX dedupe is what
+keeps the all-classes reach from becoming spam: each caller hears about
+a given ``(dashboard, state)`` once per window.
+
+Design contract (mirrors the #2550 mould):
+
+* **Advisory only.** Never gates, blocks, or fails a dispatch. The
+  fragment is appended to an already-successful response.
+* **Fail-open.** A broad guard swallows any DB or Valkey error,
+  warn-logs ``checks_alert_advisory_failed``, and returns ``{}`` -- the
+  lookup never converts a successful dispatch into a failure.
+* **Bounded.** Two round-trips per successful dispatch, whatever the
+  tenant's state: one indexed read (``check_dashboard_tenant_idx``) of
+  the memo column, capped at :data:`_ADVISORY_MAX_DASHBOARDS` rows, then
+  one pipelined Valkey batch staging at most that many ``SET`` commands.
+  A green tenant pays only the SELECT. Neither the cost nor the fragment
+  grows when a correlated failure reddens a whole tenant at once -- the
+  case both bounds exist for.
+* **``0`` disables.** ``CHECKS_ALERT_ADVISORY_WINDOW_MINUTES`` (default
+  30) short-circuits before any I/O when ``0``.
+* **Memo-backed states only.** The advisory reads the transition memo,
+  not the on-read rollup, so a staleness-derived ``unknown`` (which the
+  memo never holds mid-window) is not reflected -- a deliberate,
+  documented limitation acceptable for an awareness nudge. A ``NULL``
+  memo (never-transitioned Dashboard) yields no advisory.
+* **Server-derived fields only.** Each entry is
+  ``{dashboard_id, name, state}`` read straight off the row -- no
+  free-form text lifted from sensor evidence enters the op response
+  (the untrusted-prose discipline of Initiative #2543).
+"""
+
+from __future__ import annotations
+
+import uuid
+from typing import Any, Final
+
+import structlog
+from sqlalchemy import select
+
+from meho_backplane.auth.delegation import resolve_actor_sub
+from meho_backplane.auth.operator import Operator
+from meho_backplane.broadcast.client import get_broadcast_client
+from meho_backplane.db.engine import get_sessionmaker
+from meho_backplane.db.models import CheckDashboard
+from meho_backplane.settings import get_settings
+
+__all__ = [
+    "CHECKS_ADVISORY_EXTRAS_KEY",
+    "build_checks_alert_advisory",
+]
+
+_log = structlog.get_logger(__name__)
+
+#: The ``extras`` key the advisory rides on the ``OperationResult`` --
+#: sibling of :data:`meho_backplane.broadcast.history.ADVISORY_EXTRAS_KEY`.
+CHECKS_ADVISORY_EXTRAS_KEY: Final[str] = "checks_alert_advisory"
+
+#: Memo states that surface. ``degraded`` / ``critical`` only -- ``ok``
+#: needs no nudge, ``skip`` is a deliberate opt-out, and ``unknown`` in
+#: the memo means "cannot evaluate", which the investigator surface owns.
+_ADVISORY_STATES: Final[tuple[str, ...]] = ("degraded", "critical")
+
+#: How many non-green Dashboards one dispatch pays for. One SELECT is
+#: both the read and the entry source here, so this single cap bounds
+#: the query, the ``SET`` commands staged, and the fragment together --
+#: and keeping claims and entries equal is what makes a claimed dedupe
+#: key mean "this caller was told".
+#:
+#: Sized as an awareness nudge, not an audit: the rationale of the #2550
+#: precedent's
+#: :data:`~meho_backplane.broadcast.history._ADVISORY_MAX_ENTRIES`
+#: (``5``), which is the constant this one plays the role of. Its
+#: sibling ``_ADVISORY_SCAN_LIMIT = 100`` bounds a single ``XREVRANGE``
+#: -- one round-trip whatever the count -- so its *value* carries no
+#: argument for a cap that also decides what the agent reads. What the
+#: value buys here is bytes of unsolicited agent context, and entry 11
+#: of a nudge is worth ~nothing against that; a tenant that far into the
+#: red is the checks surface's problem, not an advisory's.
+#: ``ORDER BY name`` makes the truncation deterministic.
+_ADVISORY_MAX_DASHBOARDS: Final[int] = 10
+
+
+def _dedupe_key(
+    tenant_id: uuid.UUID,
+    principal_sub: str,
+    actor_sub: str | None,
+    dashboard_id: uuid.UUID,
+    state: str,
+) -> str:
+    """Valkey key deduping one ``(caller, dashboard, state)`` per window.
+
+    Caller identity is the ``(principal_sub, actor_sub)`` pair -- the
+    same pair the #2550 advisory uses for its self-activity drop -- so a
+    delegated agent and its human principal each get their own reminder.
+    An unbound actor renders as ``-``. A state change mints a new key,
+    so escalation (``degraded`` -> ``critical``) re-announces
+    immediately; window expiry re-reminds on unchanged state.
+
+    The two sub segments are free-form and the delimiter is ``:``, so
+    two distinct ``(principal, actor)`` pairs could in principle render
+    the same key (e.g. ``a:b``/``-`` vs ``a``/``b:-``). The UUID + closed-
+    vocabulary segments anchor both ends, the aliasing window is one
+    reminder wide, and the payload is advisory-only -- not worth an
+    escaping scheme.
+    """
+    return (
+        f"meho:checks:advisory:{tenant_id}:{principal_sub}:"
+        f"{actor_sub or '-'}:{dashboard_id}:{state}"
+    )
+
+
+async def _non_green_dashboards(
+    tenant_id: uuid.UUID,
+) -> list[tuple[uuid.UUID, str, str]]:
+    """Read ``(id, name, last_rollup_state)`` for non-green memo rows.
+
+    One SELECT riding ``check_dashboard_tenant_idx``; rows whose memo is
+    ``NULL`` (never transitioned) or outside :data:`_ADVISORY_STATES`
+    never match. Ordered by name and capped at
+    :data:`_ADVISORY_MAX_DASHBOARDS`, so the fragment is deterministic
+    and the caller's per-dispatch cost does not scale with how much of
+    the tenant is broken.
+    """
+    sessionmaker = get_sessionmaker()
+    async with sessionmaker() as session:
+        rows = await session.execute(
+            select(CheckDashboard.id, CheckDashboard.name, CheckDashboard.last_rollup_state)
+            .where(
+                CheckDashboard.tenant_id == tenant_id,
+                CheckDashboard.last_rollup_state.in_(_ADVISORY_STATES),
+            )
+            .order_by(CheckDashboard.name)
+            .limit(_ADVISORY_MAX_DASHBOARDS)
+        )
+        return [(row[0], row[1], row[2]) for row in rows.all()]
+
+
+async def _claim_unannounced(
+    candidates: list[tuple[str, dict[str, Any]]],
+    window_seconds: int,
+) -> list[dict[str, Any]]:
+    """Claim each candidate's dedupe key; return the entries that won.
+
+    Every key is claimed with ``SET key 1 NX EX <window-seconds>``, all
+    of them staged into one pipelined ``MULTI``/``EXEC`` batch: a single
+    awaited round-trip however many Dashboards are red. Batching is what
+    bounds the cost -- a claim has to be issued before anyone can know
+    whether it will succeed, so the NX dedupe cannot save the round-trip
+    it rides on, and an unpipelined loop would pay one per row on
+    *every* successful dispatch for as long as the tenant is red.
+
+    An entry survives only when its own claim succeeded
+    (``parse_set_result`` yields ``True`` for OK and ``None`` when the
+    key already exists -- verified against the pinned redis-py 8.0.1,
+    whose ``Pipeline`` shares the client's response-callback table, so
+    buffering does not change the shape).
+
+    ``transaction=True`` (MULTI/EXEC, the idiom in
+    ``broadcast/rate_limit.py``) so a mid-batch teardown cannot leave
+    some keys claimed while the caller's fail-open guard discards the
+    entries they stood for.
+    """
+    client = get_broadcast_client()
+    async with client.pipeline(transaction=True) as pipe:
+        for key, _entry in candidates:
+            pipe.set(key, "1", nx=True, ex=window_seconds)
+        claims = await pipe.execute()
+    return [entry for (_key, entry), claimed in zip(candidates, claims, strict=True) if claimed]
+
+
+async def build_checks_alert_advisory(operator: Operator) -> dict[str, Any]:
+    """Build the ``extras`` fragment naming the tenant's non-green Dashboards.
+
+    Returns ``{"checks_alert_advisory": [{"dashboard_id", "name",
+    "state"}, ...]}`` or an empty dict (no key added) when the advisory
+    does not apply. The empty-dict short-circuits, in order:
+
+    * the feature is disabled (``checks_alert_advisory_window_minutes
+      == 0`` -- checked before any DB or Valkey I/O);
+    * the tenant has no Dashboard with a ``degraded`` / ``critical``
+      memo (the indexed SELECT is the only I/O paid);
+    * every non-green Dashboard was already announced to this caller in
+      the current window (each ``SET NX`` found its key present).
+
+    Each surviving row's ``(caller, dashboard, state)`` dedupe key is
+    claimed through :func:`_claim_unannounced`, which batches every
+    claim into one Valkey round-trip and keeps only the Dashboards whose
+    key it won. The Valkey TTL key IS the delivery state: nothing
+    durable is persisted; a Valkey flush re-reminds each caller once.
+
+    Note for sizing :data:`_ADVISORY_MAX_DASHBOARDS`: ``extras`` is
+    attached after the dispatcher has already reduced the payload, so
+    this fragment never passes through JSONFlux/result-handle
+    reduction -- the cap is the only bound on the context it adds.
+
+    Fail-open: any error -- DB teardown, Valkey teardown, anything --
+    is swallowed by the broad guard, warn-logged as
+    ``checks_alert_advisory_failed``, and yields ``{}``.
+    """
+    window_minutes = get_settings().checks_alert_advisory_window_minutes
+    if window_minutes <= 0:
+        return {}
+    try:
+        rows = await _non_green_dashboards(operator.tenant_id)
+        if not rows:
+            return {}
+        actor_sub = resolve_actor_sub()
+        entries = await _claim_unannounced(
+            [
+                (
+                    _dedupe_key(operator.tenant_id, operator.sub, actor_sub, dashboard_id, state),
+                    {"dashboard_id": str(dashboard_id), "name": name, "state": state},
+                )
+                for dashboard_id, name, state in rows
+            ],
+            window_minutes * 60,
+        )
+    except Exception:
+        # Advisory is best-effort awareness; any failure (a DB or Valkey
+        # teardown included) must never convert a successful dispatch
+        # into a failure.
+        _log.warning(
+            "checks_alert_advisory_failed",
+            tenant_id=str(operator.tenant_id),
+        )
+        return {}
+    if not entries:
+        return {}
+    return {CHECKS_ADVISORY_EXTRAS_KEY: entries}
--- a/docs/codebase/checks-advisory.md
+++ b/docs/codebase/checks-advisory.md
@@ -1,0 +1,182 @@
+# Checks-alert advisory (in-band, dispatch-time)
+
+## Overview
+
+The second in-band `extras` fragment on successful dispatch responses
+(#2718, Initiative #2716 scope item 4): when the caller's tenant has a
+Dashboard whose `check_dashboards.last_rollup_state` memo is `degraded`
+or `critical`, the next successful dispatch — agent `call_operation` and
+operator CLI alike, **read ops included** — carries one compact
+`extras["checks_alert_advisory"]` entry naming it, once per
+`(caller, dashboard, state)` window. Ambient awareness for anyone
+actively working through the backplane; delivery to people *not*
+dispatching is the sibling email path (#2719), not this fragment.
+
+"Carries" is about the response envelope, which is the same on both
+fronts. What each front then *shows* differs, and only the agent front
+shows it unprompted today — see
+[Operator reach](#operator-reach-json-only-today).
+
+## Key types
+
+- `meho_backplane.checks.advisory.build_checks_alert_advisory(operator)`
+  — the builder; returns
+  `{"checks_alert_advisory": [{"dashboard_id", "name", "state"}, ...]}`
+  or `{}`.
+- `CHECKS_ADVISORY_EXTRAS_KEY = "checks_alert_advisory"` — the extras
+  key, sibling of `broadcast.history.ADVISORY_EXTRAS_KEY`
+  (`target_activity_advisory`, #2550).
+- `settings.checks_alert_advisory_window_minutes` — dedupe window,
+  default `30`, `ge=0`; `0` disables the feature before any I/O.
+
+## Control flow
+
+```text
+dispatcher._reduce_and_audit_success(...)   # after audit + broadcast
+  → build_target_activity_advisory(...)     # 2550 fragment (write-gated)
+  → build_checks_alert_advisory(operator)   # this fragment (all classes)
+      [gate — returns {} without any I/O when:]
+        · settings.checks_alert_advisory_window_minutes == 0
+      [otherwise:]
+        → SELECT id, name, last_rollup_state FROM check_dashboards
+          WHERE tenant_id = :t AND last_rollup_state IN
+          ('degraded','critical')            # check_dashboard_tenant_idx
+          ORDER BY name LIMIT 10             # _ADVISORY_MAX_DASHBOARDS
+        → one MULTI/EXEC batch, staging per hit: SET meho:checks:
+          advisory:{tenant}:{principal_sub}:{actor_sub|-}:{dashboard_id}:
+          {state} 1 NX EX <window-seconds>   # ONE awaited round-trip
+        → include the dashboard ONLY when its SET claimed the key
+  → wrap_ok_result(..., extras={**activity, **checks})
+```
+
+## Design contract
+
+- **All op classes.** Deliberate divergence from the write-gated #2550
+  precedent: that gate exists because the target-activity advisory is
+  mutation-overlap-specific. A non-green Dashboard concerns every active
+  caller; the per-caller NX dedupe is what keeps that from becoming
+  spam.
+- **Bounded per dispatch — two round-trips, flat.** The claims are
+  staged into a single pipelined `MULTI`/`EXEC` batch, so the Valkey
+  cost is one awaited round-trip however many Dashboards are red. That
+  matters because the NX dedupe *cannot* help here: a claim has to be
+  issued before anyone can know whether it will succeed, so an
+  unpipelined loop would pay a round-trip per row on **every**
+  successful dispatch, read ops included, for as long as the tenant is
+  red — worst exactly during the incidents the feature exists to
+  announce.
+- **`_ADVISORY_MAX_DASHBOARDS = 10`** caps the SELECT, and with it the
+  `SET`s staged and the bytes the fragment adds. It is sized as an
+  awareness nudge, not an audit — the rationale of the #2550
+  precedent's `_ADVISORY_MAX_ENTRIES = 5`, the constant it plays the
+  role of. (Its sibling `_ADVISORY_SCAN_LIMIT = 100` bounds a single
+  `XREVRANGE` — one round-trip whatever the count — so its *value*
+  carries no argument for a cap that also decides what the agent
+  reads.) What the value buys here is bytes of unsolicited agent
+  context, and `extras` is attached after the dispatcher's JSONFlux
+  reduction, so nothing downstream trims it. One query is both the read
+  and the entry source, so claims and entries share the cap — which is
+  also what makes a claimed key mean "this caller was told".
+  `ORDER BY name` makes the truncation deterministic.
+- **The Valkey TTL key IS the delivery state.** Nothing durable is
+  persisted; a Valkey flush re-reminds each caller once. A state change
+  mints a new key, so escalation re-announces immediately; window expiry
+  re-reminds on unchanged state.
+- **Caller identity** is the `(principal_sub, actor_sub)` pair — the
+  same pair #2550 uses for its self-drop — so a delegated agent and its
+  human principal are distinct callers.
+- **Memo-backed states only.** The advisory reads the transition memo
+  (maintained exactly-once-per-transition by `checks/investigate.py`'s
+  compare-and-swap), never the on-read rollup — so a staleness-derived
+  `unknown` is not reflected, and a `NULL` memo (never-transitioned
+  Dashboard) yields nothing. Documented limitation, acceptable for an
+  awareness nudge.
+- **Fail-open, advisory-only.** A broad guard swallows any DB/Valkey
+  error, warn-logs `checks_alert_advisory_failed`, and returns `{}`.
+  Never gates, blocks, or fails a dispatch.
+- **Server-derived fields only** — no free-form sensor evidence enters
+  the op response (Initiative #2543 untrusted-prose discipline).
+
+## Dependencies
+
+- `check_dashboards.last_rollup_state` (#2506's column, #2507's CAS
+  writer) — read-only here; this module never writes the memo.
+- `broadcast.client.get_broadcast_client()` — the shared fast Valkey
+  client (5 s socket timeout bounds the worst-case added latency).
+- redis-py 8.x `set(name, value, nx=True, ex=seconds)` — atomic
+  set-if-absent-with-TTL; returns `True` when claimed, `None` when the
+  key exists (`parse_set_result`).
+- redis-py 8.x `client.pipeline(transaction=True)` — async context
+  manager; `pipe.set(...)` stages the command and returns the pipeline
+  synchronously, `await pipe.execute()` is the one round-trip and
+  replies with a result per staged command, through the *same*
+  response-callback table as an unbuffered call. Same idiom as
+  `broadcast/rate_limit.py`.
+
+## Operator reach (`--json`-only today)
+
+The fragment is on the wire for every front — the backend sets it on
+the `OperationResult` envelope, so the HTTP response is identical
+whether an agent or `meho operation call` issued it. What the **Go
+CLI's default human render** does with it is the gap:
+
+```go
+// cli/internal/cmd/operation/call.go — printCallResult
+if r.Status == "ok" {
+    ... print r.Result ...
+    return                      // ← returns before the extras block
+}
+...
+if len(r.Extras) > 0 && ... {   // ← non-ok path only
+```
+
+`printCallResult` prints `extras` only on the **non-ok** branch, and
+this advisory only ever rides a *successful* dispatch. So today:
+
+| Front | Sees the fragment? |
+|---|---|
+| Agent / MCP `call_operation` | yes — it reads the JSON envelope |
+| `meho operation call --json` | yes — full envelope, verbatim |
+| `meho operation call` (human render) | **no** — extras dropped on `ok` |
+
+That is a deliberate scope line, not an oversight left unrecorded:
+teaching the human render to print `extras` on success is a CLI-wide
+UX change (it would also start printing #2550's
+`target_activity_advisory`, and the vendor verbs render through a
+different path — `dispatch.Render` — so a one-line edit here would
+leave the two fronts inconsistent). #2718 is backend-only; the CLI
+render decision wants its own Task.
+
+`cli/internal/cmd/operation/client_test.go`
+(`TestCallJSONCarriesChecksAlertAdvisoryOnOK`,
+`TestCallHumanRenderOmitsExtrasOnOK`) pins both halves of that table so
+the gap cannot close or widen silently.
+
+## Known issues / boundaries
+
+- The two free-form key segments (`principal_sub`, `actor_sub`) share
+  the `:` delimiter, so two distinct caller pairs could in principle
+  alias one dedupe key; the aliasing window is one reminder and the
+  payload is advisory-only — deliberately not escaped. No reachable
+  pair aliases today: an IdP-issued `sub` is a Keycloak UUID and agent
+  principals are pinned to `^[A-Za-z0-9_\-\.]+$`
+  (`auth/agent_principals.py`), neither of which admits a `:`.
+- The default human CLI render does not show the fragment — see
+  [Operator reach](#operator-reach-json-only-today).
+- Only the first 10 non-green Dashboards by name ride a fragment
+  (`_ADVISORY_MAX_DASHBOARDS`); the tail is permanently invisible to
+  this surface. The checks surface and the #2719 email path own
+  completeness. A tenant that far into the red has bigger problems than
+  advisory coverage. The fragment carries no truncation marker today —
+  an agent cannot tell a capped list from a complete one.
+
+## References
+
+- Task #2718, Initiative #2716 (parent goal #221).
+- Precedent: `docs/codebase/broadcast.md` § "Read path (dispatch-time
+  target-activity advisory)" (#2550).
+- Memo + CAS: `backend/src/meho_backplane/db/models.py`
+  (`CheckDashboard`), `backend/src/meho_backplane/checks/investigate.py`
+  (`_claim_rollup_transition`).
+- Tests: `backend/tests/test_checks_advisory.py`;
+  `cli/internal/cmd/operation/client_test.go` (the CLI-path pair).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,23 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Added — checks advisory: non-green Dashboards surface on dispatch responses (#2718)
+
+- Successful dispatch responses — agent `call_operation` and the
+  operator CLI's `--json` envelope, read ops included — now carry a
+  compact `extras["checks_alert_advisory"]` naming any Dashboard in
+  the caller's tenant whose rollup memo is `degraded`/`critical`, once
+  per (caller, dashboard, state) window (Valkey `SET NX EX` dedupe;
+  `CHECKS_ALERT_ADVISORY_WINDOW_MINUTES`, default 30, `0` disables).
+  At most 10 Dashboards ride one fragment, ordered by name — an
+  awareness nudge, not an audit — and their dedupe claims are staged
+  into one pipelined Valkey round-trip, so a tenant-wide outage adds
+  neither an unbounded read nor a per-Dashboard round-trip to every
+  successful dispatch.
+  Advisory only: fail-open, never gates a dispatch; second extras
+  fragment beside the #2550 target-activity advisory. The CLI's
+  default human render prints `extras` only for non-ok statuses, so
+  operators reach the fragment through `--json` today. (#2718)
 ### Added — docs site: Start here, reference architecture, and the install trail (#2671)
 
 - Fill the docs site's *Start here* and *Install & operate* sections
--- a/backend/src/meho_backplane/operations/_errors.py
+++ b/backend/src/meho_backplane/operations/_errors.py
@@ -1494,8 +1494,8 @@ def wrap_ok_result(
 
     *extras* carries envelope-extension fields the dispatcher computes on
     the success path -- currently the dispatch-time target-activity
-    advisory (#2550). ``None`` leaves the frozen model's ``{}`` default
-    intact.
+    advisory (#2550) and the checks-alert advisory (#2718). ``None``
+    leaves the frozen model's ``{}`` default intact.
     """
     if payload is None or isinstance(payload, (dict, list)):
         result_value: dict[str, Any] | list[Any] | None = payload
--- a/backend/src/meho_backplane/operations/dispatcher.py
+++ b/backend/src/meho_backplane/operations/dispatcher.py
@@ -226,6 +226,7 @@ import hvac.exceptions
 from meho_backplane.auth.operator import Operator
 from meho_backplane.broadcast.events import classify_op, scrub_secret_named_values
 from meho_backplane.broadcast.history import build_target_activity_advisory
+from meho_backplane.checks.advisory import build_checks_alert_advisory
 from meho_backplane.connectors import (
     OperationResult,
     ResolutionLabel,
@@ -772,12 +773,23 @@ async def _reduce_and_audit_success(
         redaction_policy_id=redaction.policy_id,
         handle_metadata=_handle_metadata_for_audit(handle),
     )
-    advisory = await build_target_activity_advisory(
+    activity_advisory = await build_target_activity_advisory(
         operator,
         op_id=descriptor.op_id,
         target_name=getattr(target, "name", None),
     )
-    return wrap_ok_result(op_id, summary, duration_ms, handle, extras=advisory)
+    # Second extras fragment beside #2550: the checks-alert advisory
+    # (#2718) applies to ALL op classes (its builder documents why) and
+    # is deduped per (caller, dashboard, state); distinct keys, so a
+    # plain merge can never clobber either fragment.
+    checks_advisory = await build_checks_alert_advisory(operator)
+    return wrap_ok_result(
+        op_id,
+        summary,
+        duration_ms,
+        handle,
+        extras={**activity_advisory, **checks_advisory},
+    )
 
 
 async def _run_branch_with_error_handling(
--- a/backend/src/meho_backplane/settings.py
+++ b/backend/src/meho_backplane/settings.py
@@ -1066,6 +1066,14 @@ class Settings(BaseModel):
     #: ``extras["target_activity_advisory"]`` on its response. ``0``
     #: disables the feature entirely (no stream read on any dispatch).
     dispatch_activity_advisory_window_minutes: int = Field(default=30, ge=0)
+    #: Dedupe window (minutes) for the dispatch-time checks-alert
+    #: advisory (#2718). A successful dispatch by a caller whose tenant
+    #: has a Dashboard with a ``degraded``/``critical``
+    #: ``last_rollup_state`` memo carries a compact
+    #: ``extras["checks_alert_advisory"]`` once per (caller, dashboard,
+    #: state) window. ``0`` disables the feature entirely (no DB read on
+    #: any dispatch).
+    checks_alert_advisory_window_minutes: int = Field(default=30, ge=0)
     # Broadcast v2 T2 (#2547) -- durable-announcement retention prune
     # knobs. ``days=0`` is the keep-forever opt-out sentinel;
     # ``enabled=False`` skips starting the background task entirely
@@ -1722,6 +1730,9 @@ def get_settings() -> Settings:
         ),
         dispatch_activity_advisory_window_minutes=int(
             os.environ.get("DISPATCH_ACTIVITY_ADVISORY_WINDOW_MINUTES", "30"),
+        ),
+        checks_alert_advisory_window_minutes=int(
+            os.environ.get("CHECKS_ALERT_ADVISORY_WINDOW_MINUTES", "30"),
         ),
         broadcast_announcement_retention_days=int(
             os.environ.get("BROADCAST_ANNOUNCEMENT_RETENTION_DAYS", "90"),
--- a/backend/tests/test_checks_advisory.py
+++ b/backend/tests/test_checks_advisory.py
@@ -1,0 +1,595 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""Dispatch-time checks-alert advisory (#2718, Initiative #2716).
+
+A successful dispatch by a caller whose tenant has a Dashboard with a
+``degraded``/``critical`` ``last_rollup_state`` memo carries a compact
+``extras["checks_alert_advisory"]`` naming it -- once per (caller,
+dashboard, state) window via a Valkey ``SET NX EX`` dedupe. Advisory
+only: it never gates a dispatch, and every failure mode fails open.
+
+Coverage mirrors the acceptance criteria on the issue, by name:
+
+* the ``0`` window knob short-circuits before any DB or Valkey I/O;
+* a green tenant (``NULL`` / ``ok`` memos) yields no advisory and no
+  Valkey call;
+* NX dedupe -- the same caller sees a given (dashboard, state) once per
+  window; a state change mints a new key and re-announces;
+* per-caller isolation -- distinct principals each get their own
+  reminder;
+* the ``_ADVISORY_MAX_DASHBOARDS`` cap bounds both the fragment and the
+  ``SET`` commands staged when a whole tenant goes red, and the claims
+  cost exactly one awaited round-trip however many rows survive it;
+* a Valkey teardown and a DB teardown both fail open (``{}``,
+  warn-logged, dispatch unaffected);
+* dispatch integration -- a successful ``dispatch()`` response carries
+  the fragment (read-class op included -- the deliberate divergence
+  from the write-gated #2550 precedent) and an immediate second
+  dispatch by the same caller does not;
+* coexistence -- the #2550 ``target_activity_advisory`` and this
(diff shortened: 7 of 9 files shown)


docs/codebase/checks-advisory.md
```diff
@@ -0,0 +1,98 @@
...
+  key exists.
+
+## Known issues / boundaries
+
+- The two free-form key segments (`principal_sub`, `actor_sub`) share
+  the `:` delimiter, so two distinct caller pairs could in principle
+  alias one dedupe key; the aliasing window is one reminder and the
+  payload is advisory-only — deliberately not escaped.
+- The CLI renders the fragment through the ordinary `OperationResult`
+  extras pass-through; no Go changes.
```
damir-topic: **[🛑 B1]** — the operator CLI does not render this fragment on the default human output, so this line is inaccurate

**What:** This says the CLI "renders the fragment through the ordinary `OperationResult` extras pass-through". It does not. `printCallResult` prints `extras` only on the non-ok branch — the `status == "ok"` branch returns before it ever gets there:

```go
// cli/internal/cmd/operation/call.go:259-272
if r.Status == "ok" {
    if len(r.Result) > 0 && string(r.Result) != "null" {
        pretty, err := prettyJSON(r.Result)
        if err == nil {
            fmt.Fprintln(w, pretty)
            return          // <- extras never printed
        }
        fmt.Fprintln(w, string(r.Result))
    }
    return                  // <- extras never printed
}
// status == "error" (or any non-ok dispatcher envelope).
... prints extras here only
```

The advisory only ever rides a **successful** dispatch, so on the default `meho operation call` render the operator sees nothing. It is reachable today only via `--json` ("emit the full OperationResult envelope as JSON instead of the human render") and via the API/MCP JSON envelope.

**Why it matters:** #2718's headline promise is "agent `call_operation` and operator CLI **alike**", and the CHANGELOG repeats "agent `call_operation` and operator CLI alike, read ops included". As shipped, the operator half of that is true of the wire envelope but not of what an operator actually sees. #2718's implementation sketch asked for exactly this to be checked ("verify with one CLI-path test or snapshot"); no CLI-path test exists in the diff, and the unverified claim was written into a new `docs/codebase/` artifact, where it will be trusted by the next person. Initiative #2716's DoD item (c) rests on the same claim.

**Required change:** pick one and make the repo self-consistent.

1. *Deliver it* — print `extras` on the ok branch of `printCallResult` (guarded by `len(r.Extras) > 0 && string(r.Extras) != "null"`, same shape as the error branch) and add a Go test asserting a `status=ok` result with an `extras` payload renders it. Then this doc line becomes true as written.
2. *Scope it honestly* — change this line to state that the fragment reaches the operator through `--json` / the JSON envelope, and that the default human render prints `extras` only for non-ok statuses; and soften the CHANGELOG's "operator CLI alike" to match. Then also record the human-render gap here under "Known issues / boundaries" so a follow-up ticket has an anchor.

**Source:** `cli/internal/cmd/operation/call.go:259-272` (render), `call.go:109-111` (`--json` flag help), task #2718 § Implementation sketch — "CLI note".

backend/src/meho_backplane/checks/advisory.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+    sessionmaker = get_sessionmaker()
+    async with sessionmaker() as session:
+        rows = await session.execute(
+            select(CheckDashboard.id, CheckDashboard.name, CheckDashboard.last_rollup_state)
+            .where(
+                CheckDashboard.tenant_id == tenant_id,
+                CheckDashboard.last_rollup_state.in_(_ADVISORY_STATES),
+            )
+            .order_by(CheckDashboard.name)
```
damir-topic, line 159: **[⚠️ M1]** — uncapped SELECT makes the per-dispatch Valkey fan-out unbounded, contradicting the module's own "**Bounded.**" contract

**What:** This query has no `LIMIT`. Every row it returns costs one sequential `await client.set(...)` in `build_checks_alert_advisory`, and that loop runs on **every successful dispatch of every op class** — the dedupe only decides whether a dashboard enters `entries`, not whether the `SET` is issued. So the cost is O(non-green dashboards in the tenant) per dispatch, permanently, not once per window.

The module docstring nonetheless states:

> * **Bounded.** One indexed read (`check_dashboard_tenant_idx`) of the memo column per successful dispatch, plus at most one `SET` per non-green Dashboard.

"At most one `SET` per non-green Dashboard" is a per-dashboard statement, not a bound — nothing caps the dashboard count.

**Why it matters:** the load correlates with the incident. A correlated failure is exactly what turns many dashboards `degraded`/`critical` at once, and that is also when operators and agents dispatch hardest. Twenty red dashboards means twenty sequential Valkey round-trips added to every read dispatch in the tenant for the duration. The #2550 precedent this module explicitly models itself on refuses this on purpose — `_ADVISORY_SCAN_LIMIT: Final[int] = 100` in `broadcast/history.py`, with the comment "this caps the entry count so a burst inside the window can't turn one dispatch into an unbounded read." That reasoning transfers verbatim; the write-class gate is the part that legitimately does not.

**Required change:** add a cap constant beside `_ADVISORY_STATES` and apply it, then make the docstring bullet state the real bound.

```suggestion
            select(CheckDashboard.id, CheckDashboard.name, CheckDashboard.last_rollup_state)
            .where(
                CheckDashboard.tenant_id == tenant_id,
                CheckDashboard.last_rollup_state.in_(_ADVISORY_STATES),
            )
            .order_by(CheckDashboard.name)
            .limit(_ADVISORY_MAX_DASHBOARDS)
```

with e.g. `_ADVISORY_MAX_DASHBOARDS: Final[int] = 10` documented as "a nudge, not an inventory — the operator has the dashboard list surface for the full picture", and the "Bounded." bullet reworded to name the cap. The existing `.order_by(name)` already makes the truncation deterministic rather than arbitrary.

**Source:** `backend/src/meho_backplane/broadcast/history.py` (`_ADVISORY_SCAN_LIMIT` and its rationale comment) — the precedent this module cites as its mould.

damir-topic (commented):
> **Posted as `COMMENT` state, not `REQUEST_CHANGES`.** GitHub rejects a formal
> change request from the PR author's own identity ("Can not request changes on
> your own pull request"). The `**Verdict:**` line below is binding — same
> convention as PRs #2710-#2715 and #2724.

**Anchor task:** #2718 — in-band `checks_alert_advisory` on dispatch responses, deduped per caller and state
**Surface area:** backend (+ devops-adjacent settings knob; one Go CLI read-only finding)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `056edb90fb1606e85bc6b3048522515cda311373`

The core is right: memo-read-only, never gates dispatch, fail-open on both
DB and Valkey, `0` short-circuits before any I/O, `SET NX EX` per
(caller, dashboard, state), all op classes, server-derived fields only.
Every acceptance criterion on #2718 is delivered and independently
verified below. Two findings block: one contract/doc mismatch on the
operator-CLI half of the task's headline promise, and one unbounded
hot-path fan-out that contradicts the module's own "**Bounded.**" claim.

### Acceptance criteria

- [x] Unit tests cover, by name: disable short-circuit; green ⇒ empty; NX dedupe; re-announce on state change; per-caller isolation; fail-open on Valkey and on DB — all six present in `backend/tests/test_checks_advisory.py`; **10 passed** locally.
- [x] Dispatch integration test: seeded `critical` memo ⇒ successful `dispatch()` carries `extras.checks_alert_advisory` with `{dashboard_id, name, state}`; immediate second dispatch by the same caller does not — `test_dispatch_success_carries_fragment_then_dedupes`, read-class op, verified.
- [x] `checks_alert_advisory` hits the dispatcher success path beside the #2550 merge; coexistence proven by `test_both_advisory_fragments_coexist_on_one_response`. `wrap_ok_result` is called from exactly one site in the dispatcher, so parity with #2550 is complete.
- [x] `settings.py` shows `checks_alert_advisory_window_minutes: int = Field(default=30, ge=0)` with the `#:` comment convention.
- [x] **The `BaseModel`-not-`BaseSettings` trap is genuinely handled.** `get_settings()` hand-maps the env var at `settings.py:1734`. Verified empirically rather than by reading: `CHECKS_ALERT_ADVISORY_WINDOW_MINUTES=7 → 7`, `=0 → 0`, unset `→ 30`, `=-1 → ValidationError`. The field is not inert.
- [x] No migration (no `alembic/` files; CI "Detect migration-related changes" green). CHANGELOG bullet present.
- [x] Full backend suite green — all 536 test files run in chunks: 3138 + 2762 + 2601 + 1138 + 571 + 267 + 142 + 36 + 23 + 12 passed. The only failures are the two known-environmental ones (`test_connectors_net_icmp.py::test_trace_reaches_loopback`, `test_connectors_net_dns_lookup.py::test_chosen_resolver_queries_that_nameserver`), which fail identically on untouched `main` on this host.

### Project-board alignment

|---|---|---|---|---|---|
| #2718 | Open | `Closes` (confirmed in `closingIssuesReferences`) | ✅ | ⚠️ near-full — see B1 | close on merge once B1 lands |
| #2716 | Open (parent initiative) | parent | — | partial (scope item 4 of 5) | stays open |
| #221 | Open (parent goal) | grandparent | — | — | stays open |

Nothing from siblings #2717 / #2719 / #2720 / #2721 leaks into this diff, and
nothing this diff delivers is left unlinked. Hierarchy coherent.

### Predicted vs. actual diff

- ✅ `backend/src/meho_backplane/checks/advisory.py` — new builder, right package (checks domain, not `broadcast/`), dependency-light as the task asked. No circular import: `checks/__init__.py` re-exports only the dependency-pure `assertions` + `evaluate`, so the dispatcher's module-scope import is safe.
- ✅ `backend/src/meho_backplane/operations/dispatcher.py` — single merge site, distinct keys, plain merge cannot clobber either fragment.
- ✅ `backend/src/meho_backplane/settings.py` — field **and** the `get_settings()` mapping.
- ✅ `backend/tests/test_checks_advisory.py` — the fake `SET NX` store models the real claim contract, so the dedupe assertions exercise the actual protocol rather than a mock's say-so.
- ✅ `docs/codebase/checks-advisory.md` + cross-ref in `docs/codebase/broadcast.md`.
- ❓ `backend/src/meho_backplane/operations/_errors.py` (2 lines) — unexpected but justified: `wrap_ok_result`'s docstring named only #2550 and would have gone stale.
- ⚠️ `docs/codebase/checks-advisory.md:87` — expected file, wrong content (B1).
- ⚠️ `backend/src/meho_backplane/checks/advisory.py:126-131` — expected file, missing the precedent's bound (M1).
- 🕳️ Nothing missing. I checked specifically for a Helm/values or config-reference file that enumerates knobs: the direct #2550 precedent (`DISPATCH_ACTIVITY_ADVISORY_WINDOW_MINUTES`) appears only in `CHANGELOG.md` + `docs/codebase/broadcast.md`, and the chart has a generic `extraEnv: []` escape hatch — so CHANGELOG + the new codebase doc is full parity. No chart change owed.

### Integration verification

- [x] `ruff check src/ tests/` — **All checks passed**
- [x] `ruff format --check src/ tests/` — 1372 files already formatted
- [x] `mypy` on all four touched source files — Success, no issues found
- [x] `pytest tests/test_checks_advisory.py` — 10 passed
- [x] `pytest` on the blast-radius neighbours (`test_operations_dispatcher`, `test_broadcast_target_advisory`, `test_settings`, `test_checks_investigate`) — 151 passed
- [x] Full backend suite in chunks — green modulo the two known-environmental failures
- [x] **Settings env-var round-trip executed** (not reasoned about) — see acceptance criteria above
- [x] **redis-py contract verified against the pinned 8.0.1 in-tree**, not from memory: `redis/_parsers/response_callbacks.py` maps `"SET"` → `parse_set_result`, and `redis/_parsers/helpers.py:1443-1454` returns `response and str_if_bytes(response) == "OK"` — so an NX miss (nil reply) yields `None` and a successful claim yields `True`. The module docstring's claim is accurate.
- [x] **Go CLI render path read end to end** — `cli/internal/cmd/operation/call.go:259-272` (this is what B1 turns on)
- [ ] ⚠️ not run — migrations replay: n/a, no migration in this PR
- [ ] ⚠️ not run — `docker compose up`: not exercised; the change adds no service, port, or startup dependency
- [ ] ⚠️ not run — live Valkey round-trip: the dedupe is proven against a fake implementing the real claim contract, not a real Valkey. Reasoned-only for the true TTL-expiry re-reminder path (Valkey's own contract, not this feature's logic)

- 🛑 **B1** — `docs/codebase/checks-advisory.md:87` asserts the operator CLI renders the fragment; it does not on the default human render. Required change: either print `extras` on the `status == "ok"` branch of `printCallResult` (plus a Go test), or correct the doc **and** the CHANGELOG "CLI alike" phrasing to state the `--json`-only reality and record the human-render gap under "Known issues / boundaries".
- ⚠️ **M1** — `backend/src/meho_backplane/checks/advisory.py:126-131`: the SELECT has no `LIMIT`, so the per-dispatch Valkey fan-out is O(non-green dashboards) on every successful dispatch, while the docstring claims "**Bounded.**". Required change: add a cap constant mirroring `_ADVISORY_SCAN_LIMIT` and apply `.limit(...)`; reword the "Bounded" bullet to state the actual cap.

- **Copilot's dedupe-key collision comment is not reachable — you can close it with a citation.** `auth/agent_principals.py` pins agent names to `^[A-Za-z0-9_\-\.]+$` (no `:`), so `principal_sub` is either a colonless Keycloak sub or `agent:<name>` with exactly one colon at a fixed offset; `actor_sub` likewise, or `-`. No two legal `(principal, actor)` pairs can render the same `{p}:{a}` substring. Worth replacing the docstring's weaker "not worth an escaping scheme" hand-wave with that grammar argument — it is a proof rather than a risk acceptance.
- A Valkey failure part-way through the loop discards the entries already collected **and** leaves their dedupe keys claimed, so those dashboards go unannounced for a full window. Returning the partial `entries` from the guard instead of `{}` would preserve the "once per window" contract under partial failure.
- `_ADVISORY_STATES` duplicates two members of `CheckState`; since that is a `Literal` alias rather than an enum there is nothing cleaner to reuse today, so this is a note rather than a request.

### Merge-readiness (not code findings — recorded for the orchestrator)

- **DCO is `ACTION_REQUIRED`**: commit `056edb90` carries no `Signed-off-by` trailer, so the PR cannot merge as-is. Known branch-level issue, handled outside this review.
- **`mergeable_state: dirty`** as of this review. `main` advanced to `c5fa843f` (#2723) after the branch was cut; `git merge-tree` reports exactly one conflicting file, `CHANGELOG.md` — both sides insert under `## [Unreleased]` at the same anchor. Mechanical resolution, keep both bullets.
- **Sibling #2724 (issue #2717) conflict risk assessed, content not reviewed** (out of scope): it also touches `settings.py` and `CHANGELOG.md`. The `settings.py` hunks do not overlap — #2726 edits ~1066-1074 and ~1731-1734, #2724 edits ~1321-1340 and ~1881-1900 — so that file auto-merges cleanly in either order. `CHANGELOG.md` will conflict at the same `## [Unreleased]` anchor whichever merges second; trivial, keep both entries.

### Docs
- [x] `docs/codebase/checks-advisory.md` — new, well structured, but carries the B1 inaccuracy
- [x] `docs/codebase/broadcast.md` — cross-ref added; the #2550 section stays accurate, nothing invalidated
- [x] No .NET analogies introduced; no `.planning/` references; `AGENTS.md` / `CLAUDE.md` untouched

- The `BaseModel`-not-`BaseSettings` trap is the one this repo reliably loses a round to, and it was handled correctly the first time — field *and* `get_settings()` mapping, with the test that would actually catch a missing mapping (`test_window_zero_short_circuits_no_db_no_valkey` asserts `db_stub.assert_not_called()`, which fails loudly if the env var were inert even though the fail-open guard swallows the AssertionError).
- The divergence from #2550's write-class gate is argued rather than asserted — the docstring explains *why* the precedent gates and why that reason does not transfer. That is exactly the shape a deliberate deviation should take.
- `_FakeNxStore` implements the real `True`/`None` claim contract instead of returning a canned truthy value, so the dedupe and re-announce tests prove the protocol rather than the mock.

backend/src/meho_backplane/checks/advisory.py
```diff
@@ -0,0 +1,189 @@
...
+    two distinct ``(principal, actor)`` pairs could in principle render
+    the same key (e.g. ``a:b``/``-`` vs ``a``/``b:-``). The UUID + closed-
+    vocabulary segments anchor both ends, the aliasing window is one
+    reminder wide, and the payload is advisory-only -- not worth an
+    escaping scheme.
+    """
+    return (
+        f"meho:checks:advisory:{tenant_id}:{principal_sub}:"
+        f"{actor_sub or '-'}:{dashboard_id}:{state}"
+    )
```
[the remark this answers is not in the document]
damir-topic: Checked this one properly rather than waving it off — the conclusion is that no reachable `(principal_sub, actor_sub)` pair can alias, so I'm leaving the key shape as-is.

The `agent:<name>` convention you're citing is the Keycloak **clientId** (`_CLIENT_ID_PREFIX` in `backend/src/meho_backplane/auth/agent_principals.py:98`), not the `sub` claim. `operator.sub` is the OIDC subject read straight off the validated JWT (`auth/jwt.py:864,902`), which Keycloak issues as a UUID — no colon. And where an agent name *does* appear anywhere in an identity string, it is pinned to `^[A-Za-z0-9_\-\.]+$` (`auth/agent_principals.py:95`, mirroring `agents/schemas.NAME_PATTERN`), which also excludes `:`. `actor_sub` comes from the same population via `resolve_actor_sub()`, or renders as the literal `-` when unbound.

So both free-form segments are drawn from a colonless alphabet, and the surrounding segments (tenant UUID, dashboard UUID, and a closed `degraded`/`critical` vocabulary) anchor both ends. The `a:b`/`-` vs `a`/`b:-` construction is well-formed as a string but not constructible from any identity this system issues.

Residual risk if an IdP ever did issue a colon-bearing `sub`: one caller could miss one advisory reminder for one window. Advisory-only payload, no gating, no durable state — still not worth an escaping scheme. I've recorded the grammar argument in `docs/codebase/checks-advisory.md` § "Known issues / boundaries" so the reasoning is discoverable rather than re-litigated on the next read.

backend/src/meho_backplane/checks/advisory.py
```diff
@@ -0,0 +1,211 @@
...
+#: ``100``) for the same reason: the window bounds *when* the advisory
+#: re-announces, but nothing bounds *how many* rows one dispatch pays
+#: for, so a correlated failure reddening a whole tenant would turn every
+#: successful dispatch into an unbounded read plus one Valkey ``SET`` per
+#: row. There the scan cap and the emitted-entry cap are separate
+#: constants because a stream read feeds a reducer; here one SELECT is
+#: both the read and the entry source, so this single cap bounds the
+#: query, the per-dispatch fan-out, and the fragment size together.
+#: ``ORDER BY name`` makes the truncation deterministic.
+_ADVISORY_MAX_DASHBOARDS: Final[int] = 100
```
damir-topic: **[🛑 B1]** — the cap lands, but `100` leaves the cost M1 was filed about essentially unmitigated on the dispatch hot path

**What:** `_ADVISORY_MAX_DASHBOARDS = 100` really does bound the SELECT — I compiled the statement and it ends `ORDER BY check_dashboards.name LIMIT :param_1`, so this is SQL-side, not a Python slice. But each surviving row is a *separate awaited round-trip* in the loop below, and the dedupe decides only whether a row enters `entries`, never whether the `SET` is issued. So a tenant with ≥100 red Dashboards still pays up to 100 sequential Valkey RTTs on **every** successful dispatch, read ops included. M1 said the fan-out was `O(non-green dashboards)`; at 100 it is `O(min(non-green, 100))` — for the incident case that motivated the finding, the same number.

The comment anchors `100` to #2550's `_ADVISORY_SCAN_LIMIT`, but that constant caps the entries returned by **one** `XREVRANGE` (`history.py:1159-1164`) — the precedent's per-dispatch Valkey cost is 1 round-trip, not 100. The constant whose *role* this one plays for the agent-visible fragment is `_ADVISORY_MAX_ENTRIES`, and its value is `5`, for the reason its own docstring gives: "an awareness nudge, not an audit — the caller who needs the full picture queries `meho.broadcast.recent`." Collapsing the two constants into one is fine and the comment is honest that it does so; taking the looser of the two values for the role that faces the agent is the part that needs changing.

**Why it matters:** two concrete costs, both worst at the worst time.

1. *Latency.* Up to 100 sequential RTTs added to the product's core verb. Co-located Valkey ≈ 10–30 ms; a network-separated Valkey at 1–2 ms RTT ≈ 100–200 ms on **every** operation for the duration of an incident. The overhead scales up exactly when a correlated failure reddens the tenant — when operators are dispatching hardest and want dispatch fastest.
2. *Agent context.* `extras` is set on the `OperationResult` after `_reduce_and_audit_success` has already reduced the payload, so this fragment bypasses JSONFlux / result-handle reduction entirely (`operations/_errors.py:1504-1511`). 100 entries ≈ 12 KB ≈ ~3k tokens of *unsolicited* context in one tool result — ~3× the ~4 KB threshold `CLAUDE.md` postulate 6 sets for set-shaped data.

And the entries buying that cost are worth ~0: an agent told "5 Dashboards are critical" acts identically to one told all 100.

**Required change:** make the fan-out and the fragment nudge-sized. Minimal form is this line:

```suggestion
_ADVISORY_MAX_DASHBOARDS: Final[int] = 10
```

plus the prose that quotes the number: this comment's `_ADVISORY_SCAN_LIMIT` parity paragraph (cite `_ADVISORY_MAX_ENTRIES = 5`'s nudge-not-audit rationale instead — it is the honest precedent for this role), `docs/codebase/checks-advisory.md` § Design contract + § Known issues, and the CHANGELOG's "At most 100 Dashboards". `test_tenant_wide_outage_caps_fragment_and_valkey_fanout` needs no structural change — it is written against the constant, which is why it keeps working. Any nudge-sized value (5–10) satisfies this; `10` is named so the fix is unambiguous.

Equally acceptable if you want to keep a 100-row read: pipeline the claims so the fan-out is ~1 round-trip **and** add a separate nudge-sized fragment cap. `client.pipeline(transaction=False)` exists on the pinned redis-py 8.0.1 async client (`redis/asyncio/client.py:519`; `Pipeline(Redis)` at `:1635` inherits `set`, so `nx`/`ex` semantics are unchanged), with in-tree precedent at `broadcast/rate_limit.py:166`. That route costs a pipeline double in `_FakeNxStore`; the single-constant change costs nothing.

Either way, keep the doc's tail-invisibility note pointed at whichever number ships — it already names the checks surface and #2719 as the completeness owners, so only the figure moves.

**Source:** `backend/src/meho_backplane/broadcast/history.py:1024-1036` (both precedent constants and their stated rationales) and `:1159-1164` (the single capped read); `backend/src/meho_backplane/operations/_errors.py:1504-1511` (extras is set post-reduction); redis-py `8.0.1` pinned in-tree (`redis/__init__.py:67`).

damir-topic (commented):




**Anchor task:** #2718 — in-band `checks_alert_advisory` on dispatch responses, deduped per caller and state
**Surface area:** backend, devops (+ Go CLI tests)
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `61721c28`

> Posted with GitHub review state `COMMENT` because GitHub 422s a
> `REQUEST_CHANGES` review submitted from the PR author's own identity.
> The `**Verdict:**` line above is the binding one — same convention as
> iteration 1 and PRs #2710–#2715 / #2724.

### Iteration-1 punch list — disposition

| # | Iteration-1 finding | Verified state |
|---|---|---|
| B1 | Doc/CHANGELOG claimed operator-CLI pass-through the CLI does not do | ✅ **Resolved** |
| M1 | Uncapped SELECT ⇒ unbounded per-dispatch Valkey fan-out | ⚠️ **Cap landed correctly; the chosen value does not mitigate the cost** → new 🛑 B1 |

**B1 — resolved, and resolved well.** The honest rescope is complete and
self-consistent across all four surfaces: `docs/codebase/checks-advisory.md`
§ "Operator reach (`--json`-only today)" with the per-front table and the
`printCallResult` excerpt, the module docstring, the CHANGELOG bullet
("the operator CLI's `--json` envelope"), and § "Known issues / boundaries".
I checked the doc's Go excerpt against `cli/internal/cmd/operation/call.go`
line by line — `if r.Status == "ok"` returns at `:266` and `:271`, and the
`len(r.Extras) > 0` block at `:279` is on the non-ok path only. The doc is
now accurate.

The test pair was **mutation-verified in both directions**, not just read:

- Taught `printCallResult` to print `extras` on the `ok` branch →
  `TestCallHumanRenderOmitsExtrasOnOK` fails, and its failure message names
  the doc section and the CHANGELOG bullet to update. ✅
- Made the `--json` branch drop `extras` → `TestCallJSONCarriesChecksAlertAdvisoryOnOK`
  fails with `--json envelope must carry extras on status=ok; got <nil>`. ✅

Both mutations reverted; `git status` clean afterwards. Both tests exercise
the real `runCall` path through the existing fake-client harness, not
`printCallResult` in isolation, so neither is tautological.

**M1 — the mechanics are right; the number is the problem.** Everything the
punch list asked for is there, and better than asked for:

- `.limit(_ADVISORY_MAX_DASHBOARDS)` is applied **in SQL**, not Python-side.
  Confirmed by compiling the statement: it ends
  `ORDER BY check_dashboards.name LIMIT :param_1`.
- The test asserts on `len(store.ex_values)` — Valkey round-trips *issued* —
  not just fragment length. Mutation-verified: moving the cap to a Python
  slice after the `SET` loop fails it with `assert 103 == 100`, which is
  exactly the hole a length-only assertion would have left open.
- The module docstring's "**Bounded.**" bullet now names the cap instead of
  asserting the property.

What the value leaves on the table is 🛑 B1 below. In short: 100 rows means
up to 100 *sequential awaited* Valkey round-trips per successful dispatch,
where the #2550 precedent's per-dispatch cost is one `XREVRANGE`; and the
precedent constant that plays this one's *fragment* role is
`_ADVISORY_MAX_ENTRIES = 5`, not `_ADVISORY_SCAN_LIMIT = 100`.

**On the judgment call the fixer flagged for scrutiny** — the answer is that
`100` is not defensible here, but not primarily for the reason the punch list
implied. The ~12 KB of agent context is the *weaker* half of the objection:
the NX dedupe makes it one-shot per caller per window. The load-bearing half
is the one the dedupe does not touch — the `SET` is issued before anyone
knows whether it will claim, so the round-trip count recurs on **every**
dispatch for as long as the tenant is red. That is the cost M1 was filed
about, and bounding it at 100 does not reduce it for the incident case.
Copilot reached the same conclusion independently (suppressed comment on
`advisory.py:199`, suggesting a pipeline).

### Acceptance criteria (#2718)

- [x] Unit tests cover by name: disable short-circuit; green ⇒ empty; dedupe NX; re-announce on state change; per-caller isolation; fail-open on Valkey and on DB — 11 passed in `tests/test_checks_advisory.py`
- [x] Dispatch integration test: seeded `critical` memo ⇒ `extras.checks_alert_advisory`; immediate second dispatch by the same caller does not — `test_dispatch_success_carries_fragment_then_dedupes`
- [x] `checks_alert_advisory` at the dispatcher success path beside the #2550 merge; coexistence test present — `dispatcher.py:783-792`, `test_both_advisory_fragments_coexist_on_one_response`
- [x] `Field(default=30, ge=0)` with the `#:` convention, plus the `get_settings()` env mapping — `settings.py:1069-1076`, `:1734-1736`
- [x] No migration; CHANGELOG bullet present; suites green
- [x] **Operator CLI carries the fragment** — the criterion iteration 1 failed. Now met by honest scoping plus the CLI-path test #2718 asked for and never got. This was the right call over the Go change: `printCallResult` returns inside its `ok` branch, so moving the extras block would also start printing #2550's `target_activity_advisory`, and the vendor verbs render through `dispatch.Render` — a CLI-wide UX decision, correctly deferred to its own Task.

**Every acceptance criterion on #2718 is now met.** The blocker below is a
promoted ⚠️ Major, not a missing criterion — promoted per the binary-verdict
rule because the residual cost sits on the hot path of the product's core
verb, scales up precisely during the incidents the feature exists to announce,
and is a ~1-line fix that would otherwise ship as a documented-but-unargued
default.

### Project-board alignment

|---|---|---|---|---|---|
| #2718 | Open | `Closes` (confirmed in `closingIssuesReferences`) | ✅ | ✅ full | close on merge once B1 lands |
| #2716 | Open | parent initiative | — | partial (scope item 4 of 5) | stays open |
| #221 | Todo | parent goal | — | partial | stays open |

No delivered-but-unlinked issue; no linked-but-undelivered issue; hierarchy coherent.

### Predicted vs. actual diff

Full PR diff vs. merge-base is 9 files / +1065 / −4 — unchanged in shape from
iteration 1. The two `main` merge commits (`37caa7c3`, `61721c28`) bring in
only `docs-site/`, `mkdocs.yml`, `.pre-commit-config.yaml` and CHANGELOG
sections from #2723 / #2725; **no backend source arrived via the merges**, so
iteration 1's full-suite green still covers the blast radius.

- ✅ `checks/advisory.py`, `operations/dispatcher.py`, `settings.py`, `operations/_errors.py`, `tests/test_checks_advisory.py`, `docs/codebase/broadcast.md`, `CHANGELOG.md`
- ✅ `docs/codebase/checks-advisory.md` — was ⚠️ in iteration 1; now accurate
- ✅ `cli/internal/cmd/operation/client_test.go` — new, and predicted (the missing CLI-path test)
- ⚠️ `checks/advisory.py:100` — the cap value, see inline
- 🚫 none undisclosed · 🕳️ none missing

### Integration verification

Run inside `.claude/worktrees/pr-2726-review` at `61721c28`.

- [x] `ruff check src/ tests/` — All checks passed
- [x] `ruff format --check` — 1372 files already formatted
- [x] `mypy` on the 4 touched source files — Success, no issues
- [x] `pytest tests/test_checks_advisory.py` — 11 passed
- [x] `pytest` blast radius (advisory + dispatcher + broadcast advisory + settings + investigate) — 162 passed
- [x] **SQL LIMIT proven, not assumed** — compiled the select; `LIMIT :param_1` present
- [x] **Mutation test (M1):** cap moved to a post-loop Python slice ⇒ test fails `assert 103 == 100` on `ex_values`
- [x] `go build ./...`, `gofmt -l`, `go vet ./...`, `go test ./...` on `cli/` — all clean
- [x] **Mutation tests (B1):** both boundary directions fail their intended test
- [x] Copilot's two new reviews (09:57, 10:00) read in full, `<details>` blocks included
- [x] GraphQL `reviewThreads` — 3 threads, all unresolved (see merge readiness)
- ⚠️ not run: full backend suite — iteration 1 ran it in chunks green at `056edb90`; no backend source has changed since except `advisory.py` (`.limit()` + docstring) and its own test file. CI's `Python (ruff + mypy + pytest)` was IN_PROGRESS at review time.
- ⚠️ not run: `docker compose up`, migrations replay (no migration; `Detect migration-related changes` SUCCESS), agent eval
- reasoned-only: live Valkey round-trip — the fake implements the real `True` / `None` claim contract; TTL expiry is Valkey's contract, not this feature's logic

- 🛑 **B1** — `_ADVISORY_MAX_DASHBOARDS = 100` bounds the SELECT correctly but still costs up to 100 sequential Valkey round-trips per successful dispatch, and ships a 20× larger agent-visible fragment than the precedent it cites; required change: set the constant to a nudge-sized value (`10`), re-anchor the rationale comment to `_ADVISORY_MAX_ENTRIES = 5` rather than `_ADVISORY_SCAN_LIMIT = 100`, and update the three prose sites that quote the number (module comment, `docs/codebase/checks-advisory.md` § Design contract + § Known issues, CHANGELOG). Alternative accepted: pipeline the claims **and** add a separate nudge-sized fragment cap.

- **Copilot's second suppressed comment is fair but is codebase parity, not a regression:** the fail-open guard warn-logs `checks_alert_advisory_failed` without `exc_info`, so a production DB/Valkey failure is undiagnosable from the log line. The #2550 precedent has exactly the same shape (`history.py:1169-1173`). Worth a small follow-up that adds `exc_info=True` to *both* guards together rather than diverging one of them here.
- A Valkey failure part-way through the `SET` loop discards the entries already collected **and** leaves their dedupe keys claimed, so those Dashboards go unannounced for a full window. Returning the partial `entries` from the guard instead of `{}` would preserve the "once per window" contract under partial failure. Carried over from iteration 1; correctly scoped out as a behaviour change.
- If a nudge-sized cap lands, a `truncated` / `more` marker on the fragment would tell an agent to go look at the checks surface. Design addition, deliberately *not* required by B1.
- `_ADVISORY_STATES` duplicates two members of `CheckState`; `CheckState` is a `Literal` alias rather than an enum, so there is nothing cleaner to reuse today.

### Merge readiness (not review findings — orchestrator's)

- **DCO is `ACTION_REQUIRED`:** commit `056edb90` carries no `Signed-off-by`. All four commits added this iteration do. Known, handled outside this review.
- **Three unresolved review threads** under the team's "all comments resolved" policy: iteration 1's B1 thread (now `isOutdated`, fixed) and M1 thread (fixed), plus Copilot's dedupe-key thread — where the author's reply is correct. I re-verified it independently rather than relaying: `auth/agent_principals.py` pins agent names to `^[A-Za-z0-9_\-\.]+$` and uses `agent:` only as the Keycloak *clientId* prefix, while `operator.sub` is the OIDC subject off the validated JWT (`auth/jwt.py`), so both free-form key segments are drawn from a colonless alphabet. All three threads are resolvable by the author without code churn.
- `mergeable: true`, `mergeable_state: blocked` — attributable to DCO plus two IN_PROGRESS checks, not to a conflict. Both `main` merges kept both CHANGELOG bullets. PR #2724 will still conflict on the `## [Unreleased]` anchor whichever merges second; trivial, keep both entries.

### Docs

- [x] `docs/codebase/checks-advisory.md` — **updated and now accurate**; the iteration-1 inaccuracy at the old line 87 is gone, replaced by a section that documents the real per-front behaviour and points at the tests that pin it. Only the `100` figure needs to follow B1.
- [x] `docs/codebase/broadcast.md` — cross-ref intact, #2550 section not invalidated
- No .NET analogies, no `.planning/` references, `AGENTS.md` / `CLAUDE.md` untouched

### What worked well

- **The rescope was the harder and better call.** Choosing option 2 meant writing a gap down rather than papering over it, and the reasoning for why the Go change is not a one-liner (the shared `ok` return, `dispatch.Render` for the vendor verbs) is correct — verified against `call.go` rather than taken on trust.
- **The B1 test pair is genuinely bidirectional.** Pinning a *gap* with a test that fails loudly, carrying a message that names the doc and CHANGELOG to update, is materially better than the single test #2718 asked for. Both mutations confirm it works.
- **The M1 test asserts the right quantity.** `len(store.ex_values)` rather than fragment length is precisely the distinction between a real fan-out cap and a cosmetic one, and the mutation proves it catches the cosmetic version.
- Flagging the cap value for reviewer scrutiny with the tradeoff written out is exactly how a judgment call should be handed over — B1 is an adjudication of that hand-off, not a surprise.

backend/src/meho_backplane/checks/advisory.py
damir-topic: Fixed in `466ddcbc` — and the diagnosis is right on the part that matters. The `SET` is issued before anyone can know whether it claims, so the NX dedupe saves nothing on the round-trip axis: the fan-out recurred on every successful dispatch, read ops included, for as long as the tenant stayed red. Agreed too that `_ADVISORY_SCAN_LIMIT = 100` was the wrong half of the precedent to borrow from — it bounds one `XREVRANGE`, so its *value* carries no argument for a cap that also decides what the agent reads.

Took both remedies, but **not option 2 verbatim**, and it's worth saying why. Keeping the 100-row read with a separate nudge-sized *fragment* cap breaks this module's load-bearing invariant: the Valkey key IS the delivery record. Claiming 100 keys while emitting 5 entries marks 95 Dashboards "announced" without announcing them — silently suppressed for a full window, which is strictly worse than the tail-invisibility the doc already accepts and names owners for. Repairing that needs either a compensating `DEL` over the unannounced claims or a claim-until-K loop that defeats the batching; neither stays small.

So: **one** cap, nudge-sized at `10`, which keeps claims and entries equal and therefore keeps a claimed key meaning "this caller was told" — **plus** the pipeline, where at that size batching is pure win. Per successful dispatch: one awaited Valkey round-trip instead of up to 100, and at most 10 entries instead of 100.

`transaction=True` (the `broadcast/rate_limit.py` idiom) additionally means a mid-batch teardown can't leave some keys claimed while the fail-open guard discards the entries they stood for — which narrows the partial-failure case carried over from iteration 1. Verified against the pinned wheel rather than assumed: `Pipeline` shares the client's response-callback table, so `parse_set_result` still yields `True`/`None` per staged `SET`, and `pipe.set(...)` stages synchronously and returns the pipeline.

On the non-blocking JSONFlux note: correct, and it's now the stated reason the cap is nudge-sized rather than a footnote — `extras` is attached after `_reduce_and_audit_success` has already reduced the payload, so the cap is the fragment's only bound. Named in the `build_checks_alert_advisory` docstring and the docs' design contract; not re-architected.

Tests keep the round-trips-issued style and now pin both halves of the bound — `len(store.ex_values)` for `SET`s staged, `store.round_trips` for the batching. Both mutations were run and observed: post-claim Python cap → `assert 13 == 10`; unpipelined loop → `assert 10 == 1`. A third test pins the magnitude (`<= 10`) so widening it back stays an explicit decision.

Rationale re-anchored at every prose site: the constant's comment, the module docstring's `**Bounded.**` bullet, `docs/codebase/checks-advisory.md` (control flow, design contract, dependencies, known issues) and the CHANGELOG bullet. The Known-issues entry also now records that the fragment carries no truncation marker.

damir-topic (commented):




**Anchor task:** #2718 — in-band `checks_alert_advisory` on dispatch responses, deduped per caller and state
**Surface area:** backend (+ devops posture on the Valkey client)
**Verdict:** APPROVE
**Head SHA reviewed:** `466ddcbc`

> Posted in `COMMENT` state because GitHub rejects `addPullRequestReview` with
> `APPROVE` from the pull request author's own identity ("Can not approve your own
> pull request") — same constraint that forced iterations 1 and 2 to post
> `REQUEST_CHANGES` as a comment. The `**Verdict:**` line above is the binding
> signal; the machine-readable result file records `"verdict": "APPROVE"`.

Iterations 1 and 2 cleared everything except the cap value. `466ddcbc` is the only
delta since, and it is the subject of this review.

### Disposition of iteration 2's B1

**RESOLVED.** The fix did two things, and both were verified against source rather
than taken on report.

**1. `_ADVISORY_MAX_DASHBOARDS` 100 → 10, re-anchored.** The constant's rationale
now argues from `broadcast/history._ADVISORY_MAX_ENTRIES = 5` — the constant whose
role it actually plays — and states outright why `_ADVISORY_SCAN_LIMIT = 100`'s
*value* carries no argument here (that one bounds a single `XREVRANGE`, one
round-trip whatever the count). Both cited constants exist at
`broadcast/history.py:1027` and `:1036` with the cited values. `grep -n 100` over
`advisory.py` and `docs/codebase/checks-advisory.md` now matches exactly one line
each — the deliberate contrast sentence. No justification-by-wrong-precedent left
anywhere.

**2. The per-hit `await client.set(...)` loop → one pipelined `MULTI`/`EXEC`.**
Verified against the pinned redis-py **8.0.1** in the worktree venv
(`redis/__init__.py:67`), not from memory:

- `redis/asyncio/client.py:1868 _execute_transaction` builds
  `cmds = (MULTI, *commands, EXEC)`, packs them with `connection.pack_commands(...)`
  and issues **one** `await connection.send_packed_command(all_cmds)`. One
  round-trip is exact, not approximate.
- `client.py:1748 execute_command` routes to `pipeline_execute_command` whenever
  `self.watching` is falsy and `args[0] != "WATCH"` — which appends to
  `command_stack` and `return self` **synchronously**. So the un-awaited
  `pipe.set(...)` is correct, not a silently-dropped coroutine.
- `_execute_transaction` applies the response callbacks itself ("We have to run
  response callbacks manually", `self.response_callbacks[command_name](r, **options)`),
  so `parse_set_result`'s `True`/`None` shape survives buffering. The docstring's
  claim is exact.
- `get_broadcast_client()` builds a standalone client via `redis.from_url(...)`
  (`broadcast/client.py:147`) — not `RedisCluster` — so a multi-key `MULTI` has no
  cross-slot constraint.
- `Pipeline.__aexit__` → `await self.reset()` (`client.py:1677`), so the connection
  returns to the pool whether or not `execute()` raised.
- `broadcast/rate_limit.py:165-168` uses the identical idiom. The cited precedent
  is real.

### The crux: are the claim set and the emitted fragment the same set?

This was the thing worth checking, because the fixer rejected the "read 100, emit
10" variant on exactly this ground. **The reasoning holds**, and it is now
test-pinned rather than merely argued:

- Every short-circuit — window `<= 0`, DB teardown, zero rows, client construction,
  `pipeline()` raising, staging raising — happens **before** any command reaches the
  wire. Nothing is claimed.
- On the success path the surviving entries are
  `[entry for (_key, entry), claimed in zip(candidates, claims, strict=True) if claimed]`
  — precisely the claims that won, in order.
- Mis-pairing is unreachable: redis-py raises `ResponseError("Wrong number of
  response items from pipeline execution")` before the strict `zip` could ever see
  a length mismatch.
- A **mid-batch** teardown cannot leave a partial claim set, because MULTI…EXEC goes
  out as one packed write and the server applies it atomically. The docstring's
  wording is accurate as written.

And the invariant is caught by the tests, not just asserted in prose — I re-created
the rejected variant (`.limit(cap * 10)` on the SELECT plus `[:cap]` on the entries)
and `test_tenant_wide_outage_caps_fragment_and_valkey_fanout` failed with
`assert 13 == 10` on `len(store.ex_values)`. Claiming more than you emit is a test
failure now.

### Mutation spot-checks (run here, not relayed)

All three mutations the fixer reported were re-run independently in the review
worktree. Each failed the intended assertion; each was reverted and the suite
re-run green:

| Mutation | Failing assertion |
|---|---|
| unpipelined per-row `execute()` | `assert 10 == 1` (`store.round_trips`) |
| cap widened back to `100` | `assert 100 <= 10` |
| claim 100, emit 10 (the rejected variant) | `assert 13 == 10` (`len(store.ex_values)`) |

Both halves of the bound are genuinely pinned: `len(store.ex_values)` counts `SET`s
actually staged (catches a post-claim Python cap) and `store.round_trips` counts
awaited `execute()`s (catches an unpipelined loop). `_FakeNxStore` is faithful to
the redis-py semantics verified above — sync `pipeline()`, sync staging `set()`
returning self, awaited `execute()` returning one `True`/`None` per staged command
— and it asserts `transaction=True`, `nx=True` and `ex > 0` on the way through.

### Acceptance criteria
- [x] Bounded per-dispatch Valkey cost — one awaited round-trip, ≤10 staged `SET`s; both mutation-verified
- [x] Cap sized on its own merits, every prose site re-anchored — constant comment, module docstring, both function docstrings, four doc sections, CHANGELOG
- [x] JSONFlux-bypass reason stated and true — `wrap_ok_result` (`_errors.py:1479`) takes the already-reduced payload and the already-built handle and attaches `extras` verbatim; nothing downstream trims it
- [x] #2718's own seven criteria still met — no contract changed; fragment shape, dedupe grammar, fail-open, all-op-class reach, disable knob and CLI reach untouched, and their tests pass
- [x] Signed-off-by on `466ddcbc`

### Project-board alignment
|---|---|---|---|---|---|
| #2718 | Open | Closes | ✅ | ✅ full | close on merge |
| #2716 | Open | parent initiative | — | ⚠️ partial (item 4 of 5) | stays open |
| #221 | Todo | parent goal | — | ⚠️ partial | stays open |

### Predicted vs. actual diff
- ✅ `checks/advisory.py`, `tests/test_checks_advisory.py`, `docs/codebase/checks-advisory.md`, `CHANGELOG.md` — the four files a cap-and-batching fix should touch, and only those
- ✅ Full PR shape unchanged at 9 files / +1189 / −4 vs merge-base; no stealth edits arrived with the fix

### Integration verification
All run inside `.claude/worktrees/pr-2726-review` at `466ddcbc`.

- [x] `ruff check src/ tests/` — All checks passed
- [x] `ruff format --check` — 1372 files already formatted
- [x] `mypy src/meho_backplane/checks/advisory.py` — Success, no issues
- [x] `pytest tests/test_checks_advisory.py` — 12 passed
- [x] blast radius (advisory + dispatcher + #2550 advisory + rate-limit + rollup + settings) — 152 passed, 1 skipped
- [x] checks/advisory slice (`-k "advisory or checks"`) — 254 passed, 1 skipped
- [x] dispatch/operation slice (`-k "dispatch or operation"`) — 1428 passed, 106 skipped
- [x] redis-py 8.0.1 pipeline semantics — read from the pinned wheel in-tree, cited above
- [x] 3 mutations run and reverted; worktree clean afterwards
- ⚠️ not run: full backend suite (546 files). Iteration 1 ran it green; the only backend source changed since is `advisory.py`, whose single consumer is `dispatcher.py:785`, and the 1428-test dispatch slice covers that seam. CI's `Python (ruff + mypy + pytest)` is the backstop and was still in flight at review time.
- n/a: frontend, migrations replay, Go (no Go files touched this iteration)

### Findings (inline)
None. No blockers, no majors, and nothing minor that carries a required change —
so nothing is posted inline, per the "every inline comment is a merge blocker"
policy.

Recorded for the trail; none blocks this merge or needs a reply here.

- **redis-py retries the batch, and `SET NX` is not idempotent in its *return
  value*.** `Pipeline.execute` wraps the call in `conn.retry.call_with_retry(...)`
  and `_disconnect_raise_on_watching` does not re-raise when `watching` is falsy, so
  a connection that dies *after* the server applied `EXEC` but before the client
  read the reply gets the whole `MULTI`/`EXEC` re-sent; the re-issued `SET NX`s find
  their keys present, every result is `None`, and the fragment is suppressed for one
  window with the keys claimed. Worth naming because it is the one residual path
  where claims and entries diverge — but it is **not introduced here**: the previous
  per-`await client.set(...)` loop carried the identical hazard on every one of its N
  commands, so the pipeline strictly *reduces* the exposure from N retryable
  operations to one. Consequence is one missed nudge inside a 30-minute window on an
  explicitly fail-open advisory. Same family as the partial-failure note carried from
  iteration 1; both would be closed by returning the collected `entries` from the
  guard rather than `{}`.
- The module docstring's `**Bounded.**` bullet opens "Two round-trips per successful
  dispatch, whatever the tenant's state" and then says "A green tenant pays only the
  SELECT" two sentences later. Self-correcting on a careful read, but "at most two"
  would say it once.
- Copilot's `exc_info` finding still stands as codebase parity — the #2550 guard at
  `history.py:1169-1173` warn-logs without exception context too. Worth one follow-up
  that adds `exc_info=True` to both rather than diverging one.
- The fragment still carries no truncation marker, so an agent cannot distinguish a
  capped list from a complete one. Now honestly recorded under "Known issues /
  boundaries" — a design addition, correctly not forced into this PR.
- The CLI human render dropping `extras` on `status=ok` is documented and test-pinned
  in both directions, but remains a real product gap that hits #2550's
  `target_activity_advisory` equally. Deserves its own Task.

### Docs
- [x] `docs/codebase/checks-advisory.md` — updated and accurate. Control-flow block,
  § Design contract (now split into the round-trip bullet and the cap's own bullet),
  § Dependencies (the pipeline contract, with the response-callback detail), and
  § Known issues all re-anchored; the new "no truncation marker" admission is a
  genuine addition rather than a restatement. No .NET analogies, no `.planning/`
  references, `AGENTS.md` / `CLAUDE.md` untouched.

### Merge readiness (orchestrator, not the author)
- DCO fails on the pre-existing commit `056edb90` (no `Signed-off-by`). `466ddcbc`
  and the four commits before it all carry the trailer. Out of this review's scope.
- Three `Python`/image checks were still pending at review time; every other
  non-skipped check is green (14 pass).
- Four review threads are open. All are resolvable without code churn: iteration 1's
  B1 (outdated, fixed), iteration 1's M1 (fixed), iteration 2's B1 (fixed by
  `466ddcbc`), and Copilot's dedupe-key thread (disputed correctly, re-verified
  independently at iteration 2). Team policy needs them closed before merge.
- `CHANGELOG.md` will conflict with sibling PR #2724 at the `## [Unreleased]` anchor
  whichever lands second. Trivial — keep both bullets.

### What worked well
- The fixer rejected the offered remedy for the right reason rather than applying it
  verbatim. "Claiming 100 keys while emitting 10 marks ~90 Dashboards
  announced-but-never-announced" is correct, and preserving claims-equals-entries is
  what keeps a claimed key meaning "this caller was told". Verified rather than
  accepted — and the invariant is now pinned by a test, so the next person cannot
  quietly break it.
- Splitting `store.ex_values` (fan-out *size*) from `store.round_trips` (fan-out
  *cost*) is the right test shape. Either assertion alone would have missed one of
  the two mutations.
- The constant's comment explains why the precedent's *value* does not transfer even
  though the precedent's *reasoning* does. That distinction is what the iteration-2
  finding was actually about, and it landed.

Issue #1778, opened 2026-06-15
Task: build the /ui/approvals approve/deny surface — notifications bell + modal over a session BFF + live SSE


damir-topic:
Parent goal: #336
Parent initiative: #1775

## Summary

Let operators **approve/deny actions in the console**: a **notifications bell/badge** in the app-shell shows new approval requests (live); clicking opens a **modal** with the request detail and **Approve / Deny** buttons. The approval queue + decision endpoints + lifecycle broadcasts already exist (Goal #800); there is no UI for them today.

## Current gap (grounded on v0.15.0)

`grep approval` across `ui/` finds only the connectors ops-matrix `requires_approval` *badge* — no approvals page, bell/badge, or approve/deny modal; no `/ui/approvals` route/template.

## Backend surface to reuse + the load-bearing correction

- **Queue + decisions exist:** `operations/approval_queue.py` — `list_pending()` (`:884-912`), `approve_request()` (`:386-486`), `reject_request()` (`:489-556`), self-approval guard `_check_self_approval()` (`:802-831`, requester≠approver, #1401), and `publish_approval_event()` (`:937-984`, emits `approval.pending|approved|rejected|expired` BroadcastEvents).
- **⚠️ Do NOT POST from the browser to `/api/v1/approvals/*`.** Those endpoints (`api/v1/approvals.py:288-533`) are **Bearer-gated** (`_require_operator`) — a session cookie + CSRF cannot authenticate them. **Add a session BFF:** new `/ui/approvals` sub-routes (`require_ui_session`, CSRF-protected) that call the `approval_queue` service in-process (`list_pending`/`approve_request`/`reject_request`) — the standard console-surface pattern.
- **Live notifications:** subscribe the bell to the existing session-gated **`/ui/broadcast/stream`** SSE bridge filtered to `op_class=approval` (`ui/routes/broadcast/stream.py:235-293`) — no new SSE endpoint. Use the `dashboardFeedTray` hidden-sink + Alpine pattern (`static/src/app/dashboard-feed.js`) so event fields stay inert.

## Proposed implementation

1. `ui/routes/approvals/routes.py` `build_approvals_router()` (`require_ui_session`):
   - `GET /ui/approvals` → list page (pending requests via `list_pending`), and/or a `GET /ui/approvals/{id}` modal fragment with the request detail (op_id, connector_id, `proposed_effect`, requester `principal_sub`, created_at).
[...]

PR #2729, opened 2026-08-01
feat(checks): checks.transition broadcast events + checks op-class on the tenant feed

### LINKED ISSUE #2720: Task: checks.transition broadcast events + checks op-class on the tenant feed

damir-topic:
## Summary

- Every claimed Dashboard rollup edge now publishes one
  `checks.transition` event to the tenant broadcast feed, in **both**
  directions — a watcher who saw a Dashboard go `critical` sees it clear.
  Publishing sits on #2719's compare-and-swap claim-win branch, so
  exactly-once-per-transition across replicas is inherited from the CAS,
  not re-implemented.
- New `checks` op-class on `classify_op`, so `meho.broadcast.recent` /
  `.watch` with `filter.op_class=checks`, `GET /api/v1/feed`, and
  `/ui/broadcast/stream?op_class=checks` all narrow to check state
  changes server-side — the shape the approvals console already uses for
  `approval.*`. No new consumer built here.
- Fail-open twice over, and the two halves log in **different** places:
  `publish_event` swallows Valkey errors and reports them on its existing
  feed-wide `broadcast_publish_failed` / `broadcast_publish_errors_total`
  signals (it never re-raises, so an outage never reaches this module),
  while the new publisher wraps its body in a second guard that logs
  `checks_transition_broadcast_failed` for a failure to *build* the event.
  Either way the committed `last_rollup_state` memo, the #2719 email, and
  the runner's persist path are untouched.
- No migration, no new settings, no console bell, no event-outbox
  routing, no durable event table — the Valkey stream is the carrier,
  exactly as for `approval.*`.

## Two AC details corrected

Both recorded in full on the issue
([comment](https://github.com/evoila/meho/issues/2720#issuecomment-5151571152));
neither changes what the task delivers. The item-1 rationale below is the
corrected one — see the follow-up section for what was wrong with the
original.

**1. Exact-membership allowlist, not a `checks.` prefix branch.** The
`checks.` prefix is already taken: `api/v1/checks.py` binds
`checks.assignment.put` / `.get` and `checks.results.post` as
`audit_op_id`s.

What is *not* at risk is what those rows persist. Each of the three
routes also binds an explicit `audit_op_class` contextvar (`write` /
`read` / `write`), and `resolve_broadcast_detail` takes that
`op_class_override` in preference to `classify_op`. The same value is
what the audit row stores in `payload["op_class"]`, and that stored value
is what `meho.audit.query` filters on in SQL — so `classify_op` never
decides the gateway rows' class on the write path, and no saved
`op_class=write` query would change its results.

The hazard is on the **read** side: `classify_op` is re-run at render
time against the stored op-id by the audit drawer and the broadcast event
drawer, supplying the displayed class and feeding `is_aggregate_only`. A
prefix branch would relabel every `/api/v1/checks/*` row there —
retroactively, since the class is derived on read rather than read back
off the row — and would sweep in any future non-transition `checks.*`
op-id by construction. Implemented with `_CHECK_EVENT_OPS`, the same
idiom `_CREDENTIAL_MINT_OPS` already uses in the same function.

**2. `broadcast_recent` did *not* pick the class up for free.**
`filter.op_class` carries `"enum": list(OP_CLASS_ENUM)` and
`mcp/handlers.py:337` validates every `tools/call` against the advertised
`inputSchema` with `jsonschema.Draft202012Validator` — verified against
the installed 4.26.0 that an out-of-enum value is rejected before the
handler runs. `checks` added to `OP_CLASS_ENUM`; the AC-4 test posts a
full `tools/call` rather than calling the handler, so the enum is what is
under test.

## Test plan

- [x] `ruff format --check src/ tests/` — 1382 files already formatted
- [x] `mypy src/` — 1 pre-existing darwin-only error
      (`connectors/net/icmp.py` `MSG_ERRQUEUE`); CI runs Linux
- [x] **AC1** `classify_op("checks.transition") == "checks"` —
      `test_checks_transition_maps_to_checks`, plus the doctest and
      `test_gateway_checks_op_ids_are_not_swept_into_checks`, which pins
      what `classify_op` returns for the three gateway op-ids *in
      isolation* (`write` / `read` / `other`) — the drawer label, not the
      class those rows ship with (`write` / `read` / `write`, bound by
      their routes)
- [x] **AC2** forced `ok→critical` publishes exactly ONE event under two
      raced `investigate_on_transition` calls; recovery publishes the
      `critical→ok` event —
      `test_concurrent_transitions_publish_exactly_one_event`. Plus
      `test_second_persist_on_an_unchanged_rollup_does_not_republish`
      guarding against the publish being wired outside the claim-win
      branch
- [x] **AC3** raising publish seam → memo committed, mail still sent,
      `checks_transition_broadcast_failed` logged —
      `test_broadcast_failure_leaves_the_memo_and_the_mail_intact`,
      `test_publisher_failure_is_swallowed_and_logged`. The injected
      failure stands in for a *pre-publish* fault; a real Valkey outage is
      absorbed by `publish_event` and reports on
      `broadcast_publish_failed` / `broadcast_publish_errors_total`
      instead
- [x] **AC4** `meho.broadcast.recent` with `filter.op_class=checks`
      returns the event over a full JSON-RPC `tools/call` —
      `test_filter_op_class_checks_reaches_the_handler`
- [x] **AC5** no migration, no new settings; CHANGELOG bullet present
- [x] Backend suite green (`tests/`, minus two known-environmental
      network deselects and one pre-existing `test_ui_audit.py` teardown
      error reproduced on `main`)

## Out of scope

- Checks console notification bell / replacing the `/ui/checks` 30 s HTMX
  poll — enabled by this event, not built here.
- Event-outbox routing (matcher is a no-op pending #826) and any durable
  event table.
- #2721 (per-dashboard investigator prompt + emailed findings) is landing
  concurrently in the same area.

## Adjacent findings (not touched)

- `OP_CLASS_BADGE_CLASSES` / `OP_CLASS_FILTER_OPTIONS`
  (`ui/routes/broadcast/feed.py:104,121`) omit `approval`,
  `credential_write`, and now `checks`. Unlisted classes fall back to
  `badge-ghost` and stay reachable via an explicit query parameter, so
  nothing breaks — the UI dropdown just cannot name them. Same gap
  `approval` has had since #1778; worth one follow-up fixing all three.
- `meho.audit.query` shares `OP_CLASS_ENUM` as one taxonomy (pinned by
  `test_mcp_tools_list_shape_conventions.py`), so it now advertises an
  `op_class=checks` filter that is always empty — a transition has no
  audit row. Documented in the constant's docstring and in
  `docs/codebase/checks-broadcast.md`.
- `ui/routes/broadcast/aggregate_gate.py:104` states that `classify_op`
  "yields the same class the broadcast publisher computed". Untrue for
  any route binding an explicit `audit_op_class` — the same misconception
  that produced finding B2. Pre-existing and outside this PR's blast
  radius; worth a follow-up so the next author does not inherit it.
- `_process_transition` crosses the 60-line code-quality *warning*
  threshold (62) with the third claim consumer. Left alone on simplicity
  grounds — it blocks at 100, and splitting a three-consumer fan-out loop
  to reclaim two lines is churn for its own sake.

---

Addressed the two blockers and the minor from review iteration 1.
**Prose only — no executable line under `src/` changed**; both blockers
were operator-facing contract accuracy, not implementation defects.
Verified with the targeted suites (`test_checks_broadcast.py`,
`test_broadcast_events.py`, `test_mcp_tool_broadcast_recent.py`, plus the
adjacent broadcast / checks / UI suites — 422 passed, 10 skipped) and the
full lint gates.

- **B1** `db2523a2` — `checks_transition_broadcast_failed` could never
  fire on the Valkey outage the CHANGELOG, the new doc, and the AC-3 line
  all named it for: `publish_event` swallows every transport exception,
  logs `broadcast_publish_failed`, increments
  `broadcast_publish_errors_total`, and returns. An operator following
  those release notes would have built a permanently silent alert. The
  two halves of the fail-open posture are now described separately in the
  CHANGELOG, `docs/codebase/checks-broadcast.md`, the
  `checks/broadcast.py` module docstring, and this PR body; the test that
  injected `RuntimeError("valkey down")` at the publish seam is renamed
  and reworded to say what it actually stands in for.
- **B2** `db2523a2` — `_CHECK_EVENT_OPS` kept (it is the right
  implementation), but both stated reasons were false and had spread to
  seven places. All three `/api/v1/checks/*` routes bind an explicit
  `audit_op_class`, which `resolve_broadcast_detail` honours over
  `classify_op`, and `meho.audit.query` filters the persisted
  `payload["op_class"]` in SQL — so neither the deploy-boundary split nor
  the result-ingest volume argument was possible, and the CHANGELOG's
  claim that `checks.results.post` "keeps" class `other` was false (it
  ships as `write`). Every rationale is re-anchored to the true hazard —
  `classify_op` is the read-time classifier in the audit and broadcast
  event drawers — across `events.py` (`_CHECK_EVENT_OPS` plus the
  `classify_op` step-4c docstring), the CHANGELOG,
  `docs/codebase/checks-broadcast.md`, `docs/codebase/broadcast.md`, the
  regression test's name and docstring, and this PR body.
  `BroadcastEvent.op_class` now enumerates `approval` and `checks` and
  records the `audit_op_class` override precedence.
- **m1** `db2523a2` — `test_no_claim_publishes_nothing` renamed to
  `test_second_persist_on_an_unchanged_rollup_does_not_republish`: its
  first call wins a real `None → critical` claim and must publish, which
  the old name and docstring denied. Assertions unchanged.

Not taken, matching the reviewer's own assessment and recorded as
adjacent findings above: the `OP_CLASS_BADGE_CLASSES` /
`OP_CLASS_FILTER_OPTIONS` gap, the `aggregate_gate.py:104` pre-existing
misconception, and `asyncio.gather` on the per-claim publishes. The
`_process_transition` line-count justification was corrected — #2721 does
not touch that function, so the restraint stands on simplicity grounds
alone.

Closes #2720

## Summary by CodeRabbit

* **New Features**
  * Dashboard check state changes now broadcast `checks.transition` events for both worsening and recovery transitions.
  * Events include Dashboard identity and previous and new states.
  * Broadcast streams and APIs support filtering by the `checks` operation category.
  * Concurrent processing prevents duplicate transition events.

* **Bug Fixes**
  * Broadcast service failures no longer interrupt state updates, email notifications, or runner persistence.

* **Documentation**
  * Added documentation for check transition events and filtering behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,12 +104,16 @@ connector-related release-notes line.
   uses for `approval.*`. Exactly one event per transition across
   replicas (the existing compare-and-swap claim is the dedupe) and no
   configurable floor: narrowing a feed is the consumer's job.
-  Publishing is fail-open — a Valkey outage logs
-  `checks_transition_broadcast_failed` and leaves the committed state
-  memo, the email notification, and the runner's persist path
-  untouched. The `/api/v1/checks/*` gateway op-ids
-  (`checks.assignment.put` / `.get`, `checks.results.post`) keep their
-  existing `write` / `read` / `other` classes. No migration, no new
+  Publishing is fail-open and leaves the committed state memo, the
+  email notification, and the runner's persist path untouched. A Valkey
+  outage is absorbed by the shared broadcast publisher and stays on its
+  existing feed-wide signals — the `broadcast_publish_failed` warning
+  and the `broadcast_publish_errors_total` counter; alert on those, not
+  on the new `checks_transition_broadcast_failed` warning, which fires
+  only if the event cannot be built before it reaches the publisher.
+  The `/api/v1/checks/*` gateway op-ids (`checks.assignment.put` /
+  `.get`, `checks.results.post`) are untouched and keep the `write` /
+  `read` / `write` classes their routes bind. No migration, no new
   settings. (#2720)
 
 ### Checks notifications — Dashboard email on state transitions (#2719)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,28 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Checks broadcast — transition events on the tenant feed (#2720)
+
+- Every claimed Dashboard rollup edge now publishes one
+  `checks.transition` event to the tenant broadcast feed, in **both**
+  directions — a watcher who saw a Dashboard go `critical` sees it
+  clear. The payload names the Dashboard (id + name) and the
+  `previous → new` states. Events carry a new `checks` op-class, so
+  `meho.broadcast.recent` / `meho.broadcast.watch` with
+  `filter.op_class=checks`, `GET /api/v1/feed?op_class=checks`, and
+  `/ui/broadcast/stream?op_class=checks` all narrow to check state
+  changes server-side — the same shape the approvals console already
+  uses for `approval.*`. Exactly one event per transition across
+  replicas (the existing compare-and-swap claim is the dedupe) and no
+  configurable floor: narrowing a feed is the consumer's job.
+  Publishing is fail-open — a Valkey outage logs
+  `checks_transition_broadcast_failed` and leaves the committed state
+  memo, the email notification, and the runner's persist path
+  untouched. The `/api/v1/checks/*` gateway op-ids
+  (`checks.assignment.put` / `.get`, `checks.results.post`) keep their
+  existing `write` / `read` / `other` classes. No migration, no new
+  settings. (#2720)
+
 ### Checks notifications — Dashboard email on state transitions (#2719)
 
 - A Dashboard can now name an operator to mail when its rollup crosses
--- a/backend/src/meho_backplane/broadcast/events.py
+++ b/backend/src/meho_backplane/broadcast/events.py
@@ -174,6 +174,31 @@ _CREDENTIAL_WRITE_OPS: Final[frozenset[str]] = frozenset(
     }
 )
 
+#: Op-ids that classify as ``checks`` — the checks subsystem's own
+#: state-change events (#2720), published by
+#: :func:`~meho_backplane.checks.broadcast.publish_check_transition_event`
+#: rather than by the audit middleware.
+#:
+#: An explicit allowlist, **not** a ``checks.`` prefix branch, for two
+#: reasons. The ``/api/v1/checks/*`` gateway routes already own three
+#: shipped op-ids under that prefix (``checks.assignment.put`` /
+#: ``checks.assignment.get`` / ``checks.results.post``,
+#: :data:`meho_backplane.api.v1.checks._CHECKS_OP_IDS`), so:
+#:
+#: 1. A prefix branch would reclassify them out of ``write`` / ``read`` /
+#:    ``other``, splitting the historical meaning of ``audit_log.op_class``
+#:    at the deploy boundary and silently dropping the assignment write
+#:    out of any saved ``op_class=write`` query.
+#: 2. ``checks.results.post`` is the runner result-ingest data plane — one
+#:    call per runner per poll cycle, the highest-volume path in the
+#:    subsystem. Folding it into ``checks`` would bury the rare transition
+#:    edges this class exists to surface, defeating the
+#:    ``op_class=checks`` filter before it shipped.
+#:
+#: The allowlist is the same idiom :data:`_CREDENTIAL_MINT_OPS` uses for
+#: the same reason: an exact pin that beats a broader structural match.
+_CHECK_EVENT_OPS: Final[frozenset[str]] = frozenset({"checks.transition"})
+
 #: Op-id suffixes that imply mutation. Append to this tuple when a new
 #: write-shaped verb spelling lands. ``.put`` is the KV-v2 write verb
 #: (``vault.kv.put`` — G3.3-T1 #545); without it a secret write would
@@ -406,6 +431,19 @@ def classify_op(op_id: str) -> str:
        ``decision`` / the request id, never secret material), so it
        stays full-detail like ``other`` did -- see
        :data:`~meho_backplane.broadcast.overrides._SENSITIVE_OP_CLASSES`.
+    4c. ``checks`` — the checks subsystem's own state-change events
+       (:data:`_CHECK_EVENT_OPS`; today just ``checks.transition``,
+       emitted by
+       :func:`~meho_backplane.checks.broadcast.publish_check_transition_event`).
+       Its own class for the same reason ``approval`` has one: a feed
+       watcher filters ``op_class=checks`` server-side to follow
+       Dashboard rollup edges without reading every unclassified op.
+       Matched by **exact membership**, not by ``checks.`` prefix — the
+       ``/api/v1/checks/*`` gateway routes already own op-ids under that
+       prefix and must keep their ``write`` / ``read`` / ``other``
+       classes; see :data:`_CHECK_EVENT_OPS` for the full rationale.
+       Not sensitive (the payload is dashboard id / name / the two
+       states), so it stays full-detail.
     5. ``read`` / ``write`` — HTTP-method-prefixed ingested op IDs
        (e.g. ``GET:/api/v2.0/systeminfo``). ``GET:`` and ``HEAD:``
        map to ``read``; ``POST:``, ``PUT:``, ``PATCH:``, ``DELETE:``
@@ -454,6 +492,10 @@ def classify_op(op_id: str) -> str:
     'approval'
     >>> classify_op("approval.approved")
     'approval'
+    >>> classify_op("checks.transition")
+    'checks'
+    >>> classify_op("checks.assignment.put")
+    'write'
     >>> classify_op("GET:/api/v2.0/systeminfo")
     'read'
     >>> classify_op("DELETE:/api/v2.0/projects/myproj/repositories/myrepo")
@@ -481,6 +523,12 @@ def classify_op(op_id: str) -> str:
     # not sensitive, so it stays full-detail like the `other` fall-through.
     if op_id.startswith("approval."):
         return "approval"
+    # Checks-subsystem state-change events (checks.transition), so a feed
+    # watcher can filter op_class=checks for Dashboard rollup edges (#2720).
+    # Exact membership, not a `checks.` prefix: the /api/v1/checks/* gateway
+    # routes own op-ids under that prefix and keep their existing classes.
+    if op_id in _CHECK_EVENT_OPS:
+        return "checks"
     # Ingested ops use HTTP-method prefixes (e.g. "GET:/api/v2.0/systeminfo");
     # GET/HEAD are safe reads by HTTP semantics. The synthetic net.* probes
     # (#2406) ride the same arm: every net.* op is a non-mutating
--- a/backend/src/meho_backplane/broadcast/events.py
+++ b/backend/src/meho_backplane/broadcast/events.py
@@ -179,21 +179,33 @@ _CREDENTIAL_WRITE_OPS: Final[frozenset[str]] = frozenset(
 #: :func:`~meho_backplane.checks.broadcast.publish_check_transition_event`
 #: rather than by the audit middleware.
 #:
-#: An explicit allowlist, **not** a ``checks.`` prefix branch, for two
-#: reasons. The ``/api/v1/checks/*`` gateway routes already own three
-#: shipped op-ids under that prefix (``checks.assignment.put`` /
+#: An explicit allowlist, **not** a ``checks.`` prefix branch. The
+#: ``/api/v1/checks/*`` gateway routes already own three shipped op-ids
+#: under that prefix (``checks.assignment.put`` /
 #: ``checks.assignment.get`` / ``checks.results.post``,
-#: :data:`meho_backplane.api.v1.checks._CHECKS_OP_IDS`), so:
+#: :data:`meho_backplane.api.v1.checks._CHECKS_OP_IDS`).
 #:
-#: 1. A prefix branch would reclassify them out of ``write`` / ``read`` /
-#:    ``other``, splitting the historical meaning of ``audit_log.op_class``
-#:    at the deploy boundary and silently dropping the assignment write
-#:    out of any saved ``op_class=write`` query.
-#: 2. ``checks.results.post`` is the runner result-ingest data plane — one
-#:    call per runner per poll cycle, the highest-volume path in the
-#:    subsystem. Folding it into ``checks`` would bury the rare transition
-#:    edges this class exists to surface, defeating the
-#:    ``op_class=checks`` filter before it shipped.
+#: What is *not* at risk is what those rows persist. All three routes
+#: bind an explicit ``audit_op_class`` contextvar (``write`` / ``read`` /
+#: ``write``), and
+#: :func:`~meho_backplane.broadcast.overrides.resolve_broadcast_detail`
+#: takes that ``op_class_override`` in preference to this function — the
+#: same value the audit row stores in ``payload["op_class"]``, which is
+#: what ``meho.audit.query`` filters on in SQL. So a prefix branch would
+#: not move a single stored class, and no saved ``op_class=write`` query
+#: would change its results.
+#:
+#: The hazard is on the **read** side. ``classify_op`` is re-run at render
+#: time against the stored op-id by the audit drawer
+#: (:mod:`meho_backplane.ui.routes.audit.routes`) and the broadcast event
+#: drawer (:mod:`meho_backplane.ui.routes.broadcast.event`), where it
+#: supplies the row's displayed class and badge and feeds
+#: :func:`~meho_backplane.ui.routes.broadcast.aggregate_gate.is_aggregate_only`.
+#: A prefix branch would therefore relabel every gateway row as
+#: ``checks`` — retroactively, because that class is derived on read
+#: rather than read back off the row — and would sweep in any future
+#: non-transition ``checks.*`` op-id by construction, diluting the class
+#: whose whole purpose is surfacing rare transition edges.
 #:
 #: The allowlist is the same idiom :data:`_CREDENTIAL_MINT_OPS` uses for
 #: the same reason: an exact pin that beats a broader structural match.
@@ -344,7 +356,13 @@ class BroadcastEvent(BaseModel):
     op_id: str
     #: One of ``"read"`` / ``"write"`` / ``"credential_read"`` /
     #: ``"credential_mint"`` / ``"credential_write"`` / ``"audit_query"``
-    #: / ``"other"``. Derived from :func:`classify_op` at publish time.
+    #: / ``"approval"`` / ``"checks"`` / ``"other"``. Usually derived
+    #: from :func:`classify_op` at publish time — but a route that binds
+    #: the ``audit_op_class`` contextvar overrides it:
+    #: :func:`~meho_backplane.broadcast.overrides.resolve_broadcast_detail`
+    #: takes ``op_class_override`` in preference to :func:`classify_op`,
+    #: and the same value is what the audit row persists under
+    #: ``payload["op_class"]``.
     op_class: str
     #: One of ``"ok"`` / ``"error"`` / ``"denied"``. The handler
     #: produces it; the broadcast publisher does not re-classify.
@@ -440,10 +458,11 @@ def classify_op(op_id: str) -> str:
        Dashboard rollup edges without reading every unclassified op.
        Matched by **exact membership**, not by ``checks.`` prefix — the
        ``/api/v1/checks/*`` gateway routes already own op-ids under that
-       prefix and must keep their ``write`` / ``read`` / ``other``
-       classes; see :data:`_CHECK_EVENT_OPS` for the full rationale.
-       Not sensitive (the payload is dashboard id / name / the two
-       states), so it stays full-detail.
+       prefix, and because this function is re-run at render time on the
+       stored op-id, a prefix branch would relabel their rows (historical
+       ones included) in the drawers; see :data:`_CHECK_EVENT_OPS` for
+       the full rationale. Not sensitive (the payload is dashboard id /
+       name / the two states), so it stays full-detail.
     5. ``read`` / ``write`` — HTTP-method-prefixed ingested op IDs
        (e.g. ``GET:/api/v2.0/systeminfo``). ``GET:`` and ``HEAD:``
        map to ``read``; ``POST:``, ``PUT:``, ``PATCH:``, ``DELETE:``
--- a/backend/tests/test_checks_broadcast.py
+++ b/backend/tests/test_checks_broadcast.py
@@ -304,8 +304,15 @@ async def test_audit_id_is_the_nil_uuid(monkeypatch: pytest.MonkeyPatch) -> None
 async def test_publisher_failure_is_swallowed_and_logged(
     monkeypatch: pytest.MonkeyPatch,
 ) -> None:
-    """A raising publisher never reaches the caller; it logs and returns."""
-    _install_publisher(monkeypatch, exc=RuntimeError("valkey down"))
+    """A raising publish never reaches the caller; it logs and returns.
+
+    The injected failure stands in for what the guard actually covers: a
+    fault *before* the ``XADD``, e.g. lineage resolution or
+    ``BroadcastEvent`` validation. A Valkey outage never gets this far --
+    :func:`~meho_backplane.broadcast.publisher.publish_event` swallows it
+    and logs ``broadcast_publish_failed`` on its own.
+    """
+    _install_publisher(monkeypatch, exc=RuntimeError("event construction failed"))
 
     with capture_logs() as logs:
         await publish_check_transition_event(
@@ -368,16 +375,20 @@ async def test_concurrent_transitions_publish_exactly_one_event(
 
 
 @pytest.mark.asyncio
-async def test_publisher_outage_leaves_the_memo_and_the_mail_intact(
+async def test_broadcast_failure_leaves_the_memo_and_the_mail_intact(
     monkeypatch: pytest.MonkeyPatch,
 ) -> None:
-    """A broadcast outage must not fail the persist seam or the notifier.
+    """A failing broadcast must not fail the persist seam or the notifier.
 
     The committed ``last_rollup_state`` memo is the durable truth; the
     feed is the at-most-once real-time view. This is the fail-open
-    contract the whole publish path is built on.
+    contract the whole publish path is built on. Raising at the publish
+    seam is the strongest form of the failure -- a real Valkey outage is
+    weaker still, since
+    :func:`~meho_backplane.broadcast.publisher.publish_event` swallows it
+    before this module's guard ever sees it.
     """
-    _install_publisher(monkeypatch, exc=RuntimeError("valkey down"))
+    _install_publisher(monkeypatch, exc=RuntimeError("broadcast unavailable"))
     mail = _install_transport(monkeypatch)
     monkeypatch.setattr(inv, "_schedule_investigation", lambda **_: None)
     await _seed_tenant()
@@ -396,13 +407,18 @@ async def test_publisher_outage_leaves_the_memo_and_the_mail_intact(
 
 
 @pytest.mark.asyncio
-async def test_no_claim_publishes_nothing(monkeypatch: pytest.MonkeyPatch) -> None:
-    """A persist that moves no rollup state publishes no event.
+async def test_second_persist_on_an_unchanged_rollup_does_not_republish(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Only the persist that moves the rollup state publishes.
 
-    The Dashboard is already ``critical`` in the memo, so the
-    compare-and-swap finds nothing to claim. Guards against the publish
-    being wired outside the claim-win branch, which would emit one event
-    per Sensor evaluation instead of one per edge.
+    The Dashboard starts with a NULL memo, so the first
+    ``investigate_on_transition`` wins a real ``ok -> critical`` claim and
+    publishes once. The second call finds the memo already ``critical``,
+    the compare-and-swap claims nothing, and the count must stay at one.
+    Guards against the publish being wired outside the claim-win branch,
+    which would emit one event per Sensor evaluation instead of one per
+    edge.
     """
     fake = _install_publisher(monkeypatch)
     _install_transport(monkeypatch)
@@ -412,7 +428,7 @@ async def test_no_claim_publishes_nothing(monkeypatch: pytest.MonkeyPatch) -> No
     await _seed_dashboard(name="prod-health", sensor_ids=[sid])
 
     await inv.investigate_on_transition(sensor_id=sid, tenant_id=_TENANT)
-    assert len(fake.events) == 1
+    assert len(fake.events) == 1, "the first persist claims the edge and publishes it"
 
     await inv.investigate_on_transition(sensor_id=sid, tenant_id=_TENANT)
     assert len(fake.events) == 1, "a no-op persist must not re-publish the edge"
--- a/backend/tests/test_checks_broadcast.py
+++ b/backend/tests/test_checks_broadcast.py
@@ -1,0 +1,418 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""Tests for the ``checks.transition`` broadcast publisher (#2720).
+
+Initiative #2716 (parent goal #221), Task #2720. Coverage:
+
+* **Event shape** -- op-id, op-class, tenant, principal, and the four
+  payload fields; plus the pin that the op-id the publisher emits is the
+  one :func:`~meho_backplane.broadcast.events.classify_op` recognises.
+* **Fail-open** -- a raising publisher never reaches the caller, and
+  emits the structured warning operators alert on.
+* **Integration through the persist seam** -- two raced
+  ``investigate_on_transition`` calls on one Dashboard publish exactly one
+  event (the compare-and-swap claim is the dedupe, same proof shape as
+  #2719's email test), the recovery edge back to ``ok`` publishes a
+  second, and a publisher outage leaves the memo committed and the mail
+  sent.
+
+:func:`~meho_backplane.broadcast.publisher.publish_event` is replaced by a
+recording fake in every test, so no Valkey connection is opened; the DB
+layer is real against the SQLite engine from :mod:`tests.conftest`.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from collections.abc import Iterator
+from datetime import UTC, datetime, timedelta
+from typing import Any
+from uuid import UUID, uuid4
+
+import pytest
+from structlog.testing import capture_logs
+
+import meho_backplane.checks.broadcast as checks_broadcast
+import meho_backplane.checks.investigate as inv
+import meho_backplane.checks.notify as notify
+from meho_backplane.broadcast.events import BroadcastEvent, classify_op
+from meho_backplane.broadcast.history import OP_CLASS_ENUM
+from meho_backplane.checks.broadcast import (
+    CHECK_TRANSITION_OP_ID,
+    publish_check_transition_event,
+)
+from meho_backplane.connectors.mail.transport import MailSendResult
+from meho_backplane.db.engine import get_sessionmaker
+from meho_backplane.db.models import (
+    CheckDashboard,
+    CheckDashboardSensor,
+    Sensor,
+    Tenant,
+)
+from meho_backplane.settings import get_settings
+
+_TENANT = UUID("cccccccc-cccc-cccc-cccc-cccccccccccc")
+_DASHBOARD = UUID("11111111-1111-1111-1111-111111111111")
+
+_ASSERTION: dict[str, Any] = {
+    "select": {"path": "$.count"},
+    "compare": {"type": "threshold", "op": "lt", "critical": 10},
+}
+
+
+# ---------------------------------------------------------------------------
+# Fixtures + doubles
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(autouse=True)
+def _settings_env(monkeypatch: pytest.MonkeyPatch) -> Iterator[None]:
+    """Pin the env :class:`Settings` requires so ``get_settings()`` constructs."""
+    monkeypatch.setenv("KEYCLOAK_ISSUER_URL", "https://kc.test/realms/meho")
+    monkeypatch.setenv("KEYCLOAK_AUDIENCE", "meho")
+    monkeypatch.setenv("VAULT_ADDR", "https://vault.test")
+    monkeypatch.setenv("SCHEDULER_ENABLED", "false")
+    monkeypatch.setenv("SENSOR_RUNNER_ENABLED", "false")
+    get_settings.cache_clear()
+    yield
+    get_settings.cache_clear()
+
+
+@pytest.fixture(autouse=True)
+def _reset_background_tasks() -> Iterator[None]:
+    """Clear the in-flight notification / investigation sets per test."""
+    yield
+    notify._NOTIFICATIONS.clear()
+    inv._INVESTIGATIONS.clear()
+
+
+class _RecordingPublisher:
+    """Stands in for ``publish_event``; records every event, optionally raises."""
+
+    def __init__(self, *, exc: Exception | None = None) -> None:
+        self.events: list[BroadcastEvent] = []
+        self._exc = exc
+
+    async def __call__(self, event: BroadcastEvent) -> None:
+        self.events.append(event)
+        if self._exc is not None:
+            raise self._exc
+
+
+def _install_publisher(
+    monkeypatch: pytest.MonkeyPatch, *, exc: Exception | None = None
+) -> _RecordingPublisher:
+    """Patch the publisher the checks module imported (not the source module)."""
+    fake = _RecordingPublisher(exc=exc)
+    monkeypatch.setattr(checks_broadcast, "publish_event", fake)
+    return fake
+
+
+class _RecordingTransport:
+    """Stands in for ``send_email`` so the notifier opens no SMTP session."""
+
+    def __init__(self) -> None:
+        self.calls: list[dict[str, Any]] = []
+
+    async def __call__(self, *, to: list[str], subject: str, body: str) -> MailSendResult:
+        self.calls.append({"to": list(to), "subject": subject, "body": body})
+        return MailSendResult(sent=True)
+
+
+def _install_transport(monkeypatch: pytest.MonkeyPatch) -> _RecordingTransport:
+    fake = _RecordingTransport()
+    monkeypatch.setattr(notify, "send_email", fake)
+    return fake
+
+
+# ---------------------------------------------------------------------------
+# DB seeds (mirror tests/test_checks_notify.py)
+# ---------------------------------------------------------------------------
+
+
+async def _seed_tenant(tenant_id: UUID = _TENANT) -> None:
+    sessionmaker = get_sessionmaker()
+    async with sessionmaker() as session:
+        if await session.get(Tenant, tenant_id) is None:
+            session.add(Tenant(id=tenant_id, slug="tenant-bcast", name="Tenant Broadcast"))
+            await session.commit()
+
+
+async def _seed_sensor(*, name: str, last_state: str = "critical") -> UUID:
+    now = datetime.now(UTC)
+    sensor_id = uuid4()
+    sessionmaker = get_sessionmaker()
+    async with sessionmaker() as session:
+        session.add(
+            Sensor(
+                id=sensor_id,
+                tenant_id=_TENANT,
+                name=name,
+                connector_id="vmware-rest-9.0",
+                op_id="vmware.vm.list",
+                target=None,
+                params={},
+                assertion=_ASSERTION,
+                status="active",
+                cadence_kind="interval",
+                interval_seconds=60,
+                cron_expr=None,
+                timezone="UTC",
+                next_fire_at=now + timedelta(seconds=60),
+                severity="critical",
+                for_seconds=0,
+                last_state=last_state,
+                last_value=7,
+                last_evidence={"observed": 7, "bound": 10},
+                last_evaluated_at=now - timedelta(seconds=30),
+                state_since=now - timedelta(hours=1),
+                identity_sub="__sensor__",
+                created_by_sub="op-admin",
+            )
+        )
+        await session.commit()
+    return sensor_id
+
+
+async def _seed_dashboard(
+    *,
+    name: str,
+    sensor_ids: list[UUID],
+    dashboard_id: UUID = _DASHBOARD,
+    notify_email: str | None = "oncall@example.com",
+) -> UUID:
+    sessionmaker = get_sessionmaker()
+    async with sessionmaker() as session:
+        session.add(
+            CheckDashboard(
+                id=dashboard_id,
+                tenant_id=_TENANT,
+                name=name,
+                description=None,
+                last_rollup_state=None,
+                notify_email=notify_email,
+                notify_min_state="critical",
+                created_by_sub="op-admin",
+            )
+        )
+        for sid in sensor_ids:
+            session.add(CheckDashboardSensor(dashboard_id=dashboard_id, sensor_id=sid))
+        await session.commit()
+    return dashboard_id
+
+
+async def _set_sensor_state(sensor_id: UUID, state: str) -> None:
+    sessionmaker = get_sessionmaker()
+    async with sessionmaker() as session:
+        row = await session.get(Sensor, sensor_id)
+        assert row is not None
+        row.last_state = state
+        row.state_since = datetime.now(UTC) - timedelta(hours=1)
+        row.last_evaluated_at = datetime.now(UTC)
+        await session.commit()
+
+
+# ---------------------------------------------------------------------------
+# Event shape
+# ---------------------------------------------------------------------------
+
+
+def test_op_id_is_the_one_classify_op_recognises() -> None:
+    """The publisher's op-id and the classifier's allowlist cannot drift.
+
+    The two live in different layers on purpose --
+    :mod:`meho_backplane.broadcast.events` must not import from
+    :mod:`meho_backplane.checks` -- so the literal is written twice. This
+    pins them together, the same way the op-class enum is pinned to the
+    tool schemas.
+    """
+    assert classify_op(CHECK_TRANSITION_OP_ID) == "checks"
+
+
+def test_checks_is_filterable_from_the_agent_surface() -> None:
+    """``checks`` is in the op_class filter vocabulary the MCP tools advertise.
+
+    ``inputSchema`` validation is jsonschema-enforced at the dispatcher,
+    so a class absent from :data:`OP_CLASS_ENUM` is a ``-32602`` before
+    the handler runs -- the class would be published but unreachable from
+    ``meho.broadcast.recent`` / ``.watch``.
+    """
+    assert "checks" in OP_CLASS_ENUM
+
+
+@pytest.mark.asyncio
+async def test_event_carries_the_edge_and_the_dashboard(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """One publish per call, with the documented op-id / class / payload."""
+    fake = _install_publisher(monkeypatch)
+
+    await publish_check_transition_event(
+        tenant_id=_TENANT,
+        dashboard_id=_DASHBOARD,
+        dashboard_name="prod-health",
+        previous_state="ok",
+        new_state="critical",
+    )
+
+    assert len(fake.events) == 1
+    event = fake.events[0]
+    assert event.kind == "operation"
+    assert event.op_id == "checks.transition"
+    assert event.op_class == "checks"
+    assert event.tenant_id == _TENANT
+    assert event.principal_sub == "__checks__"
+    assert event.result_status == "ok"
+    assert event.payload == {
+        "op_class": "checks",
+        "result_status": "ok",
+        "dashboard_id": str(_DASHBOARD),
+        "dashboard_name": "prod-health",
+        "previous_state": "ok",
+        "new_state": "critical",
+    }
+
+
+@pytest.mark.asyncio
+async def test_audit_id_is_the_nil_uuid(monkeypatch: pytest.MonkeyPatch) -> None:
+    """No audit row backs a rollup edge, so the FK is the nil sentinel.
+
+    A fabricated id would be indistinguishable from a stale one to the UI
+    drawer, which resolves ``/ui/broadcast/event/{audit_id}`` against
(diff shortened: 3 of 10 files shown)


damir-topic (commented):




**Anchor task:** #2720 — `checks.transition` broadcast events + a `checks` op-class on the tenant feed
**Surface area:** backend
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `dc0c96ea`

> Posted in COMMENT state because GitHub 422s a formal APPROVE/REQUEST_CHANGES from the PR author's own identity (this automation shares the author's account, as on #2724 / #2726 / #2728). The **Verdict** line above is binding.

### Adjudicating the Phase 4 pushback

Both corrections were verified independently against the code, not taken on the author's word.

**Claim 2 — `broadcast_recent` did NOT pick the class up for free: UPHELD, in full.**

Verified end to end. `mcp/tools/broadcast.py:308` (and `:1236` for `.watch`) advertise `"enum": list(OP_CLASS_ENUM)` on `filter.op_class`; `mcp/handlers.py:336-346` runs `jsonschema.validate(..., cls=Draft202012Validator)` over every `tools/call` before the handler is reached. I proved the consequence empirically rather than by reading: removing `"checks"` from `OP_CLASS_ENUM` in a scratch worktree makes `test_filter_op_class_checks_reaches_the_handler` **fail** (restored immediately after; `git status` clean). The task's Desired-state item 4 was wrong, the enum change was load-bearing, and posting a full JSON-RPC `tools/call` rather than calling the handler is exactly the right test shape — a handler-level call would have passed with the class missing. Nothing to change here.

**Claim 1 — exact-membership allowlist instead of a `checks.` prefix branch: right conclusion, wrong reasoning.**

The *decision* stands — keep `_CHECK_EVENT_OPS`. But both stated reasons are factually incorrect, and they are now baked into a 20-line docstring, the user-facing CHANGELOG, a new permanent `docs/codebase/` page, and a test docstring. Verified by executing the real path:

```
persisted payload: {'op_class': 'write', 'op_id': 'checks.results.post'}
op_id/override   : ('checks.results.post', 'write')
classify_op says : other
```

All three `/api/v1/checks/*` routes bind an **explicit `audit_op_class`** alongside `audit_op_id` (`api/v1/checks.py:124-126`, `:156-158`, `:186-188` → `"write"` / `"read"` / `"write"`). `_resolve_audit_payload` lifts that into `payload["op_class"]`, and `broadcast/overrides.py:357` reads `op_class = op_class_override if op_class_override else classify_op(op_id)`. So for these three op-ids **`classify_op` is never consulted on the publish or audit-write path**, and `meho.audit.query` filters the persisted `payload["op_class"]` in SQL (`audit_query/query.py:240-242`) — never `classify_op`. A `checks.` prefix branch therefore could not have moved an audit query or a broadcast `op_class` for those routes at all. Consequences: reason 1 (a saved `op_class=write` query stops seeing the assignment write) cannot happen; reason 2 (folding the high-volume result-ingest plane into `checks`) cannot happen; and `checks.results.post` ships as **`write`**, not the `other` that the CHANGELOG, the doc, and the regression test all assert it "keeps". See 🛑 B2.

The allowlist is still correct, for a narrower reason worth writing down instead: `classify_op` *is* the read-time classifier at `ui/routes/audit/routes.py:439` and `ui/routes/broadcast/event.py:141`, both feeding `is_aggregate_only`. A prefix branch would relabel every historical **and** future `checks.assignment.*` / `checks.results.post` drawer row as `checks` (retroactively — it is derived at read time, so not a deploy-boundary split either), and would silently sweep in any future `checks.*` op-id that is not a transition event.

**Is the regression test adequate?** Mechanically yes — `test_gateway_checks_op_ids_keep_their_classes` fails if anyone adds a prefix branch, which is its job. But it pins `classify_op` in isolation while its docstring claims to guard `audit_log.op_class`, which `classify_op` does not determine for these rows, and it enshrines `checks.results.post → "other"` as though that were the shipped class. Fix the docstring with B2.

### The three self-flagged judgement calls

**`audit_id` = nil UUID — accepted.** I walked every consumer, not just the one named. `/ui/broadcast/event/{audit_id}` parses the nil UUID fine, `fetch_audit_row` returns `None`, and the route renders `_event_drawer_not_found.html` with a 404 (`ui/routes/broadcast/event.py:139-147`) — degrades, does not break. `broadcast-feed.js:215`'s `if (!ev || !ev.audit_id)` guard does not catch the nil UUID (truthy string), so a click opens that 404 drawer; acceptable. The one I chased hardest was `connectors/_recent_ops.html:93`, which uses `ev.audit_id` as an Alpine `x-for` `:key` — a shared nil UUID across several events would be a duplicate-key collision. It cannot fire: that card subscribes to `/ui/broadcast/stream?target=<name>` and a transition event carries no target, so it is filtered server-side before reaching the key. No finding.

**`"__checks__"` principal sentinel — accepted, and the reasoning is right.** `"__sensor__"` is the *default* of an operator-overridable per-Sensor column (`checks/schemas.py:113`, `db/models.py:6261`), not a subsystem identity, so reusing it would have been wrong. A distinct `__x__` sentinel matches the existing convention.

**Inline await instead of backgrounding — accepted, verified rather than assumed.** The fast broadcast client really does pin both timeouts (`broadcast/client.py:150-151` → `socket_connect_timeout=3.0`, `socket_timeout=5.0`), so a Valkey stall is bounded at ~5 s, and `publish_event` swallows every exception (`publisher.py:228-236`) so it cannot raise into the persist seam. It also runs *after* `session.commit()` inside `_record_result`, which is itself a backgrounded per-sensor evaluation task (`checks/runner.py:433-444`) — so the await adds bounded latency to a background task, not to a synchronous request. The judgement is sound.

**Leaving `_process_transition` at 62 lines — right call, slightly overstated reason.** The threshold really is warn-at-60 / block-at-100, so this is a warning only, and splitting a three-consumer fan-out loop to reclaim 2 lines is churn CLAUDE.md rule 3 exists to prevent. One correction: #2721's hunks in `investigate.py` land at old lines 49 / 122 / 219 / 284 / 606 / 640 / 969 / 987 / 1122 / 1149 — **none inside `_process_transition`** — so a refactor there would not in fact have collided with #2721. The restraint is right on simplicity grounds; the stated conflict-avoidance grounds do not hold.

### Conflict risk with the concurrent #2721

Low and routine. Overlap is exactly two files. `investigate.py`: #2721 does not touch `_process_transition`, so the only contact point is the import block — #2729 adds `checks.broadcast` at old line 119-120, #2721 extends the `checks.notify` import at old 122-129. Adjacent context windows, so possibly a trivial textual conflict, resolvable mechanically. `CHANGELOG.md`: both append an `[Unreleased]` section — near-certain conflict, entirely routine here. No shared symbols, no shared tests, and #2721's migration `0069` is untouched by this PR.

### Acceptance criteria

- [~] **AC-1** `grep -n 'startswith("checks.")'` hits — **deliberately not met**; substituted with `_CHECK_EVENT_OPS` exact membership. Adjudicated **correct as implementation**, but the substitution's written rationale is wrong (🛑 B2). `classify_op("checks.transition") == "checks"` verified by `test_checks_transition_maps_to_checks` + doctest.
- [x] **AC-2** exactly one event per claimed edge, both directions — `test_concurrent_transitions_publish_exactly_one_event` passes: two raced `investigate_on_transition` calls publish once, recovery `critical→ok` publishes a second. `test_no_claim_publishes_nothing` guards the publish staying on the claim-win branch.
- [~] **AC-3** fail-open — behaviour verified (`test_publisher_outage_leaves_the_memo_and_the_mail_intact`: memo committed, mail sent). The **"structured warning logged"** half is misstated: the named warning cannot fire on the outage the docs name it for (🛑 B1).
- [x] **AC-4** `broadcast_recent` filtered to `op_class=checks` returns the event — verified over a full `tools/call`, and proven load-bearing by the enum-removal experiment above.
- [x] **AC-5** no migration, no new settings (`git diff --name-only main...HEAD` matches nothing under `alembic|settings|migration`); CHANGELOG bullet present; suite green.

### Project-board alignment

|---|---|---|---|---|---|
| #2720 | open, `task` | `Closes` | ✅ | ⚠️ partial — AC-1 substituted, AC-3 doc claim wrong | close on merge after B1/B2 |
| #2716 | open (parent initiative) | — | — | — | scope item 3 of 5; leave open |

`closingIssuesReferences` resolves to exactly #2720. No delivered-but-unlinked work; no linked-but-undelivered work.

### Predicted vs. actual diff

- ✅ `broadcast/events.py` (classifier arm), `broadcast/history.py` (enum), `checks/broadcast.py` (new publisher), `checks/investigate.py` (claim-win call site), the three test files
- ⚠️ `CHANGELOG.md`, `docs/codebase/checks-broadcast.md`, `broadcast/events.py` docstring — expected files, wrong content (B1, B2)
- ❓ `docs/codebase/broadcast.md` also gains an `approval` bullet that was missing before — undisclosed in the PR body, but it is a one-item gap-fill in the very list the PR had to edit; justified
- 🚫 none
- 🕳️ none

### Integration verification

- [x] `ruff format --check src/ tests/` — 1382 files already formatted
- [x] `mypy src/` — 1 error, the known darwin-only `MSG_ERRQUEUE` in `connectors/net/icmp.py`; CI runs Linux
- [x] `pytest tests/test_checks_broadcast.py tests/test_broadcast_events.py tests/test_mcp_tool_broadcast_recent.py` — 134 passed, 4 skipped
- [x] full backend suite — green (see run note in the result file)
- [x] **enum gate exercised for real**: `checks` removed from `OP_CLASS_ENUM` → AC-4 test fails → restored, worktree clean
- [x] **audit/broadcast classification path executed**, not reasoned about: contextvar → `_resolve_audit_payload` → `_resolve_op_id_and_class_override` run live to establish B2
- [⚠️ not run] migrations replay — no migration in this PR
- [⚠️ not run] `docker compose up`, UI flow, agent trace — no UI or agent-path change

### Findings (each is a merge blocker per team policy)

- 🛑 **B1** — `checks_transition_broadcast_failed` cannot fire on a Valkey outage, but the CHANGELOG, the new doc, and the PR test plan all tell operators it does. Required change: reword all three to say a Valkey/`XADD` failure logs `broadcast_publish_failed` (from `publish_event`) and that `checks_transition_broadcast_failed` covers only pre-publish construction failures.
- 🛑 **B2** — the `_CHECK_EVENT_OPS` rationale is factually wrong in four places, and `checks.results.post` is documented as class `other` when it ships as `write`. Required change: replace the rationale with the read-time-`classify_op` one and fix the class claim.
- 💡 **m1** — `test_no_claim_publishes_nothing` asserts the *second* (no-op) persist does not re-publish; the first call does publish. Required change: rename / redescribe to match.

Copilot independently raised B1 (on `CHANGELOG.md` and the doc) and m1; those three threads are still open and are resolved by the same fixes.

- The control-flow fence in `docs/codebase/checks-broadcast.md:35` has no language tag (markdownlint MD040, raised by CodeRabbit). No markdownlint runs in CI, so this gates nothing.
- Per-claim publishes are sequential; if one Sensor ever fans out to many Dashboards that is N × ≤5 s in a background task. `asyncio.gather` would flatten it. Genuinely low value at current fan-out — recorded, not required.
- `OP_CLASS_BADGE_CLASSES` / `OP_CLASS_FILTER_OPTIONS` (`ui/routes/broadcast/feed.py:104,121`) omit `approval`, `credential_write`, and now `checks` — the author already flagged this; one follow-up should fix all three.
- Pre-existing and outside this PR: `ui/routes/broadcast/aggregate_gate.py:104` claims `classify_op` "yields the same class the middleware builds", which is untrue for any route binding an explicit `audit_op_class` — the same misconception that produced B2. Worth a follow-up so the next author does not inherit it.

### Docs

- [x] `docs/codebase/checks-broadcast.md` — new, thorough, well-structured; **but** its "The `checks` op-class" and "Failure posture" sections carry B2 and B1 respectively
- [x] `docs/codebase/broadcast.md` — updated correctly
- No central `docs/codebase/` index exists, so nothing else needed updating

### What worked well

- The exactly-once proof is the right shape: rather than asserting the CAS works, `test_concurrent_transitions_publish_exactly_one_event` actually races two `investigate_on_transition` calls through the real persist seam and counts events, then drives the recovery edge in the same test. `test_no_claim_publishes_nothing` is the complementary guard that the publish sits *inside* the claim-win branch — that is the failure mode that would have shipped one event per Sensor evaluation.
- Posting AC-4 as a full JSON-RPC `tools/call` instead of a direct handler call is a deliberately harder test than the task asked for, and it is the only reason the `OP_CLASS_ENUM` gap was found at all.
- Pushing back on two acceptance criteria rather than implementing them literally was the right instinct, and the enum finding is a genuinely good catch that the task author got wrong.

CHANGELOG.md
```diff
@@ -90,6 +90,28 @@ connector-related release-notes line.
...
+  `previous → new` states. Events carry a new `checks` op-class, so
+  `meho.broadcast.recent` / `meho.broadcast.watch` with
+  `filter.op_class=checks`, `GET /api/v1/feed?op_class=checks`, and
+  `/ui/broadcast/stream?op_class=checks` all narrow to check state
+  changes server-side — the same shape the approvals console already
+  uses for `approval.*`. Exactly one event per transition across
+  replicas (the existing compare-and-swap claim is the dedupe) and no
+  configurable floor: narrowing a feed is the consumer's job.
+  Publishing is fail-open — a Valkey outage logs
+  `checks_transition_broadcast_failed` and leaves the committed state
```
damir-topic: **[🛑 B1]** — `checks_transition_broadcast_failed` cannot fire on a Valkey outage, which is the one case the release notes name it for

**What:** This bullet tells operators that "a Valkey outage logs `checks_transition_broadcast_failed`". It does not. `publish_event` catches **every** exception from `get_broadcast_client()` and `client.xadd(...)`, logs its own `broadcast_publish_failed`, increments `BROADCAST_PUBLISH_ERRORS_TOTAL`, and **returns normally** (`broadcast/publisher.py:220-236`). Because it never re-raises, the `except Exception` guard in `publish_check_transition_event` is unreachable for any transport failure — it can only catch a failure *before* the publish, i.e. in `resolve_broadcast_lineage()`, `classify_op()`, or `BroadcastEvent(...)` construction.

The same claim appears in `docs/codebase/checks-broadcast.md:111-113` ("A swallowed failure logs `checks_transition_broadcast_failed` at warning with the dashboard id and both states") and in the PR body's AC-3 line. The module docstring at `checks/broadcast.py` gets this right — it correctly separates the two guards — so only the downstream summaries drifted.

**Why it matters:** This is an operator alerting contract, published in user-facing release notes. An operator who reads this bullet and builds an alert on `checks_transition_broadcast_failed` gets a rule that stays silent through exactly the Valkey outage it was written to detect, while the real signal (`broadcast_publish_failed`, plus the `BROADCAST_PUBLISH_ERRORS_TOTAL` counter) goes unwatched. It also makes AC-3 read as proven when the test that backs it (`test_publisher_failure_is_swallowed_and_logged`) injects a raising `publish_event` — a fake that cannot occur in production.

**Required change:** Reword all three places to describe the two-layer posture accurately. In this CHANGELOG bullet, replace "a Valkey outage logs `checks_transition_broadcast_failed`" with something like: "a Valkey outage is swallowed by `publish_event` (logged as `broadcast_publish_failed`, counted in `BROADCAST_PUBLISH_ERRORS_TOTAL`), and the checks-specific `checks_transition_broadcast_failed` guard covers pre-publish event-construction failures". Apply the same correction to `docs/codebase/checks-broadcast.md:111-113` under "Failure posture" and to the AC-3 line in the PR description. No code change is needed — the behaviour is already correct and fail-open; only the description of what operators will observe is wrong.

**Source:** `backend/src/meho_backplane/broadcast/publisher.py:220-236` (the swallow-and-return). Same finding as the open Copilot threads on this line and on `docs/codebase/checks-broadcast.md:113`.

backend/src/meho_backplane/broadcast/events.py
```diff
@@ -174,6 +174,31 @@
...
+#: An explicit allowlist, **not** a ``checks.`` prefix branch, for two
+#: reasons. The ``/api/v1/checks/*`` gateway routes already own three
+#: shipped op-ids under that prefix (``checks.assignment.put`` /
+#: ``checks.assignment.get`` / ``checks.results.post``,
+#: :data:`meho_backplane.api.v1.checks._CHECKS_OP_IDS`), so:
+#:
+#: 1. A prefix branch would reclassify them out of ``write`` / ``read`` /
+#:    ``other``, splitting the historical meaning of ``audit_log.op_class``
+#:    at the deploy boundary and silently dropping the assignment write
+#:    out of any saved ``op_class=write`` query.
```
damir-topic: **[🛑 B2]** — both stated reasons for the allowlist are factually wrong, and `checks.results.post` does not have the class this documents

**What:** Keeping `_CHECK_EVENT_OPS` is the right call — I am not asking you to switch to a prefix branch. But neither reason given here holds, because all three `/api/v1/checks/*` routes bind an **explicit `audit_op_class`** next to `audit_op_id`:

- `api/v1/checks.py:124-126` → `audit_op_class="write"`
- `api/v1/checks.py:156-158` → `audit_op_class="read"`
- `api/v1/checks.py:186-188` → `audit_op_class="write"`

`_resolve_audit_payload` lifts that into `payload["op_class"]`, and `broadcast/overrides.py:357` is `op_class = op_class_override if op_class_override else classify_op(op_id)`. So `classify_op` is **never consulted** for these three op-ids on the publish or audit-write path. Executed live to confirm:

```
persisted payload: {'op_class': 'write', 'op_id': 'checks.results.post'}
op_id/override   : ('checks.results.post', 'write')
classify_op says : other
```

Therefore reason 1 is impossible: `meho.audit.query` filters the *persisted* `payload["op_class"]` in SQL (`audit_query/query.py:240-242`), never `classify_op`, so a prefix branch could not have dropped the assignment write out of a saved `op_class=write` query. Reason 2 is impossible for the same reason: `checks.results.post` broadcasts as `write` from its override, so a prefix branch could not have folded the result-ingest data plane into `checks`.

It also makes a claim about shipped behaviour that is wrong. `checks.results.post`'s effective class is **`write`**, not `other` — `other` is only what `classify_op` returns in isolation, a value that never reaches production for that route. That wrong triple is repeated in `CHANGELOG.md:110-112` ("keep their existing `write` / `read` / `other` classes"), in `docs/codebase/checks-broadcast.md:56-70`, and in `test_gateway_checks_op_ids_keep_their_classes` (`tests/test_broadcast_events.py:324`), whose docstring restates both false reasons.

**Why it matters:** This is the load-bearing "why" for a deliberate deviation from the task's AC-1, written into a permanent module docstring, a new permanent `docs/codebase/` page, user-facing release notes, and a regression test's stated contract. The next person to touch `classify_op` will reason from a model of the audit pipeline that is wrong in both directions — believing `classify_op` drives `audit_log.op_class` (it does not, wherever a route binds an override) and believing `checks.results.post` is class `other` (it is not). The CHANGELOG entry additionally tells readers something false about behaviour that shipped before this PR.

**Required change:** Keep `_CHECK_EVENT_OPS` and the test; replace the rationale with the accurate one and fix the class claim.

1. Rewrite reasons 1–2 in this docstring to the real hazard: `classify_op` is the **read-time** classifier at `ui/routes/audit/routes.py:439` and `ui/routes/broadcast/event.py:141`, both of which feed `is_aggregate_only`. A `checks.` prefix branch would relabel every `checks.assignment.*` / `checks.results.post` drawer row as `checks` — retroactively for historical rows too, since it is derived at read time, not a deploy-boundary split — and would silently sweep in any future `checks.*` op-id that is not a transition event. That is a sufficient and true reason for exact membership.
2. Fix `CHANGELOG.md:110-112`: the gateway op-ids keep their route-bound `write` / `read` / `write` classes (or drop the per-op-id enumeration and just say the gateway routes are unaffected because they bind their class explicitly).
3. Fix the same passage in `docs/codebase/checks-broadcast.md:56-70`.
4. Fix `test_gateway_checks_op_ids_keep_their_classes`'s docstring, and either scope its name/parametrisation to "what `classify_op` returns in isolation" or assert against the routes' real effective classes so it stops enshrining `other` for `checks.results.post`.
5. While you are in this file: the `BroadcastEvent.op_class` field docstring 130 lines below (`events.py:345-347`) enumerates the valid classes and does not list `checks` — this PR widens a gap that already omitted `approval`. Add both. Its closing sentence, "Derived from `classify_op` at publish time", is the same misconception this finding is about; qualify it with "unless a route bound an explicit `audit_op_class`, which takes precedence (`broadcast/overrides.py:357`)".

**Source:** `backend/src/meho_backplane/broadcast/overrides.py:357`; `backend/src/meho_backplane/api/v1/checks.py:124-126,156-158,186-188`; `backend/src/meho_backplane/audit_query/query.py:240-242`; `backend/src/meho_backplane/audit.py:211,235-255`.

backend/tests/test_checks_broadcast.py
```diff
@@ -0,0 +1,418 @@
...
+    sessionmaker = get_sessionmaker()
+    async with sessionmaker() as session:
+        row = await session.get(CheckDashboard, _DASHBOARD)
+        assert row is not None
+        assert row.last_rollup_state == "critical"
+    assert len(mail.calls) == 1, "the email notifier must be unaffected by a feed outage"
+
+
+@pytest.mark.asyncio
+async def test_no_claim_publishes_nothing(monkeypatch: pytest.MonkeyPatch) -> None:
```
damir-topic: **[💡 m1]** — the test name and docstring describe the opposite of what the first half asserts

**What:** `test_no_claim_publishes_nothing` and its docstring ("A persist that moves no rollup state publishes no event") describe a scenario in which nothing publishes. But the Dashboard is seeded with `last_rollup_state=None`, so the *first* `investigate_on_transition` call wins a real `None → critical` claim and the test asserts `len(fake.events) == 1`. Only the second call is the no-claim case, and it is the `== 1` assertion after it that carries the actual contract.

**Why it matters:** The name is what a future reader greps for when they change the claim path, and it currently promises a guarantee ("no claim → no publish") that the body only demonstrates in its second half. Worse, a regression that made the publish fire per Sensor evaluation instead of per claimed edge would be caught by the second assertion, but someone reading only the name could reasonably delete or weaken the first half as redundant setup and not realise the second assertion depends on the exact event count established by it.

**Required change:** Rename to something that states the real invariant — e.g. `test_second_persist_on_an_unchanged_rollup_does_not_republish` — and reword the docstring to say the first call establishes the claimed edge (one event) and the second, which finds nothing to claim, must leave the count at one. Keep the assertions as they are; only the name and docstring are wrong.

**Source:** Same finding as the open Copilot thread on this test.

CHANGELOG.md
damir-topic: Fixed in `db2523a2` — you're right, and the alert an operator would have built from those release notes would have been permanently silent.

Reworded so the two halves of the fail-open posture are named separately everywhere they appear: a Valkey/`XADD` failure is absorbed by `publish_event` and surfaces on `broadcast_publish_failed` + `broadcast_publish_errors_total` (the feed-wide signals to alert on), while `checks_transition_broadcast_failed` is scoped to a failure to *build* the event before it reaches the publisher. Sites: `CHANGELOG.md`, `docs/codebase/checks-broadcast.md` ("Failure posture"), the `checks/broadcast.py` module docstring, and the AC-3 line in the PR body.

No source behaviour changed. One extra thing while in there: `test_publisher_failure_is_swallowed_and_logged` was injecting `RuntimeError("valkey down")` at the publish seam, which would have contradicted the corrected prose — renamed/reworded (and its sibling to `test_broadcast_failure_leaves_the_memo_and_the_mail_intact`) to say what the injection actually stands in for.

backend/src/meho_backplane/broadcast/events.py
damir-topic: Fixed in `db2523a2`. Confirmed your finding independently before rewriting: all three routes bind `audit_op_class` (`api/v1/checks.py` `"write"` / `"read"` / `"write"`), `overrides.py:357` reads `op_class_override if op_class_override else classify_op(op_id)`, the override lands in `payload["op_class"]` via `_resolve_op_id_and_class_override`, and `audit_query/query.py:240-242` filters that stored value in SQL. So both of my original reasons were impossible, and the CHANGELOG's `checks.results.post` → `other` claim was a factually false release note.

`_CHECK_EVENT_OPS` and the regression test stay. Rationale re-anchored to the true hazard — `classify_op` is the **read-time** classifier at `ui/routes/audit/routes.py:439` and `ui/routes/broadcast/event.py:141`, supplying the displayed class/badge and feeding `is_aggregate_only`, so a prefix branch would relabel every gateway row there retroactively (derived on read, no deploy boundary) and sweep in any future non-transition `checks.*` op-id.

Corrected in seven places, not five — your list plus two more carrying the same sentence: `classify_op`'s own step-4c docstring ("must keep their `write` / `read` / `other` classes") and the new `checks` bullet in `docs/codebase/broadcast.md`. `BroadcastEvent.op_class` now enumerates `approval` and `checks` and qualifies "derived from `classify_op`" with the `audit_op_class` override precedence. The regression test is renamed to `test_gateway_checks_op_ids_are_not_swept_into_checks` and its docstring now states explicitly that the pinned values are what `classify_op` returns *in isolation*, not the class those rows ship with.

backend/tests/test_checks_broadcast.py
damir-topic: Fixed in `db2523a2` — renamed to `test_second_persist_on_an_unchanged_rollup_does_not_republish` and the docstring now says the first call wins a real `None → critical` claim and publishes, and only the second is the no-claim case. Assertions unchanged; added a message to the first `assert` so both halves read as deliberate.

damir-topic (commented):




**Anchor task:** #2720 — `checks.transition` broadcast events + `checks` op-class on the tenant feed
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `db2523a2`

> **Why this is posted in COMMENT state.** GitHub rejects both `APPROVE` and
> `REQUEST_CHANGES` from the pull request author's own identity
> ("Can not approve your own pull request"), the same fallback used on
> #2724 / #2726 / #2728. The `**Verdict:**` line above is the binding
> signal for the orchestrator, not the GitHub review state.

Both blockers from iteration 1 are resolved. No new findings, no inline
comments this round.

### The prose-only claim — verified independently, and more strictly

The fix report claims the `src/` change is prose-only, evidenced by filtering
the diff to non-comment lines. I did not take that on trust and did not
re-run the same filter. Both changed source files were parsed at `dc0c96ea`
and at `db2523a2` with `ast`, module/class/function docstrings stripped, and
`ast.dump()` compared:

```
backend/src/meho_backplane/broadcast/events.py: AST-minus-docstrings identical = True
backend/src/meho_backplane/checks/broadcast.py: AST-minus-docstrings identical = True
```

Comments never enter the AST, so this is a proof rather than a heuristic:
**zero executable lines changed under `src/`.** `_CHECK_EVENT_OPS` is a
module-level `frozenset` literal and would have shown as an AST delta had a
single member moved — it did not. Iteration 1's behavioural evidence for
AC-1…AC-5 therefore carries over intact and is not re-litigated here.

One thing that proof deliberately does *not* cover: doctests live inside
docstrings and are executable under `--doctest-modules`, and `classify_op`'s
docstring — which carries the AC-1 doctest — is one of the things that
changed. Re-run separately: 1 passed.

### B1 — resolved. An operator can now build an alert that fires.

The only bar that matters is whether someone reading the shipped text wires
a monitor that actually fires on a Valkey outage. They will, and the text
now steers them away from the wrong signal explicitly rather than merely
omitting it:

- **CHANGELOG** — "A Valkey outage is absorbed by the shared broadcast
  publisher and stays on its existing feed-wide signals — the
  `broadcast_publish_failed` warning and the `broadcast_publish_errors_total`
  counter; **alert on those, not on** the new
  `checks_transition_broadcast_failed` warning, which fires only if the
  event cannot be built before it reaches the publisher."
- **`docs/codebase/checks-broadcast.md` → Failure posture** — split into two
  paragraphs that close, respectively, with "Those are what a 'transitions
  stopped reaching the feed' alert should watch" and "It is a construction-bug
  signal, not an outage signal."
- **`checks/broadcast.py` module docstring** and the **PR body** (summary
  bullet + AC-3 line) say the same thing.

The metric name was re-verified here rather than accepted from the fix
report, and one step further than the report went. The constructor argument
at `publisher.py:103-104` is `Counter("broadcast_publish_errors_total", …)`,
and `prometheus_client` strips/re-appends `_total`, so the exposed name could
have differed. Running `generate_latest` against a scratch registry and
reading the exposition text — not just the sample-name set — gives
`broadcast_publish_errors_total 1.0`. It is also genuinely reachable: the
counter registers in the default `REGISTRY`, which `metrics.py:76` renders
via `generate_latest(REGISTRY)`, wired at `main.py:1191` `@app.get("/metrics")`.
The name in the release notes is the name on `/metrics`, end to end.

**Text over code was the right call**, for a slightly stronger reason than
the fix report gave. It is not only that `publish_event` returns `None` on
both paths. The swallow *is* the feed's contract — `publisher.py`'s own
docstring states "The caller … MUST NOT depend on this succeeding — the audit
row is the canonical record." Handing one caller a private success/failure
channel would create a second, contradictory contract on the same seam. And
the observability an operator needs already exists at the correct altitude:
feed-wide, because a Valkey outage is a broadcast problem, not a checks
problem. Per-caller outage signals would produce N alerts for one incident.

### B2 — resolved across all seven sites; nothing residual

I swept for the false sentence rather than counting the fixer's sites. Grep
for `split the historical meaning` / `deploy boundary — rows written` /
`result-ingest data plane` / `highest-volume path` across `CHANGELOG.md`,
`backend/src`, `backend/tests` and `docs` returns **nothing**. All seven
sites now state the true reason (`classify_op` is the read-time classifier,
so a prefix branch retroactively relabels already-persisted rows in the
drawers):

1. `broadcast/events.py` `_CHECK_EVENT_OPS` block — leads with the negative
   result ("What is *not* at risk is what those rows persist"), which is the
   honest framing.
2. `broadcast/events.py` `classify_op` step 4c — *(site I did not name in
   iteration 1)*.
3. `broadcast/events.py` `BroadcastEvent.op_class` field docstring — now
   enumerates `approval` + `checks` and records the `audit_op_class`
   override precedence.
4. `CHANGELOG.md` — "keep the `write` / `read` / `write` classes their routes bind".
5. `docs/codebase/checks-broadcast.md` — "The checks op-class" + the
   `meho.audit.query` known-issues bullet.
6. `docs/codebase/broadcast.md` `checks` bullet — *(site I did not name in
   iteration 1)*.
7. `tests/test_broadcast_events.py` — renamed
   `test_gateway_checks_op_ids_keep_their_classes` →
   `test_gateway_checks_op_ids_are_not_swept_into_checks`, docstring now
   separates "what `classify_op` returns **in isolation**" from "the class
   those rows ship with".

**No site claims `checks.results.post` keeps class `other` any more.** The one
surviving `write` / `read` / `other` triple, at `checks-broadcast.md:165`, is
*true* and correctly scoped — it describes what the drawers **display**. I
executed `classify_op` on all three gateway op-ids to confirm it returns
exactly `write` / `read` / `other`, and the passage labels the divergence
"pre-existing … not something this change introduces", which is accurate.

The rewrite also documents the read-vs-write divergence as a named boundary
instead of quietly routing around it. That is more than iteration 1 asked for
and is the right disposition — fixing it means deciding whether the drawers
should read `payload["op_class"]` back, a broadcast/audit-wide call well
outside #2720.

### The two unasked-for test edits — correct call

Both tests injected `RuntimeError("valkey down")` at the publish seam. After
the B1 correction that would have re-encoded the false model in the test
suite, which is the first place the next author looks. Renaming
`test_publisher_outage_…` → `test_broadcast_failure_…`, changing the messages
to `"event construction failed"` / `"broadcast unavailable"`, and stating in
each docstring that the injection stands in for a pre-publish fault is the
minimum honest fix. I diffed both: **no assertion changed**, no coverage lost
— the only behavioural delta is a message string in a monkeypatched raise
that nothing asserts on. In scope (the PR's own new test file). Not doing it
would have left the contradiction B1 exists to remove.

### The correction comment on #2720 — defensible

The PR body links `issuecomment-5151571152`, which carries the original false
rationale verbatim and is reachable from this PR permanently. The correction
sits as the next comment in the same thread, headed "Correction to the AC-1
substitution rationale above", links back to what it corrects, states the
claim, why neither reason is possible (with citations I re-verified), and
names the real hazard. Editing the original would have rewritten the record;
a follow-up preserves it. Note that this skill's prohibition on issue comments
binds the *reviewer*, whose rationale — the orchestrator reads the review
directly, so an issue comment is noise — does not apply to a correction whose
whole purpose is to reach a human reading the issue. Good judgement.

### Acceptance criteria
- [x] AC-1 `classify_op("checks.transition") == "checks"` — unchanged from iteration 1 (AST-identical); doctest + `test_checks_transition_maps_to_checks` re-run green. The allowlist substitution stands; its rationale is now true.
- [x] AC-2 exactly one event per claimed edge under concurrency, both directions — unchanged executable code; `test_concurrent_transitions_publish_exactly_one_event` green.
- [x] AC-3 fail-open with a structured warning — behaviour unchanged; the previously-misstated half now names the signal that actually fires, verified against the exposed metric name.
- [x] AC-4 `broadcast_recent` filtered to `op_class=checks` — unchanged; proven load-bearing by iteration 1's enum-removal negative test.
- [x] AC-5 no migration, no new settings, CHANGELOG bullet, suite green — CHANGELOG bullet reworded in place (no new bullet, no new section).

### Project-board alignment
|---|---|---|---|---|---|
| #2720 | open (task) | Closes | ✅ | ✅ full | close on merge |
| #2716 | open (parent initiative) | none | — | n/a | leave open — this is scope item 3 |

`closingIssuesReferences` still resolves to exactly #2720 and nothing else.
Hierarchy coherent: #2720 → #2716 → Goal #221. No delivered-but-unlinked
work: this iteration touched only prose.

### Predicted vs. actual diff
- ✅ `CHANGELOG.md`, `backend/src/meho_backplane/broadcast/events.py`, `backend/src/meho_backplane/checks/broadcast.py`, `docs/codebase/checks-broadcast.md`, `docs/codebase/broadcast.md` — exactly the B1/B2 sites, all prose
- ✅ `backend/tests/test_broadcast_events.py`, `backend/tests/test_checks_broadcast.py` — m1 rename + the two disclosed test edits
- ❓ none unexpected
- 🚫 none undisclosed
- 🕳️ none missing

### Integration verification
All commands ran inside `.claude/worktrees/pr-2729-review` at `db2523a2`.

- [x] `ruff check src/ tests/` — All checks passed
- [x] `ruff format --check` — 1382 files already formatted
- [x] `mypy src/` — 1 error, the known darwin-only `connectors/net/icmp.py:208` `MSG_ERRQUEUE` on an untouched file; 691 source files checked
- [x] `pytest --doctest-modules broadcast/events.py` — 1 passed (re-run specifically because docstrings were the only thing that moved)
- [x] `pytest` affected 3 files — 134 passed, 4 skipped (identical to iteration 1)
- [x] `pytest` adjacent 7 files (broadcast classifier / overrides / publisher / history, checks investigate, ui audit, ui broadcast feed) — 131 passed, 2 skipped, 1 error = the known-environmental RESPX teardown in `test_ui_audit.py`
- [x] `code-quality.py --files` on both changed source files — 0 blockers, 3 warnings, all pre-existing (`events.py` 863 lines, `redact_payload` 69, `publish_check_transition_event` 67)
- [x] CI on `db2523a2` — polled to completion: Python (ruff + mypy + pytest) SUCCESS, Python (integration testcontainers) SUCCESS, backplane image build SUCCESS, plus CodeQL / Semgrep SAST / Semgrep OSS / TruffleHog / Helm / CLI API snapshot freshness / DCO / CodeRabbit all SUCCESS. Zero failures.
- [x] Metric-name exposition check — `generate_latest` output read directly, not inferred
- ⚠️ not run: full 17-minute local suite. Justified by the AST proof (zero executable lines under `src/`) plus separately re-run doctests; iteration 1's green full run on `dc0c96ea` transfers, and CI's own full leg is green on `db2523a2`.
- n/a: frontend, migrations replay, docker compose, agent eval, UI flow — no such surface touched

None. No inline comments posted this iteration.

Recorded for the review trail; none block merge or require a response here.

- Copilot's suppressed note on `broadcast/history.py:144` is factually
  correct: `OP_CLASS_ENUM` is a *proper subset* of `classify_op`'s outputs
  (it omits `credential_write` and `approval`, which `agent_events.py:188`
  documents for its own enum), and the docstring this PR extended does not
  say so. Pre-existing — both omissions predate #2720. Pairs with the item
  below into one follow-up.
- `ui/routes/broadcast/feed.py:104,121` — `OP_CLASS_BADGE_CLASSES` /
  `OP_CLASS_FILTER_OPTIONS` omit `approval`, `credential_write` and now
  `checks`. Carried forward from iteration 1.
- `ui/routes/broadcast/aggregate_gate.py:104` — the `resolve_op_id` docstring
  still claims `classify_op` "yields the same class the broadcast publisher
  computed", untrue for any route binding an explicit `audit_op_class`. This
  is the direct origin of the B2 misconception; worth fixing before the next
  author inherits it.
- The PR body still links the uncorrected `#2720` comment rather than the
  correction. They are adjacent in the same thread and the correction's
  heading points back at it, so this is cosmetic.

### Housekeeping before merge (administrative, not code)
- The three CodeRabbit/Copilot threads (CHANGELOG, docs failure posture, test
  naming) are the same three findings and are all fixed in `db2523a2`, but the
  threads are still unresolved and unreplied. My own three threads carry
  "Fixed in `db2523a2`" replies and are also still open. Under the repo's
  "all comments resolved before merge" policy these need resolving —
  `mergeStateStatus` is `BLOCKED`.

### Docs
- [x] `docs/codebase/checks-broadcast.md` — updated; both previously-wrong
  sections now match the code, and the read-vs-write divergence is recorded
  as an explicit boundary
- [x] `docs/codebase/broadcast.md` — updated; the `checks` bullet now states
  the true rationale
- No .NET analogies, no `.planning/` references, `AGENTS.md` / `CLAUDE.md` untouched

- The B2 rewrite leads with what is *not* at risk before naming the real
  hazard. That ordering is what stops the next reader re-deriving the wrong
  model, and it is why the seven sites now read as one coherent explanation
  rather than seven patched sentences.
- Verifying the *exposed* Prometheus name rather than the constructor
  argument was the right instinct — that distinction is exactly where a
  well-intentioned runbook line goes wrong.
- Catching the two `"valkey down"` test injections unprompted. That
  contradiction was created by the B1 fix itself and would have quietly
  outlived it.

CHANGELOG.md
```diff
@@ -90,6 +90,28 @@ connector-related release-notes line.
...
+  memo, the email notification, and the runner's persist path
```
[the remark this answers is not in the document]
damir-topic: Fixed in `db2523a2` — and "adjust the wording" was the correct half of your either/or.

The bullet now names the real signal and steers explicitly away from the wrong one: a Valkey outage "is absorbed by the shared broadcast publisher and stays on its existing feed-wide signals — the `broadcast_publish_failed` warning and the `broadcast_publish_errors_total` counter; alert on those, not on the new `checks_transition_broadcast_failed` warning, which fires only if the event cannot be built before it reaches the publisher."

The other branch — changing the publisher integration so the checks-specific warning fires on Valkey failures — was deliberately not taken. The swallow *is* the feed's contract (`publisher.py`: "The caller ... MUST NOT depend on this succeeding — the audit row is the canonical record"), so giving one caller a private success/failure channel would put a second, contradictory contract on the same seam. And a Valkey outage is feed-wide, not a checks problem: a per-caller outage signal would produce N alerts for one incident. `broadcast_publish_errors_total` was verified to be the name actually exposed on `/metrics` by reading `generate_latest()` output rather than the constructor argument, since prometheus_client rewrites `_total` suffixes.

docs/codebase/checks-broadcast.md
```diff
@@ -0,0 +1,166 @@
...
+subsystem identity.
+
+## Failure posture
+
+Fail-open twice over. `publish_event` already swallows every Valkey error
+(at-most-once delivery is the feed's documented contract), and
+`publish_check_transition_event` wraps construction in a second guard so
+a malformed field cannot reach the caller either. A swallowed failure
+logs `checks_transition_broadcast_failed` at warning with the dashboard
+id and both states.
```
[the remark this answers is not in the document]
damir-topic: Fixed in `db2523a2`. Your parenthetical — that `checks_transition_broadcast_failed` only fires for errors before/around event construction — is essentially the wording the fix adopted.

"Failure posture" is now split into one paragraph per guard. The publisher half: "a Valkey outage or a failing `XADD` does **not** reach this module's guard: it surfaces on the existing feed-wide signals, the `broadcast_publish_failed` warning and the `broadcast_publish_errors_total` counter ... Those are what a 'transitions stopped reaching the feed' alert should watch." The module half: "That guard, and only that guard, logs `checks_transition_broadcast_failed` at warning with the dashboard id and both states. It is a construction-bug signal, not an outage signal."

The same correction was applied everywhere the claim had propagated — `CHANGELOG.md`, the `checks/broadcast.py` module docstring, and the AC-3 line in the PR body — and the two tests that injected `RuntimeError("valkey down")` at the publish seam were reworded ("event construction failed" / "broadcast unavailable", plus a docstring clause naming what the injection stands in for) so the suite stops encoding the model you flagged. No assertion changed and no source behaviour changed; the fail-open behaviour was already correct.

backend/tests/test_checks_broadcast.py
```diff
@@ -0,0 +1,418 @@
...
+    """A persist that moves no rollup state publishes no event.
+
+    The Dashboard is already ``critical`` in the memo, so the
+    compare-and-swap finds nothing to claim. Guards against the publish
+    being wired outside the claim-win branch, which would emit one event
+    per Sensor evaluation instead of one per edge.
+    """
```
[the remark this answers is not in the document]
damir-topic: Fixed in `db2523a2` — renamed to `test_second_persist_on_an_unchanged_rollup_does_not_republish`, which is the invariant the body actually proves.

The docstring now says it outright: the Dashboard starts with a NULL memo, so the first `investigate_on_transition` wins a real `ok -> critical` claim and publishes once; the second call finds the memo already `critical`, the compare-and-swap claims nothing, and the count must stay at one.

Assertions are unchanged. The previously bare first `assert` also gained a message — "the first persist claims the edge and publishes it" — so that half reads as a deliberate precondition rather than deletable setup, which was the concrete risk in leaving the old name in place.

CHANGELOG.md
damir-topic: Closing this out — the fix was re-verified at review iteration 2 against the only bar that matters here: does an operator reading the shipped text build a monitor that actually fires on a Valkey outage? Yes, and the text now steers away from the wrong signal explicitly rather than merely omitting it.

Two things were checked rather than taken from the fix report. (1) `broadcast_publish_errors_total` is the name really exposed on `/metrics`: `generate_latest()` exposition text was read directly (`# TYPE broadcast_publish_errors_total counter`), because prometheus_client strips and re-appends `_total` and the constructor argument alone would not have proven it; the reachability path default `REGISTRY` -> `metrics.py:76` -> `main.py:1191 @app.get("/metrics")` was walked too. (2) The text-over-code call was upheld on a stronger reason than "both paths return `None`": the swallow is the publisher's documented contract, so a private per-caller success channel would create a second contradictory contract on that seam, and outage observability already exists at the right altitude — feed-wide, because a Valkey outage is a broadcast problem, not a checks problem.

Resolving.

backend/src/meho_backplane/broadcast/events.py
damir-topic: Closing this out — re-verified at review iteration 2, with two checks beyond the rewrite itself.

`classify_op` was executed live on all three gateway op-ids: `checks.assignment.put` -> `write`, `checks.assignment.get` -> `read`, `checks.results.post` -> `other` (and `checks.transition` -> `checks`). That is what establishes the one surviving `write / read / other` triple — at `docs/codebase/checks-broadcast.md:165` — as a *true* statement rather than a residual site: it describes what the audit/broadcast drawers **display**, and the passage labels the divergence "a pre-existing read-vs-write divergence for any route binding an explicit `audit_op_class`, not something this change introduces."

A residual-claim sweep was also run instead of trusting a site count: grep for "split the historical meaning", "deploy boundary — rows written", "result-ingest data plane" and "highest-volume path" across `CHANGELOG.md`, `backend/src`, `backend/tests` and `docs` returns nothing. Every remaining site anchors to the true reason (`classify_op` is the read-time classifier, re-run at render time by `ui/routes/audit/routes.py:439` and `ui/routes/broadcast/event.py:150`), and the cited symbols were spot-checked in the worktree — including `overrides.py:357` `op_class = op_class_override if op_class_override else classify_op(op_id)`.

Recording the read-vs-write divergence as a named boundary in the docs rather than routing around it was above the ask and the right disposition: actually removing it means deciding whether the drawers should read `payload["op_class"]` back, which is a broadcast/audit-wide call well outside #2720.

Resolving.

backend/tests/test_checks_broadcast.py
damir-topic: Closing this out — verified at review iteration 2. The rename and docstring land as described: the docstring now states that the Dashboard starts with a NULL memo so the first `investigate_on_transition` wins a real claim (surfaced as `ok -> critical`) and publishes, and only the second call — which finds the memo already `critical` and claims nothing — is the no-claim case. Assertions are unchanged, and the first `assert` carries a message so neither half reads as redundant setup. `test_checks_broadcast.py` and its two sibling suites are green (134 passed, 4 skipped — byte-identical counts to the pre-fix run).

Resolving.

damir-topic:
### Disposition of Copilot's suppressed note on `broadcast/history.py:144`

Copilot's second review generated no new comments but suppressed one, so it has no thread to reply in. Recording the disposition here rather than letting it fall off the record.

**The claim is true.** `OP_CLASS_ENUM` is a proper subset of `classify_op`'s outputs: it omits `credential_write` (`broadcast/events.py:536`) and `approval` (`:544`), and the docstring this PR extended does not say so. Verified by reading both, not accepted on the bot's word.

**It is out of scope for #2720, deliberately.**

1. *Both omissions predate this PR.* On `origin/main`, `classify_op` already returns `credential_write` and `approval` while `OP_CLASS_ENUM` already omits both; #2720 only adds `checks`. The gap is not widened here in any way that makes something this PR asserts false.
2. *Widening it is not a docstring edit.* `mcp/tools/broadcast.py:308,1236` advertise `list(OP_CLASS_ENUM)` as the `filter.op_class` `enum` on `meho.broadcast.recent` and `meho.broadcast.watch`, and `mcp/handlers.py:337` validates every `tools/call` against that `inputSchema` before the handler runs. Adding members changes the published agent-facing contract of two shipped tools. Per the tuple's own docstring the taxonomy is shared with `meho.audit.query` and pinned by `tests/test_mcp_tools_list_shape_conventions.py`, so the change lands on both surfaces at once, and it pairs with `ui/routes/broadcast/feed.py:104,121`, whose `OP_CLASS_BADGE_CLASSES` / `OP_CLASS_FILTER_OPTIONS` omit the same classes.

Doing that on a PR whose acceptance criteria do not need it would be exactly the speculative scope creep CLAUDE.md rule 3 forbids. It is carried instead in this PR's review record as an adjacent improvement (`.claude/auto/review-results/pr-2729-iter-2.json`, `adjacent_improvements`), alongside the `feed.py:104,121` badge/filter gap and the stale `ui/routes/broadcast/aggregate_gate.py:104` docstring — the three want one change, one review, one place where the taxonomy is made consistent. To be precise about status: no tracking issue is open for them yet.

Issue #2700, opened 2026-07-29
`tools/call` is accepted without a negotiated MCP session → the audit row carries `agent_session_id: null` and the whole trace is invisible to session-lineage replay, while `/status` reports `mcp_session_id_capture: "always"`


ilhan-karic:
We drove a session-lineage isolation test against `/audit/sessions/{id}/replay` and found the surface works well — but that it can only ever see a subset of MCP traffic, in a way nothing at read time reveals.

**What we expected.** With `MCP_REQUIRE_SESSION_ID` enforced and `/status` reporting `mcp_session_id_capture: "always"`, we expected every `method=MCP` audit row to carry a session id, and therefore to be replayable.

**What actually happens.** A `tools/call` sent with **no prior `initialize` and no `Mcp-Session-Id` header** is executed normally, and the row it writes has a null lineage column. One request reproduces it:

```
POST /mcp
Authorization: Bearer <jwt>
Content-Type: application/json
Accept: application/json, text/event-stream

{"jsonrpc":"2.0","method":"tools/call","id":1,
 "params":{"name":"meho.status","arguments":{}}}
```

→ `HTTP 200`, no `error` key, full result — and the result body itself reports `mcp_session_id_capture: "always"`. The audit row for that call:

```
method            = MCP
path              = /mcp/tools/call/meho.status
status_code       = 200
agent_session_id  = null
```

This is not limited to a synthetic probe. In our tenant, one shared service identity that reaches `/mcp` via a minted JWT from a script has **389 of 389** `method=MCP` rows null over an 8-day window — including `approval.reject`, `meho.connector.delete` and `meho.connector.ingest`, i.e. governed writes. Over the same period an MCP client that *does* perform the handshake has 32 of 32 rows populated. The split is per-client, and it is invisible unless you go looking for it.

**Why we think this matters more than a missing column.**

1. **An empty replay forest is indistinguishable from an empty history.** `/audit/sessions/{id}/replay` returns `{root: [], row_count: 0}` both for a session that did nothing and for a principal whose client never negotiated a session. An investigator asking "what did this identity do?" via session lineage gets silence and no indication that silence is unreliable. That is the one property a forensic surface most needs not to have.

[...]

Issue #2699, opened 2026-07-29
sensor.create accepts an arbitrary unvalidated identity_sub; dispatch audit rows are attributed to the spoofed principal


ilhan-karic:
## Summary

`sensor.create`'s `identity_sub` is caller-supplied free text with ZERO validation — it accepts
another principal's `sub` (or a wholly non-existent string). Every audit row and broadcast event
the sensor's scheduled dispatches emit then carries `operator_sub = <the spoofed sub>`, so a
forensic `query_audit principal=<victim>` attributes the sensor's recurring reads to a principal
who never created or ran anything.

Observed live 2026-07-28 against a v0.25.0 deploy (chart 0.25.0 / git `4a70391a`), on both the
MCP path (`meho.sensor.create`) and the REST path (`POST /api/v1/sensors`).

## What was observed

Three sensor-create calls as a `tenant_admin` principal, each on the safe op `argocd.app.list`
against a registered ArgoCD target:

    identity_sub supplied                        result
    (omitted)                                    201 — defaults to "__sensor__"
    <the sub of a real, DIFFERENT principal>      201 — accepted verbatim
      (a shared service-account principal in
       the same tenant)
    "this-principal-does-not-exist-anywhere"      201 — accepted verbatim
      (not a UUID, not a principal, free text)

No existence check, no ownership check, no capability check, no format check beyond a 256-char
cap. `created_by_sub` recorded the true caller correctly while `identity_sub` took the bogus value
verbatim — i.e. the two fields disagree exactly as described. All three sensors were deleted
immediately; the tenant is back to zero sensors.

## What it is NOT (scoped deliberately — this is not privilege escalation)

Source-read of `evoila/meho` @ v0.25.0 (`4a70391a`) settles the severity honestly. `identity_sub`
is **attribution-only**, not authorization-bearing:

- `checks/runner.py:278-294` `_sensor_operator()` builds a synthetic `Operator` where **every
  authority-bearing field is a hardcoded literal** — only `sub` and `tenant_id` come from the
  sensor row. `capabilities=frozenset()`, `platform_admin=False`, `raw_jwt=""`.
- **No credential selection.** `connectors/_shared/vault_creds.py:376-381` refuses any
  operator-context credential read when `raw_jwt` is empty, and the connector credential/session
  caches key on `(tenant_id, target_id)` with no identity component
  (`connectors/_shared/cache_key.py:41-54`). Naming a principal does not borrow its secrets.
[...]

PR #2766, opened 2026-08-03
fix(dispatcher): name the target's product in the no_connector error

### LINKED ISSUE #2701: A target registered with a `product` no connector advertises is accepted at registration and fails every dispatch; the `no_connector` error names the caller's `connector_id` instead of the target's product
**Up front: the root cause here is our own misconfiguration.** We registered a Kubernetes target with `product: kubernetes` while the connector advertises `product: k8s`. That is on us, and our fix is to re-register the target. We are filing it because the product turned a one-character config mistake into a genuinely puzzling debugging session, and two small changes would have made it a non-event.

**What happens.** Every operation against that target fails identically, regardless of op:

```
call_operation(connector_id="k8s-1.x", op_id="k8s.about",
               target="<k8s-target>", params={})

{"status": "error",
 "error": "no_connector: no implementation for product='k8s' version='1.x'",
 "extras": {"error_code": "no_connector",
            "product": "k8s",
            "version": "1.x",
            "exception_message":
              "no connector advertises support for (product='kubernetes', version=None)"}}
```

Confirmed on `k8s.about` and `k8s.deployment.info` — not op-specific. The connector is ingested, enabled, and dispatchable (25 of 25 ops enabled); the target resolves fine. Only the join between them is broken.

**Problem 1 — the error blames the caller.** Read the top-level `error`: *"no implementation for product='k8s' version='1.x'"*. Those are the values **we passed**, echoed back. The natural reading is "your `connector_id` is wrong", so that is what we chased — we re-listed connectors, tried `kubernetes-1.31` and other spellings, and confirmed `k8s-1.x` was correct and enabled. The actual cause is in `extras.exception_message`, which reports the *resolved* lookup: `product='kubernetes'`. That one line is the whole answer and it is the field a caller reads last.

Concretely: the top-level message should name the resolved target product and say the mismatch out loud — something like *"target `<name>` declares product='kubernetes'; no connector advertises that product (did you mean 'k8s'?)"*. The information is already in hand at the point the error is raised.

**Problem 2 — the misconfiguration is accepted at write time.** Target registration takes a free-text `product` and stores it without checking that any connector advertises it. The failure then surfaces at *every subsequent dispatch*, arbitrarily far from the change that caused it, to whoever happens to call next rather than to whoever mis-registered. Rejecting an unknown product at registration (or warning loudly, if unknown-product-now / connector-ingested-later is a supported ordering) moves the error to the moment it can be understood and fixed in seconds.

**Worth noting** — our own kb has warned about this exact spelling for over a year (*"`product: k8s`, not `kubernetes` — that's the canonical pitfall"*), and we still shipped it. A documented footgun that a careful team steps into anyway is usually better fixed with a validation than with more documentation.

We also see a migration in the history that backfilled product-spelling drift, so the class of problem is known. This looks like either a row that escaped it or a mapping that does not include `kubernetes` → `k8s`. Whether the right fix is normalisation, a resolver alias set, or write-time rejection is your call — we would just like the failure to name itself.

**Impact on us, for calibration:** this made the "k8s reads via the backplane" workflow unusable, so we fell back to our local `kubectl` wrapper and filed this rather than silently working around it. Low urgency; nothing is blocked that the fallback does not cover.

damir-topic:
## Summary

- A target registered under a `product` no connector advertises (the classic
  `product: kubernetes` where the connector advertises `product: k8s`) misses
  at connector resolution on **every** dispatch. The residual `no_connector`
  error echoed the caller's `connector_id`-parsed `(product, version)` — e.g.
  `no implementation for product='k8s' version='1.x'` — i.e. the values the
  caller *passed*, not the target's product the resolver actually rejected.
  It read as "your `connector_id` is wrong" and sent the reporter chasing
  connector ids instead of the mis-registered target row.
- The fix names the **target's** registered product and the version the
  resolver read (which may be `None`) in the top-level `error` and aligns
  `extras['product']` / `extras['version']` to the target. It falls back to
  the `connector_id` pair only when there is no target product to name
  (e.g. an ingested op dispatched with `target=None`).
- Diagnosability-only. Register-time rejection of an unadvertised `product`
  already shipped (#1144); this is the Part B follow-on that makes the
  residual runtime failure name itself.
- `resolve_target_version` is promoted from private (`_resolve_target_version`)
  so the dispatcher names exactly the version the resolver saw, without
  duplicating the fingerprint-vs-operator-asserted priority logic.

## Test plan

- [x] New async dispatcher test `test_no_connector_error_names_target_product_not_connector_id`
      in the exact #2701 shape (connector_id product `k8s` ≠ target product
      `kubernetes`, descriptor lookup by `connector_id` succeeds, no connector
      advertises `kubernetes`). Asserts `error.startswith('no_connector:')`,
      `product='kubernetes'` in the error, `product='k8s'` **not** presented as
      the unsupported product, `version=None` (not the connector_id's `1.x`),
      and `extras['product'] == 'kubernetes'`. **Proven to fail on the
      pre-fix dispatcher and pass after** (reverted the call-site change and
      re-ran: 1 failed → restored → 3 passed).
- [x] `pytest tests/test_operations_dispatcher.py tests/test_operations_connector_error_redaction.py tests/test_connectors_resolver.py` — 96 passed
- [x] `ruff check` — clean
- [x] `mypy` (dispatcher, _errors, resolver, connectors/__init__) — Success, no issues
- [x] `docs/codebase/error-message-shape.md` updated — the #2701 note added as
      the Part B resolution of Signal 5 (the same `kubernetes`/`k8s` scenario
      already catalogued there).

The core fix names the target's product (the DoD's non-breaking option a). Two
richer/more-invasive alternatives were deliberately **not** taken autonomously:

1. **"Did you mean 'k8s'?" near-match suggestion.** The issue asks for a
   suggestion clause. A string-distance near-match (`difflib`) against
   `registered_product_tokens()` does **not** catch the flagship case:
   `SequenceMatcher('kubernetes', 'k8s').ratio() ≈ 0.31`, well below any sane
   cutoff — `kubernetes`↔`k8s` is an abbreviation/alias relationship, not a
   typo. Delivering a real "did you mean 'k8s'?" needs a **curated alias map**
   (`kubernetes → k8s`), which is a new maintained data structure and a wider
   surface. Left out to keep this change surgical; easy to add on top if the
   team wants the alias table.

2. **Resolver alias / normalization so the mis-registered target actually
   resolves.** The issue floats "normalisation, a resolver alias set, or
   write-time rejection" as options. Making the resolver treat `kubernetes`
   as `k8s` would let the previously-broken row dispatch — a **behavior
   change**, not just a message change: it would silently repair (and thereby
   mask) a class of mis-registration that #1144 chose to reject loudly at
   write time. That is a policy call for a human, so it is out of scope here;
   this PR only makes the failure name itself.

## Out of scope

- **Part A — reject an unadvertised `product` at registration.** Already
  shipped in #1144 (`_canonicalise_and_validate_product` → structured 422
  `unknown_product`); untouched here.

Closes #2701

### DIFF
--- a/backend/src/meho_backplane/connectors/__init__.py
+++ b/backend/src/meho_backplane/connectors/__init__.py
@@ -43,6 +43,7 @@ from meho_backplane.connectors.resolver import (
     ResolutionLabel,
     resolve_connector,
     resolve_connector_or_label,
+    resolve_target_version,
 )
 from meho_backplane.connectors.schemas import (
     AuthModel,
@@ -99,6 +100,7 @@ __all__ = [
     "register_connector_v2",
     "resolve_connector",
     "resolve_connector_or_label",
+    "resolve_target_version",
     "shim_kind",
     "split_version",
     "validate_execution_profile",
--- a/backend/src/meho_backplane/connectors/resolver.py
+++ b/backend/src/meho_backplane/connectors/resolver.py
@@ -325,7 +325,7 @@ def resolve_connector(target: Any) -> type[Connector]:
             f"target has no product slug; cannot resolve a connector: target={target!r}"
         )
 
-    target_version = _resolve_target_version(target)
+    target_version = resolve_target_version(target)
     preferred_impl_id = getattr(target, "preferred_impl_id", None)
 
     candidates = _filter_candidates(product, target_version)
@@ -532,7 +532,7 @@ def _run_tie_break_ladder(
     return None, "ambiguous", candidates
 
 
-def _resolve_target_version(target: Any) -> str | None:
+def resolve_target_version(target: Any) -> str | None:
     """Pull the target's product version from fingerprint or operator hint.
 
     Two sources, in priority order:
--- a/backend/src/meho_backplane/db/models.py
+++ b/backend/src/meho_backplane/db/models.py
@@ -929,7 +929,7 @@ class Target(Base):
     # can still be created and probed; once the probe succeeds the
     # authoritative ``fingerprint.version`` takes precedence at resolver
     # time (see
-    # :func:`~meho_backplane.connectors.resolver._resolve_target_version`).
+    # :func:`~meho_backplane.connectors.resolver.resolve_target_version`).
     # G0.15-T6 (#1215) ships this column to break the chicken-and-egg
     # the v0.7.0 dogfood surfaced (RDC #753, signal 6): every typed
     # connector except K8s required ``fingerprint.version`` to resolve,
--- a/backend/src/meho_backplane/operations/_errors.py
+++ b/backend/src/meho_backplane/operations/_errors.py
@@ -285,11 +285,20 @@ def result_ambiguous_target(
 def result_no_connector(
     op_id: str,
     product: str,
-    version: str,
+    version: str | None,
     duration_ms: float,
     exception_message: str | None = None,
 ) -> OperationResult:
     """Resolver miss -- no registered impl for *(product, version)*.
+
+    ``product`` / ``version`` name the **failing input** the resolver
+    rejected -- i.e. the *target's* registered product and the version
+    the resolver read off it (which may be ``None`` when the target
+    carries no fingerprint or operator-asserted version), **not** the
+    ``connector_id`` the caller passed. The dispatcher's connector_id
+    already resolved a valid descriptor before this builder is reached
+    (else ``result_unknown_op`` fires), so echoing it here would blame
+    the wrong input (evoila/meho#2701).
 
     ``exception_message`` (added by G0.14-T1 #1142) carries the
     :exc:`~meho_backplane.connectors.NoMatchingConnector` exception text
--- a/backend/src/meho_backplane/operations/dispatcher.py
+++ b/backend/src/meho_backplane/operations/dispatcher.py
@@ -232,6 +232,7 @@ from meho_backplane.connectors import (
     ResolutionLabel,
     ResultHandle,
     resolve_connector_or_label,
+    resolve_target_version,
 )
 from meho_backplane.connectors._shared.vcf_auth import ConnectorAuthError
 from meho_backplane.connectors.base import Connector, shim_kind
@@ -2091,10 +2092,28 @@ async def dispatch(
             # faults, which only arise when a target *was* supplied).
             return result_target_required(op_id, _elapsed_ms(started))
         if resolution_error == "no_connector":
+            # #2701: the resolver miss is on the TARGET's product/version,
+            # not the ``connector_id`` the caller passed. That connector_id
+            # already resolved a valid descriptor at Step 2 (else
+            # ``result_unknown_op`` fired), so echoing its parsed
+            # ``(product, version)`` here names the wrong input and reads as
+            # "your connector_id is wrong" -- the exact misattribution that
+            # sent the reporter chasing connector ids. Name the target's
+            # registered product and the version the resolver actually read
+            # (which may be ``None``). Fall back to the connector_id pair
+            # only when there is no target product to name (e.g. an ingested
+            # op dispatched with ``target=None``).
+            target_product = getattr(target, "product", None)
+            if isinstance(target_product, str) and target_product:
+                miss_product = target_product
+                miss_version: str | None = resolve_target_version(target)
+            else:
+                miss_product = product
+                miss_version = version
             return result_no_connector(
                 op_id,
-                product,
-                version,
+                miss_product,
+                miss_version,
                 _elapsed_ms(started),
                 exception_message=exception_message,
             )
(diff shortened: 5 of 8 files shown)


damir-topic (commented):




**Anchor task:** #2701 — `no_connector` error names the caller's `connector_id` instead of the target's product (G0.38 Part B)
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `db523754561eba2fc17527802dc09f2558e6df1b`

### Acceptance criteria (grounded DoD for #2701)
- [x] Residual `no_connector` at dispatch names the **target's** registered product and the version the resolver saw (may be `None`), not the `connector_id`-parsed pair — `dispatcher.py:2094` now uses `target.product` + `resolve_target_version(target)`, falling back to the connector_id pair only when the target has no product.
- [x] `extras['product']` / `extras['version']` aligned to the target — verified via the builder (`_errors.py`) which sources both `error` string and `extras` from the same `(product, version)` args.
- [x] Graceful fallback when target / `target.product` absent — `getattr(target, "product", None)` + `isinstance(... str) and ...` guard; `target=None` yields the connector_id pair, no crash.
- [x] New async dispatcher test in the exact #2701 shape (connector_id `k8s` ≠ target `kubernetes`, descriptor lookup succeeds, no connector advertises `kubernetes`) asserting `error.startswith('no_connector:')`, `'kubernetes'` present, `'k8s'` not presented as the unsupported product, `version=None`, `extras['product']=='kubernetes'`. **Verified it fails on pre-fix code (reverted the call-site: 1 failed) and passes after.**

### Design decision (honored)
Non-breaking **option (a)** taken — name the target's product as primary. Both richer alternatives are surfaced in the PR body as human-decision items and deliberately **not** shipped: (1) the "did you mean 'k8s'?" near-match (correctly rejected as a message-only enrichment that needs a curated alias map — `SequenceMatcher('kubernetes','k8s')≈0.31`), and (2) the **behavior-changing** resolver alias/normalization that would silently repair a mis-registration #1144 chose to reject loudly. This matches the grounded `design_decision` recommendation exactly.

### Project-board alignment
|---|---|---|---|---|---|
| #2701 | Open | Closes | ✅ | ✅ full (Part B / DoD scope) | close on merge |

Part A (register-time reject) already shipped in #1144 and is correctly out of scope. #2766 is the only open PR referencing #2701.

### Predicted vs. actual diff
- ✅ `dispatcher.py` no_connector branch (the fix), `_errors.py` `version: str | None`, new dispatcher test, `docs/codebase/error-message-shape.md` note
- ❓ `resolver.py` + `connectors/__init__.py` (promote `_resolve_target_version` → public `resolve_target_version`) and `db/models.py` + `test_connectors_k8s_e2e.py` docstring-reference updates — justified: reusing the resolver's own helper is how the dispatcher names *exactly* the version the resolver saw without duplicating the fingerprint-vs-operator priority logic
- 🚫 none (no undisclosed/stealth edits)
- 🕳️ none (nothing predicted-but-missing)

### Integration verification
- [x] `ruff check` (4 changed src files) — clean
- [x] `mypy` (dispatcher, _errors, resolver, connectors/__init__) — Success
- [x] `pytest tests/test_operations_dispatcher.py tests/test_operations_connector_error_redaction.py tests/test_connectors_resolver.py` — 96 passed
- [x] New test proven to catch the bug — fails against pre-fix call site (1 failed), passes after; restore left a clean tree
- [x] Sibling ghost/ghost test (`:633`) re-verified non-regressed (target product == connector_id product → same output)

### Findings
None — no blockers, no majors.

- The issue's illustrative message also names the **target name** ("target `<name>` declares product=…"); the DoD only requires the product/version, so this is a possible future enrichment, not required here.
- `resolve_target_version(target)` runs a second time on the error path (the resolver already computed it during resolution). It's a cheap pure `getattr` on a cold path — negligible; left as-is for the "name exactly what the resolver saw" clarity the PR intends.

### Docs
- [x] `docs/codebase/error-message-shape.md` — updated (the #2701 note filed as the Part B resolution of the catalogued `kubernetes`/`k8s` Signal 5)

### Scope note on #2699 / #2700
Out of scope for this PR and correctly **not** claimed. #2766 touches only the dispatch error-formatting substrate (dispatcher / _errors / resolver / models-docstring / tests / docs); it does not touch the #2699 sensor `identity_sub` choke point (`checks/service.py`, `checks/runner.py`) or the #2700 MCP session-capture path (`mcp/server.py`, `mcp/audit.py`). Those audit/session-integrity children remain open and are neither closed nor regressed here.

- The new test is a genuine regression guard, not a happy-path echo — independently confirmed to fail on the pre-fix dispatcher.
- Both invasive alternatives were surfaced with concrete rationale rather than silently shipped, exactly matching the grounded design decision.

Issue #2642, opened 2026-07-19
Task: give the in-process check-runner a credential identity on per-operator-WIF GSM deploys — background dispatch fail-closes on the empty-JWT guard


damir-topic:
Parent initiative: #2656 (G0.37 v0.25.0 ops-feedback hardening)
Parent goal: #221

## Summary

Follow-on from #2595 (shipped in v0.25.0): target-bound Sensors on a `credentialBackend=gsm` + **per-operator-WIF** deploy now dispatch but fail credential resolution — every evaluation returns `connector_error: VaultCredentialsReadError`. The GSM WIF path mints a credential by exchanging **the calling operator's Keycloak JWT** at `sts.googleapis.com`; the in-process check-runner dispatches as a synthetic operator with `raw_jwt=""`, so the shared loader's empty-JWT guard fail-closes before the exchange ever happens. On a pod with no ambient GCP identity (on-prem Kubernetes: no GKE Workload Identity, no mounted SA key), **0% of credentialed Sensors can evaluate** — the feature works interactively but not on its own scheduler, which is its whole point. Give scheduled/background dispatch a real credential identity.

## Problem

- `backend/src/meho_backplane/connectors/_shared/gsm_creds.py` — Phase-2 per-operator WIF exchanges `operator.raw_jwt` at `sts.googleapis.com` (docstring: "a system-initiated call (`raw_jwt=""` — health probe, scheduler) never reaches the WIF exchange").
- `backend/src/meho_backplane/checks/runner.py:281-291` — the runner's synthetic operator explicitly sets `raw_jwt=""` ("the runner forwards no bearer token").
- Net: `POST /api/v1/operations/call` succeeds (carries the operator's JWT) while the runner tick fails for the identical `(connector_id, op_id, target)` — dashboards render with every target-bound member `unknown`/`degraded` forever.
- Compounding diagnosability bug: the GSM backend reuses `vault_creds`'s shared loader **and its exception class**, so a no-Vault deploy surfaces `VaultCredentialsReadError` — Vault-named errors on a deploy with `vault: not_configured`.

## Approach (decided direction)

**Primary — runner service principal.** Give the in-process check-runner a service-principal identity (client-credentials JWT from Keycloak) to present as the WIF subject token for background dispatch — analogous to the satellite-runner principal (#2415), but for the in-process path. This keeps the WIF model intact (no static keys enter the flow) and gives GCP-layer audit a stable, attributable service identity for scheduled evaluations.

[...]

PR #2767, opened 2026-08-03
fix(checks): reject unowned sensor identity_sub at the create choke point

### LINKED ISSUE #2699: sensor.create accepts an arbitrary unvalidated identity_sub; dispatch audit rows are attributed to the spoofed principal

damir-topic:
## Summary

- A Sensor's `identity_sub` was caller-supplied free text with only a length cap and **no principal validation**. It is the `sub` the check runner dispatches every scheduled evaluation under, and that sub is stamped verbatim onto `AuditLog.operator_sub` and `BroadcastEvent.principal_sub`. A tenant-admin could therefore attribute a high-volume recurring dispatch stream to any principal they named — corrupting the audit ledger (#2699).
- Fix validates `identity_sub` at the shared `SensorAdminService.create` choke point (a pure in-memory ownership check that runs first, before any DB read): accept only the `"__sensor__"` sentinel or the creating operator's own sub (`created_by_sub`); any other value is refused with a new `SensorIdentitySubForbiddenError` → **422 `sensor_identity_sub_forbidden`** at REST and the same code as invalid-params at MCP. The service is the one place REST + MCP + CLI all route through; the `SensorCreate` Pydantic model cannot enforce this (it has no access to the authenticated operator).
- **Attribution-only, not privilege escalation**: `identity_sub` selects no credentials (the cache keys on `(tenant_id, target_id)` with no identity component; the runner's downstream service-principal token is a separate axis, #2642). The runner is left unchanged — it faithfully dispatches under the stored, now-validated sub, so the audited principal is guaranteed real.
- The existing pin test `test_dispatch_uses_custom_identity_sub` — which asserted the spoofable behaviour was intended — is **inverted**.

## Test plan

- [x] `ruff check` — clean (all touched files)
- [x] `mypy src` (strict) — clean on all touched files (one pre-existing, unrelated `connectors/net/icmp.py` `MSG_ERRQUEUE` error only reproduces on darwin; CI runs on Linux)
- [x] **Service rejection** (`test_service_create_rejects_foreign_identity_sub`): `create(created_by_sub="alice", identity_sub="bob-victim-sub")` raises `SensorIdentitySubForbiddenError` (`error_code == "sensor_identity_sub_forbidden"`).
- [x] **Creator's own sub allowed** (`test_service_create_allows_creator_own_identity_sub`) and **default sentinel** (`test_service_create_defaults_to_sentinel_identity_sub`).
- [x] **Dispatch-path proof** (inverts `tests/test_sensor_runner.py:547`): a spoofed sub is refused before the row is written (`test_service_create_refuses_spoofed_identity_sub_before_dispatch`), and a legitimately-created sensor (`identity_sub == created_by_sub`) dispatches under that real principal (`test_dispatch_uses_creator_owned_identity_sub` — `captured["operator"].sub == "op-owner"`, the value `_audit.py` stamps onto `AuditLog.operator_sub`).
- [x] **REST + MCP boundary wiring** (`test_rest_create_foreign_identity_sub_returns_422`, `test_mcp_create_foreign_identity_sub_surfaces_code`).
- [x] Full checks/sensor blast-radius suite — **223 passed** (`test_checks_*`, `test_ui_*`, `test_mcp_tool_sensors`, `test_sensor_runner`, `test_db_sensor`).

This PR implements **Option A** (the non-breaking option): validate `identity_sub` and keep the per-sensor knob.

**Option B — drop the `identity_sub` field/column entirely** (always dispatch as `__sensor__`, treat `created_by_sub` as the durable real-author record). This is arguably the cleaner end state: the runner's authority is fixed regardless (a synthetic no-capability operator), so a per-sensor identity knob buys nothing today except the ability to mis-attribute. It was **not** taken here because it is a **breaking change** — it removes a wire field (`SensorCreate.identity_sub`), an MCP tool property, and the Go CLI `--identity-sub` flag, and needs a DB migration on `sensor.identity_sub`. If Option B is preferred, note that the dispatch audit path only reads `operator.sub` today, so surfacing the human author on the audit row would additionally require plumbing `created_by_sub` onto the `AuditLog` row (does not exist today). Deferred as a human decision.

## Out of scope

- **Pre-existing spoofed rows**: the guard is create-time only (there is no update route). Any `sensor` row persisted before this landed with a spoofed `identity_sub` keeps dispatching under it — no data migration normalises historical rows. Deployments that ran the pre-guard build should re-audit existing sensors' `identity_sub`. Noted in `docs/codebase/sensor.md`.
- **Cross-tenant platform-admin create** (`body.tenant_id` set): the allowed set stays `{sentinel, operator.sub}` — a platform admin cannot name an arbitrary sub in the target tenant either. This is the safe default (attribute to the real actor or the sentinel).
- **The scheduler shares the same `identity_sub` free-text pattern** (`scheduler/service.py`, `scheduler/repository.py`) and is a candidate for the same guard, but is a separate surface outside #2699's scope. Adjacent finding.
- **`raw_jwt` / credential resolution** (#2642) is a deliberately separate axis and is untouched.

Closes #2699

## Summary by CodeRabbit

* **New Features**
  * Added validation to ensure sensor identity attribution uses the system sentinel or the creating operator’s own identity.
  * Added clear error reporting when an unauthorized identity is provided across REST and MCP interfaces.

* **Documentation**
  * Clarified identity ownership rules, error behavior, and creation-time limitations.

* **Bug Fixes**
  * Prevented unauthorized identity values from being persisted or dispatched.


damir-topic:




**Anchor task:** #2699 — Sensor `identity_sub` was caller-supplied free text, letting a tenant-admin attribute a recurring dispatch stream (and its `AuditLog.operator_sub` / broadcast `principal_sub`) to any principal.
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `7fd3acc2`

### Acceptance criteria (grounded #2699 final_dod)
- [x] `sensor.create` accepts `identity_sub` only as the `"__sensor__"` sentinel or the creator's own `created_by_sub`; any other value refused — verified at `checks/service.py:298` (`_require_owned_identity_sub` runs first) → `service.py:231` (`identity_sub not in (SENSOR_IDENTITY_SENTINEL, created_by_sub)`).
- [x] Enforced at the shared service choke point, **not** the `SensorCreate` Pydantic validator — verified: guard is in `SensorAdminService.create`, schemas.py only length-caps.
- [x] REST + MCP + Go CLI covered in one place — REST (`api/v1/sensors.py:179` → 422 `sensor_identity_sub_forbidden`) and MCP (`mcp/tools/sensors.py:214` → `McpInvalidParamsError`) both route through the service; the Go CLI POSTs to REST. `create_sensor` (repository) has exactly one production caller (`service.py:321`).
- [x] Every scheduled-dispatch audit row records the sentinel or creator, never an arbitrary sub — closed at the root: create is the only write path for `Sensor.identity_sub` (no update route; service exposes only create/list/get/delete). Runner builds `Operator(sub=snap.identity_sub)` (`runner.py:315`), and `_audit.py:583` stamps `operator_sub=operator.sub` / `_audit.py:655` `principal_sub=operator.sub`. A rejected value can never be persisted, so it can never be stamped.
- [x] A test proves a spoof cannot set the dispatch `operator.sub` — `test_service_create_refuses_spoofed_identity_sub_before_dispatch` (refused before the row is written) + `test_dispatch_uses_creator_owned_identity_sub` (a validated sub reaches dispatch as `operator.sub == "op-owner"`, the value `_audit.py` stamps).
- [x] The pin test `test_dispatch_uses_custom_identity_sub` (which asserted `sub == "svc:sensor-x"` was intended) is inverted — renamed to `test_dispatch_uses_creator_owned_identity_sub`, assertion flipped to `op-owner`.

### Design decision (grounded #2699 design_decision)
- [x] **Option A (non-breaking)** shipped: validate + keep the per-row knob.
- [x] **Option B (breaking)** — drop the field/column, always dispatch `__sensor__`, DB migration — surfaced in the PR body's "Alternative (needs human decision)" and in `docs/codebase/sensor.md`, **not** shipped. Correct posture.
- [x] Cross-tenant platform-admin sub-decision resolved: allowed set stays `{sentinel, operator.sub}` (documented in "Out of scope").
- [x] Migration sub-decision resolved: create-time-only, pre-existing spoofed rows explicitly out of scope, documented in `sensor.md`.
- [x] Attribution-only framing preserved (`raw_jwt` / #2642 not conflated).

### Project-board alignment
|---|---|---|---|---|---|
| #2699 | In Review | Closes | ✅ | ✅ full | close on merge |

No stealth changes; all 8 files are within the grounded blast radius. **#2700 (MCP un-negotiated-session forensic hole) is a separate bug and is NOT in scope for this PR** — the PR only `Closes #2699` and touches no `mcp/server.py` session path.

### Predicted vs. actual diff
- ✅ `checks/service.py` (guard + error), `checks/schemas.py` (docstring), `api/v1/sensors.py` (422 map), `mcp/tools/sensors.py` (error map + schema desc), 3 test files, `docs/codebase/sensor.md`
- ❓ none unexpected
- 🚫 none undisclosed
- 🕳️ none missing (Go CLI `--identity-sub` needs no change under Option A — it forwards to the now-guarded REST path; MCP JSON-schema description was updated)

### Integration verification (ran inside worktree on `7fd3acc2`)
- [x] `ruff check` (4 touched src files) — clean
- [x] `mypy` (4 touched src files) — clean
- [x] `pytest tests/test_checks_admin.py tests/test_mcp_tool_sensors.py tests/test_sensor_runner.py` — **67 passed** (includes the new service/REST/MCP guard tests and the inverted dispatch test)
- [n/a] migrations — none in this PR (CI "Detect migration-related changes" green; create-time-only guard, no schema change)
- [n/a] frontend / docker compose / agent eval — no such surface touched

### Findings
None blocking. No inline comments posted.

- `SENSOR_IDENTITY_SENTINEL = "__sensor__"` (service.py) is now a third copy of the `"__sensor__"` literal alongside `SensorCreate.identity_sub`'s Pydantic default and the `Sensor.identity_sub` column default. All three agree today and the guard is correct, so this is a latent drift risk, not a defect. A future PR could hoist the sentinel to a single shared constant and reference it from all three sites. (Also raised by the Copilot bot review.)
- The scheduler surface (`scheduler/service.py`, `scheduler/repository.py`) shares the same `identity_sub` free-text pattern and is a candidate for the same guard — correctly flagged out-of-scope for #2699 in the PR body.

### Docs
- [x] `docs/codebase/sensor.md` — updated: guard added to control flow (step 2), steps renumbered, create-time-only caveat + Option-B deferral recorded. No .NET analogies.

- Fix is at the correct architectural locus (the single service choke point), so REST/MCP/CLI are covered without per-transport duplication — and it closes the latent second-order `ApprovalRequest.principal_sub` / self-approval surface at the root by construction.
- Root-cause closure rather than symptom patch: because create is the sole write path, a rejected `identity_sub` is unpersistable and therefore unspoofable in the audit ledger — the integrity guarantee is real, not merely plausible.
- The breaking alternative is surfaced with a precise cost (wire field + MCP property + CLI flag + migration) and deferred as a human decision, exactly as the grounded design_decision required.

Issue #1147, opened 2026-05-26
Task: G0.14-T6 decouple audit-replay session-id capture from MCP_REQUIRE_SESSION_ID enforcement — G8.2 lights up automatically (signal 11)


damir-topic:
Parent goal: #221
Parent initiative: #1139

## Problem

`claude-rdc-hetzner-dc#697` signal 11
(`agent-session-id-never-captured-blocks-audit-replay`)
verified that the entire G8.2 audit-replay-as-tree feature
stack is **inert in the consumer's v0.6.0 deploy:**

- Migration 0014 (the column + index) is present.
- `POST /api/v1/audit/query` with `has_agent_session: true`
  returns 0 rows across the entire deploy lifetime.
- MCP `initialize` does NOT include an `Mcp-Session-Id` header
  in the response.
- `MCP_REQUIRE_SESSION_ID` env var is NOT set on the running
  pod.
- `GET /api/v1/audit/sessions/{id}/replay` returns 404 for
  every session-shaped id tried.

The whole G8.2 stack ships and works — but the upstream
session-id capture only fires when an env var is set,
which isn't called out in the v0.6.0 release notes or
`docs/RELEASING.md`. Operators discover the dependency
empirically by hitting the surface and finding no data.

Looking at the source at
[mcp/server.py:521-588](backend/src/meho_backplane/mcp/server.py#L521):

```python
def _bind_mcp_session_id(...):
    """Capture the Mcp-Session-Id header into a structlog contextvar.
    ...
    is True (MCP_REQUIRE_SESSION_ID env), a missing/empty header
    """
    if not session_id:
        _log.warning("mcp_session_id_required_but_missing")
        ...
        return 400, "invalid request: Mcp-Session-Id header is required"
    structlog.contextvars.bind_contextvars(mcp_session_id=str(session_id))
```

The gating logic conflates **capture** and **enforcement**.
The two should be independent:

- **Capture** = if the client sends `Mcp-Session-Id`, bind it
  to the contextvar so downstream audit rows carry it. **Always
  on.**
- **Enforce** = require the header; reject the request if
  missing. **Gated on `MCP_REQUIRE_SESSION_ID`.**

Decoupling these means audit-replay lights up automatically
for any client that sends the header (Claude Code does by
default per the consumer's `RDC Wall W7` reference in their
config), without operators needing to flip
`MCP_REQUIRE_SESSION_ID` and risk breaking older clients.

## Current state

[...]

PR #2768, opened 2026-08-03
fix(mcp): honest mcp_session_id_capture label + excluded null-session count on replay

### LINKED ISSUE #2700: `tools/call` is accepted without a negotiated MCP session → the audit row carries `agent_session_id: null` and the whole trace is invisible to session-lineage replay, while `/status` reports `mcp_session_id_capture: "always"`

damir-topic:
## Summary

- `/status` (`GET /api/v1/health`) reported `mcp_session_id_capture: "always"`, which reads as a guarantee that every `method=MCP` audit row carries a session id. It is not: capture is handshake-gated. A client that skips `initialize` and POSTs `tools/call` directly never receives an `Mcp-Session-Id` to echo, so its row lands `agent_session_id = NULL` and is invisible to G8.2 session-lineage replay — an empty replay forest was indistinguishable from an empty history (#2700).
- Ships the **non-breaking honest-label** resolution (the reporter's stated acceptable fallback, report option **b**): the label now tells the truth, and the replay surface reveals what it structurally cannot see.
- `mcp_session_id_capture_mode()` returns `"when_negotiated"` (not `"always"`) in the default lenient mode; `"enforced"` is unchanged.
- `GET /api/v1/audit/sessions/{id}/replay` now returns `excluded_null_session_count` alongside `row_count` — the tenant-wide tally of `method=MCP` rows with a NULL `agent_session_id`. A `row_count=0` response with a non-zero excluded count distinguishes an empty forest from an empty history.
- Lenient accept-and-null is preserved, so header-less script callers (the 389-row service identity in the report) are **not** broken.

## Design decision

Per the grounded contract, the load-bearing fork is *fail-loud-and-reject* (breaking) vs *keep-lenient-but-honest* (non-breaking). This PR implements the non-breaking option and describes the breaking one below for a human to decide. Synthetic per-call session ids are explicitly **not** an option — decision #1147 removed them because they pollute replay with uncorrelated one-row sessions.

## Test plan

- [x] `mypy src/meho_backplane/` — clean on changed files (1 pre-existing, unrelated `socket.MSG_ERRQUEUE` error in `connectors/net/icmp.py`, a darwin-only platform gap)
- [x] `code-quality.py --diff` — 0 blockers (only pre-existing size warnings)
- [x] `GET /api/v1/health` reports `mcp_session_id_capture == "when_negotiated"` and `!= "always"` in default mode — `tests/test_api_v1_health.py::test_mcp_session_id_capture_default_when_negotiated`, `tests/test_mcp_server.py::test_mcp_session_id_capture_mode_default_is_when_negotiated`
- [x] Enforced mode still reports `"enforced"` — unchanged, `test_mcp_session_id_capture_mode_enforced_when_env_set`
- [x] Replay surface returns a **non-zero** `excluded_null_session_count` for a tenant with header-less MCP rows, while `row_count == 0` for an unrelated session (empty forest ≠ empty history); non-MCP NULL rows and another tenant's rows are excluded — `tests/test_api_audit_routes.py::test_replay_reports_excluded_null_session_count`
- [x] Envelope-shape / exact-body assertions updated for the new field — `test_api_audit_routes.py`, `tests/acceptance/test_audit_replay_acceptance.py`
- [x] Full affected suites green: `test_api_audit_routes.py`, `test_api_v1_health.py`, `test_mcp_audit.py`, `test_mcp_server.py`, `test_audit_replay_acceptance.py`, `test_features.py`, `test_ui_audit_replay.py`, `test_settings.py`, `test_approval_queue.py` (enforced-mode lineage)
- [x] CLI OpenAPI snapshot + Go client regenerated (`make snapshot-openapi && make generate`) so the `cli-api-snapshot-freshness` CI drift gate stays green; `go build ./...` + `go test ./internal/cmd/audit/...` pass

**Option (a) — hard-reject session-less post-handshake calls (fail-loud, BREAKING).**

The reporter's first-choice ask: with the mode enforced, a `tools/call` / `resources/read` arriving with no negotiated session is refused with a JSON-RPC `-32600` **before dispatch** and writes **no** audit row, making the "every MCP row is replayable" guarantee real. `/status` would then report a session-required value (e.g. `"enforced"`), never `"always"`.

Why it is **not** shipped autonomously here:

- It is a **breaking change** to header-less `curl`/script callers — the report describes one service identity with 389/389 NULL rows over 8 days that would start failing. Migrating those callers to the handshake (or to supply a session id) is part of adopting this option.
- The reject **must be method-scoped**: `initialize` (and `ping` / `notifications/*`) must be exempt, because the handshake issues the session id in the first place. The existing enforced gate at `server.py:_bind_mcp_session_id` rejects **all** header-less methods including `initialize`, so it is chicken-and-egg-broken as-is and would need to become method-aware (and `mcp_dispatch` ordering adjusted so the reject can see the method) before this option is safe to enable.

Recommendation: needs a human/product call on whether to break header-less callers. If chosen, it is a separate PR that (1) makes the reject method-aware, (2) migrates the known header-less callers, and (3) relabels `/status` accordingly.

## Out of scope (adjacent findings)

- `features.py::_audit_replay_block` hardcodes `capture_mode: "enforced"` (stale pre-T6 constant; never wired to `mcp_session_id_capture_mode()`), so `/ready` and `/health` already disagree on `main`. Left untouched to stay surgical — #2700 is about the `/status` `"always"` label, which `/ready` never carried. A follow-up could derive `_audit_replay_block`'s `capture_mode` from the shared helper. `docs/RELEASING.md` (lines ~226/353) and `test_features.py` docstrings carry the same stale "flip to always" narrative.
- The MCP `meho_audit_replay` / `query_audit` (`shape="tree"`) tools and the UI replay route build their own `{root, session_id, tenant_id, row_count}` envelope and do **not** yet carry `excluded_null_session_count`. The issue's forensic surface is the REST `/audit/sessions/{id}/replay` route, which is what this PR covers; parity on the MCP/UI surfaces is a candidate follow-up.
- Pre-existing `code-quality` size warnings on `mcp/server.py` (1193-line file; `_bind_mcp_session_id` / `_dispatch_to_handler` / `mcp_dispatch` > 60 lines) — not introduced by this change.

Closes #2700

## Summary by CodeRabbit

* **New Features**
  * Audit replay responses now report MCP audit records excluded because they lack a negotiated session.
* **Bug Fixes**
  * Session capture reporting now correctly defaults to `when_negotiated`.
  * Header-less MCP requests remain accepted and are recorded without a session, while enforced mode continues requiring one.
* **Documentation**
  * Updated health, audit replay, and API documentation to describe session capture modes and excluded records.


damir-topic:




**Verdict: APPROVE**

**Anchor task:** #2700 — MCP `mcp_session_id_capture: "always"` label is a false forensic guarantee; un-negotiated (header-less) `method=MCP` rows land `agent_session_id = NULL` and are invisible to G8.2 session-lineage replay, making an empty replay forest indistinguishable from an empty history.
**Surface area:** backend (+ generated Go CLI client, docs)
**Head SHA reviewed:** `4b4fe68c98e6f323845d35350fb2a3d7609edd66`

### Grounded-contract verdict (object #2700)

- **`final_dod` satisfied — option (b) HONEST-LABEL (non-breaking).**
  - `mcp_session_id_capture_mode()` returns `"when_negotiated"` (not `"always"`) in default lenient mode; `"enforced"` unchanged. Wired to the real operator surface (`meho status` → `GET /api/v1/health`, `status.go:125`).
  - `GET /api/v1/audit/sessions/{id}/replay` now returns `excluded_null_session_count` alongside `row_count` — tenant-scoped tally of `method=MCP` rows with a NULL `agent_session_id`. An empty forest (`row_count=0`) with a non-zero excluded count is now distinguishable from an empty history.
  - Lenient accept-and-null preserved → header-less script callers are **not** broken.
- **`design_decision` honored.** Ships the non-breaking honest-label option; the breaking option (a) hard-reject `-32600` is surfaced in the PR body ("Alternative — needs human decision") with the method-scoping / chicken-and-egg caveat and caller-migration note — described, not shipped. Synthetic per-call ids explicitly rejected (decision #1147). Correct fork.
- **`suggested_test` added** — and stronger than the minimum:
  - `test_mcp_session_id_capture_default_when_negotiated` asserts `== "when_negotiated"` **and** `!= "always"`.
  - `test_replay_reports_excluded_null_session_count` seeds 3 header-less MCP NULL rows, asserts the count is exactly 3, and proves method-scoping (a `POST` NULL row is excluded) and tenant isolation (tenant B's MCP NULL row is excluded), with `row_count == 0` on an unrelated session.
- **Session integrity actually closed (not merely plausible).** Verified the excluded count is **not inert**: MCP audit rows are genuinely written with literal `method="MCP"` (`mcp/handlers.py`, `mcp/audit.py:25`), so the `method='MCP' AND agent_session_id IS NULL` filter matches the real un-negotiated rows the report describes. DISPATCH lineage children are correctly not double-counted; other-method NULLs and other tenants are correctly excluded (proven by test).

### Acceptance criteria
- [x] `/status` (`meho status` → `/api/v1/health`) reports `mcp_session_id_capture != "always"` (`"when_negotiated"`) — `test_api_v1_health.py`, `test_mcp_server.py`.
- [x] Replay surface returns non-zero `excluded_null_session_count` for a tenant with header-less MCP rows — `test_api_audit_routes.py::test_replay_reports_excluded_null_session_count`.
- [x] Non-breaking: enforced-mode reject gate untouched; lenient accept-and-null preserved.

### Predicted vs. actual diff
- ✅ `mcp/server.py` (label + docstring), `api/v1/health.py` (docstring), `api/v1/audit.py` (+`_count_excluded_null_session_rows`, returns field), `api/v1/audit_models.py` (field) — expected and correct.
- ✅ Tests updated/added; `CHANGELOG.md`; `docs/codebase/error-message-shape.md` (signal 11) + `docs/cross-repo/audit-replay.md`.
- ❓ `cli/api/openapi.json` + `cli/internal/api/client.gen.go` — regenerated for the additive field (freshness gate); justified. `go build ./...` passes.
- 🕳️ None missing.
- 🚫 None undisclosed.

### Integration verification
- [x] `ruff check` / `ruff format --check` (changed files) — clean.
- [x] `mypy api/v1/audit.py` — clean.
- [x] `pytest` — `test_api_audit_routes.py`, `test_api_v1_health.py`, `test_mcp_server.py`, `test_mcp_audit.py`, `acceptance/test_audit_replay_acceptance.py`, `test_features.py` → 145 passed, 19 skipped.
- [x] `go build ./...` — passes (regenerated client compiles).
- [⚠️ not run] full backend suite / CLI-snapshot-freshness gate — deferred to CI (currently in progress); the committed `openapi.json` and `client.gen.go` are internally consistent and build.

### Bot review
Copilot (`COMMENTED`) posted 4 inline comments, all the same point: docstrings say "the deploy's `/status`" but the route is `/api/v1/health`. **False positives** — `meho status` is a real CLI verb backed by `GET /api/v1/health` (`status.go:24,50,125`), so "/status" is the established operator-facing name (the grounded contract itself says "make /status honest"). The docs asserting "no /status route" refer to a different service (MEHO.Knowledge / docs-addon). No change required. CodeRabbit hit its review limit (no review produced).

### Docs
- [x] `docs/codebase/error-message-shape.md` (signal 11) and `docs/cross-repo/audit-replay.md` updated to the honest label + new field. `docs_status: updated`.

- `features.py::_audit_replay_block` hardcodes `capture_mode: "enforced"` (stale pre-T6 constant, never wired to `mcp_session_id_capture_mode()`), so `/ready` already disagreed with `/status` on `main` **before** this PR. Pre-existing; correctly disclosed and left out of scope. `final_dod` (b) does not require `/ready`. A follow-up should derive it from the shared helper and refresh the "flip to always" narrative in `docs/RELEASING.md` and `test_features.py` docstrings.
- The MCP `meho_audit_replay` / `query_audit(shape="tree")` tools and the UI replay route build their own `{root, session_id, tenant_id, row_count}` envelope and do not yet carry `excluded_null_session_count`. The DoD's forensic surface is the REST route, which is covered; parity is a candidate follow-up (disclosed in the PR body).
- `_count_excluded_null_session_rows` is a tenant-wide `COUNT` on `AuditLog` per replay call; on very large audit tables an index on `(tenant_id, method, agent_session_id)` may help. Low priority — replay is a low-frequency, `tenant_admin`-gated forensic action and mirrors the existing count-first guard's cost profile.

- Correct, disciplined read of the grounded fork: non-breaking option shipped, breaking option surfaced with the exact method-scoping caveat, synthetic-id anti-pattern explicitly ruled out.
- Scope is genuinely surgical; the one pre-existing inconsistency (`features.py`) is disclosed rather than silently bundled.
- The excluded-count test proves method-scoping and tenant isolation, not just the happy path.

_Note: the reviewing `gh` identity is the PR author, so a GitHub formal `--approve` is structurally blocked (self-review). Verdict of record is APPROVE in the machine-readable result file; this comment carries the human-readable trail._

Issue #1856, opened 2026-06-18
Task: generic redaction-safe params-echo default for proposed_effect previews


damir-topic:
Parent initiative: #1853
Parent goal: #800

## Problem

`operations/_preview.py::build_proposed_effect` returns `None` for any op_id without a registered builder, so the caller falls back to an **identifier-only** approval row (`{op_id, connector_id, target_id}`). Only k8s / vmware / argocd register builders — every other approval-gated op (keycloak, nsx, sddc, vcfa, …) shows the approver no params at all.

## Desired state

A **generic params-echo default**: when no bespoke builder is registered (and the op is not in `_SENSITIVE_CLASSES`), echo the requested params into `proposed_effect` — redaction-safe, reusing the existing credential-class suppression + redaction discipline — so every connector gets param-level legibility for free, with bespoke builders layered on where richer previews help.

## Acceptance criteria

- [ ] Parking an approval-gated op with no registered builder surfaces its requested params in `proposed_effect` rather than the bare identifier default.
- [ ] Secret-bearing params are scrubbed; credential-class ops (`_SENSITIVE_CLASSES`) remain suppressed (no regression of the existing suppression).
- [ ] Ops with a bespoke builder keep using it (no double-echo).
- [ ] Regression test for both the echo and the redaction.

## References

- `backend/src/meho_backplane/operations/_preview.py` (`build_proposed_effect`, `_SENSITIVE_CLASSES`). Migrated from evoila-bosnia/meho-internal #109.

PR #2769, opened 2026-08-03
fix(approvals): uniform op_class + proposed_effect envelope across argocd write ops

### LINKED ISSUE #2681: proposed_effect metadata inconsistent across argocd write ops: preview flag contradicts content, op_class drift, per-op envelope variance
First, credit where due: the preview *content* for argocd writes is at keycloak parity. `app.set` parks carry a full `before_spec`/`after_spec` snapshot (it caught a single-byte diff in our test), and `app.delete` on a nonexistent app honestly discloses its failed cascade snapshot (`preview_unavailable: true` + the verbatim `preview_error`). We explicitly did NOT file the "thin proposed_effect" finding for argocd — it's rich.

What's inconsistent is the metadata around it, three ways (all live-verified on v0.25.0 by parking refresh/set/delete and inspecting each payload):

1. `argocd.app.refresh` reports `preview_populated: true` while the served payload contains no `preview` key at all. `app.set` has flag true AND preview present; `app.delete` has flag false AND `preview_error` — refresh is the odd one out, claiming content the approver can't see.
2. `op_class` drifts across sibling writes: `app.delete` classifies `write`, but `app.set` and `app.refresh` classify `other`. Anything filtering by op_class (broadcast filters, audit queries, approval dashboards) sees two of three argocd writes under the wrong class. Related: the audit row for a denied approve attempt classifies `op_class: "read"` — the audit enum seems to lack an approval class.
3. The proposed_effect envelope field-set differs per op (`app.delete` embeds `op_id`/`target_id`/`connector_id` inside proposed_effect; the others don't) — consumers can't rely on one schema per connector.

Asks (binary):
- `preview_populated` is true iff a `preview` key is present in the served payload.
- Every `requires_approval` connector write op classifies `op_class: write` (or a documented approval class) — no `other` fallback.
- One proposed_effect envelope schema per connector.

Unrelated small ask spotted during the same campaign: `result_query` has no field projection — answering "which of 23 apps are OutOfSync?" means paging ~0.5 MB of full unprojected objects through the agent. A `fields`/projection argument would fix the recurring "is X in the list" shape.

Consumer signal: `approval-proposed-effect-metadata-inconsistent-argocd-writes` (evoila-bosnia/claude-rdc-hetzner-dc, campaign #1556 / #2039; probe parks all three ops, checks the assertions, self-rejects — queue left clean).

damir-topic:
## Summary

Fixes the three `proposed_effect` metadata inconsistencies #2681 reports across the argocd write ops — all three root-caused to the **shared** park-time envelope machinery, not the argocd connector:

- **op_class drift (claim 2).** `classify_op` keys purely off the op-id verb suffix. `.set` / `.sync` / `.rollback` / `.refresh` were not in `_WRITE_SUFFIXES`, so 4 of the 7 approval-gated argocd writes classified `other` while `.delete` / `appproject.create` / `appproject.update` classified `write`. Now all seven classify `write`.
- **Per-op envelope variance (claim 3).** The op-identity fields (`op_id` / `connector_id` / `target_id`) rode only the builder-declines/raises paths, so the same op parked against a nonexistent vs a live app produced a different field-set (runtime-state-dependent). `dispatcher._build_proposed_effect` now stamps the identity fields **and** `op_class` onto every envelope uniformly.
- **preview flag vs content (claim 1).** No code change: `preview_populated: true` for `argocd.app.refresh` is working-as-designed since #1856/#2332 — the generic `params_echo` is a legitimate populated content key, distinct from a computed `preview`. The reporter read the absent `preview` key as "content the approver can't see," but the content is under `params_echo`. The tests now encode the correct contract.

## Design decisions

**op_class locus — non-breaking option chosen.** The genuine choice was (a) extend `_WRITE_SUFFIXES` vs (b) make `classify_op` descriptor-aware. I took (a) — the smaller blast radius and the non-breaking option — plus one refinement:

- `.set` / `.sync` / `.rollback` are added to the global write-suffix set. They are unambiguous mutation verbs and no non-argocd op uses them today, matching the existing precedent (`.add`/`.remove` for bind9, `.assign` for keycloak, `.restart` for rke2 were all added the same way).
- `argocd.app.refresh` is pinned by **exact op-id** in a new `_WRITE_OPS` frozenset, *not* via a `.refresh` suffix — see the two callouts below. Same exact-pin idiom as the existing `_CHECK_EVENT_OPS` / `_CREDENTIAL_MINT_OPS`.

**(a) Render-time relabeling of stored op-ids.** `classify_op` re-runs at render time on stored op-ids in the console drawers (`ui/routes/broadcast/event.py`, `ui/routes/audit/routes.py`), so extending the suffix set relabels *historical* broadcast/audit rows for every connector whose op-ids end in the new suffixes — not just new rows. This is intended (the whole point is consistent classification), and the added suffixes are collision-free today (only argocd uses `.set`/`.sync`/`.rollback`). The label change is cosmetic for redaction: both `other` and `write` are full-detail, non-sensitive classes, so no payload that was previously redacted becomes exposed or vice-versa.

**(b) Is `.refresh` honestly `write`?** For `argocd.app.refresh`, yes — it is an HTTP GET (`?refresh=hard`) that forces ArgoCD to reconcile live-vs-desired state, i.e. an approval-gated action with a cluster-side side effect. On the approval/audit surface that is a write even though the HTTP verb is a read; I classified it `write` (the DoD's sanctioned policy call) rather than minting a new approval op-class, to keep the blast radius minimal. Crucially, `.refresh` is **not** honestly a write in general: `topology.refresh` is a genuine read that emits only aggregate node/edge counts and explicitly stores `op_class="read"`. A blanket `.refresh` suffix would relabel every historical `topology.refresh` row as `write` at render time — so `.refresh` is pinned to the one op that earns it instead of going global.

The **breaking** alternative for the op_class locus is making `classify_op` **descriptor-aware** — classifying any `requires_approval=True` op as `write` (or a new documented approval op-class). It is semantically cleaner (no per-suffix maintenance, no `.refresh` special-case) but:

- changes the `classify_op(op_id: str)` signature to take a descriptor, rippling to ~9 call sites (broadcast, audit, preview, approval_queue, the two UI render paths, checks/broadcast, …);
- **breaks at render time**, where only a stored op-id string exists and no descriptor is available — forcing a migration to persist `op_class` on the row and a fallback classifier anyway.

That is a larger, breaking change with a data-migration tail; it should be a deliberate human call, so this PR ships the non-breaking suffix/pin approach. If a dedicated `approval` op-class for reconcile-shaped GET verbs is wanted (so `.refresh` isn't labelled a literal `write`), that is the natural place to introduce it.

## Test plan

- [x] `ruff check` — clean on changed files
- [x] `mypy src/` — clean on changed files (one pre-existing `net/icmp.py` `MSG_ERRQUEUE` error is a Linux-only socket constant mypy flags on macOS; passes on CI's Linux runner)
- [x] `pytest tests/test_connectors_argocd_write_e2e.py` — 20 passed, incl. the new parametrised park-envelope test over all 7 write ops (asserts `op_class == "write"`, the `preview_populated` content-key contract, and an identical op-identity/metadata field-set) and the nonexistent-app delete case (the identifier-leak branch)
- [x] Full unit suite (`pytest -n 3 --ignore=tests/integration --ignore=tests/migrations`) — green, incl. the updated dispatcher / vault-kv / vmware-composite park tests that had locked the old per-path envelope shape
- [x] Per the DoD, does **not** assert `set(effect.keys())` equality across ops — the `preview` XOR `params_echo` content-key split is intended

## Docs

- `docs/codebase/approvals.md` — new "Uniform op-identity envelope (#2681)" section documenting the one-schema-per-connector contract and the classifier change.

## Out of scope

- `result_query` field projection (unrelated small ask in the issue body — separate feature).
- The reporter's "denied approve attempt classifies `op_class: read`" note: the approve/reject audit bindings already bind `write` (`api/v1/approvals.py`, `mcp/tools/approvals.py`); the read-classifying path could not be reproduced. Left untouched.

Closes #2681

---

## Follow-up (auto-implement-issue, iteration 2, 2026-08-03)

Addressed review finding from `.claude/auto/review-results/pr-2769-iter-1.json` (REQUEST_CHANGES):

- **B1** (blocker) `5d6d0b28` — `.set` / `.sync` / `.rollback` were added to `_WRITE_SUFFIXES`, but `_MEHO_FLAT_WRITE_SUFFIXES` is auto-derived from it (`.` → `_`), so the flat MCP-tool arm gained `_set` / `_sync` / `_rollback` and reclassified `meho_broadcast_overrides_set` from `other` to `write` — turning `test_broadcast_events.py::TestClassifyOp::test_meho_broadcast_prefix_is_not_audit_query` red and relabelling that op's historical rows at render time. Fixed via the review's preferred shape: `argocd.app.set` / `.sync` / `.rollback` are now exact-pinned in `_WRITE_OPS` alongside the already-pinned `.refresh`, and removed from `_WRITE_SUFFIXES`. ArgoCD op-ids are always dotted, never `meho_`-flat, so exact-pinning classifies all seven argocd writes `write` with zero flat-mirror bleed. `docs/codebase/approvals.md` updated to describe the exact-pin rationale.

Verified: `ruff check` / `ruff format --check` / `mypy` clean; `test_broadcast_events.py` + `test_connectors_argocd_write_e2e.py` + dispatcher / vault-kv / vmware-composite preview + classifier-coverage suites 199 passed / 4 skipped (integration guards); code-quality `--diff` gate clean.

### DIFF
--- a/backend/src/meho_backplane/broadcast/events.py
+++ b/backend/src/meho_backplane/broadcast/events.py
@@ -260,39 +260,47 @@ _WRITE_SUFFIXES: Final[tuple[str, ...]] = (
     # unclassified event. Classifying it ``write`` keeps the mutation signal
     # accurate on the feed.
     ".restart",
-    # ArgoCD write verbs (#2681). ``argocd.app.set`` (PUT the spec),
-    # ``argocd.app.sync`` (POST a sync) and ``argocd.app.rollback`` (POST a
-    # rollback) are approval-gated mutations that fell through to ``other``
-    # before this — so op_class-filtered broadcast/audit/approval surfaces
-    # saw them under the wrong class while their ``.delete`` /
-    # ``appproject.create`` / ``appproject.update`` siblings classified
-    # ``write``. All three verb spellings are unambiguous mutations (no
-    # non-argocd op uses them today), so the global suffix is safe. The
-    # fourth argocd write verb, ``.refresh``, is deliberately NOT here: it
-    # is an HTTP-GET-shaped verb and a blanket ``.refresh`` suffix would
-    # also relabel ``topology.refresh`` (a read that emits only aggregate
-    # counts) as ``write`` at render time. It is pinned individually in
-    # :data:`_WRITE_OPS` instead.
-    ".set",
-    ".sync",
-    ".rollback",
 )
 
 #: Exact-match op-ids that classify ``write`` despite not carrying a
-#: write-shaped verb suffix. ``argocd.app.refresh`` is an HTTP GET
-#: (``GET /api/v1/applications/{name}?refresh=hard``) that forces ArgoCD to
-#: reconcile the live vs. desired state — an approval-gated action with a
-#: cluster-side side effect, so on the approval/audit surface it is a write
-#: even though the HTTP verb is a read. It is pinned here rather than via a
-#: ``.refresh`` suffix in :data:`_WRITE_SUFFIXES` because that suffix is
-#: shared: ``topology.refresh`` is a genuine read (it emits only aggregate
-#: node/edge counts, explicitly stored ``op_class="read"``), and because
-#: :func:`classify_op` re-runs at render time on stored op-ids, a blanket
-#: suffix would silently relabel every historical ``topology.refresh`` row
-#: as ``write`` in the console drawers. Same exact-pin idiom as
-#: :data:`_CHECK_EVENT_OPS` / :data:`_CREDENTIAL_MINT_OPS`: a precise pin
-#: that beats a broader structural match.
-_WRITE_OPS: Final[frozenset[str]] = frozenset({"argocd.app.refresh"})
+#: write-shaped verb suffix. The four ArgoCD write verbs (#2681) are pinned
+#: here rather than added to :data:`_WRITE_SUFFIXES` because both spellings
+#: that a suffix would need — dotted and its underscore mirror — collide:
+#:
+#: * ``argocd.app.set`` / ``.sync`` / ``.rollback`` are approval-gated
+#:   mutations that fell through to ``other`` (their ``.delete`` /
+#:   ``appproject.create`` / ``appproject.update`` siblings already
+#:   classified ``write`` via existing suffixes). A ``.set`` / ``.sync`` /
+#:   ``.rollback`` suffix would work for the dotted argocd op-ids, but
+#:   :data:`_MEHO_FLAT_WRITE_SUFFIXES` is auto-derived from
+#:   :data:`_WRITE_SUFFIXES` by ``.`` → ``_``, so it would ALSO add
+#:   ``_set`` / ``_sync`` / ``_rollback`` to the flat MCP-tool arm and
+#:   silently relabel ``meho_``-prefixed ops such as
+#:   ``meho_broadcast_overrides_set`` from ``other`` to ``write``. ArgoCD
+#:   op-ids are always dotted (``argocd.app.set``), never ``meho_``-flat,
+#:   so an exact pin classifies them ``write`` with zero flat-mirror bleed.
+#: * ``argocd.app.refresh`` is an HTTP GET
+#:   (``GET /api/v1/applications/{name}?refresh=hard``) that forces ArgoCD
+#:   to reconcile the live vs. desired state — an approval-gated action with
+#:   a cluster-side side effect, so on the approval/audit surface it is a
+#:   write even though the HTTP verb is a read. A ``.refresh`` suffix is
+#:   shared with ``topology.refresh``, a genuine read (it emits only
+#:   aggregate node/edge counts, explicitly stored ``op_class="read"``), and
+#:   because :func:`classify_op` re-runs at render time on stored op-ids a
+#:   blanket suffix would silently relabel every historical
+#:   ``topology.refresh`` row as ``write`` in the console drawers.
+#:
+#: Same exact-pin idiom as :data:`_CHECK_EVENT_OPS` /
+#: :data:`_CREDENTIAL_MINT_OPS`: a precise pin that beats a broader
+#: structural match.
+_WRITE_OPS: Final[frozenset[str]] = frozenset(
+    {
+        "argocd.app.set",
+        "argocd.app.sync",
+        "argocd.app.rollback",
+        "argocd.app.refresh",
+    }
+)
 
 #: Op-id suffixes that imply non-mutating read. ``.ls`` and ``.about``
 #: are the CLI-shaped verbs (``meho vsphere ls``, ``meho meho about``)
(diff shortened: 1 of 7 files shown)


damir-topic (commented):




**Anchor task:** #2681 — uniform `op_class` + `proposed_effect` envelope across the seven approval-gated argocd write ops
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `5d6d0b28`

### Acceptance criteria (grounded DoD #2681)
- [x] **(a) `op_class == "write"` for all seven argocd writes** — `sync` / `rollback` / `set` / `refresh` were `other`; now `write` via the exact-pin `_WRITE_OPS` frozenset in `broadcast/events.py`. Verified: `test_park_envelope_uniform_across_write_ops` asserts `write` for all 7; doctests updated and green.
- [x] **(b) `preview_populated` defined against the real content key** — True iff `preview` OR `params_echo` present, False + reason when neither, False on `preview_unavailable`. `describe_preview_provenance` (`_preview.py:504-516`) reads the raw hook output (not the merged base). #1856/#2332 `params_echo` semantics preserved (`test_park_no_builder_ops_use_generic_params_echo` green).
- [x] **(c) One uniform op-identity + metadata field-set across all seven** — `_build_proposed_effect` now `setdefault`-stamps `op_id`/`connector_id`/`target_id` + `op_class` onto every non-None envelope, before which they rode only the declines/raises path. Verified by the identity/meta field-set equality across all 7 ops plus the nonexistent-app delete case (the identifier-leak branch).

### Design decision (honored)
Non-breaking option taken. The DoD framed (a) extend `_WRITE_SUFFIXES` vs (b) descriptor-aware `classify_op`. Iteration 2 ships a refinement of (a): exact-pin the four dotted argocd verbs in `_WRITE_OPS` (no global suffix), which is strictly more surgical — it relabels only those four exact op-ids, no other connector, and avoids the flat-mirror bleed that failed iteration 1. The breaking alternative (descriptor-aware `classify_op` with a render-time migration tail) is surfaced in the PR body under "Alternative (needs human decision)" and deliberately not shipped.

### Iteration-1 blocker resolution
- **B1 (iter-1 blocker)** — adding `.set`/`.sync`/`.rollback` to `_WRITE_SUFFIXES` bled into the auto-derived `_MEHO_FLAT_WRITE_SUFFIXES` (`.`→`_`), reclassifying `meho_broadcast_overrides_set` `other`→`write`. **Resolved:** the three verbs moved to the exact-pin `_WRITE_OPS` (consulted only on the dotted arm, not the `meho_` flat arm) and removed from `_WRITE_SUFFIXES`. Verified green: `test_broadcast_events.py::…::test_meho_broadcast_prefix_is_not_audit_query` (asserts `meho_broadcast_overrides_set == "other"`).

### Predicted vs. actual diff
- ✅ `broadcast/events.py` — `_WRITE_OPS` pin + `classify_op` tail + doctests
- ✅ `operations/dispatcher.py` — uniform `setdefault` stamping in `_build_proposed_effect`
- ✅ `tests/test_connectors_argocd_write_e2e.py` — the two new #2681 tests + the `:539`→`write` update
- ✅ `tests/{test_operations_dispatcher,test_connectors_vault_kv_write_preview,test_connectors_vmware_rest_composites_write_preview}.py` — updated to the uniform-identity envelope
- ✅ `docs/codebase/approvals.md` — new "Uniform op-identity envelope (#2681)" section
- 🕳️ Missing: none

### Integration verification (ran in a venv on the identical head SHA)
- [x] `ruff check` (changed src) — clean
- [x] `ruff format --check` (6 changed files) — clean
- [x] `mypy` (both changed src files) — clean
- [x] `pytest` argocd e2e + broadcast classifier + vault/vmware preview — 145 passed
- [x] `pytest` dispatcher park/preview — 5 passed
- [x] `pytest` broadcast classifier coverage — 7 passed
- [x] `classify_op` doctests — 1 passed
- [x] migrations replay — n/a (no schema change; `proposed_effect` is JSON, additive keys)

### Findings

- `tests/test_connectors_argocd_write_e2e.py:653-654` — the comment "`#2681 added `.set` to the write-suffix set`" is stale after the iteration-2 pivot: `.set` is now exact-pinned in `_WRITE_OPS`, not in the write-suffix set. Cosmetic; worth a one-line comment fix in a future touch. (Also flagged by Copilot.)
- Copilot suggested pinning `topology.refresh` in a new `_READ_OPS` set so `classify_op` returns `read` (it returns `other` today). Pre-existing behavior, orthogonal to #2681 — a reasonable follow-up, out of scope here.

### Scope note — #2699/#2700 audit/session integrity
Out of scope for this PR and untouched by it. #2769 is scoped to #2681 (`op_class` classification + `proposed_effect` envelope shape). It does not modify `AuditLog.operator_sub` / `BroadcastEvent.principal_sub` (`operations/_audit.py` is not in the diff), sensor `identity_sub`, or MCP session capture — the surfaces #2699/#2700 concern. The only audit-adjacent effect is the `op_class` label (a display/redaction class; both `other` and `write` are full-detail, non-sensitive), so no previously-redacted payload becomes exposed. No cross-contamination, no regression.

### Docs
- [x] `docs/codebase/approvals.md` — updated (new #2681 section), accurate against the implementation, no .NET analogies.

### What worked well
- The iteration-1 blocker was fixed at exactly the right altitude: exact-pin on the dotted arm keeps every non-argocd op (and the `meho_` flat mirror) untouched, which is a cleaner outcome than the suffix approach the DoD offered.
- Test correctly follows `final_dod` over the grounding's stale `suggested_test` phrasing: it asserts the identity/meta field-set is uniform, and explicitly does NOT assert `set(effect.keys())` equality (which would wrongly fail on the intended `preview` XOR `params_echo` split).

Issue #2644, opened 2026-07-19
Task: strip top-level JSON-Schema unions from published tool schemas — broadcast.watch anyOf 400s every Anthropic agent run at model-init


damir-topic:
Parent initiative: #2656 (G0.37 v0.25.0 ops-feedback hardening)
Parent goal: #221

## Summary

Every `meho agents run` on provider `anthropic` dies at model-init with `status_code: 400 … tools.5.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level` (`turns: 0`), independent of the agent's own toolset — reproduced with an empty toolset `{}`. The offending tool is `meho.broadcast.watch`: its MCP `inputSchema` carries a top-level `anyOf` enforcing the `cursor`/`since_cursor` XOR, and the hosted-agent bridge deep-copies the **registered** schema verbatim into the Anthropic tool list — bypassing `ToolDefinition.to_wire()`, which already strips root-level combinators for exactly this reason. Route the bridge through the wire shape and add a sweep test over *published* schemas so no bridged tool can regress.

## Problem

- `backend/src/meho_backplane/mcp/registry.py:187,256` — `_wire_safe_input_schema` already strips top-level `oneOf`/`allOf`/`anyOf`, and `ToolDefinition.to_wire()` applies it. The MCP wire surface is therefore already correct; only the agent bridge is not.
- `backend/src/meho_backplane/agent/toolset_broadcast.py:82` — `_mcp_broadcast_tool` returns `copy.deepcopy(definition.inputSchema)`, i.e. the **registered** schema, bypassing `to_wire()`. The bridge hands it to `Tool.from_schema` (`agent/toolset.py:362-368`), and pydantic-ai forwards it to the Anthropic Messages API, which requires each tool's `input_schema` to be a plain object schema at the root.
- `backend/src/meho_backplane/mcp/tools/broadcast.py:1310` — `meho.broadcast.watch`'s registered `inputSchema` ends with a top-level `"anyOf": [{"required": ["cursor"]}, {"required": ["since_cursor"]}]`. This is **not** belt-and-braces: `mcp/handlers.py:337-339` runs `jsonschema.validate(..., schema=defn.inputSchema)`, so the registered union is the server-side validation layer #910 deliberately added. Deleting it would narrow validation for all 8 combinator-carrying tools.
- Default tool order is the 3 operation meta-tools + announce/recent/watch → `watch` is exactly the `tools.5` in the error.
[...]

PR #1031, opened 2026-05-24
feat(agent-runtime): G11.1-T6 agent_run record + enforced lifecycle + cancellation

damir-topic:
## Summary

- Adds the **`agent_run`** record (table + migration `0015` + ORM model + lifecycle service) — one durable row per in-process LLM-agent invocation (G11.1-T6, Initiative #802, Goal #800).
- The run `id` doubles as the **`agent_session_id` lineage key** that G11.4/C2 binds into per-tool-call audit rows (same shape as `audit_log.agent_session_id`, migration `0014`).
- The `status` lifecycle is an **explicit, enforced state machine**: `ALLOWED_TRANSITIONS` is the single source of truth and `transition()` rejects any illegal edge *before* any DB write (a DB `CHECK` only constrains the value set, not the edges).
- A `pending` / `running` / `awaiting_approval` run is **cancellable by an authorized operator** (`TenantRole.OPERATOR`+); cancellation records durable intent (`status` + `ended_at`) so it survives a process restart — the in-flight loop interruption is left to the AgentRun seam (G11.1-T1).

## Design notes

- `agent_definition_id` (table lands in parallel in #809) and the self-referential `parent_run_id` are **soft-FKs** (no clause) — same discipline as `audit_log.target_id` / `parent_audit_id`; `tenant_id` is a real FK (clean-slate substrate).
- `AgentRunStatus` / `AgentRunTrigger` are closed `StrEnum`s backed by DB `CHECK` constraints that move in lock-step (drift guards assert model enum == model tuple == migration literal tuple).
- `cost` is a `Numeric(12,6)` stub recorded NULL until G11.5/C3; `succeed_run` already accepts a `cost` arg so C3 lands without a signature change.
- New `docs/codebase/agent-run.md` documents the model, state machine, and scope boundaries.

## Test plan

- [x] `ruff check` — clean
- [x] `ruff format --check` — clean
- [x] `mypy src/meho_backplane/` — Success, no issues in 271 files
- [x] `pytest tests/test_db_agent_run.py tests/test_migration_0015_agent_run.py tests/test_agent_run_lifecycle.py` — 62 passed
- [x] Lifecycle enforced — every legal edge accepted, every illegal edge raises `IllegalTransitionError` with no persisted change (parametrized over the full status×status matrix)
- [x] Inspectable + cancellable while running — authorized cancel across pending/running/awaiting_approval; read-only rejected (403-class), already-terminal rejected (409-class), missing id (404-class)
[...]

PR #2818, opened 2026-08-05
fix(agents): persist run turns counter at finalize (#2743)

### LINKED ISSUE #2743: Task: agents — persist the run turns counter (increment_turns has no caller; run_status reads 0 on every run, colliding with the #2644 outage fingerprint)
Parent goal: #221

G0.39 cycle seed — filed from v0.26.0/v0.27.0 ops feedback; attach to the G0.39 initiative when that cycle is opened (predecessors: #2761 G0.38, #2656 G0.37).

## Reporter's evidence (original report, verbatim)

> Small one, but it cost us a confused half hour while verifying the #2644 fix on v0.26.0.
>
> A one-shot agent run with an empty toolset that completes successfully — `status: succeeded`, real model output in the result — reports `turns: 0` in `meho.agents.run_status`. That's presumably because the field counts tool-use turns, and a no-tool run has none. The problem is that `turns: 0` was exactly the fingerprint of the pre-#2644 model-init failure ("400 invalid_request_error ... turns: 0"), so any consumer that keys on "did the run reach turn 1" — which our verification probe contract did, and which seems like a natural reading — concludes a perfectly healthy run never started.
>
> Suggested fix, either works for us: count a completed model reply as a turn (a succeeded run should never read 0), or document the field as tool-turns-only and expose an explicit reached-model discriminator. `status` plus the presence of output already disambiguates in practice, but the field name invites the wrong read.
>
> Repro: create any agent with `toolset: {}`, run it with a trivial prompt, check `run_status`. We track this as `agents-run-status-turns-zero-on-successful-no-tool-run`.

## Context (triage grounding, 2026-08-03)

The reporter's diagnosis is charitable — the reality is worse: **the `turns` counter is never incremented for any run, tool-using or not.** It is a dead column that has read 0 since it shipped:

- `agent_run.turns` — Integer NOT NULL default 0, documented as "The count of tool-use turns the loop has executed. The runtime increments it per turn" (`backend/src/meho_backplane/db/models.py:3394-3397`, column at `:3524`). That docstring describes intent, not reality.
- The single write site is `increment_turns` (`backend/src/meho_backplane/operations/agent_run.py:794-804`, `row.turns += 1`), shipped with the lifecycle service in #1031 (G11.1-T6). **It has zero production callers** — `grep -rn increment_turns backend/src` matches only its definition and `__all__` entry; the only callers are its own unit tests (`backend/tests/test_agent_run_lifecycle.py:503-509`).
- The runtime already counts what's needed and drops it at the finalize seam: `AgentRunResult` lifts `request_count` (model-request turns) and `tool_call_count` from the framework's usage accounting "so the T6 run record ... [has] the turn + tool totals" (`backend/src/meho_backplane/agent/run.py:530-556`) — but the invoke path builds only `TokenUsage` from the result and calls `_finalize_run` without ever writing turns (`backend/src/meho_backplane/agent/invocation.py:816-828`). The SSE path (`stream_events`, `invocation.py:1529-1541`) likewise consumes only FINAL/ERROR events, never TURN events (`AgentRunEventKind.TURN`, `run.py:568,579`).
- Every surface then faithfully reports the constant 0: MCP `meho_agents_run_status` (`backend/src/meho_backplane/mcp/tools/agent_runs.py:270`), REST (`backend/src/meho_backplane/api/v1/agent_runs.py:261,407`), `/ui` (`backend/src/meho_backplane/ui/routes/agents/runs/views.py:142,162`).

So the collision the reporter hit — healthy run indistinguishable from the pre-#2644 model-init-failure fingerprint (`turns: 0`) — applies to **all** runs, and any consumer probe keying on "reached turn 1" mis-verdicts everything.

## Change

Adopt the reporter's first option, which the code is already one seam away from: count model-request turns.

1. At run finalize, persist the loop's turn total onto the row — either write `result.request_count` (and consider `tool_call_count` alongside; the column exists for turns, a `tool_calls` column is optional and NOT required by this task) in `_finalize_run`'s success path, or wire `increment_turns` per `AgentRunEventKind.TURN` event. The finalize-write is the smaller diff and covers the non-streamed path; the streamed path (`stream_events`) must also end with a correct total.
2. Semantics: `turns` = model requests made by the loop. A succeeded run therefore always reports `turns >= 1`, restoring the "turn 0 means the model was never reached" discriminator the #2644 lineage consumers rely on.
3. Fix the drifted docs: the column docstring (`db/models.py:3394-3397` — "runtime increments it per turn" becomes true; clarify model-request semantics) and the `meho_agents_run_status` tool description (`mcp/tools/agent_runs.py:285-299`) so the field's meaning is stated on the wire surface agents actually read.

## Acceptance criteria

- [ ] Reporter's repro inverted: an agent with `toolset: {}` run to `status: succeeded` reports `turns >= 1` via `meho_agents_run_status` (test at the invocation-service level, not a hand-minted double).
- [ ] A tool-using run reports `turns` equal to the framework's `request_count` for the run (assert against `AgentRunResult.request_count` in the runtime seam test).
- [ ] A run that fails at model-init (the #2644 class) still reports `turns: 0` — the outage fingerprint stays meaningful.
- [ ] `grep -rn increment_turns backend/src` shows a production caller, or the helper is removed in favor of the finalize-write (no second dead path left behind).
- [ ] Column + tool-description docs state the model-request-turn semantics.

## Out of scope

- A separate `tool_calls` counter column / new wire field (nothing consumes it today; file separately if a consumer asks).
- Backfilling `turns` on historical run rows.
- The #2077 zero-tool-call hallucination guard (closed, different mechanism).

## References

- Reporter's consumer-side tracking id: `agents-run-status-turns-zero-on-successful-no-tool-run`.
- Dead counter: `backend/src/meho_backplane/operations/agent_run.py:794-804` (shipped #1031, never wired); column docs `backend/src/meho_backplane/db/models.py:3394-3397,3524`.
- Available totals: `backend/src/meho_backplane/agent/run.py:530-556` (`AgentRunResult.request_count` / `tool_call_count`); finalize seam `backend/src/meho_backplane/agent/invocation.py:816-828`.
- Read surfaces: `mcp/tools/agent_runs.py:270`, `api/v1/agent_runs.py:261,407`, `ui/routes/agents/runs/views.py:142,162`.
- Lineage: #2644 (model-init 400 outage whose fingerprint was `turns: 0`), #1031 (lifecycle service), #2077 (adjacent zero-tool-call finding, closed).

damir-topic:
## Summary

- The `agent_run.turns` column shipped with the lifecycle service (#1031) but was **never wired**: `increment_turns` had zero production callers, so `run_status` reported a constant `0` for every run — tool-using or not.
- That collided with the pre-#2644 model-init-failure fingerprint (`turns: 0`), making a healthy run indistinguishable from an outage to any consumer keying on "did the run reach turn 1" (our verification probe did).
- Fix: persist the loop's **model-request turn total** at run finalize. The non-streamed path writes `AgentRunResult.request_count`; the streamed path tallies `TURN` events (one per model-request node → the same count). Both go through `succeed_run`, so `turns` lands atomically with the terminal transition.
- Written **only on the success path**, so a run that fails before reaching the model (the #2644 class) keeps `turns == 0` and the outage fingerprint stays meaningful. A succeeded run always made ≥1 model request → `turns >= 1`.
- The now-superseded `increment_turns` helper is removed (no second dead path). Column + `meho_agents_run_status` docs now state the model-request-turn semantics.

## Test plan

- [x] `mypy` — clean (4 changed source modules)
- [x] `pytest` — 277 passed across lifecycle / invocation / runtime-seam / MCP / REST + UI read surfaces / consumption / scheduler; **0** failures
- [x] `code-quality.py --diff` — 0 blockers (remaining warnings are pre-existing legacy file/function sizes)

Acceptance criteria:

- [x] **Reporter's repro inverted** — `toolset: {}` run to `succeeded` reports `turns >= 1` via `meho_agents_run_status` (real `AgentInvoker` over a `FunctionModel`, not a hand-minted double): `test_run_status_reports_turns_for_no_tool_run`, `test_no_tool_success_records_one_turn`.
- [x] **Tool-using run** — `turns == AgentRunResult.request_count`, asserted against the runtime seam's result: `test_tool_using_run_persists_turns_equal_to_request_count`.
- [x] **Model-init failure** — still reports `turns: 0`: `test_model_init_failure_keeps_turns_zero`.
- [x] **No dead path** — `grep -rn increment_turns backend/src` is empty; the helper is removed in favor of the finalize-write. Lifecycle coverage moved to `test_succeed_run_records_turn_count` / `..._without_turns_leaves_counter_at_default`.
- [x] **Docs** — `agent_run.turns` column docstring and the `meho_agents_run_status` tool description state the model-request-turn semantics; `docs/codebase/agent-run.md` + `agent-runtime.md` updated.
- [x] Streamed (SSE) path also ends with a correct total: `test_streamed_run_persists_turn_total`.

## Out of scope

- **Child-run turns.** The `_finalize_child_run` / `ChildRunFinalizer` protocol does not carry the loop's `request_count`, so agent-invoked child rows keep `turns == 0`. Wiring that needs a protocol-signature change and is not among the issue's cited seams — file separately if a consumer needs child-run turns.
- A separate `tool_calls` counter column / new wire field (nothing consumes it today).
- Backfilling `turns` on historical run rows.

Closes #2743

## Summary by CodeRabbit

* **New Features**
  * Agent run status now reports the number of model-request turns.
  * Successful runs record at least one turn; streamed runs report their total turn count.
  * Runs that fail during model initialization retain a turn count of zero.

* **Documentation**
  * Updated agent run lifecycle and runtime documentation to clarify turn-count behavior.


backend/src/meho_backplane/mcp/tools/agent_runs.py
```diff
@@ -288,7 +288,11 @@ async def _run_status_handler(
             "#802). Operator-level. Returns {run_id, status, turns, "
             "provider, model, output, error, agent_definition_id, "
             "agent_name}; output/error are set once the run reaches a "
-            "terminal state; agent_definition_id / agent_name are null for "
+            "terminal state. `turns` is the number of model-request turns "
+            "the loop made: a succeeded run always reports `turns >= 1`, "
+            "so `turns: 0` on a failed run means the model was never "
```
damir-topic: **[🛑 B1]** — `run_status` description makes a false claim about `turns: 0` on failed runs

**What:** The description says a succeeded run reports `turns >= 1`, "so `turns: 0` on a failed run means the model was never reached." But `turns` is persisted only on the success path — `_finalize_run` routes failures to `fail_run`, which never writes `turns`. A run that fails *after* one or more model requests (budget exhaustion, a tool raising, a mid-stream model error) also reports `turns: 0`. The streamed path even counts `TURN` events and then discards the tally on the failure branch.

**Why it matters:** This is the agent-facing wire surface, and this PR's whole purpose is to stop `turns: 0` from being misread. Claiming `turns: 0` proves "the model was never reached" reintroduces that exact misread for the failed-run case — an operator/agent would wrongly conclude a budget-exhausted run never called the model.

**Required change:** State the real guarantee instead: a *succeeded* run always reports `turns >= 1`; `turns` is recorded only on success, so any non-succeeded run reports `0` regardless of how many model requests it made (the #2644 model-init failure is one such case). Apply the same correction to the `AgentRun.turns` column docstring in `db/models.py`, which carries the same implication.

damir-topic:
> **Note:** Formal `REQUEST_CHANGES` review-state can't be applied here — GitHub blocks reviewing your own PR and the automation shares the author identity (review-App not yet provisioned). This comment + the inline finding are the review of record; machine-readable verdict in the result file.

**Anchor task:** #2743 — persist the run turns counter (`increment_turns` had no caller; `run_status` read 0 on every run)
**Surface area:** backend
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `233fce1b`

### Acceptance criteria
- [x] `toolset:{}` succeeded run reports `turns >= 1` via `meho_agents_run_status` — verified (`test_run_status_reports_turns_for_no_tool_run`: real `AgentInvoker` + `FunctionModel` through the MCP dispatcher)
- [x] tool-using run reports `turns == AgentRunResult.request_count` — verified (`test_tool_using_run_persists_turns_equal_to_request_count` asserts against the runtime seam's result, not a constant)
- [x] model-init failure still reports `turns: 0` — verified (`test_model_init_failure_keeps_turns_zero`)
- [x] `increment_turns` removed, no dead path — verified (`grep -rn increment_turns backend/src backend/tests` empty; lifecycle coverage moved to `succeed_run` turns tests)
- [❌] column + tool-description docs state the semantics — docs updated, **but** the tool description + column docstring overclaim the failed-run `turns: 0` case (**B1**)

### Project-board alignment
|---|---|---|---|---|---|
| #2743 | Open (bug/task) | Closes | ✅ | full | close on merge |

### Predicted vs. actual diff
- ✅ `succeed_run` gains `turns`; `_finalize_run` threads it; non-streamed writes `request_count`; streamed tallies `TURN` events; `increment_turns` removed; column + tool docs; tests across all ACs
- ✅ read surfaces (REST / UI / `poll`) need no change — they read `row.turns`, now populated (regression tests green)
- 🚫 none undisclosed
- 🕳️ none missing

### Integration verification (clean-room in review worktree @ `233fce1b`)
- [x] `ruff check` — clean
- [x] `mypy` — clean
- [x] `pytest` — 188 passed (lifecycle / invocation / runtime-seam / mcp / REST + UI read surfaces / consumption)
- [x] `code-quality.py --diff` — 0 blockers (remaining warnings are pre-existing legacy file/function sizes)
- [n/a] migrations — no schema change (`turns` column pre-exists); CI migration-detect job green

- 🛑 **B1** — `run_status` tool description + `AgentRun.turns` column docstring claim `turns: 0` proves "the model was never reached"; that is false for runs that fail *after* model requests (`turns` is written only on success). Required change: reword to "a succeeded run always reports `turns >= 1`; `turns` is recorded only on success, so any non-succeeded run reports `0` regardless of requests made (model-init failure is one such case)."

- **Pre-existing (out of scope):** `stream_events`' `finally` finalizes with `output=terminal_output, error=terminal_error` — an SSE stream that ends without a `FINAL`/`ERROR` event (e.g. client disconnect) is recorded `succeeded` with empty output. This predates this PR — the base branch finalizes identically at the same seam; this PR only adds the `turns` tally to that existing call. Recommend a **separate ticket** to finalize a terminal-event-less stream as `failed`/`cancelled`. Not blocking here — this diff introduces no new behavior class.
- Child agent-runs keep `turns == 0`: `_finalize_child_run` / `ChildRunFinalizer` protocol doesn't carry `request_count`. Correctly out of scope per the issue's cited seams (invocation.py:816-828 + 1529-1541 are the top-level paths); file separately if a consumer needs child-run turns.

### Docs
- [❌] `docs/codebase/agent-run.md` + `agent-runtime.md` updated; the wire-surface tool description + column docstring need the B1 accuracy fix.

- AC2's test asserts the persisted `turns` against the actual `AgentRunResult.request_count` from the runtime seam — not a hand-picked constant.
- Both finalize paths converge on the same semantic (model requests); the streamed path discards its tally on the failure branch, so streamed/non-streamed failed runs stay consistent (both `0`).
- `increment_turns` fully removed — no second dead path left behind, exactly per the issue's AC4.

Issue #2763, opened 2026-08-03
checks: sensor evaluation loop stalled fleet-wide for 37 minutes, no self-detection


damir-topic:
Parent initiative: #2780

## Ops report (verbatim, v0.26.0)

Running v0.26.0 with a fleet of 24 interval sensors (net.tcp_check, 300s cadence) evaluated by the background check runner. This morning the whole evaluation loop went quiet for ~37 minutes and nothing surfaced it.

Timeline, reconstructed from the audit ledger (DISPATCH rows, principal `__sensor__` — which turned out to be a great forensic surface for this, by the way):

- batches of 24 `net.tcp_check` rows land every ~5 minutes up to 10:27:49Z
- then nothing at all until 11:04Z — about 7 consecutive missed cadence windows, across every sensor
- at 11:04 it resumes by itself; no restart, no operator action (we did roll the deployment at 11:11 for an unrelated values change, i.e. *after* recovery)

While it was stalled the dashboards did the right thing per the rollup rules — every member read `raw=unknown` once past `next_fire_at + 60s` grace — but that's only visible if a human happens to GET the dashboard in that window. The notifier can't fire (no evaluations → no transitions), so the one thing the checks plane cannot alert on is its own outage. Had a lab actually gone down inside those 37 minutes, we'd have learned about it from users, not from checks.

I can't tell you what the runner was doing during the gap — the pod has since been replaced and we don't run log aggregation on that cluster yet, so the audit timeline above is all the evidence I have. If it recurs I'll capture logs. Two smaller observations from the same morning that may or may not be related: batch spacing drifts (nominal 300s, observed gaps of 6 and even 9 minutes between batches), and one row in the resumed window carried duration_ms ~5600 for a plain TCP check.

This looks like a sibling of #2245 (agent scheduler stops firing after a failed run — same silent-stall shape, different loop).

The ask, in order of value to us:

1. a watchdog on the evaluation loop: log + structured event (`checks_scheduler_stalled` or similar) when no batch has completed for N × cadence
2. a queryable staleness surface — a `/ready` facet or a built-in meta-sensor exposing "seconds since last completed evaluation batch", so an *external* prober (which is how we monitor the monitoring) can catch it
3. ideally: when the loop recovers from a detected stall, a notify to dashboards whose members transitioned while blind

## Current state (grounded 2026-08-03)

[...]

Issue #2756, opened 2026-08-03
checks: sensors discard per-tick evidence — no retention, no trend query


damir-topic:
Parent initiative: #2780

## Ops report (verbatim, v0.26.0)

We run the checks layer as our lab-fleet monitoring (25+ sensors, dashboards + email notify, on 0.26.0) and hit this during a post-incident review: the sensor evaluation history would have answered "when did this start degrading" and "how fast is it filling" — but every tick overwrites `last_value`/`last_evidence`/`state_since`, so the history the runner already computed is thrown away. In our case the monitored platform's own stats had aged into 30-minute rollups for the incident window, so the sensor record was the only shot at fine-grained truth, and it doesn't exist.

Verified live on 0.26.0: the sensor surface is create/list/delete with `last_*` fields only; dashboards carry current rollup state. If there's a retention mechanism we missed, happy to be corrected.

The ask, deliberately bounded:

1. **Persist per-tick evidence rows** — `(sensor_id, ts, state, evidence_value, reason)` — with a deploy-level retention window (e.g. `CHECKS_EVIDENCE_RETENTION_DAYS`, conservative default so nobody's DB grows surprisingly).
2. **A trend query op** (MCP + CLI + REST) with binary filters only: `sensor_id` (exact), `from`/`to`, optional `state` enum, `limit`; deterministic `ts ASC` ordering; raw rows, client aggregates. No server-side smoothing, downsampling or scoring knobs — same determinism reasoning as the #1177 retrieval decision.

Non-goals we explicitly do *not* want from meho: a PromQL-style engine, per-VM metric cardinality, or dashboard rendering — that's a TSDB's job and we're standing one up next to meho for exactly that. This ask is only "don't discard what the runner already computed": it makes sensor history forensic (when did X first flap) and gives every governed check a free trend, which a external TSDB can't reconstruct after the fact.

## Current state (grounded 2026-08-03)

- The report is correct, and the latest-only shape is a *recorded decision*: the `Sensor` model docstring section "Latest-state projection (not a results table)" (`backend/src/meho_backplane/db/models.py:6098-6109`) states a results table was rejected on retention/pruning-cost grounds. This task supersedes that decision with the reporter's bounded shape — a deploy-level retention window answers exactly the cost objection.
[...]

Issue #2733, opened 2026-08-01
Task: dedicated machine identity for automated reviews — restore the two-party review property


damir-topic:
Parent goal: #221

## Summary

The autonomous implement/review skills run under a maintainer's own GitHub identity, which is also the identity that authors their PRs. GitHub refuses `APPROVE` and `REQUEST_CHANGES` on your own pull request, so every automated review degrades to a `COMMENT` carrying a prose `**Verdict:**` line. That leaves `reviewDecision` at `REVIEW_REQUIRED` forever, so a human must click merge on every PR, and — more seriously — the automation's "approval" is indistinguishable from a real one in prose while carrying none of GitHub's two-party guarantee. Give the automation its own machine identity so its reviews are real reviews.

## Current state

- `/auto-review-pr` posts its verdict as a GitHub review. On PRs authored by the same account it runs as, GitHub returns 422 (`"Can not approve your own pull request"` / `"Can not request changes on your own pull request"`), and the skill falls back to `COMMENT` state with a binding `**Verdict:**` line in the body.
- Observed across the whole of Initiative #2716: PRs #2724, #2726, #2728, #2729, #2730 each ended `mergeable=MERGEABLE`, all checks green, all review threads resolved, and `mergeStateStatus=BLOCKED` purely on `reviewDecision=REVIEW_REQUIRED`. Every one required a manual merge click. The same is recorded on the earlier #2710-#2715 series.
- Consequences, in order of severity:
  1. **The two-party review property is not actually held.** Branch protection asks for one approving review; what exists is a comment written by automation running as the author. The merge click is the only real second-party step, and it is not a review.
  2. `--auto-merge` on `/auto-implement-initiative` can never complete a run unattended, which is the flag's entire purpose.
  3. A reader skimming the PR sees "Verdict: APPROVE" and may reasonably believe the PR carries an approval it does not.
- Adjacent symptom from the same root: DCO. Commits authored by the automation lack `Signed-off-by` unless every worker remembers `git commit -s`; two PRs in #2716 needed a squash and force-push to repair, because this repo has no `.github/dco.yml` and the DCO app's remediation-commit path is therefore disabled.

## Desired state

A dedicated machine identity (GitHub App installation token preferred over a PAT-backed machine user — App tokens are short-lived, scoped per-installation, and attribute cleanly) that:

1. Is the identity the review skills authenticate as, so `APPROVE` / `REQUEST_CHANGES` succeed and `reviewDecision` reflects the automated review's actual verdict.
[...]

Issue #2680, opened 2026-07-23
Approved re-dispatch of argocd.app.sync (dry_run) fails with an opaque 500, and the DISPATCH audit row records no error detail


ilhan-karic:
We ran the first human-pair approve→execute roundtrip on the argocd connector (v0.25.0) and the good news first: the whole approval mechanic works. Park (`argocd.app.sync` with `dry_run: true` against a Synced app), a second operator approves, and the re-dispatch fires automatically ~110ms later with a clean parent/child audit lineage (park → decision → DISPATCH rows linked via `parent_audit_id`). No manual re-call needed — nice.

The bad news, in two halves:

1. The re-dispatched op failed: the DISPATCH audit row shows `status_code: 500, result_status: error` after ~80ms, and no ArgoCD operation was ever created on the target (the app's `operationState` never moved; the app was untouched). 80ms rules out the poll timeout, so it looks like the sync handler dies before or at the POST — possibly the `dry_run` param handling in the typed handler (SyncRequest body shape?). We can't tell from our side, because of half 2.

2. The DISPATCH audit row carries no error detail — just `result_status: error` and a params hash. The approver's CLI/UI gets the `dispatch` block with the real error, but that surface is ephemeral. For a product whose pitch is the audit trail, a failed approved write with no recorded reason is a forensics gap. Normal (non-approval) dispatches return an excellent structured error envelope (`error_code`, `http_status`, `upstream_message`) — that same envelope should land in the audit payload.

Asks (binary):
- Approved re-dispatch of `argocd.app.sync` with `dry_run: true` succeeds and returns the SyncResult (or documented dry-run equivalent).
- A failed dispatch's audit row carries the same structured error envelope the caller receives — never just `result_status: error`.

Repro: park `argocd.app.sync {name: <synced-app>, dry_run: true}` against any registered ArgoCD target, approve as a second operator, read the DISPATCH audit row.

Consumer signal: `approved-redispatch-argocd-sync-dry-run-500` (evoila-bosnia/claude-rdc-hetzner-dc, campaign #1556 / #2039; probe checks the latest `argocd.app.sync` DISPATCH row's `result_status`).

PR #2832, opened 2026-08-06
fix(connectors): netdiag allowlist refusal is a dispatch error, not a reading (#2784)

### LINKED ISSUE #2784: checks: netdiag probe-allowlist refusal is indistinguishable from a down host — sensors go false-critical on a config regression
We just spent a morning chasing a "down" vCenter that was up the whole time. A redeploy had reverted `MEHO_NETDIAG_PROBE_ALLOWLIST` to an older value, and every tcp sensor pointed at a host that fell off the list flipped critical within one cadence window — and stayed there for 11 hours.

The mechanism: `net.tcp_check` on a non-allowlisted host returns a *successful* dispatch with a false reading:

```
net.tcp_check {"host": "192.0.2.1", "port": 443}
→ {"status": "ok", "result": {"connected": false, "reason": "not_in_probe_allowlist", "latency_ms": null}}   (~4 ms, no dial attempted)
```

A sensor asserting on `$.connected` sees `false` and goes critical. The `reason` field — the only thing distinguishing "policy refused to dial" from "host didn't answer" — is dropped at the assertion select, so the evidence stored on the sensor is just `observed: false`. On the dashboard that renders as a down host, indistinguishable from a real outage. Reproduced on v0.27.0 with the TEST-NET address above.

The contrast that makes this feel like a bug rather than a taste question: when a credentialed sensor can't read its Vault creds, the dispatch *fails* and the sensor lands in `unknown` with `reason: dispatch_not_ok` — a truthful "I can't tell". The allowlist refusal is the same category of problem (the runner is blocked from observing, the target's state is unknown) but it masquerades as an observation.

Ask, in preference order:

1. Treat an allowlist refusal as a dispatch error (`connector_error`, like the credential path) so sensors go `unknown` instead of producing reading-shaped output. Still fail-closed — but "unknown: policy refused" tells the operator the truth.
2. Failing that (or additionally): carry `reason` into the sensor's stored evidence, so the dashboard shows `connected: false (not_in_probe_allowlist)` — one glance instead of a morning of forensics.

Related: #2518 (the probe allowlist's extraEnv-only chart surface — the same deploy-config fragility that caused our regression), #2763 (the fleet degradation this rode in alongside; commented there separately).

damir-topic:
## Summary

- A `net.*` probe whose destination is outside `MEHO_NETDIAG_PROBE_ALLOWLIST` returned a **successful** dispatch carrying a reading-shaped payload (`{"connected": false, "reason": "not_in_probe_allowlist", ...}`, `status="ok"`) without dialing anything. A Sensor asserting on `$.connected` read `false` and flipped **critical**, indistinguishable from a genuine outage — and the `reason` sibling was dropped at the assertion select, so the stored evidence was a bare `observed: false`. A redeploy that reverted the allowlist value paged as an 11-hour fleet outage that never happened.
- The refusal now **propagates** out of every `net.*` handler and the dispatcher classifies it as a structured `connector_probe_refused` error — the probe-policy sibling of `connector_vault_forbidden` (#2091) and `connector_unsupported` (#1627). Sensors need **no change**: the check-runner's existing non-`ok` mapping lands them in `unknown` with `reason: dispatch_not_ok` and the allowlist named in `dispatch_error`. **Zero checks-layer diffs.**
- The operator-facing message names the env var, states that no socket was opened (so the result is *not* a reachability answer), and carries the add-the-CIDR-or-hostname remediation plus the `netdiag.probeAllowlist` chart key that renders it. `extras` carry `error_code` / `allowlist_env` / `host` / `detail` / `exception_class`.
- `net.http_probe`'s mid-chain **redirect** re-gate is the deliberate carve-out and stays `status="ok"` / `reachable: true` / `reason: "blocked_redirect"` — the prior hop answered, so that is a real observation.

## Why ask (1) and not ask (2)

The issue asks, in preference order, for (1) treating the refusal as a dispatch error, or (2) carrying `reason` into the sensor's stored evidence. This PR implements **(1) only**, and (2) becomes unnecessary rather than skipped:

- **(2) does not fix the reported harm.** It keeps the state `critical`. A truthful-evidence false alarm still notifies, still fires the diagnose-only investigator, and still poisons rollup and flap suppression. The operator still gets paged for a host that was never probed.
- **(2) has no clean generic implementation.** `evaluate_assertion` navigates the select path to a scalar and builds evidence from `{path, aggregate, comparator, expect, observed}`. Carrying a payload sibling into evidence needs either per-connector knowledge in the pure evaluator or unbounded payload snapshots in the evidence rows (which collides with #2756's evidence retention).
- **(1) delivers (2)'s stated benefit anyway.** The sensor's stored evidence now carries `dispatch_error` with the full structured message naming `MEHO_NETDIAG_PROBE_ALLOWLIST` — the "one glance instead of a morning of forensics" the issue asked for — and the rollup's `unknown` → `degraded` mapping turns the config regression into attention rather than a false page.

## Design decisions

- **Where the taxonomy arm lives.** `ProbeNotAllowedError` subclasses `ValueError`, which no earlier dispatcher arm catches, so it is narrowed out of the generic `except Exception` arm via `is_probe_refused` rather than a dedicated `except`. The exception type is resolved through a **deferred, cached** import: `connectors/net/__init__.py` queues every `net.*` registrar (pulling dnspython / pyOpenSSL / httpx *and* `operations.typed_register`), so a module-scope import from `operations/_errors.py` would close an import cycle back through `operations/__init__`. The resolver returns `None` if the import ever fails, preserving the dispatcher's never-raises contract.
- **The message stays address-free**, preserving `allowlist.py`'s documented never-echo-a-destination posture. The caller's own destination string rides `extras.host` instead (`assert_probe_allowed` resolves nothing, so echoing it back is not a DNS/topology oracle). That also keeps the connector's *audit-visible host* foundation intact: an error row writes no `raw_payload`, but #2680 threads `extras` into `audit_log.payload`.
- **Agent-visible contract change**, so it ships with a CHANGELOG behaviour note including an adopter migration line. The reason code `not_in_probe_allowlist` no longer appears on any `net.*` response.

## Ops affected

`net.tcp_check`, `net.dns_lookup`, `net.tls_inspect`, `net.http_probe`, `net.ntp_check`, `net.ping`, `net.trace`, `net.path_mtu`.

## Test plan

- [x] `ruff format --check src/ tests/` — 1410 files already formatted
- [x] `mypy src/meho_backplane/ --ignore-missing-imports` — only the pre-existing macOS-only `socket.MSG_ERRQUEUE` error in `icmp.py` (identical on `main`; `MSG_ERRQUEUE` is Linux-only, so CI is clean)
- [x] `pytest -n 3 --dist loadscope --ignore=tests/integration --ignore=tests/migrations tests/` — **11120 passed, 181 skipped**. Three failures, each verified **pre-existing on clean `main`** in this macOS sandbox: `test_trace_reaches_loopback` (Linux-only `IP_RECVERR`), `test_chosen_resolver_queries_that_nameserver` (`OSError: [Errno 49] Can't assign requested address`), `test_chart_mcp_resource_uri.py::test_no_mcp_env_keys_when_ingress_disabled_and_nothing_set`
- [x] **AC: every initial-target `net.*` refusal returns `status="error"` with the structured code, names the env var + remediation, never echoes a resolved address** — asserted across all six `test_connectors_net_*.py` suites; `test_connectors_net_probe.py` additionally asserts the message contains `MEHO_NETDIAG_PROBE_ALLOWLIST` + `netdiag.probeAllowlist` and does **not** contain the probed host
- [x] **AC: `blocked_redirect` behaviour unchanged, pinned by test** — `test_redirect_to_non_allowlisted_host_is_blocked_and_never_dialed` now also asserts `extras.error_code is None` alongside `status="ok"` / `reachable=true`
- [x] **AC: integration test through `_run_evaluation`** — new `test_netdiag_allowlist_refusal_persists_unknown_not_critical` in `tests/test_sensor_runner.py` drives the **real** dispatcher (no monkeypatched `dispatch`) with the issue's TEST-NET-1 repro (`192.0.2.1:443`) and asserts `last_state == "unknown"`, `evidence.reason == "dispatch_not_ok"`, and `dispatch_error` naming the allowlist. Verified it **fails on the pre-fix code** with `AssertionError: assert 'critical' == 'unknown'` — it reproduces the reported bug, it does not merely restate the fix
- [x] **AC: schemas / `llm_instructions` / contract docstrings / `docs/codebase/connectors-net-diagnostics.md` updated** — response-schema `reason` prose, `output_shape`, `when_to_use`, registrar `description`s, and the `allowlist.py` / `ops.py` contract docstrings all re-stated; the doc gains a *Refusal is a dispatch error, not a reading* section
- [x] **AC: CHANGELOG behaviour-change entry** — under `[Unreleased]` with the adopter migration recipe
- [x] **AC: existing `net.*` suites green with refusal tests updated** — 216 passed across the six `net` suites + `test_sensor_runner.py` + `test_operations_dispatcher.py`

## Docs

- `docs/codebase/connectors-net-diagnostics.md` — new *Refusal is a dispatch error, not a reading (#2784)* section; every stale `not_in_probe_allowlist` reason-code reference corrected across the tcp / dns / tls / http / ntp / icmp passages
- `docs/codebase/error-message-shape.md` — audit-table row for `connector_probe_refused` (per that doc's own "add the code to the catalogue in the same PR" rule)
- `docs/codebase/sensor.md` — Known-issues bullet generalising the hazard: a connector that refuses to execute must fail the dispatch, because assertion evidence carries no payload siblings and so a reading-shaped refusal has no in-band way to announce itself

## Out of scope / adjacent findings

- `register_net_icmp_operations` is a pre-existing **184-line** registrar, already far over the code-quality block limit of 100. Three cosmetic `description` additions inside it tripped the gate, so they were dropped — the load-bearing corrections for those ops (the response-schema `reason` prose) live at module scope and are included. Splitting that registrar into three is its own refactor task.
- `test_chart_mcp_resource_uri.py::test_no_mcp_env_keys_when_ingress_disabled_and_nothing_set` fails identically on clean `main` in this sandbox — unrelated to this change, not investigated here.
- The `net.*.refused` structlog lines are removed (their `_log` / `structlog` imports in `ops.py` / `tls.py` / `icmp.py` became dead). The refusal is now recorded on the durable audit row instead, which is strictly more durable than an INFO line.
- #2518 (typed `netdiag.probeAllowlist` chart key) already shipped; it makes the allowlist schema-validated and discoverable, but a reverted allowlist *value* still lands as a false page regardless of how the env var is set — which is what this state fix addresses. #2763's watchdog (shipped via #2791) would not have caught this either: the loop was healthy, it was lying, not dead.

Closes #2784

---

## Follow-up (iteration 1, 2026-08-06)

Addressed the findings from the iteration-1 review (verdict of record `REQUEST_CHANGES`; delivered in comment mode because GitHub rejects a formal `REQUEST_CHANGES` from the PR author — the shared-automation-identity limitation tracked in #2733):

- **B1** `9b59b7a0` — the adopter-visible body-shape change was filed under `### Fixed` only. This repo's own CHANGELOG category rules put "body-shape changes … any other contract change that requires adopters to update their client code" under `### Breaking changes`, above `Added`, with the migration recipe on the bullet — and this file is the authoritative source for the published GitHub Release notes, so an adopter scanning a release for breaking changes would have found nothing. Adds the `Breaking changes` section (migration recipe, the eight affected ops, the unchanged `blocked_redirect` carve-out) and trims the `Fixed` entry's `Adopters` bullet to a pointer so the recipe is single-sourced.
- **M1** `11822da3` — `docs/deploying.md` § `netdiag.probeAllowlist` is the operator-facing page for the exact knob whose reverted value caused the incident, and described only the allow/deny semantics. Now names `connector_probe_refused`, states that nothing is dialed so the result is not a reachability answer, and spells out the sensor consequence (`unknown`, rolling up as `degraded`, never `critical`).

Disputed: none. Deferred: none.

Re-verified after the fixes: `ruff check` clean, `ruff format --check` clean (1410 files), `code-quality.py --diff` 0 blockers. Both commits are docs-only — no source or test behaviour changed since the reviewed head.

---

## Follow-up (iteration 2, 2026-08-06)

Addressed the finding from the iteration-2 review:

- **B2** `2d6d97e9` — `connector_probe_refused` was the only structured dispatch cause without a `test_operations_connector_<cause>.py` module, and its narrowing predicate was untested. Unlike its siblings (which discriminate on `hvac.exceptions.Forbidden`, an `ssl`-caused `httpx.ConnectError`, `NotImplementedError`), `is_probe_refused` discriminates on a `ValueError` subclass from inside the catch-all `except Exception` arm — the widest discriminator in the family — so nothing asserted that a plain `ValueError` escaping a handler still lands as `connector_error`. Adds `backend/tests/test_operations_connector_probe_refused.py` (12 tests): the narrowing through the real dispatch path, the classification, the parametrised predicate, the builder's message contract (env var + `netdiag.probeAllowlist` + both doc refs present, destination absent), and the never-raises degenerate branches (no `host`, non-`str` `host`, over-cap message, empty message leaving no dangling `Allowlist said:` tail).

The narrowing test was **mutation-checked**: widening `is_probe_refused` to `isinstance(exc, ValueError)` fails `test_dispatch_plain_value_error_falls_through_to_connector_error` and the `plain-value-error` predicate case, and nothing else — so it pins the boundary rather than restating the implementation.

Disputed: none. Deferred: none.

Re-verified: `ruff check src/ tests/` clean, `ruff format --check` clean, `mypy src/` unchanged (only the pre-existing macOS-only `MSG_ERRQUEUE` attr), `code-quality.py --diff` 0 blockers, new module 12 passed.


CHANGELOG.md
```diff
@@ -90,6 +90,45 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Fixed — a refused `net.*` probe no longer reads as a down host (#2784)
```
damir-topic, line 114: **[🛑 B1]** — an adopter-visible contract change is filed under `Fixed` only, so the migration recipe never surfaces under `Breaking changes`

**What:** this entry is the sole CHANGELOG record of the change, and it sits under `### Fixed`. The change alters the response body shape of eight agent-callable ops for one class of outcome: a caller that branched on `result.reason == "not_in_probe_allowlist"` against a `status="ok"` body now gets `status="error"` and must read `extras.error_code` instead. The entry even carries the migration line ("**Adopters:** … switch the check to the error envelope"), which is precisely the artefact this repo's `Breaking changes` category is defined to hold.

**Why it matters:** this file's own category rules define **Breaking changes** as "schema renames, **body-shape changes**, removed endpoints, or any other contract change that requires adopters to update their client code. Each bullet includes a migration recipe … **Surfaces above `Added`** in the release section so adopters reading top-to-bottom see migrations before features." This file is also the authoritative source for the published GitHub Release notes — `.github/workflows/cli-release.yml` extracts the matching section verbatim and passes it to GoReleaser via `--release-notes`. An adopter who scans a release for `Breaking changes` before upgrading will find nothing here and ship a client that silently mis-reads a refusal as a successful dispatch — the same failure class this PR exists to remove, displaced from the sensor onto the SDK consumer. Category placement is the delivery mechanism for the migration, not bookkeeping.

**Required change:** add a `### Breaking changes` section for #2784 above this one, carrying the migration recipe (the smallest concrete client edit), and trim the `Adopters:` bullet here to a pointer so the recipe is single-sourced. Keep the `### Fixed` narrative — the change is genuinely both, and Keep a Changelog permits one change appearing under two categories. Suggested shape:

```markdown
### Breaking changes — `net.*` refusals move from a `status="ok"` body to a dispatch error (#2784)

- A `net.*` probe whose destination is outside `MEHO_NETDIAG_PROBE_ALLOWLIST`
  now returns `status="error"` with `extras.error_code ==
  "connector_probe_refused"`. It previously returned `status="ok"` with a
  reading-shaped body carrying `reason: "not_in_probe_allowlist"`, which the
  `### Fixed` entry below explains was indistinguishable from a real outage.
  Affects `net.tcp_check`, `net.dns_lookup`, `net.tls_inspect`,
  `net.http_probe`, `net.ntp_check`, `net.ping`, `net.trace`, `net.path_mtu`.
- **Migration:** replace `result["reason"] == "not_in_probe_allowlist"` on an
  `ok` body with `result["extras"]["error_code"] == "connector_probe_refused"`
  on an `error` envelope. Sensors need no change — the check-runner already
  maps any non-`ok` dispatch to `unknown`.
- **Unchanged:** `net.http_probe`'s mid-chain redirect re-gate still returns
  `status="ok"` with `reachable: true` / `reason: "blocked_redirect"`.
```

**Source:** `CHANGELOG.md` "How entries are added" § Categories (the `Breaking changes` definition and its above-`Added` ordering rule); `.github/workflows/cli-release.yml` (release-notes extraction).

damir-topic (commented):
> **Verdict of record: `REQUEST_CHANGES`.** GitHub rejects a formal
> `REQUEST_CHANGES` submission from the PR author
> (`422: Review Can not request changes on your own pull request`) — the
> shared-automation-identity limitation tracked in #2733 — so this review
> is delivered in comment mode. The binding verdict is the one stated
> below and in `.claude/auto/review-results/pr-2832-iter-1.json`; the
> review/fix dance treats it as `REQUEST_CHANGES` and will not exit
> merge-ready until B1 is addressed and re-reviewed.

**Anchor task:** #2784 — netdiag probe-allowlist refusal is indistinguishable from a down host
**Surface area:** backend
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `05b26cf7561f3578677304b22c08f299cb2b9e11`

### Acceptance criteria

- [x] Every initial-target `net.*` allowlist refusal returns dispatch `status="error"` with the structured error code; the message names `MEHO_NETDIAG_PROBE_ALLOWLIST` + the remediation and never echoes a resolved address — verified across all six `test_connectors_net_*.py` suites; `test_connectors_net_probe.py` additionally asserts the summary contains the env var and `netdiag.probeAllowlist` and does **not** contain the probed host.
- [x] `blocked_redirect` behaviour unchanged (`status="ok"`, `reachable=true`) — pinned by test: `test_redirect_to_non_allowlisted_host_is_blocked_and_never_dialed` now also asserts `extras.error_code is None`.
- [x] Integration test through `_run_evaluation` — `test_netdiag_allowlist_refusal_persists_unknown_not_critical` drives the real dispatcher with the issue's TEST-NET-1 repro and asserts `unknown` / `dispatch_not_ok` / allowlist named in `dispatch_error`. Independently confirmed to fail on the pre-fix tree with `assert 'critical' == 'unknown'`, so it reproduces the reported bug rather than restating the fix.
- [x] Response-schema `reason` prose, `llm_instructions.output_shape`, `allowlist.py` / `ops.py` contract docstrings, and `docs/codebase/connectors-net-diagnostics.md` updated to the new contract.
- [⚠️] CHANGELOG behaviour-change entry — present and thorough, but filed under the wrong category. See **B1**.
- [x] Existing `net.*` suites green with refusal tests updated to the new envelope — 365 passed in the review worktree.

### Project-board alignment

|---|---|---|---|---|---|
| #2784 | Open (`bug`, `task`, `priority:high`, `effort:medium`, `connector`) | `Closes` | ✅ | ✅ full | close on merge |

Deliberately standalone per the triage disposition (a G0.39 cycle seed, explicitly kept out of #2780's scope), so there is no parent Initiative to walk to — the hierarchy correctly terminates at the task. Nothing else in the diff closes another issue; no other open PR touches `connectors/net/` or `operations/_errors.py`.

### Predicted vs. actual diff

Predicted from the issue independently of the PR: the seven `net.*` refusal sites, a new dispatcher taxonomy arm + error builder, the six `net` test suites, a checks-layer integration test, the net-diagnostics doc, the error-shape catalogue, and a CHANGELOG note.

- ✅ `connectors/net/{allowlist,ops,tls,ntp,icmp,http_probe}.py`, `operations/_errors.py`, `operations/dispatcher.py`, the six `test_connectors_net_*.py`, `test_sensor_runner.py`, `docs/codebase/connectors-net-diagnostics.md`, `docs/codebase/error-message-shape.md`, `CHANGELOG.md`
- ❓ `docs/codebase/sensor.md` — unexpected but justified: it generalises the hazard (a connector that refuses to execute must fail the dispatch, because assertion evidence carries no payload siblings) on the page an operator reads when a sensor state looks wrong. Correctly framed as a boundary, not a defect.
- 🕳️ `docs/deploying.md` — predicted, absent. See **M1**.
- No 🚫 undisclosed edits. Every file in the diff is accounted for by the issue's acceptance criteria; the removal of the now-dead `_log` / `structlog` in `ops.py` / `tls.py` / `icmp.py` is disclosed in the PR body.

### Integration verification

Run inside `.claude/worktrees/pr-2832-review` at the PR head:

- [x] `ruff format --check src/ tests/` — 1410 files already formatted
- [x] `pytest -n 3 --dist loadscope` over the six `net` suites + `test_sensor_runner.py` + `test_operations_audit_error_envelope.py` + `test_operations_connector_auth_failed.py` + `test_operations_connector_error_redaction.py` + `test_dispatcher_redaction_integration.py` + `test_checks_{assertions,rollup,sensor_results}.py` — **365 passed**, 2 failed
- [x] Both failures confirmed **pre-existing on `main`** in this macOS sandbox and unrelated to the diff: `test_trace_reaches_loopback` (Linux-only `IP_RECVERR`) and `test_chosen_resolver_queries_that_nameserver` (`OSError: [Errno 49] Can't assign requested address`).
- [x] `mypy` — reasoned + spot-checked: the tree's only error is the pre-existing macOS-only `socket.MSG_ERRQUEUE` attr in `icmp.py`, byte-identical on `main`.
- [x] Regression reproduction executed, not assumed: the new sensor test was run against the pre-fix source and fails with `assert 'critical' == 'unknown'`.
- [x] Consumer sweep executed: no `not_in_probe_allowlist` references survive outside deliberate historical prose; no caller invokes a `net_*` handler outside `connectors/net/`; `_run_branch_with_error_handling` has exactly one call site so every `source_kind` branch is covered; the `mail` connector only cites the allowlist as a semantic analogue and shares no code.
- [x] Import-cycle claim verified against the installed tree, not assumed — `connectors/net/__init__.py` imports all five op modules and `operations.typed_register`, so the deferred-import rationale holds.
- [⚠️ not run] migrations replay — n/a, no schema change.
- [⚠️ not run] `docker compose up` — n/a, no deploy-surface change in this diff (which is itself **M1**).

- 🛑 **B1** — the contract change is filed under `### Fixed` only, contradicting this repo's own CHANGELOG category rule for adopter-visible body-shape changes; required change: add a `### Breaking changes` section carrying the migration recipe, placed per the documented ordering.
- ⚠️ **M1** — `docs/deploying.md` § `netdiag.probeAllowlist` is the operator-facing page for the exact knob whose regression caused this incident, and it still describes only the allow/deny semantics; required change: state what a refusal now looks like (`connector_probe_refused`, sensors read `unknown` not `critical`) so an operator debugging a misconfigured allowlist from that page reaches the answer. Not anchorable inline — `docs/deploying.md` is not in the diff.

- `register_net_icmp_operations` is a pre-existing 184-line registrar, far past the 100-line code-quality block limit. The PR correctly declined to refactor it and correctly declined to plant an escape-valve comment, instead keeping its diff out of that function — but it means `net.ping` / `net.trace` / `net.path_mtu` carry slightly thinner catalogue `description` text than the other five ops. Splitting that registrar into three is its own task.
- `test_chart_mcp_resource_uri.py::test_no_mcp_env_keys_when_ingress_disabled_and_nothing_set` fails identically on clean `main` in this sandbox — pre-existing, unrelated, worth its own look.

### Docs

- [x] `docs/codebase/connectors-net-diagnostics.md` — updated (new *Refusal is a dispatch error, not a reading* section; every stale reason-code reference corrected across the tcp / dns / tls / http / ntp / icmp passages)
- [x] `docs/codebase/error-message-shape.md` — updated (audit-table row for `connector_probe_refused`, satisfying that doc's own "add the code to the catalogue in the same PR" rule)
- [x] `docs/codebase/sensor.md` — updated (generalised boundary)
- [❌] `docs/deploying.md` — see **M1**

- The regression test is the real thing: it drives the **actual** dispatcher rather than a monkeypatched `dispatch`, uses the incident's own TEST-NET-1 address, and asserts the state the issue reported. Most fixes of this shape ship a test that only restates the new behaviour.
- The `extras["host"]` decision is the non-obvious catch. Turning a `status="ok"` refusal into an error silently drops `raw_payload`, which would have quietly deleted the connector's documented audit-visible-host foundation. Carrying the caller's own literal on the exception — and keeping it *out* of the free-text summary to preserve the never-echo-a-destination posture — resolves both constraints instead of trading one for the other.
- The deferred-import rationale is correct and load-bearing rather than defensive hand-waving: a module-scope import really would close a cycle through `operations/__init__`, and the `None`-on-failure fallback keeps the dispatcher's never-raises contract intact on the error path itself.

backend/src/meho_backplane/operations/_errors.py
```diff
@@ -534,6 +538,114 @@ def result_connector_error(
...
+    :func:`result_connector_error`, mirroring :func:`is_auth_failed_status`'s
+    role for the auth-class statuses. Type-based, not message-based —
+    ``ProbeNotAllowedError`` subclasses :class:`ValueError`, which no earlier
+    dispatcher arm catches.
+    """
+    exc_type = _probe_refused_error_type()
+    return exc_type is not None and isinstance(exc, exc_type)
+
+
+def result_connector_probe_refused(
```
damir-topic, line 579: **[🛑 B2]** — the sixth structured dispatch cause ships without its per-cause test module, leaving the narrowing property untested

**What:** every other structured cause in this taxonomy has a dedicated behavioural module — `test_operations_connector_auth_failed.py`, `..._http_403.py`, `..._tls_verify_failed.py`, `..._unsupported.py`, `..._vault_forbidden.py` — each covering (a) the builder's shape and its degenerate inputs and (b) a `..._falls_through_to_connector_error` test proving the arm does **not** capture its neighbours. `connector_probe_refused` adds a builder *and* a new narrowing predicate (`is_probe_refused`) with neither. The end-to-end coverage in the six `test_connectors_net_*.py` suites proves the happy path fires, but nothing anywhere asserts the negative.

**Why it matters:** the narrowing is the safety property of this arm, and it is the widest one in the family. The siblings key off distinctive types (`hvac.exceptions.Forbidden`, `httpx.ConnectError` + an `ssl` cause, `NotImplementedError`); this one keys off a subclass of `ValueError` — the single most commonly raised exception type in the codebase — sitting inside the catch-all `except Exception` arm. If `is_probe_refused` ever widened to `isinstance(exc, ValueError)` (a plausible future "simplification", since the deferred import looks like ceremony), every bad-input `ValueError` from every connector would be relabelled "your probe allowlist refused this" and would ship an allowlist remediation for an unrelated bug. That is precisely the mislabelling `test_operations_connector_tls_verify_failed.py::test_dispatch_non_ssl_connect_error_falls_through_to_connector_error` and `test_operations_connector_vault_forbidden.py::test_dispatch_other_hvac_errors_fall_through_to_connector_error` exist to prevent — and it is the same class of "a plausible-looking answer for something that did not happen" that this PR was written to remove. The builder's defensive `getattr(exc, "host", None)` path is likewise unexercised: no test raises the error without a `host`, so nothing proves the `None` branch produces a valid envelope rather than a `TypeError` inside a never-raises path.

**Required change:** add `backend/tests/test_operations_connector_probe_refused.py` following the sibling modules' structure. Minimum contents:

1. **Narrowing (the load-bearing one):** a typed handler raising a plain `ValueError("bad input")` still yields `extras["error_code"] == "connector_error"` — assert `!= "connector_probe_refused"` explicitly, mirroring `test_operations_dispatcher.py:2030`.
2. **Classification:** a handler raising `ProbeNotAllowedError` yields `connector_probe_refused` with `allowlist_env`, `host`, `detail`, `exception_class` populated.
3. **Builder shape:** `result_connector_probe_refused` names `MEHO_NETDIAG_PROBE_ALLOWLIST`, `netdiag.probeAllowlist`, and both doc references; asserts the summary does **not** contain the host.
4. **Degenerate input:** `ProbeNotAllowedError("refused")` with no `host` → `extras["host"] is None` and a well-formed envelope (the `getattr` branch).
5. **Oversized/empty detail:** an over-cap message truncates via `_sanitize_free_text`; an empty message leaves no dangling `Allowlist said:` tail — the shape both `test_builder_caps_oversized_message` and `test_builder_empty_message_has_no_dangling_tail` pin for the vault builder. Note that the current builder appends the tail under `if detail:`, so the empty case is already handled — the test pins it against regression.

**Source:** `backend/tests/test_operations_connector_vault_forbidden.py` (module docstring + `test_dispatch_other_hvac_errors_fall_through_to_connector_error`); `backend/tests/test_operations_connector_tls_verify_failed.py::test_dispatch_non_ssl_connect_error_falls_through_to_connector_error`; `docs/codebase/error-message-shape.md` § builder-shape discipline.

damir-topic (commented):
> **Verdict of record: `REQUEST_CHANGES`.** GitHub rejects a formal
> `REQUEST_CHANGES` submission from the PR author
> (`422: Review Can not request changes on your own pull request`) — the
> shared-automation-identity limitation tracked in #2733 — so this review
> is delivered in comment mode. The binding verdict is the one stated
> below and in `.claude/auto/review-results/pr-2832-iter-2.json`.

**Anchor task:** #2784 — netdiag probe-allowlist refusal is indistinguishable from a down host
**Surface area:** backend
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `11822da303e101f020d7d856f8e0747b125222e3`

### Iteration-1 findings — disposition

| Finding | Status | Evidence |
|---|---|---|
| 🛑 B1 — contract change filed under `Fixed` only | ✅ resolved | `9b59b7a0` adds `### Breaking changes` for #2784 carrying the migration recipe, the eight affected ops and the unchanged `blocked_redirect` carve-out; it sits above both `Fixed` and `Added`, satisfying the documented ordering. The `Fixed` entry's `Adopters` bullet is trimmed to a pointer, so the recipe is single-sourced. |
| ⚠️ M1 — `docs/deploying.md` § `netdiag.probeAllowlist` silent on the refusal shape | ✅ resolved | `11822da3` names `connector_probe_refused`, states nothing is dialed so it is not a reachability answer, and spells out the sensor consequence (`unknown` rolling up as `degraded`, never `critical`). Accurate against `checks/rollup.py`'s UNKNOWN→degraded mapping. |

The iteration-1 delta is docs-only — no source or test file changed between `05b26cf7` and `11822da3`, so iteration 1's correctness verification carries forward unmodified.

### New finding

Surfaced by re-walking the taxonomy the PR extends, rather than the diff:

- 🛑 **B2** — `connector_probe_refused` is the sixth structured dispatch cause, and it is the only one without a `test_operations_connector_<cause>.py` module. Its narrowing predicate `is_probe_refused` keys off a `ValueError` subclass inside the catch-all `except Exception` arm — much wider than any sibling's discriminator — and nothing asserts that a plain `ValueError` from a handler still lands as `connector_error`. The builder's defensive `getattr(exc, "host", None)` branch is likewise unexercised. Required change: add the module per the sibling structure (full contents in the inline comment).

This is a genuine gap, not iteration-1 rework: the sibling convention is uniform (`auth_failed` / `http_403` / `tls_verify_failed` / `unsupported` / `vault_forbidden` all ship one), and each of those modules carries an explicit `..._falls_through_to_connector_error` test for exactly this property.

### Acceptance criteria

All six of the issue's criteria remain satisfied — see the iteration-1 review. The CHANGELOG criterion is now fully met (B1 resolved). B2 is a repo-convention gap in test structure, not an unmet acceptance criterion.

### Project-board alignment

Unchanged from iteration 1: #2784 linked via `Closes`, delivered full, deliberately standalone per the triage disposition. No other issue is closed by this diff; no other open PR touches these files.

### Predicted vs. actual diff (delta only)

- ✅ `CHANGELOG.md` — expected and correct (B1 fix)
- ✅ `docs/deploying.md` — this was iteration 1's 🕳️ missing entry; now present and correct (M1 fix)
- 🕳️ `backend/tests/test_operations_connector_probe_refused.py` — predicted from the taxonomy's uniform per-cause test convention, absent. See **B2**.
- No 🚫 undisclosed edits; the PR body's follow-up section accurately records both commits.

### Integration verification

Run inside `.claude/worktrees/pr-2832-review` at the new head `11822da3`:

- [x] `ruff format --check src/ tests/` — 1410 files already formatted
- [x] `pytest -n 3 --dist loadscope` over `test_connectors_net_probe.py`, `test_sensor_runner.py`, `test_operations_connector_vault_forbidden.py`, `test_operations_connector_tls_verify_failed.py` — **69 passed** (the two sibling taxonomy suites included deliberately, to confirm the new arm does not perturb them)
- [x] `DCO` check — green on the new head; both fix commits carry `Signed-off-by` under the same identity as the original commit, so no remediation is needed (this repo's DCO app does not accept remediation commits)
- [x] Delta scoped and re-read in full: docs-only, no behavioural surface touched
- [⚠️ not run] full unit sweep — re-running the 11-minute suite adds nothing over a docs-only delta; iteration 1's full-suite result (11120 passed) stands

- 🛑 **B2** — add `backend/tests/test_operations_connector_probe_refused.py` covering, at minimum: a plain `ValueError` still classifying as `connector_error` (the narrowing), the `ProbeNotAllowedError` → `connector_probe_refused` classification with populated `extras`, the builder's message content (env var + chart key + both doc refs, host absent), the `host=None` degenerate branch, and the oversized/empty-detail shapes.

Carried forward unchanged from iteration 1: the pre-existing 184-line `register_net_icmp_operations`, and the pre-existing `test_chart_mcp_resource_uri.py` failure in this sandbox.

### Docs

- [x] `docs/codebase/connectors-net-diagnostics.md` — updated
- [x] `docs/codebase/error-message-shape.md` — updated
- [x] `docs/codebase/sensor.md` — updated
- [x] `docs/deploying.md` — updated (M1 resolved)
- [x] `CHANGELOG.md` — updated, now correctly categorised (B1 resolved)

- Both iteration-1 fixes landed as separate, self-describing commits with the rationale in the body rather than being squashed into a "address review comments" blob — the commit log now explains *why* the CHANGELOG category matters, which is the part a future reader needs.
- The B1 fix resisted the easy shortcut of simply relabelling the `Fixed` heading. Keeping the `Fixed` narrative and adding a separate `Breaking changes` section with only the migration recipe is the correct reading of Keep a Changelog, and the `Adopters` bullet was trimmed to a pointer so the recipe cannot drift between two copies.

CHANGELOG.md
damir-topic: Resolved in `9b59b7a0`. Added a `### Breaking changes` section for #2784 above both `Fixed` and `Added`, carrying the migration recipe (`result["reason"] == "not_in_probe_allowlist"` on an `ok` body → `result["extras"]["error_code"] == "connector_probe_refused"` on an `error` envelope), the eight affected ops, and the unchanged `blocked_redirect` carve-out. The `Fixed` entry's `Adopters` bullet is trimmed to a pointer so the recipe is single-sourced and cannot drift between two copies; the `Fixed` narrative stays, since the change is genuinely both.

backend/src/meho_backplane/operations/_errors.py
damir-topic: Resolved in `2d6d97e9` — `backend/tests/test_operations_connector_probe_refused.py`, 12 tests covering all five requested items: the narrowing through the real dispatch path, the classification with populated `extras`, the parametrised predicate, the builder's message contract (env var + `netdiag.probeAllowlist` + both doc refs present, destination absent), and the never-raises degenerate branches (no `host`, non-`str` `host`, over-cap message, empty message leaving no dangling `Allowlist said:` tail).

The narrowing test was mutation-checked rather than assumed: widening `is_probe_refused` to `isinstance(exc, ValueError)` fails `test_dispatch_plain_value_error_falls_through_to_connector_error` and the `plain-value-error` predicate case, and nothing else — so it pins the boundary instead of restating the implementation. It drives the same `net.tcp_check` op as the positive test and changes only the exception type `asyncio.open_connection` raises, isolating the discriminator.

damir-topic (commented):
> **Verdict of record: `APPROVE`.** GitHub also rejects a formal
> self-approval (`422`) — the shared-automation-identity limitation
> tracked in #2733 — so this review is delivered in comment mode. The
> binding verdict is the one stated below and in
> `.claude/auto/review-results/pr-2832-iter-3.json`. `reviewDecision`
> will therefore stay unset; the merge gate is the orchestrator's.

**Anchor task:** #2784 — netdiag probe-allowlist refusal is indistinguishable from a down host
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `2d6d97e9f970aad27d4e38e78a6ced4db3f972e0`

### Prior-iteration findings — all resolved

| Finding | Iter | Status | Evidence |
|---|---|---|---|
| 🛑 B1 — contract change filed under `Fixed` only | 1 | ✅ resolved | `9b59b7a0` — `### Breaking changes` for #2784 with the migration recipe, above both `Fixed` and `Added` per the documented ordering; the `Fixed` entry's `Adopters` bullet trimmed to a pointer so the recipe is single-sourced |
| ⚠️ M1 — `docs/deploying.md` silent on the refusal shape | 1 | ✅ resolved | `11822da3` — names `connector_probe_refused`, states nothing is dialed, spells out the `unknown`-not-`critical` sensor consequence |
| 🛑 B2 — no per-cause test module; narrowing untested | 2 | ✅ resolved | `2d6d97e9` — `test_operations_connector_probe_refused.py`, 12 tests covering all five requested items |

Both inline threads replied to and resolved.

### B2 verification — the fix was checked, not taken on trust

The blocker was "the narrowing is untested", so a test that merely restates the implementation would not have closed it. Two things were checked:

1. **Mutation.** `is_probe_refused` was temporarily widened to `isinstance(exc, ValueError)` — the exact regression the finding described. `test_dispatch_plain_value_error_falls_through_to_connector_error` and the `plain-value-error` predicate case both fail; the other ten pass. The test pins the boundary rather than the implementation.
2. **Isolation.** The narrowing test drives the same real `net.tcp_check` op through the same real dispatch path as the positive test and changes only the exception type `asyncio.open_connection` raises. Nothing but the discriminator varies between the two cases.

The module also covers what the finding called out as unexercised: the builder's `getattr(exc, "host", None)` branch (no `host`, plus a non-`str` `host`), the over-cap message, and the empty-message no-dangling-tail shape. It now matches the structure of all five sibling per-cause modules.

### Acceptance criteria

- [x] Every initial-target `net.*` allowlist refusal returns `status="error"` with the structured code; the message names `MEHO_NETDIAG_PROBE_ALLOWLIST` + the remediation and never echoes a resolved address — asserted across the six `net` suites and now directly in the per-cause module, which explicitly asserts the destination is absent from the message and present only in `extras`.
- [x] `blocked_redirect` unchanged (`status="ok"`, `reachable=true`) — pinned by `test_redirect_to_non_allowlisted_host_is_blocked_and_never_dialed`, which also asserts `extras.error_code is None`.
- [x] Integration test through `_run_evaluation` — `test_netdiag_allowlist_refusal_persists_unknown_not_critical` drives the real dispatcher with the issue's TEST-NET-1 repro; independently confirmed to fail pre-fix with `assert 'critical' == 'unknown'`.
- [x] Response schemas' reason prose, `llm_instructions.output_shape`, the `allowlist.py` / `ops.py` contract docstrings, and `docs/codebase/connectors-net-diagnostics.md` updated.
- [x] CHANGELOG behaviour-change entry — present under both `Breaking changes` (migration recipe) and `Fixed` (narrative).
- [x] Existing `net.*` suites green with refusal tests updated to the new envelope.

### Project-board alignment

Unchanged: #2784 linked via `Closes`, delivered in full, deliberately standalone per the triage disposition. No other issue is closed by this diff; no other open PR touches these files.

### Predicted vs. actual diff

- ✅ All 20 files expected and correct. Iteration 1's 🕳️ `docs/deploying.md` and iteration 2's 🕳️ `test_operations_connector_probe_refused.py` are both now present.
- ❓ `docs/codebase/sensor.md` — unexpected but justified (generalises the hazard on the page an operator reads when a sensor state looks wrong).
- No 🚫 undisclosed edits; the PR body records every commit and its rationale.

### Integration verification

Run inside `.claude/worktrees/pr-2832-review` at `2d6d97e9`:

- [x] `ruff format --check src/ tests/` — 1411 files already formatted
- [x] `mypy src/` (the exact CI target) — only the pre-existing macOS-only `socket.MSG_ERRQUEUE` attr in `icmp.py`, byte-identical on `main`; `MSG_ERRQUEUE` is Linux-only so CI is clean
- [x] Whole taxonomy family re-run together (`probe_refused` + `vault_forbidden` + `tls_verify_failed` + `unsupported` + `auth_failed` + `http_403`) plus the six `net` suites, `test_sensor_runner.py`, `test_operations_dispatcher.py`, `test_operations_audit_error_envelope.py` — 312 passed
- [x] Mutation check on the new narrowing test (see above)

### Full unit sweep at the final head

Re-run in full rather than carried forward, because iteration 2 added an
executable module and a new test module can perturb global registry state
across xdist workers:

```
pytest -n 3 --dist loadscope --ignore=tests/integration --ignore=tests/migrations tests/
3 failed, 11132 passed, 181 skipped in 557.71s
```

11132 passed is exactly +12 over iteration 1's 11120 — the new module's
twelve tests and nothing else moved. All three failures are the known
pre-existing sandbox set, each verified against clean `main`:
`test_trace_reaches_loopback` (Linux-only `IP_RECVERR`),
`test_chosen_resolver_queries_that_nameserver` (macOS bind restriction),
`test_chart_mcp_resource_uri.py::test_no_mcp_env_keys_when_ingress_disabled_and_nothing_set`.

**Forward-compatibility check.** `main` moved 8 dependency bumps ahead of
this branch's base while the PR was in flight, one of them `ruff
0.16.0 → 0.16.1`. The branch's lock still pins 0.16.0, so CI on the PR head
runs the old version — `ruff@0.16.1 check` and `format --check` were run
against this tree explicitly and are clean, so the merge will not
surface a lint delta. None of the eight bumps touch a file in this diff
(`pyproject.toml` / `uv.lock` / two workflows), and GitHub reports
`mergeable: true` — no rebase needed, and no force-push performed at any
point in this run.

### Findings

None. Zero blockers, zero majors, zero minors, zero questions.

Carried forward across all three iterations; neither blocks merge nor requires a response on this PR:

- `register_net_icmp_operations` is a pre-existing 184-line registrar, far past the 100-line code-quality block limit. The PR correctly declined both the out-of-scope refactor and the escape-valve comment, keeping its diff out of that function instead — at the cost of slightly thinner catalogue `description` text on `net.ping` / `net.trace` / `net.path_mtu`. Splitting it into three registrars is its own task.
- `test_chart_mcp_resource_uri.py::test_no_mcp_env_keys_when_ingress_disabled_and_nothing_set` fails identically on clean `main` in this sandbox — pre-existing, unrelated, worth its own look.

### Docs

- [x] `docs/codebase/connectors-net-diagnostics.md` — updated (new *Refusal is a dispatch error, not a reading* section; every stale reason-code reference corrected across the tcp / dns / tls / http / ntp / icmp passages)
- [x] `docs/codebase/error-message-shape.md` — updated (audit-table row for `connector_probe_refused`, per that doc's own same-PR catalogue rule)
- [x] `docs/codebase/sensor.md` — updated (generalised boundary)
- [x] `docs/deploying.md` — updated (M1)
- [x] `CHANGELOG.md` — updated and correctly categorised (B1)

- The regression test earns its keep twice over: it drives the real dispatcher rather than a monkeypatched `dispatch`, uses the incident's own TEST-NET-1 address, and was verified to fail on the pre-fix tree. Most fixes of this shape ship a test that only restates the new behaviour.
- The `extras["host"]` decision is the non-obvious catch. Converting a `status="ok"` refusal into an error silently drops `raw_payload`, which would have quietly deleted the connector's documented audit-visible-host foundation. Carrying the caller's own literal on the exception — and keeping it *out* of the free-text summary to preserve the never-echo-a-destination posture — satisfies both constraints instead of trading one against the other.
- Scope discipline held under pressure at the code-quality gate: rather than refactor a 184-line pre-existing registrar or plant an escape-valve comment to silence the check, the PR kept its diff out of that function and disclosed the tradeoff in the body. That is the right call in both directions.
- The PR body argues ask (1) against ask (2) from the issue text with the state model as the evidence, instead of silently implementing the easier half — which is what the issue's preference ordering actually asked for.

### Verdict

**APPROVE.** Every acceptance criterion is proven by an executed command, all three prior findings are resolved and verified (B2 by mutation, not by inspection), both inline threads are resolved, and every gate that should have run ran green. Merge-ready — the merge itself stays with the orchestrator.

Issue #2993, opened 2026-08-17
Task: vcf-fleet / vcf-logs spec availability — pin + lane, or record evidenced exclusion


damir-topic:
Parent goal: #221
Parent initiative: #2979

Depends-on: #2980

## Problem

`vcf_fleet` (~8 path literals) and `vcf_logs` (~2) carry hand-coded paths with no pinned spec and unknown spec availability (VCF 9 fleet-management and VCF Operations for Logs may publish OpenAPI as part of the VCF 9 API bundle — the shelf's `sddc-manager-9.0` / `vcf-operations-9.0` dirs may already cover part of these surfaces).

## Deliverable

Determine per connector: (a) an existing shelf spec already serves these paths → point the lane at it; (b) a vendor spec exists but isn't pinned → pin it (consumer-shelf PR, vendor-license signoff extension per the ADR) and add the lane; (c) no machine-readable spec exists → record that explicitly in the #2980 standard doc's exclusion list with the evidence, and close.

## Acceptance criteria

- [ ] Both connectors have either a green lane or a recorded, evidenced exclusion.

Issue #2986, opened 2026-08-17
Task: assert + fix vmware _OP_GET_CLUSTER_DRS and complete read-composite lane coverage (residual from #2970)


damir-topic:
Parent goal: #221
Parent initiative: #2979

## Problem

#2970's adjacent finding (recorded in `.claude` result files and the PR #2974 report): `_OP_GET_CLUSTER_DRS` in `backend/src/meho_backplane/connectors/vmware_rest/composites/_read.py` is also not served by the real `vcenter.yaml` — it escaped the 11-op fix only because no lane currently asserts it.

## Deliverable

1. Add `_OP_GET_CLUSTER_DRS` (and any other read-composite op_ids the existing lanes don't cover — sweep `_read.py` exhaustively) to the vmware reconcile-lane assertions.
2. Fix the unserved path(s) per the #2970 pattern: locate what the pinned spec actually serves (vim analogue likely, cf. the DRS-recommendations decision in PR #2974); never invent a path; per-op decision with rationale in the PR body.

## Acceptance criteria

- [ ] Every op_id constant in `composites/_read.py` is asserted by a lane.
- [ ] All five (or six) vmware spec lanes green against the shelf; skip-clean without env.

Issue #2981, opened 2026-08-17
Task: nsx real-spec reconcile lane — assert hand-coded paths against the pinned nsx-9.0 spec


damir-topic:
Parent goal: #221
Parent initiative: #2979

Depends-on: #2980

## Problem

`backend/src/meho_backplane/connectors/nsx/` carries ~11 hand-coded vendor-API path literals that nothing asserts against the real product API. #2970 found 11 of 13 such vmware paths unserved by the real spec — same defect class, currently invisible here.

## Deliverable

1. A reconcile lane using the #2980 harness: enumerate every hand-coded `METHOD:/path` in the nsx connector (typed ops + any composite sub-ops) and assert each is served by the pinned `nsx-9.0` spec on the shelf; uniform skip-with-reason when the shelf is absent.
2. Widen the `ci.yml` spec-shelf sparse-checkout + verify step to include `docs/nsx-9.0` (one line each; expect trivial conflicts with sibling wave-3 tasks — rebase, don't battle).
3. Extend the `docs/decisions/vendor-spec-ci-provisioning.md` signoff with the one-paragraph nsx/nsx-9.0 entry per the #2980 standard doc.
4. If the lane is red on first run: fix mechanically-wrong paths in the same PR (the #2970 pattern — never invent a path; per-op decisions with rationale in the PR body); split a fix task only if a repoint needs a design decision. Never merge red.

## Acceptance criteria

- [ ] Lane green locally against the shelf; skip-clean without env.
- [ ] Any path fixes verified against the pinned spec, decisions recorded in the PR body.
- [ ] CI runs the lane (armed repo) — verify step passes with the widened checkout.

damir-topic:
**Signoff attestation (licensing determination of record — NSX + wave-3 blanket).**

Determination by the maintainer/operator (@damir-topic, 2026-08-18): *fetching the Broadcom/VMware NSX 9 OpenAPI specs (`nsx_api.json` / `nsx_policy_api.json`, obtained from the operator's own lab NSX manager and pinned to the private spec-shelf) into ephemeral CI jobs is acceptable under the operator's vendor license* — same conditions as the 2026-08-17 vSphere determination (evoila/meho#2949 comment 5316438336): sparse read-only checkout, ephemeral runner workspace, never committed to the public tree, never in artifacts or logs, same-repo runs only, secret withheld from fork PRs and the Dependabot store.

[...]

Issue #2949, opened 2026-08-17
Task: provision the vendor vCenter spec-shelf into CI so the real-spec ingest/reconcile lanes run on PRs (all skip today)


damir-topic:
Parent goal: #221

Split out of #2944 (see that ticket's rescope): wiring the vendor spec-shelf into CI is a licensing + secret-provisioning decision, not a code-only change, so it is its own task.

## Why

Five vCenter-spec-backed test lanes validate our OpenAPI ingest + op_id reconciliation against the **real** pinned `vcenter.yaml` / `vi-json.yaml` — but **every one `pytest.skip()`s in CI**, because the vendor-licensed spec-shelf is not provisioned there (verified: zero `MEHO_VCENTER_OPENAPI*` / `MEHO_CONSUMER_DOCS_ROOT` references in `.github/workflows/`):

- `backend/tests/acceptance/test_g07_vsphere_canary.py` — the G0.7 vSphere ingest + dispatch canary
- `backend/tests/integration/test_operations_ingest_vcenter.py` + `test_operations_ingest_vi_json.py` — real-spec ingest
- `backend/tests/acceptance/test_portgroup_audit_op_id_reconcile.py` — the #1602 read-side audit reconcile
- `backend/tests/test_connectors_vmware_rest_composites_l2_ingest_reconcile.py` — the #2944 write-composite `_SUB_OPS_*` real-spec reconcile guard

So the "does every declared op_id / sub-op path actually exist in the real vCenter spec?" check runs only where an operator has the shelf locally — never automatically on a PR. A shape-correct-but-nonexistent path (typo, renamed endpoint, wrong API version) sails through CI and only fails against a live vCenter. Provisioning the shelf into CI **once** lights up all five lanes together.

The resolver is already built: `backend/tests/acceptance/_vcenter_spec.py` resolves the specs from either `MEHO_VCENTER_OPENAPI_VCENTER` + `MEHO_VCENTER_OPENAPI_VI_JSON` (explicit paths) or `MEHO_CONSUMER_DOCS_ROOT` (a directory), and the tests skip cleanly when unset. This task only has to make the specs available in CI and set those env vars — no test or resolver change.

## The decision this ticket is really about

The vCenter OpenAPI specs (`vcenter.yaml` ~961 paths, `vi-json.yaml` ~2,195 paths) are **Broadcom/VMware vendor-licensed** and deliberately NOT committed to this public repo — they live in the operator's private spec-shelf. Wiring them into public-repo CI means:

[...]

PR #2995, opened 2026-08-17
feat(tests): generalize the spec-lane harness + record the spec-reconcile-guard standard (#2980)

### LINKED ISSUE #2980: Task: generalize the spec-lane harness + record the spec-reconcile-guard standard (decision doc)
Parent goal: #221
Parent initiative: #2979

## Problem

The shelf-lane machinery is vcenter-shaped: `backend/tests/acceptance/_vcenter_spec.py` resolves exactly one spec dir, and the reconcile-lane pattern (`test_connectors_vmware_rest_composites_l2_ingest_reconcile.py`) hard-codes it. Every wave-3 lane task would otherwise re-invent resolution, skip semantics, and path-extraction.

## Deliverable

1. A parameterized shelf resolver (e.g. `_spec_shelf.py`): given a product dir name (`nsx-9.0`, …) under `MEHO_CONSUMER_DOCS_ROOT`, return the parsed spec or a uniform skip-with-reason. Keep the vcenter lanes' behavior byte-identical (they migrate to it or stay — decide, don't duplicate).
2. A reusable assert helper: given a set of `METHOD:/path` strings and a parsed OpenAPI spec, report unserved entries with actionable diffs (the #2944 lane's core, extracted).
3. `docs/decisions/spec-reconcile-guards-standard.md`: the standard (every hand-coded vendor path asserted against a pinned spec where one exists; ingested connectors immune by construction), the extension mechanics (sparse-checkout widening, verify step, per-vendor signoff extension in `vendor-spec-ci-provisioning.md`), and the red-lane-is-a-finding protocol (#2944/#2970 precedent).

## Acceptance criteria

- [ ] Existing five vcenter lanes still green against the shelf and skip-clean without env.
- [ ] A trivial example lane (any product) demonstrates the harness in <30 lines.
- [ ] Decision doc filed; `docs/codebase/` index updated if the testing surface doc exists.

damir-topic:
## Summary
- Generalizes the vcenter-only spec-shelf machinery into the product-agnostic harness all wave-3 reconcile lanes (#2981-#2993) build on: `backend/tests/_spec_shelf.py` — `require_shelf_spec` (uniform skip-with-reason for any `<product-dir>/<file>` under `MEHO_CONSUMER_DOCS_ROOT`), `openapi_served_op_ids` (served set via the real ingest parser, so op_ids are byte-for-byte `endpoint_descriptor.op_id` strings), `assert_op_ids_served` (actionable diff with near-miss repoint leads + vacuous-empty guard).
- `tests/acceptance/_vcenter_spec.py` keeps its public API, reason string, and explicit/legacy env-var priority; only its shelf-root branch delegates to the shared resolver — the five vcenter lanes are untouched and behavior-identical.
- Files the standard: `docs/decisions/spec-reconcile-guards-standard.md` (every hand-coded vendor path asserted against a pinned spec where one exists; ingested connectors immune by construction; extension mechanics; red-lane-is-a-finding protocol per #2944/#2970). `docs/codebase/devops.md` spec-shelf section updated to match.
- Example lane `tests/test_spec_shelf_example_lane.py` (29 lines) is the copy-paste template for the sibling lanes and doubles as the harness's CI-armed integration proof against the real `vcenter-9.0` shelf.

## Test plan
- [x] `ruff check src/ tests/` — clean; `ruff format --check` — clean
- [x] `mypy src/` — no src changes; sole error (`net/icmp.py` `MSG_ERRQUEUE`) is a pre-existing Darwin-only typeshed artifact, absent on Linux CI
- [x] Acceptance: five vcenter lanes green against the shelf and skip-clean without env, **byte-identical to main** — with `MEHO_CONSUMER_DOCS_ROOT`: 35 passed / 28 skipped on both this branch and main; without env: 22 passed / 41 skipped on both; skip reason strings unchanged
- [x] Acceptance: example lane demonstrates the harness in <30 lines (29) — green against the local shelf, skips clean without env
- [x] Acceptance: decision doc filed; no dedicated testing-surface doc exists in `docs/codebase/` (verified), `devops.md` updated instead
- [x] New harness unit tests: 13 passed (resolver skip/None semantics, parser extraction incl. `?action=` keying, actionable-red / near-miss / empty-guard, vcenter delegation pins)
- [x] Full CI-equivalent unit sweep (`-n 3 --dist loadscope`, ignore integration/migrations): 11578 passed; only failures are pre-existing environment flakes reproducing identically on clean main (`net_icmp` loopback trace, `net_dns_lookup` resolver, chart mcp env — all Darwin/local-env, all unrelated to this diff)
- [x] `.claude/hooks/code-quality.py --diff` — exit 0

## Out of scope
- Per-vendor lanes, ci.yml sparse-checkout widening, and signoff extensions — the sibling tasks #2981-#2993 (each is one `sparse-checkout` + one verify line + one signoff paragraph, per the standard doc).
- `test_operations_ingest_vcenter.py`'s self-contained legacy resolver — predates `_vcenter_spec.py`, has its own env-var contract; migrating it would churn a required-gate lane for zero behavior change.
- vmware `_OP_GET_CLUSTER_DRS` residual — #2986.

Closes #2980

---

## Follow-up (auto-implement-issue, fix iteration 1, 2026-08-17)

Armed CI (run 32038537229) errored `test_canary_ingest_meets_operation_count` at setup: `InvalidSpecError: spec URI must use the https scheme; got ''`.

**Root cause — not this PR's delegation.** The canary constructed content-less `SpecSource(uri=str(<shelf path>))` entries; the backend ingest path is https-only by design (G0.16-T8 #95 SSRF guard), so a bare local path is rejected at the scheme check before any read. This was latent since the guard landed: with no CI spec shelf the spec fixtures always skipped. The first run after `SPEC_SHELF_TOKEN` was provisioned (#2949/#2966, 2026-08-17 afternoon) surfaced it on **every** branch — run 32044865650 on `feat/2862-targets-add-verb` (pre-#2980 resolver, no delegation) fails identically at 16:24, and the "passing" run for #2978 (13:09) never ran the shelf checkout (token absent), so its canary skipped. The #2980 delegation is behavior-identical (see the delegation pins in `tests/test_spec_shelf_harness.py`).

**Fix** — `0e08e3e5`: the canary's four `SpecSource` sites now read the resolved spec file client-side and upload the bytes via `content=` (helper `_inline_spec_source`), the same on-ramp the CLI uses for `docs:`/`file://` sources (#1535/#1572) and `tests/integration/test_operations_ingest_vcenter.py` has used since the guard landed. The bare path stays the `uri` audit label, so the `spec:<path>` row tags are unchanged and provenance classifies as `inline`.

**Re-verified** against the real shelf: `read_spec_info_version` resolves `9.0.0.0` for both specs (version cross-check vs `9.0`: `exact`); `parse_openapi_with_provenance` emits 1275 / 2195 ops (floors 1200 / 2000); five vcenter lanes 35 passed / 28 skipped with `MEHO_CONSUMER_DOCS_ROOT`, 22 / 41 without — byte-identical to main; canary with only `MEHO_CONSUMER_DOCS_ROOT` set: no setup error (remaining subtests are Docker/opt-in gated locally, exercised by armed CI).

Note for other open PRs: every armed CI run on any branch fails this way until this fix reaches main.

---

## Follow-up (auto-implement-issue, fix iteration 2, 2026-08-17)

With the spec now parsed (6711 collected vs 6336), armed CI (run 32048056557) advanced to the next layer: `test_canary_ingest_meets_operation_count` errored at setup in the LLM grouping pass — `LlmOutputInvalid`: `group_key 'vm-managed-objects' must match '^[a-z][a-z0-9_]*$'` raised in `parse_proposal_response` (Pass-1 parser, `_llm_grouping_internals.py:438`).

**Diagnosis: B — deterministic stub output, not a live LLM call.** The unit-job canary fixture chain (`ingested_canary`) drives `IngestionPipelineService` with `llm_client_factory=lambda: _PathPrefixStubLlmClient()` — a hand-authored in-test stub whose static Pass-1 taxonomy (`_PROPOSE_RESPONSE`) is exactly the "rich, plausible" array in the CI log. The live-LLM variant (`real_llm_ingested_canary`) is gated behind `MEHO_G07_CANARY_LIVE_LLM=1` + `ANTHROPIC_API_KEY`, which `ci.yml` never sets; no LLM credential exists in the unit job. The output is byte-deterministic across runs.

**Root cause.** The stub's taxonomy declared four vi-json Managed-Object groups with kebab-case `group_key`s (`vm-managed-objects`, `host-managed-objects`, `cluster-managed-objects`, `datastore-managed-objects`) — introduced by #520 (G3.1-T3 two-spec canary) and violating the snake_case contract the production `propose_groups.md.j2` prompt mandates and `_GROUP_KEY_PATTERN` enforces. Latent for the same reason as iteration 1: the ingest never actually ran in CI until the shelf was armed (#2949/#2966), and never runs locally without Docker. Not introduced by this PR — this branch only touched `SpecSource` construction; `git log -S` pins the keys to #520 on main.

**Fix** — `eaccb895`:
- Renamed the four keys to snake_case in both `_PROPOSE_RESPONSE` and `_PATH_RULES` (Pass-1 proposals and Pass-2 assignments stay consistent; the validator is untouched).
- Added `test_stub_taxonomy_passes_production_proposal_validation` — fixture-less (no DB, no shelf), round-trips the stub's static taxonomy through the production `parse_proposal_response` and checks every `_PATH_RULES` target is a proposed key (or `none`). This is the capture-time-validation gap made concrete: stub drift of this class now fails at unit speed in every environment instead of only in a shelf-armed, Docker-backed CI run.

**Re-verified locally:** chokepoint proof — the pre-fix kebab payload raises the byte-identical `LlmOutputInvalid` frame, the fixed taxonomy parses 14 proposals; five vcenter lanes 36 passed / 28 skipped with `MEHO_CONSUMER_DOCS_ROOT`, 23 / 41 without (iteration-1's 35/28 and 22/41 plus exactly the one new fixture-less self-check, which runs in both matrices); harness+example lane 14 passed (byte-identical); canary-only with shelf env: 1 passed / 28 skipped, 0 errors; `ruff check` + `ruff format --check` clean; `code-quality.py --diff` exit 0. Only CI can prove the full DB-backed two-spec ingest (Postgres testcontainers + shelf); the grouping pass it runs now consumes the validated taxonomy.

**Adjacent finding (not implemented here):** the production pipeline has no retry budget and no kebab→snake normalization on `LlmOutputInvalid` — `llm_groups.py` documents that the orchestrator deliberately does not catch it, so a *live* model emitting one kebab key fails the entire ingest of any future connector with only the operator-facing error envelope as recourse. The Pass-1 prompt already mandates snake_case; whether a deterministic normalization (kebab→snake before validation, validator kept as backstop) or a bounded re-prompt is warranted is a product-scope question for a follow-up ticket, out of scope for this test-harness PR.

---

## Follow-up (auto-implement-issue, fix iteration 3, 2026-08-17)

Armed CI's next layer: with iteration 1+2's fixes in place, the canary's full two-spec ingest ran inside the **unit** sweep for the first time ever — and the "Python (ruff + mypy + pytest)" job (timeout-minutes: 25) was timeout-killed on four consecutive attempts on head `eaccb895` (~25m20s wall each; the last, 19:47–20:12Z, on a quiet cluster with pytest at 81% and progressing). Baseline: the 16:57Z run (canary still erroring fast at the grouping layer) finished the whole sweep in 478 s. Cause: the `ingested_canary` fixture is function-scoped by design (the module docstring documents why), so each of the canary's 25 armed test cases re-runs the full ~3,470-op ingest (parse + register + real embeddings + grouping) — minutes per test, and `--dist loadscope` serializes the whole module onto one worker.

**Fix — lane relocation, per ci.yml's own perf-budget discipline (#898/#793/#827: relocate/amortize the slow path, never bump `timeout-minutes` or `budget_s`):**

- The canary's 13 full-ingest test functions (25 cases) now carry a collection-time `skipif` on `MEHO_SKIP_SPEC_INGEST_TESTS` (strict truthy parsing mirroring `MEHO_RUN_SLOW_TESTS`), with a fixture-level backstop in `vcenter_spec_path` for any future heavy test that forgets the marker. Collection-time (not fixture-side) so an opted-out lane never spins the module-scoped Postgres container the fixture chain would otherwise instantiate first.
- `ci.yml` `python-lint-test` (and the push-only `python-coverage` sweep, which re-runs the same selection serially) set `MEHO_SKIP_SPEC_INGEST_TESTS: "1"`.
- `ci.yml` `python-integration` now selects `tests/acceptance/test_g07_vsphere_canary.py` alongside `tests/integration/` — the single armed lane for the full ingest (60-min cap, containers first-class, already green with the canary-era work at 17:19–17:32Z). Estimated added wall: ~15–20 min serial, well under cap.
- The fixture-less self-checks (`test_stub_taxonomy_passes_production_proposal_validation`), the example lane, the harness tests, and every seconds-cheap reconcile lane **stay in the unit sweep** — no armed coverage is lost anywhere: every shelf lane still runs on every same-repo PR, each in exactly one armed lane.
- Lane-placement rule recorded in `docs/decisions/spec-reconcile-guards-standard.md` §"Lane placement (CI budget)" so the 12 dependent lane tasks (#2981–#2993) inherit the distinction explicitly: their reconcile lanes are parse-only and seconds-cheap — they live in the unit sweep and need **no gating**; only full-ingest canaries take the integration-lane placement. `docs/codebase/devops.md` spec-shelf section updated to match.

**Verified locally** (no Docker — the armed heavy path itself is CI-proven by design, as in iterations 0–2):

- Canary module, shelf env + `MEHO_SKIP_SPEC_INGEST_TESTS=1` (CI unit-lane matrix): 1 passed / 28 skipped in 1.7 s — exactly 25 cases skipped with the opt-out reason, 3 with their pre-existing opt-in reasons, the taxonomy self-check runs.
- Canary module, shelf env, no opt-out var (local-dev / integration matrix): 1 passed / 28 skipped — byte-identical counts and reasons to `eaccb895`.
- Five vcenter lanes, shelf env, no opt-out: **36 passed / 28 skipped** — byte-identical to the iteration-2 baseline; without env: **23 passed / 41 skipped** — byte-identical. Harness + example lane: 14 passed.
- Integration-lane selection (`pytest tests/integration/ tests/acceptance/test_g07_vsphere_canary.py --collect-only` with shelf env): 333/334 collected (1 = the pre-existing `load`-marker deselection), all 29 canary tests in the selection.
- Full CI-equivalent unit sweep (`-n 3 --dist loadscope --maxfail=1`, shelf env + opt-out env): **11,477 passed / 167 skipped, wall 583 s (9:41)** on the local dev box — back in the pre-armed envelope (CI baseline 478 s; `budget_s` 1320). Only failures: the same pre-existing Darwin/local-env flakes documented since the fresh iteration (`net_icmp` loopback trace, `net_dns_lookup` resolver, chart mcp env — re-run in isolation, reproduce identically to clean main) plus one load-flaky event-loop-lag assertion (`test_ingest_large_spec_event_loop`, 0.5 s wall-clock ceiling under 3 saturated workers; 3/3 passed re-run in isolation, module untouched by this diff).
- `ruff check` + `ruff format --check` clean; `code-quality.py --diff` exit 0; ci.yml parses clean.

**Sequencing note for the orchestrator (#3001):** PR #3001 extracts only the iteration-1 inline-content canary fix. If #3001 merges **without** this PR's lane relocation, main trades the canary's fast setup ERROR for this same unit-lane 25-min TIMEOUT on every armed run (the inline-content fix is exactly what lets the full ingest proceed into the unit sweep). #3001 should either merge together with / after this PR, or cherry-pick the lane-relocation commit as well.

---

## Follow-up (auto-implement-issue, fix iteration 4 — operator-authorized, 2026-08-18)

**One remaining failure** (run 32067450471, integration lane, 1 of 303): `test_canary_govc_parity_benchmark[list-clusters]` — `GET:/vcenter/cluster` lost its raw top-3 to three vi-json `ClusterComputeResource` ops. The operator chose **re-scope over xfail**: the benchmark asserted raw corpus-wide search — the flow architecture postulate 4 tells agents *not* to use at thousands-of-ops scale. Raw-corpus relevance is now #3006's question; the benchmark asserts the sanctioned flow (pick connector → list groups → search within the group).

**The re-scope** (`backend/tests/acceptance/test_g07_vsphere_canary.py` + `docs/cross-repo/g07-vsphere-canary.md`):

- `GOVC_PARITY_BENCHMARK` is now `(query, group, expected_op_id)` triples; each case drives `search_operations` with the real `group` filter (`meta_tools.py` → `resolve_group_id` → `hybrid_search(group_id=...)`) scoped to the stub-taxonomy group an agent would pick. Case ids unchanged, so CI history stays comparable; collection node-IDs verified byte-identical to `08d3e1af` (29 tests).
- Measured against the real two-spec corpus (two independent local ingest+probe samples, Docker PG + pgvector, real bge-small embeddings): **9 of 13 cases pass group-scoped** — `list clusters` (the CI failure) at rank **1** both samples; datacenter/datastore/network/host/session/perf at ranks 1–2; `get virtual machine info` at rank 3. Group-scoped ranks were identical across both samples; raw ranks would have failed **7 of 13** cases, confirming the postulate-4 re-scope beyond the single CI data point.

**Real findings surfaced by the re-scope (not papered over):**

1. **Benchmark case `revert vsphere snapshot` was mis-authored.** Its expected op `POST:/VirtualMachine/{moId}/RevertToSnapshot_Task` **does not exist in vi-json.yaml** — in the corpus (as in the VIM API) `RevertToSnapshot_Task` is a method of `VirtualMachineSnapshot`; `VirtualMachine` only carries `RevertToCurrentSnapshot_Task`. The case could never pass and was latent because the integration lane runs `pytest -x`, which stopped on `list-clusters` before this case ever executed — it has never actually run armed. Corrected to the spec truth `POST:/VirtualMachineSnapshot/{moId}/RevertToSnapshot_Task` (group `vm_managed_objects`), which passes at group-scoped rank 2 (both samples).
2. **`tail vsphere events` misses top-3 even group-scoped — stable in-group rank 4** (both samples) of the 8-op `events` group: "tail" matches no EventManager op text, "events" matches all of them, and the short property-reads (`latestEvent`, `description`) plus `LogUserEvent` out-score `QueryEvents` on BM25 text density. Also never previously executed armed (same `-x` shadow). This is a genuine vi-json per-op description-quality finding, xfail-documented (non-strict) with the measured rank in the reason string — same limitation class (weak per-op text, no T3 per-op `llm_instructions`) as the vcenter cardinal trio; recorded in *Known gaps* §2 and feeding #3006's evidence.
3. **The vm-cardinal trio misses top-3 even inside its own group** — `GET:/vcenter/vm` at in-group rank 7, `power?action=start/stop` at 5–6, within the 148-op `vm` group (both samples): group scoping removed the cross-spec noise but not the intra-group sub-path crowding. The pre-existing xfail markers stay, with reasons upgraded to the measured group-scoped ranks. Fix path remains per-op description enrichment (T3 `llm_instructions` follow-up), not search-flow changes.
4. **Opportunistic latent-bug fix in touched code:** the opt-in real-LLM eyeball (`test_g07_canary_real_llm_eyeball`, env-gated, never in CI) iterated all 13 benchmark queries against its vcenter-only ingest — the 3 vi-json expected ops could never appear, so any opt-in run was guaranteed to fail. It now iterates the 10 vcenter queries, and deliberately stays **raw** (it measures exactly #3006's question — whether LLM-curated hints lift corpus-wide ranking).

Per the operator's constraint: no raw-search hard assert was added and no raw-search xfail exists in the file — the four xfails are group-scoped, measured, documented product limitations, each with its specific reason.

**Verified locally** (Docker PG + real embeddings + real specs from the shelf clone):

- Re-scoped pytest cases run for real (full two-spec ingest each, 408 s wall): `[list-clusters]` **PASSED** (the CI failure), `[revert-vsphere-snapshot]` **PASSED** (corrected op), `[tail-vsphere-events]` **XFAIL** with the measured reason — `2 passed, 1 xfailed`, exit 0.
- Opted-out lane (`MEHO_SKIP_SPEC_INGEST_TESTS=1`): 1 passed / 28 skipped — byte-identical to `08d3e1af` (unit-lane gating untouched).
- Collection: 29 node-IDs byte-identical to `08d3e1af`; with-shelf / env-less lane matrices untouched by construction (no test added/removed/renamed).
- `ruff check` + `ruff format --check` clean; `code-quality.py --diff` exit 0.

---

## Follow-up (auto-implement-issue, iteration 5, 2026-08-18)

Addressed review finding **B1** from iteration 4 ([review](https://github.com/evoila/meho/pull/2995#issuecomment-5326008028), [inline](https://github.com/evoila/meho/pull/2995#discussion_r3802624574)) — CONFIRMED empirically by run 32117732263: the integration merge gate was timeout-killed at 60m19s on head `d150d28d` because the function-scoped `ingested_canary` fixture re-ran the full two-spec ingest (measured ~164 s on CI runners) for each of the 25 armed tests: ~68 min of ingest against a 60-min cap. Direction followed exactly: **amortize the ingest; never raise `timeout-minutes`**.

**B1** `4146b4f6` — shared-corpus amortization:

- **Module-scoped `_canary_corpus`** ingests the shelf specs ONCE per module. One corpus (not two): all 25 armed cases consume the same two-spec state; single-spec fallback stays inside it. It is a *sync* fixture running the async pipeline via `asyncio.run` on a private event loop — module-scoped *async* fixtures would couple to pytest-asyncio loop-scope config, and an engine's pooled asyncpg connections are loop-affine, so every DB resource is created and disposed inside that one loop. Module-scoped fixtures instantiate before all function-scoped autouse pins, so the fixture pins its own env (`_CHASSIS_ENV`, container `DATABASE_URL`, fastembed cache) via a module-lifetime `pytest.MonkeyPatch`, brackets its own structlog silencing (leak-sweep protection), and resolves the real embedding singleton once (also deleting the old per-test ONNX reload).
- **Thin function-scoped `ingested_canary` binder** — same name, same `_CanaryIngestState`; all 25 test signatures and node-IDs unchanged. Per test it re-pins a fresh engine to the same module-scoped container **without truncating**: the repo has no shared-ingested-corpus precedent (the closest are the module-scoped container fixtures: `async_pg_url`, valkey/harbor/k8s), and both `pg_engine` fixtures TRUNCATE per test — exactly what forced function scope before (the fixture's old docstring said so). The root conftest's autouse `_default_database_url` also disposes/resets the engine cache after every test, so per-test re-pinning is mandatory regardless. The pattern is now documented in `docs/decisions/spec-reconcile-guards-standard.md` §Lane placement as the rule for future heavy ingest lanes.
- **Isolation semantics:** all 25 armed cases are read-only against the corpus (SELECTs, DB-backed meta-tool reads, pure-function checks) — verified per test body; no reset between cases is needed. State-mutating tests (the env-gated vcsim-dispatch and real-LLM opt-ins, never in CI) keep their own function-scoped ingest against the truncating `pg_engine`, and are defined after every shared-corpus consumer. The corpus tests run with an empty v2 registry (the autouse `_reset_module_state` clears it before each body); verified safe — `_require_dispatchable_connector` short-circuits on `connector_exists` (DB) before any registry lookup.
- **Lane gating untouched:** `MEHO_SKIP_SPEC_INGEST_TESTS` semantics, the collection-time markers, and every CI lane selection are unchanged (the ci.yml delta is comment-only). Opted-out lane re-verified: 1 passed / 28 skipped in 0.82 s, no container. Collection node-IDs byte-identical to `d150d28d` (29 tests). Skip reasons and counts identical in every lane.

**Real finding surfaced (not caused) by B1's fix** — making the full armed module reachable for the first time (`pytest -x` had died on `list-clusters` in both armed runs) exposed one latent case with zero armed execution history: `test_canary_two_spec_grouping_unassigned_ratio` fails its `< 50%` bar against the real corpus — **2831/3470 ops unassigned (81.6%**; per-spec: vcenter 949/1275, vi-json 1882/2195; measured via an instrumented run). Reproduced **identically on the pre-amortization `d150d28d` shape** (86.79 s single-test run, same assertion failure) — latent mis-calibration, not a regression: the `< 50%` bar was authored from the 14-family stub taxonomy's coverage claim without an armed measurement, and the real specs' long tail (vcenter `appliance/`/`esx/`/`hvc/`/`content/`; ~100 vi-json MO types beyond the six mapped) classifies to `none` by design. Handled exactly like iteration 4's `tail-vsphere-events`: xfail (non-strict) with the full measured numbers in the reason; the assertion still executes every armed run; the substrate is proven by the 639 assigned ops populating all 14 groups. Documented in *Known gaps* §4 of `docs/cross-repo/g07-vsphere-canary.md`. The four xfails from `d150d28d` are unchanged.

**Projected integration-lane math** (from measured per-phase costs):

| Phase | Cost |
|---|---|
| Integration baseline (run 32049795686, pre-canary) | ~710 s |
| Canary module container boot + migrations + model load | ~30 s |
| ONE shared two-spec ingest (measured ~164 s on CI runners) | ~164 s |
| 25 read-only case bodies + per-test engine re-pins (measured: whole armed module locally = 90.7 s incl. ingest → bodies ≈ 25 s; CI margin ×4) | ~100 s |
| **Projected total** | **~1,000 s ≈ 17 min** |

vs. ~59 min available under `timeout-minutes: 60` — ~42 min of honest headroom (target was ≤40 min total). The previous shape projected ~80 min.

**Verified locally** (Docker PG + real embeddings + real shelf specs):

- Full armed module, single shared ingest: **21 passed, 3 skipped (opt-in), 4 xfailed, 1 xpassed in 90.7 s** (the xpass is a benchmark case drifting into top-3 — pgvector IVFFlat variance, exactly why the markers are non-strict).
- Latent-failure adjudication: same single test on stashed `d150d28d` → same failure (86.79 s).
- Opted-out lane: 1 passed / 28 skipped, 0.82 s, no container.
- Collection node-IDs byte-identical to `d150d28d`; `ruff check` + `ruff format --check` clean; `code-quality.py --diff` exit 0; ci.yml parses (comment-only delta).

What only CI can prove: the wall-clock of the shared ingest on merge-gate runners (~164 s measured there previously vs ~60 s locally) and the end-to-end integration job total under the 60-min cap — this PR's next run adjudicates the projection.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -246,7 +246,10 @@ jobs:
       # Dependabot runs, not-yet-provisioned repos) the directory does
       # not exist, the resolvers return None, and the spec-backed lanes
       # pytest.skip exactly as they did before #2949. When the checkout
-      # ran, the lanes run for real. The docs-root form (not
+      # ran, the seconds-cheap reconcile lanes run for real in this job;
+      # the G0.7 canary's full-ingest tests are opted out here and run
+      # armed in python-integration instead (#2980 — see this job's
+      # MEHO_SKIP_SPEC_INGEST_TESTS env row). The docs-root form (not
       # MEHO_VCENTER_OPENAPI_VCENTER / _VI_JSON) is deliberate: it is
       # the only single setting that lights every spec lane, because
       # test_operations_ingest_vcenter.py predates the explicit
@@ -478,6 +481,26 @@ jobs:
           # (en_core_web_lg) once that lane provisions the heavier
           # model.
           MEHO_REDACTION_SPACY_MODEL: en_core_web_sm
+          # Lane placement (#2980): opt this sweep out of the G0.7
+          # canary's full-ingest spec tests. With the spec shelf armed
+          # (#2949/#2966) each of those tests re-runs the full two-spec
+          # ingest (~3,470 ops through parse + register + embedding +
+          # grouping — the fixture is function-scoped by design), which
+          # took the whole sweep from 478 s (2026-08-17 16:57Z, canary
+          # still erroring fast) past this job's 25-min cap (four
+          # consecutive timeout kills on head eaccb895, the last on a
+          # quiet cluster with pytest at 81% and progressing). Per this
+          # workflow's perf-budget discipline (#898/#793/#827): the slow
+          # path is relocated, not the cap raised. The canary's
+          # full-ingest tests run armed in the python-integration job
+          # (which selects the file explicitly, see its Pytest step);
+          # its fixture-less self-checks and every seconds-cheap
+          # reconcile lane stay in this sweep. Honored by a
+          # collection-time skipif marker in tests/acceptance/
+          # test_g07_vsphere_canary.py (fixture-level backstop in
+          # vcenter_spec_path); rule of record in
+          # docs/decisions/spec-reconcile-guards-standard.md.
+          MEHO_SKIP_SPEC_INGEST_TESTS: "1"
         # PARALLELIZED via pytest-xdist (#585 unblocked this). The
         # serial sweep was the job's dominant cost (~49 min in CI vs
         # ~5 min local). #585 added process-wide global-registry
@@ -758,6 +781,15 @@ jobs:
         timeout-minutes: 45
         env:
           MEHO_REDACTION_SPACY_MODEL: en_core_web_sm
+          # Same lane-placement opt-out as the required unit lane's
+          # Pytest step (#2980) — this job re-runs that sweep, serially
+          # and under a 45-min step cap, so the canary's full-ingest
+          # tests are opted out here for the same budget reason. Today
+          # this is belt-and-suspenders (this job has no spec-shelf
+          # checkout, so those tests skip on shelf absence anyway);
+          # it keeps a future shelf-arming of this job from silently
+          # re-importing the multi-minute ingest wall.
+          MEHO_SKIP_SPEC_INGEST_TESTS: "1"
           # Each xdist worker is a subprocess; coverage's a1_coverage.pth hook
           # calls coverage.process_startup() when COVERAGE_PROCESS_START points
           # at a config, so every worker writes its OWN parallel data file
@@ -954,14 +986,28 @@ jobs:
         run: uv sync --locked --all-groups
 
       - name: Pytest (integration testcontainers)
-        # Only the tests/integration/ subtree. No --cov here: coverage
-        # is emitted by the unit job (see its NOTE); combined coverage
-        # is a #564 follow-up. -x for fast-fail on the first broken
-        # integration case. Deliberately NOT parallelized: the unit
-        # job is the ~49min bottleneck (now split-parallelized); this
-        # job is ~6min and its container-heavy fixtures under xdist
-        # are unverified — out of scope for the perf change.
-        run: uv run pytest -x tests/integration/
+        # The tests/integration/ subtree plus the G0.7 vSphere canary
+        # module. No --cov here: coverage is emitted by the unit job
+        # (see its NOTE); combined coverage is a #564 follow-up. -x for
+        # fast-fail on the first broken integration case. Deliberately
+        # NOT parallelized: the unit job is the ~49min bottleneck (now
+        # split-parallelized); this job is ~6min and its
+        # container-heavy fixtures under xdist are unverified — out of
+        # scope for the perf change.
+        #
+        # The canary module is selected here (#2980) because its
+        # armed-shelf full-ingest tests cost minutes per test (full
+        # two-spec ingest per test by design) — they blew the unit
+        # lane's 25-min cap the first day the shelf was armed, so this
+        # job (60-min cap, containers first-class) is their single
+        # armed lane; the unit sweep opts out via
+        # MEHO_SKIP_SPEC_INGEST_TESTS (see its Pytest step's env
+        # comment). Shelf absent (fork/Dependabot/unprovisioned) → the
+        # canary skips exactly as it always did; its fixture-less
+        # self-checks also run (again) here — seconds, harmless. Lane
+        # placement rule: docs/decisions/spec-reconcile-guards-
+        # standard.md.
+        run: uv run pytest -x tests/integration/ tests/acceptance/test_g07_vsphere_canary.py
 
   python-migration-tests:
     name: Python (database migrations)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -996,17 +996,20 @@ jobs:
         # scope for the perf change.
         #
         # The canary module is selected here (#2980) because its
-        # armed-shelf full-ingest tests cost minutes per test (full
-        # two-spec ingest per test by design) — they blew the unit
-        # lane's 25-min cap the first day the shelf was armed, so this
-        # job (60-min cap, containers first-class) is their single
-        # armed lane; the unit sweep opts out via
+        # armed-shelf full-ingest tests need the container-heavy
+        # budget: the ~164 s two-spec ingest runs ONCE per module
+        # (module-scoped _canary_corpus fixture, amortised in PR #2995
+        # after per-test ingests — 25 x ~164 s — timeout-killed this
+        # job's 60-min cap on the first armed run; per-test ingests
+        # had already blown the unit lane's 25-min cap the day the
+        # shelf was armed). This job is the canary's single armed
+        # lane; the unit sweep opts out via
         # MEHO_SKIP_SPEC_INGEST_TESTS (see its Pytest step's env
         # comment). Shelf absent (fork/Dependabot/unprovisioned) → the
         # canary skips exactly as it always did; its fixture-less
         # self-checks also run (again) here — seconds, harmless. Lane
-        # placement rule: docs/decisions/spec-reconcile-guards-
-        # standard.md.
+        # placement rule + amortisation pattern:
+        # docs/decisions/spec-reconcile-guards-standard.md.
         run: uv run pytest -x tests/integration/ tests/acceptance/test_g07_vsphere_canary.py
 
   python-migration-tests:
--- a/backend/tests/_spec_shelf.py
+++ b/backend/tests/_spec_shelf.py
@@ -1,0 +1,188 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""Product-agnostic spec-shelf resolver + reconcile-assert harness (#2980).
+
+Vendor API specs are pinned in the operator's private spec-shelf repo
+(``evoila-bosnia/claude-rdc-hetzner-dc``, ``docs/<product>-<version>/``)
+rather than this public chassis repo, so the vendor-licensed schema
+corpus never lands in public history. CI provisions the shelf through a
+secret-gated sparse checkout and points ``MEHO_CONSUMER_DOCS_ROOT`` at
+it (see ``docs/decisions/vendor-spec-ci-provisioning.md``); local dev
+sets the same env var at a full shelf clone.
+
+This module generalises the vcenter-only resolver
+(:mod:`tests.acceptance._vcenter_spec`, which now delegates its
+shelf-root branch here) into the harness every per-vendor reconcile
+lane builds on (#2981-#2993). A lane is three calls:
+
+    spec_path = require_shelf_spec("nsx-9.0", "nsx-openapi.json")
+    served = openapi_served_op_ids(spec_path)
+    assert_op_ids_served(declared, served, spec_label="nsx-9.0/nsx-openapi.json")
+
+Skip semantics are uniform: when ``MEHO_CONSUMER_DOCS_ROOT`` is unset,
+or the product directory / spec file is absent from the configured
+shelf (e.g. CI's sparse checkout has not been widened to that product
+yet), :func:`require_shelf_spec` skips with a reason naming the exact
+missing file and this module — never fails. Wherever the shelf is
+wired, the lane runs for real; a red lane there is the guard surfacing
+a real finding, not harness noise (the #2944/#2970 protocol — see
+``docs/decisions/spec-reconcile-guards-standard.md``).
+"""
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+import pytest
+
+from meho_backplane.operations.ingest import parse_openapi
+
+if TYPE_CHECKING:
+    from collections.abc import Collection, Iterable
+
+__all__ = [
+    "SHELF_ENV_VAR",
+    "assert_op_ids_served",
+    "openapi_served_op_ids",
+    "require_shelf_spec",
+    "resolve_shelf_file",
+    "shelf_skip_reason",
+]
+
+#: The single env var every shelf-backed lane resolves through. CI sets
+#: it unconditionally on both Python jobs; when the secret-gated shelf
+#: checkout was skipped the directory does not exist and lanes skip.
+SHELF_ENV_VAR = "MEHO_CONSUMER_DOCS_ROOT"
+
+#: Hint cap per missing entry in :func:`assert_op_ids_served` failures.
+_NEAR_MISS_LIMIT = 3
+
+
+def shelf_skip_reason(product_dir: str, filename: str) -> str:
+    """Uniform skip reason for an unconfigured / unprovisioned shelf spec."""
+    return (
+        f"{product_dir}/{filename} spec not configured. Set {SHELF_ENV_VAR} to a "
+        f"spec-shelf docs directory containing {product_dir}/{filename}. "
+        "See tests/_spec_shelf.py for the resolver contract."
+    )
+
+
+def resolve_shelf_file(product_dir: str, filename: str) -> Path | None:
+    """Return the local path to ``<shelf>/<product_dir>/<filename>``, or ``None``.
+
+    ``None`` covers every unprovisioned shape the same way: env var
+    unset, shelf root missing, product directory absent (sparse
+    checkout not widened), or the file itself absent. Callers convert
+    ``None`` into a ``pytest.skip`` with :func:`shelf_skip_reason` —
+    or use :func:`require_shelf_spec`, which does both.
+    """
+    shelf_root = os.getenv(SHELF_ENV_VAR)
+    if not shelf_root:
+        return None
+    candidate = Path(shelf_root).expanduser() / product_dir / filename
+    return candidate if candidate.is_file() else None
+
+
+def require_shelf_spec(product_dir: str, filename: str) -> Path:
+    """Resolve a shelf spec file or skip the calling test with the uniform reason."""
+    resolved = resolve_shelf_file(product_dir, filename)
+    if resolved is None:
+        pytest.skip(shelf_skip_reason(product_dir, filename))
+    return resolved
+
+
+def openapi_served_op_ids(spec_path: Path, *, spec_source: str | None = None) -> set[str]:
+    """Parse an OpenAPI spec file into the set of served ``METHOD:/path`` op_ids.
+
+    Runs the file through the real ingest parser
+    (:func:`~meho_backplane.operations.ingest.parse_openapi`), so the
+    returned strings are byte-for-byte the ``endpoint_descriptor.op_id``
+    values an operator's ingest of the same spec would produce — the
+    exact strings dispatch pre-flight resolves against. ``content=``
+    feeds the bytes verbatim (the https-only SSRF guard applies only to
+    the URL-fetch path); the ``file://`` URI is just the audit label.
+    """
+    spec_text = spec_path.read_text(encoding="utf-8")
+    rows = parse_openapi(
+        f"file://{spec_path}",
+        spec_source=spec_source or f"spec:{spec_path.name}",
+        content=spec_text,
+    )
+    return {row.op_id for row in rows}
+
+
+def _near_misses(op_id: str, served: Collection[str]) -> list[str]:
+    """Served op_ids sharing the missing path's last literal segment.
+
+    The #2970 repoint pattern: a hand-coded path is usually wrong in its
+    prefix (renamed API family, wrong version root), while its resource
+    name survives somewhere in the real spec — e.g. the unserved
+    ``GET:/vcenter/network/distributed-switches`` had its resource under
+    ``/vcenter/namespace-management/…/distributed-switches``. Surfacing
+    those candidates turns a bare failure into a grounded repoint lead.
+    """
+    _, _, path = op_id.partition(":")
+    literal_segments = [
+        segment
+        for segment in path.partition("?")[0].split("/")
+        if segment and not (segment.startswith("{") and segment.endswith("}"))
+    ]
+    if not literal_segments:
+        return []
+    needle = literal_segments[-1]
+    return sorted(candidate for candidate in served if needle in candidate)[:_NEAR_MISS_LIMIT]
+
+
+def assert_op_ids_served(
+    declared: Iterable[str],
+    served: Collection[str],
+    *,
+    spec_label: str,
+) -> None:
+    """Fail with an actionable diff unless every declared op_id is served.
+
+    *declared* is the lane's enumeration of hand-coded ``METHOD:/path``
+    strings (introspected from the connector's live constants, never a
+    hardcoded mirror); *served* is the pinned spec's op_id set, usually
+    from :func:`openapi_served_op_ids` (union the sets when a product
+    pins multiple specs). An empty *declared* fails rather than passing
+    vacuously — a lane that enumerates nothing guards nothing.
+
+    The failure message lists each unserved entry with up to
+    3 near-miss candidates from the served set, plus the
+    red-lane-is-a-finding protocol pointer. Never invent a repoint: fix
+    paths only to targets the pinned spec actually serves.
+    """
+    declared_set = set(declared)
+    if not declared_set:
+        pytest.fail(
+            f"reconcile lane for {spec_label} declared no op_ids — the enumeration "
+            "wiring broke (a vacuously green lane guards nothing)."
+        )
+    missing = declared_set - set(served)
+    if not missing:
+        return
+
+    lines = [
+        f"{len(missing)} hand-coded op_id(s) are not served by the pinned {spec_label}:",
+        "",
+    ]
+    for op_id in sorted(missing):
+        lines.append(f"  {op_id}")
+        candidates = _near_misses(op_id, served)
+        if candidates:
+            lines.extend(f"    near miss: {candidate}" for candidate in candidates)
+        else:
+            lines.append("    near miss: none — no served path shares its resource name")
+    lines += [
+        "",
+        "A red lane here is the guard surfacing a real finding (a typo, a renamed "
+        "endpoint, or a wrong-API-version path — the defect class fixture-based "
+        "tests cannot catch). Repoint each op_id at a path the pinned spec "
+        "actually serves; never invent one. Protocol: "
+        "docs/decisions/spec-reconcile-guards-standard.md.",
+    ]
+    pytest.fail("\n".join(lines), pytrace=False)
--- a/backend/tests/acceptance/_vcenter_spec.py
+++ b/backend/tests/acceptance/_vcenter_spec.py
@@ -20,6 +20,10 @@ local path to each spec by checking, in priority order:
    is expected to contain ``vcenter-9.0/vcenter.yaml`` and
    ``vcenter-9.0/vi-json.yaml``. Local-dev convenience for the
    maintainer with the consumer repo cloned at a known sibling path.
+   This branch delegates to the product-agnostic shelf resolver
+   (:mod:`tests._spec_shelf`, #2980) that every per-vendor reconcile
+   lane shares; only the vcenter-specific explicit/legacy env vars
+   above stay here.
 
 When no source resolves, the test skips with a pointer to this
 docstring rather than failing — the canary verifies a substrate that
@@ -32,6 +36,8 @@ from __future__ import annotations
 
 import os
 from pathlib import Path
+
+from tests._spec_shelf import resolve_shelf_file
 
 __all__ = [
     "VCENTER_SPEC_REASON",
@@ -75,12 +81,7 @@ def resolve_vcenter_yaml() -> Path | None:
             return legacy_path
         if legacy_path.is_dir() and (legacy_path / "vcenter.yaml").is_file():
             return legacy_path / "vcenter.yaml"
-    consumer_root = os.getenv("MEHO_CONSUMER_DOCS_ROOT")
-    if consumer_root:
-        candidate = Path(consumer_root).expanduser() / "vcenter-9.0" / "vcenter.yaml"
-        if candidate.is_file():
-            return candidate
-    return None
+    return resolve_shelf_file("vcenter-9.0", "vcenter.yaml")
 
 
 def resolve_vi_json_yaml() -> Path | None:
@@ -88,9 +89,4 @@ def resolve_vi_json_yaml() -> Path | None:
     explicit = _expand_optional_path(os.getenv("MEHO_VCENTER_OPENAPI_VI_JSON"))
     if explicit is not None:
         return explicit
-    consumer_root = os.getenv("MEHO_CONSUMER_DOCS_ROOT")
-    if consumer_root:
-        candidate = Path(consumer_root).expanduser() / "vcenter-9.0" / "vi-json.yaml"
-        if candidate.is_file():
-            return candidate
-    return None
+    return resolve_shelf_file("vcenter-9.0", "vi-json.yaml")
--- a/backend/tests/acceptance/test_g07_vsphere_canary.py
+++ b/backend/tests/acceptance/test_g07_vsphere_canary.py
@@ -30,19 +30,20 @@ both under one connector triple, and asserts:
 * :meth:`ReviewService.enable_connector` (T4) cascades every staged
   group to ``review_status='enabled'``, every staged op to
   ``is_enabled=True``, and writes one connector-level audit row.
-* The govc-parity benchmark: 10 of 13 representative vSphere
-  operator queries (10 ``vcenter.yaml`` + 3 ``vi-json.yaml``) return
-  the canonical operation in the top-3 hits via
-  :func:`search_operations` (T8, #438) over the PG hybrid
-  BM25+cosine RRF ranking. The three failing queries (all vcenter
-  cardinal ops) are marked ``xfail`` (non-strict, because pgvector's
-  IVFFlat approximation makes the failure non-deterministic between
-  runs) — they target cardinal operations whose spec descriptions
-  are vendor-schema-heavy and lose to short sub-path descriptions in
-  BM25 ranking. The three vi-json queries skip in single-spec mode
-  and are NOT marked xfail in two-spec mode (their target ops carry
-  descriptive method names like ``RevertToSnapshot_Task`` that BM25
-  picks up cleanly). See *Known gaps* in
+* The govc-parity benchmark: 13 representative vSphere operator
+  queries (10 ``vcenter.yaml`` + 3 ``vi-json.yaml``) return the
+  canonical operation in the top-3 hits via the sanctioned
+  group-scoped agent flow — :func:`search_operations` (T8, #438)
+  with the ``group`` filter, over the PG hybrid BM25+cosine RRF
+  ranking. The benchmark deliberately does NOT assert raw
+  corpus-wide relevance: architecture postulate 4 says raw search
+  ranks poorly at thousands-of-ops scale, the first armed two-spec
+  run proved it (#3006), and the agent flow the meta-tools instruct
+  is pick connector → list groups → search within a group. Cases
+  that miss top-3 even group-scoped are marked ``xfail``
+  (non-strict — pgvector IVFFlat variance) with the measured
+  in-group rank in the reason. The three vi-json queries skip in
+  single-spec mode. See *Known gaps* in
   ``docs/cross-repo/g07-vsphere-canary.md``.
 * A vi-json ``{moId}`` path substitutes cleanly through the
   production dispatcher helper
@@ -115,8 +116,8 @@ A real grouping pass on vcenter.yaml issues
 API. That's a non-trivial budget the canary should not consume on
 every CI run. The default stub returns deterministic group proposals
 + classifies each batch's ops by path prefix; the assertion is
-"the pipeline wires together correctly + the 10-query benchmark
-ranks every canonical op in the top-3", not "the LLM produces good
+"the pipeline wires together correctly + the 13-query benchmark
+ranks every canonical op in its group's top-3", not "the LLM produces good
 groups" (the latter is a manual operator-review step). An opt-in
 ``ANTHROPIC_API_KEY``-gated live-LLM run via
 ``MEHO_G07_CANARY_LIVE_LLM=1`` exercises the real Anthropic adapter.
@@ -276,12 +277,20 @@ _MIN_OPERATION_COUNT: int = _MIN_VCENTER_OPERATION_COUNT
 #: to re-call after partial completion.
 _MAX_STUB_LLM_CALLS: int = 128
 
-#: Govc-parity benchmark — the (query, expected_op_id) pairs the
-#: canary asserts. Each query is a natural-language phrase an
-#: experienced vSphere operator might type; the expected ``op_id`` is
-#: the canonical match for that workflow in the parsed corpus. Top-3
-#: ranking is asserted via :func:`search_operations` over the PG
-#: hybrid BM25 + cosine RRF index.
+#: Govc-parity benchmark — the (query, group_key, expected_op_id)
+#: triples the canary asserts. Each query is a natural-language phrase
+#: an experienced vSphere operator might type; ``group_key`` is the
+#: operation group an agent would pick from ``list_operation_groups``
+#: for that workflow (the stub taxonomy's ``when_to_use`` hints make
+#: the pick unambiguous); the expected ``op_id`` is the canonical
+#: match for the workflow in the parsed corpus. Top-3 ranking is
+#: asserted via :func:`search_operations` with the ``group`` filter —
+#: the sanctioned agent flow (architecture postulate 4), not raw
+#: corpus-wide search. Re-scoped from raw search when the first armed
+#: two-spec CI run (PR #2995) measured exactly the failure postulate 4
+#: predicts: cross-spec near-duplicates (vi-json ClusterComputeResource
+#: ops vs ``GET:/vcenter/cluster``) crowding the top-3 at 3,470-op
+#: scale. Raw-corpus relevance is tracked in #3006, not asserted here.
 #:
 #: The first 10 entries target ``vcenter.yaml``; the last 3 target
 #: ``vi-json.yaml`` Managed-Object operations (snapshot revert, event
@@ -289,32 +298,42 @@ _MAX_STUB_LLM_CALLS: int = 128
 #: parametrised test cases when only the vcenter env var resolves —
 #: see :data:`_VI_JSON_BENCHMARK_QUERIES`.
 #:
-#: Three queries are marked ``xfail`` (non-strict) — see
-#: :data:`_XFAIL_BENCHMARK_QUERIES`.
-GOVC_PARITY_BENCHMARK: tuple[tuple[str, str], ...] = (
-    ("list virtual machines", "GET:/vcenter/vm"),
-    ("list clusters", "GET:/vcenter/cluster"),
-    ("list datacenters", "GET:/vcenter/datacenter"),
-    ("list datastores", "GET:/vcenter/datastore"),
-    ("list networks", "GET:/vcenter/network"),
-    ("list hosts", "GET:/vcenter/host"),
-    ("power on virtual machine", "POST:/vcenter/vm/{vm}/power?action=start"),
-    ("power off virtual machine", "POST:/vcenter/vm/{vm}/power?action=stop"),
-    ("create login session", "POST:/session"),
-    ("get virtual machine info", "GET:/vcenter/vm/{vm}"),
+#: Queries measured to miss top-3 even group-scoped are marked
+#: ``xfail`` (non-strict) — see :data:`_XFAIL_BENCHMARK_QUERIES`.
+GOVC_PARITY_BENCHMARK: tuple[tuple[str, str, str], ...] = (
+    ("list virtual machines", "vm", "GET:/vcenter/vm"),
+    ("list clusters", "cluster", "GET:/vcenter/cluster"),
+    ("list datacenters", "datacenter", "GET:/vcenter/datacenter"),
+    ("list datastores", "datastore", "GET:/vcenter/datastore"),
+    ("list networks", "network", "GET:/vcenter/network"),
+    ("list hosts", "host", "GET:/vcenter/host"),
+    ("power on virtual machine", "vm", "POST:/vcenter/vm/{vm}/power?action=start"),
+    ("power off virtual machine", "vm", "POST:/vcenter/vm/{vm}/power?action=stop"),
+    ("create login session", "session", "POST:/session"),
+    ("get virtual machine info", "vm", "GET:/vcenter/vm/{vm}"),
     # vi-json Managed-Object operations. The path shape comes from the
     # parsed vi-json.yaml corpus (``/<ManagedObjectType>/{moId}/<Method>``
-    # with no server-prefix). These queries target ops with descriptive
-    # method names — ``RevertToSnapshot_Task`` literally contains
-    # "revert" and "snapshot"; ``QueryEvents`` contains "events";
-    # ``QueryPerf`` contains "perf"/"performance". They should rank
-    # top-3 cleanly without xfail discipline. If the first run finds
-    # any of them under-ranking, the canary surfaces the failure
-    # rather than silently absorbing it via xfail (see the
-    # *Acceptance criteria* in #503).
-    ("revert vsphere snapshot", "POST:/VirtualMachine/{moId}/RevertToSnapshot_Task"),
-    ("tail vsphere events", "POST:/EventManager/{moId}/QueryEvents"),
-    ("get vm performance metrics", "POST:/PerformanceManager/{moId}/QueryPerf"),
+    # with no server-prefix); group_keys come from the stub taxonomy's
+    # MO families (`/VirtualMachineSnapshot` falls under the
+    # ``/VirtualMachine`` path rule → ``vm_managed_objects``).
+    #
+    # The snapshot-revert op_id is the spec truth, verified against
+    # vi-json.yaml: ``RevertToSnapshot_Task`` is a method of the
+    # *VirtualMachineSnapshot* managed object (as in the VIM API —
+    # ``vim.vm.Snapshot.RevertToSnapshot_Task``); VirtualMachine only
+    # carries ``RevertToCurrentSnapshot_Task``. The previous expected
+    # op ``POST:/VirtualMachine/{moId}/RevertToSnapshot_Task`` does
+    # not exist in the corpus and could never rank — latent because
+    # the armed integration lane runs ``pytest -x`` and died on
+    # ``list-clusters`` before ever executing this case (first armed
+    # two-spec run, 2026-08-17).
+    (
+        "revert vsphere snapshot",
+        "vm_managed_objects",
+        "POST:/VirtualMachineSnapshot/{moId}/RevertToSnapshot_Task",
+    ),
+    ("tail vsphere events", "events", "POST:/EventManager/{moId}/QueryEvents"),
+    ("get vm performance metrics", "performance", "POST:/PerformanceManager/{moId}/QueryPerf"),
 )
 
 #: Subset of :data:`GOVC_PARITY_BENCHMARK` queries that target the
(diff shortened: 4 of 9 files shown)


damir-topic:
> satisfied by this comment.

**Anchor task:** #2980 — generalize the spec-lane harness + record the spec-reconcile-guard standard
**Surface area:** backend (tests + docs)
**Verdict:** APPROVE
**Head SHA reviewed:** `cee8a9f0245da1619e822d81d9c6279156dd35e8`

### Acceptance criteria
- [x] Five vcenter lanes green against the shelf and skip-clean without env — **independently re-proven, byte-identical to clean main** (`26b0bd9e`): with `MEHO_CONSUMER_DOCS_ROOT` pointed at the real shelf, 35 passed / 28 skipped on both branch and main; without env, 22 passed / 41 skipped on both; the full `-rs` skip-summary sections (reason strings included) diff empty in both configurations. The `_vcenter_spec.py` delegation is also a line-for-line semantic match of the removed shelf-root branch (same falsy-env check, same `expanduser()`, same `is_file()` gate).
- [x] Example lane demonstrates the harness in <30 lines — 29 lines; **runs for real** (passes, not skips) against the real `vcenter-9.0` shelf and skips with the uniform reason without env. In armed CI it executes in the unit job: both Python jobs carry the secret-gated shelf checkout + unconditional `MEHO_CONSUMER_DOCS_ROOT`, and the file is not in either ignore list.
- [x] Decision doc filed; docs/codebase index — no testing-surface doc exists in `docs/codebase/` (verified; `spec-ingestion.md` is the ingest-pipeline doc), `devops.md` spec-shelf section updated instead and stays accurate.

### Decision-doc mechanics vs. what #2966 actually shipped
Verified against `ci.yml` on this head: `sparse-checkout: docs/vcenter-9.0` + fail-loud verify step present on **both** Python jobs (unit ~L300/L318, integration ~L913/L924), `SPEC_SHELF_TOKEN_PRESENT` presence-boolean gating (secrets context unavailable in step-level `if:`), `MEHO_CONSUMER_DOCS_ROOT` set unconditionally on both jobs. The doc's "one sparse-checkout line + one verify line per product dir, same `SPEC_SHELF_TOKEN`, no new secret" extension mechanics match that shape exactly. The `nsx-9.0/MANIFEST.md` no-OpenAPI-spec claim (2026-04-29) verified against the real shelf.

### Project-board alignment
|---|---|---|---|---|---|
| evoila/meho#2980 | Open (task, priority:high) | Closes | ✅ | ✅ full | close on merge |

Hierarchy coherent: #2980 → Initiative #2979 → Goal #221; sibling lane tasks #2981–#2993 depend on this harness and are correctly out of scope here. No other open PR touches these files.

### Predicted vs. actual diff
- ✅ `backend/tests/_spec_shelf.py` — new harness (resolver + parser extraction + assert), expected and correct
- ✅ `backend/tests/acceptance/_vcenter_spec.py` — shelf-root branch delegation only; public API, reason string, and env-var priority untouched
- ✅ `backend/tests/test_spec_shelf_example_lane.py` — the AC-2 template lane
- ❓ `backend/tests/test_spec_shelf_harness.py` — unexpected but justified: 13 always-on unit tests incl. the vcenter delegation pins
- ✅ `docs/decisions/spec-reconcile-guards-standard.md` — the AC-3 standard
- ✅ `docs/codebase/devops.md` — spec-shelf section updated to match
- 🕳️ none missing; 🚫 none undisclosed. `ci.yml` untouched by design (vcenter-9.0 already armed; widening belongs to the sibling lane tasks).

### Integration verification
- [x] `ruff check src/ tests/` — clean; `ruff format --check` — clean (1422 files)
- [x] `mypy src/` — sole error is the pre-existing Darwin-only `net/icmp.py` `MSG_ERRQUEUE` typeshed artifact (present on clean main, absent on Linux CI)
- [x] Five-lane matrix (branch × main × with/without shelf) — counts and skip-reason strings byte-identical, see AC-1
- [x] Harness + example lane: 14 passed with shelf; example lane skips clean without env, reason names `vcenter-9.0/vcenter.yaml` + `MEHO_CONSUMER_DOCS_ROOT`
- [x] Full CI-equivalent unit sweep (`-n 4 --dist loadscope`, ignoring integration/migrations): **11578 passed**, 196 skipped; the only 3 failures (`net_icmp` loopback trace, `net_dns_lookup` resolver, chart mcp env) were re-run on clean main and fail identically there — pre-existing local-env flakes, not this diff
- [x] `parse_openapi(..., content=)` contract re-read at source: inline content bypasses the https-only SSRF fetch path exactly as the harness docstring claims; returned op_ids are the `EndpointDescriptorProto.op_id` strings
- [x] pytest 9.1.1 idioms (`pytest.fail(pytrace=False)`, `pytest.skip` in call phase, `pytest.fail.Exception` / `pytest.skip.Exception`) confirmed against current pytest reference docs; both exception-attribute patterns have repo precedent
- ⚠️ not run: migrations replay / docker compose / frontend — n/a, no src, migration, or frontend changes
- ⚠️ PR CI: all jobs on this head failed at **"Set up job"** (runner provisioning — the known ARC wedge shape), zero code signal; needs a re-run. Local verification above stands in; merge remains gated on green CI.

None — no inline comments posted.

- `openapi_served_op_ids`'s `spec_source=` kwarg is inert for this function's observable output (it only tags rows, and only `op_id`s are returned). Dropping it in a follow-up would keep the 12-lane template surface minimal; a lane needing provenance calls `parse_openapi` directly. The copy-paste template (example lane) doesn't use it, so propagation risk is low.
- The #2944 lane (`test_connectors_vmware_rest_composites_l2_ingest_reconcile.py`) still carries its own bespoke reconcile core and `_vi_json_path_item_has_post` text checks. Leaving it untouched was the right call under the byte-identical constraint; a post-wave-3 consolidation task could migrate it onto `assert_op_ids_served`.
- Skips raised inside `require_shelf_spec` report `tests/_spec_shelf.py:93` as their location (the raise site) instead of the lane's line, unlike the legacy lanes which skip at the call site. The reason string names the exact product/file, so it stays actionable either way — recording for awareness only.

### Docs
- [x] `docs/codebase/devops.md` — updated, accurate against `ci.yml` on this head
- [x] `docs/decisions/spec-reconcile-guards-standard.md` — filed; cross-link to `vendor-spec-ci-provisioning.md` resolves; every checkable claim in it verified (see mechanics section)

- The vcenter delegation pins in `test_spec_shelf_harness.py` turn the byte-identical constraint into permanent regression coverage rather than a one-time manual proof.
- `assert_op_ids_served`'s vacuous-empty guard and near-miss leads reproduce the exact #2970 failure shape in a unit test — the failure UX the 12 sibling lanes inherit is genuinely actionable.

damir-topic:
> satisfied by this comment.

**Anchor task:** #2980 — generalize the spec-lane harness + record the spec-reconcile-guard standard
**Surface area:** backend (tests-only delta)
**Verdict:** APPROVE
**Head SHA reviewed:** `0e08e3e5b1fccec4aaac9f5fd4260223ea78d25e`

Iteration 1 (head `cee8a9f0`) approved with zero findings; the sole delta since is `0e08e3e5` — a test-only fix (+28/−4 in `backend/tests/acceptance/test_g07_vsphere_canary.py`) for the latent canary bug armed CI exposed. This iteration verified the fix's claims at source and re-proved the byte-identical constraint in the review worktree.

### Delta verification (the fix's claims, checked at source)

- **Root-cause claim holds.** `SpecSource` (`backend/src/meho_backplane/operations/ingest/api_schemas.py`) documents the contract exactly as the fix states: content-less sources are *fetched* under the G0.16-T8 (#95) https-only SSRF guard; `_load_spec_bytes` (`openapi.py`) rejects bare paths at the scheme check before any read. A content-less `SpecSource(uri=str(<shelf path>))` could never have worked; it was latent only because the spec fixtures always skipped pre-shelf-arming.
- **Same on-ramp claim holds.** The CLI reads `file://`/`docs:` sources client-side and uploads bytes via `content` (`cli/internal/cmd/connector/ingest.go:291`, `src.Content = &content`); `tests/integration/test_operations_ingest_vcenter.py:88` feeds the same inline-content channel (`parse_openapi(..., content=spec_text)`). `_inline_spec_source` is the same shape.
- **Audit-label contract holds.** Pipeline calls `parse_openapi_with_provenance(spec.uri, spec_source=spec.uri, content=spec.content)` and tags rows `spec:<uri>`; with `uri` kept as the bare absolute path, `test_canary_every_row_tagged_with_spec_source`'s `endswith("vcenter.yaml")` / `endswith("vi-json.yaml")` keys are unchanged. `classify_spec_origin` → `inline` (content set, non-`spec:` label), exactly as the PR body states; no canary test asserts a `fetched` origin, so nothing else moves.
- **Call-site completeness.** All four `SpecSource` sites in the canary switched (`ingested_canary` ×2, `real_llm_ingested_canary`, `vcsim_ingested_canary`); the helper at line 737 is now the file's only constructor. Repo-wide sweep for other content-less local-path `SpecSource`s: `tests/test_operations_ingest_pipeline_version_check.py:125` is the only other local-path construction and it is **not** the same latent bug — an autouse fixture deliberately monkeypatches `read_spec_info_version` citing the #95 guard, and only `_validate_spec_versions` is exercised. `tests/integration/test_operations_ingest_github.py` uses a live public https URL behind an explicit `MEHO_GH_INGEST_LIVE=1` opt-in. UI modal / catalog constructions are the by-design fetch path.
- **Safety of the client-side read.** Both resolvers return a path only when `is_file()` (`tests/_spec_shelf.py:86`, `tests/acceptance/_vcenter_spec.py:64`), so `read_text` cannot hit a missing file; both shelf specs (7.5 MB / 10.5 MB) sit well under the 20 MiB `content` char cap and `_MAX_SPEC_BYTES`.

### Acceptance criteria (re-verified on the new head)

- [x] Five vcenter lanes green against the shelf and skip-clean without env — re-run in the review worktree: **35 passed / 28 skipped** with `MEHO_CONSUMER_DOCS_ROOT` (all skips Docker-socket/opt-in gated), **22 passed / 41 skipped** without env. Shelf clone verified clean against origin.
- [x] Example lane <30 lines demonstrates the harness — harness + example lane: **14 passed** with the real shelf.
- [x] Decision doc filed + `devops.md` updated — unchanged since iteration 1 (verified there); delta invalidates neither.

### Project-board alignment

|---|---|---|---|---|---|
| #2980 | Open (task, priority:high, effort:medium, connector) | Closes | ✅ | ✅ | close on merge |

Unchanged from iteration 1; #2995 is the only open PR referencing #2980.

### Predicted vs. actual diff (delta)

- ✅ `backend/tests/acceptance/test_g07_vsphere_canary.py` — exactly the predicted fix shape: one helper + four call-site switches, nothing else touched.

### Integration verification (this iteration, in the review worktree)

- [x] `ruff check` + `ruff format --check` on the changed file — clean
- [x] Five vcenter lanes with shelf env — 35 passed / 28 skipped (matches main-equivalent counts from iteration 1)
- [x] Five vcenter lanes without env — 22 passed / 41 skipped
- [x] Harness + example lane with shelf — 14 passed
- [x] Direct proof of the fixed on-ramp: both real shelf specs parsed through the inline-content channel with the bare-path label — `vcenter.yaml` 1275 ops (floor 1200), `vi-json.yaml` 2195 ops (floor 2000), `info.version=9.0.0.0`, `origin=inline`, uri label preserved
- ⚠️ mypy — n/a for the delta (CI runs `mypy src/`; no src changes)
- ⚠️ PR CI on `0e08e3e5` — **pending at post time**: Semgrep, CodeQL, TruffleHog, DCO, Go, Helm, chart, license checks all green; `Python (ruff + mypy + pytest)` and `Python (integration testcontainers)` in progress (runners healthy — no Set-up-job wedge this time). The armed-CI canary path (Docker-gated subtests) is exercised only there; merge should wait for these two jobs to complete green.

### Findings

None. `counts: {blocker: 0, major: 0, minor: 0, question: 0}`.

### Adjacent improvements (non-blocking, carried from iteration 1 — all still valid)

- `openapi_served_op_ids`'s `spec_source=` kwarg is inert for its observable output; drop in a follow-up.
- Post-wave-3 consolidation candidate: migrate the #2944 lane's bespoke reconcile core onto `assert_op_ids_served`.
- Skips raised inside `require_shelf_spec` report the raise site as location; reason string keeps it actionable — awareness only.

### Docs

- [x] `docs/decisions/spec-reconcile-guards-standard.md` + `docs/codebase/devops.md` — verified in iteration 1; the tests-only delta invalidates neither.

### What worked well

- The fix's docstring is a model incident record: names the guard, the CLI on-ramp precedent, the integration-lane precedent, and why the bug was latent — every claim in it checked out at source.
- Minimal blast radius: four call sites through one helper, audit-label semantics deliberately preserved so the tag-reconcile assertions needed no touch.

damir-topic:
> satisfied by this comment.

**Anchor task:** #2980 — generalize the spec-lane harness + record the spec-reconcile-guard standard
**Surface area:** backend (tests-only delta)
**Verdict:** APPROVE
**Head SHA reviewed:** `eaccb8953bbbd400d8b8e31fa52da7b9230a370e`

Iterations 1 (`cee8a9f0`) and 2 (`0e08e3e5`) both approved the full diff; this iteration reviews the delta — commit `eaccb895`, the stub-taxonomy fix for the `LlmOutputInvalid` failure armed CI run 32048056557 exposed.

### Delta verification (commit `eaccb895`)
- [x] **Key renames consistent** — the four kebab-case `group_key`s (`vm-managed-objects`, `host-managed-objects`, `cluster-managed-objects`, `datastore-managed-objects`) are renamed to snake_case in **both** `_PROPOSE_RESPONSE` and `_PATH_RULES`. Repo-wide grep: zero occurrences of the old names remain anywhere in `backend/`. (Kebab-case `group_key`s elsewhere in the test tree — `vm-lifecycle`, `rke2-node-write`, etc. — are typed-connector/DB-row keys outside the LLM-grouping Pass-1 contract, untouched on main; the `^[a-z][a-z0-9_]*$` pattern is enforced only in `parse_proposal_response`.)
- [x] **New test round-trips the production validator, not a copy** — `test_stub_taxonomy_passes_production_proposal_validation` imports `parse_proposal_response` + `NONE_GROUP_KEY` directly from `meho_backplane.operations.ingest._llm_grouping_internals` (the module whose `GroupProposal` field validator enforces `_GROUP_KEY_PATTERN` at src line 248). `_PROPOSE_RESPONSE` is a `json.dumps(...)` string — exactly the `raw_output: str` the parser takes, so the round-trip exercises fence-stripping, JSON decode, pydantic validation, and the duplicate-key check.
- [x] **Path-rules-subset assertion correct** — `rule_keys - proposed_keys - {NONE_GROUP_KEY}` must be empty. All 14 `_PATH_RULES` targets are proposed keys today; the `NONE_GROUP_KEY` allowance correctly tolerates a future explicit `none` rule (the stub's `_group_for_path` fallback).
- [x] **Docstring 8→14** matches reality: 14 `"group_key"` entries in the static taxonomy (8 vcenter.yaml families + 6 vi-json groups).
- [x] **Chokepoint proof re-run both ways** — the fixed taxonomy parses to exactly 14 proposals; re-injecting one kebab key raises `LlmOutputInvalid` with the byte-matching `group_key 'vm-managed-objects' must match '^[a-z][a-z0-9_]*$'` frame.

### Diagnosis spot-check (PR appendix, fix iteration 2)
- [x] **Deterministic stub, no live LLM in the unit job** — the canary fixture chain builds `IngestionPipelineService(..., llm_client_factory=lambda: stub_client)` with `stub_client = _PathPrefixStubLlmClient()`; the live variant asserts `ANTHROPIC_API_KEY` and is gated behind `MEHO_G07_CANARY_LIVE_LLM`; `.github/workflows/ci.yml` sets neither (grep: no match).
- [x] **Introduced by #520** — `git log -S "vm-managed-objects"` on the merge-base pins the kebab keys to `476ef1c7` (#520, G3.1-T3 two-spec canary). Not introduced by this PR.

### Integration verification (all re-run in the review worktree at `eaccb895`)
- [x] New test fixture-less: passes with **no** `MEHO_CONSUMER_DOCS_ROOT`, no DB, no Docker (1 passed, 1.6 s)
- [x] Five vcenter lanes **with** shelf env: **36 passed / 28 skipped** — exactly iteration-2's 35/28 plus the one new test
- [x] Five vcenter lanes **without** env: **23 passed / 41 skipped** — exactly 22/41 plus the new test (it runs in both matrices)
- [x] Harness + example lane with shelf: 14 passed (unchanged)
- [x] `ruff check` + `ruff format --check` on the changed file — clean
- [x] Shelf clone freshness: `docs/vcenter-9.0` identical to origin (last shelf change 2026-04-29)
- ⚠️ not run locally: the DB-backed two-spec ingest (Postgres testcontainers + shelf) — armed CI only; that is precisely the path the renamed taxonomy now feeds, and run 32049795686's Python jobs are the proof

Environment note (my harness, not the PR): the first five-lane run picked up a *newer* `meho_backplane` src via the shared editable install (origin/main advanced with #2994's `_SUB_OPS_VM_DEPLOY_FROM_LIBRARY` mid-review), which failed the merge-base-pinned reconcile lane. Re-run with the worktree's own src pinned — version-coherent counts above. Post-merge coherence holds: #2994 updated that reconcile test and its src in lockstep on main, and this PR touches neither.

### Project-board alignment
|---|---|---|---|---|---|
| #2980 | Open (task, priority:high, effort:medium) | Closes | ✅ | ✅ | close on merge |

Ordering note: #3000 / PR #3001 extract the canary inline-spec fix (this PR's `0e08e3e5` content) to main independently so other branches' armed CI stops failing. Whichever lands second touches the same `test_g07_vsphere_canary.py` hunks — expect a rebase or redundancy resolution; content of both is compatible.

### Findings
None. Zero blockers, zero majors, zero minors, zero questions.

- Carried from iterations 1–2: inert `spec_source=` kwarg on `openapi_served_op_ids`; #2944-lane consolidation onto `assert_op_ids_served`; `require_shelf_spec` skip-location awareness.
- The PR appendix's adjacent finding stands on its own merits: the production grouping pass has no retry/normalization on `LlmOutputInvalid`, so a live model emitting one kebab key fails an entire ingest — worth its own ticket (product scope, out of scope here).

### Docs
- [x] `docs_status: verified_unchanged` for the delta — the stub taxonomy is test-internal; the decision doc + `devops.md` updates from iteration 1 remain accurate.

### What worked well
- The fixture-less self-check is the right durable fix: it converts a shelf-armed, Docker-gated CI-only failure class into a unit-speed failure in every environment, and the Pass-2 consistency assertion closes the sibling gap (a rule targeting a non-proposed key would silently fall to `none`).

### Merge gate
CI run 32049795686 on `eaccb895` was **in progress at post time**: Go, CLI snapshot, Helm, migration-detect all green; coverage + migrations skipped by design; **Python (ruff + mypy + pytest) and Python (integration testcontainers) still running** — these two are the merge gate, and the armed-CI canary ingest that motivated this fix runs only there. Do not merge before both complete green. `reviewDecision` cannot flip from this comment (#2733).

damir-topic:
**auto-parked: fix-iteration budget exhausted (3/3)** — /auto-implement-initiative #2979 run.

Cycle history (every fix was a REAL latent bug the newly-armed spec-shelf CI exposed — none introduced by this PR's own logic):
1. `0e08e3e5` — canary fixtures built content-less bare-path `SpecSource`s → rejected by the https-only SSRF guard. Latent since the canary was written; never ran armed before today.
2. `eaccb895` — canary stub taxonomy carried kebab-case `group_key`s (introduced by #520) → production snake_case validator. Latent for the same reason.
3. `08d3e1af` — the armed canary's true cost (~15 min two-spec ingest) blew the unit job's 25-min timeout (4 timeout-kills measured, incl. on a quiet cluster) → relocated to the integration lane per the workflow's own perf-budget discipline; unit sweep measured back to 9:41.

Remaining failure (1 of 303 in the integration lane, first-ever armed execution of this test):
`test_canary_govc_parity_benchmark[list-clusters]` — asserts `GET:/vcenter/cluster` in raw-search top-3 for 'list clusters'; with the full two-spec corpus (1275 + 2195 ops) vi-json ClusterComputeResource ops outrank it. Note: this benchmark asserts raw-corpus-wide search relevance, which the architecture (CLAUDE.md postulate 4) explicitly says ranks poorly at this scale — the sanctioned agent flow is group-scoped search. The real corpus has effectively proven the postulate.

Operator decision needed — options:
a) Re-scope the benchmark to the group-scoped flow (align the test with postulate 4), one more fix iteration.
b) Keep the raw-search assertion as a genuine ranking-quality bar → file a search-relevance ticket, mark the case xfail-with-ticket, land the PR.
c) Take over manually.

Everything else is green: 301/303 integration (36:41, under the 60-min cap), unit lane 9:41, three APPROVE verdicts of record (iters 1–3, iter-4 review in flight on this head).

damir-topic:
> satisfied by this comment.

## /auto-review-pr summary (iteration 4)




**Anchor task:** #2980 — generalize the spec-lane harness + record the spec-reconcile-guard standard
**Surface area:** backend + devops
**Verdict:** REQUEST_CHANGES
**Head SHA reviewed:** `d150d28d8fa6db4390eb1f06f1fc2a10d947a936`
**Delta reviewed:** commits `08d3e1af` (lane placement / MEHO_SKIP_SPEC_INGEST_TESTS gating) and `d150d28d` (benchmark re-scope to group-scoped triples) — neither previously reviewed (iters 1–3 stopped at `eaccb895`).

### Verdict in one paragraph

Both commits are individually well-built and were verified in depth (below) — the truthy parsing is byte-faithful to the `MEHO_RUN_SLOW_TESTS` precedent, the re-scope drives the real production group path, the corrected op is spec-verified, every xfail reason matches independently reproduced measurements, and the unit merge-gate job is now **proven green on this head** (run 32117732263, success at 08:59Z, pytest 916 s vs 1320 s budget). The blocker is arithmetic, not craftsmanship: the same run's **integration** merge-gate job cannot fit the newly-unblocked workload inside its 60-minute cap. Measured on these runners (run 32067450471): ~164 s per full ingest, and the module re-ingests per test by design — 25 armed ingest-bearing cases. `d150d28d` fixing `[list-clusters]` is exactly what removes the `-x` early-stop that had been hiding 16 of those 25: projected pytest wall ≈ 710 s (integration subtree) + 25 × 164 s ≈ **80 min against the ~59 min available**. Iteration 3's "~15–20 min added, well under cap" estimate is contradicted by its own first armed run (which added ~25 min while executing only 9 of 25 ingests). The in-flight integration job (pytest started 08:43:16Z, kill at 09:42:28Z) adjudicates this: at the measured rate it cannot finish in time. If it defies the projection and completes green with real headroom, B1 is refuted and the orchestrator can override to APPROVE — if it squeaks under at >90 % of cap, the fix is still needed per this workflow's own discipline ("timeout caps a runaway gate at **twice** its observed cost", ci.yml header; #898/#793/#827).

### Acceptance criteria (delta scope)

- [x] Unit-lane timeout fixed (08d3e1af) — proven: Python (ruff+mypy+pytest) green on this head, 17:11 job wall vs 25-min cap.
- [x] Canary full-ingest runs in exactly one armed lane — verified: opt-out env set in `python-lint-test` + `python-coverage` only; `python-integration` selects the module; env read only by the canary module (repo-wide grep); example lane / harness / reconcile lanes unaffected.
- [x] Benchmark asserts the sanctioned group-scoped flow (d150d28d) — verified at source: `search_operations` → `resolve_group_id` → `hybrid_search(group_id=…)` (meta_tools.py); unknown group short-circuits to empty hits, which the test's `assert hits` fails loudly.
- [x] Corrected op is spec truth — verified against the shelf: `/VirtualMachineSnapshot/{moId}/RevertToSnapshot_Task` exists (vi-json.yaml:67648); `/VirtualMachine/{moId}/` carries only `RevertToCurrentSnapshot_Task` (:64993). The old expected op does not exist in the corpus.
- [x] Xfails are non-strict, measured, non-convenience — all four carry the probe-measured in-group rank in the reason string, matching the fix-4 result file's probe table; independently reproduced (below).
- ❌ Integration-lane budget honest — **B1**. The lane the full ingest was relocated INTO cannot hold it.

### Integration verification (ran in the review worktree at d150d28d)

- [x] `ruff check` + `ruff format --check` — clean on the changed test file.
- [x] Five-lane matrix, shelf env + opt-out: **36 passed / 28 skipped** — matches claim.
- [x] Five-lane matrix, env-less: **23 passed / 41 skipped** — matches claim.
- [x] Canary module, shelf + `MEHO_SKIP_SPEC_INGEST_TESTS=1`: **1 passed / 28 skipped in 0.74 s**, no PG container spun; exactly 25 opt-out-reason skips + 3 opt-in-reason skips.
- [x] Collection: 29 node-IDs, benchmark case ids unchanged (byte-identical to 08d3e1af).
- [x] Integration-lane selection collect (shelf env): 333/334, 1 `load`-marker deselection intact — `pyproject.toml` `addopts -m "not load"` not clobbered.
- [x] **Real 3-case spot-check with full two-spec ingest per case** (Docker PG + pgvector + real bge-small + shelf specs): `[list-clusters]` **PASSED**, `[revert-vsphere-snapshot]` **PASSED**, `[tail-vsphere-events]` **XFAIL** with the measured reason — `2 passed, 1 xfailed in 328.44s`. Independent reproduction of the worker's probe claims.
- [x] CI on this head: unit job **success**; migrations **success**; integration **in progress at post time** (see B1).

### Findings

- 🛑 **B1** — `python-integration` (required merge gate) is projected past its 60-min `timeout-minutes` once the canary module runs to completion; required change: amortize the per-test full ingest (shared module/class-scoped ingest for the read-only cases — note `pg_engine`'s between-test TRUNCATE is the documented reason for function scope, so the shared path needs a non-truncating engine fixture — keeping function scope only for state-mutating tests), or, by explicit maintainer decision, move the full-ingest canary to a scheduled armed lane outside the required per-PR gate. Raising `timeout-minutes` is not an acceptable fix (#898/#793/#827; this PR's own decision-doc §Lane placement). See the inline comment on `ci.yml` for the full arithmetic.

### Project-board alignment

|---|---|---|---|---|---|
| #2980 | Open | Closes | ✅ | ✅ full (modulo B1) | close on merge |
| #3006 | Open | Refs | — | n/a (evidence sink) | none |

Ordering note carried from iter-3: #3001 (canary inline-spec extract) is still open; whichever of #2995/#3001 lands second touches the same hunks.

### Docs

- [x] `docs/decisions/spec-reconcile-guards-standard.md` §Lane placement — correctly instructs the 12 dependent lane tasks (#2981–#2993): reconcile lanes = ungated, seconds-cheap, unit sweep; full-ingest canaries = integration-only. Consistent with `devops.md` and ci.yml comments.
- [x] `docs/cross-repo/g07-vsphere-canary.md` — matches the new contract: triples table with groups, 9-of-13 group-scoped, four measured xfails, corrected op #11, eyeball stays raw for #3006. (Note the B1 outcome may oblige a further update if the fixture scoping changes.)

- Carried from iters 1–3: inert `spec_source=` kwarg on `openapi_served_op_ids`; #2944-lane consolidation onto `assert_op_ids_served`; `require_shelf_spec` skip-location awareness.
- The two opt-in gates use inconsistently-prefixed env vars (`MEHO_G07_CANARY_LIVE_LLM` vs `G07_CANARY_REAL_LLM` without the `MEHO_` prefix). Code and docs agree with each other today; pre-existing, unify in a follow-up.
- Carried product-scope ticket candidate: no retry budget / kebab→snake normalization on `LlmOutputInvalid` in the production grouping pass.

- The re-scope surfaced and honestly documented two real findings (mis-authored snapshot-revert op; QueryEvents in-group rank 4) instead of papering over them — and the corrected op was verified against the spec, not assumed.
- The opt-out marker's collection-time placement (so an opted-out lane never boots the module-scoped Postgres container) is exactly right, and reproduced: 0.74 s for the opted-out module.

.github/workflows/ci.yml
```diff
@@ -954,14 +986,28 @@ jobs:
...
+        # lane's 25-min cap the first day the shelf was armed, so this
+        # job (60-min cap, containers first-class) is their single
+        # armed lane; the unit sweep opts out via
+        # MEHO_SKIP_SPEC_INGEST_TESTS (see its Pytest step's env
+        # comment). Shelf absent (fork/Dependabot/unprovisioned) → the
+        # canary skips exactly as it always did; its fixture-less
+        # self-checks also run (again) here — seconds, harmless. Lane
+        # placement rule: docs/decisions/spec-reconcile-guards-
+        # standard.md.
+        run: uv run pytest -x tests/integration/ tests/acceptance/test_g07_vsphere_canary.py
```
damir-topic, line 1013: **🛑 B1** — this required merge-gate lane is projected past its 60-min cap once the canary module actually runs to completion

**What:** Relocating the full-ingest canary here was measured against a run that never finished the module. Run 32067450471 (head `08d3e1af`) spent 2201.65 s in pytest = ~710 s for `tests/integration/` (the eaccb895 baseline: 710.04 s) + ~1490 s for the **9** ingest-bearing canary cases it executed before `-x` stopped on `[list-clusters]` — i.e. **~164 s per full ingest on these runners**. The module has **25** armed ingest-bearing cases (the `ingested_canary` fixture is function-scoped by design), and `d150d28d` fixing `[list-clusters]` is precisely what removes the early-stop that hid the other 16. Projected pytest wall: 710 + 25 × 164 ≈ **4,810 s ≈ 80 min**, against ~59 min available under `timeout-minutes: 60` (setup is ~50 s). The iteration-3 estimate in the PR body ("~15–20 min serial, well under cap") is contradicted by its own first armed run, which added ~25 min while executing only 9 of 25 ingests.

**Why it matters:** This job is a required merge gate (#698). At the measured rate the in-flight run on this head (32117732263, pytest started 08:43:16Z, kill at 09:42:28Z) cannot finish in time — the PR trades the unit lane's timeout for the same timeout here, and every armed run on any branch inherits it after merge. Even in the best plausible case (my Mac measured 109 s/ingest — these runners measured 50% slower) the lane lands at ≥97% of its cap, violating this workflow's own header discipline ("`timeout-minutes` caps a runaway gate at twice its observed cost").

**Required change:** Amortize the per-test ingest before the canary rides this lane — split `ingested_canary` into a shared (module- or class-scoped) ingest consumed by the read-only cases (all 13 benchmark cases plus the read-only assertions), keeping function scope only for tests that mutate connector state. Note the documented reason for function scope is `pg_engine`'s between-test TRUNCATE (`tests/integration/conftest.py`), so the shared path needs a non-truncating engine fixture (or a truncation-exempt module) and the module docstring's scoping rationale updated in the same commit. That drops the armed cost from 25 ingests to ~3–5 (~10–15 min total, honest headroom). Acceptable alternative by explicit maintainer decision: move the full-ingest canary out of the required per-PR gate into a scheduled armed lane, keeping the seconds-cheap lanes armed per-PR. NOT acceptable: raising `timeout-minutes` (#898/#793/#827; `docs/decisions/spec-reconcile-guards-standard.md` §Lane placement — this PR's own rule).

**Source:** run 32067450471 job 95502754228 pytest summary (`2201.65s (0:36:41)`, progression `........xF`); run 32049795686 job 95486117806 (`710.04s`, integration subtree only); local reproduction at `d150d28d`: 3 benchmark cases = 328.44 s (109 s/ingest on a faster box).

**Adjudication:** if run 32117732263's integration job completes **success** with real headroom before this is acted on, B1 is refuted — treat this comment as resolved by that evidence. A timeout, or a pass at >90% of cap, confirms it.

damir-topic:
> satisfied by this comment.

## /auto-review-pr summary (iteration 5)




**Anchor task:** #2980 — generalize the spec-lane harness + record the spec-reconcile-guard standard
**Surface area:** backend (tests) + devops
**Verdict:** APPROVE
**Head SHA reviewed:** `4146b4f6843f0126c84e510bb2b6a4cd441e00ae`

### B1 closure (the sole iteration-4 blocker)

B1 (integration lane projected past its 60-min cap under 25 per-test full ingests; confirmed empirically by the 60m19s timeout kill of run 32117732263) is **closed** by `4146b4f6`, verified structurally and empirically:

- **Exactly one ingest per module run.** The module-scoped `_canary_corpus` fixture performs the two-spec ingest once via `asyncio.run` on a private event loop; the function-scoped `ingested_canary` binder only re-pins a fresh engine per test. Independently reproduced: the **full armed module ran locally in 121.65 s** (Docker PG + real bge-small + real shelf specs) — `21 passed, 3 skipped, 5 xfailed` — vs ~25 min the per-test-ingest shape would cost locally. (The author's 4-xfail + 1-xpass vs this run's 5-xfail on one benchmark case is the documented pgvector IVFFlat variance — exactly why the markers are non-strict.)
- **The binder cannot truncate the shared corpus.** Walked the fixture graph: `ingested_canary` is `pg_engine` minus the TRUNCATE + seed — it creates an engine bound to the module container and assigns `engine_module._engine` directly; no TRUNCATE statement exists in it. The root conftest's autouse `_default_database_url` only touches per-test SQLite files and the process engine cache (dispose + reset, no PG writes), which is precisely why the per-test re-pin is mandatory and sufficient.
- **Fixture ordering sound.** Module-scoped fixtures instantiate before every function-scoped autouse pin (pytest scope-ordering), which is why `_canary_corpus` pins its own `_CHASSIS_ENV` / `DATABASE_URL` / model-cache env via a module-lifetime `MonkeyPatch` and brackets its own structlog floor. Engine lifecycle is loop-affine-clean: the corpus engine is created and disposed inside the private `asyncio.run` loop; each test's engine is created and disposed on that test's own loop.
- **Read-only isolation verified test-by-test.** All 25 shared-corpus cases are SELECTs, DB-backed meta-tool reads (`search_operations` / `list_operation_groups` / `list_ingested_connectors`), result-object inspection, or pure functions (`_substitute_path`). The empty-v2-registry consequence is safe: `_require_dispatchable_connector` returns on `connector_exists` (DB) before any registry lookup (meta_tools.py:613–619). The three mutating opt-in variants (live-LLM, real-LLM eyeball, vcsim dispatch) keep their own truncating `pg_engine` + function-scoped ingest and are all defined after every shared-corpus consumer.
- **Projected CI wall honest:** ~710 s integration baseline + ~30 s container/model + ~164 s single ingest + ~100 s bodies ≈ 17 min vs ~59 min available — the local 121.65 s whole-module run is consistent with the bodies-cost term.

### Acceptance criteria (delta scope — iterations 1–4 previously verified)
- [x] B1 required_change implemented as specified (amortize, not `timeout-minutes`) — fixture graph + empirical run above
- [x] No test contract changed: collection node-IDs byte-identical to `d150d28d` (29 tests, diff empty)
- [x] Opted-out lane byte-identical: 1 passed / 28 skipped in 0.83 s, no container, 25 opt-out-reason + 3 opt-in-reason skips
- [x] Envless matrix: 1 passed / 28 skipped
- [x] `ci.yml` delta comment-only (the `run:` line unchanged; canary still in exactly one armed lane)
- [x] New xfail is honest: non-strict (no `strict=True`, no `xfail_strict` in pyproject), the `< 50%` assertion still executes every armed run, reason carries the measured numbers (2831/3470 = 81.6%; 949/1275 + 1882/2195 = 2831/3470; 639 assigned — internally consistent). Pre-existence on `d150d28d` holds by construction: the delta touches only fixture scoping — the corpus (same specs, same deterministic path-prefix stub, same pipeline) is identical, so the ratio is identical. This is a stub-taxonomy coverage finding surfaced by making the full module reachable, not a convenience xfail; documented in Known gaps §4 of `docs/cross-repo/g07-vsphere-canary.md`.

### Project-board alignment
|---|---|---|---|---|---|
| #2980 | Open (task) | Closes | yes | full | close on merge |
| #3006 | Open (task) | Refs | — | follow-up (raw-corpus relevance) | stays open |
| #3001 | Open | sequencing note in PR body | — | — | orchestrator-managed |

### Predicted vs. actual diff
- ✅ `backend/tests/acceptance/test_g07_vsphere_canary.py` — module-scoped corpus + binder + xfail
- ✅ `docs/decisions/spec-reconcile-guards-standard.md` — §Shared-corpus amortisation (the pattern of record for future heavy lanes)
- ✅ `docs/codebase/devops.md`, `docs/cross-repo/g07-vsphere-canary.md` — both updated to match
- ✅ `.github/workflows/ci.yml` — comment-only
- No unexpected files, nothing missing.

### Integration verification (ran vs reasoned)
- [x] `ruff check` + `ruff format --check` on the changed test file — clean (ran)
- [x] `code-quality.py --diff` — exit 0 (ran)
- [x] Full armed module, Docker PG + real embeddings + real shelf: 21 passed / 3 skipped / 5 xfailed in 121.65 s, single shared ingest (ran)
- [x] Opted-out lane matrix: 1 passed / 28 skipped, 0.83 s, no container (ran)
- [x] Envless matrix: 1 passed / 28 skipped (ran)
- [x] Collection node-ID diff vs `d150d28d`: empty, 29 tests (ran)
- [x] `_require_dispatchable_connector` DB-first short-circuit — verified in source (meta_tools.py:613)
- mypy: n/a (tests/docs/ci-comment delta; CI runs mypy on src/ only)
- ⚠️ CI on head `4146b4f6` (run 32125729197): "Python (database migrations)" success; "Python (ruff + mypy + pytest)" and "Python (integration testcontainers)" **in progress at post time** (started 10:14Z). The integration job's wall against the 60-min cap is the final empirical B1 headroom evidence; the orchestrator gates the merge on it separately.

### Findings
None. B1 is closed; no new blockers, majors, minors, or questions.

- Carried from iters 1–4: inert `spec_source=` kwarg on `openapi_served_op_ids`; #2944-lane consolidation onto `assert_op_ids_served`; `require_shelf_spec` skip-location awareness; opt-in env-var prefix inconsistency (`MEHO_G07_CANARY_LIVE_LLM` vs `G07_CANARY_REAL_LLM`); no retry budget / kebab→snake normalization on `LlmOutputInvalid` in the production grouping pass (product-scope ticket candidate).
- Envless local run with Docker available still boots the PG container before the shelf skip fires (`async_pg_url` instantiates before `_canary_corpus`'s body can skip) — pre-existing behavior, identical under the old fixture chain; cosmetic dev-ergonomics only.

### Docs
- [x] `docs/decisions/spec-reconcile-guards-standard.md` — updated (§Shared-corpus amortisation)
- [x] `docs/codebase/devops.md` — updated
- [x] `docs/cross-repo/g07-vsphere-canary.md` — updated (Known gaps §4)

### What worked well
- The B1 fix follows the required_change exactly and then makes the pattern reusable: the sync-fixture + `asyncio.run` + private-loop shape sidesteps the pytest-asyncio loop-scope trap cleanly, and recording it as the standard for future heavy ingest lanes is exactly the right altitude.
- The unassigned-ratio finding was surfaced honestly — measured, reproduced on the prior shape, documented in three places — instead of being quietly bar-lowered.

The iteration-4 inline B1 thread on `ci.yml` is addressed by `4146b4f6` and can be resolved.

PR #2806, opened 2026-08-04
fix(ci): python-coverage — step-level pytest timeout so a hang can't flip the run to 'cancelled'

damir-topic:
## Summary

- A hang in the push-only `python-coverage` job no longer flips the whole CI run conclusion to `cancelled`. The hang backstop moves from the job-level `timeout-minutes` (expiry **cancels** the job; cancellation — unlike failure — leaks past `continue-on-error` into the run conclusion) to a **step-level `timeout-minutes: 45`** on the pytest step: step timeout → step **failure** → job failure → absorbed by the existing whole-job `continue-on-error: true` → run conclusion stays `success`, quality-gate.yml's `workflow_run` still fires, and a skipped coverage upload stays loud via the #2513 missing-artifact annotation.
- The job-level cap rises 50 → 55 min as the strictly-outer backstop for a wedged non-pytest step (never observed; both 2026-08-04 hangs were in pytest). Step cap 45 + observed pre-pytest setup (35–75 s) + combine/upload (~15 s) means the step timeout always fires first for a pytest hang.
- Incident driving this: run [30902356757](https://github.com/evoila/meho/actions/runs/30902356757) (2026-08-04, merge commit `da1cc71b`) — three attempts, all concluded `cancelled`, 2h10m with zero conclusive CI verdict on main while v0.28.0 pre-flight was blocked.
- ci.yml design-intent comments updated to name the cancelled-vs-failure distinction; `docs/codebase/sonarcloud.md` + `docs/RELEASING.md` pre-flight note updated; CHANGELOG bullet under `[Unreleased]` → Fixed.

## Semantics (verified 2026-08-04)

- `jobs.<job_id>.timeout-minutes` — "The maximum number of minutes to let a job run before GitHub **automatically cancels it**" (github/docs, workflow-syntax reference).
- `jobs.<job_id>.steps[*].timeout-minutes` — "The maximum number of minutes to run the step before **killing the process**" (ibid.); the resulting step conclusion is **Failed**: `actions/runner` `src/Runner.Worker/StepsRunner.cs` sets `step.ExecutionContext.Result = TaskResult.Failed` with the annotation "The action '…' has timed out after N minutes" when the job cancellation token is not requested.
- `jobs.<job_id>.continue-on-error` — "Prevents a workflow run from failing when a job **fails**" (ibid.) — failure-only absorption; a job-cancel leaks through. That is the bug this PR fixes.

## Test plan

[...]

Issue #2513, opened 2026-07-15
Task: G0.35-T3 restore the Python coverage import — un-break the main-only coverage job + fail-visible artifact contract


damir-topic:
Parent goal: #221
Parent initiative: #2510

## Problem

**Backend Python coverage has had no working signal anywhere since 2026-06-20.** The unit lane deliberately dropped `--cov` for OOM reasons (#1982), leaving the dedicated main-only "Python coverage (SonarCloud)" CI job as the sole producer — and that job has been dying at ~47 minutes (runner lost/OOM, null step conclusions) on every main push since 2026-06-20, invisibly, because the job is non-blocking and the downstream consumer degrades silently (`continue-on-error` on the artifact download in quality-gate.yml). SonarCloud's overall coverage collapsed from 65.8% (last good import, 2026-06-19) to ~24%, and backend test-coverage regressions are currently undetectable. The backend tree measured ~88% of its ~42k coverable lines when the import last worked.

## Current state (verified in code + CI history, 2026-07-15, main @ v0.22.0)

- [ci.yml](.github/workflows/ci.yml) coverage job (~lines 520-545): runs on push-to-main or a PR labeled `ci-coverage-validate` (fork-guarded), `runs-on: meho-runners-ci-heavy`, `timeout-minutes: 50`, non-blocking. Its own comment block ([ci.yml:526-528](.github/workflows/ci.yml#L526-L528)) already names the durable fix: *"If the suite grows enough to push -n 1 over the limit, the durable fix is a larger-memory runner pool (infra, out of band) — this job is non-blocking so an OOM merely skips one coverage upload until then."* That contingency is now the steady state.
- The job dies at ~47 min with null step conclusions (runner lost — consistent with OOM/eviction on the heavy pool), i.e. **before** its 50-min timeout; raising `timeout-minutes` would be a no-op.
- [quality-gate.yml:53-62](.github/workflows/quality-gate.yml#L53-L62): `actions/download-artifact` for `python-coverage` with `continue-on-error: true` — "missing artifact degrades to 'no coverage', never blocks". Correct for resilience, but it is also why a 100%-failure streak stayed invisible for 3+ weeks.
- Last good SonarCloud coverage analysis: 2026-06-19T18:55:33Z at 65.8%; first collapsed one: 2026-06-20T10:46:33Z at 24.9%.

## Desired state

[...]

PR #3034, opened 2026-08-19
fix(ci): stop coverage-job pytest hangs cancelling the CI run

### LINKED ISSUE #2865: Task: CI coverage job — pytest hang evades the #2806 step timeout, run flips to cancelled again (2× on the v0.28.0 cut)
## Problem

The `Python coverage (SonarCloud)` job hung twice in a row on the v0.28.0 release-cut commit `fc37709f` (a **docs-only** CHANGELOG roll whose parent `e991b49c` had just passed the full suite, coverage job included), and both times the hang escaped the #2806 backstop: the run conclusion flipped to **`cancelled`** — exactly the failure mode #2800/#2806 was meant to end.

Evidence (run 31199593016):
- Attempt 1, job `92936108608`: cancelled at the job-level cap; step 9 "Pytest (unit) with offline coverage data" still `in_progress`, steps 10+ `pending`.
- Attempt 2 (job-level re-run), job `92951892836`: runner `meho-runners-ci-heavy-j2jw6-runner-n6pcd`, started `17:53:55Z`, killed `18:53:55Z` — exactly the cap; same step-9-in_progress signature.
- Every substantive job on the same run was green (ruff+mypy+pytest, integration, migrations, Go, Helm, CLI snapshot).

Per #2806, the pytest step carries `timeout-minutes: 45` precisely so a hang becomes a **step failure** absorbed by the job's `continue-on-error: true`. That did not happen — the job-level cap (cancellation, which leaks past `continue-on-error`) fired first, both times. Either the step-level timeout didn't apply to the step that actually hung, or the accounting (setup steps eating into wall-clock vs the job cap) means the job cap strictly precedes the step timeout in practice.

Release impact: the v0.28.0 cut tripped the pre-flight "cancelled ≠ green" gate on a commit whose substantive CI was fully green; the tag proceeded on explicit operator judgment. Without a fix this recurs on every release-cut where the coverage duplicate hangs.

Likely environmental trigger: same code passed the identical job an hour earlier; both hangs landed on the `meho-runners-ci-heavy` ARC pool. The runner-side health question is the consumer repo's territory, but the meho-side contract — "a coverage hang can never flip the run conclusion" — is this repo's to enforce.

## Desired state

1. Reproduce the timeout math: confirm whether the step-level `timeout-minutes` on the pytest step can ever fire before the job-level cap given the setup steps' wall-clock, and fix the margins so the step timeout strictly wins (e.g. step 35 min / job 55 min), or move the guard in-process (`pytest-timeout`) so a hang dies as a test failure.
2. On step-timeout, capture a diagnostic before death (e.g. `py-spy dump` or faulthandler output as an artifact) so the next hang says *where* it hung.
3. Verify the #2513 fail-visible contract still holds when the step fails (missing-artifact annotation loud, run conclusion `success`).

## Acceptance criteria

- [ ] A deliberately-hung pytest in the coverage job (temporary test sleep) produces: step **failure** (not job cancellation), run conclusion **success**, hang diagnostic artifact uploaded.
- [ ] The timeout margins are documented in the workflow comment (why step < job with slack).
- [ ] CHANGELOG `[Unreleased]` entry per convention.

## References

- #2800 / PR #2806 — the step-level-timeout fix this escapes.
- Run 31199593016, jobs 92936108608 + 92951892836 (2026-08-07/08, v0.28.0 cut).
- #2513 fail-visible coverage contract; #1987 AC5 (job non-required/non-blocking).

damir-topic:
## Summary

- The push-only `python-coverage` job (SonarCloud coverage producer) could **hang / overrun** in its pytest step and flip the whole CI run conclusion to **`cancelled`** — twice on the v0.28.0 cut (run `31199593016`), and it is still happening on `main` today (see the before/after below). That trips `docs/RELEASING.md`'s "cancelled ≠ green" pre-flight and is release-critical.
- Root cause is **not** a too-short timeout: when the runner is wedged, the #2806 step-level `timeout-minutes` never fires, the job runs to the ~60-min pod reap, and a **`cancelled`** conclusion leaks past the job's `continue-on-error`. A **runner-enforced** timeout cannot end a hang the runner is too wedged to reap — so the guard moves **in-process**.

## Fix

Run the coverage pytest step under **pytest-timeout**, with the tee exit propagated:

```yaml
shell: bash          # self-hosted pool's default shell lacks pipefail (see below)
run: |
  uv run pytest -n 1 --dist loadscope -p no:cov \
    --timeout=300 --timeout-method=signal --max-worker-restart=0 \
    --ignore=tests/integration tests/ 2>&1 | tee pytest-coverage.log
  exit "${PIPESTATUS[0]}"
```

- **`--timeout=300` (signal method)** — a per-test SIGALRM fires inside the pytest process, so a wedged test dies as a `Failed` with the offending test's traceback after 300 s, **independent of runner health**. `continue-on-error` absorbs the failure → run stays `success`.
- **`signal`, not `thread`** — load-bearing under `-n 1` xdist. `thread`'s `os._exit()` reads to xdist as a worker **crash**, which it re-runs up to `--max-worker-restart` times (~5×: one 300 s hang balloons into ~25 min → job-cap cancel) **and** whose stack dump xdist swallows. `signal` raises in the worker's main thread → one clean `Failed`, traceback preserved, no re-run. (Both proven on the real Linux pod — see Test plan.)
- **`--max-worker-restart=0`** — caps any *other* worker death (segfault / OOM-kill; this job peaks ~12.5 GiB, ~1.2 GiB under the pod limit) at zero retries, so a single failure can never multiply into a job-cap overrun.
- **`shell: bash` + `exit "${PIPESTATUS[0]}"`** — the self-hosted pool's default shell does **not** enable `pipefail`, so `pytest | tee` was returning tee's `0` and masking the hang as step SUCCESS (caught by CI validation run `32230484451`). Forcing bash + propagating pytest's own exit makes a hang read as a step **failure**, not a false green.
- **Hang diagnostic artifact** — the step tees to `backend/pytest-coverage.log`, uploaded as `coverage-pytest-log`, so the next in-process-caught hang says *where* it hung.
- **Docs** — the step/job `timeout-minutes` are re-documented in `ci.yml` as the layer-2/3 backstops behind the in-process guard (and why the step cap stays 45, not the 35 floated in #2865: the honest `-n 1` wall exceeds 35 min). `docs/codebase/sonarcloud.md` updated to match. The `coverage.xml` production path is **unchanged**, so the #2513 fail-visible contract still holds.

## Test plan — validated on the real `meho-runners-ci-heavy` pod

- [x] **Local hang-proof** (pytest 9.1.1 + xdist 3.8 + pytest-timeout 2.4.0, `-n 1`): `signal` → clean `Failed` at the timeout with the hung line + preserved traceback, exit 1, no re-run; `thread` reproduced the ~5× worker-crash re-run and swallowed the dump — confirming the `signal` choice.
- [x] **Deliberately-hung pytest on CI (AC-1)** — a temporary hang canary produced a coverage **step failure** (run `32231149189`, step 9 = `failure`), the `coverage-pytest-log` artifact carried `Failed: Timeout (>15.0s) from pytest-timeout` + the exact hung line, the combine + both uploads still ran, and the coverage **job** failure was absorbed by `continue-on-error`. Canary removed before merge (net diff = fix only).
- [x] **Honest full-suite run on this PR (`ci-coverage-validate`)** — run `32231461352`: the coverage job overran the 45-min step cap and reported **`failure`**, and the **run overall concluded `success`** (not cancelled). This is the fix working on the real suite.
- [x] **Before/after proof** — recent **`main`** pushes (no fix) show the coverage job `cancelled` → **run `cancelled`**: `32223700123` (`e7b91676`), `32188022949` (`64d4f1a2`), `32167341560` (`ad652d0a`), `32182712327` (`f52271c0`). Same overrun on this PR → **run `success`**.
- [x] `actionlint` clean on all changed lines; `uv lock --check` no drift; `pytest-timeout` verified **inert** in every other lane (only the coverage step passes `--timeout`; no `timeout` ini anywhere).

## Out of scope (ambient, ops-owned — not this PR)

- The coverage job's honest `-n 1` serial suite now runs ~45 min (grown past the 35:32 baseline) and OOM-brushes the pod near ~47 min, so it fails/degrades to "no coverage this push" on normal runs — the recurring issue `docs/codebase/sonarcloud.md` documents, whose durable fix is a **larger-memory runner pool** (ops-owned, per the job's `# TODO(ops)`). This PR does **not** make coverage complete; it ensures that failure is **absorbed (run stays green)** instead of cancelling the run — which is exactly the meho-side contract #2865 scopes ("a coverage hang can never flip the run conclusion"; runner health is "the consumer repo's territory").
- Pre-existing `.github/actionlint.yaml` gap: `meho-runners-ci-heavy` isn't in its label list (only `meho-runners` / `meho-runners-ci`); actionlint is not wired into CI. Not touched here.

Closes #2865


damir-topic:
> **DEGRADED MODE — machine review identity unavailable.** This verdict is advisory prose, **not** a GitHub review; `reviewDecision` is unchanged and branch protection is not satisfied by this comment. (`op` is not signed in in this autonomous session; the `meho-review-app` token could not be minted — #2733.)

**Anchor task:** #2865 — coverage-job pytest hang evades the #2806 step timeout, run flips to `cancelled`
**Surface area:** devops / CI
**Verdict:** APPROVE
**Head SHA reviewed:** `95bf7569`

### Acceptance criteria
- [x] **A deliberately-hung pytest in the coverage job produces step failure (not job cancellation), run conclusion success, hang diagnostic artifact uploaded** — proven on the real pod. Canary run `32231149189`: coverage **step 9 = `failure`**, coverage **job absorbed by `continue-on-error`**, `coverage-pytest-log` artifact carried `Failed: Timeout (>15.0s) from pytest-timeout` + the exact hung line. (First canary run `32230484451` surfaced a `pytest | tee` pipefail-masking bug — the pool's default shell lacks `pipefail`; fixed with `shell: bash` + `exit "${PIPESTATUS[0]}"` and re-proven.)
- [x] **Timeout margins documented (why step < job with slack)** — `ci.yml` now carries the three-layer defense comment (in-process 300 s → step 45 min → job 55 min), incl. why the step cap stays 45 (the honest `-n 1` wall exceeds 35 min, so the #2865-floated 35 would kill honest runs).
- [x] **CHANGELOG `[Unreleased]` entry** — added under `### Fixed`.
- [x] **(desired-state) #2513 fail-visible contract still holds** — the `coverage.xml` combine/upload path is unchanged; a missing artifact still fires the loud annotation, run stays `success`.

### Integration verification (the coverage job is the integration path — exercised on the real pod)
- [x] **Honest full-suite PR run `32231461352`** — coverage job overran the 45-min step cap → coverage **job `failure`** → **run overall `success`** (not cancelled).
- [x] **Before/after** — `main` (no fix) cancels the run on the same overrun: `32223700123`/`e7b91676`, `32188022949`/`64d4f1a2`, `32167341560`/`ad652d0a`, `32182712327`/`f52271c0` all **run `cancelled`**. This PR: **run `success`**.
- [x] Local hang-proof (installed pytest 9.1.1 + xdist 3.8 + pytest-timeout 2.4.0): `signal` clean-fails; `thread` reproduced the ~5× worker-crash re-run + swallowed dump.
- [x] `actionlint` clean on changed lines; `uv lock --check` no drift; `pytest-timeout` **inert** in every other lane (only the coverage step passes `--timeout`; no `timeout` ini).

### Predicted vs. actual diff
- ✅ `.github/workflows/ci.yml` — coverage step in-process guard + `shell: bash`/PIPESTATUS + diagnostic upload + three-layer timeout docs
- ✅ `backend/pyproject.toml` + `backend/uv.lock` — `pytest-timeout>=2.4.0` dev dep
- ✅ `CHANGELOG.md` — Fixed entry
- ✅ `docs/codebase/sonarcloud.md` — hang-guard section rewritten
- 🕳️ none missing; 🚫 none undisclosed (validation canary reverted; net diff = fix only)

### Findings
No blockers, majors, or minors requiring a code change. No inline comments.

- The honest `-n 1` coverage suite now runs ~45 min (required lane `-n 3` = 18 min; grown past the 35:32 baseline) and OOM-brushes the pod near ~47 min, so the coverage job now fails/degrades on most honest runs and rarely produces coverage. This is the recurring, ops-owned issue `sonarcloud.md` already documents (durable fix = larger-memory runner pool). Out of scope for #2865 — but the growing wall makes that pool provisioning more urgent.
- The hang-diagnostic artifact is captured on the **in-process** timeout path (proven R1/R2). On the **runner step-timeout** path (R3) the abrupt kill left steps 10-12 unrun, so no artifact that run. A future improvement could arm a `faulthandler.dump_traceback_later()` to a file just under the step cap, though reliably uploading after an abrupt runner kill is hard.

### Docs
- [x] `docs/codebase/sonarcloud.md` — updated (three-layer hang defense). No .NET analogies.

### What worked well
- Root-caused past the symptom: identified that a runner-enforced timeout can't reap a wedged runner, so the guard had to move in-process — not a bare timeout bump (which the issue explicitly rejects).
- The `signal`-vs-`thread` decision is load-bearing and empirically justified (the `thread` re-run/​swallow trap was reproduced), and the pipefail-masking bug was caught by real-CI validation rather than shipped.

PR #3048, opened 2026-08-19
fix(connectors): bind9 record.add/remove view-qualified resolution + view-aware verify

### LINKED ISSUE #2897: bind9.record.add fails on zones included in multiple views: zone auto-resolution 'ambiguous', then atomic-apply verify rollback
Hit this on a bind9 9.18 where the same zone is `include`d into two views (a default view plus a per-client view — a fairly common split-horizon layout).

Two failures in sequence with `bind9.record.add`:

1. Without an explicit `zone` param: `connector_error: ZoneResolutionError, exception_message: "ambiguous"` — fair enough, the longest-suffix match sees the zone twice. Passing `zone` explicitly gets past this.
2. With `zone` set: `connector_error: AtomicApplyError, atomic-apply failed at step='verify'` and the change rolls back. The underlying cause (found by running the same edit by hand): `rndc reload <zone>` on a multi-view server fails with `zone '<zone>' was found in multiple views` — the zone-scoped reload needs `rndc reload <zone> IN <view>` (once per view), or a full `rndc reload`. So the zonefile edit + serial bump are fine, the running named just never picks them up, and the dig verify correctly finds nothing → rollback.

The atomic rollback behaved exactly right here, which is nice — no half-applied state. But the op is unusable on multi-view servers.

Suggestions:

- In the reload step, detect the "found in multiple views" failure and either iterate `rndc reload <zone> IN <class> <view>` over the views that include the zone, or fall back to a full `rndc reload` (that's what unblocked us: `bind9.config.reload` after the file edit, and the record served immediately).
- `exception_message` for the AtomicApplyError came through as an empty string after `step='verify': ` — whatever detail the verify step had (the dig output, the rndc stderr) would have saved a manual SSH to diagnose.

Worked around meanwhile with the file edit + full `bind9.config.reload`, so not blocking — filing for the connector's multi-view story.

damir-topic:
## Summary

- Fixes `bind9.record.add` / `record.remove` on split-horizon nameservers where the same zone is declared in more than one `view` block (#2897). Two failures were happening in sequence: zone auto-resolution flattened the per-view rows and reported the single zone name as `ambiguous`; and with an explicit `zone`, the atomic-apply verify used a view-blind `dig @localhost`, which is answered by whichever view matches the loopback source — potentially a different view than the one whose zonefile was staged — so verify failed and the change rolled back even though the edit was correct.
- Adds an optional `view` param (op params + CLI `--view` on `record add` / `record remove`) threaded through a new pure `resolve_zone_target`: it dedupes zone names (so a zone in N views no longer ties with itself), rejects a multi-view zone with an actionable `ambiguous_view` / `view_not_found` that **names the candidate views**, and otherwise edits the named view's zonefile. `ZoneResolutionError` now renders a human-readable sentence rather than the bare reason code (the issue flagged the opaque `"ambiguous"`).
- Makes verify **view-aware**: when the caller passes `view`, verification switches to a view-precise `rndc zonestatus <zone> IN <view>` serial-match (poll-bounded) — `dig` cannot target a view by name. On a no-views / single-view target the record-level `dig` check is kept. Both predicates now **echo their observed state on failure**, so the `AtomicApplyError` detail is no longer empty (the second #2897 complaint).
- The commit-path reload stays a whole-server `rndc reload` — it already reloads every view from disk, and a full reload was the reporter's working workaround; the multi-view fix is in resolution + verify, so the `_atomic.py` pipeline is untouched.

## Test plan

- [x] `ruff check` / `ruff format --check` (`src/`, `tests/`) — clean
- [x] `mypy src/meho_backplane/connectors/bind9/` — clean (repo-wide mypy shows one **pre-existing, macOS-only** `net/icmp.py` `MSG_ERRQUEUE` false positive — Linux-only socket constant, untouched by this PR, passes in CI)
- [x] `pytest tests/test_connectors_bind9_*` — **231 passed**. New: `TestResolveZoneTarget` (all resolution branches incl. `ambiguous_view` / `view_not_found`), `TestViewAwareVerifyBuilders` (verify-command shapes + serial round-trip + metachar quoting), `TestRecordViewHandling` (view param end-to-end through add/remove: pre-stage rejection, view-file selection, zonestatus-vs-dig verify choice)
- [x] Broader sweep `-k "bind9 or classify_op or typed_op or broadcast"` — 371 passed; `test_broadcast_events.py` — 95 passed
- [x] Go CLI `go build / vet / test ./internal/cmd/bind9/` — pass
- [x] `code-quality.py --diff` — 0 blockers (size warnings only, under the hard limit; module carries an existing `code-quality-allow`)
- [x] `rndc reload <zone> IN <view>` and `rndc zonestatus <zone> IN <view>` syntax + the un-indented `serial:` output line verified against ISC/BIND docs (grep anchor excludes the DNSSEC `signed serial:` line)

## Acceptance criteria (#2897)

- [x] Multi-view zone no longer resolves as spurious `ambiguous`; `view` disambiguates, or a clear `ambiguous_view` error names the views.
- [x] Atomic-apply verify is view-aware, so `record.add` works on multi-view zones.
- [x] Empty `exception_message` fixed — verify predicates emit diagnostics on failure.

## Out of scope / follow-ups

- A containerised split-horizon **integration** lane: the existing `tests/integration/test_connectors_bind9_container.py` fixture is single-view. The logic is fully unit-tested and the `rndc` syntax verified against ISC docs, but a real multi-view container test couldn't be validated in this sandbox (no Docker) — sensible follow-up.
- Pre-existing code-quality size warnings on `parse_dig_answer` (untouched) and the record handlers (already covered by the module's `code-quality-allow` header).

Closes #2897

### DIFF
--- a/backend/src/meho_backplane/connectors/bind9/ops_record.py
+++ b/backend/src/meho_backplane/connectors/bind9/ops_record.py
@@ -783,11 +783,15 @@ def _zonestatus_serial_verify(zone_name: str, view: str, serial: int) -> str:
     # Anchored, whitespace-tolerant so ``serial: <N>`` matches whatever
     # indentation the rndc build emits and never a ``signed serial:`` line.
     serial_re = f"^[[:space:]]*serial: {serial}[[:space:]]*$"
+    # ``if ... then exit 0; fi`` rather than ``... && exit 0`` so the
+    # predicate is unambiguously safe under the pipeline's ``set -e``:
+    # a grep miss must fall through to the next poll and, after the
+    # loop, to the diagnostic -- never abort the script early.
     return (
         "STATUS=''; "
         "for _ in 1 2 3 4 5; do "
         f"STATUS=$({zonestatus} 2>&1) || true; "
-        f'printf "%s\\n" "$STATUS" | grep -qE "{serial_re}" && exit 0; '
+        f'if printf "%s\\n" "$STATUS" | grep -qE "{serial_re}"; then exit 0; fi; '
         "sleep 0.3; "
         "done; "
         'printf "zone %s view %s not at staged serial '
@@ -809,9 +813,11 @@ def _dig_add_verify(fqdn: str, ip: str, record_type: str) -> str:
     """
     quoted_fqdn = shlex.quote(fqdn)
     quoted_ip = shlex.quote(ip)
+    # ``if ... then exit 0; fi`` so a grep miss falls through to the
+    # diagnostic under the pipeline's ``set -e`` (not an early abort).
     return (
         f"ANSWER=$(dig @localhost {quoted_fqdn} {record_type} +short 2>&1) || true; "
-        f'printf "%s\\n" "$ANSWER" | grep -qxF {quoted_ip} && exit 0; '
+        f'if printf "%s\\n" "$ANSWER" | grep -qxF {quoted_ip}; then exit 0; fi; '
         f'printf "expected {record_type} %s in the answer for %s; dig +short returned:'
         f'\\n%s\\n" {quoted_ip} {quoted_fqdn} "$ANSWER"; '
         "exit 1"
--- a/backend/src/meho_backplane/connectors/bind9/ops_record.py
+++ b/backend/src/meho_backplane/connectors/bind9/ops_record.py
@@ -14,20 +14,33 @@
 G3.4-T2 (#588) of Initiative #367 landed the read op (``dig @localhost
 <fqdn> [<type>]``). G3.4-T3 (#589) adds the symmetric write ops:
 
-* ``bind9.record.add <fqdn> <ip> [--zone <name>] [--type A|AAAA]`` --
-  atomic stage-validate-commit-reload-verify-rollback against the
-  affected zonefile via :mod:`._atomic`. Verify predicate runs ``dig
-  @localhost <fqdn>`` and asserts the new IP appears in the answer.
-  ``safety_level=caution`` (mutation; the production-path gate is
-  G7/G10 policy territory).
-* ``bind9.record.remove <fqdn> [--zone <name>]`` -- symmetric remove
-  with verify predicate = ``dig`` no longer resolves the FQDN.
+* ``bind9.record.add <fqdn> <ip> [--zone <name>] [--type A|AAAA]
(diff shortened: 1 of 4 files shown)


damir-topic:
> **Advisory note — not a GitHub approval.** Automation shares the PR-author identity (evoila/meho#2733), so a formal review would 422; `reviewDecision` stays `REVIEW_REQUIRED` and branch protection still needs a human approval. This comment records the autonomous review outcome for the merger.

## Merge-ready — autonomous review complete (APPROVE)

**Head:** `5bfd698c` · **Closes #2897**

### Self-review caught + fixed one issue
The initial verify predicates used `<pipeline> && exit 0`, which relies on subtle `set -e` AND-OR-list semantics on the safety-critical atomic-apply verify path (a miss could skip the retry poll / failure diagnostic on some shells → empty-detail regression). Hardened to `if <pipeline>; then exit 0; fi` and **verified end-to-end**: rendering the real predicates and executing them under `set -e -o pipefail` with stub `rndc`/`dig` — match exits 0, miss exits 1 with a non-empty diagnostic.

### Verification
- CI: **all green** — Python unit, **Python integration (testcontainers, real bind9 dispatch)**, Go, DCO, image build.
- Local: `ruff`/`ruff format`/`mypy` clean; `pytest` bind9 231 + broad sweep 371 + broadcast 95; `go build/vet/test` on the CLI; `code-quality --diff` 0 blockers.
- `rndc reload/zonestatus <zone> IN <view>` syntax + the un-indented `serial:` output line verified against ISC/BIND docs (grep anchor excludes the DNSSEC `signed serial:` line).

### Acceptance (#2897)
- [x] Multi-view zone no longer resolves as spurious `ambiguous` (view names deduped; `ambiguous_view` names the candidate views).
- [x] `view` param (op + CLI `--view`) disambiguates split-horizon zones.
- [x] Verify is view-aware (`rndc zonestatus <zone> IN <view>` serial-match) so `record.add` works on multi-view zones.
- [x] Empty `exception_message` fixed — verify predicates emit diagnostics on failure.

### Follow-up (non-blocking)
A containerised split-horizon integration lane (the current bind9 container fixture is single-view). Logic is fully unit-tested and `rndc` syntax doc-verified; a real multi-view container test is a sensible follow-up.

Issue #912, opened 2026-05-22
Task: assign offline_access optional scope to the MCP client (bootstrap helper + onramp docs) — Claude Code requests it for refresh tokens (RDC Wall W7)


damir-topic:
Parent initiative: #803

Parent goal: #221 (G0 Foundational substrate)
Originating: RDC v0.5.0 onramp feedback 2026-05-22, **Wall W7** (newest finding). Follow-up to #790 (onramp docs) + #791 (`bootstrap-clients` helper), both shipped in v0.5.0.

## Problem

Claude Code's MCP client **always requests the `offline_access` scope** during its auth-code + PKCE flow (to obtain a refresh token). On a hardened realm where the public MCP client does not have `offline_access` assigned, Keycloak rejects the authorization request with `invalid_scope`. The RDC deployer hit this today against a v0.5.0-provisioned realm — it's the one onramp wall not covered by the existing recipe (#790) or the `bootstrap-clients` helper (#791).

This is reported alongside confirmation that the MCP surface now works end-to-end: with the client provisioned correctly, `meho_status` over MCP returns the full chain (operator identity from the token, Vault reachable + KV read, DB migrated) — auth-code + PKCE + audience + tenant claims all flow. W7 is the last assignment a deployer must add by hand to get there.

## Current state

- `cli/internal/cmd/admin/keycloak/bootstrap.go` sets the MCP client's `DefaultClientScopes = openid/profile/email + basic/roles/web-origins/acr` (`requiredDefaultScopes()`), plus the `mcp:read`/`mcp:execute` scopes. **No `offline_access`, and no optional-scope reconciliation path** (`reconcileDefaultScopes` exists; there is no `reconcileOptionalScopes`).
- `deploy/values-examples/README.md` onramp recipe + 4-wall matrix don't mention `offline_access`.

## Desired state

A bootstrap-provisioned (or hand-built per the doc) public MCP client carries `offline_access` as an **optional** client scope, so Claude Code's flow — which requests it explicitly — succeeds and gets a refresh token; the onramp doc names W7 in the troubleshooting matrix.

## Fix shape

- **Bootstrap helper:** assign the realm-built-in `offline_access` to the MCP client as an **optional** client scope (add an optional-scope reconcile analogous to `reconcileDefaultScopes`, writing `OptionalClientScopes`). Optional — not default — because the client requests it explicitly; making it default would mint a refresh token into every flow including the device-code CLI client, which doesn't need one.
[...]

PR #3051, opened 2026-08-19
feat(cli): opt-in bounded offline-token path for meho login (#2902)

### LINKED ISSUE #2902: meho-cli has no long-lived-token path — operators re-login on every SSO idle timeout
## What's happening

`meho login` (the RFC 8628 device-code flow) leaves the CLI with a token that dies whenever the Keycloak SSO session idles out. In practice an operator gets kicked back to a full browser device-code login surprisingly often — any time they go more than the realm's idle window without running a `meho` command. And that's common: you're SSHing, editing, running other tools, and the CLI token just sits there aging.

For a governance CLI that operators are meant to reach for many times per task, re-authenticating through a browser several times a day is real adoption friction.

## Why (dug into it on a live deployment)

- Realm `ssoSessionIdleTimeout` = 30 min, `accessTokenLifespan` = 5 min. Fine on its own — the 5-min access token is meant to refresh silently.
- But the refresh token `meho login` receives is an **online** one, bound to the SSO session. Sit unused past the idle window and the SSO session ends, the refresh token goes "not active", and the next call fails with `oauth2: invalid_grant "Token is not active"` → forced re-login.
- The clincher: the `meho-cli` public client is provisioned with `defaultClientScopes` that do **not** include `offline_access`, and `optionalClientScopes: []`. So the CLI has **no way to request an offline token at all** — `meho login --scope offline_access` is a no-op against this client. It's locked into short online sessions by config.
- `revokeRefreshToken: false`, `refreshTokenMaxReuse: 0` — so this is not a rotation/persistence bug. It's purely the online-session + no-offline-path combination.

## How other tools handle this (the bar to match)

- **gcloud / kubectl (oidc-login) / gh** — request an offline/long-lived refresh token (`access_type=offline` / `offline_access`), store it, and **silently mint short access tokens** from it. You log in once and it works for weeks; the refresh token dies only on revocation / long inactivity. This is the common standard for a sporadically-used CLI.
- **Vault** — doesn't use OIDC refresh tokens for its CLI at all: `vault login` mints a **first-class, renewable Vault token** with an explicit `ttl`/`max_ttl` (renew silently up to `max_ttl`, e.g. via Vault Agent). Lifetime is a deliberate policy knob rather than an accident of session idle.
- **Telling detail:** meho's own **MCP** consumer (Claude Code, per the `meho-cli` client description) uses this *same* client and therefore the *same* online-only token — it only stays authenticated because it's exercised on nearly every turn, continuously resetting the SSO idle timer. Leave it idle past the window and it hits the same wall. So the CLI and MCP share the limitation; MCP just masks it by constant use.

`meho-cli` currently matches **neither** the gcloud-style (offline refresh + silent renewal) nor the Vault-style (renewable native token, sane max TTL) model.

## I don't want to presume the fix

Token longevity for a CLI is a genuine threat-model decision, and I'd rather it be made deliberately than have someone (me) reach for the quickest lever. The standard directions, with tradeoffs:

- **Offline tokens for the device-code client** — add `offline_access` to `meho-cli` (default or optional) and have `meho login` request it; matches gcloud/kubectl. Tradeoff: a long-lived credential on the operator's disk, so the OAuth 2.0 Security BCP (draft-ietf-oauth-security-topics) guidance is to pair it with **refresh-token rotation + reuse detection** (`revokeRefreshToken: true`) to bound a stolen token, keep keyring-first storage (the CLI already prefers the OS keyring), etc.
- **A dedicated, longer CLI session policy** — keep online tokens but give the `meho-cli` client its own client-level session idle/max overrides tuned for CLI use instead of inheriting the 30-min human-browser default. Narrower blast radius than offline tokens, capped at `ssoSessionMaxLifespan`.
- **Something else** that fits meho's threat model — you own the auth design; I'm flagging the gap and the standards, not choosing.

Whatever the direction, it should land in the Keycloak **config-as-code** (the realm/clients are `keycloak-config-cli`-managed, so an admin-console hand-edit just gets reverted on the next sync).

Happy to help validate whichever way you pick against a live deployment.

_(Reported from a downstream deployment; sanitized of host/realm/tenant specifics per the public-repo convention. Config values quoted are observed defaults, not deployment secrets.)_

damir-topic:
## Summary

Gives `meho login` a **long-lived, opt-in, bounded** offline-token path so operators stop re-authenticating through a browser device dance on every SSO idle timeout — implementing the direction the maintainer locked in on #2902 (opt-in bounded offline token, not a realm-wide idle bump, not wontfix).

- **`bootstrap-clients --cli-offline-access[=<seconds>]`** — assigns the realm-built-in `offline_access` scope to the `meho-cli` client as an **optional** scope and sets the per-client `client.offline.session.idle.timeout` attribute (bare flag = 48h / `172800`s). Mirrors the #912 MCP-client precedent (`reconcileOptionalScopes` + the client `Attributes` map). **Off by default** — a run that doesn't opt in makes *no extra Keycloak API calls* and leaves the `meho-cli` client exactly as before.
- **`meho login --offline`** — appends `offline_access` (ensuring `openid` alongside) so Keycloak mints a refresh token that survives the idle window; the CLI's existing lazy 401-retry refresh then renews access tokens silently.
- **Security posture** (a long-lived token is sensitive): scoped (optional-only, requested per-login — never minted implicitly), revocable (RFC 7009 endpoint / per-client offline sessions / user consents), and **bounded** by the per-client offline-session idle timeout (default 48h — far tighter than the weeks-long gcloud/kubectl norm, directly answering the documented Wall W7 blast-radius concern). The existing auth model is unchanged; nothing is weakened for non-opters.
- **Docs** updated from "deliberately not given" to "opt-in and bounded, off by default": `deploy/values-examples/README.md` (W7 client-scope table + prose + five-walls matrix) and `docs/codebase/cli.md` (bootstrap step 6, login flow, token-storage note on the macOS keyring→file convergence).

## Framework-research note (refresh.go — please read)

The issue's scope listed "include `offline_access` in the refresh-exchange scopes (refresh.go hardcodes `["openid"]`)". Grounding this in the **installed `golang.org/x/oauth2@v0.27.0`** (`tokenRefresher.Token`, `oauth2.go:275`), the refresh grant sends only `grant_type` + `refresh_token` and **never forwards `Config.Scopes`** — so the hardcoded `["openid"]` was **inert**, and adding `offline_access` there would have been a no-op. It would also be *wrong* to actually send a narrowed scope on refresh: RFC 6749 §6 says an omitted scope preserves the original grant (this is what keeps an offline token offline), whereas explicitly narrowing scope is exactly what Keycloak rejects with `invalid_scope` for offline sessions (keycloak/keycloak#27878, #28463). So I removed the inert line, documented the correct behaviour, and added a regression test asserting the refresh POST carries **no** `scope` param. Refresh-token rotation is deliberately **deferred** per the issue decision.

## Test plan

- [x] `go vet` (changed packages) — clean
- [x] `gofmt -l` (changed files) — clean
- [x] `golangci-lint run` @ **v2.12.2** (CI-pinned) — **0 issues**
- [x] `go test -race ./internal/...` — all pass
- [x] New: `TestBootstrap_CLIOfflineAccessOptInAssignsScopeAndIdle` — opt-in assigns `offline_access` as *optional* (not default) + `client.offline.session.idle.timeout=172800`
- [x] New: `TestBootstrap_DefaultRunLeavesCLIOfflineAccessOff` — off-by-default guarantee (no scope, no attribute)
- [x] New: `TestBootstrap_RejectsNegativeCLIOfflineIdle` — validation rejects a negative bound
- [x] New: `TestTokenBox_Refresh_OmitsScopeToPreserveGrant` — refresh sends `grant_type=refresh_token` and **no** `scope=` (offline grant preserved)
- [x] Existing `TestBootstrap_MCPClientHasOfflineAccessOptionalScope` / `TestBootstrap_IdempotentReRunIsNoOp` still green (no regression)

## Out of scope / notes

- **Refresh-token rotation** (`revokeRefreshToken` / `refreshTokenMaxReuse`) — explicitly deferred per the issue (realm-level, and concurrent CLI+MCP refreshes can trip reuse detection; revisit after the bounded path soaks).
- `govulncheck` reports 4 **Go stdlib** advisories (asn1/net-http, fixed in go1.26.6) reached via pre-existing `internal/discovery/discovery.go` and `internal/api/client.go` — present on `main`, not introduced here; a toolchain bump is out of scope.

Closes #2902


damir-topic:
> satisfied by this comment. (evoila/meho#2733 — the autonomous run
> shares the author identity, so a formal self-approval 422s.)

**Anchor task:** #2902 — meho-cli has no long-lived-token path
**Surface area:** devops (Go CLI + Keycloak config-as-code) + docs
**Verdict:** APPROVE (advisory — see banner)
**Head SHA reviewed:** `31e9c428`

### Acceptance criteria (from the maintainer's #2902 decision comment)
- [x] `bootstrap-clients` opt-in flag assigns `offline_access` as an **optional** scope on `meho-cli` AND sets `client.offline.session.idle.timeout` — verified `TestBootstrap_CLIOfflineAccessOptInAssignsScopeAndIdle`
- [x] Off by default (documented posture unchanged for non-opters) — verified `TestBootstrap_DefaultRunLeavesCLIOfflineAccessOff` + existing `TestBootstrap_MCPClientHasOfflineAccessOptionalScope` still green; the opt-off path makes no extra Keycloak API calls
- [x] `meho login --offline` appends `offline_access` (ensuring `openid`) — `slices.Contains` dedupe; build/vet/lint clean
- [x] Refresh keeps the offline grant alive — `TestTokenBox_Refresh_OmitsScopeToPreserveGrant` (no `scope` param on refresh)
- [x] Documented rationale updated (README W7 + cli.md) from "deliberately not given" → "opt-in and bounded, off by default"

### Predicted vs. actual diff
- ✅ `bootstrap.go` (opt-in scope + attribute + validate), `keycloak.go` (flag + NoOptDefVal + NoArgs), `login.go` (`--offline`), `refresh.go` (inert-scope removal), tests, docs — all expected and correct
- ❓ `cobra.NoArgs` on `bootstrap-clients` — unexpected but **justified + disclosed**: it turns the `--cli-offline-access <n>` space-form (which pflag's `NoOptDefVal` would misparse as a positional) into an error rather than a silent drop. No cobra-level test passes positional args, so it is non-breaking.
- 🕳️ Missing: none. 🚫 Undisclosed: none.

### Integration verification (run in the worktree on head `31e9c428`)
- [x] `go vet` (changed pkgs) — clean
- [x] `gofmt -l` (changed files) — clean
- [x] `golangci-lint run` @ **v2.12.2** (CI-pinned) — **0 issues**
- [x] `go test -race ./internal/...` — all pass (incl. 4 new tests)
- [x] Framework grounding: oauth2 `tokenRefresher` (v0.27.0) never forwards `Config.Scopes`; Keycloak `CLIENT_OFFLINE_SESSION_IDLE_TIMEOUT="client.offline.session.idle.timeout"` (OIDCConfigAttributes 26.x); pflag `NoOptDefVal` `=`-form requirement
- [~] CI (GitHub Actions) — in progress at review time; DCO ✅, CodeQL ✅, TruffleHog ✅; Go + Python jobs queued (local run mirrors the CI Go gate exactly)

### Security review
- Scoped: `offline_access` is **optional**, minted only when `--offline` is explicitly requested — never implicitly
- Bounded: per-client `client.offline.session.idle.timeout` (default 48h) tightens below the realm offline idle (realm remains the upper bound) — far tighter than the gcloud/kubectl weeks-long norm
- Revocable: RFC 7009 endpoint / per-client offline sessions / user consents
- Off by default; existing online-session + refresh behaviour unchanged; keyring-first / 0600-file storage unchanged

### Findings
None. No blockers, no required-change findings.

- Reconcile is additive (matching existing repo convention): *disabling* offline after a prior enable leaves the `offline_access` optional scope attached while clearing the idle attribute. Out of scope for an opt-in-to-enable task; noted for a future prune pass if ever wanted.

### Docs
- [x] `docs/codebase/cli.md` (bootstrap step 6, login flow, token storage) + `deploy/values-examples/README.md` (W7 table/prose/matrix) — updated

### What worked well
- Grounding the refresh-scope decision in the *installed* oauth2 source rather than the issue's assumption caught that the literal ask was inert (and that narrowing scope on refresh would have been an active regression for offline tokens).

PR #3052, opened 2026-08-19
feat(connectors): k8s connector supports vSphere Supervisor (WCP) SSO targets (#2905)

### LINKED ISSUE #2905: k8s connector: support vSphere Supervisor (WCP) targets — they issue no static/durable credential
We tried to register a **vSphere Supervisor** (the WCP "Supervisor cluster" fronting a vSphere workload domain) as a `k8s-1.x` target and hit a wall that needs a connector feature, not a config workaround.

The Supervisor's kube-API is reachable and read-auth works fine — you authenticate with vSphere SSO creds via the `kubectl-vsphere` login (the WCP `/wcp/login` token exchange) and get a working context that can `list pods --all-namespaces` etc. The problem is **durability of the credential**:

- A WCP Supervisor **won't let you mint a static credential.** WCP owns the cluster RBAC, and even `administrator@vsphere.local` is **not** cluster-admin on the Supervisor — live-checked, `auth can-i create serviceaccount` / `create secret` / `*/*` all return **no**. So the read-only-SA + non-expiring-token approach that works for a normal appliance/kubeadm cluster simply isn't available.
- The only credential a Supervisor issues is the **short-lived vSphere-SSO token** from the login exchange (it expires with the SSO session). The `k8s-1.x` connector expects a static kubeconfig/token, so it can't consume this durably.

Net: a Supervisor can't currently be a first-class k8s target, even though it's a real k8s cluster people want to observe. It's the one k8s tier the connector can't reach across our fleet (appliance-internal clusters and everything else register fine).

## The ask

A **vSphere-Supervisor / SSO auth mode for the k8s connector**: the target's secret is vSphere SSO `{username, password}`, and the connector performs the WCP login exchange itself to **mint → cache → refresh** a short-lived Supervisor token, transparently. This is exactly the shape `kubelogin` (OIDC) and `kubectl-vsphere` already implement — the connector holds the credential and does the token dance, instead of expecting a pre-minted static token.

## Two field notes (things you'd only find by trying it live)

- **Internal-VIP redirect.** When the Supervisor is reached over a NAT'd alias (common in a lab or behind an operator VPN), `kubectl-vsphere login` establishes the top-level Supervisor context fine, then tries to open per-namespace/workload sessions against the Supervisor's **raw internal VIP** — which isn't reachable from outside, so those sub-sessions time out. The connector should authenticate against the reachable host and use the top-level Supervisor kube-API context, and not follow the internal-VIP workload-session redirects.
- **Least privilege.** The SSO super-admin works but isn't ideal; the right long-term credential is a scoped read-only vSphere SSO user granted a read-only vSphere-Namespace role. A first cut can accept the SSO admin.

Happy to test a build against a live Supervisor.

_(Reported from a downstream deployment; sanitized of host/realm/tenant specifics.)_

damir-topic:
## Summary

- A **vSphere Supervisor (WCP)** issues no static/durable credential — WCP owns the cluster RBAC and even `administrator@vsphere.local` is not cluster-admin, so the only credential is the short-lived vSphere-SSO token from `POST /wcp/login`. The k8s connector's static-kubeconfig path could not consume that durably, so a Supervisor could not be a first-class `k8s-1.x` target even though it is a real cluster people want to observe.
- Adds a **WCP SSO auth mode**: the connector holds the vSphere SSO `{username, password}` and performs the `/wcp/login` exchange itself to **mint → cache → refresh** the Supervisor bearer token transparently, via `kubernetes_asyncio`'s async `refresh_api_key_hook` (proactive, JWT-`exp`-aware with a short TTL fallback + single-flight locking). The cached `ApiClient` stays put; only its bearer rotates.
- **Fits the target/fingerprint model, no special-casing**: the credential protocol is discriminated by Vault **payload shape** (`kubeconfig` vs `username`+`password`), mirroring the in-repo GitHub App-vs-PAT precedent. The Supervisor still fingerprints as `product="k8s"`, resolves through the single `("k8s","1.x","k8s")` connector, and needs **no target-schema change**. `auth_model` (the identity enum) is untouched; the agent surface is unchanged.
- Honours the two field notes: dials the operator-reachable `target.host` (never the login response's **internal-VIP** `server`, and no `guest_cluster_*` per-workload sub-sessions), and threads `verify_tls` / `tls_ca_pin` through the login TLS + the kube-API `Configuration` (CA-chain verified; hostname assertion relaxed only because the alias, not the cert's VIP SAN, is dialed). No SSO credential or token is ever logged.

New module `connectors/kubernetes/wcp.py` (login + expiry-stamped token + self-refreshing `Configuration`); discriminator `load_kubernetes_credential` in `kubeconfig.py`; connector branches in `_get_api_client` / `_get_ws_api_client`. Legacy `kubeconfig_loader=` injection is preserved (adapted onto the new credential contract), so the static path is byte-identical.

## Test plan

- [x] `ruff check` + `ruff format --check` — clean (5 files)
- [x] `mypy` — clean (19 source files)
- [x] `pytest tests/test_connectors_k8s_wcp.py` — **34 passed** (new suite)
- [x] `pytest -k "k8s or kubernetes or typed_connectors_metadata"` — **453 passed, 1 skipped** (no regressions; legacy `kubeconfig_loader=` path intact)
- [x] AC — connector performs `/wcp/login` and dispatches with the minted token: `test_connector_builds_wcp_client_from_sso_credential`, `test_connector_fingerprint_over_wcp`
- [x] AC — mint→cache→refresh of the short-lived token: `test_configuration_refreshes_token_past_expiry`, `test_configuration_refresh_is_single_flight`, JWT-`exp` vs TTL-fallback expiry tests
- [x] AC — top-level Supervisor context (HTTP Basic, no body), not the internal-VIP redirect: `test_wcp_login_success_top_level_context`, `test_build_configuration_dials_alias_not_internal_vip`
- [x] AC — payload-shape discrimination fits the target model: `test_discriminator_*`

## Out of scope

- **Reactive 401 re-mint** via the dispatcher's `invalidate_session` recovery arm: k8s ops raise `kubernetes_asyncio.ApiException` (not `ConnectorAuthError`), which does not route to that arm today — wiring it is a larger dispatcher change. The proactive refresh (with a safety margin) covers normal SSO-session expiry; a token rejected earlier than its `exp` is a documented follow-up.
- **Least-privilege scoped SSO user**: the connector is credential-agnostic (uses whatever SSO pair is staged); the first cut accepts the SSO admin, and a scoped read-only vSphere-Namespace user is the recommended operator-side credential (documented).
- Sibling **#2902** (operator-CLI login-token longevity) is a different surface (Keycloak OIDC) with no shared implementation.

Closes #2905

### DIFF
--- a/backend/src/meho_backplane/connectors/kubernetes/__init__.py
+++ b/backend/src/meho_backplane/connectors/kubernetes/__init__.py
@@ -40,12 +40,23 @@ from meho_backplane.connectors.kubernetes.connector import (
     product_from_git_version,
 )
 from meho_backplane.connectors.kubernetes.kubeconfig import (
+    CredentialLoader,
+    KubeconfigCredential,
     KubeconfigLoader,
+    KubernetesCredential,
     KubernetesTargetLike,
+    WcpSsoCredential,
     load_kubeconfig_from_vault,
+    load_kubernetes_credential,
     parse_kubeconfig_yaml,
 )
 from meho_backplane.connectors.kubernetes.ops import KUBERNETES_OPS, KubernetesOp
+from meho_backplane.connectors.kubernetes.wcp import (
+    WcpLoginError,
+    WcpToken,
+    build_wcp_api_configuration,
+    wcp_login,
+)
 from meho_backplane.connectors.registry import (
     register_connector,
     register_connector_v2,
@@ -88,14 +99,23 @@ async def register_kubernetes_typed_operations(
 
 __all__ = [
     "KUBERNETES_OPS",
+    "CredentialLoader",
+    "KubeconfigCredential",
     "KubeconfigLoader",
     "KubernetesConnector",
+    "KubernetesCredential",
     "KubernetesOp",
     "KubernetesTargetLike",
+    "WcpLoginError",
+    "WcpSsoCredential",
+    "WcpToken",
+    "build_wcp_api_configuration",
     "load_kubeconfig_from_vault",
+    "load_kubernetes_credential",
     "parse_kubeconfig_yaml",
     "product_from_git_version",
     "register_kubernetes_typed_operations",
+    "wcp_login",
 ]
 
 # v1 entry -- retained for ``get_connector("k8s")`` callers (the
--- a/backend/src/meho_backplane/connectors/kubernetes/connector.py
+++ b/backend/src/meho_backplane/connectors/kubernetes/connector.py
@@ -69,9 +69,13 @@ from meho_backplane.auth.operator import Operator
 from meho_backplane.connectors._shared.system_operator import synthesise_system_operator
 from meho_backplane.connectors.base import Connector
 from meho_backplane.connectors.kubernetes.kubeconfig import (
+    CredentialLoader,
+    KubeconfigCredential,
     KubeconfigLoader,
+    KubernetesCredential,
     KubernetesTargetLike,
-    load_kubeconfig_from_vault,
+    WcpSsoCredential,
+    load_kubernetes_credential,
 )
 from meho_backplane.connectors.kubernetes.ops import KUBERNETES_OPS
 from meho_backplane.connectors.kubernetes.ops_core import (
@@ -80,6 +84,7 @@ from meho_backplane.connectors.kubernetes.ops_core import (
     namespace_row,
     node_row,
 )
+from meho_backplane.connectors.kubernetes.wcp import build_wcp_api_configuration
 from meho_backplane.connectors.schemas import (
     FingerprintResult,
     OperationResult,
@@ -279,6 +284,22 @@ def product_from_git_version(git_version: str) -> str:
     return "vanilla"
 
 
+def _adapt_kubeconfig_loader(loader: KubeconfigLoader) -> CredentialLoader:
+    """Adapt a legacy kubeconfig-dict loader onto the credential-loader contract.
+
+    Preserves the ``kubeconfig_loader=`` injection every existing test
+    uses: the dict the loader returns is wrapped in a
+    :class:`~meho_backplane.connectors.kubernetes.kubeconfig.KubeconfigCredential`,
+    so the static path is byte-identical and only vSphere Supervisor
+    (WCP) targets ever take the new SSO branch.
+    """
+
+    async def _wrapped(target: KubernetesTargetLike, operator: Operator) -> KubernetesCredential:
+        return KubeconfigCredential(await loader(target, operator))
+
+    return _wrapped
+
+
 class KubernetesConnector(Connector):
     """Kubernetes connector -- reads kubeconfig per target, caches the client."""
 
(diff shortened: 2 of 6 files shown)


damir-topic:
> **DEGRADED MODE — machine review identity unavailable.** This verdict is advisory prose, **not** a GitHub review; `reviewDecision` is unchanged and branch protection is not satisfied by this comment. (No `MEHO_REVIEW_APP_TOKEN`; a self-review from the author identity would 422 — evoila/meho#2733.)

**Anchor task:** #2905 — k8s connector: support vSphere Supervisor (WCP) targets (no static/durable credential)
**Surface area:** backend (Python)
**Verdict:** APPROVE
**Head SHA reviewed:** `c86fc28e`

### Acceptance criteria
- [x] Connector holds vSphere SSO `{username,password}` and performs `/wcp/login` to mint the Supervisor token — `wcp_login` (HTTP Basic, no body → top-level context); `test_connector_builds_wcp_client_from_sso_credential`, `test_wcp_login_success_top_level_context`.
- [x] mint → cache → refresh of the short-lived token, transparently — async `refresh_api_key_hook` re-mints inside the expiry margin (JWT-`exp`-aware, TTL fallback, single-flight); `test_configuration_refreshes_token_past_expiry`, `test_configuration_refresh_is_single_flight`.
- [x] Does NOT follow the internal-VIP workload-session redirects; stays on the top-level Supervisor context reached via `target.host` — host overridden to the reachable alias, never the login response's `server`; `test_build_configuration_dials_alias_not_internal_vip`.
- [x] Fits the fingerprint/target model, not special-cased — payload-shape discrimination (`kubeconfig` vs `username`+`password`), the in-repo GitHub App-vs-PAT precedent; product stays `k8s`, no schema change; `test_discriminator_*`.
- [x] Docs updated — `docs/codebase/connectors-kubernetes.md` gains a WCP SSO auth-mode section.

### Project-board alignment
|---|---|---|---|---|---|
| #2905 | Open | Closes | ✅ | ✅ full | close on merge |

Sibling #2902 (operator-CLI OIDC token longevity) is a different surface — correctly not touched.

### Predicted vs. actual diff
- ✅ `connectors/kubernetes/wcp.py` (new — login + expiry-stamped token + self-refreshing Configuration)
- ✅ `connectors/kubernetes/kubeconfig.py` (discriminated `KubernetesCredential` + `load_kubernetes_credential`)
- ✅ `connectors/kubernetes/connector.py` (`_get_api_client` / `_get_ws_api_client` branch on `WcpSsoCredential`; legacy `kubeconfig_loader=` adapted)
- ✅ `connectors/kubernetes/__init__.py` (public exports)
- ✅ `tests/test_connectors_k8s_wcp.py` (34 tests)
- ✅ `docs/codebase/connectors-kubernetes.md`
- 🕳️ Missing: none. 🚫 Undisclosed: none.

### Integration verification (run locally in the worktree on the reviewed head)
- [x] `ruff check` + `ruff format --check` — clean (5 files)
- [x] `mypy` — clean (19 source files)
- [x] `pytest tests/test_connectors_k8s_wcp.py` — 34 passed
- [x] `pytest -k "k8s or kubernetes or typed_connectors_metadata"` — 453 passed, 1 skipped (legacy path intact)
- [x] CI on the PR head — DCO green; Python/Go/Semgrep/CodeQL green (see checks)

### Findings

- `assert_hostname = False` is set (only) when `verify_tls=True` for a WCP client: this is deliberate and load-bearing — the connector dials the operator-reachable alias while the Supervisor cert's SAN is the internal VIP, so hostname assertion cannot hold; the CA chain is still verified. Documented inline + in `wcp.py`/docs. If a future target grows a real SNI/expected-hostname field, this could tighten to a hostname override instead of a blanket disable.
- The WCP path reads `verify_tls` / `tls_ca_pin` via `getattr(target, ...)` with safe defaults rather than widening the narrow `KubernetesTargetLike` Protocol (which would churn every test stub). The concrete `Target` model always carries both; documented in `_wcp_configuration`.
- Reactive 401 re-mint (dispatcher `invalidate_session` arm) is intentionally out of scope: k8s ops raise `ApiException`, not `ConnectorAuthError`, so they don't route to that arm today. Proactive margin-based refresh covers normal expiry; a follow-up could wire the reactive path.

### Docs
- [x] `docs/codebase/connectors-kubernetes.md` — updated (WCP SSO auth mode section). No .NET analogies.

- The token dance rides `kubernetes_asyncio`'s own async `refresh_api_key_hook` seam rather than reinventing a cache-eviction path — the cached `ApiClient` is untouched, only the bearer rotates.
- Payload-shape discrimination reuses the established GitHub App-vs-PAT pattern, so a Supervisor stays a plain `product=k8s` target with zero schema/resolver change — the right altitude for an auth-only difference.
- Tests exercise the real seams (the actual `get_api_key_with_prefix` refresh path, real `load_kube_config_from_dict`, `respx` for the HTTP shape) rather than over-mocking, including error, expiry-fallback, and single-flight cases.

Issue #3096, opened 2026-08-23
feat(connectors): vm.create placement passthrough (resource_pool/datastore/host) — folder-only placement can't pin multi-host clusters


damir-topic:
## Summary

`vmware.composite.vm.create` builds its CreateSpec placement as `{"folder": <moid>}` only — no way to pin resource pool, datastore, or host. On a multi-host cluster with host-local datastores (or any deliberate placement policy), the create either fails at the vendor (placement underspecified for the target inventory) or lands wherever vCenter defaults it. The composite is the ONLY governed path for a nested-hypervisor-ready VM (its `nested_hv` leg — #3093 — rides the vim substrate REST cannot express), so placement must be expressible here.

## Fix

Add optional `resource_pool`, `datastore`, `host` (moids) to `VM_CREATE_PARAMETER_SCHEMA`; thread present values into `create_spec["placement"]` alongside `folder`. Absent → byte-identical behavior. Update response/preview/CLI snapshot/docs (nested-ESXi recipe: placement now first-class).

## Acceptance criteria

- [ ] All three placement fields pass through to the vendor placement spec; regression test asserts the built body.
- [ ] Param-absent byte-identical (exact-envelope test).
- [ ] Schema + preview + CLI snapshot + recipe doc updated.

Issue #2298, opened 2026-07-09
Task: key filter.* listing params off the mount flavor — modern /api rejects legacy filter.-prefixed query params (portgroup.audit + datastore.usage enrichment broken on real vCenter)


damir-topic:
## Summary

Every filtered vCenter listing sends **legacy `/rest`-style `filter.*` query params regardless of mount flavor** — and the modern `/api` surface (what real vCenter 8.x serves) expects bare param names (`datastores=`, `hosts=`, `names=`, …) and **400s the prefixed form**. Live-proven by the RDC re-run on #2294: `datastore.usage`'s vm-placement enrichment 400s on **every** row against vCenter 8.0.3 (`GET:/vcenter/vm?filter.datastores=…`), degrading gracefully to a never-populated field. Grounding on main shows the defect is systematic: `network.portgroup.audit` **always** sends `filter.types` (fully broken on modern mounts), and the typed `host.usage`/`host.network_uplinks` listing legs send `filter.hosts` whenever a host filter is supplied. Unfiltered listings (`GET:/vcenter/host` bare) work — which is why the rest of the re-run passed.

## Current state (verified in code, main)

- Sub-call seam passes `query` verbatim onto the mounted path for both flavors: [`composites/_read.py:262-268`](backend/src/meho_backplane/connectors/vmware_rest/composites/_read.py#L262-L268) (`mount_op_path` → `_get_json(params=query)`).
- `filter.*` senders: `_read.py:523` (`filter.names`), `:580` (`filter.datastores` — the enrichment; its own docstring at `:486-492` anticipated "a vCenter that rejects the filter.datastores query" and treated it as environmental), portgroup.audit's `filter.types`/`filter.networks`/`filter.vdses` legs (`:621+`), host-filter legs elsewhere in `_read.py`; typed ops: [`typed_ops.py:334`](backend/src/meho_backplane/connectors/vmware_rest/typed_ops.py#L334) and [`typed_ops_host_network_uplinks.py:251`](backend/src/meho_backplane/connectors/vmware_rest/typed_ops_host_network_uplinks.py#L251) (`filter.hosts`).
- Mount flavor is already known at the seam: `mount_op_path` resolves `/api` (modern) vs `/rest` (legacy/vcsim) off the established session. vcsim serves the **legacy** flavor, where `filter.*` is the *correct* style — which is why CI never caught this: fixtures/vcsim exercise `/rest`.

## Desired state

[...]

PR #3119, opened 2026-08-24
fix(connectors): vim task fault extraction falls back through the fault body (#3116)

damir-topic:
## Summary

- `vim_task.py::_fault_message` now falls back through the fault body instead of requiring a plain-str `error.localizedMessage`: `localizedMessage` → `fault.faultMessage[*].message` (joined with `; `) → `fault._typeName`. `error_message` degrades to `None` (the callers' `<no fault reported>` path) only when the `TaskInfo` truly carries no fault content — fixing the live 8.0.3 CreateVM_Task fault that surfaced as `<no fault reported>` while the platform fully described it.
- The fallback order is grounded in the vim25 VMODL contract: `LocalizedMethodFault.localizedMessage` is optional on the wire, `MethodFault.faultMessage` is an optional `LocalizableMessage[]` whose per-entry `message` is itself optional, and the concrete fault DataObject always carries its `_typeName` discriminator on live VI-JSON.
- Issue ask 1 (box-tolerant unwrap) was **verified rather than re-added**: `unwrap_vim_value` (#3109) is recursive and `_extract_task_info` funnels the whole `TaskInfo` through it before any field read, so boxed nested primitives (the #3106 class) are already bare inside `_fault_message`. The new tests pin that composition end-to-end so a future shallow-unwrap regression fails loudly.
- Every vim-arm composite (create / reconfigure / clone / disk.grow / snapshot / migrate / maintenance) inherits the improvement through `VimTaskResult.error_message`; `docs/codebase/connectors-vmware-rest.md` records the fallback chain.

## Test plan

- [x] `ruff check src/ tests/` — clean
- [x] `ruff format --check src/ tests/` — 1574 files already formatted
- [x] `mypy src/` — clean except one pre-existing darwin-only artifact in untouched `net/icmp.py` (`socket.MSG_ERRQUEUE` is Linux-only; CI runs Linux)
- [x] `pytest tests/test_connectors_vmware_rest_vim_task.py tests/test_connectors_vmware_rest_composites_write.py` — 150 passed
- [x] Ask 1 (box tolerance): `test_poll_fault_with_boxed_nested_primitives_surfaces_fallback_chain` — fully boxed fault content (boxed `state`, `ArrayOfLocalizableMessage` box, boxed `key`/`message` strings) surfaces the joined message
[...]

Issue #3116, opened 2026-08-24
fix(connectors): vim task fault messages swallowed — '<no fault reported>' while TaskInfo carries a localizedMessage


damir-topic:
## What happened

`vmware.composite.vm.create` (vim arm) failed with:

```
"rollback_reason": "CreateVM_Task creating '<name>' faulted: <no fault reported>"
```

But the task on the vCenter side carried a perfectly good localized message ("The input arguments had entities that did not belong to the same datacenter") — visible immediately via a SOAP client's task list. The composite's poll saw the terminal `error` state correctly (it returned fast, no timeout) but extracted **no fault message**, so the operator gets `<no fault reported>` for a fault the platform fully described.

## Where

`vim_task.py::_fault_message` requires `info["error"]` to be a plain dict and `localizedMessage` to be a plain `str`:

```python
error = info.get("error")
if not isinstance(error, dict):
    return None
localized = error.get("localizedMessage")
return localized if isinstance(localized, str) else None
```

Two gaps, both live-plausible on 8.0.3:

1. **Boxed nested primitives (the #3106 class).** `TaskInfo` arrives as a RetrievePropertiesEx `anyType` val; live 8.0.3 boxes nested primitives (`{"_typeName": "string", "_value": ...}`). The `state` read is box-tolerant (that's why the terminal detection worked), but `localizedMessage` inside the `error` LocalizedMethodFault is read raw — a boxed value fails the `isinstance(str)` check and degrades to `None`.
2. **No fallback chain.** Some vim faults genuinely lack `localizedMessage`; the fault body still carries `fault._typeName` (e.g. `InvalidArgument`) and `fault.faultMessage[].message`. `_fault_message` discards all of it.

## Asks

1. Run the fault extraction through the same box-tolerant unwrap the state/progress reads use.
2. Fallback chain: `localizedMessage` → `fault.faultMessage[*].message` (joined) → `fault._typeName` — never `<no fault reported>` when the TaskInfo has *any* fault content.
3. Regression fixture: a faulted TaskInfo with boxed nested primitives, asserting the message survives to the composite's `rollback_reason`.

Every vim-arm composite (create/reconfigure/clone/disk.grow/migrate) inherits this extractor, so one fix improves the debuggability of the whole family.

PR #3120, opened 2026-08-24
fix(connectors): vm.create folder lookup is datacenter-aware and fail-loud on ambiguity

### LINKED ISSUE #3115: fix(connectors): vm.create folder_lookup resolves display name across ALL datacenters — wrong-DC create on multi-DC vCenter
## What happened (live, governed nested-host build on vCenter 8.0.3)

`vmware.composite.vm.create` with `folder_name: "vm"` plus placement pins (`resource_pool` / `host` / `datastore`, all in datacenter A) dispatched clean, but `CreateVM_Task` faulted:

> The input arguments had entities that did not belong to the same datacenter

The vCenter task log showed the create was attempted **in a different datacenter** (B). Root cause: the composite's `folder_lookup` step resolves the display name via `GET:/vcenter/folder?filter.names=<name>` **unscoped** — and every datacenter has a default VM folder named `vm`. On a multi-DC vCenter the lookup returns multiple matches and the composite picked one from the wrong datacenter, then mixed it with the caller's datacenter-A placement pins.

The failure is silent at lookup time — `folder_lookup` reports success (`steps_succeeded: ["folder_lookup"]`) and the composite only fails later inside the vim task (and the fault message is currently swallowed too — filed separately).

## Why this matters

Any multi-datacenter vCenter + the default folder name = guaranteed wrong-DC create for at least one DC. `folder_name: "vm"` is the most natural spelling an agent will produce.

## Asks

1. **Scope the folder lookup when placement pins are present.** `host` / `resource_pool` / `datastore` pins identify the datacenter; pass `filter.datacenters` on the folder list call (the REST filter supports it) so the lookup can only resolve within the pins' DC.
2. **Fail loud on ambiguity.** If the name still matches >1 folder, return a structured `rolled_back` with the candidate moids instead of silently picking the first row.
3. Optionally accept a folder **moid** (`folder` param alongside `folder_name`) as an explicit pin, mirroring the #3096 placement-pin pattern.

Workaround used live: created a uniquely-named folder in the target DC and passed that name — works, but needed an out-of-band write because folder create has no governed path (filed separately).

damir-topic:
## Summary

- `vmware.composite.vm.create` resolved `folder_name` through an **unscoped** `GET:/vcenter/folder` name filter and silently took the first row. Every datacenter ships a default VM folder named `vm`, so on a multi-datacenter vCenter the most natural spelling picked a wrong-datacenter folder, mixed it with the caller's placement pins, and `CreateVM_Task` faulted only later with "entities that did not belong to the same datacenter" (proven live on vCenter 8.0.3 — #3115).
- Folder resolution is now a ladder shared by both create arms (REST and pre-9.0 vim): an explicit **`folder` moid pin** (new optional param, mirroring the #3096 placement-pin pattern; one of `folder`/`folder_name` required via `anyOf`) skips the lookup entirely; a **unique unscoped match** stays byte-identical (zero extra reads); a **multi-match with placement pins** reverse-maps a pin to its datacenter (datacenter listing + one identity∩`datacenters` intersection probe per datacenter, host → resource-pool → datastore priority — no REST listing row carries a parent-datacenter key, so the FilterSpec intersection is the REST-native reverse map) and re-issues the lookup scoped via `filter.datacenters`; **residual ambiguity refuses** with a structured `rolled_back` carrying `candidate_folders` instead of guessing.
- Queries stay authored in the composite's existing `filter.*` style — `adapt_op_query` (#2298) already rewrites to the bare-name dialect live 8.0.x `/api` mounts require.
- The three scoping reads (`GET:/vcenter/datacenter`, `GET:/vcenter/resource-pool`, `GET:/vcenter/datastore`) joined `_SUB_OPS_VM_CREATE`, so the ingest-reconcile guard pins them against the spec shelf's `vcenter.yaml` (verified locally: 49 passed with the shelf configured).
- Docs: `connectors-vmware-rest.md` vm.create + preview rows, nested-ESXi recipe folder-resolution note, CHANGELOG `[Unreleased]` bullet. Preview echoes the new `folder` pin so approvers see the full blast radius.

## Test plan

- [x] `ruff check src/ tests/` + `ruff format --check src/ tests/` — clean
- [x] `mypy src/` (strict) — clean for this change (the sole error on this machine is the pre-existing Linux-only `socket.MSG_ERRQUEUE` in `net/icmp.py`, absent on darwin; CI runs Linux)
- [x] Full backend suite `pytest tests/` — green locally
- [x] Multi-DC ambiguity with pins → scoped pick: `test_vm_create_multi_dc_ambiguous_folder_rescopes_via_placement_pins` (REST arm, exact call/query sequence: unscoped lookup → datacenter listing → per-DC host probes → rescoped lookup → create in the pins' DC)
- [x] Ambiguity without pins → structured rollback: `test_vm_create_ambiguous_folder_without_pins_returns_structured_rolled_back` (`candidate_folders` + reason naming every moid; envelope validates against `VM_CREATE_RESPONSE_SCHEMA`; no DC reads attempted; nothing gated)
- [x] `folder` moid pin honored on both arms: `test_vm_create_folder_moid_pin_skips_lookup_rest_arm` + `test_vm_create_pre9_folder_moid_pin_skips_lookup` (zero lookup reads, `folder_lookup` omitted from the ledger, pin named in the durable gate params)
- [x] Vim arm scoping + fail-loud: `test_vm_create_pre9_ambiguous_folder_rescopes_then_refuses_residual` (resource-pool probe fallback when no host pin; same-DC duplicates still refuse before any vim traffic)
- [x] Unique-name and pin-absent paths byte-identical: every pre-existing vm.create test passes unmodified (only the register `required` assertion moved to the new `anyOf` shape and the preview echo grew the `folder` key)
- [x] Reconcile lanes against the real pinned `vcenter.yaml` — 49 passed with `MEHO_CONSUMER_DOCS_ROOT` set

## Out of scope

- Scoping the vm.create folder lookup to `filter.type=VIRTUAL_MACHINE` folders (folder.create already does) — live 8.0.x dialect for the bare `type` param is unproven; noted as an adjacent finding.
- The swallowed vim create fault message and the ungoverned folder-create path called out in #3115 as "filed separately".

Closes #3115


damir-topic:
> satisfied by this comment.

**Anchor task:** #3115 — vm.create `folder_lookup` resolves display names across ALL datacenters, wrong-DC create on multi-DC vCenter
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `3984341be9527d64f397bf0266a675f5fd89003a`

### Acceptance criteria

- [x] Placement pins scope the folder list call by datacenter (`filter.datacenters`) — `test_vm_create_multi_dc_ambiguous_folder_rescopes_via_placement_pins` asserts the exact call/query ladder (unscoped lookup → datacenter listing → per-DC host∩datacenter probes → rescoped lookup → create in the pins' DC); the vim arm is covered by `test_vm_create_pre9_ambiguous_folder_rescopes_then_refuses_residual` (resource-pool probe fallback). Both arms share `_resolve_folder_moid`, and queries keep the composite's `filter.*` authoring — the #2298 mount adapter already rewrites to the bare dialect the live 8.0.x `/api` mount requires (the issue's dialect note).
- [x] Fail loud on residual ambiguity with candidate moids — `test_vm_create_ambiguous_folder_without_pins_returns_structured_rolled_back`: structured `rolled_back` with `candidate_folders` + a reason naming every moid and the disambiguation path; envelope validated against `VM_CREATE_RESPONSE_SCHEMA`; nothing dispatched, nothing gated.
- [x] Optional `folder` moid pin skips the lookup on both arms — `test_vm_create_folder_moid_pin_skips_lookup_rest_arm` + `test_vm_create_pre9_folder_moid_pin_skips_lookup`: zero lookup reads, `folder_lookup` omitted from the ledger, pin named in the durable gate params; schema `anyOf` requires one of `folder`/`folder_name` (vm.resize precedent).

### Project-board alignment

|---|---|---|---|---|---|
| #3115 | Open (`evoila/meho`, backplane board) | Closes | ✅ | ✅ full (all three asks) | close on merge |

The two "filed separately" siblings named in #3115 are correctly out of scope: the swallowed fault message landed independently as #3116/#3119 (this branch merged it in), and the ungoverned folder-create path is not claimed.

### Predicted vs. actual diff

- ✅ `_write.py` (op ids + `_SUB_OPS_VM_CREATE` + resolution ladder + both arm wirings), `schemas.py` (param + response), `_write_preview.py` (`folder` echo), three test files, `connectors-vmware-rest.md`, nested-ESXi recipe, CHANGELOG — all predicted, all present and correct.
- ❓ Merge commit `3984341b` (origin/main, #3119 CHANGELOG collision) — disclosed, resolution keeps both `[Unreleased]` sections; no history rewrite.
- 🚫 none; 🕳️ none.

### Integration verification

- [x] `ruff check src/ tests/` — clean (review worktree)
- [x] `mypy` (strict) — connector tree clean; the only `src/` error on the review machine is the pre-existing Linux-only `socket.MSG_ERRQUEUE` in `net/icmp.py` (darwin artifact; CI runs Linux)
- [x] `pytest` — 300 passed across composites write/preview/register/gate, vim_task, nested-ESXi recipe, vcsim integration, and both reconcile lanes **armed against the real pinned `vcenter.yaml`** (spec shelf configured — the three new sub-ops `GET:/vcenter/datacenter`, `GET:/vcenter/resource-pool`, `GET:/vcenter/datastore` resolve against the pinned spec, not just the synthetic fixture)
- [x] CI-mirror unit sweep (`-n 3 --dist loadscope`, implementer run): 12,603 passed; every local red proven environmental (fails identically on clean `origin/main` and/or passes standalone)
- [⚠️ not run] live multi-DC vCenter dispatch — reasoned only; the regression tests encode the live evidence's exact shape
- GitHub CI at post time: 6 pass / 10 pending / 0 fail (DCO ✅, CodeQL ✅)

None. No inline comments posted.

- The lookup could additionally scope `filter.type=VIRTUAL_MACHINE` (folder.create's parent resolution already does) — deliberately skipped: the bare `type` param dialect is unproven on live 8.0.x `/api`, and a regression there would break every named-folder create. Candidate follow-up once live-proven.
- `_resolve_pin_datacenter` (64 lines) and `_resolve_folder_moid` (75 lines) sit above the 60-line warn threshold — docstring-heavy, logic is compact; fine for now.
- When both `folder` and `folder_name` are supplied the pin silently wins (documented in both param descriptions). A refusal would be stricter, but the documented precedence matches the pin pattern.

### Docs

- [x] `docs/codebase/connectors-vmware-rest.md` — vm.create row + preview table updated
- [x] `docs/codebase/connectors-vmware-rest-nested-esxi-recipe.md` — folder-resolution note added
- [x] `CHANGELOG.md` — `[Unreleased]` Fixed bullet (#3115)

- The lazy-rescope ladder keeps the unique-name path byte-identical (zero extra reads) — every pre-existing vm.create test passes unmodified, which is the strongest regression signal in the diff.
- The probe design is spec-grounded: `Vcenter.Host.FilterSpec.datacenters` is documented as "Datacenters that **must contain** the hosts for the hosts to match the filter" (pinned `vcenter.yaml`), so the identity∩datacenter intersection is the vendor-contracted reverse map, not an inferred one.

Issue #1627, opened 2026-06-11
Task: connector NotImplementedError on dispatch flattens to a bare connector_error: NotImplementedError — promote to a structured cause + remediation


damir-topic:
Parent goal: #221
Parent initiative: #1626

## Problem

The RDC cycle-8 report ("enabled spec-ingested vmware L2 ops raise a bare
`NotImplementedError` on dispatch — the `composite_l2_disabled` remediation dead end")
is **misdiagnosed as written**. The framing is "raw-L2 execution is unimplemented in
vmware-rest." That is **false**: `VmwareRestConnector` (priority 1, hand-rolled)
dispatches raw ingested L2 reads successfully — `tests/acceptance/test_vmware_rest_dispatch_smoke.py`
asserts `GET:/vcenter/vm`, `GET:/vcenter/network`, et al. return `status:ok`.

The real defect is the **error surface**. The bare `NotImplementedError` the consumer
saw is `auth_headers()` failing — either the target's `auth_model` is not
`SHARED_SERVICE_ACCOUNT`, or an unconfigured `GenericRestConnector` auto-shim is
resolving instead of the hand-rolled class — and `result_connector_error` then
**flattens it to an opaque `connector_error: NotImplementedError`**, dropping the
(already descriptive) exception message into `extras.exception_message` where the
operator never sees it, with no remediation. This is exactly the opaque-error class the
structured-error convention exists to prevent.

## Current state (verified in code, v0.13.0 / `f6ee330`)

- The two `NotImplementedError` sites both carry actionable messages already:
  - [`connectors/vmware_rest/connector.py:278`](backend/src/meho_backplane/connectors/vmware_rest/connector.py#L278)
    — `auth_headers()` raises with "only supports auth_model=SHARED_SERVICE_ACCOUNT;
    target … requested auth_model=…".
  - [`operations/ingest/connector_registration.py:141`](backend/src/meho_backplane/operations/ingest/connector_registration.py#L141)
    — the auto-shim raises "must be replaced with a per-product Connector subclass before
    dispatch is enabled".
- The flattening happens in
  [`operations/_errors.py:424-443`](backend/src/meho_backplane/operations/_errors.py#L424-L443)
  — `result_connector_error` sets `error=f"connector_error: {type(exc).__name__}"` and
  buries `str(exc)` in `extras`, with no remediation field.
- Raw L2 reads DO execute on the hand-rolled connector:
[...]

PR #3125, opened 2026-08-25
feat(connectors): installer air-gapped depot lifecycle — depot settings, trusted certificates, bundle downloads

### LINKED ISSUE #3121: installer connector: no ops for depot settings + trusted certificates — blocks governed air-gapped 9.1 bring-up at the Versions-and-Bundles gate
## Summary

During a governed VCF 9.1 management-domain bring-up, two required installer-appliance configuration steps had no corresponding operations on the `installer-rest-9.1` typed connector and had to be performed out-of-band via direct API calls (documented fallback, this issue is the required capability-gap filing):

1. **Depot settings** — `GET/PUT /v1/system/settings/depot`. A greenfield VCF 9.x installer cannot pass its "Versions and Bundles" validation check without a connected release depot; in air-gapped labs this is an offline depot. The connector exposes the bringup family (`installer.sddc.spec.validate` / `validation.status` / `bringup.start` / `bringup.status`) but no depot configuration op, so the operator must leave the governed path at exactly the step that gates everything else.
2. **Trusted certificates** — `GET/POST /v1/sddc-manager/trusted-certificates`. When the offline depot serves TLS with a self-signed/private-CA certificate, the depot PUT fails with `VMWARE_DEPOT_CONNECT_FAILURE: Secure protocol communication error` until the depot's certificate is imported into the appliance trust store. The trusted-certificates endpoint exists on the installer and works (validated live: cert POST → 200, subsequent depot PUT → `DEPOT_CONNECTION_SUCCESSFUL`, HTTP 202) — it just isn't surfaced as a connector op.

## Asks

- Add typed ops to `installer-rest-9.1`: `installer.system.depot.get` / `installer.system.depot.set` (params: hostname, port, offline account credential ref) and `installer.system.trusted-certificates.add` (param: PEM certificate; safety_level for the mutating ops should park for approval like the bringup gate).
- The depot-set op should treat `VMWARE_DEPOT_CONNECT_FAILURE` as a structured error carrying the upstream message (same #1627/#1649/#1804 pattern), since the TLS-trust failure mode is the common first-run outcome and the remediation (import cert first) should be discoverable from the error.

## Validated fallback (for the record)

Direct appliance API, token auth (`POST /v1/tokens`), then `POST /v1/sddc-manager/trusted-certificates` with `{"certificate": "<PEM>", "certificateUsageType": "TRUSTED_FOR_OUTBOUND"}`, then `PUT /v1/system/settings/depot` with `{"offlineAccount": {...}, "depotConfiguration": {"isOfflineDepot": true, "hostname": "<depot>", "port": 443}}`. Note this path needs NO appliance shell access — it supersedes the older SDDC-Manager-era `lcm.depot.adapter.httpsEnabled=false` property-file workaround for HTTPS-capable offline depots.

## Context

Found during the governed nested-VCF bring-up program (same series as #3071/#3081 lineage). A follow-up would also cover the 9.1 SddcSpec surface (`vspClusterSpec`, `vcfOperationsSpec`, `vcfOperationsFleetManagementSpec`, `vcfOperationsCollectorSpec`) in the connector's `llm_instructions` for `installer.sddc.spec.validate`, since the 9.1 spec-shape requirements (VSP cluster FQDN + IPv4 pool `{ipRange: {startIpAddress, endIpAddress}}`, ops/fleet/proxy hostnames) are currently undocumented there and had to be reverse-engineered from validator errors.

damir-topic:
## Summary

Implements #3121 (+ its addendum): the three steps that forced the live governed VCF 9.1 bring-up out-of-band now have governed ops — configuring the release depot, trusting its TLS certificate, and explicitly pre-downloading the release bundles. New group `installer-depot`, six typed primitives:

| op | wire | posture |
|---|---|---|
| `installer.system.depot.get` | `GET /v1/system/settings/depot` | safe (passwords scrubbed at every depth) |
| `installer.system.depot.set` | `PUT /v1/system/settings/depot` | caution + approval |
| `installer.system.trusted-certificates.list` | `GET /v1/sddc-manager/trusted-certificates` | safe |
| `installer.system.trusted-certificates.add` | `POST /v1/sddc-manager/trusted-certificates` | caution + approval |
| `installer.bundles.list` | `GET /v1/bundles` | safe |
| `installer.bundles.download` | `PATCH /v1/bundles/{id}` per id | caution + approval (one batch) |

## Design points

- **`depot.set` is the verdict**: the appliance runs its connection check synchronously, so vendor rejection maps to a structured `depot_error` (http_status + errorCode + message) per the #1627/#1649/#1804 pattern — the common `VMWARE_DEPOT_CONNECT_FAILURE` TLS-trust case carries its remediation inline (→ `trusted-certificates.add`, retry). Raw `401`s still propagate to the #2067 session-recovery arm (pinned by test).
- **`bundles.download` batches**: one approval covers the id set; per-bundle vendor rejections are collected (`status="partial"`), not fatal; re-triggering in-flight bundles is harmless, which also makes the dispatcher's 401 re-dispatch safe mid-batch.
- **Secret hygiene**: `depot.get`/`depot.set` echoes are scrubbed recursively by key; park-time previews echo depot identity + account username / cert SHA-256 fingerprint / bundle-id batch (truncated past 20) — never a password or PEM body. All preview builders live in `bringup.py` (the connector's single preview home) with no-secret pins in tests.
- **Institutional knowledge shipped**: the `spec` parameter hint on validate/start(/retry) and the new group's `when_to_use` document the 9.1 SddcSpec rules this program paid for live — the three-unique-`vspClusterSpec`-FQDNs rule (omitting `fleetFqdn` passes validation but 500s bring-up at 'Save VCF Management Components'), the ≤20-char core passwords, the derived `vcf-identity` FQDN, `esaConfig` for nested/all-flash — plus the mandatory explicit GA pre-download and its catalog-pinning rationale.
- **Reconcile lane**: all six wire paths verified against the vendored `vcf-installer-9.1` OpenAPI (`updateDepotSettings`, `addTrustedCertificate`, `getBundles`, `startBundleDownloadByID`); read-constant and write-set pins extended.

## Tests

85 passed locally across the installer modules (scrub behavior, verbatim PUT body vs scrubbed echo, depot_error mapping, 401 propagation on both writes, batch partial-failure collection, preview registration + content no-secret pins, posture map, reconcile pins); ruff + format + mypy clean.

Closes #3121

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,33 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Added — air-gapped depot lifecycle ops on the installer connector (#3121)
+
+- A governed air-gapped VCF 9.1 bring-up had to leave the governed path at
+  exactly the step that gates everything else: configuring the release
+  depot, trusting its TLS certificate, and explicitly pre-downloading the
+  release bundles all had no connector ops (performed out-of-band during
+  the live bring-up program, documented on the issue). Six new typed
+  primitives close the gap (group `installer-depot`):
+  `installer.system.depot.get`/`.set` (settings read is password-scrubbed
+  at every depth; the PUT runs the vendor connection check synchronously
+  and maps rejections to a structured `depot_error` — the common
+  `VMWARE_DEPOT_CONNECT_FAILURE` TLS-trust case carries its remediation
+  inline), `installer.system.trusted-certificates.list`/`.add` (the
+  depot-trust fix; preview echoes a SHA-256 fingerprint, never the PEM),
+  and `installer.bundles.list`/`.download` (the mandatory explicit GA
+  pre-download — one approval per batch, per-bundle vendor rejections
+  collected as `partial`, not fatal). Writes are `caution` +
+  `requires_approval` with secret-free park-time previews. The submit
+  primitives' `spec` hint and the new group's `when_to_use` now carry the
+  live-proven 9.1 SddcSpec field rules (three unique `vspClusterSpec`
+  FQDNs each resolving outside the services pool — omitting `fleetFqdn`
+  passes validation but fails bring-up at "Save VCF Management
+  Components"; ≤20-char core passwords; the derived `vcf-identity` FQDN;
+  `esaConfig` for all-flash/nested), so the next operator reads at
+  discovery time what this program learned from validator errors and
+  inventory stack traces.
+
 ### Added — governed bring-up retry: `installer.sddc.bringup.retry` (#3123)
 
 - A failed VCF management-domain bring-up was unrecoverable through the
--- a/backend/src/meho_backplane/connectors/vcf_installer/bringup.py
+++ b/backend/src/meho_backplane/connectors/vcf_installer/bringup.py
@@ -61,6 +61,7 @@ secrets (e.g. the GOSC composites). It is *not* redacted.
 from __future__ import annotations
 
 import asyncio
+import hashlib
 import time
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
@@ -72,6 +73,9 @@ from meho_backplane.connectors.schemas import OperationResult
 from meho_backplane.connectors.vcf_installer.typed_writes import (
     INSTALLER_BRINGUP_RETRY_OP_ID,
     INSTALLER_BRINGUP_START_OP_ID,
+    INSTALLER_BUNDLES_DOWNLOAD_OP_ID,
+    INSTALLER_DEPOT_SET_OP_ID,
+    INSTALLER_TRUSTED_CERTIFICATE_ADD_OP_ID,
 )
 from meho_backplane.operations._preview import PreviewContext, register_preview_builder
 from meho_backplane.operations.composite import enforce_subop_policy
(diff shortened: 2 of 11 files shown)


damir-topic:
## Review of record (self-review 422 overlay)

Reviewed the full diff against the #3121 asks (+ addendum) and the connector's established patterns:

- **Wire correctness**: all six `METHOD:/path` pairs verified against the vendored 9.1 OpenAPI before implementation (`updateDepotSettings`, `addTrustedCertificate`, `getBundles`, `startBundleDownloadByID`); reconcile read-constant + write-set pins extended so drift is loud. Body shapes match the vendor schemas (`DepotSettings`, `TrustedCertificateSpec`, `BundleUpdateSpec`).
- **Error posture**: `depot.set` maps vendor rejections to structured `depot_error` per the #1627 family, with the `VMWARE_DEPOT_CONNECT_FAILURE` remediation inline; 401s propagate raw on BOTH writes (pinned by two dedicated tests) so the #2067 recovery arm keeps working — the mapping can never swallow the session-expiry signal.
- **Secret hygiene**: recursive key-scrub on the depot read AND the depot.set echo (verbatim wire body vs scrubbed governed echo pinned in one test); previews echo identity/fingerprint/batch only — no-secret pins on all three; PEM never appears in a preview.
- **Batch semantics**: `bundles.download` collects per-bundle failures (`partial`), preserving order; 401 mid-batch re-raises and the re-dispatch is safe because re-triggering downloadNow is per-bundle harmless (documented in the docstring).
- **Registration invariants**: posture map extended to 11 ops; the result-scalars round-trip test now discriminates declared-vs-undeclared hints and pins the hinted count at 5 so it can't go vacuous; the row-fetch filter widened from `installer.sddc.%` to `installer.%` (was silently excluding the new families).
- **Knowledge capture**: `_SDDC_SPEC_HINT` and the `installer-depot` `when_to_use` carry today's live-proven 9.1 rules — reviewed for accuracy against the actual failure evidence (fleetFqdn/FLEET_LCM inventory translation, duplicate-key validation crash, catalog-pinning V&B behavior).

Local: 85 passed + 1 skipped (shelf lane, runs in CI); ruff check/format + mypy clean.

**Verdict: APPROVE**

Issue #3084, opened 2026-08-21
fix(connectors): vcf_installer validate/status JSONFlux reduction swallows the Validation id — submit→poll undrivable through MEHO alone


damir-topic:
## Summary

`installer.sddc.spec.validate` and `installer.sddc.validation.status` return the vendor `Validation` whose set-shaped `validationChecks` correctly rides the JSONFlux reduction — but the reduction **replaces the whole result**, swallowing the top-level scalars: `id`, `executionStatus`, `resultStatus`, `description`. The caller therefore never receives the validation `id` it must pass to `installer.sddc.validation.status` — the submit→poll loop cannot be driven through MEHO alone.

Observed live (first governed validate against a real VCF Installer 9.0.2): dispatch `ok`, JSONFlux handle with all 14 checks and drill-in — but `result` contained only `{row_count, total, sample_rows_returned, sample_bytes, source_key: "validationChecks"}`; the poll id had to be recovered out-of-band (vendor `GET /v1/sddcs/validations` listing).

## Fix

Surface the poll-relevant scalars alongside the reduced check rows — the same pattern `vm.deploy_from_library` uses (`vm_id` stays top-level while `issues` reduce). Either:
- have the two impls return a shaped envelope: `{id, executionStatus, resultStatus, validationChecks: [...]}` with a reduction hint that only wraps `validationChecks`, or
- if the reducer supports scalar-preservation natively, register the hint accordingly.

Same audit for `installer.sddc.bringup.start` / `installer.sddc.bringup.status`: the vendor `SddcTask` scalars (`id`, `status`, phase fields) must survive any reduction of subtask lists — verify and pin with tests.

## Acceptance criteria

- [ ] Governed `spec.validate` response carries the validation `id` + `executionStatus`/`resultStatus` as top-level result fields; checks remain JSONFlux-reduced with drill-in.
- [ ] `validation.status` likewise (id echoes back, statuses top-level).
- [ ] `bringup.start`/`bringup.status` audited for the same scalar-swallow; `SddcTask.id`/`status` pinned top-level by tests.
- [ ] Regression tests assert the scalars are present exactly when the checks/subtasks exceed the reduction threshold.

Issue #2113, opened 2026-06-30
k8s.pod.info: result-reduction silently drops container_statuses (keeps only conditions) for pods above the size threshold


ddzafic:
## Summary

On the typed k8s connector (`k8s-1.x`), `k8s.pod.info` returns the full pod detail object for **small** pods, but for a pod whose serialized detail is large enough to trip result-reduction the response is reduced to a JSONFlux handle exposing **only the `conditions` sub-array** (`source_key: "conditions"`). `containers`, `container_statuses`, `volumes`, `node`, `qos_class`, etc. are absent from both the inline sample and the handle, and paging the handle (`result_query`) yields only more `conditions`. The single most operationally useful section — `container_statuses` (image / ready / restart_count / state) — is exactly what's dropped.

Observed on meho **v0.19.0** (chart 0.19.0, git `077740220a`) via MCP `call_operation`.

## Reproduction

Two `k8s.pod.info` calls against the same k8s target:

**A. A small pod** (a StatefulSet pod with empty `env`, e.g. running `mongo:8.0`) → full flat object inline, **no handle**:
```jsonc
{ "name": "...", "namespace": "...", "status": "Running", "node": "...", "qos_class": "BestEffort",
  "containers": [{ "name": "mongo", "image": "mongo:8.0", "ports": [...], "env": [], "volume_mounts": [...] }],
  "container_statuses": [{ "name": "mongo", "image": "docker.io/library/mongo:8.0", "ready": true, "restart_count": 0, "state": "running" }],
  "init_containers": [], "volumes": [...], "conditions": [ ...5 rows... ] }
```

**B. A larger application pod** (a Deployment pod with a populated `env`/spec) → **reduced**:
```jsonc
"result": { "row_count": 5, "total": 5,
            "sample": [ {"type":"PodReadyToStartContainers","status":"True"}, ... 5 conditions ],
            "source_key": "conditions" },
"handle": { "summary_md": "...Available Tables: `result` (5 rows)...", "total_rows": 5 }
```
`result_query(handle, offset=0, limit=50)` → the same 5 `conditions`, `total_rows: 5`, `truncated: false`. No `containers` / `container_statuses` retrievable anywhere.

## Impact

- `pod.info`'s documented projection (containers image/ports/env, per-container statuses, node, QoS, conditions) is silently truncated to just `conditions` precisely for **real application pods** — the ones with populated env/spec that exceed the threshold. Trivial pods return full data, so the gap is easy to miss in testing.
[...]

PR #3126, opened 2026-08-25
fix(connectors): installer.sddc.bringup.status reduces the ~260-row sddcSubTasks array (#3122)

### LINKED ISSUE #3122: fix(connectors): installer.sddc.bringup.status returns the full ~260-row sddcSubTasks array inline — set-shaped response bypasses JSONFlux reduction
## Summary

While polling a live VCF 9.1 management-domain bring-up through governed dispatch (`call_operation` → `installer.sddc.bringup.status` on `installer-rest-9.1`), every poll returned the complete SDDC task document inline — including the full `sddcSubTasks` array, ~260 rows on a 9.1.0.0 greenfield bring-up, i.e. tens of KB per response. No result handle, no reduction.

Per v0.1-spec §4 (postulate 6), set-shaped responses beyond ~50 rows / 4 KB must come back as a handle plus a reduced view. The sibling validation ops on this same connector already reduce — #3084 was precisely about that reduction (and about it swallowing the `id` the poller needs). `bringup.status` appears to bypass the reduction entirely and pass the vendor document through raw.

## Ask

Reduce `installer.sddc.bringup.status` to a poller-sufficient summary + handle:

- **Inline (reduced view):** task `id`, `status`, `creationTimestamp`, per-state subtask counts (e.g. `{INITIALIZED: n, IN_PROGRESS: n, POSTVALIDATION_COMPLETED_WITH_SUCCESS: n, FAILED: n}`), the names of currently-`IN_PROGRESS` subtasks, and any `FAILED` / `COMPLETED_WITH_FAILURE` subtasks **with their `errors`** — failures must stay inline, because the restart-from-failed-subtask flow depends on seeing them without a second call.
- **Behind a result handle:** the full `sddcSubTasks` array for `result_query` drill-in.

The #3084 lesson generalizes: reduction must preserve drivability. A reduced view carrying only `status` would force a drill-in on every poll and be strictly worse than today's raw pass-through — the counts + active + failures shape above is exactly what a 150 s poll loop needs and nothing more.

## Evidence

Live 9.1.0.0 management-domain bring-up (same program as #3121 / #3084, observed 2026-08-25): the bring-up task carries ~260 subtasks; every 150 s status poll through the governed path returned all of them inline for the whole multi-hour run.

damir-topic:
## Problem

Polling a live VCF 9.1 management-domain bring-up through governed dispatch (`call_operation` → `installer.sddc.bringup.status` on `installer-rest-9.1`) returned the whole vendor `SddcTask` document inline on every 150 s poll — the full `sddcSubTasks[]` array (~260 rows, tens of KB), with **no JSONFlux handle and no reduction**. Per v0.1-spec §4 / postulate 6, set-shaped responses past ~50 rows / 4 KB must return a handle + reduced view.

## Root-cause gap

The reducer *can* reduce this payload (the `test_connectors_vcf_installer_jsonflux_scalars` #3084 tests prove it), but the **live dispatch path never invoked reduction** for this op. In `jsonflux_reducer._detect_collection`, a live `SddcTask` carries **both** `sddcSubTasks[]` **and** a populated `milestones[]` — two *real* list fields — so it hit the #2113 dict-of-arrays **detail-object exemption** and returned `(None, None)`, passing the whole document through verbatim. The `result_scalars` hint (#3084) never even fired, because the payload was never reduced in the first place. The prior `test_bringup_task_with_milestones_is_a_detail_object_passthrough` test had actually *pinned* this buggy behavior (it fed a scalars-only context and asserted `handle is None`).

## Fix

A new opt-in **`result_digest`** reducer hint, threaded exactly like the existing `result_scalars` / `result_ordering` / `pagination_hint` hints (descriptor `llm_instructions` → `dispatcher._result_digest_from_descriptor` → reducer context). It does two jobs:

1. **Collection override** — names the dominant collection, so `_detect_collection` reduces it directly (overriding the multi-list exemption). The exemption is untouched for any op that does not register the hint (re-pinned by test).
2. **Inline digest** — `_row_digest` computes, over the **full** set: `status_counts` (per-state counts; `sum == total_rows`), `in_progress` (names of `IN_PROGRESS` sub-tasks when `active_states` configured), and `failed` (every `FAILED` / `COMPLETED_WITH_FAILURE` sub-task preserved **whole, with its `errors[]`**) when `failed_states` configured.

Failures ride inline so the restart-from-failed flow (`installer.sddc.bringup.retry`) never needs a drill-in — the #3084 drivability lesson generalized. The full `sddcSubTasks` array stays behind the result handle for `result_query`. Registered on the three `SddcTask` bring-up ops (`start` / `retry` / `status`) — symmetric with the shared `_SDDC_TASK_RESULT_SCALARS`.

The hint validation follows the never-raise discipline of the sibling hints: `collection` + `status_field` are load-bearing (a bad value disables the hint and logs once → falls back to heuristic detection); `name_field` / `active_states` / `failed_states` degrade gracefully. An under-threshold task still passes through verbatim (the hint changes *which* collection reduces, never *whether* the thresholds are met).

## Tests

- `test_operations_jsonflux_reducer.py`: collection-override vs the multi-list exemption, absent-collection fall-through, `_row_digest` counts/active/failed-with-errors over the full set, digest skipped when the reduced collection isn't the named one, malformed-hint rejection + graceful defaults, end-to-end reduce of a two-list payload, and malformed-hint → heuristic fallback (never raises).
- `test_connectors_vcf_installer_jsonflux_scalars.py`: digest-hint registration on the 3 SddcTask ops (and its absence on the validation ops); the live two-list gap now **reduces + digests** (260 sub-tasks, 3 in-progress, 2 failed with `errors[]`); the #2113 exemption re-pinned for the non-hinted (scalars-only) context; a small two-list task still passes through even with the digest hint.

`uv run pytest tests/test_connectors_vcf_installer_*.py tests/test_operations_jsonflux_reducer.py tests/test_operations_dispatcher.py tests/test_operations_reducer.py tests/test_vault_kv_list_jsonflux.py tests/test_connectors_k8s_dispatcher_shim.py tests/test_dispatcher_redaction_integration.py` → all green. `ruff check` + `ruff format` clean; `mypy` clean on the three touched source packages.

Closes #3122

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -117,6 +117,26 @@ connector-related release-notes line.
   discovery time what this program learned from validator errors and
   inventory stack traces.
 
+### Fixed — `installer.sddc.bringup.status` reduces the ~260-row sub-task array (#3122)
+
+- Polling a live VCF 9.1 management-domain bring-up returned the whole vendor
+  `SddcTask` document inline on every 150 s poll — including the full
+  `sddcSubTasks[]` array (~260 rows, tens of KB) — with no JSONFlux handle and
+  no reduction, violating v0.1-spec §4. Root cause: the live `SddcTask` carries
+  **both** `sddcSubTasks[]` and a populated `milestones[]`, so the reducer's
+  #2113 multi-list detail-object exemption classified it as a dict-of-arrays
+  and passed it through unreduced — the `result_scalars` hint (#3084) never even
+  fired. New opt-in `result_digest` reducer hint (threaded like `result_scalars`
+  / `result_ordering` / `pagination_hint`): an op names its dominant collection,
+  so the reducer reduces it despite the sibling array, and writes a drivable
+  inline digest — per-`status` counts, the names of `IN_PROGRESS` sub-tasks, and
+  every `FAILED` / `COMPLETED_WITH_FAILURE` sub-task preserved **whole with its
+  `errors[]`** so the restart-from-failed flow (`installer.sddc.bringup.retry`)
+  never needs a drill-in. The full `sddcSubTasks` array stays behind the result
+  handle for `result_query`. Applied to the three `SddcTask` bring-up ops
+  (`start` / `retry` / `status`); the #2113 exemption is unchanged for every op
+  that does not register the hint.
+
 ### Added — governed bring-up retry: `installer.sddc.bringup.retry` (#3123)
 
 - A failed VCF management-domain bring-up was unrecoverable through the
--- a/backend/src/meho_backplane/connectors/vcf_installer/typed_ops.py
+++ b/backend/src/meho_backplane/connectors/vcf_installer/typed_ops.py
@@ -144,6 +144,7 @@ def _instructions(
     output_shape: str,
     parameter_hints: dict[str, str],
     result_scalars: tuple[str, ...] | None = None,
+    result_digest: dict[str, Any] | None = None,
 ) -> dict[str, Any]:
     instructions: dict[str, Any] = {
         "when_to_use": when_to_use,
@@ -157,6 +158,13 @@ def _instructions(
         # summary so the submit → poll loop stays drivable through the
         # governed surface (the reducer's ``_preserved_scalars`` contract).
         instructions["result_scalars"] = {"keys": list(result_scalars)}
+    if result_digest is not None:
+        # #3122: the row-digest hint — names the collection to reduce (so a
+        # two-list SddcTask reduces sddcSubTasks[] instead of passing through
+        # the reducer's multi-list detail-object exemption) and drives the
+        # inline per-state counts / IN_PROGRESS names / FAILED-rows-with-errors
+        # digest (the reducer's ``_row_digest`` contract).
+        instructions["result_digest"] = dict(result_digest)
     return instructions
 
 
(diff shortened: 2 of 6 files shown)


damir-topic:
## Review of record (self-review 422 overlay; implementation by a delegated worker, reviewed independently)

Verified against the #3122 asks and the live evidence (the ~260-row raw pass-through observed on every poll of a real bring-up today):

- **Root cause is correct and now pinned the right way round**: `_detect_collection`'s #2113 multi-list detail-object exemption fired because a live `SddcTask` carries `sddcSubTasks[]` beside a populated `milestones[]` — and the prior test had actually pinned the buggy pass-through. That test now asserts the reduction; the #2113 exemption is re-pinned for non-hinted ops, so both behaviors are load-bearing.
- **Mechanism is generic and opt-in**: `result_digest` threads descriptor → dispatcher → reducer context exactly like `result_scalars` (no schema import at the dispatcher layer; reducer validates). The named-collection override is additive with heuristic fall-through when the named key is absent — the hint can never hard-fail a reduce.
- **Drivability contract honored**: digest computed over the FULL collection (not the sample) — `status_counts` (UNKNOWN-bucketed so counts always sum to row_count), `in_progress` names, `failed` rows WHOLE with `errors[]` inline; `digest_spec.collection == envelope_key` guard prevents digesting a fallback collection; digest merges before reducer-owned bookkeeping so reserved keys still win.
- **Never-raise discipline**: malformed state lists degrade to counts-only with a single warning; verified the resolve happens once per reduce.
- Registered on all three `SddcTask` ops (start/retry/status); `typed_writes.py` untouched and `typed_ops.py` changes localized as briefed (will rebase against #3125's typed_ops changes at merge time).

Independently re-ran: 59 tests across the reducer/scalars/register modules in the PR worktree — green; ruff + mypy clean. Worker's fuller sweep (dispatcher + sibling jsonflux consumers, 210+) also reported green and CI will confirm.

**Verdict: APPROVE**

damir-topic:
Re-verification after rebase onto `df78c9c0` (post-#3125, shared `typed_ops.py`): the only conflict was CHANGELOG-only (both [Unreleased] bullets kept); the `typed_ops.py` auto-merge was inspected — the #3126 `result_digest` kwarg + its three call sites and the #3125 hint wiring + depot op set coexist as intended. Local verification on the rebased head: 90 installer/reducer tests green, ruff clean; full CI green on `ced7aec1`.

**Verdict: APPROVE** (re-affirming the pre-rebase verdict for the rebased head; content unchanged apart from the CHANGELOG seam).

PR #3049, opened 2026-08-19
fix(connectors): guard datastore.usage against global-VM-list enrichment (#2975)

damir-topic:
## Summary

- `vmware.composite.datastore.usage` populated each datastore row's `vm_count`/`vm_names` directly from `GET:/vcenter/vm` with **no cross-row identity check** (`_read.py` `datastore_usage_composite`). When the target silently ignores the per-datastore filter, that sub-op returns the whole-inventory VM list, so every row carries the **identical global list** — the reported symptom (#2975): `vm_count` = cluster-wide total, identical `vm_names` on all 50 rows, poisoning the "which VMs live on this filling-up datastore?" triage question with the same wrong answer everywhere.
- Adds a **cross-row identical-sets guard**: a VM's working directory lives on exactly one datastore, so genuinely-distinct datastores never share a non-empty VM set. An identical **non-empty** set across *every* enriched row is treated as failed enrichment — `vm_count`/`vm_names` nulled + an `enrichment_note` — reusing the existing best-effort null+note contract (#1908), instead of emitting a populated-but-wrong global list.
- The guard's non-empty condition excludes the legitimate all-empty case (every datastore really has zero VMs), and it needs ≥2 enriched rows to compare, so a single-datastore result is never touched.
- Schema (`enrichment_note`/`vm_count` descriptions) and `docs/codebase/connectors-vmware-rest.md` updated to document the guard alongside the existing best-effort-error path.

## Scope split — guard-half only (param root-cause deferred, human-gated)

Per @damir-topic's grounding comment on #2975, the work splits cleanly and **only the guard-half is implemented here** — it "stands regardless of root cause":

- **Done (this PR):** the identical-sets guard, which protects the data on any target no matter *why* the filter was ignored.
- **Deferred (NOT in this PR):** the `GET:/vcenter/vm` filter-param / root-cause fix. The actual cause is **undetermined from code** — either the session mounted legacy `/rest`, or the 8.0.3 target ignores the bare `datastores=` param — and per the grounding comment it **must be diagnosed from the dispatch audit row's request URL/params by a human before the param-half is coded**. The param path is left exactly as-is (`{"filter.datastores": [ds_id]}` through `adapt_op_query`, the #2298 seam). Diff confirms zero changes to the param path.

## Test plan

- [x] `ruff check src/ tests/` — clean
- [x] `ruff format --check` (changed files) — clean
[...]

Issue #2758, opened 2026-08-03
checks: sampled-envelope composite ops aren't assertable per-entity — datastore.usage needs an exact-match filter


damir-topic:
Parent initiative: #2780

## Ops report (verbatim, v0.26.0)

Trying to build the capacity tier of our lab monitoring on Sensors (the actual "datastore is filling / capacity halved" predictors), we hit this: `vmware.composite.datastore.usage` takes no params (`additionalProperties: false` — verified live on 0.26.0: `datastore`/`name`/`filter` all rejected) and returns the sampled envelope `{row_count, total, sample_rows_returned, sample_bytes, source_key}`. A Sensor assertion can JSONPath the envelope — we shipped an inventory-count sensor on `$.row_count`, which works — but there is **no way to assert on a single datastore's free/capacity**, because the rows are behind the sampling and there is no way to narrow the op to one entity.

Ask, in the smallest shape that fits the determinism postulate: an **exact-match filter param** (e.g. `datastore: <name>`) on `datastore.usage` — filtered result small enough to return unsampled, so a Sensor can select `$.rows[0].free_gb` (or equivalent) and threshold it. Filters, not weights: no fuzzy matching, no top-N, no server-side ranking. The same pattern presumably generalizes to the other composite list ops, but datastore.usage is the one that monitors the failure class that has now taken out two of our labs (a vSAN silently losing half its capacity and then filling to the write-stall wall).

## Current state (grounded 2026-08-03) — the asked-for filter already exists

The report's core premise is stale, through no fault of the probe: **`vmware.composite.datastore.usage` is not param-less.** An exact-match `filter_names` param (array of datastore names, forwarded server-side as `filter.names` on the `GET:/vcenter/datastore` listing) has been on the op since it landed (#524) and is present in the v0.26.0 tag (`backend/src/meho_backplane/connectors/vmware_rest/composites/schemas.py:204-220`). The live probe guessed `datastore`/`name`/`filter` — all correctly rejected by `additionalProperties: false` (`schemas.py:219`) — and never hit `filter_names`. That is itself the finding: a competent operator probing live could not discover the param.

What the mechanics say a one-name filter should already deliver:

- Rows carry what the sensor needs: `{id, name, type, capacity, free_space}` (bytes) built at `composites/_read.py:581-587`, returned whole as `{"datastores": [...]}` (`_read.py:621`). Filter consumption at `_read.py:536-540`.
[...]

PR #3184, opened 2026-08-29
fix(connectors): scope datastore.usage VM placement per datastore off vim (#2975)

### LINKED ISSUE #2975: datastore.usage vm-placement enrichment returns the GLOBAL vm list on every row (vm_count/vm_names identical across all datastores)
## What happened

`vmware.composite.datastore.usage` against a vCenter 8.0.3 target on a v0.28.0 deploy (git_sha fc37709f): every one of 50 datastore rows carries `vm_count=<global VM total>` and an IDENTICAL `vm_names` sample — the vCenter's whole inventory, not per-datastore placement. Verified on BOTH surfaces (MCP `call_operation` and REST `/api/v1/operations/call`), twice, hours apart — the count even tracked the real global total as new VMs deployed (111 → 119).

## Correction of the initial report

The first version of this issue claimed a surface divergence (MCP populated vs REST null). That was a consumer-probe artifact: the probe read `.result.sample_rows`, which does not exist on the REST envelope — the reduced rows live under `.handle.sample_rows`. With the right path, REST shows the same populated-global-list. No divergence claim stands. (Same wrong path also invalidates the consumer's earlier "vm_count null" verdict history for #2324 — when the global-list behavior started is unknown.)

## Why it matters

1. **Silently wrong data**: 'which VMs live on this 95%-full datastore' (the classic datastore-full triage question) gets N wrong answers on every row.
2. **It flips #2324's fix-check falsely**: any FIXED-condition shaped as 'vm_count non-null' passes while the data is wrong.

## Hypothesis

The enrichment sub-op `GET /api/vcenter/vm?filter.datastores=<id>` (legacy /rest param, #2324): the target vCenter appears to ignore the unknown param and return the unfiltered VM list. Fix per #2324 (bare `datastores=` param) plus a guard: if per-row VM sets come back identical across rows, treat enrichment as failed (null + note) rather than emitting a populated wrong list.

## Context

Found during a consumer clean-lane session (evoila-bosnia/claude-rdc-hetzner-dc#2485; admitted via #2495). Consumer signal: `datastore-usage-vm-enrichment-global-list-per-row`. Sibling: #2324.

damir-topic:
## Summary

- `vmware.composite.datastore.usage` returned the **GLOBAL** VM list on every
  datastore row (identical `vm_count`/`vm_names`), because its VM-placement
  enrichment scoped each row with a per-datastore
  `GET:/vcenter/vm?datastores=<ds>` filter that some vCenter builds silently
  ignore — returning the whole inventory on every call. The classic "which VMs
  live on this filling-up datastore?" triage question got the same wrong answer
  on every row (#2975).
- Per @damir-topic's grounding pass, the originally-filed param-style root cause
  is **refuted**: the code cannot emit `/api` + `filter.datastores` (the #2298
  seam strips the prefix on modern mounts off the same recorded session path),
  and the real cause is undetermined from code (legacy `/rest` mount vs the
  build ignoring the bare `datastores=` param). So the fix stops depending on
  the server-side filter altogether.
- **The real fix:** `_read_vm_placement` resolves placement authoritatively —
  one unfiltered `GET:/vcenter/vm` (moid + name) plus one `RetrievePropertiesEx`
  over the vim `VirtualMachine.datastore` property (the union of datastores a
  VM's config/disks/snapshots/swap sit on), joined client-side into
  per-datastore `vm_count`/`vm_names`. Each row now carries only the VMs vim
  reports on that datastore regardless of whether any REST filter is honoured; a
  VM whose disks span two datastores legitimately appears under both. Two
  batched reads replace N per-datastore calls.
- The **#3049 identical-sets guard stays** as belt-and-braces. With
  authoritative placement an identical non-empty set across every row cannot
  arise in practice (it would require every VM to sit on every datastore), so
  the guard's firing condition is made **impossible in practice** — exactly the
  ask — while the guard remains to refuse any populated-but-identical anomaly.

## Test plan

- [x] `ruff format --check` (changed files) — clean
- [x] `mypy src/meho_backplane/connectors/vmware_rest/composites/` — Success, no issues
- [x] `code-quality.py --diff` — 0 blockers (4 pre-existing/adjacent function-size warnings)
- [x] **Reproduction test written first, failed on old code, passes on the fix:**
  `test_datastore_usage_scopes_vms_per_datastore_from_authoritative_placement`
  — a fixture where the raw VM listing is the whole inventory (filter ignored)
  yet two datastores hosting disjoint VM sets get different lists.
- [x] New/updated unit tests: authoritative-placement happy path, spanning-VM
  union semantics, batched best-effort degradation, all-empty short-circuit,
  belt-and-braces guard via identical placement, legacy-envelope tolerance,
  legacy-mount listing filter — all green.
- [x] vcsim (respx) integration: modern + legacy mounts assert per-datastore
  scoping off the vim placement read.
- [x] Targeted suite (composites read/reconcile/register + vcsim + per-entity
  sensor): 73 passed, 2 skipped.
- [x] Broad `-k "vmware_rest or datastore or composite"`: 721 passed, 36 skipped.

## Out of scope

- The `network_portgroup_audit` composite has the same per-row `vm_names`
  pattern on a different (networks) surface — untouched, not part of #2975.
- Live-appliance reverification against a real 8.0.3 target (the deploy that
  reported the symptom).

Closes #2975

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,7 +90,7 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
-### Fixed — `datastore.usage` scopes VM placement per datastore off vim, not an ignorable filter (#2975)
+### Fixed — `datastore.usage` scopes VM placement per datastore off vim, not an ignorable filter (#2975 / PR #3184)
 
 - `vmware.composite.datastore.usage` enriched every datastore row with the
   **whole-inventory** VM list — identical `vm_count` (the cluster-wide total)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,24 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Fixed — `datastore.usage` scopes VM placement per datastore off vim, not an ignorable filter (#2975)
+
+- `vmware.composite.datastore.usage` enriched every datastore row with the
+  **whole-inventory** VM list — identical `vm_count` (the cluster-wide total)
+  and identical `vm_names` on all rows — because it scoped placement with a
+  per-datastore `GET:/vcenter/vm?datastores=<ds>` filter that some vCenter
+  builds silently ignore, returning the unfiltered list on every call. The
+  classic "which VMs live on this filling-up datastore?" triage question got
+  the same wrong answer on every row. The composite now resolves placement
+  authoritatively: one unfiltered `GET:/vcenter/vm` (moid + name) plus one
+  `RetrievePropertiesEx` over the vim `VirtualMachine.datastore` property,
+  joined client-side into per-datastore `vm_count`/`vm_names`. Each row carries
+  only the VMs vim reports on that datastore regardless of whether any REST
+  filter is honoured (a VM whose disks span two datastores appears under both),
+  and the two batched reads replace N per-datastore calls. The #3049
+  identical-sets guard stays as belt-and-braces; authoritative placement makes
+  its firing condition impossible in practice.
+
 ### Added — REST + CLI parity for `result_query`: read JSONFlux handles back off MCP (#3179 / PR #3181)
 
 - `POST /api/v1/operations/call` can *mint* a reduced result handle for any
--- a/backend/src/meho_backplane/connectors/vmware_rest/composites/_read.py
+++ b/backend/src/meho_backplane/connectors/vmware_rest/composites/_read.py
@@ -204,6 +204,15 @@ _PROP_DRS_CONFIG = "configurationEx.drsConfig"
 # vm.migrate DRS lookup already relies on.
 _PROP_DRS_RECOMMENDATION = "drsRecommendation"
 
+# vim constants for the datastore-usage VM-placement read. The
+# ``VirtualMachine.datastore`` property is the vim-authoritative set of
+# datastores a VM's files sit on (config + disks + snapshots + swap,
+# unioned) -- the placement source the datastore-usage composite scopes
+# each row from, instead of trusting a server-side ``GET:/vcenter/vm``
+# datastore filter that some builds silently ignore (#2975).
+_VIRTUAL_MACHINE_MO_TYPE = "VirtualMachine"
+_PROP_VM_DATASTORE = "datastore"
+
 # Per-composite sub-op-id tuples. Each tuple lists the raw-REST /
 # vi-json sub-ops the composite issues directly on the connector
 # session. Pre-#2253 these fed the L2 pre-flight check that guarded a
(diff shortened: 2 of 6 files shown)


damir-topic:
> **DEGRADED MODE — machine review identity unavailable.** This verdict is advisory prose, **not** a GitHub review; `reviewDecision` is unchanged and branch protection is not satisfied by this comment. GitHub refuses a formal self-review (422), so the verdict of record is stated explicitly below.

**Verdict: APPROVE** — 0 blockers, 0 majors, 0 minors.

**Anchor task:** #2975 — datastore.usage VM-placement enrichment returned the GLOBAL VM list on every row
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `2e88d258`

### Acceptance criteria
- [x] Reproduction test written first, fails on old code, passes on fix — `test_datastore_usage_scopes_vms_per_datastore_from_authoritative_placement` (confirmed RED pre-fix, GREEN post-fix).
- [x] Each datastore row carries only the VMs placed on it — disjoint fixtures produce different lists; a disk-spanning VM appears under both datastores (union semantics of `VirtualMachine.datastore`).
- [x] Guard #3049 firing condition made impossible in practice, guard retained — placement is authoritative; the guard is still exercised via a contrived identical-placement fixture.
- [x] Best-effort degradation preserved — a failed batched placement read nulls `vm_count`/`vm_names` + notes on every row while keeping `capacity`/`free_space`.
- [x] Mount routing — vcsim modern (`/sdk/vim25`) and legacy (`/rest`) both scope per datastore off the vim read.

### Predicted vs. actual diff
- ✅ `_read.py` (composite rewrite + `_read_vm_placement` + `_extract_props_by_moid` + `VirtualMachine.datastore` constants + manifest), unit/vcsim/sensor tests, docs, CHANGELOG — all expected.
- ❓ None unexpected. `schemas.py` correctly untouched (response shape unchanged); reconcile lane untouched (both `GET:/vcenter/vm` and `RetrievePropertiesEx` were already pinned/served).
- 🕳️ None missing.

### Integration verification
- [x] `ruff format --check` (changed) — clean
- [x] `mypy` (composites package) — clean
- [x] `pytest` targeted (composites read/reconcile/register + vcsim + per-entity sensor) — 73 passed / 2 skipped
- [x] `pytest -k "vmware_rest or datastore or composite"` — 721 passed / 36 skipped
- [x] CI `Python (ruff + mypy + pytest)` and `Python (integration testcontainers)` — green

### Findings (inline)
None. No blockers, majors, or minors.

- A one-datastore `filter_names` query (the #2758 sensor path) now fetches the whole VM inventory + one `RetrievePropertiesEx` over all VMs to enrich a single row, where the old per-datastore filter fetched only that datastore's VMs. Correctness-first tradeoff (the old path returned wrong data); a follow-up could skip VM enrichment when a caller only needs capacity, or chunk very large inventories.
- `_read_vm_placement` is a 67-line function-size WARN (docstring-heavy) — non-blocking.

### Docs
- [x] `docs/codebase/connectors-vmware-rest.md` — updated (best-effort-enrichment section rewritten to describe the authoritative vim-placement mechanism; guard reframed as belt-and-braces).

- The reproduction test was written first and demonstrably fails on the old code — the bug is pinned, not just asserted around.
- Reusing the already-reconciled `GET:/vcenter/vm` + `RetrievePropertiesEx` sub-ops meant zero reconcile-lane churn and the batched read is *cheaper* (2 calls vs N) in the broad case.

Issue #229, opened 2026-05-11
Initiative: G7.1 Tenant conventions + Layer 2 starter — conventions table + 4 verbs + session preamble assembler + MCP resource + docs/examples/consumer-onboarding/CLAUDE.md


damir-topic:
Parent goal: #218

## Scope

Ship the **two-layer tenant-conventions surface** decided in v0.2 planning:

- **Layer 1 — Server-side conventions** (the database-backed rules auto-loaded into the session preamble for any agent connecting *through* MEHO): `tenant_conventions` table, CLI verbs (`meho conventions list / show / edit / history`), session-preamble assembler, MCP resource (`meho://tenant/<id>/conventions`).
- **Layer 2 — Local agent starter** (the in-repo `CLAUDE.md` template consumer repos copy in to teach their local Claude sessions to prefer MEHO features): `docs/examples/consumer-onboarding/CLAUDE.md` + an onboarding guide explaining the layering.

Plus seed content for the `rdc-internal` tenant: the *operational* paragraphs from the consumer's existing `CLAUDE.md` migrated per decision #4 (operational rules only — Vault canonical, secret-handling, no AI-tool-names-in-identifiers; repo-discipline rules stay in repo).

This is **Phase-1 Track C** of the locked v0.2 sequencing — runs in parallel with G3 (Track A) + G6 (Track B) after G0.

**Area:** [backend/src/meho_backplane/conventions/](backend/src/meho_backplane/conventions/) (new module), [backend/src/meho_backplane/api/v1/conventions.py](backend/src/meho_backplane/api/v1/conventions.py) (new router), [cli/internal/cmd/conventions/](cli/internal/cmd/conventions/) (new CLI verb tree), [docs/examples/consumer-onboarding/](docs/examples/consumer-onboarding/) (new doc directory with CLAUDE.md template + ONBOARDING.md guide), new Alembic migration adding `tenant_conventions` table.

The work items grouped under this Initiative:

[...]

PR #3208, opened 2026-08-29
feat(ui): resolve audit + broadcast drawer reference GUIDs to substance

damir-topic:
## What & why

The meho-backplane console's **Audit row** and **Broadcast event** detail
drawers rendered their reference columns raw — `operator_sub` as an OIDC
sub, `target_id` / `parent_audit_id` / `agent_session_id` / `run_id` as
bare UUIDs, `op_id` as a machine handle — so a human reading a drawer saw
identifiers, not "who did what to which target and what came back"
(operator finding, internal#236).

Both drawers now share **one reference-resolution helper + Jinja partial**
that turns every reference into named, linked substance.

## Changes

**New shared modules**
- `ui/references.py` — `resolve_audit_references(...)` resolves an
  `audit_log` row's references, tenant-scoped, reading only stores the
  backplane already has (`targets`, `endpoint_descriptor` /
  `operation_group`, `runbook_runs`, `audit_log`).
- `ui/roles.py` — the tenant-admin role-lift, extracted so both drawers
  gate the replay deep-link identically (previously private to the audit
  routes).
- `templates/_references.html` — the shared macros both drawers render.

**Resolved substance (both drawers)**
- `principal_sub` → display name (from the row's own payload), with a
  **service** marker for a non-interactive agent / delegated-actor
  principal (derived from `agent_session_id` / `actor_sub`).
- `target_id` → target name linked to `/ui/connectors/<name>`.
- `op_id` → the operation's human summary + operation group.
- `parent_audit_id` → a labelled lineage link (`parent: <op> on <target>`).
- `agent_session_id` → the session-replay link (tenant-admin-gated).
- `run_id` → the originating runbook run (state + template).
- `work_ref` → the change ticket, linked out for the `gh:owner/repo#N`
  shorthand.
- A plain-language **"what happened"** line (op + target + outcome), a
  status rendered with meaning (not just the code), and a humanized
  timestamp (absolute + relative).
- The broadcast event drawer is now **one click from its fully-resolved
  audit row** (`BroadcastEvent.audit_id`).

**Graceful degradation** — a reference that can't be resolved (deleted,
cross-tenant, missing, or a store that carries no name for it, e.g. a
chassis HTTP `op_id`) falls back to the raw id with an `unresolved`
marker; the raw id stays reachable everywhere (chip `title` attributes +
the Identifiers section). Never crashes, never blanks.

**No invented data** — every lookup is verified against real code and
tenant-scoped so a resolution can never surface another tenant's name.

Folds in the label-humanization pass (internal#230) across both drawers,
**superseding** it. Out of scope by design: the Broadcast live-feed filter
bug (internal#229).

## Tests

- New `test_ui_references.py`: pure mappers (`resolve_status`,
  `humanize_relative`) + DB resolution — happy-path across every store,
  unresolved / cross-tenant / no-target fallbacks, admin gating.
- Extended `test_ui_audit_drawer.py` and `test_ui_broadcast_filters.py`
  with rendered-substance assertions: resolved names/links, the
  `unresolved` marker on a missing entity, run + parent lineage, the
  what-happened line, the full-audit-row cross-link, and a
  no-raw-GUID-as-primary assertion. Fixtures use synthetic names/ids only
  (public-repo leak hygiene).
- Local gates green: `ruff check src/ tests/`, `ruff format --check`,
  `mypy src/` (clean on Linux; a `socket.MSG_ERRQUEUE` note is a
  darwin-only artifact in an untouched file), and the audit/broadcast UI
  test suites.

## CHANGELOG

Added a `### Changed` bullet under `[Unreleased]`.

Refs evoila-bosnia/meho-internal#236


damir-topic:
## Independent review — verdict of record

An independent reviewer audited this PR against the internal#236 spec and the repo's quality bar. GitHub refuses a formal self-approval (author == approver, 422), so the binary verdict is recorded here as the sanctioned comment-with-verdict overlay.

**Scope reviewed:** resolution correctness (tenant-scoping on every join, graceful degradation, naive/aware datetime, `gh:owner/repo#N` parsing, HTTP-op-id non-match handled), data honesty (only real stores; columns verified against `db/models.py`), security/leak (synthetic fixtures only; decision-#3 payload gate preserved in both drawers), StrictUndefined/Jinja safety (every context key provided, incl. the audit deep-link `{% with %}`), role-lift refactor (byte-identical semantics, all call sites updated), tests (happy-path, unresolved-fallback, cross-tenant no-leak, lineage links, no-raw-GUID-as-primary), and scope discipline (folds in #230, leaves #229 untouched).

**Independent gates:** `ruff check src/ tests/` clean · `ruff format --check` clean · `pytest` (references + audit-drawer + broadcast-filters) 59 passed · `mypy` clean on the changed sources.

Two non-blocking cosmetic observations (no changes required):
- `_resolve_operation` `.limit(1)` may surface a summary from a different impl sharing the same `op_id`; the raw op_id is always shown.
- `TimeRef.absolute` from a SQLite-naive `occurred_at` yields an offset-less ISO string in `<time datetime>` — consistent with pre-existing console behaviour.

Follow-up after review: regenerated `cli/api/openapi.json` (the `_replay_handler` docstring feeds the OpenAPI description; the CLI-snapshot-freshness gate required the re-snapshot). Go client unchanged.

**Verdict: APPROVE**

Issue #3349, opened 2026-09-04
grants: target selectors for runtime-created targets + explicit null-target semantics + governed-subop discovery


damir-topic:
## Context

Service-principal grants clear the service-principal park gate on a per-dispatch basis. `find_live_grant` (`operations/service_grants.py:405-441`) matches `(tenant, principal_sub, op_id, connector_id, target_id)` **exactly, with no wildcard widening** — `target_id` included, and the `NULL` case is matched literally (a null-target grant matches *only* target-less dispatches; docstring + `docs/codebase/service-principal-grants.md:86-92`).

Three related gaps make it impossible to grant a blueprint run by a service principal ahead of time when the blueprint **registers its own targets during the run** (appliances it just deployed): a target-scoped grant cannot pre-exist for a target that does not yet exist, and a null-target grant does not match a target-scoped dispatch — so every write op against a runtime-created target **parks by construction**.

### Ask 1 — null-target is not a wildcard: make it loud at runtime (or add an explicit opt-in)

The matching rule is documented (`docs/codebase/service-principal-grants.md:86-92` — "`target_id` is matched **exactly**, including the targetless (`NULL`) case — no any-target wildcard"), but there is **no runtime signal**. A grant created with `target_id=null` in the belief that null means "any target" simply never matches a target-scoped dispatch; the op parks with a generic service-principal-park reason and no hint that the null was matched literally. This is precisely how a recent provisioning run misfired.

- Return a 4xx / park-reason hint when a service-principal dispatch parks while the only candidate grant for the same `(principal, op_id, connector)` is **null-target**: "a null-target grant does not match a target-scoped dispatch — scope the grant to this target, or opt into any-target."
- OR add an explicit `any_target: true` opt-in on the grant (deny-by-default preserved; the wildcard is requested, never implied by `NULL`).

**Acceptance**
- [ ] A parked service-principal dispatch whose only candidate grant is null-target returns a hint naming the mismatch.
- [ ] (If chosen) `any_target: true` grants match any target for the `(principal, op_id, connector)`; `NULL` without the flag stays target-less-only.
- [ ] Doc updated to state the runtime behavior, not just the matching rule.

### Ask 2 — a target selector for grants (product / name pattern)

[...]

PR #3352, opened 2026-09-04
fix(grants): accept literal query-string op ids (?action=…) in standing grants

### LINKED ISSUE #3350: grants: literal query-string op ids (?action=…) are refused as wildcards — service principals cannot hold grants for power/deploy sub-ops
## Symptom

Creating a service-principal standing grant whose `op_id` contains a literal
query string is rejected with HTTP 422:

```
op_id must be an exact value; wildcards are not permitted (got
'POST:/vcenter/vm/{vm}/power?action=start'). A standing grant is the
operator's explicit per-op review, not a pattern.
```

## Cause

`backend/src/meho_backplane/operations/service_grants.py::_reject_wildcards`
(around :102-119) refuses **any** `?` (and `*`) in a grant scope field as a
glob metacharacter. But several governed op ids legitimately carry a literal
`?action=` query string as part of their exact id:

- `POST:/vcenter/vm/{vm}/power?action=start|stop|reset|suspend` and
  `POST:/vcenter/vm/{vm}/guest/power?action=shutdown|reboot`
  (`connectors/vmware_rest/composites/_write.py::_power_vm_op_id` /
  `_guest_power_vm_op_id`)
- `POST:/vcenter/ovf/library-item/{ovfLibraryItemId}?action=deploy`
  (`_OP_DEPLOY_OVF_LIBRARY_ITEM`)
- `POST:/esx/settings/hosts/{host}/software?action=apply&vmw-task=true`
- every `?action=` endpoint the ingest parser emits
  (`operations/ingest/openapi.py`, `op_id = f"{method}:{path}"`)

## Impact

A service principal can never hold a standing grant for these ops, so every
composite that gates on one — the `vm.create` power-on leg, `vm.power`,
`vm.deploy_from_library` — always parks for a human when run by a service
principal, even after an operator has reviewed and granted it. Grant matching
is exact string equality (`find_live_grant` :405-441), so allowing the literal
`?` is safe.

## Fix

In `_reject_wildcards`: still reject `*` anywhere; reject `?` only when it is
not a well-formed literal query string on an HTTP-style op id (exactly one
`?`, no `*`, one or more `key=value` pairs joined by `&`). Keep the existing
message for real wildcards; add a distinct message for a malformed `?`. No
schema/pydantic-pattern change is needed (op_id has no regex constraint), so
the CLI OpenAPI snapshot is unaffected.

- Labels: bug, priority:high, effort:small, policy

damir-topic:
## Why

Creating a service-principal standing grant whose `op_id` carried a literal
query string was rejected with HTTP 422:

```
op_id must be an exact value; wildcards are not permitted (got
'POST:/vcenter/vm/{vm}/power?action=start').
```

`_reject_wildcards` (`operations/service_grants.py`) treated every `?` as a
glob. But several governed vCenter op ids carry a literal `?action=` query
string as a part of their **exact** id — the `vm.power` / guest-power
composite sub-ops, `POST:/vcenter/ovf/library-item/{ovfLibraryItemId}?action=deploy`,
`POST:/esx/settings/hosts/{host}/software?action=apply&vmw-task=true`, and
every `?action=` endpoint the ingest parser emits (`op_id = f"{method}:{path}"`).
A service principal could therefore never hold a standing grant for them, so
the `vm.power` / `vm.create` power-on / `vm.deploy_from_library` composites
always parked for service principals — even after an operator reviewed and
granted the op. Grant matching is exact string equality (`find_live_grant`),
so allowing the literal `?` is safe.

## Change

- `_reject_wildcards`: still reject `*` anywhere; reject `?` only when it is
  not a well-formed literal query string on an HTTP-style op id (exactly one
  `?`, no `*`, one or more `key=value` pairs joined by `&`; keys/values allow
  `-`/`.` so `?action=apply&vmw-task=true` is covered). A malformed `?` gets a
  distinct message; real wildcards keep the existing message.
- No pydantic `pattern` constraint exists on `op_id` (only length), and the
  REST route / CLI do no separate op-id validation, so nothing else in the
  grants path needed aligning and the **CLI OpenAPI snapshot is unaffected**.
- Docs (`docs/codebase/service-principal-grants.md`) + CHANGELOG `[Unreleased]`
  Fixed bullet.

## Tests

- `test_service_grants.py`: new parametrized accept cases (power/guest-power,
  OVF deploy, multi-param with hyphenated key), new reject cases (`*` in
  path/value, bare `?`, `?a`, `?a=b?c=d`, `*.delete`), and a round-trip test
  that a query-param grant is created and then matched by `find_live_grant`
  keyed by the same string.
- Green locally: `test_service_grants.py`, `test_service_principal_grant_gate.py`,
  `test_api_v1_service_grants.py` (68) + the approvals suites (122). ruff +
  mypy clean.

Closes #3350
Refs #3349

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,20 @@ connector-related release-notes line.
 
 ## [Unreleased]
 
+### Fixed — service-principal grants: literal query-string op ids (`?action=…`) were refused as wildcards (#3350)
+
+- A standing grant whose `op_id` carried a literal query string (e.g.
+  `POST:/vcenter/vm/{vm}/power?action=start`,
+  `POST:/vcenter/ovf/library-item/{ovfLibraryItemId}?action=deploy`) was
+  rejected with a 422 "wildcards are not permitted", because `_reject_wildcards`
+  treated every `?` as a glob. Several governed vCenter ops key each `?action=`
+  verb as its own exact op id, so a service principal could never hold a grant
+  for them and the `vm.power` / `vm.create` power-on / `vm.deploy_from_library`
+  composites always parked. Grant creation now accepts a well-formed literal
+  query string on an HTTP-style op id (exactly one `?`, no `*`, `key=value`
+  pairs joined by `&`) and matches it by exact string equality; `*` anywhere and
+  a malformed `?` are still refused.
+
 ## [0.33.0] - 2026-09-04
 
 ### Added — docs site: new guide pages, client onramps + MCP surface tiers, and a What's new page (#3334 / #3335 / #3342 / #3343 / #3346)
--- a/backend/src/meho_backplane/operations/service_grants.py
+++ b/backend/src/meho_backplane/operations/service_grants.py
@@ -1,5 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright (c) 2026 evoila Group
+# code-quality-allow: cohesive CRUD + enforcement unit (create-time review,
+# lookup, dispatch-time consult) that predates this change at ~650 lines;
+# splitting the grant service from its enforcement path is out of scope for a
+# surgical bug fix.
 
 """Standing scoped auto-approval grants for service principals (#3151 / #3152).
 
@@ -43,6 +47,7 @@ tag.
 
 from __future__ import annotations
 
+import re
 import uuid
 from datetime import UTC, datetime
 from decimal import Decimal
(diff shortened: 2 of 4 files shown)


damir-topic:
**VERDICT: APPROVE** — independent review of `fd599397`.

Surgical, correct, and well-tested. The fix narrows `_reject_wildcards` so `*` is still refused anywhere while a `?` is accepted only for a well-formed literal query string on an HTTP-style `op_id`; matching stays exact string equality, so a literal `?` cannot widen a grant.

**What I verified**

1. **Regex (`service_grants.py:118-121`).** `^[A-Z]+:/[^?*]+\?…$` is fully anchored; the path class `[^?*]+` excludes `?`/`*` so there is exactly one `?` and no `*` reaches the query part; key/value classes `[A-Za-z0-9_.-]` cover hyphenated keys (`vmw-task`). Empirically confirmed accept for the four governed shapes and reject for `*` in path/value, bare `?`, `?a`, `?a=b?c=d`, `*.delete`. Malformed `?` gets the distinct message; the original wildcard message string is unchanged (the API suite still asserts on it).
2. **No contradicting layer.** `ServiceGrantCreate` (`service_grant_schemas.py:53`) constrains `op_id` by length only — no `pattern`. The REST route (`api/v1/service_grants.py:143`) delegates straight to `service.create`; there is no separate Python CLI validator. `find_live_grant` (`service_grants.py:475`) matches with SQL `op_id == op_id` — exact equality preserved.
3. **Delete-shape refusal intact.** In `create()` (`:289-303`) `_reject_wildcards` runs first, then the delete-shape pattern/descriptor checks. Confirmed `DELETE:/vcenter/vm/{vm}?x=y` passes the wildcard gate but is still refused by `fnmatchcase(subject, "DELETE:*")`; fnmatch treats the literal `?` in the subject as a literal (metachars are pattern-only), so the interaction is correct.
4. **Tests exercise the public seam.** New cases call `svc.create(...)` (4 accept / 7 reject) and the round-trip uses `find_live_grant`; the round-trip key `POST:/vcenter/vm/{vm}/power?action=start` is exactly `_power_vm_op_id("start")`, and the accept params match `_guest_power_vm_op_id`, `_OP_DEPLOY_OVF_LIBRARY_ITEM`, and `_OP_HOST_SOFTWARE_APPLY`. The API suite also carries route-level reject tests.
5. **Docs + CHANGELOG.** `service-principal-grants.md` paragraph is accurate; `[Unreleased]` holds exactly one Fixed bullet and nothing else — clean for the imminent roll. No schema change, so no CLI OpenAPI snapshot churn.
6. **Hygiene.** Single commit, DCO signed. `# code-quality-allow` marker is justified (module was exactly 650 lines on main) and follows the existing repo convention; it does not relax ruff/mypy. Only the four expected files change; no leaks.

**Minor (non-blocking)**

- M1 (`service_grants.py:118-121`): query values are restricted to `[A-Za-z0-9_.-]`. This covers every currently-governed `?action=` op id; a future value needing other characters would be rejected and require a one-line widening. Appropriately scoped for a surgical fix — noting the boundary only.

**What I ran** (SQLite-backed, `uv run`): `test_service_grants.py` + `test_service_principal_grant_gate.py` + `test_api_v1_service_grants.py` → 68 passed; approvals suites (api/expiry/queue/result-surface/agent-resume/topology/secret-move/mcp/examples/ui) → 193 passed. `ruff check`, `ruff format --check`, `mypy` on the touched files → all clean.

CI note acknowledged: a shard-1 perf test (`test_ingest_large_spec_event_loop…`, timing) is flaky and unrelated to this change; this verdict is on the code.

PR #3357, opened 2026-09-04
fix(vmware-rest): send inherited:false on DVS portgroup vlan (and security) InheritablePolicy specs

### LINKED ISSUE #3356: vmware: portgroup.create VLAN spec omits inherited:false — vCenter ignores vlan_id and creates an untagged (VLAN 0) portgroup
## Symptom

`vmware.composite.network.portgroup.create` returns `status: "created"` with the
requested `vlan_id` echoed, but the portgroup it creates is **untagged**. Reading
the new portgroup back (`config.defaultPortConfig.vlan`) shows:

```json
{ "_typeName": "VmwareDistributedVirtualSwitchVlanIdSpec", "inherited": true, "vlanId": 0 }
```

The call succeeds, yet the VLAN was silently dropped. Verified live against
vCenter 8.0.3.

## Root cause

`_build_portgroup_vlan_spec` builds the `defaultPortConfig.vlan` spec without the
required `inherited` field:

- `backend/src/meho_backplane/connectors/vmware_rest/composites/_write.py`
  — `_build_portgroup_vlan_spec` (access `VmwareDistributedVirtualSwitchVlanIdSpec`
  and trunk `VmwareDistributedVirtualSwitchTrunkVlanSpec` branches).

Both spec kinds derive from `InheritablePolicy` (via
`VmwareDistributedVirtualSwitchVlanSpec`), whose `inherited` field the pinned
`vi-json.yaml` marks **required**. When `inherited` is omitted, vCenter defaults
it to `true` ("inherit the switch default") and **discards the explicit `vlanId`
/ trunk ranges** — the create returns ok but the portgroup reads back
`inherited: true, vlanId: 0` (untagged).

The unit test `test_network_portgroup_create_access_vlan_spec` asserted only the
mocked wire body (which lacked `inherited`), so it passed against the defective
shape.

## Impact

Every DVS portgroup created through the backplane with a VLAN is silently
untagged; trunk ranges are dropped the same way. This breaks any L2 substrate
recipe (e.g. nested-ESXi labs) that relies on the created portgroup carrying its
VLAN — the operation reports success while producing a VLAN-0 portgroup.

## Fix

Add `"inherited": False` to each VLAN spec branch in `_build_portgroup_vlan_spec`
(access + trunk). `NumericRange` entries are unaffected (not an InheritablePolicy).

## Audit of the other InheritablePolicy specs in `_write.py`

- `vmware.composite.network.portgroup.security.set` — the `DVSSecurityPolicy`
  spec and its per-field `BoolPolicy` values (`allowPromiscuous` / `macChanges` /
  `forgedTransmits`) **already** carry `inherited: false` (`_bool_policy` +
  the `security_policy` dict). **Not affected.**
- No PVLAN spec is built (the composite exposes access + trunk only).
- No other InheritablePolicy specs are assembled in `_write.py`.

So the fix is scoped to the two VLAN spec branches.

## Acceptance

- The `CreateDVPortgroup_Task` wire body for a `vlan_id` (access) create carries
  `"inherited": false` on the `VmwareDistributedVirtualSwitchVlanIdSpec`.
- The `CreateDVPortgroup_Task` wire body for a `vlan_trunk_ranges` (trunk) create
  carries `"inherited": false` on the `VmwareDistributedVirtualSwitchTrunkVlanSpec`.
- Tests assert the wire body carries `inherited: false` for both branches, using
  the VI-JSON `InheritablePolicy` semantics.
- `network.portgroup.security.set` BoolPolicy handling is confirmed unchanged
  (already correct).

damir-topic:
## Why

`vmware.composite.network.portgroup.create` returned `status: created` with the
requested `vlan_id` echoed, but the portgroup it created was **untagged**. Live
against vCenter 8.0.3, the new portgroup read back
`config.defaultPortConfig.vlan = {inherited: true, vlanId: 0}`.

Root cause: `_build_portgroup_vlan_spec` built the access
`VmwareDistributedVirtualSwitchVlanIdSpec` and trunk
`VmwareDistributedVirtualSwitchTrunkVlanSpec` without the `inherited` field.
Both derive from `InheritablePolicy` (via `VmwareDistributedVirtualSwitchVlanSpec`),
whose `inherited` field the pinned `vi-json.yaml` marks required. Omitted, vCenter
defaults `inherited: true` ("inherit the switch default") and discards the
explicit `vlanId` / trunk ranges — so the create succeeds while producing a
VLAN-0 portgroup, breaking any L2 substrate that relies on the created VLAN.
The existing unit test asserted only the mocked wire body, so it passed against
the defective shape.

## Change

- `_build_portgroup_vlan_spec`: send `"inherited": False` on both the access and
  trunk VLAN spec branches.
- Docstring + the `_typeName`-discriminator comment now record that the VLAN
  specs are `InheritablePolicy` subtypes carrying `inherited: false`.
- Docs (`connectors-vmware-rest.md`) + `[Unreleased]` CHANGELOG bullet.

**Audit of the other InheritablePolicy specs in `_write.py`:**
`network.portgroup.security.set` already sends `inherited: false` on both the
`DVSSecurityPolicy` and its per-field `BoolPolicy` values (`allowPromiscuous` /
`macChanges` / `forgedTransmits`) — **not affected, left unchanged.** No PVLAN
spec is built. The fix is scoped to the two VLAN branches.

## Tests

- Strengthened `test_network_portgroup_create_trunk_happy_path` and
  `test_network_portgroup_create_access_vlan_spec` to require `"inherited": false`
  in the `CreateDVPortgroup_Task` wire body (trunk + access).
- `uv run pytest` over the vmware composites write / register / write-body-reconcile
  / l2-ingest-reconcile modules: 230 passed, 21 skipped.
- `ruff check`, `ruff format --check`, and `mypy` clean on the touched files.

Closes #3356


damir-topic:
**APPROVE** — independent review of `51e2eac7`, Closes #3356. Correct, minimal, well-tested fix; no gaps found.

**What I verified**

1. **Both VLAN branches fixed.** `_build_portgroup_vlan_spec` now emits `"inherited": False` on the access `VmwareDistributedVirtualSwitchVlanIdSpec` and the trunk `VmwareDistributedVirtualSwitchTrunkVlanSpec`. The `_typeName` discriminators (VLAN-id / trunk / NumericRange) are intact and nothing else in the two bodies changed. No PVLAN branch exists — the composite exposes access + trunk only, so nothing is missing there.

2. **Audit of every other InheritablePolicy spec confirmed.** The security path (`network.portgroup.security.set`) already sends `inherited: false` on the `DVSSecurityPolicy` and on each per-field `BoolPolicy` (via `_bool_policy`) — verified by reading the code, not just the PR claim. A repo-wide grep for InheritablePolicy subtypes (`DVSSecurityPolicy`, `BoolPolicy`, the two VLAN specs, plus TrafficShaping / UplinkPortTeaming / Long / String / Int policies) shows those three constructions are the *only* InheritablePolicy specs assembled anywhere under the connector — no untreated write-path spec remains.

3. **Regression coverage is real.** Both wire-body tests assert the created `CreateDVPortgroup_Task` body by exact dict equality *including* `"inherited": False` (trunk: full-body equality; access: `defaultPortConfig.vlan` equality). Dropping the field again would fail both — the omission that shipped the bug is now caught.

4. **Read-back present.** The `created` envelope reads `config.defaultPortConfig.vlan` back off the new portgroup and surfaces it as `observed` (`{name, vlan}`), so an untagged (VLAN 0) result is visible to the caller — consistent with the family's verify-by-read-back pattern.

5. **Docs/CHANGELOG accurate.** One `[Unreleased]` Fixed bullet, scoped to #3356 and nothing else (v0.33.1 already sits below it). The `connectors-vmware-rest.md` `network.portgroup.create` row now records the InheritablePolicy `inherited: false` requirement accurately.

6. **Hygiene clean.** DCO sign-off on the single commit; exactly four files touched (source, test, CHANGELOG, docs); no unrelated edits; no CLI/OpenAPI snapshot churn (builder-body change only, no schema change); no sensitive references.

**What I ran (in an isolated worktree on the PR head):**
- `uv run pytest` over the composites write / register / write-register / write-body-reconcile / l2-ingest-reconcile / read-reconcile modules → **232 passed, 23 skipped**.
- `ruff check` → all checks passed; `ruff format --check` → already formatted; `mypy` on the touched source → no issues.

Issue #3143, opened 2026-08-26
Task: field-test the Claude client onramps — Code plugin + Desktop .mcpb on three operator machines


damir-topic:
Parent initiative: #3128

> **Human-verification task** — assigned to named testers; **not** a candidate for `/auto-implement` or any autonomous execution. Everything here runs on the testers' own machines.

## Target

Field-test the two Claude client onramps shipped by #3128 on three real operator machines:

- **Part A** — the Claude Code plugin ([`clients/claude-code-plugin/`](../blob/main/clients/claude-code-plugin/), installed via the in-repo marketplace)
- **Part B** — the Claude Desktop `.mcpb` bundle ([`clients/claude-desktop-mcpb/`](../blob/main/clients/claude-desktop-mcpb/))

Testers: @ilhan-karic (Ilhan Karic), @ddzafic (Dzenan Dzafic), @setaelmedin (Elmedin Šeta)

## Context

All seven #3128 tasks are merged (PRs #3136–#3142), CI-verified, and reviewed — but every install check that needs a live VPN machine, a real Claude account, and a real browser OAuth flow is still open. This task is that check. Two specific risks need real hardware:

1. **Windows OAuth handshake** (PR #3138 review, non-blocking adjacent note): on `win32` the launcher spawns `npx.cmd` with `shell:true`, and cmd.exe may mangle the always-present JSON OAuth args (embedded quotes + a space in the scope). **If any tester has a Windows machine, their Part B run is the most valuable data point in this task.** A failure here is an *expected possible outcome* — report it loudly, it becomes a follow-up fix, not a reason to retry silently.
2. **Fresh-user friction** — all sandbox testing was done by the machine that wrote the code. Three fresh installs will surface anything the docs assume silently.

## Prerequisites (all testers, both parts)

- A machine on the VPN that can reach the backplane (same posture as the proven #2666 shim recipe).
- Node.js ≥ 18 with `npx` on `PATH`.
- Your **own** Claude account (Claude Code login for Part A, Claude Desktop login for Part B) — that's the whole point of the flipped direction.
- Your **own** MEHO identity on the realm (you log in via the Keycloak browser page during OAuth; no shared credentials).
- The backplane base URL for the lab tenant — take it from the team channel or `docs/cross-repo/mcp-client-setup.md`; **don't paste it into this public issue.**
- Optional but valuable for Part A: the `meho` CLI installed and logged in (the SessionStart digest only renders with it; testing *without* it is equally useful — see A7).

[...]

Issue #2666, opened 2026-07-22
Task: Claude Desktop → internal MEHO smoke test — first empirical run of the mcp-remote shim path (over VPN)


damir-topic:
Parent initiative: #2665 (clean-room evaluation round 1)
Parent goal: #2661

## Summary

Empirically execute a Claude Desktop → MEHO backplane connection for the first time, over the path that fits MEHO's deployment model: the backplane stays **internal-only** (VPN / internal-CA, never publicly exposed — a design invariant), and Claude Desktop reaches its `/mcp` surface through a **local `mcp-remote` stdio shim** running on an operator machine that is already on the VPN. The surface exists in code and docs — Streamable HTTP at `POST /mcp`, OAuth 2.1 + PKCE with RFC 9728 protected-resource metadata — but no one has run it from Claude Desktop. Roughly one hour of work de-risks the clean-room evaluation program and the v1.0 Desktop story.

> **Why not the remote Custom Connector.** claude.ai / Claude Desktop *remote* Custom Connectors are brokered through **Anthropic's cloud**, which fetches the RFC 9728 metadata and runs the OAuth handshake — so that path structurally requires a **publicly reachable** backplane with real public TLS. MEHO is internal-only by design and is never publicly exposed (not even a sacrificial copy), so the remote Custom Connector is **not a valid consumption mode** for this product. The reachable path is the local stdio→HTTP shim (`mcp-remote` or equivalent — post-#2744, Step 3 § Claude Desktop in `docs/cross-repo/mcp-client-setup.md`): the shim runs on the operator's VPN-connected machine, holds a Bearer token acquired out-of-band (`meho login --print-token`), and forwards Streamable-HTTP calls to the internal `/mcp` with the `Authorization` header injected. This is already how the dogfood adopters connect (`.mcp.json`: VPN + loopback PKCE / internal CA) — the internal path, not the public one, is the real consumption mode.

## Problem

Three unverified assumptions currently carry the fresh-user plan:

1. **The shim flow completes at all from Claude Desktop** — a local `mcp-remote` stdio server (configured in `claude_desktop_config.json`) forwarding to an internal `/mcp` with an injected Bearer token has never been observed end-to-end from Claude Desktop specifically. (Note a client-naming drift to resolve while here: docs/values-examples say `meho-mcp-client`; both adopters registered `meho-mcp`.)
[...]

Issue #3128, opened 2026-08-25
Initiative: Claude extension distribution + first-reflex enablement — .mcpb bundle, Claude Code plugin, and backplane reflex levers


damir-topic:
Parent goal: #2661

## Scope

Package the proven Claude-client onramp as installable artifacts, and make the backplane's coordination levers (broadcast, knowledge, memory) the agent's *default reflex* instead of an on-request behavior. Two distribution artifacts — a one-click Claude Desktop `.mcpb` bundle and an in-repo Claude Code plugin + marketplace — plus three reflex levers that reach every client surface: when-to-use curation of the meta-tool descriptions, a dispatch-time reflex advisory / opt-in announce gate on the server, and reflex-adoption KPIs computed from `audit_log`.

**Area**: `.claude-plugin/` + `clients/` (new), [`backend/src/meho_backplane/mcp/tools/`](../blob/main/backend/src/meho_backplane/mcp/tools/), [`backend/src/meho_backplane/operations/dispatcher.py`](../blob/main/backend/src/meho_backplane/operations/dispatcher.py), [`docs/examples/consumer-onboarding/`](../blob/main/docs/examples/consumer-onboarding/), [`.github/workflows/cli-release.yml`](../blob/main/.github/workflows/cli-release.yml)

## Motivation

The connection path is proven — Claude Desktop reaches an internal backplane through the local `mcp-remote` stdio shim (#2666), and the surface is Desktop-conformant (#2745, #2774). But installing it today means hand-editing `claude_desktop_config.json` or copy-merging the Layer-2 template, which sits directly on Goal #2661's fresh-user story. And once connected, agents use the coordination levers mostly when pushed: tool descriptions, the tenant preamble, and the repo template are all advisory; nothing enforces read-before-start or announce-before-mutate; and nothing measures whether the discipline actually happens.

Direction constraint (verified 2026-08-25 against current Anthropic docs): third-party products may not offer claude.ai subscription login (Claude Agent SDK terms), and the remote claude.ai Custom Connector requires a publicly reachable backplane — MEHO is internal-only by design ([`docs/cross-repo/mcp-client-setup.md`](../blob/main/docs/cross-repo/mcp-client-setup.md)). So Claude is the host (each user brings their own Claude account) and MEHO is the extension, reached over the VPN with the operator's own MEHO identity.

## Definition of done

- [ ] A fresh user installs the Desktop onramp by opening a `.mcpb` file — no JSON editing; the bundle is a published release asset.
[...]

Issue #2661, opened 2026-07-22
Goal: Road to v1.0.0 — fresh-user install, Claude Desktop, published docs, enforced stability promise


damir-topic:
## Goal

**A fresh operator can install MEHO from published documentation, connect Claude Desktop, and do real work — under an enforced, honestly-scoped stability promise. That is what v1.0.0 means; this Goal carries the path to it.**

Feature construction is complete (every feature Goal is closed) and the tracker holds no 1.0 planning artifact — searches for `v1.0`/`GA`/`stability promise` return nothing. A four-repo grounding review (2026-07-21, evoila/meho + meho-internal + both dogfood repos + the MFC adopter; adversarially fact-checked) found the gap is no longer features, it is productization:

- **What provably works** is the read/governance/knowledge plane (audit, memory/knowledge, targets, typed reads, `net.*`) — daily drivers in the field.
- **What is landed-but-broken in the field** is everything autonomous: hosted-agent runs 400 at model-init on v0.25.0 (#2644), Sensors are dead on per-operator-WIF GSM deploys (#2642), and the scheduler's Vault token dies after days (evoila-bosnia/claude-rdc-hetzner-dc#2003, 3rd recurrence — #2652 fixed only the diagnosis).
- **The dogfood signal is structurally murky**: the consumer repos carry 40–65 wrapper scripts and 39–52 skills that silently absorb every gap (the one near-fresh customer, ECP, stalled at the auth walls and adopted ~nothing).
- **The Claude Desktop Custom Connector path has never been executed anywhere** — every deploy is VPN-internal, and the path requires a publicly-reachable backplane with real TLS.
- **No published documentation exists** (no docs site, no publishing workflow) and **no surface marks feature maturity** — tracker state is the de-facto signal, and it demonstrably overstates reality.

## Milestone plan

- **v0.26 — unbreak the agents**: G0.37 (#2656) complete; durable machine credential identity on both backends (#2642 + the Vault token-lifecycle task below); first empirical Claude Desktop connection (smoke-test task below).
- **v0.27 — fresh-user install + clean-room eval round 1**: cold-install cleanup, realm-bootstrap automation, MCP-native target/secret registration, docs site v1 live, maturity labels shipping, then run the clean-room evaluation initiative below.
[...]

PR #3385, opened 2026-09-05
fix(clients): run .mcpb mcp-remote in-process, not a spawned execPath child (F6, #3341)

### LINKED ISSUE #3341: Task: fix(clients): .mcpb launcher must spawn a real node, not process.execPath, under Claude Desktop (F6, blocks #3143 Part B)
Parent initiative: #3128

## Goal

Make the Claude Desktop `.mcpb` launcher load the MEHO tool surface on
current Claude Desktop builds by spawning the vendored `mcp-remote`
through a **real `node`** runtime instead of `process.execPath`. This
clears field-test finding **F6** (#3143) and unblocks Part B of that
field test — the last open line on Initiative #3128's "fresh user
installs the Desktop onramp by opening a `.mcpb` file" definition of
done.

## Why

#3144 (merged as PR #3146) removed the `spawn npx ENOENT` failure (F1)
by vendoring `mcp-remote@0.1.38` and spawning it via the bundle's own
runtime: `spawn(process.execPath, [mcpRemoteEntry, …], { env: { …,
ELECTRON_RUN_AS_NODE: "1" } })`. That fix assumed `ELECTRON_RUN_AS_NODE=1`
makes the child "run as plain Node whether `execPath` is a standalone
`node` or Claude Desktop's Electron helper."

Field-test finding **F6** (#3143, @ilhan-karic, reproduced 2026-09-01 on
Claude Desktop **1.40609.0**, macOS 14.6.1) shows that assumption is
**false on current Desktop builds**, so #3144's fix is necessary but not
sufficient — the Desktop onramp still cannot load tools:

- Under Claude Desktop the extension's `process.execPath` is the Electron
  **`Claude Helper (Plugin)`** binary, not a standalone `node`. On build
  1.40609.0 that binary **cannot be run as Node even with
  `ELECTRON_RUN_AS_NODE=1`** — spawning it standalone FATALs:
  `[FATAL:electron/shell/app/electron_main_delegate_mac.mm:66] Unable to
  find helper app` (repeated), plus `Network service crashed` /
  `GPU process exited`.
- Reproduced in isolation, no `mcp-remote` involved:
  `ELECTRON_RUN_AS_NODE=1 "…/Claude Helper (Plugin)" -e
  'console.log(process.version)'` → the same helper-app FATAL.
- Net effect matches the observed logs: `index.mjs`'s own stdio to
  Desktop comes up ("Server started and connected successfully"), but the
  `mcp-remote` child never runs, so the MCP `initialize` request is
  forwarded nowhere → Desktop cancels after **exactly 60 s**
  (`notifications/cancelled`) → `tools/list` never runs → **no tools
  ever appear**.
- **Corrects the earlier root cause on record.** @ilhan-karic's first F6
  comment (2026-08-27) attributed the crash to "the vendored mcp-remote
  spawns a descendant via `process.execPath`." That was **wrong** —
  `mcp-remote` coordinates via lockfiles and contains no
  `spawn`/`fork`/`process.execPath`. The `Era probe verdict:
  legacy/modern` and `Era probe: sibling spawn failed` strings are
  **Claude Desktop's own** MCP transport-negotiation code (they live in
  `Claude.app/Contents/Resources/app.asar`), not the bundle's. The true
  cause is the launcher's own `spawn(process.execPath, …)` in
  `server/index.mjs`.
- **Why Part A (Claude Code) is unaffected:** there `mcp-remote` runs
  under a real `node`, so `process.execPath` is `node` and the spawn is a
  genuine Node process. Only Claude Desktop (execPath = Electron helper)
  hits F6.

This gates the full chain: F6 blocks **#3143 Part B** → Initiative
**#3128** (Claude extension distribution — its Desktop-onramp DoD line)
and sibling Initiative **#2665** (clean-room evaluation round 1 —
fresh-user Claude Desktop program) → Goal **#2661** (Road to v1.0.0).

## Affected files

- [`clients/claude-desktop-mcpb/server/index.mjs`](../blob/main/clients/claude-desktop-mcpb/server/index.mjs)
  — L91–95 (the `ELECTRON_RUN_AS_NODE=1` assumption) and L116–127 (the
  `spawn(process.execPath, [mcpRemoteEntry, …])` call). This is the line
  the fix must change: the child must launch under a real `node`, not
  `process.execPath`.
- [`clients/claude-desktop-mcpb/build.sh`](../blob/main/clients/claude-desktop-mcpb/build.sh)
  — L41–58 (staging + `npm ci --omit=dev` + `mcpb pack`). Option A adds
  per-platform `node` vendoring into the staged bundle.
- [`clients/claude-desktop-mcpb/manifest.json`](../blob/main/clients/claude-desktop-mcpb/manifest.json)
  — L12–24 (`server.mcp_config`) and L53–55 (`compatibility.platforms`:
  `["darwin","win32"]`). Option A turns a single bundle into a
  per-platform matrix.
- [`clients/claude-desktop-mcpb/package.json`](../blob/main/clients/claude-desktop-mcpb/package.json)
  / `package-lock.json` — the vendored dependency set.
- [`clients/claude-desktop-mcpb/test/index.test.mjs`](../blob/main/clients/claude-desktop-mcpb/test/index.test.mjs)
  — the behavioral suite currently asserts the child argv is
  `[process.execPath, <entry>, …]` with `ELECTRON_RUN_AS_NODE=1`; it must
  be updated to assert a real-node spawn and add a regression assertion
  that the launcher never spawns `process.execPath` when execPath is an
  Electron helper.
- [`clients/claude-desktop-mcpb/README.md`](../blob/main/clients/claude-desktop-mcpb/README.md)
  — "How the launcher works" (L61–109) currently documents the exact
  false assumption ("runs as plain Node whether `execPath` is a
  standalone `node` or Claude Desktop's Electron helper"); update it.
- [`.github/workflows/mcpb-bundle.yml`](../blob/main/.github/workflows/mcpb-bundle.yml)
  (PR dry-run, currently single `runs-on: meho-runners-ci`) and
  [`.github/workflows/cli-release.yml`](../blob/main/.github/workflows/cli-release.yml)
  (L281–292, release-asset build) — Option A requires a per-platform
  build matrix and per-platform release assets.

## Current state

`server/index.mjs` L116–127 spawns the vendored `mcp-remote` with
`spawn(process.execPath, [mcpRemoteEntry, url.href, "--static-oauth-client-info",
…], { stdio: "inherit", env })` where `env.ELECTRON_RUN_AS_NODE = "1"`
(L95). Under Claude Desktop 1.40609.0 `process.execPath` is the
`Claude Helper (Plugin)` Electron binary, which FATALs when spawned as
Node — so the `mcp-remote` child never starts and the Desktop `initialize`
handshake stalls to a 60 s cancel. The launcher has never successfully
loaded tools inside Claude Desktop; F1 (#3144) removed the `npx` failure
but exposed this deeper one.

## Desired state

- The launcher spawns `mcp-remote` under a **real `node` binary**, so
  the child and all its descendants run as plain Node under Claude
  Desktop — no reliance on `ELECTRON_RUN_AS_NODE` making the Electron
  helper behave as Node.
- Internal-CA deploys keep working: `NODE_EXTRA_CA_CERTS` must be honored
  on whichever path ships. It is delivered before the launcher starts via
  the manifest `env` block — the same mechanism that already reaches the
  launcher as `MEHO_CA_CERT` — which is what makes the in-process spike in
  the approach section viable.
- All existing launcher guards are preserved exactly: https-URL
  validation, `NODE_EXTRA_CA_CERTS` only on a non-empty CA path,
  `MEHO_CA_CERT` / `MEHO_MCP_CLIENT_ID` / `MEHO_MCP_SCOPES` scrubbed from
  the child env, and the static-OAuth invocation contract
  (`--static-oauth-client-info` with the `meho-mcp` / `client_id` seam,
  `--static-oauth-client-metadata` scope).
- Claude Desktop loads the MEHO tool surface end-to-end: OAuth completes,
  `initialize` completes, `tools/list` returns, and one read operation
  round-trips (the #2666 bar).

### Approach — decided 2026-09-03 (Damir)

**1. Spike the in-process option first (timeboxed ~1 day).** Before any
vendoring, prove or kill the lightest path: import `mcp-remote`'s CLI
entry point *inside the launcher process* — set `process.argv` to the
`mcp-remote` invocation, then `import` it. Nothing is spawned, so there is
no `process.execPath` child and F6 cannot recur. Internal-CA trust is
supplied via `NODE_EXTRA_CA_CERTS` in the manifest `env` block, which
Claude Desktop applies to the launcher's environment *before the launcher
process starts* — the same mechanism that already delivers `MEHO_CA_CERT`
to the launcher today — so the "CA cannot be injected after Node startup"
objection does not apply to an in-process import whose env is already set
before Node boots. The spike must prove:
  - **(a)** clean in-process start/stop and full stdio ownership (the
    launcher's stdio to Desktop becomes `mcp-remote`'s stdio) with **both**
    the currently pinned `mcp-remote` version **and** the pending version
    bump; and
  - **(b)** CA delivery works when the optional `user_config` CA field is
    left empty (the default fresh-user install).

**2. Fallback — Option A (vendor a real `node`), per-platform assets.** If
the spike fails on either (a) or (b), fall back to vendoring a real `node`
per platform and spawning `mcp-remote` under it, shipped in per-platform
asset form (a per-platform CI matrix + release assets). The size and
matrix costs below apply only to this fallback.

**3. Option B (resolve a system `node`) is rejected as a default.** It
breaks the README's "no system Node required" promise and the v1.0
fresh-user goal, so it is not the default path; it may be revisited only
by an explicit product decision.

**4. The manifest-runtime idea is not viable.** MCPB
`compatibility.runtimes.node` is a *version constraint only* — it declares
which Node the host must satisfy, it does not make the host ship or select
a runtime — so it cannot supply the runtime this launcher needs.

| Option | Upside | Downside |
|---|---|---|
| **In-process import** (spike first) | Least code; no child, so F6 cannot recur; no host `node`, no size cost | Must prove stdio ownership + CA delivery across the current and pending `mcp-remote` versions (the spike) |
| **A. Vendor a real `node`** per platform, spawn that (fallback) | Robust; works with internal-CA; no host dependency | +~60 MB/platform, per-platform CI matrix + release assets, larger download |
| **B. Resolve a system `node`** | Small, light change | **Rejected as default** — breaks the "no system Node required" promise + the v1.0 fresh-user goal; product decision only |

The spike's outcome is recorded on this issue before any implementation
starts (see acceptance criteria).

## Acceptance criteria

- [ ] Spike outcome recorded on this issue before implementation starts.
- [ ] `server/index.mjs` no longer spawns `process.execPath` for the
      `mcp-remote` child; `grep -n "process.execPath" clients/claude-desktop-mcpb/server/index.mjs`
      returns no spawn-relevant match (execPath may still be read for
      diagnostics only).
- [ ] Behavioral suite (`node --test clients/claude-desktop-mcpb/test/index.test.mjs`)
      passes and includes a regression case proving that when `execPath`
      points at a non-Node (Electron-helper-shaped) binary the launcher
      still starts `mcp-remote` under a real `node` — i.e. the F6 failure
      mode cannot recur.
- [ ] Existing guards still asserted green: https-URL validation,
      `NODE_EXTRA_CA_CERTS` set only on a non-empty CA path,
      `MEHO_CA_CERT` / `MEHO_MCP_CLIENT_ID` / `MEHO_MCP_SCOPES` scrubbed
      from the child env, static-OAuth argv unchanged.
- [ ] `./build.sh <version>` packs a bundle that boots `mcp-remote` under
      a real `node` on Claude Desktop; the chosen option (A or B) is
      documented in the PR, and for Option A the resulting per-platform
      bundle size(s) are stated.
- [ ] `mcpb-bundle.yml` PR dry-run lane green (updated to the chosen
      approach; per-platform matrix if Option A); README "How the
      launcher works" no longer claims `ELECTRON_RUN_AS_NODE=1` makes the
      Electron helper run as Node.
- [ ] **#3143 Part B re-test item in the PR:** on Claude Desktop
      **≥ 1.40609.0** (macOS), install the freshly built `.mcpb`,
      complete OAuth, and confirm the MEHO tools appear and one read
      operation round-trips — i.e. no `Unable to find helper app` FATAL,
      no 60 s `initialize` cancel, `tools/list` returns. Post the result
      on #3143.

## Non-goals

- The Claude Code plugin wrapper (`clients/claude-code-plugin/`) — Part A
  runs under a real `node` and is unaffected by F6; unchanged.
- Signing / notarizing the bundle (the "not verified by Anthropic"
  install warning is a separate field-test friction follow-up).
- Upgrading `mcp-remote` off the pinned `0.1.38` (upstream latest is
  0.8.3) — out of scope; pin stays unless the chosen approach requires
  otherwise.
- Any change to the OAuth realm / redirect-URI configuration (that was a
  separate, already-resolved #3143 blocker).
- The reflex-hook, digest, and tool-surface findings from #3143 (F2–F5)
  — each already has its own follow-up.

## Verification steps

```bash
cd clients/claude-desktop-mcpb
node --test test/index.test.mjs                       # launcher spawns real node; F6 regression case green
grep -n "process.execPath" server/index.mjs           # no spawn-relevant match
./build.sh 0.0.0-dev /tmp/out                          # bundle packs under the chosen option
# Then the manual #3143 Part B re-test on Claude Desktop >= 1.40609.0 (macOS).
```

## References

- Field report + reproduced root cause (credit: @ilhan-karic):
  #3143 comment of 2026-09-01 ("F6 correction + reproduced root cause")
  and the earlier 2026-08-27 F6 note it corrects.
- Predecessor fix this builds on: #3144 (closed) / PR #3146 — vendored
  `mcp-remote`, introduced the `process.execPath` spawn F6 breaks.
- Parent initiative: #3128 (Desktop-onramp DoD line); sibling initiative
  #2665; goal #2661.
- Electron `ELECTRON_RUN_AS_NODE` semantics — verify fresh at
  implementation time against current Electron docs
  (https://www.electronjs.org/docs/latest/api/environment-variables#electron_run_as_node);
  F6 is the evidence it is not honored by a standalone Electron helper on
  Claude Desktop 1.40609.0.
- MCPB packing / MANIFEST (research fresh for per-platform vendoring):
  https://github.com/modelcontextprotocol/mcpb
- [DevOps best practices — supply chain & pinning](../blob/main/.claude/skills/implement-issue/devops_best_practices.md#supply-chain)
  (per-platform vendored runtime, integrity-checked; keep the pinned,
  reproducible build discipline #3146 established).

damir-topic:
## Summary

- Fixes field-test **F6** (#3143 Part B): the Claude Desktop `.mcpb`
  launcher never loaded the MEHO tool surface because it spawned
  `mcp-remote` via `spawn(process.execPath, …)` and, on Claude Desktop
  ≥ 1.40609.0, `process.execPath` is the `Claude Helper (Plugin)` Electron
  binary that FATALs when spawned as Node (`Unable to find helper app`) even
  with `ELECTRON_RUN_AS_NODE=1` — so the child never started, `initialize`
  stalled to a 60 s cancel, and no tools appeared.
- **Runs `mcp-remote` in-process** instead: the launcher sets `process.argv`
  to the `mcp-remote` invocation and **imports** the vendored CLI entry
  (which reads `process.argv` at module top level and has no main-guard),
  spawning nothing. `execPath` is never used to launch a process, so the F6
  spawn cannot recur, and `mcp-remote` runs under the exact Node runtime
  Claude Desktop already started, owning the launcher's stdio.
- **Internal-CA trust** moves to `NODE_EXTRA_CA_CERTS` delivered by the
  manifest `env` block (applied before Node boots). This is the only route
  that works in-process — `NODE_EXTRA_CA_CERTS` is boot-only and there is no
  child env to export it to — and it is the same pre-boot delivery route
  `MEHO_CA_CERT` used.
- **Chosen option: in-process import** (the spike-first path). Option A
  (vendor a real `node` per platform) was **not** taken — no per-platform CI
  matrix, no release-asset split, no size cost. The bundle stays a single
  **1.4 MB** artifact.

## Phase-1 spike (recorded on #3341 before implementation)

A timeboxed in-process spike proved both required criteria against **both**
the pinned `mcp-remote@0.1.38` and the pending `0.8.3` (node v26.5.1,
macOS, minimal GUI-shaped `PATH`). Evidence:
[#3341 spike comment](https://github.com/evoila/meho/issues/3341#issuecomment-5554555959).

- **(a) start/stop + stdio ownership** — importing `dist/proxy.js` runs
  `mcp-remote`'s main against the launcher's argv; **zero child processes**
  (`pgrep -P` → `[]`); stdout stays the clean MCP channel while logs go to
  the launcher's stderr; graceful `process.exit(0)` on stdin-EOF / SIGINT
  via `mcp-remote`'s own handlers. Neither version self-spawns a node child
  (0.1.38 has no `spawn` in `dist`; 0.8.3's only `spawn` is the OS
  browser-opener — not an `execPath` self-exec).
- **(b) CA delivery, empty field** — `NODE_EXTRA_CA_CERTS` is boot-only
  (runtime mutation has no effect); empty `""` (the fresh-install
  substitution) is a silent no-op with the default 120-cert store intact; a
  real path is loaded and trusted (store 120→121, subject verified).

## Design note — one deliberate deviation from the issue's literal criteria

The issue was framed around a spawn-based fix with a child env; two of its
"preserved guard" lines describe launcher **code** that manages
`NODE_EXTRA_CA_CERTS`/`MEHO_CA_CERT`. In-process that code is structurally
impossible: `NODE_EXTRA_CA_CERTS` is consumed at Node boot, before launcher
code runs. So CA delivery moves entirely to the manifest `env` block, and
the guard's *purpose* — an unset optional CA must not corrupt TLS trust —
is now satisfied by Node's own empty-value handling (spike-proven no-op)
rather than a code branch. The `client_id` / `scopes` seam scrubbing is
preserved (now `delete process.env.*` before the import, since `mcp-remote`
inherits `process.env`).

## Test plan

- [x] `node --test clients/claude-desktop-mcpb/test/index.test.mjs` — 12
      passed. Includes the **F6 regression**: `mcp-remote` runs in the
      launcher's own process (`ppid` = test runner, never a spawned child)
      + a source guard that the launcher imports no process-spawning module.
- [x] `grep -n "process.execPath" clients/claude-desktop-mcpb/server/index.mjs`
      — **zero matches** (no spawn-relevant reference; execPath is not used
      at all).
- [x] Existing guards asserted green — https-only endpoint, static-OAuth
      argv unchanged, `client_id` / `scopes` override + fallback + scrub,
      empty/non-empty `NODE_EXTRA_CA_CERTS` passed through untouched.
- [x] `./build.sh 0.0.0-dev <out>` — packs a schema-valid MANIFEST-0.3
      bundle (1.4 MB); the `env` change validates.
- [x] Real vendored `mcp-remote@0.1.38` imported by the shipped `index.mjs`
      against a local endpoint — zero children, clean MCP stdout,
      `mcp-remote`'s own logs on stderr.
- [x] CI wording guard (`no "custom connector"`) clean; `mcpb-bundle.yml`
      dry-run lane updated to the in-process contract (single runner — no
      per-platform matrix needed).

## Needs a human re-test (#3143 Part B, Claude Desktop ≥ 1.40609.0, macOS)

Everything provable with local node is green; these require the real GUI +
a live backplane and are **not** claimed proven here — please run on install
of the freshly built `.mcpb` and post the result on #3143:

1. OAuth completes, `initialize` completes (no 60 s cancel), `tools/list`
   returns, and one read op round-trips (the #2666 bar) — no
   `Unable to find helper app` FATAL.
2. Claude Desktop's exact substitution of an **untouched optional `file`**
   config into `NODE_EXTRA_CA_CERTS`: `""` is fully silent; a whitespace or
   literal-placeholder substitution would print one **non-fatal** stderr
   `Warning: Ignoring extra certs …` (TLS otherwise unaffected). Confirm
   which the host does.

## Out of scope

- `mcp-remote` pin stays `0.1.38` (spike validated 0.8.3 forward-compat; no
  bump here).
- Bundle signing / notarizing; the F2–F5 findings from #3143 — each has its
  own follow-up.

Closes #3341


damir-topic:
> satisfied by this comment.

**Anchor task:** #3341 — .mcpb launcher must not spawn `process.execPath` under Claude Desktop (F6, blocks #3143 Part B)
**Surface area:** devops (Node launcher + MCPB bundle + CI)
**Verdict:** APPROVE
**Head SHA reviewed:** `bc153d5c`

### Acceptance criteria
- [x] Spike outcome recorded on the issue **before** implementation — issue comment 5554555959 (in-process import proven against `mcp-remote@0.1.38` and `0.8.3`; zero child spawn; stdio ownership; graceful stdin-EOF shutdown; `NODE_EXTRA_CA_CERTS` boot-only, empty = no-op).
- [x] `server/index.mjs` no longer spawns `process.execPath` — `grep` returns no spawn-relevant match; `execPath` is not read at all. Verified.
- [x] Behavioral suite passes with an F6 regression case — `node --test` → **12/12 pass** locally. Regression is structural: `ppid == test-runner` proves in-process, plus a comment-stripped source guard forbids `child_process` / `spawn(`.
- [x] Existing guards asserted green — https-only endpoint, `client_id` / `scopes` override + fallback + scrub, CA value passed through untouched. Verified in the suite.
- [x] `build.sh` packs under the chosen option (in-process; A/B not taken, documented) — CI "Pack .mcpb bundle (dry-run)" lane is SUCCESS; local `npm ci --omit=dev` vendors the full tree.
- [x] `mcpb-bundle.yml` PR dry-run updated to the in-process contract; README "How the launcher works" no longer claims `ELECTRON_RUN_AS_NODE=1` makes the Electron helper run as Node.
- [ ] **Human re-test (#3143 Part B, Claude Desktop ≥ 1.40609.0, macOS)** — correctly **not** claimed proven here; deferred with a clear re-test recipe. Requires the real GUI + live backplane.

### Project-board alignment
|---|---|---|---|---|---|
| #3341 | Open | Closes | ✅ | ✅ (machine-provable parts; human Part-B re-test deferred) | close on merge |

No delivered-but-unlinked or linked-but-undelivered issues. Hierarchy coherent (#3341 → #3128 → #2661).

### Predicted vs. actual diff
- ✅ `server/index.mjs` — spawn → in-process import; `ELECTRON_RUN_AS_NODE` and in-code CA guard dropped; seams scrubbed on `process.env`.
- ✅ `test/index.test.mjs` — in-process contract, F6 ppid regression, source-guard, CA pass-through, negative `ELECTRON_RUN_AS_NODE`.
- ✅ `manifest.json` — CA delivery moved `MEHO_CA_CERT` → `NODE_EXTRA_CA_CERTS` in the `env` block.
- ✅ `README.md`, `build.sh`, `mcpb-bundle.yml` — narrative/comments updated to the in-process contract; old assumption documented as the bug fixed.
- ✅ `CHANGELOG.md` — one bullet under `[Unreleased]` → `Fixed`, links #3341, accurate.
- No unexpected/undisclosed files; nothing predicted-but-missing.

### Framework research (verified against real sources, not recall)
- `mcp-remote@0.1.38` `dist/proxy.js`: runs `parseCommandLineArgs(process.argv.slice(2), …)` at module top level with **no** `import.meta`/main-guard → importing runs `main`. Confirmed against the published tarball.
- `--static-oauth-client-info` / `--static-oauth-client-metadata` parsed via `args.indexOf(...)` reading the next arg as a JSON string — exactly the argv shape the launcher builds. Confirmed.
- Zero `spawn`/`fork`/`execPath`/`child_process` in 0.1.38 `dist/` → confirms the corrected root cause (the spawn was the launcher's, never `mcp-remote`'s).
- `setupSignalHandlers` registers `SIGINT` + stdin-`end` → `cleanup(); process.exit(0)`; `process.stdin.resume()`. Confirmed graceful shutdown path.
- No `exports` field in the package → `require.resolve("mcp-remote/package.json")` resolves. Confirmed.
- `pathToFileURL(entry).href` is the correct idiom for dynamic-import of an absolute path (Windows-safe; valid on POSIX).
- Node `NODE_EXTRA_CA_CERTS` on v26.5.1: `""` → silent no-op, default store 120 intact; whitespace / literal-placeholder / bad path → **non-fatal** `Warning: Ignoring extra certs …`, default store still 120. Confirms the dropped in-code guard's worst case is a cosmetic stderr warning, never broken TLS.

### Integration verification
- [x] `node --test clients/claude-desktop-mcpb/test/index.test.mjs` — **12 passed**.
- [x] **Real vendored `mcp-remote@0.1.38` imported in-process** by the shipped `index.mjs` against a blackhole endpoint — **zero child processes** (`pgrep -P` empty), **stdout clean (0 bytes)**, `mcp-remote`'s own logs (`Using static OAuth client information from string`, `Discovering OAuth server configuration…`) on stderr with the launcher's own PID → argv drove it and it owns stdio.
- [x] Import-failure path exercised — a missing transitive dep surfaced as the launcher's exact `failed to launch mcp-remote: …` stderr + exit 1.
- [x] `NODE_EXTRA_CA_CERTS` empty/whitespace/bad-path behavior executed on node v26.5.1 (see framework research).
- [x] Wording guard (`no "custom connector"`) — clean; `manifest.json` valid JSON; CI dry-run pack SUCCESS.
- n/a Python lanes (ruff/mypy/pytest) — PR touches zero `meho_app/` Python; the in-progress Python shards in CI are unrelated required lanes.

### Findings
None blocking. No blockers, no majors, no inline-eligible minors.

- `test/index.test.mjs` line 73 — the stub still emits a `mehoCaCert: process.env.MEHO_CA_CERT` diagnostic field that nothing asserts any more and that the launcher no longer reads. Harmless dead diagnostic; could be dropped in a future touch alongside the other renamed CA assertions. Not merge-blocking.

### Docs
- [x] `clients/claude-desktop-mcpb/README.md` — updated to the in-process contract; the old `ELECTRON_RUN_AS_NODE` assumption is now framed as the bug fixed. `docs/cross-repo/mcp-client-setup.md` needs no change (it documents the bundle at the download-and-open level, not the launcher's internal mechanism).

### What worked well
- Spike-first discipline: the load-bearing runtime facts (in-process import runs `main`, boot-only `NODE_EXTRA_CA_CERTS`, empty = no-op) were proven and recorded on the issue before a line of implementation, and every one reproduces against the real published `mcp-remote` here.
- The one deliberate deviation from the issue's literal "guard in code" criterion (CA guard → Node's empty-value no-op) is called out honestly in the PR body, code comments, README, and rests on verified Node behavior whose worst case is non-fatal.
- The F6 regression is structural (`ppid` + source guard), not a brittle string check, so it actually prevents the failure mode from recurring.

PR #3382, opened 2026-09-05
ci(secret-scan): warm trufflehog image with bounded retry pull (#3310)

damir-topic:
## Summary

- `TruffleHog Secret Scan` — a **required** merge-queue check on `meho-runners-ci` — runs the scan through `docker run ghcr.io/trufflesecurity/trufflehog:latest`. That implicit image pull intermittently failed with a **TLS-handshake timeout (exit 125)** when the shared runner egress saturated under concurrent CI load (the same reliability class as the `codeload.github.com` 429s), ejecting PRs from the merge queue on a transient network fault rather than a real finding (evidence: runs `33635647838` / `33635809093` on #3306 / #3303).
- This PR warms the pinned TruffleHog image with a **bounded exponential-backoff `docker pull`** (5 attempts, 5→60s) before the scan step. `docker run`'s default `--pull missing` policy then makes the scan a **local cache hit**, so a stalled handshake becomes a short retry instead of a failed required check.
- **The security gate is not weakened.** The scan step is byte-for-byte unchanged — same pinned action, same `--results=verified,unknown --fail`, same `base..head` range on every push, PR and `merge_group` ref (the #3308 fix). The new step only isolates the flaky part (the network pull) from the deterministic part (the scan) and retries the pull.

## Scope — why only `secret-scan.yml`

The issue's triage listed several jobs as *potentially* exposed; the surgical, evidence-backed scope is `secret-scan.yml` alone:

- **`secret-scan.yml`** — the sole observed failure and the only ghcr.io-image-pull-at-runtime site on an affected runner (`meho-runners-ci`) that is a required merge-queue check. **Fixed here.**
- **`chart.yml`** — its `docker pull`s (`valkey`, `bitnamilegacy/postgresql`) live in the `helm-test` job, which runs on GitHub-hosted **`ubuntu-latest`**, not `meho-runners-ci`; it is not subject to this shared-egress saturation, and its backplane image is built locally (`load: true`) + `kind load`ed, never pulled from ghcr.io. Left untouched (out of the issue's runner scope).
- **`pr-smoke.yml`** — its `ghcr.io/evoila/meho:pr-<n>-<sha>` pull happens **in-cluster via kubelet** (`helm upgrade --install` to rke2-infra), which already retries with `ImagePullBackOff` inside the rollout window; there is no shell `docker pull` step to wrap. Left untouched.
- No ARC infrastructure or runner-count changes (out of this repo's remit).

## Test plan

[...]

Issue #3310, opened 2026-09-02
ci: ghcr.io image pulls flake with TLS handshake timeouts under concurrent CI load (trufflehog pull, backplane image jobs)


damir-topic:
## Summary

CI jobs that pull a container image from `ghcr.io` at runtime
intermittently fail with a **TLS handshake timeout** during the pull,
exiting **125** before the job's real work runs. This is
non-deterministic and load-correlated: it strikes under concurrent CI
load and clears on re-run. Because the affected jobs include required
checks (e.g. `TruffleHog Secret Scan`), a flaked pull can eject a PR
from the merge queue or block a release PR until a manual re-run.

This is **not** a workflow-logic bug — it is a runner **egress
reliability** problem, the same shared-egress-saturation class already
seen with `codeload.github.com` 429s. The action-archive cache
(`ACTIONS_RUNNER_ACTION_ARCHIVE_CACHE`) fixed the codeload path, but
`ghcr.io` image pulls are **uncached** and still hit the open egress.

## Evidence

Two failing runs, both on branches based on current `main`
(`733466279f`), same error class, same exit code, different endpoints —
a network-layer flake, not a repo-specific one:

- **#3306** (head `274a5d0d`), run `33635647838` —
  `TruffleHog Secret Scan`:
  ```
  docker: Error response from daemon: failed to resolve reference
  "ghcr.io/trufflesecurity/trufflehog:latest": failed to do request:
  Head "https://ghcr.io/v2/trufflesecurity/trufflehog/manifests/latest":
  net/http: TLS handshake timeout
  ##[error]Process completed with exit code 125.
  ```
- **#3303** (head `78a2d05b`), run `33635809093` —
  `TruffleHog Secret Scan`:
  ```
  docker: Error response from daemon: failed to copy: httpReadSeeker:
  failed open: failed to authorize: failed to fetch anonymous token:
  Get "https://ghcr.io/token?scope=repository%3Atrufflesecurity%2Ftrufflehog%3Apull&service=ghcr.io":
  net/http: TLS handshake timeout
  ##[error]Process completed with exit code 125.
  ```

Non-determinism confirmation: in the same window, PRs **#3305** and
**#3309** passed `TruffleHog Secret Scan` on the same `main`-with-fix —
so the failure is transient, not a code regression. (Context: this
surfaced while investigating a suspected `secret-scan.yml` regression;
the workflow logic was cleared — #3307 / #3308 — and the true cause is
this pull flake.)

[...]

PR #3404, opened 2026-09-05
test(ingest): de-clock the event-loop responsiveness pin — measure the DB-commit window

### LINKED ISSUE #3399: Task: ci: event-loop responsiveness test is load-sensitive — stalls >0.5s on loaded runners eject merge-queue builds
Parent goal: #221

## Goal

Make `test_nondryrun_large_spec_ingest_keeps_event_loop_responsive` load-robust so it stops ejecting unrelated PRs from the merge queue, while still enforcing the responsiveness guarantee it exists to protect (the non-dry-run large-spec register/commit pass yields per-op rather than blocking the event loop for the whole batch).

## Why

The test asserts an **absolute wall-clock ceiling** — no single event-loop heartbeat gap during the ingest may exceed `0.5 s` (`_MAX_GAP_CEILING_S`, `backend/tests/test_ingest_large_spec_event_loop.py:296-300`). That ceiling is measured on shared self-hosted ARC runners (`meho-runners-ci`), so it is load-sensitive by construction: under a saturated pool, ambient scheduler jitter alone crosses the ceiling while the code under test is working perfectly.

It has now false-failed twice, both times while the ingest offload was correct:

- **PR #3384** — unit shard 1/3, run **33990228779**: `event loop stalled for 1.203s during the ingest (ceiling 0.5s)`; passed on re-run.
- **PR #3382** (a workflow-only change) — merge_group build, unit shard 3/3, run **33991177823**: `1.003s` stall, **4914 other tests passed**, and the false failure **ejected #3382 from the merge queue**.

This is the same failure mode #2706 already fixed for the *sibling* test in this exact file (`test_blocking_parse_stays_off_the_event_loop`): it took `max(gaps)` over the run and compared against an absolute magnitude, which "conflates two unrelated things: the loop being blocked, and the runner simply being busy." That PR recorded ambient jitter of **1.522s** on a saturated pool. Crucially, that observed jitter (~1.5s) **overlaps** the ~1.7s freeze a non-yielding register batch would produce, so **no** absolute-or-ratio magnitude threshold can separate load from regression on a shared runner. #2706's remedy was to stop measuring the clock and measure the loop directly (a presence-vs-absence tick count). This test never got the same treatment — it still carries both the load-robust tick-count assertion **and** the load-fragile magnitude ceiling.

Precedent for the class: #1434 (closed) de-clocked a topology perf gate that "fails ~2.5x over 100ms on CI runner" for the same reason.

## Current state

`backend/tests/test_ingest_large_spec_event_loop.py` drives the real parse + register (DB-commit) passes for a 1275-op spec against sqlite while a heartbeat coroutine measures event-loop gaps. The test already carries the load-robust structural signal — the loop ticked ≥ 50 times *during* the ingest — but **also** asserts the load-fragile absolute ceiling:

```python
# lines 292-300
assert len(gaps) >= _MIN_TICKS, (          # structural: loop ran throughout (load-robust)
    f"heartbeat only ticked {len(gaps)} times during the ingest; "
    "the event loop was starved (a non-yielding commit pass would do this)"
)
worst = max(gaps)
assert worst < _MAX_GAP_CEILING_S, (        # magnitude: load-FRAGILE — the failing line
    f"event loop stalled for {worst:.3f}s during the ingest "
    f"(ceiling {_MAX_GAP_CEILING_S}s); the parse/commit pass is blocking the loop"
)
```

The two assertions detect the **same** regression (a register pass that stopped yielding → one big synchronous batch). The register pass yields per-op — `await session.flush()` + `await encode_endpoint_text(...)` inside `upsert_one_operation` (`backend/src/meho_backplane/operations/ingest/register_ingested.py:561-602`, `_upsert.py:371-382`), ~1275 yields for the 1275-op fixture, one `commit` per spec. If that pass regressed to a synchronous batch, the ~1.7s freeze would collapse the tick count (only the ~parse-window ticks survive, ~10, well under 50) — so the tick floor catches it load-robustly. The magnitude ceiling adds **no** detection the tick count lacks, and is the only line that false-fails under load.

## Desired state

Adopt fix shape (a): keep the deterministic structural assertion as **the** responsiveness guarantee and drop the absolute wall-clock ceiling.

- Remove the `_MAX_GAP_CEILING_S` constant, the `worst = max(gaps)` computation, and the `assert worst < _MAX_GAP_CEILING_S` assertion.
- Keep (and re-document as the guarantee) `assert len(gaps) >= _MIN_TICKS`: the loop yielded ≥ 50 times *during* the ingest. This is load-robust in the correct direction — a yielding loop under contention produces *more* ticks (stretched wall time), a frozen loop produces near-zero during the freeze — and `_MIN_TICKS = 50` is empirically safe under the exact contention that broke the ceiling (the tick assertion passed on both failing runs; only the magnitude line failed).
- Update the module docstring and the `_MAX_GAP_CEILING_S`/tick-count comments so the rationale matches: cite #2706's precedent, state that a magnitude threshold cannot separate ambient jitter (~1.5s observed) from a register-freeze regression (~1.7s) on a shared runner, and that the tick count is the load-robust proof.

```python
# target: single structural assertion, no wall-clock ceiling
assert len(gaps) >= _MIN_TICKS, (
    f"heartbeat only ticked {len(gaps)} times during the ingest "
    f"(need >= {_MIN_TICKS}); the event loop was starved — a register/commit "
    "pass that stopped yielding (one big synchronous batch) would collapse this. "
    "Magnitude of any single gap is deliberately NOT asserted: on a shared runner "
    "ambient jitter (~1.5s, #2706) overlaps a register-freeze regression (~1.7s), "
    "so only the tick count separates load from a real regression."
)
```

Do **not** replace the ceiling with a ratio-against-baseline magnitude check: on `meho-runners-ci` ambient jitter and the regression magnitude overlap, so any magnitude form reintroduces the same fragility (analysed in #2706). The tick count is the honest signal.

## Acceptance criteria

- [ ] `_MAX_GAP_CEILING_S` and the `assert worst < _MAX_GAP_CEILING_S` assertion are removed from `backend/tests/test_ingest_large_spec_event_loop.py` (`grep -n _MAX_GAP_CEILING_S backend/tests/test_ingest_large_spec_event_loop.py` returns empty).
- [ ] The structural assertion `assert len(gaps) >= _MIN_TICKS` remains and is documented as the responsiveness guarantee; `_MIN_TICKS` is unchanged at 50.
- [ ] Mutation check documented in the PR: making the register pass run synchronously (e.g. temporarily remove the per-op `await` yielding / run the batch without yielding) still fails the tick-count assertion — the test keeps its teeth. `grep -n "Verified by mutation" ...` or an equivalent note in the PR body.
- [ ] `test_nondryrun_large_spec_ingest_keeps_event_loop_responsive` still asserts the non-dry-run path landed all ops (`result.ingestion.inserted_count == _OP_COUNT`, `connector_registered is True`) — unchanged.
- [ ] The other two tests in the file (`test_blocking_parse_stays_off_the_event_loop`, `test_catalog_entry_ingest_rides_the_async_job_path`) are untouched.
- [ ] `uv run pytest backend/tests/test_ingest_large_spec_event_loop.py -q` passes locally, including a run under CPU contention (e.g. `-p no:randomly` alongside a background `stress`/`yes` load) to confirm the removed ceiling was the only load-sensitive gate.
- [ ] Module docstring / comments updated to cite #2706 and state why magnitude is not asserted.
- [ ] CHANGELOG `[Unreleased]` gains a `### Fixed` bullet referencing this issue and the PR.
- [ ] If the test is documented in `docs/codebase/spec-ingestion.md` (the responsiveness section around L1988), a one-line note reflects that responsiveness is pinned structurally (tick count), not by a wall-clock ceiling.

## Non-goals

- No change to the ingest pipeline itself (`operations/ingest/pipeline.py`, `register_ingested.py`, `_upsert.py`) — this is a test-only change; the production offload (`asyncio.to_thread`, per-op yielding, 202 + job handle) is correct and stays as-is.
- No new `perf`/`load` pytest marker or non-required lane (fix shape (b)) — shape (a) keeps the full guarantee in the required lane deterministically, so the extra lane is unnecessary.
- No change to `_MIN_TICKS`, `_HEARTBEAT_INTERVAL_S`, `_OP_COUNT`, or the sibling tests.
- Not addressing the separate ghcr TLS-pull flake (#3310) or general runner-contention CI hardening.

## References

- Failing runs: PR #3384 run 33990228779 (unit shard 1/3); merge_group run 33991177823 (unit shard 3/3, ejected PR #3382).
- Precedent (same file, same failure class): #2706 `test(ingest): make the parse-offload guard measure the loop, not the clock` — commit `063cd4a2`. Introducing pin: #2333 commit `34fcc133`.
- Class precedent: #1434 (closed) — de-clocked topology depth-16 perf gate.
- Code path under test: `backend/src/meho_backplane/operations/ingest/register_ingested.py:561-602`, `backend/src/meho_backplane/operations/ingest/_upsert.py:371-382` (per-op yield), `backend/src/meho_backplane/operations/ingest/pipeline.py:900` (`asyncio.to_thread` parse offload).
- Docs: `docs/codebase/spec-ingestion.md` L1988-1998 (async-by-default / responsiveness rationale).
- Best practices — Python pack (`.claude/skills/explore-and-refactor/python_best_practices.md`): "No `time.sleep(n)` / wall-clock as a synchronization/assertion primitive; use deterministic signals" (L337) and "CPU-bound work offloaded with `asyncio.to_thread`" (L279, the property under test). DevOps pack (`.claude/skills/implement-issue/devops_best_practices.md`): flakes hide real signal (L200).

damir-topic:
## Summary

- `test_nondryrun_large_spec_ingest_keeps_event_loop_responsive` asserted an **absolute wall-clock ceiling** (no single heartbeat gap > 0.5 s). On the shared `meho-runners-ci` ARC pool that is load-sensitive by construction: it false-failed at **1.203 s on #3384** (unit shard 1/3, run 33990228779; passed on re-run) and **1.003 s on the #3382 merge_group build** (unit shard 3/3, run 33991177823, 4914 other tests passed) — **ejecting a workflow-only PR from the merge queue** — while the ingest offload was working correctly.
- Ambient scheduler jitter on a saturated pool (~1.5 s, observed in #2706) **overlaps** the ~1.7 s a non-yielding register batch would freeze for, so **no** magnitude threshold — absolute or ratio — can separate load from a regression on a shared box. This is exactly the failure #2706 fixed for the sibling `test_blocking_parse_stays_off_the_event_loop` in this same file.
- **Fix (shape a):** replace the ceiling with a load-robust *structural* signal. Wrap the `register_ingested_operations` call to stamp its window and assert the heartbeat ticked ≥ `_MIN_TICKS_DURING_REGISTER` (10) **inside** it — measuring the DB-commit pass directly (the leg the crash signal implicated) rather than inferring it from a whole-run count the parse (in `asyncio.to_thread`) can pad. A yielding register ticks ~72 in the window healthy (~269 under 8× contention — a slower runner *stretches* the window and ticks *more*); a synchronous batch freezes it and ticks ~0. A secondary whole-run liveness floor (`_MIN_TICKS`) is kept.
- **Test-only change.** The production offload (`asyncio.to_thread` parse, per-op `await session.flush()` yielding, 202 + job handle) is unchanged.

## Why not just raise the ceiling / add a marker lane

A whole-run tick count alone is a *weak* signal — the parse runs off-loop in `asyncio.to_thread` and pads the total even if the register froze (confirmed: injecting a 2 s freeze before the per-op loop still passed a whole-run count). Measuring inside the register window is the honest guarantee with real teeth. Shape (b) (a `perf` marker + non-required lane) was rejected because shape (a) keeps the full guarantee deterministically in the required lane with no extra CI machinery.

## Test plan

- [x] `mypy` — no new findings (mypy is scoped to `src`; this change is test-only. The one bare-mypy error, `icmp.py:205 MSG_ERRQUEUE`, is a pre-existing macOS-only artifact — `socket.MSG_ERRQUEUE` is Linux-only, resolves in CI — in a file this PR does not touch)
- [x] `code-quality.py --diff` — pass (exit 0)
- [x] `pytest tests/test_ingest_large_spec_event_loop.py` — 3 passed, run 3/3 healthy
- [x] **Load-robustness:** passes under 8× CPU contention (`yes` ×8) — the exact condition that broke the old ceiling
- [x] **Teeth (verified by mutation):** patching `upsert_one_operation` to a non-yielding synchronous batch (per-op `time.sleep`, no `await`) freezes a 1.943 s register window to **1** in-window tick (< floor 10) → the assertion fails, while `inserted_count` still lands 1275. The whole-run count would have missed this.
- [x] Sibling tests `test_blocking_parse_stays_off_the_event_loop` and `test_catalog_entry_ingest_rides_the_async_job_path` untouched and passing

## Out of scope

- The ghcr TLS-pull flake (#3310) and general runner-contention CI hardening.
- Any change to the ingest pipeline itself — the offload is correct.

Closes #3399


damir-topic:
## /auto-review-pr summary — verdict: APPROVE (degraded-comment)




**Delivered as a plain comment, not a formal GitHub approval** — the PR author cannot submit an APPROVE review on their own PR (#2733). This records the autonomous review verdict; the formal App-identity approval + merge stays with the orchestrator.

**Anchor task:** #3399 — de-clock the event-loop responsiveness pin
**Surface area:** backend (test) + devops (CI reliability)
**Head SHA reviewed:** `f0afb356`

### Verdict path
Iteration 1 (`9b005fb3`) → **REQUEST_CHANGES** (1 blocker) → fix → Iteration 2 (`f0afb356`) → **APPROVE**.

**B1 (fixed in `f0afb356`):** the first cut kept a secondary whole-run `assert len(beats) >= 50`. That is itself an absolute count floor over a short (~0.8 s / ~80-beat) run, so a ~1.6× faster runner would dip below 50 and false-fail — a milder instance of the exact fragility this PR removes — and it added no teeth the in-window count lacks. Removed it and the orphaned `_MIN_TICKS` constant; the in-register-window tick count is now the sole responsiveness assertion (it scales with the window, so it is speed-robust, not just load-robust).

### Acceptance criteria
- [x] `_MAX_GAP_CEILING_S` + its assertion removed — `grep` returns empty
- [x] Load-robust structural guarantee retained **with teeth** — the DB-commit call is wrapped to stamp its window and the loop must tick ≥ 10 times inside it (~72 healthy, ~269 under 8× contention). Verified by mutation: a non-yielding register collapses the in-window count to **1** (< 10) and fails the test, while `inserted_count` still lands 1275
- [x] Passes under 8× CPU contention — the exact condition that broke the old ceiling
- [x] Sibling tests `test_blocking_parse_stays_off_the_event_loop` / `test_catalog_entry_ingest_rides_the_async_job_path` untouched and passing
- [x] CHANGELOG `[Unreleased] ### Fixed` entry (#3399 / #3404)

### Integration verification
- [x] `mypy` — no new findings (mypy is scoped to `src`; this is a test-only change). The lone bare-mypy error `icmp.py:205 MSG_ERRQUEUE` is a pre-existing macOS-only artifact (Linux CI resolves it) in an untouched file
- [x] `pytest tests/test_ingest_large_spec_event_loop.py` — 3 passed, ×3 healthy + under 8× contention

### Predicted vs. actual diff
- ✅ `backend/tests/test_ingest_large_spec_event_loop.py`, `CHANGELOG.md` — exactly as predicted, nothing else touched

### Docs
- [x] `docs/codebase/spec-ingestion.md` — verified unchanged (its responsiveness section documents the production async-by-default path, not this test's assertion mechanism; the test is self-documented by its module docstring, which was updated)

### What worked well
- The design directly reuses #2706's in-window presence-vs-absence pattern and measures the DB-commit pass itself, rather than a whole-run count the `to_thread` parse can pad — a genuinely stronger signal than the ceiling it replaces.

Issue #262, opened 2026-05-12
Task: G0.4-T5 /api/v1/retrieve operator-facing diagnostic route + query_hash-only audit row (no raw query stored)


damir-topic:
Parent initiative: #225

## Target

`/api/v1/retrieve` operator-facing diagnostic route + corresponding MCP resource:

- `backend/src/meho_backplane/api/v1/retrieve.py` — new router. `POST /api/v1/retrieve` with `{query, source?, kind?, limit?}` body.
- `backend/src/meho_backplane/main.py` — register the router.
- `backend/src/meho_backplane/mcp/resources/retrieve.py` — MCP resource registration for diagnostic retrieval (optional in v0.2).
- `backend/tests/test_api_v1_retrieve.py` — new file. RBAC + tenant-scoping + happy/sad paths.

## Context

T5 ships the **HTTP diagnostic surface** for retrieval. The primary callers are G4/G5 (which call `retrieve()` directly in-process), but the API route is useful for:

- Operator diagnostics: "is retrieval returning what I expect against the current kb corpus?"
- Testing/iteration during G4 + G5 implementation.
- Future MCP tool registrations (G4 may register `meho.kb.search` as an MCP tool that wraps this surface).

**Route shape:**

```
POST /api/v1/retrieve
Body: {
  "query": "kubernetes ingress troubleshooting",
  "source": "kb",          // optional
  "kind": "kb-entry",      // optional
  "limit": 10              // optional, default 10, max 50
}

Response 200: {
  "hits": [
    {
      "document_id": "uuid",
      "source": "kb",
      "source_id": "kubernetes-ingress",
      "kind": "kb-entry",
      "body": "...",
      "metadata": {...},
      "fused_score": 0.032,
      "bm25_score": 0.85,
      "cosine_score": 0.72,
      "bm25_rank": 2,
      "cosine_rank": 1
    },
    ...
  ],
  "query_duration_ms": 45.7
}
```

**RBAC: `operator` role minimum.** Diagnostic retrieval is operator-level; `read_only` operators don't get it (they typically use G4 `meho kb search` as their entry point, which v0.2.next can authorize differently).

**Tenant-scoped automatically.** Every call passes `operator.tenant_id` to `retrieve()`. Cross-tenant retrieval impossible by construction.

**Pagination via `limit` only.** Cursor pagination would require carrying the RRF state across requests; v0.2 keeps it simple — call with limit=50 for the top 50, that's the max.

[...]

PR #3435, opened 2026-09-06
fix(audit): enforce audit_log append-only at the datastore (S18)

damir-topic:
## Summary

- Enforce `audit_log` append-only **at the datastore**, not by code convention (security review 2026-09-06 finding S18). CLAUDE.md postulate 7 / v0.1-spec §6 makes the audit trail append-only and tamper-evident, but until now nothing at the datastore stopped a direct `UPDATE`/`DELETE` — anyone with write access to Postgres (the app DB role, a DB operator, or in-process code execution) could silently rewrite or erase the audit row of the operation just performed.
- Migration `0100` installs a `BEFORE UPDATE OR DELETE ... FOR EACH ROW` trigger (`audit_log_append_only` + function `audit_log_reject_mutation()`) that `RAISE`s. `INSERT`/`SELECT` are untouched, so the synchronous audit write path keeps working; a direct `UPDATE`/`DELETE` fails under **every** role, including a superuser (a row-level trigger fires regardless of privilege).
- Dialect-guarded: the plpgsql DDL runs only on `postgresql`. The SQLite dev/test lane is a clean no-op, so the always-on unit lanes stay green (they assert the app-level convention; the trigger is the production-datastore backstop).
- Documented the new "constraint enforcement via triggers (Postgres-only)" migration pattern in `docs/codebase/migrations.md`.

## Already on `main` (not re-done here)

- The app-level convention already held on `main` (write path is INSERT-only; the only other `AuditLog(...)` is the topology-delete append; a codebase grep finds only `select(AuditLog)` reads — reconfirmed against `src/`). This task adds the **datastore** enforcement that was missing; no prior containment PR touched `audit_log` immutability.

## Design notes

- **Trigger over REVOKE.** The AC accepts a trigger *and/or* an `INSERT`+`SELECT`-only app role. The trigger is self-contained (no external role provisioning / ops runbook dependency) and enforces under every role including the owner, which is exactly what the container test asserts. The role split is left to the retention-role work called out as out-of-scope.
- **Retention stays out of scope.** No retention/reaper path deletes `audit_log` today (the flight-recorder reaper deletes `DispatchTrace`/`DispatchTraceSpan`, not `audit_log`). When retention lands it drops whole **partitions** — `DROP TABLE` on a partition and `TRUNCATE` both bypass row-level triggers — or runs under a dedicated role, so this row-level trigger needs no bypass hatch now.
- **Blast radius.** No `src/` change. The only test paths that remove `audit_log` rows are on SQLite (`test_topology_history_hook`, where the trigger is a no-op) or use `TRUNCATE` (integration/acceptance conftests — `TRUNCATE` does not fire a row-level `DELETE` trigger). Verified there is no `UPDATE`/`DELETE`/`delete(AuditLog)` on `audit_log` in `src/`.

## Out-of-scope cleanup carried to satisfy AC5

- `alembic/versions/0023`: collapsed `sa.text("'{}'") if is_postgres else sa.text("'{}'")` (both branches identical → RUF034) to `sa.text("'{}'")`. Byte-identical DDL. This pre-existing lint error was the **only** thing failing AC5's directory-scoped `uv run ruff check alembic/versions` (CI lints `src/`+`tests/`, never `alembic/`, so it had never surfaced). One-line, zero-risk, same subsystem.

## Test plan

- [x] AC1 — `grep -rniE "trigger|revoke" backend/alembic/versions | grep -i audit_log` returns the new revision `0100_audit_log_append_only_trigger.py`.
- [x] AC2 — new Postgres-container test `TestAuditLogAppendOnlyTrigger::test_audit_log_update_delete_raises` (in `tests/migrations/test_migration_rollback.py`, reusing that file's Docker-socket skip). **Ran green against real Postgres** (Docker present in the dev box): raw `UPDATE`/`DELETE` both raise with the append-only message; raw `INSERT`/`SELECT` succeed and the row survives unchanged. Skips where Docker is absent; runs in CI.
- [x] AC3 — `uv run alembic upgrade head && uv run alembic downgrade -1 && uv run alembic upgrade head` completes (ends at `0100 (head)`).
- [x] AC4 — `uv run pytest tests/test_db_models.py tests/migrations -q --ignore=tests/migrations/test_migration_rollback.py` → 434 passed.
- [x] AC5 — `uv run ruff check alembic/versions` clean; `uv run mypy src/meho_backplane/db/models.py` → success.
- [x] `ruff check` + `ruff format --check` on `src/ tests/` (CI scope) — clean; migration-compat guard passes.

## Out of scope

- Cryptographic tamper-evidence (hash-chaining / signed rows) — a stronger model than immutability; separate.
- `audit_log` retention / `raw_payload` data-minimization (companion sibling under #262).
- Connector code-exec / SSRF findings (F03/F05/F13) — tracked separately under #262.
- Other trails (`broadcast_event`, topology history) — this task is `audit_log` only.

Closes evoila-bosnia/meho-internal#306

### DIFF
--- a/backend/alembic/versions/0023_create_approval_request.py
+++ b/backend/alembic/versions/0023_create_approval_request.py
@@ -225,7 +225,7 @@ def upgrade() -> None:
             "proposed_effect",
             proposed_effect_type,
             nullable=False,
-            server_default=sa.text("'{}'") if is_postgres else sa.text("'{}'"),
+            server_default=sa.text("'{}'"),
         ),
         # Lifecycle columns.
         sa.Column(
--- a/backend/alembic/versions/0100_audit_log_append_only_trigger.py
+++ b/backend/alembic/versions/0100_audit_log_append_only_trigger.py
@@ -1,0 +1,115 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 evoila Group
+
+"""Enforce ``audit_log`` append-only at the datastore (security S18).
+
+Revision ID: 0100
+Revises: 0099
+Create Date: 2026-09-07
+
+CLAUDE.md postulate 7 (v0.1-spec section 6) makes the audit trail
+synchronous, append-only and tamper-evident -- the load-bearing property
+of the whole governance model. Until this migration that property was a
+**code convention only**: the write path is INSERT-only and no code path
+issues an ``UPDATE`` / ``DELETE`` against ``audit_log``, but nothing at the
+datastore stopped one. Anyone with write access to the Postgres instance
+(the app DB role, a DB operator, or in-process code execution) could
+silently rewrite or erase the audit row of the operation just performed --
+the free cover-up a compromise would otherwise inherit.
+
+Security review 2026-09-06 finding S18 (Task meho-internal#306 under
+Initiative #262). Defence-in-depth: an invariant the application depends on
+belongs at the datastore that owns the data, not scattered as convention.
+
+What this migration adds (PostgreSQL only)
+------------------------------------------
+
+* ``audit_log_reject_mutation()`` -- a ``plpgsql`` trigger function that
+  ``RAISE``s on any row it fires for. ``TG_OP`` names the rejected
+  operation in the error message.
+* ``audit_log_append_only`` -- a ``BEFORE UPDATE OR DELETE ... FOR EACH
+  ROW`` trigger that calls the function. ``INSERT`` and ``SELECT`` are
+  untouched, so the synchronous audit write path keeps working; a direct
+  ``UPDATE`` or ``DELETE`` fails at the datastore under **every** role,
+  including the app role and a DB superuser (a row-level trigger fires
+  regardless of privilege).
+
+Dialect guard
+-------------
+
+The enforcement is a production-datastore guarantee, so the DDL runs only
+on ``postgresql``. The SQLite dev/test lane (aiosqlite) cannot execute
+plpgsql trigger DDL, and the unit lanes assert the app-level convention
+rather than the datastore trigger; ``upgrade()`` / ``downgrade()`` are a
+clean no-op there. This keeps the SQLite ``alembic upgrade head`` smokes in
+``tests/test_db_models.py`` green.
+
+Out of scope (tracked separately under #262)
+--------------------------------------------
+
+* Retention / archival ``DELETE`` of aged rows. No retention job exists
+  today (a codebase grep finds only ``select(AuditLog)`` reads). When one
+  lands it drops whole **partitions** -- ``DROP TABLE`` on a partition and
+  ``TRUNCATE`` both bypass row-level triggers -- or runs under a dedicated
+  retention role, so this row-level trigger does not need a bypass hatch.
+* Cryptographic tamper-evidence (hash-chaining / signed rows) is a stronger
+  model than immutability and is not this change.
+
+Reversibility contract
+----------------------
+
+``downgrade()`` drops the trigger then the function (reverse of create
+order), Postgres only. Additive on PostgreSQL (a new trigger + function, no
+ALTER of an existing object); a no-op on SQLite.
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0100"
+down_revision: str | None = "0099"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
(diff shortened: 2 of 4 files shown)


damir-topic:
> satisfied by this comment.

**Anchor task:** meho-internal#306 — enforce `audit_log` append-only at the datastore (security review S18)
**Surface area:** backend / devops (Alembic migration + testcontainers)
**Verdict:** APPROVE
**Head SHA reviewed:** `1c53865a57ab9c536f2214a75ea18cd03035a8db`

### Acceptance criteria
- [x] **AC1** — `grep -rniE "trigger|revoke" backend/alembic/versions | grep -i audit_log` returns the new revision `0100_audit_log_append_only_trigger.py`.
- [x] **AC2** — new Postgres-container test `TestAuditLogAppendOnlyTrigger::test_audit_log_update_delete_raises` asserts raw `UPDATE`/`DELETE` both raise and raw `INSERT`/`SELECT` succeed. **Ran green against real Postgres** (pgvector:pg16, 1 passed). It skips when Docker is absent, matching the existing `_DOCKER_AVAILABLE` discipline.
- [x] **AC3** — `alembic upgrade head && downgrade -1 && upgrade head` completes, ending at `0100 (head)` (verified on SQLite; the trigger DDL is a clean no-op there, and the Postgres path is exercised by the container test).
- [x] **AC4** — `pytest tests/test_db_models.py tests/migrations -q --ignore=tests/migrations/test_migration_rollback.py` → **434 passed**.
- [x] **AC5** — `ruff check alembic/versions` clean; `mypy src/meho_backplane/db/models.py` → success.

### Regression bar (security fix)
- **Fails closed:** the `BEFORE UPDATE OR DELETE ... FOR EACH ROW` trigger `RAISE`s (SQLSTATE P0001); mutation is rejected, not silently allowed.
- **Test would fail without the fix:** independently verified — neutralising the trigger creation in `upgrade()` makes `test_audit_log_update_delete_raises` FAIL (`update_raised` false); restoring the trigger makes it pass. The test genuinely guards the invariant.
- **Fires under every role:** a row-level trigger fires regardless of privilege, including the app role and a superuser — the correct enforcement for the S18 threat model.

### Project-board alignment
|---|---|---|---|---|---|
| meho-internal#306 | In Review | Closes | ✅ | ✅ full | close on merge |

### Predicted vs. actual diff
- ✅ `0100_audit_log_append_only_trigger.py` — dialect-guarded plpgsql trigger + function, reversible downgrade (drop trigger then function).
- ✅ `test_migration_rollback.py` — testcontainers slice mirroring the existing `TestForwardCompatRollback` pattern; per-statement transaction isolation so a rejected mutation doesn't poison the INSERT verification.
- ✅ `docs/codebase/migrations.md` — new "Constraint enforcement via triggers (Postgres-only)" section.
- ❓ `0023_create_approval_request.py` — one-line RUF034 collapse of an identical-branch ternary; **byte-identical DDL**, disclosed in the PR body, and necessary because AC5 scopes ruff at the `alembic/versions` directory (CI lints `src/`+`tests/` only, so it had never surfaced). Justified out-of-scope, zero-risk.

### Integration verification
- [x] `ruff check alembic/versions` — clean; changed files clean; `ruff format --check` clean
- [x] `mypy src/meho_backplane/db/models.py` — success
- [x] `pytest` new S18 container test — 1 passed against real Postgres
- [x] `pytest` SQLite unit lanes — 434 passed
- [x] Alembic round-trip (AC3) — upgrade/downgrade/upgrade ends at `0100 (head)`
- [x] Regression check — test fails with trigger neutralised, passes with fix
- [x] Adjacent-behaviour walk — no `UPDATE`/`DELETE`/`delete(AuditLog)` on `audit_log` in `src/`; `test_topology_history_hook`'s `delete(AuditLog)` runs on SQLite (trigger no-op); integration/acceptance conftests clear the table with `TRUNCATE`, which does not fire row-level `DELETE` triggers in Postgres — per-test cleanup unaffected

### Findings (inline)
None. No blockers, majors, minors, or questions.

### Docs
- [x] `docs/codebase/migrations.md` — updated with the Postgres-only trigger-enforcement pattern.

- The regression test is honest: it asserts the RAISE *message* content and the survived-row state, not just that "something raised", and isolates each mutation in its own transaction so a rolled-back RAISE can't mask the INSERT verification.
- Scope discipline is tight — no `src/` change, the out-of-scope 0023 touch is one byte-identical line and is explicitly disclosed with rationale, and retention/TRUNCATE bypass semantics are correctly reasoned in both the migration docstring and the PR body.

_Note: CHANGELOG intentionally untouched — batched by the orchestrator at merge time, not a defect._

PR #297, opened 2026-05-12
feat(mcp): reference tool meho.status + reference resource meho://tenant/<id>/info (G0.5-T4)

zdamir:
## Summary

T4 ships the two reference implementations downstream G3-G9 connector tools and resources will copy as templates:

- **`meho.status` (tool)** — the MCP transport equivalent of `GET /api/v1/health`. `mcp/tools/meho_status.py` registers it against T3's registry; the handler calls `meho_backplane.api.v1.health.build_health_response()` (extracted during T4) so the chassis route and the MCP tool share the same federation-proof probe chain rather than diverging. `inputSchema` uses `additionalProperties: false` so extra args fail validation under the `Draft202012Validator` pinned in T3.
- **`meho://tenant/{tenant_id}/info` (resource)** — establishes the `meho://` URI scheme and the tenant-boundary enforcement pattern every downstream resource will follow. The handler validates the bound `tenant_id` as a UUID, checks it matches `operator.tenant_id` *before* hitting the DB (so a probe can't learn whether an arbitrary tenant exists), queries `Tenant`, and returns `{id, slug, name, operator_role}`.
- **Error code shape.** Cross-tenant reads, invalid UUIDs, and missing rows all surface as `McpInvalidParamsError` (-32602) — the JSON-RPC transport carries error codes, not HTTP statuses, so the issue body's mention of "400" / "403" maps onto INVALID_PARAMS at this layer.

## Test plan

- [x] `ruff check src/ tests/` — clean
- [x] `ruff format --check` — 75 files formatted
- [x] `mypy src/ --ignore-missing-imports` — 38 source files (was 36 pre-T4), no issues
- [x] `pytest tests/ -q --deselect …test_metrics_endpoint…` — **345 passed** (+7 vs main: the 7 new T4 tests), 1 deselected (pre-existing macOS-only flake on `process_resident_memory_bytes`, file unchanged on this branch)

### AC coverage

- [x] `tools/list` returns `meho.status` with well-formed Draft 2020-12 `inputSchema` — `test_tools_list_exposes_meho_status_with_well_formed_input_schema`
- [x] `tools/call meho.status` returns wire-identical health bundle — `test_tools_call_meho_status_returns_health_response_shape`
- [x] `tools/call` with `arguments={"foo": "bar"}` → -32602 (extra prop fails `additionalProperties: false`) — `test_tools_call_meho_status_rejects_extra_arguments`
[...]

Issue #275, opened 2026-05-12
Marketing Task M.01: README audit for stale claims (first run + baseline)


damir-topic:
## What

Audit the current `evoila/meho` README. Compare every claim against the actual state of the repo (issues, code, releases) and produce a delta document.

## Why

Bi-weekly accuracy audits are the operational practice that keeps the README honest. The first one establishes the baseline and the workflow.

## Done-when

- Read the entire README at https://github.com/evoila/meho once, end-to-end.
- For every claim that asserts a specific fact (e.g. "the chart is on GHCR," "per-PR CI deploys to a real cluster," "first connector is in flight"): verify against the public repo state. Note which issue or PR or release backs the claim, or note that the claim is stale.
- Output: a short markdown delta document (≤ 1 page) with three sections: ✅ Claims that check out; 🟡 Claims that are subtly wrong or imprecise; ❌ Claims that are flatly wrong or unverifiable.
- Attach the delta as a comment on this issue. If anything is flatly wrong, file follow-up issues to fix the README (or escalate to your mentor).

## Skill level

`student-doable`. No senior approval needed to publish the audit document — it's an internal-team artefact, not external content.

## Estimated time

≤ 2 hours.

## Parent

Goal: MEHO's marketing stays in honest sync with the product's actual state (G_M5).

damir-topic:
Migrated to the private marketing repo: **evoila-bosnia/meho-marketing#13**.

Marketing tracking is private; this issue was created in the public product repo by mistake. Closing here, no further updates.

PR #3459, opened 2026-09-07
fix(connectors): cap the dispatch-transport response body (S09 / #297)

damir-topic:
## Summary
- The shared generic-connector transport buffered every vendor response fully into memory before checking its size: both `_request_json` and `_post_json` called `client.request(...)` in default non-streaming mode, so httpx read the whole body and `json_payload_or_empty` then expanded it again into a Python object graph. The pooled client carried only an `httpx.Timeout` — no size guard — so one oversized body could OOM-kill the single-replica pod (S09, evoila-bosnia/meho-internal#297).
- Add a configurable response-body byte cap on the outbound-dispatch path, mirroring the streamed idiom the spec-fetch (`openapi._MAX_SPEC_BYTES`) and event-ingest (`events_ingest._read_capped_body`) paths already enforce. New `_read_capped_json_response` streams via `client.stream(...)` (through the same `_SameOriginRedirectClient.send` override, so same-origin-redirect + destination-pinning behaviour is unchanged), does a `Content-Length` fast-reject, then a running byte total across `aiter_bytes()` that aborts before the body is materialised.
- Over-cap bodies raise `ResponseTooLargeError` — a plain `Exception` that flattens through the dispatcher's generic arm into a structured `connector_error` and, being outside the `httpx.ConnectError` / `httpx.HTTPStatusError` families, stays out of the idempotent-path retry budget (deterministic verdict, never retried).
- Under-cap responses are re-materialised into a fully-read `httpx.Response` (same status/headers/request/extensions), so the flight-recorder span, `raise_for_status`, and `json_payload_or_empty` are **byte-identical** to the previous non-streamed path. Default cap 100 MiB, tunable via `MEHO_CONNECTOR_MAX_RESPONSE_BYTES`. Scoped to the two dispatch seams; the deliberate streaming binary paths (`library_download.py`, `net/http_probe.py`) are untouched.

## Already met on main (drift check)
- The sibling SSRF destination-binding hardening (#275 / F13) on this same file — `SsrfBlockedError`, `build_pinned_async_transport`, the connect-time re-screen — had already landed. This PR builds on that: `_read_capped_json_response` streams **through** the pinned transport unchanged, so the two hardenings compose (one bounds *where* a connection may go, this bounds *how much* it may buffer). No other acceptance criterion of #297 was already present — the outbound-dispatch byte cap was still absent.

## Test plan
- [x] `ruff check src/meho_backplane/connectors/adapters/http.py tests/test_connectors_http_adapter.py` — clean (whole-tree `ruff check` + `ruff format --check` also clean)
- [x] `mypy src/meho_backplane/connectors/adapters/http.py` — clean
- [x] `pytest tests/test_connectors_http_adapter.py tests/test_operations_dispatcher.py -q` — 146 passed
- [x] AC grep — `grep -nE "_MAX_RESPONSE_BYTES|content-length|iter_bytes|client.stream" …/http.py` shows the module-level cap constant, the `Content-Length` reject, and the streamed running-total guard
- [x] New `test_request_json_rejects_oversize_response` returns an over-cap **invalid-JSON** body via respx and asserts `ResponseTooLargeError` — proving the body is refused before `resp.json()` would have raised `JSONDecodeError`
- [x] New tests cover the no-`Content-Length` streaming-guard path, the not-retried invariant, `_post_json` oversize, and under-cap regression guards on both seams (parsed payload unchanged), plus `MEHO_CONNECTOR_MAX_RESPONSE_BYTES` env parsing
- [x] Full backend unit lane run (`tests/` minus `integration/`+`migrations/`, `-n auto`): the only failures were in unrelated modules and are environmental/concurrency (approvals-BFF tests pass in isolation; `net` ICMP/DNS raw-socket tests; `test_chart_mcp_resource_uri` fails because the local box has **Helm v4** and the chart renders under Helm 3). My diff touches zero chart/net/approvals files.

## Docs
- No dedicated `docs/codebase/` artifact exists for the shared HTTP-adapter transport internals (its sibling SSRF hardening #275 added none either); the change is documented in-code via docstrings on the cap constant, `ResponseTooLargeError`, `_reject_oversize_content_length`, and `_read_capped_json_response`, matching the `_MAX_SPEC_BYTES` precedent. Docs verified — nothing invalidated.

## Out of scope
- SSRF destination-binding (#275 / F13); spec-ingest / inbound-MCP / event-ingest caps (already present); per-tenant response-quota / rate-limit design; chart-side memory tuning (this fix is app-layer and holds regardless of the pod limit).

Closes evoila-bosnia/meho-internal#297

### DIFF
--- a/backend/src/meho_backplane/connectors/adapters/http.py
+++ b/backend/src/meho_backplane/connectors/adapters/http.py
@@ -83,6 +83,7 @@ from __future__ import annotations
 
 import asyncio
 import hashlib
+import os
 import ssl
 from typing import Any
 
@@ -124,6 +125,30 @@ class SsrfBlockedError(httpx.ConnectError):
     dispatcher plumbing. Explicitly excluded from :func:`_retryable`:
     the rejection is deterministic policy, not a transient network
     failure, so retrying would only re-run the DNS lookup.
+    """
+
+
+class ResponseTooLargeError(Exception):
+    """Dispatch refused: a vendor response body exceeded the byte cap.
+
+    Raised by :func:`_read_capped_json_response` when a vendor response —
+    declared via ``Content-Length`` or measured while streaming — exceeds
+    :data:`_MAX_RESPONSE_BYTES`, *before* the body (or its parsed object
+    graph) is materialised. The shared generic-connector transport
+    otherwise buffers every vendor response fully into memory, so a single
+    oversized body — from a compromised/malicious appliance, a MITM on a
+    ``verify_tls=false`` channel, or a pathologically large legitimate
+    listing — could exhaust the single-replica pod's heap and OOM-kill it
+    (evoila-bosnia/meho-internal#297).
+
+    A plain :class:`Exception` (not an :class:`httpx.HTTPError`): it is a
+    policy rejection, not a transport fault, so it flattens through the
+    dispatcher's generic ``except Exception`` arm into the structured
+    ``connector_error`` shape carrying this message. Being outside the
+    ``httpx.ConnectError`` / ``httpx.HTTPStatusError`` families, it is
+    automatically excluded from :func:`_retryable` (a deterministic
+    over-cap verdict never resolves on replay), with no explicit exclusion
+    needed.
     """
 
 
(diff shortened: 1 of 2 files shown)


damir-topic:
> **DEGRADED MODE — machine review identity unavailable.** This verdict is advisory prose, **not** a GitHub review; `reviewDecision` is unchanged and branch protection is not satisfied by this comment. The orchestrator posts the formal App approval at merge time (verdict-relay).

**Anchor task:** evoila-bosnia/meho-internal#297 — S09: configurable response-body byte cap on the outbound generic-connector dispatch transport
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `8ac29242d586caceb3c986f9a5855f71d2f3dcae`

### Acceptance criteria
- [x] Module-level byte-cap constant + `Content-Length` fast-reject + streamed running-total guard on the dispatch path — verified `grep -nE "_MAX_RESPONSE_BYTES|content-length|iter_bytes|client.stream"` matches `_DEFAULT_MAX_RESPONSE_BYTES`/`_MAX_RESPONSE_BYTES` (http.py:345/370), `_reject_oversize_content_length` reading `content-length` (:382), and `client.stream(...)` + `aiter_bytes()` running total (:431/:444).
- [x] `test_request_json_rejects_oversize_response` returns an over-cap **invalid-JSON** body via respx and asserts `ResponseTooLargeError` — proving the body is refused before `resp.json()` would have raised `JSONDecodeError` (this is what makes the test fail without the fix). Body never `resp.json()`-parsed.
- [x] Under-cap regression guards for both `_request_json` and `_post_json` (parsed payload returned unchanged), plus a `_post_json` oversize test and a no-`Content-Length` streaming-guard test.
- [x] `pytest tests/test_connectors_http_adapter.py tests/test_operations_dispatcher.py -q` → **146 passed** (ran in a fresh PR-head worktree).
- [x] `ruff check` + `ruff format --check` + `mypy src/.../http.py` → all clean.

### Project-board alignment
|---|---|---|---|---|---|
| #297 | In Review | Closes | ✅ | ✅ full | close on merge |

### Predicted vs. actual diff
- ✅ `connectors/adapters/http.py` — cap constant, env resolver, Content-Length fast-reject, streamed `_read_capped_json_response` wired into both `_request_json` and `_post_json`.
- ✅ `tests/test_connectors_http_adapter.py` — 8 new tests covering the cap, both seams, retry-exclusion, env parsing, and under-cap regression.
- No unexpected/undisclosed files; the PR commit touches exactly these two. (The `library_download.py` / `net/http_probe.py` deliberate streaming paths are correctly left untouched.)

### Integration verification
- [x] `ruff check` — clean
- [x] `mypy` — clean
- [x] `pytest` (adapter + dispatcher modules) — 146 passed
- [x] AC grep — all four idioms present
- [x] Fail-closed reasoning traced end-to-end: `ResponseTooLargeError` is a plain `Exception` → falls past the dispatcher's `httpx.HTTPStatusError` / `httpx.ConnectError` / `hvac.Forbidden` arms into the generic `except Exception` → `result_connector_error`; `is_probe_refused()` is False for it. `_retryable()` returns False (verified in source + asserted by test), so the idempotent path issues exactly one call.
- [x] SSRF/redirect composition confirmed: `client.stream()` routes through `_SameOriginRedirectClient.send(stream=True)`, so #275/F13 destination-pinning + same-origin-redirect behaviour is unchanged.
- [x] Byte-identical downstream confirmed: under-cap body re-materialised into `httpx.Response(status, headers, content, request, extensions)`; `.json()` reads stored content verbatim (no double content-decoding, since `aiter_bytes()` already applied the decoder), so flight-recorder span, `raise_for_status`, and `json_payload_or_empty` see the same response as before.

### Findings
None.

- The DevOps reference on #297 floated an optional memory gauge/metric for the exhaustion condition; the reject path is already logged + audited, so a dedicated metric is a future enhancement, correctly out of scope here.
- Under-cap re-materialisation briefly holds ~2× the body (chunk list + joined bytes), bounded by the cap and no worse in kind than the pre-fix full buffer.

### Docs
- [x] `docs/codebase/` — n/a. No dedicated artifact exists for the shared HTTP-adapter transport internals (sibling #275 added none); the change is documented in-code via docstrings on the cap constant, `ResponseTooLargeError`, `_reject_oversize_content_length`, and `_read_capped_json_response`, matching the `_MAX_SPEC_BYTES` precedent. Nothing existing is invalidated.

### What worked well
- The invalid-JSON over-cap body is a genuinely rigorous regression proof: it distinguishes "refused before parse" from "parsed then rejected" in a single assertion.
- Fail-safe env parsing (`MEHO_CONNECTOR_MAX_RESPONSE_BYTES`): a typo or non-positive value falls back to the default rather than silently disabling the guard.
- Reusing httpx's own `next_request` through the existing redirect override means the two hardenings (where-you-connect / how-much-you-buffer) compose without either re-implementing the other.

PR #3524, opened 2026-09-09
fix(connectors): buffer raw wire bytes in the dispatch body cap so compressed vendor responses decode once

### LINKED ISSUE #3521: v0.34.0 regression: DecodingError "incorrect header check" on every gzip-compressed vendor response (dispatch body-cap double-decodes)
## Problem

Since v0.34.0, every read through an `HttpConnector`-based connector whose vendor returns a compressed response body fails with `httpx.DecodingError: Error -3 while decompressing data: incorrect header check` (surfaced as `connector_error`, `exception_class=DecodingError`). Reproduced against a vCenter 8.x target behind an ingested vmware-rest connector: every REST read fails, small and large alike, because vCenter gzips every response. Uncompressed or empty-body responses (e.g. 204 No Content write acks) are unaffected, which is why the failure is selective.

## Root cause

The dispatch-transport body-size cap added in PR #3459 (commit `79b65024`) rewrote the shared HTTP seam (`backend/src/meho_backplane/connectors/adapters/http.py`, `_read_capped_json_response`, v0.34.0 lines 396-459) to stream and re-materialise the response body:

- `http.py:444` collects chunks from `resp.aiter_bytes()` inside `client.stream(...)`. httpx's `aiter_bytes()` runs the Content-Encoding decoder, so the collected chunks are the **already-decompressed** plaintext body.
- `http.py:452-457` rebuilds `httpx.Response(status_code=…, headers=resp.headers, content=b"".join(chunks), …)`, copying `Content-Encoding: gzip` verbatim over a body that is now plaintext.
- On httpx 0.28.1 a `content=`-Response is read eagerly in `__init__` (`Response.__init__ → self.read() → iter_bytes()`), so the gzip decoder runs a **second** time over plaintext and zlib raises `-3`. The exception fires at the `httpx.Response(...)` **construction** (http.py:452), not at the later `.json()` access — so `record_vendor_call` (flight recorder) and `raise_for_status` are never reached for a gzip response. The failure is independent of body size (the header check is at byte 0).

## Scope

All `HttpConnector`-based connectors — the generic ingested base and the typed HTTP subclasses — whose vendor sends `Content-Encoding: gzip`/`deflate`. The capped reader is called **unconditionally** from both `_request_json` and `_post_json`. vCenter REST gzips every response, so **all vmware-rest reads fail** regardless of size.

Unaffected: uncompressed/identity-encoded bodies (the `content=` round-trip is a no-op) and 204 / empty-body write acks (`json_payload_or_empty` short-circuits, and the crash is at construction). Vault KV reads use a separate `hvac` client, not `HttpConnector`, so they bypass this path entirely.

Operational note: vCenter power-action POSTs and `DELETE /vcenter/vm/{vm}` return 204 and themselves ack, but the inventory/get reads used to plan or verify a governed teardown return gzip JSON and **fail** — so a governed teardown breaks at its read/resolve steps even though the raw power-off/delete verbs would ack.

## Fix

Buffer the **raw** wire bytes — change `resp.aiter_bytes()` to `resp.aiter_raw()` at the cap loop (http.py:444). `aiter_raw()` yields the still-gzip-encoded wire bytes, so `content=b"".join(chunks)` matches the retained `Content-Encoding: gzip` header and httpx decodes **exactly once** at construction — byte-identical to the pre-v0.34.0 `client.request()` path. The cap stays coherent: it continues to bound the *compressed* transfer bytes, which is exactly #3459's "bound a multi-GB transfer" intent. Also update the two stale `aiter_bytes` docstring mentions (`http.py:419`, `test_connectors_http_adapter.py`).

Add a **gzip-encoded response regression fixture** — the existing adapter tests use plain bodies only, which is why CI missed the regression.

**No config/env workaround exists**: `_read_capped_json_response` runs unconditionally and `MEHO_CONNECTOR_MAX_RESPONSE_BYTES` only sets the cap *size* (lowering it converts the failure to `ResponseTooLargeError`; raising it changes nothing — the header check is at byte 0). The only non-code mitigation is rolling the deployed image back to v0.33.5, which is a rollback, not a config workaround, and also drops the v0.34.0 hardening. **Target fix: v0.34.1.**

## Acceptance criteria

- [ ] Fix merged with tests covering gzip, deflate, identity (plain) and an over-cap compressed body (asserts `ResponseTooLargeError`, locking in that the running byte total measures compressed bytes after the switch); existing 204 / empty-body and plain-JSON cases stay green.
- [ ] v0.34.1 tagged.
- [ ] A gzip vendor read through an ingested vmware-rest connector succeeds on the patched instance.

## Found by

Found during a governed lifecycle rebuild readiness read on 2026-09-09 (fix-first decision).

Regression of #3459 (S09).

damir-topic:
## Problem

Starting in v0.34.0, every `HttpConnector`-based read against a vendor that
returns a `gzip`- or `deflate`-encoded response fails with:

```
httpx.DecodingError: Error -3 while decompressing data: incorrect header check
```

This breaks all generic ingested connectors and typed HTTP connectors
(vmware-rest, vcf-fleet, gcloud, vcd, vra8, loki, harbor, the argocd HTTP
path) whenever the vendor compresses its response. Vault (hvac, not an
`HttpConnector`) is unaffected. Reproduced live against a vCenter 8.x target
behind an ingested `vmware-rest` connector, and locally against httpx 0.28.1.

## Root cause

PR #3459 (the S09 body-size cap) introduced `_read_capped_json_response` in
`backend/src/meho_backplane/connectors/adapters/http.py`. Its cap loop
collected chunks via `resp.aiter_bytes()` — which yields **already-decompressed**
bytes — then rebuilt the response with:

```python
httpx.Response(status_code=..., headers=resp.headers, content=b"".join(chunks), ...)
```

keeping the upstream `Content-Encoding: gzip` header over a now-plaintext body.
httpx then attempts to decode a **second** time on first content access
(`resp.content` / `resp.json()`) → zlib `-3`. Both `_request_json` and
`_post_json` route through this helper, so every compressed vendor read broke.

## Fix

Iterate `resp.aiter_raw()` in the cap loop instead of `resp.aiter_bytes()`. The
buffered content is then the raw wire bytes that match the retained
`Content-Encoding` header, so the re-materialised response decodes exactly once
— byte-identical to the pre-#3459 `client.request()` path.

Cap semantics are preserved and made internally consistent: the running total
now counts **compressed** transfer bytes, matching the existing
`Content-Length` fast-reject (`_reject_oversize_content_length`), which already
measured the compressed length. No headers are stripped, no feature flag is
added, and the SSRF/destination-pinning transport is untouched. Docstrings on
the helper and the affected test updated to reflect `aiter_raw`.

## Tests

`backend/tests/test_connectors_http_adapter.py`, using the existing respx/httpx
fixture style:

- `test_request_json_gzip_encoded_response_decodes_once_regression_3459` — a
  `Content-Encoding: gzip` body (`gzip.compress(json)`) parses correctly through
  `_request_json` (the named regression guard for #3459).
- `test_post_json_gzip_encoded_response_decodes_once` — same via `_post_json`.
- `test_request_json_deflate_encoded_response_decodes_once` — a `deflate`
  (`zlib.compress`) body parses.
- `test_request_json_rejects_oversize_compressed_body_by_content_length` — the
  cap still rejects an oversize compressed body by its (compressed)
  `Content-Length` and quotes that compressed length.
- The oversize-streamed-body-without-`Content-Length` rejection and the plain
  (identity) under-cap parse remain covered by the pre-existing
  `test_request_json_rejects_oversize_response_without_content_length` and
  `test_{request,post}_json_under_cap_returns_parsed_payload` tests (docstring of
  the former updated from `aiter_bytes` to `aiter_raw`).

Commands (run from `backend/`, as CI does):

- `uv run pytest tests/test_connectors_http_adapter.py` → **99 passed** (was 95; +4).
- Full backend unit suite (`find tests -name 'test_*.py' -not -path 'tests/integration/*' -not -path 'tests/migrations/*'`, `uv run pytest -n auto --dist loadscope`) — green.
- `uv run ruff check src/ tests/` → clean; `uv run ruff format --check src/ tests/` → clean; `uv run mypy src/` → clean for the touched file.

## CHANGELOG entry (for the roll PR)

```
### Fixed
- Compressed (`gzip`/`deflate`) vendor responses no longer fail with `httpx.DecodingError` ("incorrect header check"): the dispatch body-size cap (#3459) now buffers raw wire bytes so the body decodes exactly once, restoring the pre-cap behaviour for every HttpConnector read. (#3521)
```

Closes #3521

### DIFF
--- a/backend/src/meho_backplane/connectors/adapters/http.py
+++ b/backend/src/meho_backplane/connectors/adapters/http.py
@@ -416,11 +416,22 @@ async def _read_capped_json_response(
     same-origin redirect and destination-pinning transport behaviour is
     unchanged) and enforces a two-part cap: a ``Content-Length`` fast-reject
     up front, then a running byte total across
-    :meth:`~httpx.Response.aiter_bytes` chunks that aborts the moment the
+    :meth:`~httpx.Response.aiter_raw` chunks that aborts the moment the
     body would exceed :data:`_MAX_RESPONSE_BYTES`. At most the cap (plus one
     trailing chunk) is ever held in memory, and an over-cap body raises
     :exc:`ResponseTooLargeError` before :func:`json_payload_or_empty` can
     parse it.
+
+    The cap iterates :meth:`~httpx.Response.aiter_raw` — the raw wire bytes,
+    still carrying whatever ``Content-Encoding`` the vendor applied — rather
+    than :meth:`~httpx.Response.aiter_bytes`, which yields already-decoded
+    bytes. Buffering the raw bytes keeps them consistent with the retained
+    ``Content-Encoding`` header, so the re-materialised response decodes
+    exactly once on first access (a plaintext body under a ``gzip`` header
+    would otherwise make httpx attempt a second decode and raise
+    :exc:`httpx.DecodingError`). The running total therefore counts
+    compressed transfer bytes, matching the ``Content-Length`` fast-reject
+    above, which measures that same compressed length.
 
     An under-cap body is joined and re-materialised into a fully-read
     :class:`httpx.Response` carrying the original status, headers, request,
@@ -441,7 +452,7 @@ async def _read_capped_json_response(
         _reject_oversize_content_length(resp)
         chunks: list[bytes] = []
         total = 0
-        async for chunk in resp.aiter_bytes():
+        async for chunk in resp.aiter_raw():
             total += len(chunk)
             if total > _MAX_RESPONSE_BYTES:
                 raise ResponseTooLargeError(
--- a/backend/tests/test_connectors_http_adapter.py
+++ b/backend/tests/test_connectors_http_adapter.py
@@ -53,11 +53,14 @@ Per-target base URL (evoila/meho#2873):
 from __future__ import annotations
 
 import datetime as _dt
+import gzip
+import json
 import socket as _socket
 import ssl
 import tempfile as _tempfile
 import threading as _threading
 import types
+import zlib
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 from typing import Any
 from unittest.mock import patch
@@ -695,10 +698,9 @@ async def test_request_json_rejects_oversize_response_without_content_length(
 ) -> None:
     """An oversize body with no ``Content-Length`` is caught by the streamed total.
 
-    An absent (or understated / content-encoded) length header cannot smuggle
-    an oversize body past the guard: the running byte total across
-    ``aiter_bytes`` chunks aborts once the cap is exceeded, before the whole
-    body is buffered.
+    An absent (or understated) length header cannot smuggle an oversize body
+    past the guard: the running byte total across ``aiter_raw`` chunks aborts
+    once the cap is exceeded, before the whole body is buffered.
     """
     monkeypatch.setattr(http_adapter, "_MAX_RESPONSE_BYTES", 1024)
     conn = _ConcreteHttpConnector()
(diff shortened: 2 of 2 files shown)


damir-topic:
## Review of record (independent, orchestrator meho-78)

**Verdict:** APPROVE

PR #3524 · branch `fix/http-cap-double-decode` · reviewed head `d29086e16571f409188317fd0c02c2365db60a04` · base `main` · closes #3521.
Read-only review from a throwaway worktree at the PR head. No GitHub review approval submitted, nothing merged/pushed (App approval is the orchestrator's).

The one-line functional change — `resp.aiter_bytes()` → `resp.aiter_raw()` in `_read_capped_json_response` (`backend/src/meho_backplane/connectors/adapters/http.py:455`) — is correct, restores the pre-#3459 behaviour for compressed vendor responses, and is proven by test to fix exactly the production regression. Scope, hygiene, lint, and typecheck are clean. One non-blocking hardening observation on cap semantics (SHOULD-FIX / follow-up), consistent with the locked brief.

---

### A. Correctness — CONFIRMED

- **Loop consumes raw wire bytes.** `async for chunk in resp.aiter_raw():` (`http.py:455`) inside `client.stream(...)`. In httpx 0.28.1 `aiter_raw` yields the untouched transport stream, marks `is_stream_consumed`, and closes the response on exhaustion (`.venv/.../httpx/_models.py:1037-1065`). The buffered `chunks` are therefore the compressed bytes still matching the retained `Content-Encoding` header.
- **Rebuilt Response decodes exactly once.** `httpx.Response(status_code=..., headers=resp.headers, content=b"".join(chunks), ...)` (`http.py:463-470`). httpx `Response.__init__` with `content=` bytes calls `_prepare(headers)` (which uses `setdefault`, so the retained headers win) and then eagerly `self.read()` because the body is a `ByteStream` (`_models.py:516-555`). `read()` → `iter_bytes()` → `_get_content_decoder()` builds the decoder from the `Content-Encoding` header (`_models.py:699-721`, `876-909`) and decompresses **once** into `_content`. Root cause reproduced: the old `aiter_bytes()` path buffered already-decoded plaintext, so this second decode of plaintext-under-a-gzip-header raised `DecodingError` — matching the reported symptom.
- **No `Content-Encoding` → byte-identical to before.** With no encoding header `_get_content_decoder()` returns `IdentityDecoder` (`_models.py:719`), a pass-through; for an identity body `aiter_raw` and `aiter_bytes` yield the same bytes, so the reconstructed content and the parsed payload are unchanged. Verified by the pre-existing identity under-cap and oversize tests still passing.
- **Both dispatch paths covered.** `_request_json` (`http.py:932`) and `_post_json` (`http.py:1040`) both call `_read_capped_json_response`; the new tests exercise both.
- **Downstream decodes once.** `resp.raise_for_status()` then `json_payload_or_empty(resp)` (`http.py:309-332`) reads `resp.content`/`resp.json()`, which return the already-populated `_content`; no re-decode. `204`/empty still maps to `{}`.
- **Content-Length coherence.** The rebuilt content is the compressed bytes; the retained `Content-Length` (when present) is the compressed length; `encode_response` computes the same length, so `_prepare`'s `setdefault` is a no-op and there is no mismatch for httpx to reject or truncate.

### B. Cap semantics — CONFIRMED (one non-blocking hardening note)

- Running total now counts **compressed** transfer bytes (raw chunks), consistent with the `Content-Length` fast-reject in `_reject_oversize_content_length` (`http.py:373-393`), which always measured the compressed length. Internally consistent now; it was not before.
- Oversize rejection type/message unchanged (`ResponseTooLargeError`, same two message shapes). Still a plain `Exception`, so still excluded from `_retryable` and still flattens to `connector_error` (covered by `test_oversize_error_is_connector_error_shaped_and_not_retryable`, `test_oversize_response_not_retried`).
- **SHOULD-FIX (hardening, non-blocking, follow-up):** the cap now bounds only the *compressed* transfer size. The rebuilt `Response.read()` decompresses the whole body into `_content` with **no** cap, so a decompression bomb (small compressed → very large decoded) passes both the compressed `Content-Length` fast-reject and the compressed streamed total, then fully materialises in memory — precisely the "compromised/malicious appliance" threat the #297 cap was built to bound. This is a deliberate tradeoff per the fix brief (compressed-transfer semantics, matching the fast-reject) and is strictly better than the pre-#3459 unbounded state — and note #3459's nominal decoded-size bound never actually functioned for compressed bodies (it always double-decoded and errored), so no working guarantee is being lost. Recommend a follow-up ticket to also bound the decoded size (e.g. a decoded-byte cap during re-materialisation). NOT blocking this hotfix: the regression it fixes (every compressed vendor response failing — vCenter gzips everything) is far more severe. The PR body states cap semantics are "preserved and made internally consistent" but does not mention the now-unbounded decoded size; worth a line in the follow-up.

### C. Tests — CONFIRMED (regression guards proven to fail on broken code)

- The three decode guards send real `Content-Encoding: gzip`/`deflate` with `gzip.compress`/`zlib.compress` bodies through the respx/httpx fixture and assert the parsed JSON, across both `_request_json` and `_post_json`.
- **Proven to fail on the broken code:** reverting the single functional change (`aiter_raw` → `aiter_bytes`) in the worktree and re-running the new tests fails all three with `httpx.DecodingError: Error -3 while decompressing data: ...` — the exact production symptom. These are genuine regression guards, not tests that pass on the bug.
- The fourth new test (`..._rejects_oversize_compressed_body_by_content_length`) asserts cap-measures-compressed semantics via the `Content-Length` fast-reject; that path is unchanged by the fix, so it passes on both old and new code (correct — it is a cap-semantics assertion, not a double-decode guard). Minor: it alone would not have caught #3459; tests 1–3 do.
- The streamed-total oversize test still exercises the cap after the change (`ByteStream`, no `Content-Length`, trips the running total via `aiter_raw`) — passes. Its docstring was correctly updated `aiter_bytes` → `aiter_raw`.

### D. Scope + hygiene — CONFIRMED

- Only `backend/src/meho_backplane/connectors/adapters/http.py` (+15/−6) and `backend/tests/test_connectors_http_adapter.py` (+136) changed. No `CHANGELOG.md` change (correctly deferred to the roll PR; a suggested entry lives in the PR body only). CLI OpenAPI snapshot / docs untouched — no regen needed for a transport-internal change.
- DCO: the single commit `d29086e1` carries `Signed-off-by: Damir Topic <...>` plus `Co-Authored-By: Claude Fable 5.1`.
- Leak-free: only `*.example.com` documentation domains, generic identifiers (`vm-101`, `web-01`, `obj-1`, `session-token-shape`), synthetic all-zero/all-one UUIDs. Commit message and PR body reference only issue numbers and a generic "vCenter 8.x" product string — no lab hostnames/IPs/env names/people.
- `uv run ruff check src/ tests/` → clean; `uv run ruff format --check src/ tests/` → clean; `uv run mypy src/` → the only error is in **unchanged** `connectors/net/icmp.py:205` (`socket.MSG_ERRQUEUE`, a Linux-only constant absent on the macOS review host) — a local-platform artifact CI's Linux runners do not hit; the changed files are mypy-clean.

### E. Test run (CI-equivalent, from `backend/`)

Environment: httpx 0.28.1, Python 3.14, `MEHO_SKIP_SPEC_INGEST_TESTS=1`, `MEHO_REDACTION_SPACY_MODEL=en_core_web_sm`, `uv run pytest -n 3 --dist loadscope` (the per-PR unit-lane invocation; reconcile/full-ingest lanes opted out as CI does).

| Scope | Command | Result |
|---|---|---|
| Changed test module | `pytest ... tests/test_connectors_http_adapter.py` | **99 passed** (was 95; +4) |
| Direct blast-radius callers | `+ vmware_rest_library_download + ovf_import` | **132 passed** |
| Full connectors suite (235 files, minus 6 raw-network `net_*`) | `pytest ... test_connectors*.py ! net_*` | **5761 passed, 64 skipped, 0 failed** |
| `net_*` family isolated | 6 files | 156 passed, **2 failed** |
| Regression proof on broken code | 3 decode tests with `aiter_raw`→`aiter_bytes` reverted | **3 failed** with `httpx.DecodingError` (expected) |
| Lint / format / type | `ruff check src/ tests/` · `ruff format --check src/ tests/` · `mypy src/` | clean / clean / 1 unrelated macOS-only error |

The 2 `net_*` failures (`test_connectors_net_dns_lookup.py::test_chosen_resolver_queries_that_nameserver`, `test_connectors_net_icmp.py::test_trace_reaches_loopback`) are raw-DNS/raw-ICMP-socket tests that depend on the local network/OS; both files are unchanged by this PR and do not use the HttpConnector transport — pre-existing macOS-local environment artifacts, not regressions.

Note on scope: I did not run the entire ~13k-test backend suite across all 3 shards (full Postgres testcontainers + embedding warm + spaCy pipeline, ~18 min); I ran the complete connector blast radius (every HttpConnector subclass's tests) plus the changed module and the lint/type gates, which is the relevant surface for a shared-transport one-line change. CI runs the full suite on the PR.

### BLOCKING

None.

PR #3072, opened 2026-08-20
fix(ui): wire both broadcast filter selects, not just the first (#229)

smujkic94:
## Summary
- The broadcast live-feed **Target** filter did nothing when changed. The filter bar used `hx-trigger="change from:find select, …"`, and htmx's `from:find <sel>` binds only the **first** matching descendant — so only the **Op class** select triggered the refilter.
- Fixed by enumerating both selects by name: `change from:find select[name=op_class], change from:find select[name=target], input changed delay:300ms from:find input[name=principal]`.
- The `op_id` input keeps **no** server-trigger clause (it's deliberately client-side, excluded from `hx-include`), so the fix restores the Target filter without churning the SSE subscription on op_id blur. Same bug class as the connector-registry filter fix (#174).

## Test plan
- [x] `ruff check` — clean
- [x] `pytest tests/test_ui_broadcast_filters.py tests/test_ui_broadcast_feed.py` — 41 passed
- [x] **Target select refilters** — render-guard asserts `change from:find select[name=target]` is present and the old first-select-only `from:find select,` trigger is gone
- [x] **Op class still refilters** — `change from:find select[name=op_class]` present
- [x] **No regression** — principal debounced input + client-side op_id narrowing covered by the 41 passing tests

## Out of scope
- The filter bar's dead daisyUI-v4 `form-control` / `label-text` classes (cosmetic; noted on the issue as a follow-up).
- Sibling broadcast-view tasks evoila-bosnia/meho-internal#230 (event drawer) and #231 (overrides dialog) under initiative #228.

## Issue
Implements **evoila-bosnia/meho-internal#229** (subtask of initiative **evoila-bosnia/meho-internal#228**). Closing keywords don't cross repositories, so the reference below is a traceability link — the issue is closed manually on merge.

Closes evoila-bosnia/meho-internal#229

PR #335, opened 2026-05-12
feat(broadcast): async Valkey client + lifespan + readiness probe (G6.1-T1)

zdamir:
## Summary

- Lands the G6.1-T1 broadcast substrate: async `redis.asyncio.Redis`
  singleton (`broadcast/client.py`), lifespan-managed open/dispose, and a
  `broadcast_readiness_probe` registered against the existing health
  registry so `/ready` reflects Valkey availability. T2-T6 (#308-#312)
  build the event schema, publish-on-write, SSE, MCP resource, and
  `meho status --watch` on top.
- `Settings` adds `BROADCAST_REDIS_URL` (default `redis://localhost:6379`;
  only `redis://`/`rediss://`/`unix://` accepted — `valkey://` rejected
  because redis-py rejects it at parse time) and `BROADCAST_RETENTION_HOURS`
  (default 24). Validators run at construction so misconfigured deploys
  fail fast at startup, not on the first `/ready` poll.
- Probe never echoes the operator-supplied URL or port; failure shapes
  collapse to four structured detail strings (`reachable`, `timeout`,
  `unreachable: <ExcClass>`, `redis_error: <ExcClass>`) plus a safety-net
  `check_failed: <ExcClass>` branch — same redaction contract `auth/vault`
  and `db/migrations` enforce.

## Test plan

- [x] `ruff check` — clean on touched files.
- [x] `ruff format --check` — clean on touched files.
- [x] `mypy src/` — clean (47 source files, strict mode).
- [x] `pytest tests/` — 426 passed, 1 skipped (the slow-lane fastembed test),
      1 deselected (pre-existing Darwin-only `process_resident_memory_bytes`
      gap in `test_observability.py` — failure reproduces on `origin/main`
      and is unrelated).
- [x] `python3 .claude/hooks/code-quality.py --diff` — 0 blockers, 3
      length warnings on pre-existing large functions plus the new probe
      whose 61-line body is mostly docstring.
- [x] AC #1: `get_broadcast_client()` returns a singleton —
      `TestClientLifecycle::test_singleton_returns_same_instance`.
- [x] AC #2: lifespan opens + disposes — `dispose` idempotency tests +
      `test_connectors_registry::test_lifespan_calls_eager_import_connectors`
      patches the new helpers.
- [x] AC #3: probe `ok=True` against reachable Valkey, `ok=False`
      against unreachable — `TestBroadcastIntegration` (Docker-gated)
      runs against `valkey/valkey:8` via testcontainers; class-level
[...]

PR #3309, opened 2026-09-02
fix(ui): structure conventions "New convention" dialog (#240)

smujkic94:
## Summary
- The Conventions "New convention" create dialog (`_create_modal.html`) used daisyUI v4 `form-control` + `label-text` on all five fields (Slug, Kind, Title, Priority, Body), both removed in v5 (zero compiled rules), so the label-over-input column collapsed and the two-column Slug/Kind and Body/token-preview grids read misaligned (the Priority label detached from its input, the token-cost preview floated).
- Migrated every field to `flex flex-col gap-1` label wrappers, `text-sm font-medium` labels, helper text as `text-xs opacity-60` blocks below, and `w-full` controls. The `max-w-[12rem]` Priority cap and both `grid md:grid-cols-2` layouts are preserved.
- Presentation only — the create POST, CSRF double-submit, and the debounced token-cost preview (the Kind select's + Body textarea's `hx-post` to `/ui/conventions/preview`) are unchanged.

## Test plan
- [x] `ruff check` — clean
- [x] `pytest tests/test_ui_conventions_write.py` — 16 passed
- [x] **Aligned dialog** — all 5 fields stack label-over-input via `flex flex-col gap-1`; grids + Priority cap preserved
- [x] **No dead classes** — render-guard asserts `form-control`/`label-text` absent from `/ui/conventions/create`
- [x] **No functional regression** — create POST, CSRF, and both preview `hx-post` wirings covered by the 16 passing tests

## Out of scope
- The detail-view edit modal (`_edit_modal.html`) — sibling task evoila-bosnia/meho-internal#241 under initiative #239.

## Issue
Implements **evoila-bosnia/meho-internal#240** (subtask of initiative **evoila-bosnia/meho-internal#239**). Closing keywords don't cross repositories, so the reference below is a traceability link — the issue is closed manually on merge.

Closes evoila-bosnia/meho-internal#240

PR #3108, opened 2026-08-24
fix(ui): structure broadcast overrides "New rule" dialog (#231)

smujkic94:
## Summary
- The "New rule" create dialog on the Broadcast **Overrides** tab (`_overrides.html`) used daisyUI v4 `form-control` + `label-text` on all four fields (Op id pattern, Scope, Scope value, Detail level), both removed in v5 (zero compiled rules), so the label-over-input column collapsed and the form read cramped.
- Migrated every field to `flex flex-col gap-1` label wrappers, `text-sm font-medium` labels, helper text as `text-xs opacity-60` blocks below, and `w-full` controls. The Scope / Scope-value pair keeps its `grid md:grid-cols-2` layout.
- Presentation only — the create POST, CSRF double-submit, and the scope-value visibility toggle (keys off the untouched `data-action="scope-select"` / `data-scope-value-row` hooks) are unchanged. Clears the **last** dead daisyUI class on the broadcast surface (round 6, #228).

## Test plan
- [x] `ruff check` — clean
- [x] `pytest tests/test_ui_broadcast_overrides.py` — 6 passed
- [x] **Aligned dialog** — all 4 fields stack label-over-input via `flex flex-col gap-1`; scope grid preserved
- [x] **No dead classes** — render-guard asserts `form-control`/`label-text` absent from the tenant-admin overrides fragment
- [x] **No functional regression** — create POST, CSRF, and the scope-value toggle (guard asserts `data-action="scope-select"` / `data-scope-value-row` present) covered by the 6 passing tests

## Out of scope
- The overrides *table* timestamp (`_overrides.html`, still raw ISO) — separate follow-up if desired; not part of the "New rule dialog" scope.
- Sibling round-6 tasks #229 (filters, PR #3072) and #230 (drawer, PR #3107) under initiative #228.

## Issue
Implements **evoila-bosnia/meho-internal#231** (subtask of initiative **evoila-bosnia/meho-internal#228**). Closing keywords don't cross repositories, so the reference below is a traceability link — the issue is closed manually on merge.

Closes evoila-bosnia/meho-internal#231

PR #3527, opened 2026-09-10
fix(ui): structure scheduler create-trigger modal + fix timezone re-validation (#335)

smujkic94:
## Summary

- Migrates the Scheduler create-trigger modal (`scheduler/_create_modal.html`) off the daisyUI v4 dead classes (`form-control` / `label-text` / `label-text-alt` — zero compiled rules in v5) to the established pattern: `flex flex-col gap-1` label wrappers, `text-sm font-medium` labels, hints as `text-xs opacity-60`, `w-full` controls, inline qualifiers as `opacity-60 font-normal` spans.
- Fixes a dead htmx binding: `change from:find input[name=timezone]` on the cron input never bound (`from:find` searches descendants of the element carrying the trigger, and an `<input>` has none), so editing the timezone silently kept a stale `next_fire_at` preview. The timezone input now carries its own `hx-post="/ui/scheduler/validate-cron"` wiring (`hx-trigger="change"`, same target/include) — same bug class and fix shape as the broadcast filter bar (#3072).
- Untouched: Alpine `x-data` kind switch, the `data-fire-at` local→UTC conversion script (#2339), form-level CSRF `hx-headers`, `hx-disabled-elt`/`hx-disinherit`, `data-action` hooks.

## Test plan

- [x] `pytest tests/test_ui_scheduler.py` — 32 passed
- [x] Render guards added to `test_create_modal_renders_for_tenant_admin`: no `form-control` / `label-text` / `from:find` in the rendered modal; exactly 2 `validate-cron` posts wired (cron input + timezone input)
- [x] `grep -n "form-control\|label-text" scheduler/_create_modal.html` — clean

## References

Task: Closes evoila-bosnia/meho-internal#335 (parent initiative: evoila-bosnia/meho-internal#337, round 8 — Scheduler; sibling: evoila-bosnia/meho-internal#336). Cross-repo `Closes` does not auto-close — close #335 manually on merge.

Precedent migrations: #2906, #3108, #3309; precedent `from:find` fix: #3072.

### DIFF
--- a/backend/src/meho_backplane/ui/templates/scheduler/_create_modal.html
+++ b/backend/src/meho_backplane/ui/templates/scheduler/_create_modal.html
@@ -22,11 +22,11 @@
   kind's discriminator field is meaningful; the server populates only that
   field on the wire schema regardless of any stale hidden values.
 
-  Live cron validation: the cron <input> ``hx-post``s to
-  ``/ui/scheduler/validate-cron`` (debounced) and swaps the
-  ``#scheduler-cron-preview`` fragment, so the operator sees validity + a
-  ``next_fire_at`` preview before submit -- no free-text cron with no
-  feedback.
+  Live cron validation: the cron <input> (debounced) and the timezone
+  <input> (on change) each ``hx-post`` to ``/ui/scheduler/validate-cron``
+  and swap the ``#scheduler-cron-preview`` fragment, so the operator sees
+  validity + a ``next_fire_at`` preview before submit -- no free-text cron
+  with no feedback, and no stale preview after a timezone edit (#335).
 
   CSRF: the form carries its own ``hx-headers`` echo of the token (HTMX
   does not inherit ``hx-headers`` to children injected into a swapped
@@ -65,11 +65,11 @@
       x-data="{ kind: 'cron' }"
     >
       {# ---------- Agent definition ---------- #}
-      <label class="form-control">
-        <span class="label-text">Agent definition</span>
+      <label class="flex flex-col gap-1">
+        <span class="text-sm font-medium">Agent definition</span>
         <select
           name="agent_definition_id"
-          class="select select-bordered select-sm"
+          class="select select-bordered select-sm w-full"
           required
           aria-label="Agent definition"
         >
@@ -83,11 +83,11 @@
       </label>
 
       {# ---------- Kind switch ---------- #}
-      <label class="form-control">
-        <span class="label-text">Kind</span>
+      <label class="flex flex-col gap-1">
+        <span class="text-sm font-medium">Kind</span>
         <select
           name="kind"
-          class="select select-bordered select-sm"
+          class="select select-bordered select-sm w-full"
           x-model="kind"
           aria-label="Trigger kind"
         >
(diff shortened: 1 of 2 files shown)


smujkic94:
@coderabbitai review

smujkic94:

`/auto-review-pr` — autonomous review at head `57b6358f`, run in an isolated worktree (`.claude/worktrees/pr-3527-review`) checked out at the PR head SHA. Posted as a comment because GitHub blocks self-approval on one's own PR.

### Scope check

Two files, both in the task's declared scope (evoila-bosnia/meho-internal#335) — no strays:

- `backend/src/meho_backplane/ui/templates/scheduler/_create_modal.html`
- `backend/tests/test_ui_scheduler.py`

### Verification (isolated worktree at `57b6358f`)

| Gate | Result |
|---|---|
| `ruff format --check` | clean (1 file already formatted) |
| `pytest tests/test_ui_scheduler.py tests/test_ui_scheduler_views.py` | **42 passed** |
| Import provenance | confirmed `meho_backplane` loaded from the worktree, not the main checkout |

CI on this SHA: **all three Python unit shards SUCCESS**, `Python lint + typecheck (ruff + mypy)` SUCCESS, DCO SUCCESS, CodeQL / Semgrep / Semgrep OSS / TruffleHog SUCCESS, Helm + Go + CLI snapshot SUCCESS. Zero failed checks. Still in flight at the time of writing: `Python (integration testcontainers)` and the backplane image build — neither touches this diff's surface (a Jinja template and a render test), but a merger should confirm both land green.

### Acceptance criteria (evoila-bosnia/meho-internal#335)

- [x] **No `form-control` / `label-text` / `label-text-alt` in the rendered modal** — all nine fields migrated; verified by grep on the template and by the new render guards, which assert on the *rendered* body rather than the source.
- [x] **Controls fill the modal width, labels above with consistent spacing** — `flex flex-col gap-1` wrappers + `text-sm font-medium` labels + `w-full` on every select/input/textarea inside a column wrapper. The inline radio labels in the in-flight-policy fieldset correctly do *not* get `w-full` (they are `flex items-center gap-2` rows), matching the pattern from the memory/broadcast/conventions rounds.
- [x] **Timezone edit re-runs `/ui/scheduler/validate-cron`** — the dead `change from:find input[name=timezone]` clause is removed from the cron input and the timezone input carries its own wiring.
- [x] **Kind switch, cron live validation, fire-at UTC conversion, CSRF, submit-disable unchanged** — `x-data`/`x-show`/`x-cloak`, `data-fire-at` + `htmx:configRequest` + `toISOString` (#2339), form-level `hx-headers`, and `hx-disinherit="hx-disabled-elt"` (#2340) are all untouched and still covered by their pre-existing tests, which pass.
- [x] **Render guards added; ruff + pytest green** — see table above.

### Adversarial pass on the functional fix

The interesting part of this PR is the htmx rewiring, so I verified it structurally rather than trusting the diff:

1. **The old binding really was dead.** `from:find <sel>` resolves against *descendants* of the element carrying the trigger. That trigger sat on the cron `<input>`, which has no descendants, so the clause bound nothing and a timezone edit left a stale `next_fire_at`. Correct diagnosis, and the same bug class as the broadcast filter bar (#3072).
2. **The new binding reaches the route with the parameters it requires.** Both fields' `<label>` wrappers sit directly inside the single `<div x-show="kind === 'cron'">` (template lines 101–143), so `hx-include="closest div"` from *either* input resolves to that same container, which holds both `cron_expr` and `timezone`. That satisfies the route signature `cron_expr: str = Form(default="")`, `timezone: str = Form(default="UTC")` — the timezone-triggered POST is not missing the cron expression.
3. **CSRF still lines up.** The new `hx-post` is a descendant of the form, so it inherits the form-level `hx-headers` CSRF echo; `hx-disinherit="hx-disabled-elt"` keeps it from inheriting `find button[type=submit]` and logging `returned no matches!` (the #2340 regression) — same posture the cron input already had.
4. **No spurious requests when kind ≠ cron.** The container is `x-show`-hidden, and `change` only fires on user interaction, so a hidden timezone field cannot trigger a validate-cron POST.
5. **Bare `<fieldset>` is safe.** Tailwind preflight zeroes the UA border/padding, and daisyUI v5's `fieldset` is an opt-in class, not a bare-element style — so dropping `form-control` there does not reintroduce a browser default border.

### Notes (non-blocking)

- `assert body.count('hx-post="/ui/scheduler/validate-cron"') == 2` is a deliberately tight guard: a future third wiring will fail the test. That is the intended behavior for a regression guard on exactly this bug, and the accompanying comment explains why — worth keeping, worth updating consciously.
- Pre-existing, not introduced here: every control carries an `aria-label` that duplicates its visible label text, so assistive tech announces the `aria-label` and not the newly-styled `<span>`. The strings match, so there is no user-facing defect; flagging only so a future accessibility pass on this surface knows the visible labels are not the accessible names.

### Bot review

CodeRabbit produced no review, inline comments, or issue comments on this PR across a bounded ~7-minute poll after the `@coderabbitai review` trigger (`bot_review.received: false`) — consistent with its extended unavailability across this initiative's other PRs. Per policy the verdict proceeds on the adversarial pass plus the local and CI gates above.

### Verdict

**APPROVE** — the change is correctly scoped, the cosmetic migration follows the pattern established in rounds 5–7, and the functional fix is a real bug correctly diagnosed and correctly wired. Close evoila-bosnia/meho-internal#335 manually on merge (cross-repo `Closes` does not auto-close).

damir-topic (approved):
## Independent review — PR #3527

**Verdict: APPROVE.** Reviewed commit `57b6358f449bbe38246c5c396a3307deec1d4e5d` against evoila-bosnia/meho-internal#335 and its Scheduler initiative. No code findings.

### Validation
- pytest: 42 passed: backend/tests/test_ui_scheduler.py backend/tests/test_ui_scheduler_views.py
- ruff: pass
- code_quality_files: pass
- code_quality_diff: pass
- git_diff_check: pass

docs/codebase/ui.md accurately describes the scheduler create and validate-cron routes; unchanged

Sources: https://htmx.org/attributes/hx-trigger/, https://htmx.org/attributes/hx-include/, https://alpinejs.dev/directives/show

Independent Terra review relayed by the orchestrator on reviewed commit `57b6358f449bbe38246c5c396a3307deec1d4e5d`. The review App was unavailable; this formal review uses the maintainer identity, distinct from the PR author. Merge remains subject to required CI and the normal merge queue.

PR #3107, opened 2026-08-24
fix(ui): humanize broadcast drawer labels + format timestamp (#230)

smujkic94:
## Summary
- The broadcast event-detail drawer (`_event_drawer.html`) showed raw snake_case machine keys as detail-row labels (`operator`, `occurred_at`, `audit_id`, `request_id`, `event_id`, …) and rendered `occurred_at` as a bare `.isoformat()` string.
- Humanized every `<dt>` label (Operator / Method / Status / Occurred / Duration (ms) / Audit ID / Request ID / Event ID) and rendered `occurred_at` in the console-standard `%Y-%m-%d %H:%M UTC` display format with the full ISO in a `<time datetime>` attribute — mirroring `topology/_drawer.html`.
- Presentation only — open-on-click, the aggregate-only payload gate, identifiers shown, and the "Suppress this op" cross-link are unchanged.

## Test plan
- [x] `ruff check` — clean
- [x] `pytest tests/test_ui_broadcast_filters.py` — 24 passed
- [x] **Human labels** — guard asserts each humanized `<dt>` label present and the raw `occurred_at`/`audit_id` keys absent
- [x] **Timestamp** — guard asserts `<time datetime=` + `UTC</time>` (formatted, not a bare isoformat dump)
- [x] **No regression** — drawer open/dismiss, aggregate gate, suppress cross-link covered by the 24 passing tests

## Out of scope
- The overrides *table* timestamp (`_overrides.html`, also raw ISO) — folded into sibling task evoila-bosnia/meho-internal#231 if low-risk.
- Sibling round-6 tasks #229 (filters, PR #3072) and #231 (overrides dialog) under initiative #228.

## Issue
Implements **evoila-bosnia/meho-internal#230** (subtask of initiative **evoila-bosnia/meho-internal#228**). Closing keywords don't cross repositories, so the reference below is a traceability link — the issue is closed manually on merge.

Closes evoila-bosnia/meho-internal#230

Issue #336, opened 2026-05-12
Goal: G10 Operator web UI — server-rendered HTMX 2 + Jinja2 + Tailwind 4 + DaisyUI + Alpine.js + Cytoscape.js island, mounted at /ui/* on the backplane; broadcast / kb / topology / connectors / memory surfaces


damir-topic:
## Goal

Ship a **web UI** (Operator Console) over the MEHO backplane API — a server-rendered console mounted at `/ui/*` on the backplane FastAPI app that gives operators visual insight into the surfaces that today only have CLI + MCP access: the **knowledge base** (search + view + upload), the **topology** (tabular + graph), the **connectors + targets**, the **memory**, and especially the **activity broadcast** (live SSE feed + Slack-mirror equivalent).

The UI is the **third operator surface** alongside the CLI (`meho ...`) and MCP (agent-native). It targets human operators who want at-a-glance visibility: the team's broadcast feed, the kb's content, what's in topology, who touched what.

## Why

CLI is great for individual ops and scripting; MCP is great for agent workflows. **Neither is great for situational awareness** — "what's everyone doing right now," "what's in our kb," "what does the topology look like." The activity broadcast in particular (G6) reaches its full UX potential only with a persistent visual surface — `meho status --watch` in a terminal is fine for one operator, but a shared web view on a wall monitor + a tab on every operator's browser is the team coordination layer the consumer explicitly framed in consumer-needs.md §G6 L184 ("Slack is the team's shared UX") + L186-187 (threaded contexts post-v0.2).

The web UI is also the **operator-side UX for write surfaces** that are awkward via CLI:
- **Uploading docs to the kb** — drag-and-drop a Markdown file beats `meho kb write` with a `--body @path/to/file.md` form.
- **Editing tenant conventions** — a Markdown editor beats `meho conventions edit` opening `$EDITOR`.
- **Browsing topology** — a graph viz beats `meho topology dependents --json | jq`.

Locked decisions ([#9-#11 in v0.2-decisions.md](docs/planning/v0.2-decisions.md)): HTMX 2 + Jinja2 + Tailwind 4 + DaisyUI 5 + Alpine.js (+ Cytoscape.js island for topology); server-rendered from the backplane FastAPI process at `/ui/*` (no separate FE deploy artifact); Backend-for-Frontend auth with `HttpOnly; Secure; SameSite=Strict` session cookie against the same Keycloak (tokens stay server-side).

## Done-when (testable from the consumer)

[...]

PR #3528, opened 2026-09-10
fix(ui): humanize scheduler detail-view timestamps (#336)

smujkic94:
## Summary

- The scheduler detail page rendered all six of its instants as raw `isoformat()` output (`2026-06-18T09:30:12.483920+00:00`) — hard to scan and inconsistent with the rest of the console. All six now follow the established `<time>` convention (precedent `topology/_drawer.html:178`, broadcast drawer #3107): the ISO instant stays machine-readable in the `datetime` attribute, the visible text is `YYYY-MM-DD HH:MM UTC`.
- The **"UTC" label is verified, not assumed**: every one of these six values already passes through `coerce_utc_aware()` in `_build_context` ([detail.py:122–140](https://github.com/evoila/meho/blob/main/backend/src/meho_backplane/ui/routes/scheduler/detail.py#L122-L140)), so the suffix is accurate on both the PostgreSQL tz-aware path and the SQLite naive round-trip.
- `font-mono` dropped from the humanized cells (the formatted string is no longer a machine literal), kept on genuinely machine-shaped values — cron expression, ids, subs, work_ref. Empty states (`—`, `— (never)`, `— (no future fire)`) unchanged.
- The existing fixture only populated three of the six timestamps, so `_seed_trigger` grew optional fired/skip parameters and a new test renders a skipping `one_off` trigger to cover the other three.

## Test plan

- [x] `pytest tests/test_ui_scheduler.py tests/test_ui_scheduler_views.py` — **43 passed** (42 before, +1 new)
- [x] All six timestamps humanized — `fire_at`, `next_fire_at`, `last_fired_at`, `last_skipped_at`, `created_at`, `updated_at`; the three previously-unexercised ones are pinned to exact strings (`2026-06-18 09:30 UTC</time>` etc.) by the new test
- [x] Empty-state fallbacks preserved — `assert "(never)" in body`
- [x] No raw isoformat as visible text — `assert "+00:00</" not in body` (the ISO instant may only appear inside a `datetime` attribute); microsecond precision asserted absent from the visible text
- [x] `grep isoformat detail.html` — all six occurrences are inside `datetime="…"` attributes

## Out of scope

- Sibling task evoila-bosnia/meho-internal#335 (create-trigger modal form layout + timezone re-validation) — PR #3527.
- `coerce_utc_aware` passes an already-aware non-UTC datetime through unchanged rather than converting it. Not reachable today (the driver returns UTC), and the pre-existing contract is pinned by `test_coerce_utc_aware_passes_aware_through` — noted, not changed.

## References

Closes https://github.com/evoila-bosnia/meho-internal/issues/336

Parent initiative: evoila-bosnia/meho-internal#337 (round 8 — Scheduler).

### DIFF
--- a/backend/src/meho_backplane/ui/templates/scheduler/detail.html
+++ b/backend/src/meho_backplane/ui/templates/scheduler/detail.html
@@ -11,6 +11,14 @@
 
   The cancel button (tenant_admin-only, hidden on a terminal trigger)
   HTMX-loads the terminal-confirm dialog into the modal container.
+
+  Timestamps follow the console-wide ``<time>`` convention (precedent
+  ``topology/_drawer.html``): the machine-readable ISO instant lives in
+  the ``datetime`` attribute, the visible text is
+  ``YYYY-MM-DD HH:MM UTC`` (#336). Every instant reaching this template
+  has already passed through ``coerce_utc_aware`` in the handler, so the
+  "UTC" suffix is accurate on both the PostgreSQL (tz-aware) and SQLite
+  (naive round-trip) paths.
 -#}
 {% extends "base.html" %}
 
@@ -66,7 +74,7 @@
             <dd>{{ trigger.timezone }}</dd>
           {% elif trigger.kind == "one_off" %}
             <dt class="opacity-60">Fire at</dt>
-            <dd class="font-mono">{% if trigger.fire_at %}{{ trigger.fire_at.isoformat() }}{% else %}—{% endif %}</dd>
+            <dd>{% if trigger.fire_at %}<time datetime="{{ trigger.fire_at.isoformat() }}">{{ trigger.fire_at.strftime("%Y-%m-%d %H:%M UTC") }}</time>{% else %}—{% endif %}</dd>
           {% else %}
             <dt class="opacity-60">Event filter</dt>
             <dd>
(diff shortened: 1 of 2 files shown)


smujkic94:
@coderabbitai review

damir-topic (approved):
## Independent review — PR #3528

**Verdict: APPROVE.** Reviewed commit `96ec2622f1e13cee5393975f9226c1526b5bf243` against evoila-bosnia/meho-internal#336 and its Scheduler initiative. No code findings.

### Validation
- pytest: 43 passed: backend/tests/test_ui_scheduler.py backend/tests/test_ui_scheduler_views.py
- ruff: pass
- code_quality_files: pass
- code_quality_diff: pass
- git_diff_check: pass

docs/codebase/ui.md accurately describes the scheduler detail route; unchanged

Sources: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time

The PR body was corrected to close only evoila-bosnia/meho-internal#336; GitHub closingIssuesReferences was rechecked. The reviewed source commit is unchanged.

Independent Terra review relayed by the orchestrator on reviewed commit `96ec2622f1e13cee5393975f9226c1526b5bf243`. The review App was unavailable; this formal review uses the maintainer identity, distinct from the PR author. Merge remains subject to required CI and the normal merge queue.

Issue #3537, opened 2026-09-11
approvals: credential_write parks persist the raw secret value on the approval row — indirect it via a one-time execution handle


damir-topic:
Parent initiative: #3655

## Summary

When a `credential_write` operation (e.g. `vault.kv.put`, and other ops that
write secret material) parks for approval, its **params — including the secret
VALUE — persist verbatim on the approval / park row** so the op can execute
after approval. The proposed_effect **preview** is already redaction-safe
(#2332: approver sees path + key names, never values), but the **stored params**
that carry the value to execution are not indirected. Result: (a) the plaintext
credential is recoverable by anyone with approval-row / park-store read access,
and (b) it lingers under whatever retention the approvals / park store has.

## Ask

Redact / indirect the credential_write payload at **park time** so the value
never persists on the approval row:

- The approver sees the **path + metadata**, never the value (already true for
  the preview — extend it to the stored params).
- A **one-time, server-side handle** carries the value from park to execution
  (the value stays in a governed, access-controlled store keyed by the handle;
  the park row holds only the handle + reference), **or** credential staging
  routes through a redacted secret-broker seam.

## Why

`credential_write` parks are, by design, approver-visible (any signed-in
approver can read the park) and durably retained. Persisting the raw value there
is a credential-exposure surface and an audit-retention concern, and it is
asymmetric with how other secret material is handled — the preview is redacted
(#2332) and session tokens are encrypted at rest, but the park params for a
credential write are neither.

## Relationship to the broader secret-state review

This is a **distinct persistence sink** from the credential-exposure concerns
already being addressed in a downstream automation consumer's durable
workflow-input store — those close the consumer-engine copy; this issue closes
the **backplane approvals / park** copy for governed `credential_write` ops,
where its fix lives.

## Acceptance criteria

- [ ] A parked `credential_write` op (starting with `vault.kv.put`) does not
      store the secret value in the approval / park-row params.
- [ ] The value is carried to execution via a one-time server-side handle (or an
      equivalent broker indirection) resolved only at approve→execute time.
- [ ] Approver-facing surfaces (preview + any raw params view) show
      path / metadata only.
- [ ] Post-approval execution still succeeds (the value is resolved from the
      governed store, not the park row).
- [ ] The class is covered, not just `vault.kv.put` — the same indirection
      applies to other `credential_write` ops that carry secret material.
- [ ] Docs + CHANGELOG.

## References

- Preview redaction (already shipped): #2332 (redaction-safe proposed_effect
[...]

PR #3539, opened 2026-09-11
feat(connectors): linux.script.run server-side secret_env Vault injection

### LINKED ISSUE #3536: feat(connectors): linux-ssh linux.script.run — server-side secret-injection seam (Vault-ref env resolved at execution) so credentials never persist on the approval row
## Summary

`linux.script.run` (linux-ssh connector, #3361) is `dangerous` +
`requires_approval` + `credential_write`, and its `env`, `arguments`, and
`script` persist **verbatim** on the approval / park row. A governed script
therefore has **no safe way to receive a credential**: putting the value in
`env` / `arguments` / `script` exposes it to the approver and to audit
retention, and the only safe pattern today is passing a non-secret Vault
**path** and resolving it on-host — which requires a Vault CLI + token already
present on the target.

## Ask

A server-side secret-injection seam for `linux.script.run`: an `env` (and/or
argument) value can be declared as a **Vault reference** that the backplane
resolves **at execution time**, injecting the resolved value into the remote
process environment — so the secret never touches `params`, the approval / park
row, or the audit log. This mirrors the connector's existing sudo-password
resolution (resolved server-side and handed to the transport, never stored in
op params).

## Why

Governed remote build / bootstrap scripts routinely need a credential (registry
pulls, appliance root passwords, API tokens). Without a resolution seam the
operator must either (a) leak the value onto the approval row, or (b)
pre-provision a Vault CLI + token on every target — defeating the point of a
governed, credential-brokered connector.

## Shape (proposal)

- A param convention (e.g. `env: {NAME: {vault_ref: "secret/…#field"}}`) that
  the dispatcher resolves through the existing credential-backend seam at
  execution, after approval, injecting into the SSH command environment.
- The preview / proposed_effect and the persisted params show the
  **reference**, never the value (parity with the #2332 vault.kv.*
  preview-redaction pattern).
- Resolution honours the connector credential-backend (Vault / GSM), so it works
  on both backends (cf. #2397 / #2860 patterns).
- Bounded and typed — no arbitrary env passthrough of secret material beyond
  declared refs.

## Acceptance criteria

- [ ] `linux.script.run` can inject a Vault-referenced value into the remote
      process env / arguments, resolved server-side at execution time.
- [ ] The secret value never appears in `params`, the approval / park row, the
      proposed_effect preview, or the audit log — only the reference does.
- [ ] Works without a Vault CLI / token pre-staged on the target host.
- [ ] Resolution routes through the credential-backend seam (Vault + GSM).
- [ ] Docs + CHANGELOG.

## References

- Parent: #3359 (linux-ssh typed connector Initiative).
- The write op this extends: #3361 (linux-ssh governed writes —
  script / service / file with approvals + previews).
- Related credential-handling: #3364 (guest-ops per-VM guest credential
  resolution — sibling concern on a different tier), #2332 (redaction-safe
  proposed_effect preview for vault.kv.* — approver sees path, never value),
  #2397 / #2860 (credential-backend seam usage).

damir-topic:
Closes #3536.

## Problem

`linux.script.run` is `safety_level=dangerous`, `requires_approval=True`, and pinned `credential_write`. Its `script` / `arguments` / `env` are base64-carried on the wire but stored **verbatim** in `ApprovalRequest.params` (needed to re-dispatch the exact call on approval) and in the audit params. That is the right design for re-dispatch, but it means a governed remote build/bootstrap script that needs a credential cannot safely *receive* one: any secret passed via `arguments` / `env` is durable on the approval row and in audit.

This is the `script.run`-specific mitigation for the general "credential_write parks persist raw secret values on the approval row" problem tracked in #3537 — `secret_env` keeps the value out of `params` entirely by resolving it server-side at execution time.

The only prior workaround was to pass a non-secret Vault *path* and resolve it **on the host**, which requires a Vault CLI + token to be present on the target (a live credential parked on a third host) — the un-governed footgun this connector exists to retire. The connector already resolves the sudo password and SSH auth material server-side; `secret_env` extends that same pattern to script credentials.

## Change

Add an optional `secret_env` parameter to `linux.script.run` mapping `ENV_NAME -> "<vault-path>#<field>"`:

1. **Validated fail-closed** before any SSH/Vault I/O: `ENV_NAME` is a POSIX identifier that must not start with a digit, collide with a reserved wrapper variable (`tmp` / `args` / `f`), or repeat an `env` name; the reference must be a single `path#field` with a non-empty logical KV-v2 path (same convention `load_vault_secret_data` accepts — a `targets/<id>`-style path under the `secret` mount, **not** an API-path-shaped `secret/data/…`) and field; the mapping is bounded (≤ 32 entries).
2. **Resolved at execution time only** — the `_approved=True` resume path or a non-parking dispatch, never at park time — through the connector's single `_resolve_secret` → `load_vault_secret_data` seam (reused, not forked) under the **dispatching operator's** identity/tenant scope, exactly as the sudo-password resolution scopes. Injected into the remote wrapper as an env var like the existing `env` (base64-carried, `umask 077` temp). Reads are de-duplicated per path.
3. **Never persisted.** `ApprovalRequest.params`, audit params, and the broadcast payload carry only the `ENV -> path#field` mapping; the resolved value never enters `params`, the `OperationResult`, the preview, or a log line. The park-time bespoke preview echoes the (non-secret) reference mapping so the reviewer sees which credential each env var will receive.
4. **Fail closed** on a missing field, a forbidden / API-path-shaped path, or any Vault resolution error: a structured error naming the `ENV` + the `path#field` reference (never a value), raised before any remote execution (no partial run).

`env` behaviour is unchanged. `secret_env` is `script.run`-scoped; the connector has no `command.run` verb to extend it to.

## Tests

New coverage in `test_connectors_linux_write.py` (14 tests): parameter validation (names / syntax / reserved+env collision / count cap); resolution + injection with a mocked Vault (value reaches the wrapper env base64-carried, never `params` or the result) on both the sudo and non-sudo paths; per-path read de-duplication; resolution at execution time (reads the operator-supplied path when the handler runs); fail-closed on missing field and forbidden path (no remote execution); the op schema declares `secret_env`; the park-time preview echoes the reference mapping never a value; and a DB-backed park test asserting `ApprovalRequest.params` + `proposed_effect` carry only the mapping with the resolved-value canary absent everywhere.

Docs: a `secret_env` section in `docs/codebase/connectors-linux.md` with the approval-row-persists-params rationale and the on-host-vault-CLI anti-pattern it replaces.

## Verification

- Targeted: `test_connectors_linux_write.py` 56 passed (42 existing + 14 new).
- Guards: `test_reference_docs_drift.py` + `test_flight_recorder_typed_spans.py` (F8 typed-span family coverage) green.
- CLI OpenAPI snapshot (`cli/api/openapi.json`) regenerated — no drift (op param schemas are DB-delivered, not in the FastAPI snapshot).
- Full unit sweep (excluding the service-backed `tests/integration` + `tests/acceptance` tiers): 15439 passed; the only failures are environment-dependent (Postgres/Valkey services, network sockets, `helm template`) and unrelated to this change.

No CHANGELOG entry in this PR (merge-queue convention: CHANGELOG lands in a separate docs-only PR).

### DIFF
--- a/backend/src/meho_backplane/connectors/linux/ops_write.py
+++ b/backend/src/meho_backplane/connectors/linux/ops_write.py
@@ -636,8 +636,7 @@ async def _resolve_secret_env(
                 # The underlying errors name the target / field / secret_ref
                 # path, never a credential value; add the ENV for the operator.
                 raise LinuxWriteError(
-                    f"secret_env[{name!r}] could not resolve vault reference "
-                    f"{path}#{field}: {exc}"
+                    f"secret_env[{name!r}] could not resolve vault reference {path}#{field}: {exc}"
                 ) from exc
             cache[path] = secret_data
         if not isinstance(secret_data, dict) or field not in secret_data:
--- a/backend/src/meho_backplane/connectors/linux/ops_write.py
+++ b/backend/src/meho_backplane/connectors/linux/ops_write.py
@@ -26,7 +26,12 @@ Tier table:
   ``umask 077`` temp path and execute it (optionally under sudo), capturing
   stdout / stderr / exit. Its ``arguments`` / ``env`` may carry secrets, so it
   is pinned ``credential_write`` and previews only at park time via its
-  bespoke builder (never the body / argument values / env values).
+  bespoke builder (never the body / argument values / env values). A credential
+  the script needs is instead passed by reference via ``secret_env``
+  (``ENV_NAME -> "<vault-path>#<field>"``), resolved from Vault server-side at
+  execution time (the same ``_resolve_secret`` seam the sudo password uses) and
+  injected as an env var -- only the mapping, never the value, is persisted on
+  the approval row / audit params.
 * ``linux.sysctl.write`` -- ``dangerous`` + approval (group ``system``). Set a
   kernel parameter (runtime ``sysctl -w`` + a persistent drop-in). Non-secret,
   so it previews via ``preview_operation`` on the generic params-echo default.
@@ -56,9 +61,14 @@ from __future__ import annotations
 
 import base64
 import shlex
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
-from meho_backplane.connectors._shared.vault_creds import strip_credential_value
+from meho_backplane.auth.vault import VaultClientError
+from meho_backplane.connectors._shared.vault_creds import (
+    CredentialsReadError,
+    strip_credential_value,
+)
 from meho_backplane.connectors.linux._sudo import run_remote_bash_with_sudo
 from meho_backplane.connectors.linux.ops import (
     SSH_TRANSPORT_NOTE,
@@ -177,6 +187,18 @@ _SYSCTL_VALUE_CHARSET: frozenset[str] = frozenset(
 _ENV_NAME_CHARSET: frozenset[str] = frozenset(
     "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_"
 )
+
+#: Upper bound on the number of ``secret_env`` entries one ``script.run`` may
+#: declare -- a bounded fan-out of per-call Vault reads, not an open-ended loop.
+_MAX_SECRET_ENV: int = 32
+
+#: env-var names ``secret_env`` must NOT bind: the ``script.run`` wrapper's own
+#: shell variables (:func:`build_script_run_wrapper` sets ``tmp`` and, when
+#: arguments are present, ``args``; the sudo primitive uses ``f``). Binding one
+#: from a resolved secret would clobber the wrapper's control state -- the temp
+#: path it executes and then removes -- so a collision is refused *before* any
+#: Vault read. ``env`` is left unchanged; this guard is ``secret_env``-only.
+_RESERVED_ENV_NAMES: frozenset[str] = frozenset({"tmp", "args", "f"})
 
 #: Hard cap + default for ``linux.script.run``'s wall-clock timeout.
 _DEFAULT_TIMEOUT_S: int = 120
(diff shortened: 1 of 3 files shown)


damir-topic:
> verdict is advisory prose, **not** a GitHub review; `reviewDecision`
> is unchanged and branch protection is not satisfied by this comment.

**Anchor task:** #3536 — server-side secret-injection seam for `linux.script.run` so a credential never persists on the approval row
**Surface area:** backend (connector / credential handling)
**Verdict:** APPROVE
**Head SHA reviewed:** `d591c47c47225528d997db2468c91b8227b3de0f`

### Acceptance criteria (from #3536)
- [x] Inject a Vault-referenced value into the remote process env at execution time — `secret_env` resolves in `linux_script_run` on the `_approved=True` handler path and merges into the wrapper env (base64-carried).
- [x] Value never in `params` / approval row / `proposed_effect` preview / audit / log — proven by the DB-backed park test (`ApprovalRequest.params`, `proposed_effect` carry only the `ENV → path#field` mapping; the resolved-value canary is absent), and by the log path (`_run_command` logs only `cmd_len`; the flight-recorder redaction omits the SSH command; the sudo primitive streams on stdin).
- [x] Works without a Vault CLI/token on the target — resolution is server-side through `_resolve_secret`.
- [x] Routes through the credential-backend seam (Vault + GSM) — `_resolve_secret → load_vault_secret_data → _resolve_and_load` performs scheme dispatch, so a `vault:`/`gsm:` ref resolves through the shared backend registry.
- [x] Bounded + typed — declared refs only, ≤ 32 entries, fail-closed validation.
- [~] Docs + CHANGELOG — docs delivered (`docs/codebase/connectors-linux.md`); CHANGELOG intentionally deferred to a docs-only PR per the merge-queue convention.

### Security review (credential seam) — findings
No blocking findings. The leak-prevention contract holds end to end:

- **Resolution timing** — `linux.script.run` is `dangerous` + `requires_approval`; the handler body runs only on the approved-resume path, so the value is never resolved at park (verified by the park test) and is re-resolved from the persisted mapping when the approved call executes.
- **Fail-closed ordering** — every ref is resolved before the wrapper is built or any remote command runs; a missing field or an unresolvable/forbidden (API-path-shaped) path raises before any SSH (`sudo_mock.assert_not_awaited()` in both fail tests).
- **Env hygiene** — the reserved-variable guard `{tmp, args, f}` correctly covers both the non-sudo wrapper (`tmp`, `args`) and the sudo primitive (`f`); `env`/`secret_env` name collisions are refused, so the `{**env, **resolved}` merge overwrites nothing; the resolved value rides the wrapper base64-encoded and is not written into the uploaded script temp file.
- **Scope** — reuses the single `_resolve_secret` Vault seam; no new client path; the API-path-shape guard applies via the shared backend.

### Integration verification
- [x] `ruff check` / `ruff format --check` — clean (both files)
- [x] `mypy src/.../linux/ops_write.py` — clean
- [x] `pytest tests/test_connectors_linux_write.py` — 56 passed
- [x] F8 typed-span guard (`test_flight_recorder_typed_spans.py`) — 17 passed
- [x] reference-docs drift (`test_reference_docs_drift.py`) — 4 passed
- [x] CLI OpenAPI snapshot — no drift expected/observed (typed-op param schemas are DB-delivered, not in the FastAPI snapshot; snapshot file correctly absent from the diff)
- [x] CI rollup — 22 success, 10 skipped, 0 failing

### Docs
- [x] `docs/codebase/connectors-linux.md` — updated (new `secret_env` section + op table + preview shape)

- **Error-contract completeness for a mistyped scheme.** `_resolve_secret_env` catches `(VaultClientError, CredentialsReadError)`, so an unregistered scheme (e.g. a typo `valut:foo#bar`) raises a bare `UnknownCredentialBackendError` rather than the structured `LinuxWriteError` naming the `ENV` the docstring promises. Still fully fail-closed (no remote exec, no leak) — only the error ergonomics differ. Consider adding `UnknownCredentialBackendError` to the caught set.
- **Security-property wording vs. GSM.** The module/op/docs text describes resolution as "under the dispatching operator's identity/tenant scope … resolved from Vault." Because #3536 mandates GSM routing and the seam dispatches by scheme, on a **GSM SA-direct** backend the read runs under the deployment identity (operator ignored) and the operator-supplied ref can name an arbitrary path. This is by design per the issue, but the wording slightly overstates operator-scoping/Vault-only; softening it to name the backend-identity model would prevent an approver from over-trusting a `gsm:` reference. The approval gate + previewed reference remains the governing control.
- **Test depth.** Re-resolution is proven at the handler level (which *is* the resume path), not through a full park → approve → execute DB cycle; an end-to-end resume test would harden the "re-resolved from the stored mapping" claim.

### What worked well
- The DB-backed park test asserts against the real persisted structures (`ApprovalRequest.params` + `proposed_effect`), not just mock call args — exactly the right rigor for a persistence-leak contract.
- Reuse of the single `_resolve_secret` seam (via a minimal `_VaultRefTarget` stub) instead of forking a Vault client keeps the scoping/audit/fail-closed behavior identical to the sudo-password read.

Issue #3152, opened 2026-08-26
policy: non-agent verdict ignores safety_level — mutating caution ops execute ungated for service principals


damir-topic:
## Finding

For non-agent principals (service principals / operator REST callers), the policy gate parks **only** on `descriptor.requires_approval == True` and never consults `safety_level` — `_non_agent_verdict` in `backend/src/meho_backplane/operations/_validate.py` (~L140-178). `safety_level` gates only *agent* principals via `resolve_verdict`.

Consequence observed while walking a real service-principal workflow: **mutating** ops classed `caution` but carrying `requires_approval=False` execute completely ungated for a service principal — concretely `POST:/vcenter/vm/{vm}/hardware/disk`, `POST:/vcenter/vm/{vm}/hardware/cdrom` (both attach real resources to a VM), and a connector's validation dry-run write. An operator whose mental model is "nothing mutates without being on a reviewed list" is wrong today for this principal class.

## Ask — an explicit decision, either way

1. **Extend the non-agent gate** to consult `safety_level` (e.g. park `dangerous` always; park `caution` mutations unless covered by a standing grant), or
2. **Document the contract** that `requires_approval` is the sole non-agent gate, and audit the op inventory so every mutating op either carries `requires_approval=True` or is explicitly, deliberately flagged as ungated-by-design.

Either resolution is defensible; the current state (implicit, undocumented, discovered by reading the validator) is not.

## Related

Companion to the standing-grants proposal (#3151) — the two define the non-agent approval model together and should be decided as one.

PR #3564, opened 2026-09-13
fix(ingest): gate VMware VM hardware writes

### LINKED ISSUE #3563: vmware-rest-9.0: VM hardware-mutating generic writes (e.g. `POST:/vcenter/vm/{vm}/hardware/ethernet`) run at `caution` without an approval park — align with the composites' `dangerous` tier
## Observed

On connector `vmware-rest-9.0`, the **generic ingested** operation
`POST:/vcenter/vm/{vm}/hardware/ethernet` (add a virtual NIC to a VM —
here to a powered-on VM with `start_connected: true` and a
`STANDARD_PORTGROUP` backing) dispatched with `safety_level = caution`
and executed **immediately**: HTTP 201, **no approval park**.

By contrast, the curated composites for the same class of effect are
one tier stricter:

- `vmware.composite.vm.nic.repoint` (change an existing NIC's portgroup
  backing) is `safety_level = dangerous`, `requires_approval = true`.
- `_repoint_vm_nics_to_fallback` inside `host.detach_from_vds` also runs
  under an approval.

Attaching a new *connected* NIC to a *running* VM puts the guest on a new
L2 — the same blast-radius class as repointing, arguably larger, because
it also creates a device and mints a new MAC. So the ingested-op tier and
the curated-composite tier disagree for the same effect.

### Root cause (grounded in the ingest code)

The generic OpenAPI ingest classifies `safety_level` purely by HTTP verb,
with `requires_approval` hard-wired to `False`, and there is **no per-path
override**:

- `backend/src/meho_backplane/operations/ingest/openapi.py:1175-1181` —
  `_safety_level_for(method)`: `DELETE` → `dangerous`, `POST`/`PUT`/`PATCH`
  → `caution`, everything else → `safe`.
- `backend/src/meho_backplane/operations/ingest/openapi.py:357-361` —
  `_CAUTION_VERBS = {"POST","PUT","PATCH"}`, `_DANGEROUS_VERBS = {"DELETE"}`.
- `backend/src/meho_backplane/operations/ingest/openapi.py:1152-1163` —
  every `EndpointDescriptorProto` is built with
  `safety_level=_safety_level_for(method)` **and** `requires_approval=False`.

So `POST:/vcenter/vm/{vm}/hardware/ethernet` is a POST → `caution`,
`requires_approval=False` — a pure verb heuristic that cannot distinguish
"add a tag" from "attach a running VM to an arbitrary L2". A repo-wide
search finds no per-path / per-path-family safety override or gate-policy
table applied at ingest or registration; the only tier overrides that
exist are the hand-authored ones on the vmware_rest composites
(`connectors/vmware_rest/composites/_register.py`).

This also compounds with the closed policy finding #3152: for non-agent
principals the gate parks only on `requires_approval == True` and never
consults `safety_level` — so a `caution` + `requires_approval=False`
ingested VM-hardware write is doubly ungated for a service principal.

## Expected

VM hardware-mutating generic writes under `/vcenter/vm/{vm}/hardware/*` —
POST / PATCH / DELETE on `ethernet`, `disk`, `cdrom`, `boot`, `serial`,
`parallel`, `floppy`, `cpu`, `memory`, and the disk/adapter sub-paths —
should ingest at **at least `dangerous` / `requires_approval=true`**,
consistent with the curated `vmware.composite.vm.nic.repoint` tier for the
same blast-radius class. The pure-verb heuristic in
`_safety_level_for` (`openapi.py:1175-1181`) is the wrong altitude for
this path family and needs a per-path-family override at ingest time.

## Why it matters

Today an operator (or a service principal — see #3152) can attach a
running VM to any L2 segment, add a disk, mount a CD-ROM, or change
cpu/memory on `vmware-rest-9.0` with **no four-eyes and no approval park**,
while the curated path for a *smaller* change (repointing an existing NIC)
correctly parks for approval. The governance surface is only as strong as
its weakest tier, and the generic ingested tier is currently the weak one
for VM hardware mutation. It also trains the wrong operator mental model:
"the backplane parks destructive VM changes" is false for the ingested
path.

## Proposal

1. **Per-path-family safety override applied at ingest/registration.**
   Introduce a declarative override (a connector-scoped table keyed by
   `(method, path-glob)`, or an equivalent registration-time hook) that the
   generic ingest consults after `_safety_level_for` so a matched
   descriptor is promoted — e.g. `POST|PATCH|DELETE /vcenter/vm/{vm}/hardware/**`
   → `dangerous` + `requires_approval=true`. Keep the verb heuristic as the
   default; the override only *raises* the tier, never lowers it. This is
   the missing mechanism — there is no per-path override today.
2. **Conformance test pinning the tier for the path family** so a
   re-ingest (spec refresh / reconcile lane) cannot silently regress it —
   in the spirit of the existing spec-reconcile guards. The test asserts
   that every ingested `vmware-rest-9.0` descriptor whose path matches the
   VM-hardware family carries `safety_level >= dangerous` and
   `requires_approval == true`.
3. **Audit the sibling families and list them** (below) — apply the same
   override to any other ingested path family whose blast radius exceeds
   what a bare `caution` conveys.

### Sibling path families to audit (same connector, same risk class)

- `POST|PATCH|DELETE /vcenter/vm/{vm}/hardware/ethernet` (this report)
- `POST|PATCH|DELETE /vcenter/vm/{vm}/hardware/disk`
- `POST|PATCH|DELETE /vcenter/vm/{vm}/hardware/cdrom` (already flagged
  ungated for service principals in #3152)
- `PATCH /vcenter/vm/{vm}/hardware/cpu`, `PATCH .../hardware/memory`
- `POST|PATCH|DELETE /vcenter/vm/{vm}/hardware/serial|parallel|floppy`
- `PATCH /vcenter/vm/{vm}/hardware/boot`
- adjacent VM-state mutation to consider in the same audit:
  `POST /vcenter/vm/{vm}/power/*` (start/stop/reset/suspend),
  `DELETE /vcenter/vm/{vm}` (already `dangerous` by verb, but confirm
  `requires_approval`), and VM create `POST /vcenter/vm`.

## Acceptance criteria

- [x] A per-path-family safety override mechanism exists and is applied at
      generic ingest/registration (raises tier only; documented where it
      lives).
- [x] `POST|PATCH|DELETE` under `/vcenter/vm/{vm}/hardware/*` on
      `vmware-rest-9.0` ingest at `safety_level >= dangerous` with
      `requires_approval = true`.
- [x] Adding a connected NIC to a powered-on VM
      (`POST:/vcenter/vm/{vm}/hardware/ethernet`) parks for approval and
      does not execute immediately.
- [x] A conformance test pins the tier for the VM-hardware path family and
      fails if a re-ingest regresses it.
- [x] The sibling families above are audited and their resulting tiers are
      recorded in the PR (each either promoted or explicitly, deliberately
      left as-is with rationale).
- [x] `docs/codebase/` is updated to document the ingest safety-override
      mechanism and the VM-hardware tier contract.

## Related

- #3562 — `vmware_rest vm.nic.repoint` can target a STANDARD portgroup
  (the `STANDARD_PORTGROUP` backing extension; establishes that the
  curated repoint path is `dangerous`/`requires_approval`).
- #3152 — policy: non-agent verdict ignores `safety_level` (the
  gate-*consumption* side; this issue is the tier-*assignment* side —
  fixing both is required for the ingested VM-hardware writes to actually
  park for a service principal).
- #3151 — policy: standing scoped auto-approval grants for service
  principals (companion to #3152; the standing-grant model is how a
  promoted-tier op would be pre-authorized for unattended runs).
- #3535 — governed add of a VMXNET3 ethernet adapter is not served/curated
  (the coverage/curation ask for the same `POST .../hardware/ethernet`
  path; this issue is about the safety *tier* of the generic ingested op,
  not curation).

damir-topic:
Closes #3563

## What

Adds a connector-advertised safety floor for VMware REST generic operations. VM
hardware writes, VM create/delete, and VM power writes now register as
`dangerous` and require approval. The vendor-specific matching rules live with
the VMware connector; generic ingestion only applies advertisements registered
during connector startup.

Existing matching descriptors are corrected on re-ingest without weakening a
stricter reviewed safety tier, approval requirement, enablement, or review
state. Existing catalog rows take the corrected posture on their next
connector re-ingest; deployment alone does not rewrite stored descriptors. The
ordinary OpenAPI parser remains vendor-neutral.

## Validation

- `uv run pytest tests/test_operations_*.py -q` — 953 passed, 1 skipped
- Targeted registration, safety-change, dispatcher, and service-grant tests — 133 passed
- Ruff, mypy, and changed-file `code-quality.py --files` / `--diff` passed
  (pre-existing module-size warnings only)


damir-topic:
> satisfied by this comment.

**Anchor task:** #3563 — VM hardware-mutating generic writes run at `caution` without an approval park; align with the composites' `dangerous` tier
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `b8540d70`

### Acceptance criteria
- [x] Per-path-family safety override mechanism, applied at generic ingest/registration, raises tier only, documented where it lives — `operations/ingest/safety_floors.py` (vendor-neutral registry) + `connectors/vmware_rest/ingest_safety.py` (VMware rules) + `docs/codebase/spec-ingestion.md`.
- [x] `POST|PUT|PATCH|DELETE` under `/vcenter/vm/{vm}/hardware/*` ingest at `safety_level >= dangerous` + `requires_approval=true` — `test_vmware_hardware_writes_receive_connector_safety_floor` asserts `{("dangerous", True)}` over 9 paths × 4 verbs.
- [x] Adding a connected NIC to a powered-on VM parks and does not execute — `test_vmware_ingested_hardware_write_parks_before_http_execution` asserts `awaiting_approval`, `instance.calls == []`, `captured_events == []` (parks before the HTTP branch and before broadcast).
- [x] Conformance test pins the tier and fails on regression — registration pin + re-ingest monotonicity (`test_vmware_floor_corrects_existing_row_without_weakening_review_state`) + cold-start boot (`test_eager_connector_import_registers_vmware_safety_floor`).
- [x] Sibling families audited and tiers recorded — docs table records hardware/**, `POST /vcenter/vm`, `DELETE /vcenter/vm/{vm}`, power, and the reads/non-VMware exclusions.
- [x] `docs/codebase/` documents the mechanism + the VM-hardware tier contract — `spec-ingestion.md` "Connector safety floors".

### Predicted vs. actual diff
- ✅ `safety_floors.py` (registry), `ingest_safety.py` (VMware floor), `_upsert.py` (monotonic `_apply_safety_metadata` + `_capture_safety_change`), `register_ingested.py` (`_prepare_ingested_operations` flooring), `pipeline.py` (dry-run), `vmware_rest/__init__.py` (import lifecycle) — all expected and correct.
- ✅ Parser (`openapi.py`) deliberately untouched — vendor-neutrality preserved.
- ❓ None undisclosed.
- 🕳️ None missing.

### Integration verification
- [x] `mypy` (changed source) — Success, no issues in 5 files
- [x] `pytest tests/test_operations_register_ingested.py tests/test_operations_dispatcher.py` — 94 passed
- [x] `pytest -k "dry_run or pipeline or safety_floor or safety_change"` — 99 passed, 1 skipped (1 testcontainers error = no local Docker; that lane is green in CI)
- [x] dispatcher park test passes in isolation (floor registration is order-independent via conftest eager-import; `clear_registry()` does not clear `_FLOORS`)

### Findings
None blocking. APPROVE.

- Rollout: deployed `vmware-rest-9.0` catalog rows keep the stale `caution`/`requires_approval=False` posture until an operator re-runs ingest (`POST /api/v1/connectors/ingest`). The PR body states this, but `spec-ingestion.md` does not name the required operator step. Consider a follow-up to either backfill matching rows (one-time reconciliation) or document the re-ingest step in deploy notes — the issue's proposal scoped the fix to ingest-time, so this is genuinely out of THIS PR's scope but worth tracking given the governance intent.
- Dry-run: `_run_dry_run` computes `floored` but uses only `len(floored)`, which equals `len(parsed)`, so the floor has no operator-visible effect there; the result carries only counts. Harmless and consistent, but the docs line "the dry-run path applies the same floor before reporting its result" overstates what the operator sees.
- The VMware floor is pinned to exactly `version == "9.0"`; a future connector version label would need its own advertisement (correct-by-design, but worth noting).
- Consider auditing VM `unregister`/instant-clone action spellings (`POST /vcenter/vm/{vm}?action=...`) — clone/register land on `POST /vcenter/vm` (covered), but a `POST /vcenter/vm/{vm}` action verb would not match (only `DELETE` is floored there).

### Docs
- [x] `docs/codebase/spec-ingestion.md` — updated (mechanism + VM-hardware tier contract table)

### What worked well
- The floor is fail-safe and spec-agnostic: broad prefix match + every mutating verb accepted, so a spec refresh introducing new hardware sub-paths stays covered without a spec-pinned enumeration.
- `requires_approval=True` is set explicitly (not just `safety_level`), which is load-bearing: for a USER/non-agent principal the park keys on `descriptor.requires_approval` (`_validate.py:402`), not on the `dangerous` tier (only `destructive` parks a USER via `safety_level`). This correctly closes the #3152 tier-assignment side.
- Monotonicity is real and tested end to end: `max(existing, incoming)` on tier, OR on approval, with the audit `SafetyLevelChange` captured before the update and mirrored to the same value; enablement and review state are untouched by omission.

Issue #3571, opened 2026-09-13
connectors: `oauth2_mint` auth scheme — support an external token issuer (token_url/issuer + scope/audience) for targets whose IdP is a separate host


damir-topic:
## Goal

Extend the declarative `oauth2_mint` auth scheme so a profiled generic connector can authenticate to a target whose OAuth2 token issuer is a **separate host** (an identity provider) from the target API itself — mint a `grant_type=client_credentials` token at that external issuer and present it as `Bearer` to the target — while keeping today's behaviour byte-for-byte for the existing target-relative issuer.

## Why

`oauth2_mint` today (`backend/src/meho_backplane/connectors/_shared/profile_auth.py`) hardcodes a **target-relative** token endpoint (`/realms/master/protocol/openid-connect/token`) and sends only `grant_type`/`client_id`/`client_secret`. That models the case where the token endpoint lives on the same host as the API. It cannot model a target whose IdP is a different host — a common shape when an API's token issuer is a standalone identity provider. Without this, such a target cannot be reached by a declarative profile at all; it would force a bespoke typed subclass, widening the hand-coded surface the profiled-connector work exists to shrink.

## Scope

Extend the declarative scheme (not a typed subclass); default to today's behaviour so the existing `oauth2_mint` request shape is unchanged.

1. **`AuthSpec`** (`backend/src/meho_backplane/connectors/profile.py`) gains three **optional** fields, all valid only for `scheme="oauth2_mint"` (forbidden on every other scheme):
   - `token_url` — absolute `https` URL of the token endpoint. Absent → keep today's target-relative `/realms/master/...` derivation. Plaintext `http` accepted **only** for a cluster-internal / private-address issuer (fail-closed for public hosts).
   - `scope` — optional OAuth2 `scope`, sent as the `scope` form param.
   - `audience` — optional audience, sent as the `audience` form param (Keycloak / RFC 8693-style).
2. **Mint/cache/re-mint harness stays shared.** Only the token endpoint resolution (`login_path`) and the form body (`_oauth2_mint_body`) change. `client_id`/`client_secret` continue to resolve from the Vault-backed target credential; the token and secret are never logged or recorded in flight-recorder spans (the login POST already bypasses the recorded `_post_json` span; verify).
3. **Boot-time profile scheme-load guard** accepts the new optional fields; older profiles without them validate unchanged.
[...]

Issue #2900, opened 2026-08-11
Initiative: add-on pairing contract — pairing identity, capability advertisement, step-event push, audit parent-linkage, paired-surface activation


damir-topic:
Parent goal: evoila-bosnia/meho-internal#223 (meho-automation; product goal relocated to the internal tracker 2026-08-12 per the board-scoping rule — this initiative is the backplane-side capability and stays on this board)

## Outcome

The backplane gains a first-class **add-on pairing contract** — the integration plane that lets a sibling product (first consumer: meho-automation; meho-knowledge is the informal precedent this formalizes) pair once and become seamless, while an unpaired backplane stays byte-identical to today with no dead surfaces.

This is deliberately **not** connector registration: connectors are for managed vendor systems below the narrow waist; a paired add-on is a peer control plane beside it, integrated through the governance planes themselves.

## Scope

1. **Pairing + identity** — a one-time handshake registering the add-on: Keycloak client-credentials principal, versioned contract negotiation, minimum-version pinning both ways, health/liveness surfaced in `/status` and the console.
2. **Capability advertisement** — the add-on declares its surfaces (meta-tool family, CLI verb family, console panels, event kinds) against the contract version; the backplane activates them only while paired and healthy.
3. **Step-event push contract** — a durable outbound event subscription for the add-on's orchestration needs (approval outcomes, dispatch completions for its own work_refs) replacing SSE polling; rides the existing outbox/broadcast substrate, scoped to the paired principal's own lineage.
4. **Audit parent-linkage** — an external orchestrator run can stamp a parent linkage so its per-step audit rows group as one replay subtree (#2086 semantics for out-of-process parents), keyed to `work_ref`.
5. **Paired-surface activation** — the `automation` meta-tool family (small, stable, first-party — the knowledge/G4 precedent; add-on entity identifiers are data, never tool names, postulate 5 intact), `meho automation` CLI verbs, and console panels, all feature-gated on pairing state.

## Definition of done

- [ ] An unpaired backplane is behaviorally identical to today (conformance test); pairing/unpairing is reversible and audited
- [ ] A reference add-on (test double) can pair, advertise a meta-tool family, receive step events for its work_refs, and produce a single audit replay subtree spanning its orchestration and the resulting dispatches
- [ ] Contract versioned + documented (`docs/codebase/` file); version-skew behavior defined (older add-on vs newer backplane and vice versa)
[...]

PR #3568, opened 2026-09-13
fix(dispatch): guard unqualified VMware catalogs

damir-topic:
## Summary

Guards enabled ingested `vmware-rest` 9.0 descriptors against targets outside VMware 9.x before policy can park an approval or construct a connector. The dispatcher uses the descriptor owner's exact v2 registry identity and a connector-owned, vendor-neutral compatibility hook; VMware supplies the PEP 440 target predicate.

The early pure class resolution runs only for connectors that advertise this catalog guard. Unadvertised ingested descriptors and all typed/composite descriptors retain their existing policy, targetless, and resolution ordering. A successful guarded class selection is reused after policy, avoiding a second resolution before instance construction.

Existing catalog rows receive this behavior when dispatched after deployment; no profile field, migration, or catalog qualification claim is introduced. 9.1 and 9.1.1 avoid the guard but are not claimed as fully qualified catalogs.

Closes #3565

## Validation

- `cd backend && uv run pytest tests/test_operations_dispatcher.py tests/test_connectors_resolver.py -q` — 99 passed
- `cd backend && uv run ruff check <changed Python files>`
- `cd backend && uv run ruff format --check <changed Python files>`
- `cd backend && uv run mypy <changed source files>`
- `cd backend && uv run python scripts/validate_vcf_api_contract_manifest.py --manifest ../docs/compatibility/vcf-api-contract-manifest.yaml`
- canonical `code-quality.py --files` and `--diff` — 0 blockers; pre-existing size warnings only

PR #3590, opened 2026-09-13
feat(connectors): register the meho-automation add-on as a profile-backed generic connector

### LINKED ISSUE #3578: connectors: register the automation add-on as a profile-backed generic connector target (launch / validate / gate ops; oauth2_mint external issuer; runtime spec ingest)
## Goal / Why

Make the `meho-automation` add-on's run-launch API governed-callable through the backplane, so an operator drives launch / validate / gate-decision via `call_operation` (same auth → policy → audit → JSONFlux path as any vendor API) instead of an out-of-band header-trust path. The add-on becomes just another **versioned, profile-backed generic-connector target** reached over HTTP with a Vault-credentialed, externally-minted Bearer. No reverse code coupling — the backplane calls the add-on only over the wire.

The auth mechanism this depends on already shipped: the `oauth2_mint` **external token issuer** support (`token_url` / `scope` / `audience` on `AuthSpec`) from #3571. The P1 auth design context is #2900.

## Scope

Ship the in-repo connector-registration artifacts for the add-on as a profile-backed generic connector — **no vendored spec, no hand-coded per-op class**:

1. **Catalog entry + `ExecutionProfile`** (`operations/ingest/catalog.yaml` + `connectors/profiles/`). Auth: `oauth2_mint` with an **external issuer**. The realm token endpoint is deployment/lab-specific and MUST NOT be hard-coded in the public repo — it is resolved per-target from the Vault-backed target credential (a `token_url` credential field), alongside `client_id` / `client_secret`. `audience` = `meho-automation` ships in the profile (non-secret).
2. **Op-tier curation** via a connector-owned ingest **safety floor**: after ingest, `launch` (`POST /api/v1/runs`) and `gate` (`POST /api/v1/runs/{run_id}/gates/{node_id}/decision`) land as `caution` (the generic POST heuristic — no approval park; the add-on's own gate nodes are the human control, operator decision), and `validate` (`POST /api/v1/blueprints/{blueprint_id}/validate`) is downgraded to `safe`. Floor-pinned so a spec re-ingest cannot silently reset the tiers.
3. **Runtime / file spec on-ramp** (no vendoring): the connector is ingested at registration time from the add-on's published `/openapi.json` (OpenAPI 3.1) via the existing operator on-ramp. The backplane's server-side spec fetch is deliberately public-https-only (SSRF hardening), and the add-on is an in-cluster plaintext service, so the sanctioned path is the inline `file://` on-ramp (operator fetches `/openapi.json` and ingests its bytes). The spec is not committed to this repo.
4. **Dispatch-canonical product token.** The connector-id parse (`{impl_id}-{version}`) recovers the product from the first hyphen-segment, so a hyphenated product does not round-trip (same constraint that realigned `sddc-manager`→`sddc`, `vcf-automation`→`vcfa`, `vcf-fleet`→`fleet`). The dispatch product token is therefore a short hyphen-free slug; the add-on's Kubernetes service name and the Keycloak `audience` are unaffected.

## Acceptance criteria

- Boot guards (`load_catalog`, `validate_catalog_registry_coverage`, `validate_shipped_artifacts`, profile scheme-load) pass on the new catalog row + profile with **no vendored spec**.
- The v2 registry resolves the connector for the add-on target `(product, version)` fingerprint (boot-stamped from the shipped profile).
- A fixture-backed ingest of a **minimal synthetic 3-route OpenAPI** (generic schema names, never the real spec) lands exactly the three ops staged/disabled with the decided tiers (launch/gate `caution`, validate `safe`).
- `oauth2_mint` sources its token endpoint per-target from the credential (external-issuer URL never in-repo); existing `oauth2_mint` behaviour (target-relative keycloak parity, and #3571 profile `token_url`) is byte-identical.
- MCP surface conformance unchanged — no per-op tools; launch/validate/gate ride `op_id` under `call_operation`.
- Docs: a registration recipe (`docs/codebase/connectors-meho-automation.md`) + a `connector-auth-coverage.md` row, including the security note that launch executes without a backplane approval park by operator decision.

## References

- #3571 — `oauth2_mint` external token issuer (`token_url` / `scope` / `audience`), merged.
- #2900 — add-on pairing contract / P1 auth design.

damir-topic:
## What

Registers the **meho-automation add-on** as a versioned, **profile-backed generic connector** target so the backplane drives its launch / validate / gate-decision API through `call_operation` — the same auth → policy → audit → JSONFlux path any vendor API rides — instead of an out-of-band header-trust call. The add-on stays a peer control plane reached only over the wire; **no reverse code coupling** (the backplane imports nothing from the add-on).

Closes #3578. Builds on the `oauth2_mint` external-issuer mint (`token_url`/`scope`/`audience`) shipped in #3571; P1 auth design context is #2900.

## Shipped in-repo (no vendored spec)

The add-on's OpenAPI is **not** committed here. The connector is ingested at registration time from the add-on's published `/openapi.json` (OpenAPI 3.1) via the operator `--spec` on-ramp.

- **Catalog row** (`operations/ingest/catalog.yaml`): product `mehoauto`, version `0.1.0`, impl `mehoauto-rest`, `upstream: null`, `spec_resource: null`, `profile_resource` set, `catalog_ingest: spec-only`. The product slug is **hyphen-free** so the connector_id round-trips — `meho-automation` would parse to product `meho` (same constraint that gave `sddc-manager → sddc`, `vcf-automation → vcfa`). The add-on's Kubernetes Service name and Keycloak audience (`meho-automation`) are decoupled from the slug.
- **ExecutionProfile** (`connectors/profiles/meho_automation_minimal.yaml`): `oauth2_mint` with an external, per-target issuer; audience `meho-automation`; fingerprint `GET /api/v1/version`.
- **Connector-owned ingest safety floor** (`connectors/meho_automation/ingest_safety.py`): pins the decided tiers so a re-ingest (or a heuristic change) can't drift them — **launch** (`POST /api/v1/runs`) + **gate** (`POST /api/v1/runs/{run_id}/gates/{node_id}/decision`) → `caution` with **no approval park**, **validate** (`POST /api/v1/blueprints/{blueprint_id}/validate`) → `safe` (read-side dry-run, downgraded from the POST-default `caution`).

**Security note:** launch executes without a backplane approval park by operator decision — the add-on's own in-run **gate nodes** are the human control point. Enabling a launch-capable connector and the first live launch remain operator-gated. MEHO's own approval-decision verbs remain human-only with no MCP path (unrelated).

## oauth2_mint external issuer — credential-sourced `token_url`

The realm token endpoint is a **per-deployment** value (each deployment mints against its own realm), so it is **not** hard-coded in the shipped, public profile. It is sourced **per-target from the Vault credential** as a `token_url` field (in `secret_fields`, alongside `client_id`/`client_secret`), via a new **additive** optional `SessionSchemeSpec.login_path_with_secret` hook. Precedence: credential `token_url` → profile `auth.token_url` (#3571) → target-relative `/realms/master/...`. The credential value clears the **same** fail-closed `_validate_token_url` SSRF rule the profile field does (https for a public host; plaintext http only for a cluster-internal/private issuer). A credential bundle with no `token_url` (keycloak) is **byte-identical to today**, and the profile-only `login_path(AuthSpec) -> str` signature the typed session connectors call at import is untouched.

## Runtime-spec seam: not needed

A profile-backed catalog row with `spec_resource: null` already clears every boot guard — `validate_shipped_artifacts` only dry-run-parses the profile (nothing to parse or leak), and `validate_catalog_registry_coverage` exempts profile-backed rows. `stamp_catalog_profiled_connectors` registers the synthesised `ProfiledRestConnector_mehoauto_0_1_0` under the `(mehoauto, 0.1.0, mehoauto-rest)` v2 key, resolvable for a target of that fingerprint. So **no** `spec_source: runtime` schema seam was added; the existing `catalog_ingest: spec-only` + null `spec_resource` shape is the sanctioned "ingest at registration" path.

## Agent surface

Unchanged — the three ops ride `op_id` under `call_operation`; no per-op MCP tools are added. `test_mcp_surface_conformance` stays green.

## Tests (run locally)

- New `test_connectors_meho_automation.py` (11 tests): row/profile shape, boot guards accept the row **without a vendored spec**, fixture-backed **synthetic 3-route** ingest (generic schema names — never the real spec) lands exactly the three ops staged/disabled with the decided tiers, boot-stamp + resolver resolution for the `(mehoauto, 0.1.0)` fingerprint, floor idempotency/version-scope.
- Credential-sourced `token_url` cases added to `test_connectors_profiled_auth.py` (dialed for the mint, overrides profile, fail-closed on public http).
- Catalog-snapshot updates in `test_operations_ingest_catalog.py`.
- Green locally: the connector/profile/ingest/boot-stamp/register/safety/conformance modules (509 passed in the focused batch), plus app-boot lifespan smoke and mypy on changed source files.

## Docs

- `docs/codebase/connectors-meho-automation.md` — registration recipe (target create with base URL + credential ref, ingest from `/openapi.json`, review groups, enable) + the launch-without-park security note.
- `docs/codebase/connector-auth-coverage.md` — credential-sourced `token_url` row under the oauth2_mint external-issuer section.


damir-topic:
**Verdict:** REQUEST_CHANGES

Independent review of PR head `ca40b63625629c21be71189a82990c59883466f4`. Read the full diff line-by-line; ran the relevant test modules locally against the PR-head source (all green: `test_connectors_meho_automation.py` 11/11, `test_connectors_profiled_auth.py` + `test_operations_ingest_catalog.py` 137/137, `test_mcp_surface_conformance.py` + `test_operations_boot_stamp.py` + `test_operations_register_ingested.py` 60/60). CI on the head is fully green (all SUCCESS/SKIPPED, no pending/failed). DCO sign-off + `Co-Authored-By` present; PR body ends with the Generated-with line; no CLAUDE.md/AGENTS.md/CHANGELOG edits.

The design is sound and the QC-critical items verify clean:
- **Safety-floor mechanism is the shared one, not bespoke.** `connectors/meho_automation/ingest_safety.py` registers through the same `operations/ingest/safety_floors.register_ingest_safety_floor` that `vmware_rest/ingest_safety.py` uses (the #3568/#3564 mechanism) — identical `(version, proto) -> proto` signature, `model_copy` idempotence, and `is proto` passthrough semantics. Not a parallel path.
- **Tiers are exactly as decided:** launch (`POST /api/v1/runs`) + gate (`POST /api/v1/runs/{run_id}/gates/{node_id}/decision`) → `caution`, `requires_approval=False` (no backplane park — the add-on's own in-run gate nodes are the human control, documented with the security note); validate (`POST /api/v1/blueprints/{blueprint_id}/validate`) → `safe`. The fixture-backed synthetic 3-route ingest, boot-guards, and `(mehoauto, 0.1.0)` resolver-resolution tests genuinely exercise the real pipeline (`parse_openapi` → `register_ingested_operations` → DB; `stamp_catalog_profiled_connectors` → `resolve_connector`).
- **Credential-sourced `token_url` reuses the same SSRF rule.** `_oauth2_login_path_from_secret` calls the vetted `_validate_token_url` verbatim (https for public hosts; plaintext http only for cluster-internal/private). A credential bundle with no `token_url` (keycloak) falls straight through `login_path` — byte-identical to today; the profile-only `login_path(AuthSpec)->str` signature the typed session connectors call at import is untouched. The mint body (client_secret) is never logged — only status/target/path. MCP surface unchanged.

## Blocking

1. **Genericize the realm segment in the new test constant.** `backend/tests/test_connectors_profiled_auth.py:1069` — `_CRED_ISSUER_URL = "https://realm.example.test/realms/evba/protocol/openid-connect/token"` carries the lab realm name `evba` in a newly-added public-repo test fixture. Per the public-repo leak policy for new test code, change the realm segment to a generic placeholder, e.g. `/realms/example/`. The host is already a generic placeholder (`realm.example.test`) and the test asserts only on the constant it defines, so the change is functionally inert.

   Transparency / severity note for the maintainer: `evba` is **not** a novel exposure — it is already the repo's canonical, public test realm. `backend/src/meho_backplane/connectors/keycloak/session.py:108` ships `DEFAULT_MANAGED_REALM: str = "evba"`, and `test_auth_config.py`, `test_connectors_keycloak_auth.py`, and `backend/tests/fixtures/rdc-hetzner-dc-targets.yaml` all use `/realms/evba` (and the real `keycloak.evba.lab` host) on `main`. This is therefore a hygiene/consistency fix on newly-added code rather than a new leak, and may be waived if the team accepts the pre-existing `evba` convention. A grep of the whole diff found **no other** lab identifiers (no IPs, no `.lab` hostnames, no tenant/codenames) — this one string is the sole occurrence.

## Nits (non-blocking)

1. **Vault-credential `token_url` widens who controls the mint destination.** `_validate_token_url` only fail-closes on cleartext-to-public; it does not host-allowlist, so any `https` endpoint is accepted (same as the #3571 profile field). Moving `token_url` from the review-gated profile into the per-target Vault credential means write access to a target's credential now also controls where the client-credentials mint POSTs `client_id`/`client_secret`. This stays within the operator trust boundary (the credential writer already owns the secret, and the profile field already permitted any https host) and is documented — no change requested, but worth capturing for whoever authors the Vault ACLs: write on a target credential == control of the mint destination host.
2. **Floor is version-pinned to exactly `0.1.0`** (mirrors the vmware_rest `9.0` gate). If an operator ingests under a different `--version`, `validate` silently stays `caution` rather than `safe` (launch/gate remain `caution` via the default POST heuristic regardless). Matches the established versioned-floor pattern; noted for awareness only.

Re-review once the blocking item is addressed (or explicitly waived).

damir-topic:
Addressed the blocking finding in 3655aacc.

**Blocking (realm name in new public-repo test fixture)** — `backend/tests/test_connectors_profiled_auth.py` `_CRED_ISSUER_URL` now reads `https://realm.example.test/realms/example/protocol/openid-connect/token` (was `/realms/evba/`). Genericized the realm segment per the public-repo leak policy for new test code. The change is functionally inert — the host was already the placeholder `realm.example.test` and the test asserts only on the constant it defines; all three usages reference the constant, so the single edit covers them. Whole-diff grep re-confirmed this was the sole lab-identifier occurrence in the PR.

**Nits** — no code change, capturing here:
- *Vault-credential `token_url` widens mint-destination control*: acknowledged and within the operator trust boundary (credential writer already owns client_id/client_secret; the #3571 profile field already accepted any https host, and `_validate_token_url` fail-closes only cleartext-to-public). Worth a note for Vault-ACL authors: write on a target credential now also controls the mint-destination host. No behavior change requested.
- *Safety floor version-pinned to `0.1.0`*: intentional, mirrors the `vmware_rest` 9.0 gate. Ingesting under a different `--version` leaves `validate` at the default `caution` heuristic instead of `safe`; launch/gate stay `caution` regardless. Awareness only.

Local: 148 passed across the two touched test files (`test_connectors_profiled_auth.py`, `test_connectors_meho_automation.py`). CI re-running on the new head.

damir-topic:
**Verdict:** APPROVE

Re-review of head `3655aacc` (fix commit `3655aacc`, "genericize realm segment in profiled-auth credential fixture", on top of `ca40b636`). The sole blocking finding from the prior REQUEST_CHANGES is fixed and verified; every QC pointer re-checks clean.

### Blocking finding — RESOLVED
1. **Lab-realm leak in the new public-repo test fixture — fixed.** `backend/tests/test_connectors_profiled_auth.py:1069` `_CRED_ISSUER_URL` is now `https://realm.example.test/realms/example/protocol/openid-connect/token` (was `.../realms/evba/...`). Whole-diff grep of added lines for lab identifiers (realm names, host/IP/VLAN, codenames, `evba`, `hetzner`, `rke2`, `10./192.168`, `.lab`) is clean — every host string in the diff is a generic placeholder (`realm.example.test`, `keycloak.test`, `vault.test`, `attacker.example.com` as the SSRF case, `meho-automation` the public k8s service/audience name). Functionally inert as claimed (the host was already the placeholder; the test asserts only on the constant it defines). TruffleHog Secret Scan lane green.

### QC pointers re-verified
2. **Credential-sourced `token_url` hook** — `_oauth2_login_path_from_secret` reuses `_validate_token_url` **verbatim** (same fail-closed SSRF rule: https for public hosts, plaintext only for cluster-internal/private), presence tested with `is not None` so a declared-but-blank value fails closed rather than falling through. A credential bundle without `token_url` (keycloak) is byte-identical to today. New tests cover dial-for-the-mint, override-of-profile, and fail-closed-on-public-http; the existing no-log/no-span test stands. Threat model: `token_url` and `client_secret` live in the same per-target Vault credential written by the same target owner, so this stays inside the operator trust boundary and matches the #3571 profile field's any-https latitude — awareness item, not a defect.
3. **Ingest safety floor reuses the shared mechanism** — registers via `register_ingest_safety_floor(product, impl_id, floor)` against the unchanged `operations/ingest/safety_floors.py` registry, mirroring `vmware_rest/ingest_safety.py` exactly (version-scoped guard + `model_copy`). Not a parallel bespoke mechanism.
4. **Tiers exact + documented** — launch (`POST /api/v1/runs`) + gate (`POST .../gates/{node_id}/decision`) → `caution`, `requires_approval=False` (no backplane park; add-on gate nodes are the human control per T-dec-4); validate (`POST .../validate`) → `safe`. `docs/codebase/connectors-meho-automation.md` carries the tier table and the launch-without-park security note; `connector-auth-coverage.md` carries the credential-`token_url` row with the SSRF note.
5. **Boot guards / v2 resolution / synthetic ingest genuinely exercised** — synthetic 3-route OpenAPI uses generic schema names (never the real spec), lands exactly the 3 ops staged/disabled with the pinned tiers through the real ingest pipeline; `stamp_catalog_profiled_connectors` + `resolve_connector` prove the `(mehoauto, 0.1.0)` fingerprint resolves to the stamped `ProfiledRestConnector_mehoauto_0_1_0`. MCP surface conformance unchanged.
7. **DCO + Co-Authored-By** on both commits; PR body ends with the Generated-with line; no CLAUDE.md/AGENTS.md edits.

**Local verification (throwaway checkout of the head):** `test_connectors_meho_automation.py` + `test_connectors_profiled_auth.py` + `test_operations_ingest_catalog.py` = **148 passed**; boot-stamp + profile-stamp-gate + registry + MCP-surface-conformance + surface-filter + addon-pairing-conformance = **77 passed**; **mypy clean** on the 4 changed source files.

### Non-blocking note — CI
6. `Python unit shard 1/3` is currently **red**, but it is a flake, not a code defect: annotations show **exit code 2** + **"Event loop is closed"** (an async-teardown flake signature, not an assertion failure — that would be exit 1). The prior head `ca40b636` — the entire PR code, differing from this head only by the inert `evba`→`example` test-string change — **passed** shard 1/3, and a one-line test-constant change cannot cause an event-loop-teardown error. The diff adds no background tasks / event loops / unclosed async resources; local runs of the touched + async modules are green with zero event-loop errors. **A CI re-run should clear it** — the merge queue's allgreen gate will require green before merge. Every other lane is green (DCO, CodeQL, Semgrep OSS+SAST, TruffleHog, Go, ruff+mypy, unit shards 2/3 & 3/3, Helm/kubeconform, CLI API snapshot, build image); integration testcontainers still finishing.

APPROVE on the merits — re-run shard 1/3 to satisfy the merge-queue gate before merging.

damir-topic:
**Verdict:** APPROVE

Independent review of the fix at `a6046d55` (throwaway checkout of the PR head; read-only otherwise).

**What I verified**

1. **Root cause is correct and cited.** `apply_safety_floor` returns the proto unchanged when its key is absent (`backend/src/meho_backplane/operations/ingest/safety_floors.py:36`), keyed off the process-global `_FLOORS` dict (`:15`). `_isolate_global_registries` snapshots/restores nine registries but not `_FLOORS` (`backend/tests/conftest.py:721-820`; zero `_FLOORS`/`safety_floor` references in that file). `test_eager_connector_import_registers_vmware_safety_floor` does `safety_floors._FLOORS.clear()` then pops only `vmware_rest` from `sys.modules` and calls `_eager_import_connectors()` (`backend/tests/test_operations_register_ingested.py:414-418`); because `_eager_import_connectors` re-imports via `importlib.import_module` (a no-op for modules still cached in `sys.modules`, `connectors/registry.py`), only vmware's floor re-registers — the `(mehoauto, mehoauto-rest)` floor is gone for the rest of that worker.

2. **Fix is load-bearing (deterministic local repro).** Restoring the parent-commit (`3655aacc`) versions of `ingest_safety.py` + the test and running `pytest tests/test_operations_register_ingested.py tests/test_connectors_meho_automation.py` in one process reproduces the exact CI failure: `2 failed` with `AssertionError: assert ('caution', False) == ('safe', False)`. With the fixed files: `52 passed`. The mehoauto file alone: `11 passed`.

3. **Order-independence proven.** The strictest leak condition — `register_ingested` (which clears `_FLOORS`) running before mehoauto in the same process — passes post-fix. Validate now rides `caution` from either path (the connector floor or the generic POST heuristic both yield `('caution', False)`), so the tier is deterministic regardless of whether the floor registration survived.

4. **Floor mechanism NOT weakened or bypassed.** The fix commit touches only the connector's own floor content, the catalog doc, the connector test, and two docs. It does not touch `safety_floors.py`, `_upsert.py`, or `conftest.py`. The connector floor still never lowers an op below `caution` (`connectors/meho_automation/ingest_safety.py:69-71`), stays idempotent, and remains registered as an import side effect.

5. **Tiers + docs consistent.** launch / gate / validate all pinned `caution`, `requires_approval=False`. `connectors-meho-automation.md`, `connector-auth-coverage.md`, and `catalog.yaml` all updated to match. This matches the locked decision (launch+gate=caution, no park); validate=safe was only the design suggestion, and `caution` executes immediately with no park, so nothing changes operationally.

6. **No leaks / DCO / CI.** Zero lab hostnames, IPs, realm/tenant/codenames in the added lines (the credential fixture uses the RFC-reserved `realm.example.test` placeholder — the point of `3655aacc`). All three commits carry `Signed-off-by` + the `Co-Authored-By: Claude Opus 4.8` trailer. CI on `a6046d55`: 22 success / 10 skipped / 0 failed / 0 in-progress — previously-red `Python unit shard 1/3`, all shards, the `Python (ruff + mypy + pytest)` fan-in, integration testcontainers, and the backplane image build are all green.

**Nits (non-blocking, follow-up material — do not gate this PR)**

1. The underlying test-isolation bug is sidestepped, not fixed: `_isolate_global_registries` still does not snapshot/restore `safety_floors._FLOORS`, and `test_eager_connector_import_registers_vmware_safety_floor` still permanently corrupts the process-global `_FLOORS` for the remainder of its xdist worker. This PR only works because validate's pinned tier now equals the generic POST-heuristic fallback. Any future connector floor that pins an op *below* the generic verb heuristic (e.g. a POST op intended at `safe`) will flake the same way. Worth a follow-up ticket to add `_FLOORS` to the conftest snapshot/restore (or have the vmware test restore the full registry rather than re-importing only vmware).

2. Small coverage note, same root cause: the fix dropped the `has_safety_floor(...)` assertion, so no test now verifies the mehoauto floor is actually wired into the process-global `_FLOORS` — a broken `__init__` side-effect import would go undetected (operationally inert today since floor == heuristic). Fixing nit 1 would let that registration assertion return.

damir-topic:
**Verdict:** APPROVE

Independent review of the delta since the last approved head `a6046d55` — commits `cb33ba90` (enforce meho-automation ingest op allowlist before persistence, T3-F01) and `492d73ff` (OpenAPI snapshot + Go client regen). Head reviewed: **`492d73ffb8c41e06c0c68b5c10d7afea2607dc44`**.

This is a clean, correctly-scoped fix for the T3-F01 HIGH finding. Enforcement is real, early, fail-closed, and generic; the floor mechanism is untouched; the report/API additions are backward-compatible; docs are corrected; no leaks. I found no blocking issues.

## What I verified

**1. Enforcement is real and early — no bypass found.**
- `apply_op_allowlist` runs in `register_ingested.py:445`, filtering `operations` → `allowlist_result.kept` **before** `_prepare_ingested_operations` (`:456`, which applies the safety floor) and before the per-row `upsert_one_operation` (`:647`). Dropped ops are never handed to `_prepare_ingested_operations`, never floored, never upserted — so `enable_connector`'s `is_enabled` cascade cannot reach them.
- Mirrored on the dry-run path in `pipeline.py:922` (before `apply_safety_floor` at `:926` and before the count).
- `apply_safety_floor` has exactly two call sites in `backend/src` (`register_ingested.py:290`, `pipeline.py:926`); `apply_op_allowlist` now precedes both. Every ingest entry point (REST/MCP/CLI) traverses the single chokepoint `IngestionPipelineService.ingest()` (`pipeline.py:620`), which has exactly two branches — `_dispatch_dry_run` → `_run_dry_run` and `_dispatch_real_run` → `_register_all_specs` → `_register_one_spec` → `register_ingested_operations` (`:1074`) — both allowlist-guarded. `jobs.py` runs the same `ingest()` closure. `_prepare_ingested_operations` and `upsert_one_operation` each have a single caller, both inside the guarded function. The typed-register / composite `EndpointDescriptor` inserts (`typed_register.py`, `composite.py`) are a different (typed) path that mehoauto — a generic/ingested connector — does not use.
- **Re-ingest is bounded**: `apply_op_allowlist` runs on every `register_ingested_operations` call, proven by `test_reingest_of_the_wide_spec_stays_bounded_to_three`.
- **Cannot be widened at runtime**: the catalog is shipped package data loaded via `importlib.resources` and `@functools.lru_cache(maxsize=1)` (`catalog.py:596`); there is no catalog-write endpoint. Widening the allowlist requires redeploying a modified image.
- **Matching** is exact on `(METHOD, path)` with the method upper-cased and query string stripped (`op_allowlist.py:_proto_key`), keyed identically to the safety floor. Because the allowlist paths are the add-on's verbatim spec path templates, an unexpected/renamed path is **dropped, not admitted** — the fail-closed direction (see nit 1).

**2. Consistency / no drift.** The catalog allowlist and the connector-owned floor's pinned keys are held equal by a drift-guard test: `op_allowlist_for("mehoauto","0.1.0") == ingest_safety._CAUTION_OPS` (`test_operations_ingest_catalog.py`). A product with no allowlist is byte-identical to today (`apply_op_allowlist` returns `kept=all, dropped=()`); regression covered by `test_product_without_an_allowlist_keeps_every_op`, `test_unknown_product_keeps_every_op`, and `test_only_mehoauto_row_declares_an_op_allowlist`.

**3. Report / API.** `dropped_count: int = 0` and `dropped_ops: list = []` on `IngestionResultModel`, plus the new `DroppedOpModel` and the `op_allowlist` field on `ConnectorSpecEntry` — all additive with empty/nullable defaults, backward-compatible. The `cli/api/openapi.json` + `cli/internal/api/client.gen.go` regen is additive (pointer/`omitempty` Go fields; only a docstring line changed on `IngestionResultModel`) and the **CLI API snapshot freshness** check is green, confirming no hand edits. The result is a bounded operator-plane response, not a set-shaped `call_operation` payload — no JSONFlux/handle contract is affected.

**4. Tests (run locally in a throwaway worktree on the head, sqlite fixture).**
- The three affected files together — `test_operations_ingest_op_allowlist.py` + `test_operations_ingest_catalog.py` + `test_connectors_meho_automation.py`: **99 passed**. This includes: `op_allowlist` (5) covering keep-all / unknown-product / drop-outside / normalisation / empty-input; the catalog drift-guard + `op_allowlist` validators (empty/duplicate/relative-path/extra-field rejected, drift guard equal to `_CAUTION_OPS`, only-mehoauto-declares); and the DB-backed meho_automation tests.
- The ~34-route synthetic fixture mirrors the add-on route families incl. tenant/site/env_spec/deployment/artifact/blueprint/environment/adoption `DELETE`s, `POST /api/v1/fleet/import`, `DELETE /api/v1/runs/{run_id}`; asserts exactly 3 persisted, all `caution` + `requires_approval False` + `is_enabled False`, dropped count/list correct, re-ingest stays 3. The floor-registration assertion is order-robust (calls the idempotent `register_safety_floor()` first, does not rely on import order).
- `test_operations_register_ingested.py`: **41 passed** (no regression from the structural changes).

**5. Docs.** The false "the three ops land staged / disabled" claim is gone; the recipe now states full-spec ingest is bounded by the allowlist, that `token_url` must be set to the realm's token endpoint, that the issuer host must be on the SSRF allowlist, and that target TLS verify must stay on. **Leak check**: no IPs, lab hostnames/domains, GUIDs, realm/tenant names, or codenames anywhere in the diff (docs use generic placeholders like "the realm's `.../protocol/openid-connect/token`").

**6. Hygiene.** Both commits carry `Signed-off-by` (DCO) and `Co-Authored-By: Claude Opus 4.8`. No CHANGELOG entry; no CLAUDE.md/AGENTS.md changes. `Python lint + typecheck (ruff + mypy)` green. `_upsert.py` (the floor interlock `_apply_safety_metadata`) and `safety_floors.py` are untouched — the floor mechanism is not weakened. `edit_op` is untouched (tracked in #3604), and since dropped ops are never persisted, `edit_op` cannot resurrect them.

## Nits (non-blocking)

1. Matching is exact-string on the add-on's verbatim `(method, path)` templates, which is correctly fail-closed (a renamed/unexpected path is dropped, never admitted). The trade-off: if the add-on ever renames a path parameter in its published `/openapi.json` (e.g. `{blueprint_id}` → `{id}`), the legitimate op would be silently dropped and an ingest could land **zero** enabled ops. The synthetic test fixture reuses the same `_LAUNCH/_VALIDATE/_GATE` constants as the allowlist source, so it cannot detect divergence from the add-on's *real* spec (inherent to the "no vendored spec" design, already documented). Optional future hardening: emit a warning when an allowlisted product ingests to 0 kept ops.

## CI (head 492d73ff)

All check-runs completed and green: **22 success, 10 skipped, 0 failures / cancelled / timed-out** (in-progress = 0). Skipped lanes are the usual conditional ones for a code-only change (migrations, SonarCloud coverage, license checks, chart package/push, rke2 deploy). `Python (integration testcontainers)` = **success**; `Python unit shard 1|2|3/3`, `Go`, `ruff + mypy`, `CLI API snapshot freshness`, `DCO`, `Semgrep`, `CodeQL`, `TruffleHog`, and the backplane image build all green.

PR #3627, opened 2026-09-14
fix(image): install the hatchling build backend from the lock (--no-build-isolation) so cold-cache image builds need no fresh PyPI resolve

### LINKED ISSUE #3626: fix(image): arm64 image build fails on cold cache — uv build-isolation resolve of hatchling cannot connect under QEMU
## Summary

The `image` workflow fails on `main` at the **Build image to quarantine
digest** step whenever the `linux/arm64` builder stage is built cold. The
`linux/amd64` builder and both runtime stages progress normally; only the
emulated arm64 `uv sync --locked --no-dev --no-editable` step (builder
10/11) fails:

```
#41 [linux/arm64 builder 10/11] RUN ... uv sync --locked --no-dev --no-editable
  × Failed to build `meho-backplane @ file:///app`
  ├─▶ Failed to resolve requirements from `build-system.requires`
  ├─▶ No solution found when resolving: `hatchling`
  ├─▶ Request failed after 3 retries in 52.4s
  ├─▶ Failed to fetch: `https://pypi.org/simple/hatchling/`  (index)
  │   / `.../hatchling-1.32.0-py3-none-any.whl.metadata`     (files host)
  ├─▶ client error (Connect)
```

Reproduced on run 34883731114 (commit 9ec60167) across all three attempts.

## Root cause

`build-system.requires = ["hatchling"]` is **not** captured in `uv.lock`, so
building the local `meho-backplane` wheel makes uv perform a **fresh
build-isolation resolution** of `hatchling` over the network. Under
`qemu-user` (arm64 emulated on the amd64 CI runner cluster) that specific
resolve fails to establish a connection (`client error (Connect)`), and it
fails deterministically across retries.

This is **not** a runner/cluster egress problem and **not** a QEMU-version
regression:

- In the *same* failing run, under the *same* emulation: the runtime apt
  layer (`apt-get update`/`install`) completed, and the builder deps step
  (`uv sync --no-install-project`, builder 6/11) downloaded ~199 locked
  wheels from `files.pythonhosted.org` successfully. Only uv's
  build-isolation resolve of the build backend fails — including on the
  same `files.pythonhosted.org` host it had just downloaded from.
- The passing builds (the prior main run and the fix PR build for the
  preceding image change) used the **identical** `tonistiigi/binfmt`
  image digest and `qemu/v10.2.3`. They passed only because their arm64
  builder used a warm buildkit uv-cache mount, so `hatchling` was served
  offline and no network resolve happened. The failing run hit a builder
  whose uv-cache mount was cold, exposing the latent fetch.

The difference between passing and failing runs is buildkit uv-cache-mount
warmth on the scheduled runner, not any code change. In particular the
preceding image change touched only the **runtime** stage's apt
`--only-upgrade` layer; the **builder** stage is byte-identical, so its
layer-cache keys did not move.

## Proposed fix

Remove the fragile network resolve from the emulated stage. The
`meho-backplane` distribution is pure-Python (`py3-none-any`), so its wheel
is arch-independent: build it once on the native `$BUILDPLATFORM` in a
dedicated stage and install the prebuilt wheel into each target-arch venv
with `uv pip install --no-deps`. Installing a built wheel needs neither a
build backend nor a network resolve, so the emulated arm64 stage never
contacts PyPI. The compiled *dependency* wheels (asyncpg, cryptography,
pydantic-core, …) still install for the target arch via the existing locked
`uv sync` (which is unaffected), preserving the per-target-arch invariant
documented in the Dockerfile header.

## Acceptance criteria

- The multi-arch `image` build succeeds with a cold buildkit cache on the
  arm64 builder (no dependency on uv-cache-mount warmth).
- The emulated arm64 builder performs no fresh PyPI resolution of the build
  backend.
- The runtime image is unchanged in content (same venv, same ui/ static
  bundle); no build backend ships in the runtime venv.

damir-topic:
## Root cause

The `image` workflow fails at **Build image to quarantine digest** whenever a
stage must perform a **cache-cold build-isolation resolve of the `hatchling`
build backend**. `build-system.requires = ["hatchling"]` is not captured in
`uv.lock`, so building the local `meho-backplane` wheel makes uv resolve
`hatchling` fresh — a PyPI index query plus a PEP 658 `.whl.metadata` fetch —
and those connections fail to establish from the CI runner cluster:

```
No solution found when resolving: `hatchling`
Request failed after 3 retries in ~50s
Failed to fetch: .../hatchling-1.32.0-py3-none-any.whl.metadata
client error (Connect)  ╰─▶ operation timed out
```

This is **not** QEMU-specific and **not** a general egress outage:

- Reproduced on **native linux/amd64** (an early attempt that only relocated
  the wheel build to `$BUILDPLATFORM` still failed on the native stage), so
  emulation is not the trigger.
- In the same failing runs, `uv sync --locked` downloads 150+ pinned wheels
  from the same host successfully, and the apt layer has egress too — only the
  cache-cold *fresh resolve* fails. uv's locked direct-URL download path is
  reliable; the fresh index/metadata resolve is not.
- Passing builds used the identical QEMU/binfmt version; they passed only
  because their buildkit uv-cache mount was warm, serving `hatchling` offline.

The preceding image change (base-OS apt patch) is **not causal**: it touched
only the runtime stage, leaving the builder stage byte-identical, so its
layer-cache keys did not move. Its merge merely triggered a build that landed
on a cold-cache runner and exposed this latent fragility.

## Fix

Route the build backend through uv's reliable locked-download path and remove
the fresh resolve entirely (two commits on this branch):

- Pin `hatchling` in a non-default `build` dependency group and regenerate
  `uv.lock` (adds hatchling + tomlkit + trove-classifiers).
- In the `wheelbuilder` stage, install only that group with
  `uv sync --locked --no-install-project --only-group build`, then build the
  arch-independent (`py3-none-any`) wheel with
  `uv build --wheel --no-build-isolation` — so no fresh index/metadata
  resolution runs in any stage.
- The target `builder` stage installs the prebuilt wheel with
  `uv pip install --no-deps --offline`.

The runtime venv is unchanged — the build backend lives only in the discarded
`wheelbuilder` stage — and the compiled dependency wheels still install for the
target arch via the unchanged locked `uv sync --no-install-project`.

## Verification

- Local native `docker buildx --target builder` (amd64): the wheel-builder
  installs the build group from the lock and builds the wheel with
  `--no-build-isolation`; the builder installs it `--offline`; Tailwind
  compiles against the installed `ui/` tree. No fresh resolve.
- This PR's own multi-arch image build (run 34890568584) is **green on both
  amd64 and arm64**: amd64 builds the wheel with no resolve, arm64 installs the
  prebuilt wheel offline (`+ meho-backplane==0.1.0.dev0`). Build-only on a PR
  (`push=false`), so the trivy/promote/sign steps are skipped by design.

Note: the underlying CI egress reliability to the PyPI CDN on cold-cache fresh
resolves is worth a separate look; this PR removes the dependency so the build
no longer relies on it.

Closes #3626


damir-topic:
### Correction after the first CI run

The initial commit (relocate the wheel build to `$BUILDPLATFORM`) did **not**
fix it, and its run is the key evidence: the build failed again, but on the
**native linux/amd64** wheelbuilder — no QEMU involved:

```
#38 [linux/amd64 wheelbuilder 8/8] RUN ... uv build --wheel --out-dir /wheels
  ├─▶ No solution found when resolving: `hatchling`
  ├─▶ Request failed after 3 retries in 49.2s
  ├─▶ Failed to fetch: .../hatchling-1.32.0-py3-none-any.whl.metadata
  ├─▶ client error (Connect)
  ╰─▶ operation timed out
```

So the failing operation is uv's **cache-cold build-isolation resolve** of
the `hatchling` backend (fresh index query + PEP 658 `.whl.metadata` fetch),
which fails to connect from the CI runner cluster on **both arches** — not a
QEMU issue. In the same run, `uv sync --locked` still downloaded 157 pinned
wheels from the same host successfully, so the locked direct-URL download
path is reliable; only the fresh resolve fails.

Updated fix: pin `hatchling` in a non-default `build` dependency group and
install it via that reliable locked path, then build the wheel with
`--no-build-isolation` so no fresh resolve runs in any stage. Runtime venv is
unchanged. Verified locally on native amd64 (`docker buildx --target
builder`). Note this is a robust build-side mitigation; the underlying CI
egress reliability to the PyPI CDN is worth a separate look (see #3626).

PR #3621, opened 2026-09-14
feat(secret-broker): Keycloak client-secret source kind (#3619)

damir-topic:
## What

Adds a **client-secret SOURCE** to the secret broker's `keycloak` kind so a confidential realm client's secret can be moved into Vault server-side, alongside the existing Keycloak user-password **sink** (#1578). The `secret.move` op stays the only, approval-gated entry point — no read-and-return op is added.

- New source ref grammar: `keycloak:<target>/<realm>/clients/<clientId>#secret`. The literal `clients/` segment distinguishes it from the user-password sink ref `<target>/<realm>/<username>#password`; both parse under the one kind, dual-dispatched by `build_keycloak_secret_endpoint` on the ref's **structure** (four address segments whose third is `clients`), not its `#field`. The sink parse/behaviour is unchanged.
- `KeycloakClientSecretSourceEndpoint.read_secret` resolves `clientId`→`uuid` via the admin `GET .../clients?clientId=<id>` exact lookup (`_find_client_uuid`) and reads `GET /admin/realms/{realm}/clients/{uuid}/client-secret` through the connector's existing admin-read path (`_get_admin_json`) — reusing `KeycloakConnector`'s admin session-cache and Bearer mint, opening no new HTTP client — then wraps the value in `SecretMaterial`.
- A `#field` other than `secret` on a `clients/` ref is rejected with a clear, value-free error; `write_secret` on a client-secret ref raises (it is a source, not a sink).
- Adds `_CLIENT_SECRET_PATH` to the keycloak reconcile-lane path constants and pins it in the spec-reconcile manifest (`GET`).

## Why

Keycloak **does** serve client secrets over the Admin REST API (`GET .../client-secret` → `{"type":"secret","value":…}`), unlike user passwords, so a broker source is legitimate. The governed add-on launch path needs a confidential realm client whose secret lands in Vault without transiting any human or agent context — the value lives only inside the in-memory `SecretMaterial` between the source read and the sink write.

The value never appears in logs, audit rows, the `OperationResult`, exceptions, or the flight recorder: the source reuses the connector's redacted vendor-call span, `SecretMaterial` redacts its `__repr__`/`__str__`, and the handler returns only `{status, value_sha256, length}` (mirrors the vault source discipline and `redaction.py`).

## How tested

[...]

PR #3645, opened 2026-09-15
docs(connectors): fix meho-automation recipe CLI drift (targets import, --impl, file:// spec, three-field credential)

damir-topic:
## What

Fixes operator-recipe CLI drift in `docs/codebase/connectors-meho-automation.md`
(the meho-automation add-on connector page shipped by #3590). Docs-only.

Each command was re-verified against the built CLI (`cli/bin/meho <verb>
--help`) and the backplane source on `origin/main`; only the wrong parts were
changed.

## Why (the drift)

- **Target registration** used `meho target create --base-url ...` — neither
  the verb nor the flag exists. The verb is `meho targets add` (no `--extras`
  flag), and a plain-`http` in-cluster base URL requires `extras.scheme: http`,
  which only `meho targets import` can write (the HTTP adapter builds
  `{scheme}://{host}[:port]` from `target.extras["scheme"]`, default `https`).
  The profile-backed connector registers **no wildcard**, so the target must
  also carry an operator-asserted `version` to resolve (or even probe), and
  `targets import` spills a top-level `version` into `extras` — so the step is
  now a two-move register: `targets add --version` then `targets import
  --update` to patch `extras.scheme`. Added generic SSRF-allowlist +
  NetworkPolicy prerequisites.
- **Ingest flag** was `--impl-id`; the real flag is `--impl`.
- **Spec source** curled the add-on's in-cluster `http://.../openapi.json`. A
  server-side `--spec` URL fetch is **https-only** (SSRF guard in
  `operations/ingest/api_schemas.py`) and the add-on's ClusterIP is unreachable
  from an operator workstation, so the step now takes the add-on's committed
  OpenAPI document (or its export script) at the deployed ref and passes
  `--spec file://...` (read + uploaded client-side). The op-allowlist still
  persists exactly three ops regardless of the document's width.
- **Credential bundle**: now states explicitly that the shipped
  `mehoauto/0.1.0` profile requires three Vault fields — `client_id`,
  `client_secret`, `token_url` (`auth.secret_fields`) — and that the secret
  should land server-side via the secret broker's Keycloak client-secret
  source (#3621), not through an operator's shell.
- **Launch body**: the doc claimed `inputs` (plus `name`) alone suffice, but
  the add-on's run-launch schema requires more than that; softened to "the
  request body follows the add-on's run-launch schema; the backplane forwards
  it verbatim", keeping the caution / no-park statement.
- **Dispatch**: `meho operation call` takes `<connector_id> <op_id>` as
  **positional arguments**, not `--connector-id` / `--op-id` flags.

Refs #3578


damir-topic:
**Verdict:** REQUEST_CHANGES

Independent read-only review of the docs-only recipe fix (head `f5ef04e3`, one file, +71/-25, Refs #3578). Most of the correction is accurate and well-grounded against the shipped CLI and `origin/main` — but the new `meho targets import` example in step 1(b) does not itself run against the CLI, which is exactly the drift class this PR sets out to eliminate.

## Blocking

1. **Step 1(b)'s `automation-addon-scheme.yaml` omits `product` and `host`, so the recipe fails before it reaches the backplane.** `meho targets import` validates every YAML entry through `parseTargetsYAML` (`cli/internal/cmd/targets/import.go:331-342`), which requires `name`, `product`, **and** `host` on every entry *unconditionally* — including `--update` mode — and aborts before any HTTP call. Running move (b) verbatim against the built CLI:

   ```
   $ meho targets import --update automation-addon-scheme.yaml
   meho: unexpected_response: entry "automation-addon": missing or non-string `product` field   (exit 4)
   ```

   Fix: add `product: mehoauto` and `host: meho-automation` to the entry. Verified live — with both present the parser passes and proceeds to the PATCH.

## Nits

2. Consequent to the fix above: on `--update`, `entryToUpdateBody` strips only `name`/`product`, **not** `host` (`import.go:375-382`), so the now-required `host` also lands in the sparse PATCH body — the PATCH touches `host` + `extras`, not "only the extras column" as 1(b) states. It re-asserts `host` to the same value so it is harmless, and the companion claim "leaves the version set in (a) intact" is correct (version is not in the YAML). Consider softening the "touches only the extras column" wording once `host` is added.

## Verified correct (for the record)

- **CLI verbs/flags** (against `cli/bin/meho`): `targets add <name>` positional + `--product/--version/--host/--port/--secret-ref`; `targets import --update` = sparse PATCH via `model_dump(exclude_unset=True)`; `connector ingest --impl` (not `--impl-id`) + `--spec file://` (read+uploaded CLI-side); `connector enable <connector_id>` positional; `operation call <connector_id> <op_id>` positional with `--target`/`--params` flags.
- **Backplane claims on `origin/main`**: HTTP adapter builds `{scheme}://{host}[:port]` from `extras.scheme` default `https` (`connectors/adapters/http.py` `_effective_scheme`/`_base_url`); server-side `--spec` URL fetch is https-only, `file://`/`docs:` read+uploaded CLI-side (`operations/ingest/api_schemas.py:107-119`); `mehoauto/0.1.0` profile `auth.secret_fields` = client_id/client_secret/token_url, fail-closed (`connectors/profiles/meho_automation_minimal.yaml`); catalog op-allowlist = exactly the three launch/validate/gate `(method,path)` pairs, dropped before persistence (`operations/ingest/catalog.yaml:421-427` + `op_allowlist.py`); profiled connector synthesised with `supported_version_range=">=0.1.0,<1.0"` (no wildcard) so a version-less target cannot resolve (`connector_registration.py::synthesise_profiled_class` + `derive_supported_version_range`).
- **Hygiene**: no leak (only generic Keycloak "realm" wording + a `<realm>` placeholder token — no host/IP/tenant/codename/GUID/vault-item names); single docs file — no CHANGELOG, no CLAUDE.md/AGENTS.md; DCO `Signed-off-by` + `Co-Authored-By` trailers present; CI = 29 check-runs, all completed, none failed.

damir-topic:
Thanks for the review — both points addressed in `626128dd`.

**(1) BLOCKING — step-1(b) YAML missing `product`/`host`.** Correct. `parseTargetsYAML` (`cli/internal/cmd/targets/import.go:331-342`) requires `name`, `product`, and `host` on every entry unconditionally, before any `--update` branching, so the example failed to parse. Added `product: mehoauto` and `host: meho-automation` (repeating the values set in step-1(a)) plus a comment noting they are required even with `--update`.

Verified against the built CLI:

```
# fixed YAML
$ meho targets import --update automation-addon-scheme.yaml --dry-run
Plan: 0 to create, 1 to update, 0 to skip
  UPDATE  automation-addon        # exit 0

# pre-fix YAML (name + extras only)
$ meho targets import --update old.yaml --dry-run
meho: unexpected_response: entry "automation-addon": missing or non-string `product` field   # exit 4
```

**(2) NIT — host is re-asserted.** Correct: `entryToUpdateBody` strips only `name`/`product`, so `host` stays in the PATCH body. Softened "touches only the extras column" to "a sparse update that leaves the version set in (a) intact (name/product are stripped as immutable; host is re-asserted to the same value)".

damir-topic:
**Verdict:** APPROVE

Re-review of fix commit `626128dd` ("docs(connectors): fix targets import --update example per review"). Both prior findings are resolved; nothing else changed.

**Blocking #1 (resolved).** Step 1(b)'s `automation-addon-scheme.yaml` now carries `name`, `product: mehoauto`, and `host: meho-automation`, with an inline comment noting all three are required on every import entry (even `--update`) and that `product`/`host` must repeat the values from (a). Verified live against `cli/bin/meho`:

```
$ meho targets import --update automation-addon-scheme.yaml --dry-run
Plan: 0 to create, 1 to update, 0 to skip
  UPDATE  automation-addon
(exit 0)
```

The parser now passes and the sparse UPDATE plan builds correctly.

**Nit #2 (resolved).** The "touches only the extras column" wording is replaced with an accurate description: "a sparse update that leaves the version set in (a) intact (name/product are stripped as immutable; host is re-asserted to the same value)" — matches `entryToUpdateBody` behavior (`cli/internal/cmd/targets/import.go:375-382`).

**Scope / hygiene.** `626128dd` touches only `docs/codebase/connectors-meho-automation.md` (+7/-2 over the prior head; whole branch is that one docs file, +76/-25 vs `origin/main`). No CHANGELOG, no CLAUDE.md/AGENTS.md. No leak in the added lines (only the generic `mehoauto` slug and `meho-automation` service name, both already public). DCO `Signed-off-by` + `Co-Authored-By` trailers present. CI on `626128dd`: 29 check-runs, all completed, none failed.

All CLI-verb and backplane claims verified in the first-pass review remain unchanged by this commit. Ready to merge.

PR #3690, opened 2026-09-16
fix(ingest): keep re-ingest grouping alive for connectors with hyphenated typed group keys (#3685)

### LINKED ISSUE #3685: Task: re-ingest grouping pass crashes when a connector has typed-op groups with hyphenated keys (GroupProposal snake_case re-validation of persisted groups)
## Summary

On a connector that carries **both** ingested OpenAPI descriptors and typed operations, a re-ingest that must run the LLM grouping pass crashes in the grouping transaction (T3) every time the connector already has typed-op groups whose `group_key` is **hyphenated**. The register/upsert transaction (T2) commits first and independently, so the connector stays dispatchable — but T3 raises, rolls back, and the ingest job is reported as a flat `failed`, misleading operators into thinking the whole ingest failed. Ungrouped ingested ops on such a connector can then never be grouped by re-ingest.

Root cause: the partial-regrouping step projects **every already-persisted** `OperationGroup` of the connector back through the `GroupProposal` pydantic model, whose `group_key` field validator enforces snake_case (`^[a-z][a-z0-9_]*$`). Typed-operation registration never enforces that pattern, so hyphenated typed keys persist fine — until re-ingest re-validates them.

All citations below are as of `origin/main` @ `f4810ac0` (paths relative to repo root).

## Observed (v0.35.1, a lab instance)

Re-ingesting a builtin vSphere 9.0 connector that carries both ingested OpenAPI descriptors and typed operations:

- **T2 (register/upsert) committed correctly** — safety floor applied, `is_enabled` preserved; the connector remained usable/dispatchable afterwards.
- **T3 (LLM grouping, partial re-grouping of ungrouped ingested ops) crashed and rolled back** — a `ValueError` from `GroupProposal` snake_case re-validation of a persisted hyphenated key.
- The job ended **`failed`** (not `degraded`), even though the register work was durable — which reads as "the ingest failed."
- The connector has **7 typed-op groups with hyphenated keys**:
  `vmware-host-usage`, `vmware-host-network-uplinks`, `vmware-host-storage-devices`, `vmware-host-vsan-health`, `vmware-object-collect`, `vmware-tasks-recent`, `vmware-vm-info`.

## Root cause (source citations, `origin/main` @ `f4810ac0`)

1. **The snake_case pattern and where it is enforced.**
   `backend/src/meho_backplane/operations/ingest/_llm_grouping_internals.py:95`
   defines `_GROUP_KEY_PATTERN = re.compile(r"^[a-z][a-z0-9_]*$")`. The
   `GroupProposal` model (`:232`) enforces it on its `group_key` field via
   `@field_validator("group_key")` / `_group_key_must_be_snake_case`
   (`:245-253`), raising `ValueError` on any non-match.

2. **The pattern is re-applied to already-persisted groups, not just fresh LLM proposals.**
   `backend/src/meho_backplane/operations/ingest/llm_groups.py:476-483`
   — in `_resolve_groups_for_pass2`, the partial-regrouping branch
   (`if existing_groups:`) projects **each persisted `OperationGroup` row**
   into `GroupProposal(group_key=row.group_key, name=..., when_to_use=...)`.
   Constructing the model re-runs the snake_case validator against the
   stored key, so a hyphenated typed key raises here. This runs before the
   grouping transaction commits (`llm_groups.py:667`), so the whole T3
   session rolls back.

3. **Typed group registration has no key-format validation.**
   `backend/src/meho_backplane/operations/typed_register.py` validates only
   the `group_key`/`when_to_use` **pairing** (`_validate_when_to_use_pairing`,
   `:775-829`), never the key format; `_lookup_or_create_group` (`:841`) then
   inserts `OperationGroup(group_key=group_key, ...)` (`:907-913`) with no
   pattern check. Indeed `:906` derives the display name via
   `group_key.replace("-", " ")...`, i.e. hyphenated keys are an expected
   input on the typed path. The 7 hyphenated constants live at
   `connectors/vmware_rest/typed_ops.py:109`,
   `typed_ops_host_network_uplinks.py:69`,
   `typed_ops_host_storage_devices.py:102`,
   `typed_ops_host_vsan_health.py:62`,
   `typed_ops_object_collect.py:85`,
   `typed_ops_tasks_recent.py:74`,
   `typed_ops_vm_info.py:99`.

4. **T2 and T3 commit independently (so exactly what is durable is precise).**
   `backend/src/meho_backplane/operations/ingest/pipeline.py:818-885`
   (`_dispatch_real_run`) runs `_run_register_phase` first — per-spec
   register commits at `register_ingested.py:678` — and only then runs
   `_run_grouping_phase` on a **separate** session (grouping commit at
   `llm_groups.py:667`). A `ValueError` from step 2 propagates before the
   grouping commit, so only the grouping session rolls back; the register
   commit is already durable.

5. **A raised grouping error is reported as a flat `failed`, never `degraded`.**
   `backend/src/meho_backplane/operations/ingest/jobs.py` — `run_ingest_job`
   (`:455`) wraps the pipeline call in `except BaseException` →
   `registry.fail(...)` → status `"failed"` (`fail()` at `:325-352`). The
   terminal `degraded` state already exists (`IngestJobStatus` at `:148`;
   `degrade()` at `:291-323`) but is only reachable from
   `_reconcile_returned_result` (`:535-580`), i.e. when the pipeline
   **returns**. Because the grouping crash raises out of the pipeline call,
   the durable-T2 / failed-T3 outcome collapses to a bare `failed`.

## Why it matters

- **Mixed typed + ingested connectors are first-class** per the MEHO CLAUDE.md postulate 1 ("both kinds expose operations through the same table … the agent does NOT see the difference"). A connector that legitimately mixes both must survive re-ingest.
- **Re-ingest is the documented path to apply safety-floor updates** (the observed run applied the safety floor correctly in T2). If re-ingest can never complete its grouping pass on such a connector, ungrouped ingested ops can never be grouped, and operators lose the discovery grouping the agent flow depends on (pick connector → list operation groups → search → call).
- **A `failed` job after a durable T2 is actively misleading** — the connector is fine and dispatchable, but the operator is told the ingest failed, and the real, recurring, fixable cause (hyphenated persisted keys) is buried in a `ValueError` traceback.

## Proposal

- **(a) Do not re-validate already-persisted group keys through `GroupProposal` on projection.** The projection at `llm_groups.py:476-483` should not put stored keys back through the snake_case validator — either bypass validation for the projected-from-DB path, or normalize hyphens→underscores for the LLM round-trip only (keeping the persisted key authoritative).
- **(b) Make the two write paths agree on key format at the boundary.** Enforce (or normalize) the `group_key` format at typed registration (`typed_register.py`) so both the typed path and the ingest/grouping path share one contract, rather than one path silently accepting what the other later rejects.
- **(c) Make the job outcome reflect "registered, grouping failed" rather than a flat failure.** A grouping-phase crash after a durable register should surface as a distinct terminal outcome (the `degraded` machinery is a natural fit) carrying the register counts and a structured reason, not a bare `failed`.

## Acceptance criteria

- [ ] A unit test with a fixture connector holding a hyphenated typed-op group **plus** ungrouped ingested ops re-ingests and groups the ungrouped ops cleanly (no `ValueError` from persisted-key re-validation).
- [ ] Typed registration and the ingest/grouping path agree on `group_key` format (either both accept hyphens, or both normalize to one canonical form); a test pins the shared contract.
- [ ] The ingest job status distinguishes "T2 registered, T3 grouping failed" from a flat `failed` (durable register counts + structured reason preserved in the terminal state), with a test asserting the distinction.

Parent initiative: #3660
Part of #221

damir-topic:
## Summary

On a connector carrying **both** ingested OpenAPI descriptors and typed operations, a re-ingest that must run the LLM grouping pass (T3) crashed every time the connector already had typed-op groups whose `group_key` is **hyphenated**. Register/upsert (T2) commits first and independently, so the connector stayed dispatchable — but T3 raised, rolled back, and the async ingest job was reported as a flat `failed`, and the ungrouped ingested ops could never be grouped.

This PR keeps re-ingest grouping alive on such connectors and makes the "registered but grouping failed" outcome honest.

## Root cause

The T3 partial-regrouping branch projected **every already-persisted** `OperationGroup` back through the validated `GroupProposal` model, whose `group_key` field validator enforces snake_case (`^[a-z][a-z0-9_]*$`):

- `backend/src/meho_backplane/operations/ingest/_llm_grouping_internals.py` — `_GROUP_KEY_PATTERN` and the `GroupProposal.group_key` `@field_validator` raising `ValueError` on any non-match.
- `backend/src/meho_backplane/operations/ingest/llm_groups.py` — in `_resolve_groups_for_pass2`, the partial-regrouping branch constructed a validated `GroupProposal(group_key=row.group_key, ...)` for each persisted row, re-running the snake_case validator against the stored key.

Typed-operation registration (`typed_register.py`) never enforces that pattern — it even derives the display name via `group_key.replace("-", " ")`, i.e. hyphenated keys are an expected input on the typed path. The `vmware_rest` connector has 7 such groups (`vmware-host-usage`, `vmware-host-network-uplinks`, `vmware-host-storage-devices`, `vmware-host-vsan-health`, `vmware-object-collect`, `vmware-tasks-recent`, `vmware-vm-info`). Register (T2) and grouping (T3) commit on independent sessions, so the `ValueError` rolled back only T3 while T2 stayed durable — collapsing to a bare `failed`.

## Change

**(a) Stop re-validating already-persisted group keys.** New `_project_persisted_groups` helper in `llm_groups.py` builds the Pass-2 `GroupProposal` list via `GroupProposal.model_construct` (validation bypassed). The persisted row is authoritative, so hyphenated keys round-trip verbatim into the Pass-2 prompt and the `group_key -> id` map; no persisted group is renamed. The 7 typed constants are **not** touched (renaming persisted keys is out of scope).

**(b) Distinguish "registered, grouping failed" from a flat failure.** `IngestionPipelineService._dispatch_real_run` wraps the grouping call and re-raises an *unexpected* grouping fault as a new `GroupingPhaseFailedError` carrying the committed register `IngestionResult`. `run_ingest_job` reconciles it to the **existing** terminal `degraded` state — new `error_class="grouping_failed_after_register"`, preserving the durable register counts — instead of a bare `failed`. `LlmClientUnavailable` (503) and `LlmOutputInvalid` (400) are passed through untouched so their operator-facing REST/CLI contracts are unchanged.

Part (b) needed **no migration and no new job-row field**: the async job is an in-memory dataclass, and `degraded` is already wired end-to-end through the REST projection (`_job_to_response`) and the Go CLI `ingest-status` renderer (`renderIngestTerminal`), so the register counts + structured reason are already REST/CLI-visible.

## Tests

- `test_project_persisted_groups_preserves_hyphenated_key` — regression on the projection helper: persisted hyphenated + snake_case keys project without raising and are preserved verbatim, while the validated `GroupProposal(...)` constructor still rejects a hyphenated key (pins the contract).
- `test_run_llm_grouping_reingest_groups_ungrouped_ops_with_hyphenated_typed_group` — end-to-end: a fixture connector with a persisted hyphenated typed group plus 5 ungrouped ingested descriptors re-ingests and groups cleanly (Pass-1 skipped, LLM mocked); the ungrouped ops assign onto the hyphenated group and the persisted key is not renamed.
- `test_grouping_failure_after_register_degrades_not_fails` — a `GroupingPhaseFailedError` from the pipeline flips the job to `degraded` (not `failed`) with `error_class="grouping_failed_after_register"`, preserving the register counts; the dispatchability probe is not consulted.

Verified locally from `backend/`: `uv run pytest` (ingest grouping/jobs/pipeline/API/MCP/register suites) green, `uv run ruff check src/ tests/` and `ruff format --check` clean, `uv run mypy src/` clean for the changed files (only a pre-existing macOS-only `socket.MSG_ERRQUEUE` note in an unrelated file, which passes on Linux CI).

No CHANGELOG, agent-surface, or MCP-tool changes.

Closes #3685


damir-topic:




**Anchor task:** #3685 — re-ingest grouping crashes on connectors with hyphenated typed-op group keys
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `b6001e3b`

> Delivery note: this is a review *comment*, not a formal GitHub review event. The orchestrator posts the formal approval. No inline threads were opened, because the verdict is APPROVE (no blockers) and inline comments would create merge-blocking threads under the "all comments resolved" policy.

### Acceptance criteria
- [x] AC1 — re-ingest groups ungrouped ops even when a hyphenated typed group exists — verified by `test_run_llm_grouping_reingest_groups_ungrouped_ops_with_hyphenated_typed_group` (5 ungrouped ops assign onto the persisted `vmware-host-usage` group; Pass-1 skipped; group not renamed) + local run green.
- [x] AC2 — typed and ingest/grouping paths agree on `group_key` format — the projection now tolerates persisted keys of any shape (persisted row authoritative) while fresh LLM Pass-1 keys stay snake_case-constrained; `test_project_persisted_groups_preserves_hyphenated_key` pins both halves.
- [x] AC3 — job status distinguishes "registered, grouping failed" from a flat failure with durable counts + structured reason — `test_grouping_failure_after_register_degrades_not_fails` asserts `degraded` + `error_class=grouping_failed_after_register` + preserved register counts (`inserted_count==7`) + probe untouched.

### Verification of the review focus points
1. **`_project_persisted_groups` via `model_construct`.** No downstream re-validation or re-serialisation of the projected proposals through the strict model. `render_assign_ops_prompt` reads attributes into a plain dict; `valid_group_keys` is an attribute read; `parse_assignment_response` is pure set-membership against `valid_group_keys` (unknown keys coerced to `"none"`, regex never re-run, no `GroupProposal` reconstruction). The partial-regrouping branch never persists the projected proposals — it only sets `group_id` on descriptors — so no group is renamed and persistence never re-serialises the model. Persisted keys round-trip verbatim (prompt → LLM → membership → `group_key -> id`). Fresh LLM Pass-1 proposals remain strictly validated via `GroupProposal.model_validate` (`parse_proposal_response`). `GroupProposal` has exactly the three fields all supplied to `model_construct`, so no unset-field access risk on the frozen model.
2. **`GroupingPhaseFailedError` + degraded routing.** `run_ingest_job` catches it before the generic `except BaseException` (correct MRO ordering) → `_reconcile_grouping_failure` → `registry.degrade(...)`. REST `_job_to_response` handles `degraded` generically and passes `error_class` through as a free `str | None` — no code change needed. Go CLI `renderIngestTerminal` handles `degraded` generically, overriding its default `errClass` with `st.ErrorClass` when present — `grouping_failed_after_register` renders verbatim, no code change. `LlmClientUnavailable`/`LlmOutputInvalid` pass-through is implemented correctly (`except (LlmClientUnavailable, LlmOutputInvalid): raise` ahead of the wrap; both are in scope — `LlmClientUnavailable` is defined in `pipeline.py`, `LlmOutputInvalid` is imported). `connector_id` and `aggregated` are in scope with correct types in the wrap. Register counts stay durable and visible on the degraded job.
3. **Tests.** Three well-targeted additions; all three ACs met. See M1 for the one gap.
4. **CLAUDE.md / docs.** Agent surface untouched (no MCP tool / CLI verb changes; postulate 1 — mixed typed+ingested connectors as first-class — is honoured by the fix). Docs: see m1 (non-blocking).
5. **Scope notes.** Both acceptable — see "Scope notes" below.

### Integration verification
- [x] `mypy` (changed src) — clean
- [x] `pytest tests/test_operations_ingest_jobs.py tests/test_operations_ingest_llm_groups.py` — 58 passed, 1 skipped
- [x] `pytest tests/test_api_v1_connectors_ingest.py -k "503 or spec_error_family or degraded or grouping"` — 6 passed (503 + 400 route contracts)
- [x] `pytest` register/catalog/pipeline-version/mcp-ingest sweep — 184 passed
- [x] CI rollup on head — all required Python/Go/CodeQL/DCO checks green

### Findings
- ⚠️ **M1 (major, non-blocking under this verdict)** — the pipeline-level seam of the new grouping-outcome contract is not directly exercised by a test. `_dispatch_real_run`'s `except Exception → raise GroupingPhaseFailedError(...)` branch (the *producer* of the new exception, and the aggregated-counts payload it carries) is never executed by any test — `test_grouping_failure_after_register_degrades_not_fails` injects a hand-built `GroupingPhaseFailedError` at the jobs layer rather than driving the pipeline. Likewise the `LlmOutputInvalid` pass-through is pinned only at the route boundary (`test_ingest_spec_error_family_returns_structured_400` mocks `IngestionPipelineService.ingest` entirely). Only the `LlmClientUnavailable` pass-through is genuinely pinned (`test_ingest_returns_503_when_llm_client_unavailable` runs a real ingest with no LLM factory, so it would fail if the guard were removed). The code is correct by reading and by the local/CI runs, so this is a regression-durability gap, not a correctness defect. **Required change (recommended):** add one pipeline-level test that drives `IngestionPipelineService.ingest(dry_run=False)` (or `_dispatch_real_run`) with a register phase that commits and a grouping phase stubbed to raise (a) a generic `RuntimeError` — assert a `GroupingPhaseFailedError` propagates carrying the aggregated register counts and `connector_id`; and (b) an `LlmOutputInvalid` — assert it propagates unwrapped (not a `GroupingPhaseFailedError`).

- m1 — Docs/comment completeness for the second `degraded` cause. `docs/codebase/spec-ingestion.md` describes `degraded` only as "the pipeline ran but persisted nothing dispatchable" (`error_class=ingested_not_dispatchable`); it does not mention the new `grouping_failed_after_register` cause (where the connector *is* dispatchable but grouping rolled back). The same partial framing appears in the `IngestJobRegistry.degrade`/`IngestJob` docstrings, the `IngestJobStatus` module comment, and the Go `renderIngestTerminal` `degraded`-case comment. All are code-correct (the machinery is generic over `error_class`); only the prose enumerates a single cause. A one-line addition naming the second cause would keep the doc/comments honest.

### Docs
- [x] `docs/codebase/spec-ingestion.md` — exists; PR does not contradict it, but omits the new degraded cause (completeness only, non-blocking — see m1).

### Scope notes (focus point 5)
- **`GroupingPhaseFailedError` in `pipeline.py`, not `exceptions.py`** — acceptable. `exceptions.py` (837 lines) is over the file-size budget, but placement in `pipeline.py` is the cohesive choice regardless: the exception is an internal control-flow signal tied to the two-phase register/group structure and carries the register `IngestionResult` that `pipeline.py` already imports; hoisting it into the widely-imported `exceptions.py` would risk an import cycle.
- **No file-size marker on `jobs.py`** — acceptable. `jobs.py` is 730 lines after the change and CI's file-size gate (inside the green Python checks) passed, so it is under budget; no marker is warranted.

- The root cause is pinned precisely and the fix is minimal and surgical: it tolerates persisted keys on the projection path only, leaving the fresh-proposal snake_case contract intact rather than loosening the model globally.
- Reuse of the existing terminal `degraded` state (rather than a new job field/migration) keeps the REST projection and Go CLI renderer working unchanged — verified end-to-end.
- The `except (LlmClientUnavailable, LlmOutputInvalid): raise` guard and the explicit "does not catch `CancelledError`" reasoning show careful attention to preserving the existing 503/400 and job-terminality contracts.

PR #3691, opened 2026-09-16
fix(connectors): enable cascade covers ungrouped descriptors and review reports them (#3681)

### LINKED ISSUE #3681: Task: connector enable cascade skips ungrouped operations (47/1049 untouched, 8 writes stay default-deny)
## Summary

`POST /api/v1/connectors/{id}/enable` (CLI `meho connector enable <id> --confirm`) transitions a connector's operation **groups** from `staged` to `enabled` and cascades `is_enabled=True` onto their child operations. The cascade is scoped by `group_id IN (<transitioned groups>)`, so any descriptor the LLM grouping pass left **ungrouped** (`group_id IS NULL`) is never touched. A full enable of an ingested connector therefore silently leaves its ungrouped write/typed operations at their default-deny state, and there is no read-only way for an operator to enumerate which ones.

## Observed (v0.35.1)

Enabling an ingested vCenter 8.0 catalog (1049 descriptors) with a full `connector enable` landed at **1041 enabled**:

- **47** descriptors were ungrouped (`group_id IS NULL`) and thus skipped by the enable cascade.
- **39** of those were reads (GET/HEAD) that a prior `connector enable-reads` had already enabled — that path is group-agnostic, so it reached them.
- **8** ungrouped write/typed descriptors stayed **default-deny** after a full enable.

No read-only remedy exists to find the 8: disabled ops are absent from `search_operations`, and the ungrouped ops never appear in the `connector review` per-group op enumeration (the human CLI render doesn't even print the ungrouped *count*). The only per-op fix is `connector edit-op <id> <op_id> --enabled`, one at a time — and the operator cannot discover the op_ids to feed it.

## Root cause

File:line citations (`evoila/meho`, v0.35.1 tree):

- **Cascade is group-scoped.** `cascade_is_enabled` filters the bulk UPDATE by `EndpointDescriptor.group_id.in_(group_ids)` — `backend/src/meho_backplane/operations/ingest/_internals.py:572` (fn at `:541`). Rows with `group_id IS NULL` match no `group_id IN (...)` clause, so they are never updated.
- **Enable passes only transitioned group ids.** `POST /{id}/enable` → `enable_endpoint` (`backend/src/meho_backplane/api/v1/connectors_ingest.py:1495`) → `ReviewService.enable_connector` (`backend/src/meho_backplane/operations/ingest/service.py:809`) → `_transition_connector` (`service.py:1052`), which computes `transitioned_ids = [group.id for group in transitionable]` (`service.py:1096`) and hands only those to `cascade_is_enabled`. Ungrouped descriptors belong to no group, so they are outside the id set by construction.
- **`enable-reads` is group-agnostic (the asymmetry).** `bulk_enable_read_ops` (`_internals.py:593`) has **no** `group_id` predicate — it flips `is_enabled=True` on every `source_kind='ingested'`, GET/HEAD, currently-disabled row (`method.in_(READ_HTTP_METHODS)` at `_internals.py:630`). Route `POST /{id}/enable-reads` → `enable_reads_endpoint` (`connectors_ingest.py:1529`) → `enable_reads` → `bulk_enable_read_ops` (`service.py:897`). So `enable-reads` reaches ungrouped reads while `enable` does not reach ungrouped writes — the two commands disagree on scope.
- **Review payload can't surface the gap per-op.** `_render_payload` only files an op under a group when `op.group_id is not None and op.group_id in ops_by_group` (`service.py:493-494`); ungrouped ops are never placed in any `groups[].ops` list. The payload carries a single reconciling integer `ungrouped_op_count` (`service.py:554`) but no enumeration and no enabled/disabled breakdown, and the human CLI render never prints even that integer (`cli/internal/cmd/connector/review.go:136,152` — it prints group count + `TotalOpCount` and iterates `groups[].ops`; `UngroupedOpCount` exists only in the `--json` struct at `cli/internal/api/client.gen.go:3429`).
- **Disabled ops absent from discovery.** `search_operations` filters `EndpointDescriptor.is_enabled.is_(True)` (`backend/src/meho_backplane/operations/_search.py:82,270`), so the 8 default-deny ops cannot be found through the agent/discovery surface either.

## Why it matters

An operator who follows the documented connector lifecycle (`ingest → review → enable`) reasonably believes a full `enable` enabled the whole catalog. It does not: any descriptor the grouping pass left ungrouped stays behind, and for write/typed descriptors that means silently default-deny with no signal. `enable-reads` and `enable` disagree on scope (one covers ungrouped rows, the other doesn't), and no read-only surface lets the operator see the gap or list the affected op_ids — the only fix is a blind per-op `edit-op` for op_ids they have no way to enumerate.

## Proposal

A menu, not a spec — the fix belongs in the enable route and/or the review payload, never on the agent surface:

- Make the enable cascade also cover ungrouped descriptors — e.g. widen the enable path so it flips rows matching `connector scope AND (group_id IN (<transitioned groups>) OR group_id IS NULL)`, keeping the existing operator-override exclusion.
- And/or surface ungrouped ops in the review payload explicitly — an `ungrouped: N (M enabled)` line (and ideally an enumeration of the ungrouped op_ids + their `is_enabled`) so an operator can see and decide, rather than reconciling a bare count that the CLI doesn't even print.
- Consider making `enable` and `enable-reads` agree on scope so a connector can't end up half-enabled in a way neither command reports.

## Acceptance criteria

- [ ] A full `connector enable` either covers ungrouped descriptors, or the `connector review` output reports them explicitly (count **and** how many are enabled) so an operator can act — the two commands no longer disagree silently on scope.
- [ ] A unit test with a fixture containing ungrouped descriptors (`group_id IS NULL`), including at least one ungrouped write/typed row, asserts the chosen behaviour (covered by enable, or reported by review).
- [ ] A docs line in the connector lifecycle doc records how ungrouped descriptors are handled by `enable` / `enable-reads` / `review`.

Parent initiative: #3660
Part of #221

damir-topic:
## Summary

A full `connector enable` now enables the connector's **ungrouped** descriptors
(`group_id IS NULL`) too, and `connector review` reports the ungrouped op count
so an operator can see any the grouping pass left out. The enable/disable
cascade previously skipped ungrouped rows, silently stranding ungrouped
write/typed ops at their default-deny state with no read-only way to find them.

## Root cause

`POST /api/v1/connectors/{id}/enable` transitions operation groups
`staged → enabled` and cascades `is_enabled=True` onto operations, but the
cascade was scoped by `group_id IN (<transitioned groups>)`:

- `cascade_is_enabled` filtered the bulk UPDATE by
  `EndpointDescriptor.group_id.in_(group_ids)`, so `group_id IS NULL` rows
  matched no clause and were never updated.
- `_transition_connector` computes `transitioned_ids` from the transitionable
  groups and hands only those to the cascade; ungrouped descriptors belong to no
  group, so they were outside the id set by construction.
- `enable-reads` (`bulk_enable_read_ops`) has **no** `group_id` predicate, so it
  reached ungrouped reads while `enable` did not reach ungrouped writes — the two
  commands disagreed on scope.
- The `connector review` human render never printed the ungrouped count, so
  there was no read-only surface to see the gap.

Observed on a vCenter 8.0 catalog (1049 descriptors): 47 ungrouped, of which 39
reads had been enabled by a prior group-agnostic `enable-reads` and 8 ungrouped
writes stayed default-deny after a full enable.

## Change

- `cascade_is_enabled` and `operator_disabled_op_ids` gain an `include_ungrouped`
  flag. The enable and disable cascades pass it, so both now cover the
  connector's ungrouped rows (`group_id IS NULL`) within the same
  `(product, version, impl_id, tenant_id)` scope — matching `enable-reads`. The
  override-exclusion lookup is widened to ungrouped ops so a per-op
  `is_enabled=False` override on an ungrouped op still sticks through the
  cascade. Disable is kept symmetric so a full disable locks down exactly what a
  full enable opened (otherwise enabling then disabling would strand ungrouped
  writes enabled). The builtin/platform_admin scoping from #3646 is untouched.
- The CLI `connector review` human render now prints the ungrouped op count
  using the existing `ungrouped_op_count` payload field (previously it was not
  shown even though `--json` carried it).
- Docs: the connector lifecycle doc (`docs/codebase/spec-ingestion.md`) records
  that enable/disable cover ungrouped descriptors and review reports them.

No API schema change — the generated CLI client (`cli/api/openapi.json`,
`cli/internal/api/client.gen.go`) is deliberately untouched (verified: the CI
`make snapshot-openapi && make generate && make cli-docs` drift check is clean).
Surfacing an ungrouped *enabled* sub-count in the review payload would have added
an API field, forcing a regen of those generated files, which are concurrently
being modified by another open PR; that was left out to keep this change
conflict-free. The core gap is closed by the cascade fix itself.

## Tests

- `test_enable_connector_cascades_to_ungrouped_descriptors` — fixture with
  grouped ops plus ungrouped read, write, and typed (`method NULL`) descriptors;
  asserts a full enable flips all of them and the audit `ops_cascade_count`
  reflects grouped + ungrouped.
- `test_enable_connector_respects_operator_override_on_ungrouped_op` — a per-op
  `is_enabled=False` override on an ungrouped op survives the enable cascade.
- `test_disable_connector_cascades_to_ungrouped_descriptors` — a full disable
  also turns ungrouped ops off.
- `test_review_payload_reports_ungrouped_descriptor_count` — the review payload
  reports ungrouped reads and writes via `ungrouped_op_count`.
- Go: `TestPrintReviewTableRendersUngroupedCount` / `...OmitsUngroupedLineWhenZero`
  — the CLI render prints the ungrouped line when there are ungrouped ops and
  omits it otherwise.

Local: `ruff check`, `ruff format --check`, `mypy src/` (clean on changed
files), the review + connectors-ingest + enable-reads + MCP-admin pytest
modules, and `go build` / `go test` / `gofmt` / `go vet` on the CLI all green.

Closes #3681


damir-topic:
> verdict is advisory prose, **not** a GitHub review; `reviewDecision`
> is unchanged and branch protection is not satisfied by this comment.
> The orchestrator posts the formal approval.

**Anchor task:** #3681 — connector enable cascade skips ungrouped operations
**Surface area:** backend + devops (Go CLI render)
**Verdict:** APPROVE
**Head SHA reviewed:** `21828cd9`

### Acceptance criteria
- [x] A full `connector enable` covers ungrouped descriptors (chose the "covered by enable" arm of the AC menu), and `connector review` additionally reports the ungrouped count — the two commands no longer disagree silently on scope. Verified in code (`cascade_is_enabled` + `_transition_connector`) and by `test_enable_connector_cascades_to_ungrouped_descriptors`.
- [x] Unit test with an ungrouped fixture including a write and a typed (`method NULL`) row asserts the enable cascade covers them — `test_enable_connector_cascades_to_ungrouped_descriptors` (read + write + typed), plus override-respect and disable-symmetry tests. All green.
- [x] Docs line in the connector-lifecycle doc records how ungrouped descriptors are handled by `enable`/`enable-reads`/`review` — `docs/codebase/spec-ingestion.md` T4.

### Blast-radius walk (the review's focus)
- **`cascade_is_enabled` callers (3):** `enable_group` (per-group) correctly leaves `include_ungrouped` at its `False` default — a single-group enable must not sweep ungrouped rows; both `_transition_connector` branches pass `include_ungrouped=True`. The new `False` default preserves every pre-existing call's behaviour.
- **`operator_disabled_op_ids` callers (2):** `enable_group` default `False`; `_transition_connector` enable branch `include_ungrouped=True`, so a per-op `is_enabled=False` override on an ungrouped op is honoured (proven by `test_enable_connector_respects_operator_override_on_ungrouped_op`).
- **Ungrouped-cascade scope:** the `group_id IS NULL` predicate is ANDed with `product/version/impl_id` **and** the pre-existing `tenant_id` predicate in the same `UPDATE` (builtin → `tenant_id IS NULL`, tenant → `= tid`). The new `ungrouped_op_ids` helper applies the identical tenant predicate. No cross-tenant / cross-scope leak.
- **#3646 builtin/platform_admin write scoping:** untouched — `_transition_connector` resolves via `_resolve_writable_scope` (which applies `_require_platform_admin_for_builtin`) before any cascade, and the cascade only ever acts within the resolved scope. The diff does not touch the resolution/gate methods.
- **Disable symmetry:** `disable_connector` passes `include_ungrouped=True` with `excluded_op_ids=[]` (no override consultation, as documented). No `disable_group` exists, so there is no per-group disable asymmetry. Because both connector-level branches are symmetric and `enable_connector` accepts a `disabled` source, a `disable`→`enable` round-trip cleanly remediates connectors already stranded before this fix.
- **`enable-reads` parity:** the cascade now reaches the same group-agnostic ungrouped remainder `bulk_enable_read_ops` already covered.

### Predicted vs. actual diff
- ✅ `_internals.py` — `include_ungrouped` flag on `cascade_is_enabled` + `operator_disabled_op_ids`, new tenant-scoped `ungrouped_op_ids` helper.
- ✅ `service.py` — both `_transition_connector` branches opt in; docstrings updated.
- ✅ `test_operations_ingest_review.py` — 4 new tests + ungrouped fixture helper.
- ✅ `review.go` / `connector_test.go` — human-render ungrouped line + 2 tests.
- ✅ `spec-ingestion.md` — docs line.
- 🕳️ Missing: none. Generated CLI client (`cli/api/openapi.json`, `cli/internal/api/client.gen.go`) is correctly **not** touched — those fields (`ungrouped_op_count` / `UngroupedOpCount`) already exist, and PR #3687 concurrently owns those generated files, so deferring an `ungrouped_enabled` sub-count is the right call to stay conflict-free.

### Integration verification (all in a worktree at head `21828cd9`)
- [x] `gofmt -l` (changed files) — clean
- [x] `go vet ./internal/cmd/connector/` — clean
- [x] `go test ./internal/cmd/connector/...` — pass (incl. both new tests)
- [x] `ruff check` (changed src) — clean
- [x] `ruff format --check` (3 changed files) — clean
- [x] `pytest tests/test_operations_ingest_review.py` — 75 passed (incl. all 4 new)
- [x] `mypy` (changed src) — clean
- [x] CLI drift: `snapshot-openapi`/`generate`/`cli-docs` reasoned clean — `review.go` is a runtime human render (no golden files; `cli-docs` builds from the command tree, not sample output), and no generated/`openapi.json` files changed.

### Docs
- [x] `docs/codebase/spec-ingestion.md` — updated (T4 lifecycle).

### Adjacent improvements (non-blocking, not a required change)
- Re-running `connector enable` on an already-fully-enabled connector is idempotent (no group transition → no cascade), so a bare re-enable does not retroactively sweep ungrouped ops stranded by a pre-#3681 enable. The intended remediation for such a connector is a `disable`→`enable` round-trip, which works precisely because the disable cascade was kept symmetric. This matches the documented idempotency; an operator-facing note ("to remediate an already-enabled connector, round-trip disable→enable") could be worth a line in a future docs pass.

### What worked well
- Tenant/scope discipline: the ungrouped cascade reuses the existing `(product, version, impl_id, tenant_id)` predicate and the new helper mirrors it exactly — no cross-tenant exposure introduced.
- Deliberately staying off the generated CLI client to avoid a regen collision with in-flight #3687, with the trade-off (deferred enabled sub-count) stated plainly in the PR body.
- The disable-symmetry decision is the load-bearing detail that makes remediation of pre-existing stranded connectors possible; it is both implemented and documented.

No agent-surface change (CLAUDE.md postulate 5): `connector enable`/`review` are operator-plane CLI/REST, not the 25-tool working surface; vendor specifics stay in `op_id`.

PR #3698, opened 2026-09-16
fix(recorder): redact OVF PropertyParams values in captured deploy bodies

damir-topic:
## Summary

The flight-recorder trace of a `vmware.composite.vm.deploy_from_library`
dispatch captured the vAPI OVF deploy POST body
(`POST /api/vcenter/ovf/library-item/{item}?action=deploy`) with its
`additional_parameters[]` `PropertyParams` block — `properties[] {id, value}`
— **in plaintext**, and marked the span `redaction_uncertain=false`
(agent/same-tenant-operator readable). OVF property values are configuration
inputs that commonly carry appliance credentials, so they must never
round-trip into the 30-day trace store.

This adds a structural redaction rule to the flight-recorder body engine that
redacts every OVF `PropertyParams` property value on both request and response
bodies, keyed on the `PropertyParams` type discriminator, while leaving
non-secret fields (network mappings, name, placement) intact.

## Root cause

The recorder redacts captured bodies through
`redaction/flight_recorder/bodies.py::redact_body`, which applies two controls:
the per-connector declared dotted-path globs
(`_apply_path_redaction`) and, underneath, the shared credential-shape net
(`_content.py::scrub_content`). Neither reaches the property values:

- **The shape net is value-shape based, not key-name based.** `_content.py`
  matches only credential *shapes* (`authorization_header`, `bearer_token`,
  `jwt`, `api_key`, `kubeconfig`). A plaintext OVF property value has no
  distinguishing shape, so it survives every rule.
- **The op-family hard-exclusion doesn't fire.** `families.py::classify_body_exclusion`
  keys on op id / tags / class; a deploy op is neither credential/session/token
  nor destructive, so its body is recorded in full.
- **A positional glob cannot address it.** The property value lives at
  `deployment_spec.additional_parameters[<i>].properties[<j>].value`, where
  `additional_parameters[]` is a type-discriminated union holding a mix of
  subtypes at *dynamic* indices; only the `PropertyParams` member carries
  secrets, and a static dotted-path glob cannot condition on the sibling
  `type == "PropertyParams"` discriminator. The wire body is built by
  `connectors/vmware_rest/composites/_write.py::_build_ovf_deploy_body`
  (`_write.py` ≈ L2918, `_OVF_PROPERTY_PARAMS_TYPE`) and dispatched raw through
  `_write_sub_op` (≈ L1258) → `HttpConnector`, where the recorder captures it
  verbatim.

Net: the property values escaped every heuristic and reached the trace store
with `redaction_uncertain=false`.

## Change

Chose option (a) — a structural rule in the recorder redaction, the smallest
correct fix. `bodies.py` gains `_redact_property_params`, a non-mutating walk
that finds any object shaped `{type: "PropertyParams", properties: [...]}` and
replaces every property `value` with a new `OVF_PROPERTY_VALUE_MARKER`, keeping
the `id` and all sibling fields. It runs inside `redact_body` between the
path-glob pass and the shape net, so it covers **both** request and response
bodies and every connector automatically, with no per-connector config. All
property values are redacted (not just password-looking ids): replay analysis
does not need OVF property inputs, and over-redaction here is the safe
direction. The redaction is *certain* — the trace stays agent-readable with the
values blanked, rather than degrading to operator-only.

**Trade-off vs. reusing the connector-boundary redaction.** The boundary
redaction that produces `params_hash` runs over the logical operation *params*
(`ovf_properties: {k: v}`), not the vAPI wire body the composite builds from
them (`additional_parameters[] {type, properties[]}`); reusing it here would
mean re-mapping the wire shape back to the boundary classifier — more code than
a structural rule, and blind to any generic connector that POSTs a
`PropertyParams` body directly. A whole-body "never record" family marker
(option b) was rejected because it would drop the benign, debug-valuable fields
(name, network mappings, placement) alongside the secrets, and there is no
existing per-sub-op "uncertain" hook the composite could set as a one-liner.

## Tests

- `test_ovf_property_params_values_are_redacted_ids_and_others_survive` — a
  synthetic deploy body with a `PropertyParams` block (fake ids, fake,
  non-credential-shaped values) asserts every value is redacted, the ids and
  non-secret fields (network mappings, name, placement) survive, and no fake
  value leaks anywhere.
- `test_ovf_property_params_redacted_in_response_body_too` and
  `test_ovf_property_params_redacted_via_redact_span_both_directions` — the rule
  fires on response bodies and end-to-end through `redact_span` in both
  directions, staying `redaction_uncertain=false`.
- `test_body_without_property_params_is_unchanged` — regression: a body with no
  `PropertyParams` block (including a top-level `properties[]` list and a
  different `additional_parameters` subtype) round-trips byte-for-byte.

Ran the flight-recorder and vmware_rest composite test packages (751 passed,
23 skipped), `ruff check` + `ruff format`, and `mypy` on the changed files
(clean; the only local mypy finding is a pre-existing macOS-only
`socket.MSG_ERRQUEUE` attribute in an unrelated `icmp.py`, which resolves on
the Linux CI runners). Docs: one line added to the flight-recorder redaction
rules in `docs/codebase/redaction.md` (F2.2). No CHANGELOG, no agent-surface
change.


damir-topic:




**Anchor task:** none linked — directly-filed security fix (no `Closes/Fixes/Refs` in body; noted, not treated as blocking).
**Surface area:** backend
**Verdict:** APPROVE
**Head SHA reviewed:** `4438b21f`

This is a security review of the flight-recorder OVF `PropertyParams` redaction. The fix is correct, closes the reported leak with tests, and does not weaken existing redaction. Two non-blocking test-completeness items are recorded below.

### Security acceptance criteria
- [x] Redacts **every** `properties[].value` inside a `PropertyParams` block **regardless of key name** — keyed on the `type == "PropertyParams"` discriminator + `properties` being an array (`_is_property_params_block`), values replaced with `OVF_PROPERTY_VALUE_MARKER`, ids/siblings kept. Verified by `test_ovf_property_params_values_are_redacted_ids_and_others_survive` (3 differently-keyed values, all redacted).
- [x] Fires on **both request and response** bodies — `redact_span` runs `redact_body` over each; verified by `test_ovf_property_params_redacted_in_response_body_too` and `..._via_redact_span_both_directions`.
- [x] **Nested (real) position** covered — the composite `_write.py::_build_ovf_deploy_body` emits `{deployment_spec:{additional_parameters:[{type:"PropertyParams",properties:[{id,value}]}]}}`; the synthetic test body matches this shape exactly, so the actual captured deploy body is provably scrubbed.
- [x] **Top-level position** handled — the walk recurses through every mapping value / array item, so a root-level block is caught too. Proven by construction; top-level *reach* is exercised by the negative regression (`test_body_without_property_params_is_unchanged`, which places a `properties[]` list and a non-`PropertyParams` `additional_parameters` member at the body root). See M1 — no *positive* top-level test.
- [x] Does not weaken existing redaction — the rule is inserted inside the existing `redact_body` try, between the path-glob pass and `scrub_content`; verdict/uncertain logic untouched. `families.py` is **not** modified — the op-family exclusion is reused via `span.py::classify_body_exclusion`, and **no secret list is duplicated** (the only new constant is the single `_PROPERTY_PARAMS_TYPE` type discriminator).
- [x] Verdict semantics — after the rule fires the body parses cleanly, so the outcome stays `uncertain=False` / agent-readable (asserted in the span test). A body the rule cannot parse (non-JSON / malformed / truncated) still short-circuits to `BODY_OMITTED_MARKER` + `uncertain=True` in `_parse_body` **before** the rule runs, and any fault inside the rule is caught by the existing `except` → fail-closed uncertain.
- [x] Idempotent + crash-safe on malformed `PropertyParams` — `properties` missing/non-array falls through untouched; a property entry without `value` recurses (no fabricated key); a non-string `value` is replaced regardless of type; a second pass is stable. Correct by construction (see m1 — not directly tested).
- [x] Non-mutating — builds new structures; the shared request/response object in the span test is not corrupted across the two `redact_body` calls.
- [x] Docs accurate — `docs/codebase/redaction.md` F2.2 sentence matches the code (redacts every `PropertyParams` property value, keeps id + network mappings/name/placement).
- [x] Public-repo hygiene — tests use only fake `PLACEHOLDER-*` values and synthetic ids (`guest.password`, `resgroup-9`, `network-11`); code/docs/PR body carry no real ids, hostnames, or secrets.

### Integration verification (run in worktree on `4438b21f`)
- [x] `pytest tests/test_flight_recorder_redaction.py` — 133 passed
- [x] `ruff check` (flight_recorder/ + test) — clean
- [x] `ruff format --check` (4 changed files) — already formatted
- [x] `mypy` (bodies.py, verdict.py, __init__.py) — no issues
- [⚠️ not run] full backend suite / migrations / compose — not required for this leaf-pure change; CI covers.

### Findings (non-blocking — verdict is APPROVE)
- ⚠️ M1 (`backend/tests/test_flight_recorder_redaction.py`) — add a **positive** test for a top-level `PropertyParams` block (root-level `additional_parameters[]` not under `deployment_spec`, and/or a bare top-level block). The recursive walk provably handles it and its reach is proven negatively, but a positive regression guard on each claimed position is worth having for a security fix. Not blocking.
- 💡 m1 (`backend/tests/test_flight_recorder_redaction.py`) — add robustness tests for malformed `PropertyParams`: `properties` absent/non-array (no redaction, no crash), an entry lacking `value` (no crash, no fabricated key), a non-string `value` (replaced by marker), and an idempotence assertion. Code is correct today; the test locks the fail-safe. Not blocking.

### Adjacent improvements (non-blocking, not required)
- `_redact_property_params` rebuilds the whole JSON tree on every recorded body, even when no `PropertyParams` is present, ahead of `scrub_content`. Bodies are capped so cost is bounded, but a subtree-only rebuild would skip the allocation on the common path.

### Docs
- [x] `docs/codebase/redaction.md` — verified accurate (F2.2 structural-rule sentence).

### What worked well
- Root cause is precisely stated (positional glob cannot condition on a sibling discriminator) and the structural rule is the minimal correct fix — one pass, no per-connector config, covers generic connectors automatically.
- Test that deliberately uses non-credential-shaped placeholder values proves the *structural* rule fires independently of the shape net — exactly the right thing to assert.

CI note: at review time the CI run was still in progress (Python lint+typecheck, unit shards, TruffleHog, Semgrep, image build). Not waited on.

PR #3720, opened 2026-09-16
feat(vmware): governed inline content fetch for vm.guest.file.read (size-capped, handle-wrapped)

### LINKED ISSUE #3719: feat(connectors): governed inline byte fetch for vmware.composite.vm.guest.file.read (fetch_content)
## Goal

`vmware.composite.vm.guest.file.read` today returns the `FileTransferInformation` (size, POSIX/Windows attributes, one-time transfer URL) and `content_fetch="deferred"` — it never proxies the bytes (inline fetch was explicitly deferred when the read first shipped). This adds the governed, audited inline fetch.

## What

Two new optional params on the existing composite (no new `op_id`, no agent-surface change):

- `fetch_content: bool = false` — when true, MEHO fetches the file bytes server-side over the one-time transfer URL (the same pooled TLS client the write path already uses to reach guest-transfer hosts) and returns them.
- `max_inline_bytes: int` — default 1 MiB (1048576), hard cap 8 MiB (8388608).

Returned content is set-shaped `content_lines` (utf-8 text lines, or 76-char base64 chunks for non-decodable binary, flagged by `content_encoding`), so large content is JSONFlux-wrapped into a result handle and paged via `result_query`. Files over `max_inline_bytes` are **refused** (not truncated) with an error naming the size. Safety tier unchanged (`safe` read). Default (`fetch_content=false`) behaviour is byte-identical to today.

## Security

The one-time transfer URL and its query token are never returned on the fetch path (the result omits `url`) and never logged; flight-recorder spans already strip the URL query. Guest credentials continue to resolve from the target's `secret_ref` only.

## Acceptance criteria

- [ ] `fetch_content=false` (and default) returns the current envelope byte-identically; no bytes fetched.
- [ ] `fetch_content=true` on a small text file returns `content_lines` (utf-8) inline with the metadata envelope; no `url` in the result.
- [ ] `fetch_content=true` on a large file JSONFlux-wraps `content_lines` into a result handle (metadata preserved), drill-in via `result_query`.
- [ ] A non-decodable file returns base64 `content_lines` (`content_encoding="base64"`), round-trips to the exact bytes.
- [ ] A file larger than `max_inline_bytes` is refused with an error naming the size (checked before egress when size is known; defence-in-depth check on the received body).
- [ ] The transfer URL / token never appears in the result or logs.
- [ ] Parameter + response schemas updated; docs page + `llm_instructions` updated; unit tests cover all branches.

damir-topic:
## What

Adds an opt-in, governed, audited **inline content fetch** to the existing `vmware.composite.vm.guest.file.read` composite. Previously the read returned only the `FileTransferInformation` (size, POSIX/Windows attributes, one-time transfer URL) with `content_fetch="deferred"` and never proxied the bytes.

Two new optional params on the existing composite — no new `op_id`, no agent-surface change:

| param | type | default | bounds |
|---|---|---|---|
| `fetch_content` | bool | `false` | — |
| `max_inline_bytes` | int | `1048576` (1 MiB) | `1` … `8388608` (8 MiB hard cap) |

When `fetch_content=true`, MEHO GETs the one-time transfer URL server-side over the **same pooled TLS client the write path already uses** for its PUT (the transfer ticket rides the URL, so no auth header is attached; a `*` placeholder host resolves to the target's host), and returns the bytes.

## Why

The read was shipped with the inline byte retrieval deliberately deferred. A governed backplane read that can only hand back a transfer handle forces the caller out-of-band to fetch the bytes; the common `program.run → redirect stdout → read it back` compose in particular needs the content inline. This closes that gap without widening the agent surface.

## Handle behaviour (postulate 6)

Fetched content is returned as set-shaped `content_lines`:

- **utf-8-decodable** → decoded text lines (`content_encoding="utf-8"`), emitted **only when the whole file decodes cleanly** as UTF-8, split on newline with the trailing empty token dropped (same shaping as `k8s.logs`). These lines are **line-oriented, not byte-reversible** — a trailing newline is dropped and a CRLF file keeps a dangling `\r`.
- **non-decodable / binary** → the whole blob base64-encoded and wrapped into **wide base64 rows** (8192 chars/row, a whole base64 quantum — much wider than RFC 2045's 76 so a large binary produces far fewer rows to page), `content_encoding="base64"`; reassemble **byte-exactly** with `base64.b64decode("".join(content_lines))`. Base64 is the exact/reversible path.

Because `content_lines` is a list, the dispatcher's JSONFlux reducer spills it to a **result handle** above the size threshold (>50 rows **or** >4 KB) and the agent pages it with `result_query`. A registered `result_scalars` hint keeps the identifying/size scalars (`vm`, `file_manager_moid`, `guest_path`, `size_bytes`, `content_fetch`, `content_encoding`) on the reduced summary, so the **metadata envelope survives alongside the handle**. `attributes` (a dict) is **not** carried on that reduced summary (the reducer preserves only scalar types) — re-read with `fetch_content=false` to obtain a large file's `attributes`. No `ResultHandle` is constructed in the handler — the reducer wraps it, exactly as the sibling `process.list` / `env.read` reads do.

## Cap: refuse, don't truncate — with a bounded, streamed read

A file larger than `max_inline_bytes` is **refused** (a partial file is misleading), never truncated. The transfer GET is **streamed under a running byte budget** (`client.stream` + a running total over the response chunks), mirroring the shared transport's `_read_capped_json_response`, so at most the cap (plus one trailing chunk) is ever held in memory:

- **before any GET** when the guest-reported `FileTransferInformation.size` is known and over cap (saves server egress),
- a **`Content-Length` fast-reject** before a single body byte is read when the response advertises a length over the cap, and
- a **streamed running-total abort** that stops the read the instant the accumulated bytes exceed the cap.

This bounds memory even when a compromised/malicious transfer host — or a `verify_tls=false` MITM — omits or under-reports `FileTransferInformation.size` **and** omits or understates `Content-Length` (chunked transfer). A non-streaming read would materialise the whole body into the single-replica pod's heap first, so this is the memory-exhaustion (OOM) guard, not merely an after-the-fact size check.

## Safety tier: `caution` (not `safe`)

With `fetch_content=true` the op returns arbitrary guest file bytes read as the in-guest login from the target's `secret_ref` (commonly root / Administrator), inline, with **no allow-list and no redaction** — the least-constrained content-returning read in the tree. So the op is **`caution`** (not `safe`): it **auto-parks for agent / service principals**, while a **human seat still executes it immediately** (`requires_approval` stays `False`; caution parks only non-human principals). The tier is **op-level**, so even a metadata-only `fetch_content=false` read now auto-parks for agents — the conservative posture chosen in review over per-param tiering.

## Security

On the fetch path the one-time URL is consumed by the GET and carries the transfer token, so the result **omits `url` entirely** and the URL is never logged (flight-recorder spans already strip the URL query). The shared client's cross-origin-redirect WARNING now logs **scheme+host+path only** (query, fragment, and userinfo stripped), so a one-time transfer ticket in the URL query never lands in a log line even on an attacker-influenced cross-origin 3xx. Guest credentials continue to resolve from the target's `secret_ref` only.

## Behaviour left intact (out of scope)

`fetch_content=false` (the default) is **byte-identical** to the pre-fetch behaviour, including the raw one-time `url` in the result. Redaction of that deferred-path `url` is a separate, pre-existing recorder/redaction question. `classify_op` / `broadcast/events.py` (owned by #3718) and `CHANGELOG.md` are deliberately untouched.

## Tests

`backend/tests/test_connectors_vmware_rest_guest_ops.py` — the fake pooled client now streams the transfer GET (`stream` yields a chunked, byte-budgeted response and records how many bytes were drained). New/updated cases cover: deferred (default + explicit `fetch_content=false`) byte-identical envelope with no GET; small-file inline utf-8; large-file JSONFlux handle via the real reducer (metadata envelope preserved, `result_query` drill-in); refuse-before-GET on known over-cap size; **Content-Length fast-reject** (VIM size absent, no body byte drained); **streamed mid-body abort** when neither VIM size nor `Content-Length` bounds egress (asserts only ~one chunk past the cap is drained, far short of the full body); binary → base64 round-trip; URL/token never leaks; non-2xx GET propagates.

`backend/tests/test_connectors_http_adapter.py` — new case asserts the cross-origin-redirect WARNING logs scheme+host+path and never the query token. `backend/tests/test_connectors_vmware_rest_composites_register.py` — the safe-tier assertion now expects `caution` for `guest.file.read`. `docs/compatibility/vcf-api-contract-manifest.yaml` — moved `guest.file.read` from the `safe` to the `caution` composite bucket (pinned by `test_vcf_api_contract_manifest.py`).

Local, as CI runs them: `ruff check src/ tests/`, `ruff format --check src/ tests/`, `mypy src/` (clean on the touched files), and the guest-ops + http-adapter + composite-register + contract-manifest + reducer + coverage modules all pass. Verified locally against a throwaway rebase onto current `origin/main` (includes #3688's JSONFlux admission guard, which reads `Settings`); a per-file settings-env fixture was added to the guest-ops module so the reducer tests stay green when CI tests this branch merged with main.

## Docs

`docs/codebase/connectors-vmware-rest-guest-ops.md` (op-family table tier, safety-model split, and the "Inline file fetch" subsection — streamed cap, reversibility, `attributes`-on-handle, and the `caution` rationale), `docs-site/guides/windows-and-guest-execution.md` (op-table tier), and the registration `description` / `llm_instructions`.

Closes #3719


damir-topic (commented):
**Verdict:** REQUEST_CHANGES

Re-review of head `ae9f4070` against the previously-reviewed head `21a4a4c0` (one fast-forward commit; DCO sign-off present).

### Previous findings

- **B1 (unbounded transfer-GET body materialised before the cap):** RESOLVED as an OOM guard, but the refactor introduced a new blocking regression on the failure path (see Blocking B1-REGRESSION). The guard itself is correct: `_get_guest_file_bytes` now does `async with client.stream("GET", …)`, fast-rejects on `Content-Length` before reading a byte, and aborts the running total the instant `total > max_bytes` — before appending the offending chunk (`_guest.py:684-706`); `_file_too_large` names size+cap only, no URL (`_guest.py:640-651`); the refuse-before-egress guard on a known guest size stays (`_guest.py:481`).
- **M1 (`vmware.composite.vm.guest.file.read` tier):** RESOLVED. `safety_level="caution"`, `requires_approval=False` with an explanatory comment (`composites/_register.py`); descriptor + `llm_instructions` state caution / auto-parks-for-agents. Contract manifest moves **only** this op safe→caution — every other op keeps its bucket (`docs/compatibility/vcf-api-contract-manifest.yaml:456-457`). Per-principal semantics confirmed in `auth/permissions.py`: caution → NEEDS_APPROVAL for both `_SAFETY_DEFAULT` and `_SAFETY_CEILING` (non-grantable), so an agent/service principal auto-parks and a human seat executes; op-level, so even `fetch_content=false` parks for agents.
- **Redirect-log ticket scrub:** RESOLVED for the WARNING. New `_origin_and_path` drops query/fragment/userinfo (`adapters/http.py:547-556`); the cross-origin-redirect WARNING now logs `from_url=_origin_and_path(request.url)` — scheme+host+path only (`adapters/http.py:607-613`), asserted by the new structlog test. (Residual `location=` note below.)
- **Nit a (utf-8 `content_encoding` wording):** RESOLVED (`composites/schemas.py`) — no longer claims replacement chars; "emitted only when the whole file decodes cleanly as UTF-8".
- **Nit b (utf-8 lines line-oriented / not byte-reversible; base64 is the exact path):** RESOLVED (`schemas.py` + `_guest.py` `_shape_guest_file_content` docstring + docs).
- **Nit c (attributes dropped on the handle path):** RESOLVED (schema `attributes` description + `docs/codebase`).
- **Nit d (base64 row width 76→8192):** RESOLVED (`_guest.py:146`) — see Nits below for a downstream ergonomic note.

### Blocking

**B1-REGRESSION — non-2xx transfer GET raises uncaught `httpx.ResponseNotRead`: `dispatch()` raises and no error-audit row is written.** `_get_guest_file_bytes` calls `response.raise_for_status()` inside the `stream()` context *before* reading any body (`_guest.py:685`), so on any non-2xx (or an unfollowed cross-origin 3xx) the raised `HTTPStatusError` carries an **unread, then aclosed** streamed response. The composite does not wrap the call (`_guest.py:482`), so it surfaces to the dispatcher's `except httpx.HTTPStatusError` arm (`dispatcher.py:1141`) → `_handle_http_status_error` → `_classify_http_status_error` → `result_connector_error` / the 403 / 422 / auth builders, all of which call `_http_upstream_message(exc.response)` (`_errors.py:793`). That helper does `response.json()` / `response.text` but catches only `(ValueError, UnicodeDecodeError)` (`_errors.py:1151-1158`). An unread streamed response raises `httpx.ResponseNotRead` — a `StreamError`→`RuntimeError`, **not** `ValueError`/`UnicodeDecodeError` and **not** an `httpx.HTTPError` — so it escapes every dispatcher `except` arm. Because `_classify_http_status_error(...)` runs synchronously as an *argument* to `_audit_error_and_return(...)` (`dispatcher.py:1601`), the raise happens *before* the audit call: **no synchronous error-audit row is written** (violates the append-only-audit postulate). It then propagates through `_execute_and_audit` and `dispatch()`, whose outer `try` has only a `finally`, no `except` (`dispatcher.py:2485` / `:2884`): **`dispatch()` raises** instead of returning a structured `OperationResult` (violates the documented never-raises contract). This is a genuine regression — at `21a4a4c0` the code used `client.get()`, which buffered the body so `exc.response` was readable and `_http_upstream_message` worked. The trigger is common and is exactly the hostile/failure scenario this PR hardens: an expired/consumed one-time transfer ticket → 404/403, a transfer-host 5xx, or a refused cross-origin 3xx.

Confirmed empirically by both verifiers against the real `_http_upstream_message` / `result_connector_error` (raises `ResponseNotRead` for 404/500/429/403/422/401) and by static call-chain trace. **Secret-leak dimension:** the uncaught `ResponseNotRead` chains the original `HTTPStatusError` as `__context__`, and `HTTPStatusError.__str__` embeds the full transfer URL including `?token=<one-time-ticket>`; any top-level handler that logs the uncaught-exception traceback then leaks the ticket in cleartext — bypassing the redaction this PR just added. The new `test_file_read_fetch_get_failure_propagates` (`tests/test_connectors_vmware_rest_guest_ops.py:778`) does **not** catch this: its fake `raise_for_status` attaches a fresh, in-memory-readable `httpx.Response` and only asserts handler-level propagation, never the dispatcher's downstream body read on a real unread stream.

Fix any one of: read/`aread` the bounded error body before raising; raise a clean `RuntimeError` (no `httpx` response attached); or teach `_http_upstream_message` to guard `httpx.StreamError`/`ResponseNotRead`. Add a test that drives the dispatcher on a *real* streamed non-2xx transfer response.

### Major

**Decompression-bomb OOM — the streamed byte budget counts DECODED bytes, so a `Content-Encoding: gzip` transfer host defeats the cap the PR set out to enforce.** `_get_guest_file_bytes` iterates `response.aiter_bytes()` (decoded) for its running total and its `Content-Length` fast-reject compares the *compressed* wire length against the *decoded*-size cap (`_guest.py:700`). httpx auto-decodes on the response `Content-Encoding` regardless of the request `Accept-Encoding` and materialises one network read's decoded output in a single `aiter_bytes` iteration *before* the `total > max_bytes` check can fire. A ~200 KiB gzip payload (≈200 MiB of zeros, delivered as one chunk, no `Content-Length`) peaked at ~437 MiB of traced memory against a 1 MiB cap before aborting — attacker-scalable to a pod OOM. This contradicts the helper's own docstring ("at most `max_bytes` (plus one trailing chunk) is ever held in memory") and its claimed parity with `_read_capped_json_response`, which deliberately iterates `aiter_raw()` (compressed) to stay bomb-immune. A guest file transfer carries no legitimate transport `Content-Encoding` (the file bytes *are* the payload), so the fix is small: iterate `aiter_raw()` for the cap and/or reject a `Content-Encoding` on the transfer response. Pre-existing (not a regression — `21a4a4c0`'s `client.get().content` also fully decoded), but this change explicitly claims to bound memory against a lying/hostile transfer host and does not do so for a compressed one, so it belongs here.

### Nits

- **Base64 row width vs the result-query output budget (author awareness):** `_BASE64_LINE_WIDTH=8192` (`_guest.py:146`) against a default `result_query_max_output_bytes` of 256 KiB means a default `limit=50` read of a base64 handle raises the recoverable output-too-large error with the lower-your-limit paging remediation (~32 8 KiB rows fit per window). Graceful, not a crash; the default paging call on a binary handle just fails-and-retries once. A width nearer ~5000 would let the default `limit=50` fit; the whole-quantum 8192 (fewer total rows) is a reasonable trade — flagging only so it's a conscious choice.
- **Cross-origin-redirect WARNING still logs the unscrubbed `Location` header (pre-existing, out of this PR's changed lines):** the WARNING now scrubs `from_url` but still logs `location=response.headers.get("Location")` verbatim (`adapters/http.py:610-613`). That is the refused, attacker-chosen destination, which could itself carry a secret in its own query; the `from_url` scrub was the reported fix, so this is not a gate on #3720 — noting it for a possible follow-up.

### Verified by test vs inspection only

- **Verified by test** (author suite + throwaway repros at head, `uv run --frozen`, real `JsonFluxReducer`/`result_query`): streamed mid-body abort with no `Content-Length` (drained ≤ cap+one chunk, well below the full body); `Content-Length` fast-reject (0 bytes drained); refuse-before-GET on a known size; base64 round-trip and large-file real-reducer handle spill; `fetch_content=false` byte-identical (no GET); url/token absent from the success result; caution-tier registration + `requires_approval` False; the single manifest bucket move; the cross-origin-redirect WARNING stripping the query; and the decompression-bomb peak (437 MiB vs 1 MiB cap). Author suite green (guest_ops + http_adapter + register + manifest + reducer + recorder), ruff clean, mypy clean on the touched sources.
- **Inspection only** (not exercised by any test in the PR): the dispatcher-level non-2xx error path — the `ResponseNotRead` crash — is proven by a standalone repro against the real `result_connector_error`, but no test drives `dispatch()` on a real streamed non-2xx transfer response; and the per-principal caution enforcement is read from `auth/permissions.py`, not re-executed here.

### Roll note

Ships in the next backplane patch release together with the already-merged #3718 (whose recorder-side body exclusion for `guest.file.read` is on `main` and is brought in by the merge-queue rebase, not present at this isolated head). No migration. `vmware.composite.vm.guest.file.read` is now **caution**: agent / service principals auto-park (a grant cannot lift caution to auto-execute), a human seat executes immediately — operators must be told the tier change. Land the B1-REGRESSION fix (and, in the same change, the decompression-bomb bound) before rolling.

damir-topic (commented):
_Posted as a comment (not a formal App approval): the meho-review App installation token could not be minted this pass (1Password authorization did not complete across 5 attempts), so this cannot clear the standing App REQUEST_CHANGES. The PR therefore remains BLOCKED until a fresh meho-review App APPROVE is pinned to head `72340e30`. The review verdict and findings below are final and unchanged._

**Verdict:** APPROVE

Third pass on head `72340e30`, verifying resolution of the round-2 REQUEST_CHANGES against the round-2 head `ae9f4070`. Reviewed with two independent verifiers; every round-2 item is fixed and each is pinned by a new regression test that fails on `ae9f4070` and passes at head (mutation-confirmed).

**Round-2 findings**
- BLOCKING (streamed non-2xx GET escaped as `httpx.ResponseNotRead`, so `dispatch()` produced no structured result and no error-audit row): RESOLVED. `_get_guest_file_bytes` (`_guest.py:783-792`) drops `response.raise_for_status()`; on `not response.is_success` it reads a bounded ~4 KiB snippet via `aiter_raw` and raises a clean `_transfer_get_failed` RuntimeError **outside** any `except` (no chained `HTTPStatusError`, no URL/token in the message). Defence-in-depth: `_http_upstream_message` now catches `httpx.StreamError` on both `.json()` and `.text` (`_errors.py:1166`, `:1174`) — and `StreamError` is a `RuntimeError`, not an `HTTPError`, so it cannot mask a genuine status body.
- MAJOR (decompression-bomb amplification past the byte cap): RESOLVED. The transfer GET sends `Accept-Encoding: identity`, refuses any non-identity `Content-Encoding` before draining a byte (`_guest.py:796-798`), and counts the budget over `aiter_raw` wire bytes (`_guest.py:813`) — the same units as the Content-Length fast-reject, so raw == decoded.
- NIT (cross-origin-redirect WARNING leaked the Location query): RESOLVED. `_scrub_location_header` (`http.py:558`) reduces Location to scheme+host+path; the warning now logs the scrubbed value (`http.py:633`).
- NIT (base64 row width 8192 could bust the `result_query` default page budget): RESOLVED. `_BASE64_LINE_WIDTH` is now 4096 (`_guest.py:162`); a default 50-row page (~205 KB) stays under the 262144 output budget.

**Blocking:** none.

**Major:** none.

**Nits (non-blocking):**
- N1 — the GitHub PR body is stale vs the merged code: it still says base64 rows are "8192 chars/row", lists "non-2xx GET propagates" in Tests (head raises a clean connector_error), describes the cap over "the response chunks" with no mention of `aiter_raw` / `Accept-Encoding: identity` / the Content-Encoding refusal, and omits the four new test cases. Commit messages are accurate and CHANGELOG is correctly untouched — refresh the body so any squash-derived summary matches the code.
- V1 — two module docstrings still list `guest.file.read` as `safety_level="safe"` (`_guest.py:43-44`; `_register.py:29-32`, and the caution-rows narrative at `_register.py:54` omits it), contradicting the same PR's function docstring (`_guest.py:445`, "caution (not safe)"), the spec row (`_register.py:1536`, `safety_level="caution"`), and the contract manifest. Free-prose only, not pinned by conformance tests (CI stays green) — but it is this PR's own drift to clean up. Update both docstrings to list `file.read` as caution.
- N2 — `_bounded_error_snippet` (`_guest.py:715`) copies up to 4 KiB of the transfer host's non-2xx body into the error/audit envelope; a URI-echoing error page could in principle surface the one-time `?token=`. Low risk: the sink is `result_connector_error`, which runs `_sanitize_free_text` first (default Tier-1 policy redacts `token=<value>` → `[REDACTED:api_key]`, verified), the code never adds the URL itself, and real VMware transfer hosts return faults rather than URI-echoing pages. An echoing-body case is untested — defensive note only.

**Verified by test:** blocking clean-error + no chained response + no token/secret in the message; dispatcher structured-error with a committed error-audit row (the new `test_dispatch_streamed_non2xx_returns_structured_error_with_audit_row` drives real `dispatch()` with a genuinely-unread lazy stream — fails on `ae9f4070` with `httpx.ResponseNotRead` at the old helper, passes at head); StreamError hardening; `Accept-Encoding: identity` request header; gzip/br/deflate zero-drain refusal; wire-byte budget on a lying Content-Length; Location scrub; caution tier + contract manifest; base64 handle paging; `fetch_content=false`/omitted deferred envelope byte-identical to main. ruff check/format clean; the PR's touched test files pass at head; the mutation check confirms all five new tests fail on `ae9f4070` and pass at `72340e30`.

**Inspection only:** the `_transfer_get_failed` RuntimeError → `connector_error` mapping through the dispatcher's generic `except Exception` arm (the most basic dispatcher error path; the harder HTTPStatusError-with-unread-stream variant is exercised end to end by the new test); the delta `ae9f4070..72340e30` touches exactly the four fix items + their tests + docs (the second commit is a behaviorally-inert ruff-format reflow); `classify_op`/`events.py` and CHANGELOG untouched across the whole PR; DCO sign-off + `Co-Authored-By` on both new commits; no lab identifiers in the diff or the PR body.

**Roll note:** ships in the next backplane patch release together with the already-merged #3718; no migration. `guest.file.read` is now `caution` — agents and service principals park, humans execute — so operators must be told.

PR #2906, opened 2026-08-12
fix(ui): migrate memory create modal off dead daisyUI-v4 classes (#218)

smujkic94:
## Summary
- The **Create memory** modal (`_create_modal.html`) used daisyUI v4 `form-control` + `label-text`, both removed in v5 (zero compiled rules), so the label-over-input column collapsed — the Slug hint wrapped beside its label and, in the two-column Body/Preview and Expires/Tags grids, helper spans floated detached from their fields.
- Migrated all 7 fields (Scope, Target name, Slug, Body, Preview, Expires at, Tags) to `flex flex-col gap-1` label wrappers, `text-sm font-medium` labels, helper notes as `text-xs opacity-60` blocks below the field, and `w-full` controls. The Preview column's `<div class="form-control">` is migrated the same way.
- Presentation only — the create POST, CSRF double-submit, the debounced Markdown preview, the RBAC scope filtering (`writable_scopes`), and the target-name toggle script (keys off the untouched `data-target-name-row` / `data-action="scope-select"` hooks) are unchanged.

## Test plan
- [x] `ruff check` — clean
- [x] `pytest tests/test_ui_memory_create_promote.py tests/test_ui_memory_list.py tests/test_ui_memory_expiry_bulk.py` — 95 passed
- [x] **Aligned, no overlap** — all fields stack label-over-input via `flex flex-col gap-1`; hints render as blocks below
- [x] **No dead classes remain** — grep clean; the create-modal render-guard test asserts `form-control`/`label-text` absent from `/ui/memory/create`
- [x] **No functional regression** — create POST, markdown preview, RBAC scope filtering, and the target-name toggle covered by the 95 passing tests

## Out of scope
- The other memory templates carrying dead classes — sibling tasks evoila-bosnia/meho-internal#219 (`_promote_modal.html`), #220 (`index.html`), #221 (`_body_edit.html`), #222 (`_cards.html`) under initiative #217.
- No change to memory create behaviour, RBAC, or preview rendering.

## Issue
Implements **evoila-bosnia/meho-internal#218** (subtask of initiative **evoila-bosnia/meho-internal#217**). Closing keywords don't cross repositories, so the reference below is a traceability link — the issue is closed manually on merge.

Closes evoila-bosnia/meho-internal#218

PR #3723, opened 2026-09-17
fix(ui): structure the shared agent create/edit form fields (#348)

### LINKED ISSUE #348: MCP resource: meho://retrieve/{query} — wrap G0.4 retrieve helper as an MCP resource (G0.5-T9)
Parent goal: #221
Parent initiative: #226

## Background

G0.4-T5 (#262, merged via PR #345) deferred the MCP resource registration for `meho://retrieve/{query}` to keep T5's scope tight to the HTTP route. Both the in-process `retrieve` helper (G0.4-T4, #261 / PR #299) and the HTTP surface (`backend/src/meho_backplane/api/v1/retrieve.py`) are in place; this is the small follow-up that exposes the same retrieval through the MCP resource registry G0.5-T3 (#248 / PR #294) landed.

**Re-homed under G0.5 (#226), not G0.4 (#225).** Although it extends G0.4's retrieval surface, it depends entirely on G0.5 infrastructure — `register_mcp_resource` (#248 / PR #294) and the MCP auth contextvar (G0.5-T2, #247 / PR #290) — and G0.4 (#225) is closed/shipped. Tracked as **G0.5-T9** to avoid colliding with #226's reserved T7 (meta-tool stubs) / T8 (docs rescope).

## Scope

- Add `backend/src/meho_backplane/mcp/resources/retrieve.py` registering an MCP resource at the URI template `meho://retrieve/{query}` via `register_mcp_resource(...)` (`backend/src/meho_backplane/mcp/registry.py:330`).
- The handler reads the operator's `tenant_id` from the MCP auth context (no `tenant_id` in the URI template — tenant comes purely from the JWT), calls `retrieve(tenant_id=..., query=...)`, and returns the hits in the MCP resource shape — mirroring the sibling resources already in `mcp/resources/` (`kb.py`, `memory.py`, `docs.py`, `tenant_info.py`).
- RBAC: `operator` minimum, same as `POST /api/v1/retrieve`.
- Audit binding: reuse the `audit_query_hash` / `audit_source` / `audit_kind` / `audit_hit_count` contextvar pattern G0.4-T5 introduced (`backend/src/meho_backplane/api/v1/retrieve.py:266–281`) so the `audit_log` payload stays privacy-preserving — the raw query is never persisted.
- Tests in `backend/tests/test_mcp_resources_retrieve.py`.

## Acceptance criteria

- [ ] `meho://retrieve/{query}` registered via `register_mcp_resource(...)`; appears in `resources/list` against the running backplane.
- [ ] Resource handler returns the same `RetrievalHit` shape the HTTP route does.
- [ ] RBAC asserted by test: `read_only` MCP-auth → 403; `operator` / `tenant_admin` → hits.
- [ ] Tenant-scoping asserted by test: a query under tenant A returns no tenant-B documents (tenant from JWT, not URI).
- [ ] Audit payload mirrors the HTTP route's `{query_hash, source, kind, hit_count}` shape; test asserts the persisted `audit_log` row contains no raw query string.
- [ ] `backend/tests/test_mcp_resources_retrieve.py` passes.

## Out of scope

- The `retrieve` helper itself (G0.4-T4, shipped) — this only wraps it.
- The HTTP route `POST /api/v1/retrieve` (G0.4-T5, shipped) — unchanged.
- Reranking, streaming, multilingual retrieval — inherited out-of-scope from #225.
- MCP meta-tools — `meho://retrieve/{query}` is a *resource*, not one of the ~17 meta-tools tracked by #226's T7.

## References

- Parent Goal: #221. Parent Initiative: #226 (G0.5 MCP server bootstrap).
- Deferral source: G0.4-T5 (#262 / PR #345) — "MCP resource (optional in v0.2)".
- HTTP surface this mirrors: `backend/src/meho_backplane/api/v1/retrieve.py` (PR #345).
- `retrieve` helper: `backend/src/meho_backplane/retrieval/retriever.py` (G0.4-T4, #261 / PR #299).
- MCP resource registry: `register_mcp_resource` at `backend/src/meho_backplane/mcp/registry.py:330` (G0.5-T3, #248 / PR #294); reference resource impl to copy: `backend/src/meho_backplane/mcp/resources/tenant_info.py` (G0.5-T4, #249 / PR #297).
- MCP auth contextvar: G0.5-T2 (#247 / PR #290).
- Best-practices anchor — MCP spec, Resources (2025-06-18): https://modelcontextprotocol.io/specification/2025-06-18/server/resources

smujkic94:
## Summary

- Migrates the shared agent form partial (`agents/_agent_form_fields.html`) off the daisyUI v4 dead classes (`form-control` / `label-text` / `label-text-alt` — zero compiled rules in v5) to the established pattern: `flex flex-col gap-1` label wrappers, `text-sm font-medium` labels, `w-full` controls, and per-field validation errors as `text-xs text-error` (they were rendering unstyled, which is the worst place to lose emphasis).
- **One change fixes both dialogs.** `_create_modal.html:40` and `_edit_modal.html:40` each `{% include %}` this partial and carry no dead classes of their own, so "New agent" and "Edit agent" are covered together — that's why the two reported items are a single task rather than two PRs racing on one file.
- Preserved deliberately: the Enabled checkbox stays an inline `label` row (no column wrapper, no `w-full`), the Model tier / Turn budget `md:grid-cols-2` pairing, the `field_value` / `error_class` macros, the `data-error-for` and `data-form-error-summary` hooks, the `mode == "edit"` readonly Name plus its immutability hint, and every `aria-*` attribute.
- **Closes a test gap found while implementing:** the suite covered the edit dialog's RBAC gate and its PATCH, but never rendered it — so the `mode == "edit"` branch of the shared partial was unguarded. Added a positive edit-modal render test.

## Test plan

- [x] `pytest tests/test_ui_agents.py` — **27 passed** (+1 new test)
- [x] `pytest tests/test_ui_agent_run_console.py tests/test_ui_agent_runs.py tests/test_ui_agent_grants.py tests/test_ui_agent_principals.py` — **99 passed** (no sibling agents surface regressed)
- [x] `grep -c "form-control\|label-text" _agent_form_fields.html` → **0**
- [x] Render guards assert on the *rendered* body for both dialogs; the edit guard also pins `readonly` + `agent-name-immutable-hint`, so the dead-class assertions are non-vacuous
- [x] Verified no existing test pinned the old agents markup

## Out of scope

- Sibling tasks evoila-bosnia/meho-internal#349 (detail/card timestamps) and #350 (run console) — same initiative, different files.
- The **corpus** surface still uses these classes and has tests actively asserting their presence (`tests/test_ui_corpus_collections_table.py:618-619`). Untouched here; a candidate for a future polish round.

## References

Task: Closes evoila-bosnia/meho-internal#348 (parent initiative: evoila-bosnia/meho-internal#347, round 9 — Agents view). Cross-repo `Closes` does not auto-close — close #348 manually on merge.

Precedent migrations: #2906, #3108, #3309, #3527.


smujkic94:
@coderabbitai review

smujkic94:

`/auto-review-pr` — autonomous review at head `8bd88ed3`, run from an isolated worktree (`.claude/worktrees/pr-3723-review`) checked out at the PR head SHA. Posted as a comment because GitHub blocks self-approval.

**Anchor task:** evoila-bosnia/meho-internal#348 (parent initiative #347, round 9 — Agents view)
**Surface area:** frontend (Jinja/daisyUI templates) + backend test suite
**AI reviewer (CodeRabbit):** not received — silent across a bounded poll on both REST (`/reviews`, `/comments`, issue comments) and GraphQL (`reviews` + `reviewThreads`), consistent with its extended unavailability on this initiative. `bot_review.received: false`; verdict proceeds on the adversarial pass plus local and CI gates.

### Scope check

Two files, both in the task's declared scope — no strays:

- `backend/src/meho_backplane/ui/templates/agents/_agent_form_fields.html`
- `backend/tests/test_ui_agents.py`

Predicted-diff walk found no missing files: the create/edit modal wrappers were correctly left alone (they carry no dead classes and only `{% include %}` the partial), and no sibling agents surface needed to change for this task.

### Verification (isolated worktree at `8bd88ed3`)

| Gate | Result |
|---|---|
| `pytest tests/test_ui_agents.py tests/test_ui_agent_run_console.py` | **52 passed** |
| Import provenance | confirmed `meho_backplane` resolves to the **worktree** copy under the pytest `PYTHONPATH`, overriding the editable install that points at the main checkout |

CI on this SHA: **all three Python unit shards SUCCESS**, ruff + mypy SUCCESS, DCO SUCCESS, TruffleHog SUCCESS, zero failed checks. `Python (integration testcontainers)` and the image build were still running at review time; neither touches a Jinja template or a render test, but a merger should confirm they land green.

### Acceptance criteria (evoila-bosnia/meho-internal#348)

- [x] **No `form-control` / `label-text` / `label-text-alt` in either rendered dialog** — source grep returns 0 across `_agent_form_fields.html`, `_create_modal.html`, `_edit_modal.html`, and both render guards assert on the *rendered* body rather than the source.
- [x] **Labels stack above full-width controls; the two-column grid still pairs at `md`** — `w-full` on exactly the five column-wrapped controls, `grid ... md:grid-cols-2` intact.
- [x] **Inline field errors render at `text-xs` in the error colour; summary banner and `data-error-for` hooks still work** — the `error_for` macro keeps `role="alert"` and `data-error-for`; `data-form-error-summary` untouched.
- [x] **Edit mode still renders Name readonly with its immutability hint** — now actually pinned by a test (see below).
- [x] **Render guards + green gates** — see table.

### Adversarial pass

The interesting decision in this PR is what it *didn't* change, so I verified that against the pinned dependency rather than by eye:

1. **The premise holds for the version actually shipped.** daisyUI is vendored at **5.5.20** (`static/src/vendor/daisyui.js`, pinned by SHA256 in the Dockerfile). `form-control` and `label-text` appear **zero times** in that plugin, and zero times in the built `dist/tailwind.css` — so these really were dead classes here, not a version-recall assumption.
2. **Keeping the Enabled checkbox as an inline row is correct.** `.label` is *not* dead in v5 — the compiled CSS carries `.label { display: inline-flex; align-items: center; gap: …; &:has(input) { cursor: pointer } }`. So the row's flex context is live and `justify-start gap-3` still apply. Wrapping it in `flex flex-col gap-1` + `w-full`, the way the other five fields were handled, would have *broken* a working layout. Swapping only the dead `label-text` span was the right call.
3. **Dropping `mt-1` from the two hints is safe** — both sit inside the new `flex flex-col gap-1` wrappers, which now own that spacing. No hint was left outside a gap-bearing parent.
4. **The `error_for` change is the highest-value part of the diff.** Per-field validation messages were carrying `label-text-alt`, i.e. rendering completely unstyled — the worst place in a form to lose emphasis. They now render `text-xs text-error` while keeping `role="alert"` for assistive tech.
5. **No test regressions hidden elsewhere** — checked that no existing test pinned the old agents markup (the `label-text` assertions that do exist are on the corpus surface, untouched here), and the four sibling agents suites pass.

### Test-gap closure

Worth calling out: the suite covered the edit dialog's RBAC gate and its PATCH but **never rendered it**, so the `mode == "edit"` branch of the shared partial had no coverage at all. The new `test_edit_modal_renders_for_admin` pins `readonly` and `agent-name-immutable-hint` alongside the markup guard — which also makes the dead-class assertions non-vacuous rather than passing against an empty or errored response.

- The "Enabled" label inherits `.label`'s muted colour (`color-mix(currentcolor 60%)` in 5.5.20), so it renders slightly dimmer than the five `text-sm font-medium` field labels. This pre-dates the PR and is arguably right for a checkbox row; noting only so a future pass can decide deliberately.
- The **corpus** surface still uses these dead classes and has tests actively asserting their presence (`tests/test_ui_corpus_collections_table.py:618-619`). A future polish round will need to update those assertions in the same change.

### What worked well

- Recognising that the two reported items ("new agent dialog", "edit agent dialog") are one shared partial and shipping them as a single change, rather than two PRs racing on the same file where the second would be a no-op.
- Fixing the unstyled validation errors, which is the defect a user would actually feel, not just the label alignment that was reported.

### Verdict

**APPROVE** — correctly scoped, consistent with the rounds 5–8 pattern, verified against the pinned daisyUI build rather than assumed, and it closes a real coverage gap on the way. Merge once the integration job lands green; close evoila-bosnia/meho-internal#348 manually on merge (cross-repo `Closes` does not auto-close).

damir-topic (approved):
Reviewed commit 8bd88ed3e3f44dd98e5050f889c9ccd6869c36ab against evoila-bosnia/meho-internal#348.

The shared create/edit partial now stacks labels above full-width controls while retaining readonly Name, validation hooks, the responsive field pairing, and the inline Enabled checkbox. Browser fixtures used the actual modal wrappers and PR-built CSS: desktop layout is correct, and a true 375px viewport has no horizontal overflow.

Assisted by Codex.

PR #350, opened 2026-05-13
feat(connectors): SshConnector adapter — asyncssh + key/password auth + per-target connection caching (G0.2-T4)

kr3s0:
## Summary

- Adds `SshConnector` abstract base class in `connectors/adapters/ssh.py` — the shared SSH transport that all SSH-shaped vendor connectors (bind9, pfsense, Holodeck) will inherit from
- Uses asyncssh 2.23.0 for async-native SSH; no `asyncio.to_thread` offload required
- Auth is centralised: `_auth_config` reads `target.secret_ref` and selects key auth (PEM private key) over password auth automatically — vendor connectors never override this
- Per-target `SSHClientConnection` pool prevents redundant handshakes; `aclose()` drains all connections cleanly on lifespan teardown
- `_run_command` wraps `conn.run()` in `asyncio.wait_for` so timeouts raise `asyncio.TimeoutError`

## Test plan

- [x] `ruff check` — clean
- [x] `ruff format` — clean
- [x] `mypy src/` — clean (asyncssh override added to `pyproject.toml`)
- [x] `pytest tests/test_connectors_ssh_adapter.py` — 17 passed in 7.77s
- [x] `pytest tests/ --ignore=tests/integration` — 459 passed, 4 skipped (no regressions)
- [x] `SshConnector` importable from `meho_backplane.connectors.adapters`
- [x] Abstract class raises `TypeError` on direct instantiation
- [x] Password auth path: `secret_ref={username, password}` authenticates to in-process asyncssh server
- [x] Key auth path: `secret_ref={username, ssh_private_key}` authenticates via ed25519 key pair
- [x] Connection pool: same `target.name` returns the same `SSHClientConnection` object
- [x] Different targets get distinct connections
- [x] Closed connection triggers reconnect on next `_connect` call
- [x] Connection failure raises `OSError` at `_connect` time (mocked — deterministic, no network dependency)
- [x] Command timeout raises `asyncio.TimeoutError` (mock with `asyncio.sleep(60)`, `timeout=0.05`)
- [x] `aclose()` closes all pooled connections and empties the dict
- [x] `aclose()` is idempotent on fresh connector

## Out of scope

- Vendor-specific connectors (bind9, pfsense, Holodeck) — G3.3 / future G3 Initiatives
- Host key pinning / `known_hosts` management — v0.2.next
- SFTP / SCP file transfer, interactive shells, jump-host support

Closes #243

## Summary by CodeRabbit

* **New Features**
[...]

PR #3766, opened 2026-09-17
fix(ui): humanize agent detail + card timestamps, coerce them to UTC (#349)

### LINKED ISSUE #349: Harden chart helm-test workflow: remove continue-on-error: true after ~10 green runs (G0.4-T6 follow-up)

smujkic94:
## Summary

- The agent detail page and list cards rendered `created_at` / `updated_at` as raw `isoformat()` output (`2026-06-18T09:30:12.483920+00:00`). All three now follow the console `<time>` convention (precedent `topology/_drawer.html:178`, broadcast #3107, scheduler #3528): the ISO instant stays machine-readable in the `datetime` attribute, the visible text is `YYYY-MM-DD HH:MM UTC`.
- **The "UTC" suffix required a real fix, not just a format change.** Unlike the scheduler, the agents projection passed these timestamps through untouched (`views.py:150`, `:165-166`). They are `DateTime(timezone=True)` columns that the SQLite driver returns **naive**, so the suffix would have labelled an unzoned value and the `datetime` attribute would have carried an instant with no offset. Both `_card_context` and `_detail_context` now route them through `coerce_utc_aware`.
- **Reused the existing helper instead of adding a third copy** — `coerce_utc_aware` already lives one package over at `agents/runs/views.py:84` and is exported there. Verified no circular import: `runs/views.py` imports nothing from `agents/views.py`.
- `data-created-at` / `data-updated-at` hooks keep carrying the full ISO instant; nothing else in either template changed.

## Test plan

- [x] `ruff check src tests` — clean (an import-sort violation from the new import was caught locally and fixed before push)
- [x] `pytest tests/test_ui_agents.py tests/test_ui_agents_views.py tests/test_ui_agent_runs.py` — **52 passed**
- [x] **Coercion unit-asserted, not inferred from the rendered string** — new `tests/test_ui_agents_views.py` (3 tests) proves naive-in → tz-aware-out with the wall clock unshifted, aware-in passes through unchanged, and the card projection formats to exactly `2026-06-18 09:30 UTC`. A render-only assertion could not tell a coerced value apart from a naive one formatted with a hardcoded suffix.
- [x] Render guards in `test_detail_renders_full_definition`: `<time datetime=` and `UTC</time>` present, `+00:00<` absent (raw isoformat is no longer element text), and both `data-*` hooks still present
- [x] `grep isoformat` on both templates — every occurrence is inside a `datetime` or `data-*` attribute

## Out of scope

- Sibling tasks evoila-bosnia/meho-internal#348 (shared create/edit form layout, PR #3723) and #350 (run console) — same initiative, different files. No overlap: this PR touches neither the form partial nor the run console.
- This branch is based on `main`, which does not yet include #3723; the two PRs touch disjoint files.

## References

Task: Closes evoila-bosnia/meho-internal#349 (parent initiative: evoila-bosnia/meho-internal#347, round 9 — Agents view). Cross-repo `Closes` does not auto-close — close #349 manually on merge.


smujkic94:
@coderabbitai review

smujkic94:

`/auto-review-pr` — autonomous review at head `5cfb0b6b`, run from an isolated worktree (`.claude/worktrees/pr-3766-review`) checked out at the PR head SHA. Posted as a comment because GitHub blocks self-approval.

**Anchor task:** evoila-bosnia/meho-internal#349 (parent initiative #347, round 9 — Agents view)
**Surface area:** backend projection + Jinja templates + tests
**AI reviewer (CodeRabbit):** not received — silent across REST (`/reviews`, `/comments`) and GraphQL (`reviews` + `reviewThreads`) polls, consistent with its extended unavailability on this initiative. `bot_review.received: false`; verdict proceeds on the adversarial pass plus local and CI gates.

### Scope check

Five files, all in the task's declared scope — no strays:

`ui/routes/agents/views.py`, `templates/agents/_detail_body.html`, `templates/agents/_cards.html`, `tests/test_ui_agents.py`, `tests/test_ui_agents_views.py` (new).

**Predicted-diff walk found nothing missing within scope.** I enumerated every template under `templates/agents/` that references `created_at` / `updated_at` and every timestamp-bearing projection in `agents/views.py`. For the agent-*definition* surface the set is exactly `_detail_body.html` and `_cards.html`, and both are fixed. I also confirmed **every render path** routes through the coerced projections: `_card_context` (views.py:244) feeds both `agents/index.html` and the HTMX `_cards.html` fragment, and `_detail_context` (views.py:293) feeds both `agents/detail.html` and the HTMX `_detail_body.html` fragment — so there is no page/fragment split where one path renders an uncoerced value.

### Verification (isolated worktree at `5cfb0b6b`)

| Gate | Result |
|---|---|
| `ruff format --check` | clean (3 files) |
| `mypy src/.../agents/views.py` | **Success: no issues** |
| `pytest test_ui_agents, test_ui_agents_views, test_ui_agent_runs, test_ui_agent_grants` | **76 passed** |
| Import provenance | confirmed `meho_backplane` resolves to the **worktree** copy under the pytest `PYTHONPATH`, overriding the editable install |

CI on this SHA: **all three Python unit shards SUCCESS**, ruff + mypy SUCCESS, DCO SUCCESS, TruffleHog SUCCESS, **zero failed checks**. `Python (integration testcontainers)` and the image build were still running at review time; neither exercises a Jinja template or this projection, but a merger should confirm they land green.

### Acceptance criteria (evoila-bosnia/meho-internal#349)

- [x] **Timestamps reach the template tz-aware on the SQLite path, unit-asserted** — see the non-vacuousness check below.
- [x] **Both detail cells and the card timestamp render `YYYY-MM-DD HH:MM UTC` inside `<time datetime="…">`** — verified in the rendered body by the guards, and by grep that every remaining `isoformat()` in both templates sits inside a `datetime` or `data-*` attribute.
- [x] **`data-created-at` / `data-updated-at` hooks unchanged** — both still carry the full ISO instant; asserted in the render guard.
- [x] **No raw isoformat as visible element text** — `assert "+00:00<" not in body`.
- [x] **Guards + projection test green** — see table.

### Adversarial pass

The claim this PR turns on is "the UTC suffix is now truthful", so I attacked that specifically rather than reading the diff for style:

1. **The coercion test is non-vacuous — verified empirically, not assumed.** The obvious failure mode is Pydantic normalizing naive datetimes at model construction, which would make `assert created.tzinfo is UTC` pass regardless of what the projection does. I constructed an `AgentDefinitionRead` from a naive datetime against the installed **Pydantic 2.13.4** and confirmed `created_at.tzinfo is None` on the way in. The assertion therefore genuinely proves the projection's coercion.
2. **The test also pins that coercion does not shift the wall clock** (`created.replace(tzinfo=None) == _NAIVE`), which is the other way this could silently go wrong — attaching UTC to a value that was already local would move the displayed time.
3. **No import cycle.** `agents/views.py` now imports `coerce_utc_aware` from `agents/runs/views.py`; `runs/views.py` imports nothing from `agents/views.py`, and mypy + the full suite import cleanly.
4. **Reuse over duplication is the right call here** — the helper was already exported from the sibling package, and a fourth copy of a six-line function is how these drift apart.
5. **The `<time>` element is the correct vehicle** — the machine-readable instant lives in `datetime` while the human string is the text node, so the existing `data-*` hooks and the new attribute carry identical ISO values and nothing downstream loses precision.

Recorded for planning; neither blocks this merge, and both are outside #349's declared scope (which names `_detail_body.html` + `_cards.html`).

- **Three sibling sub-surfaces under `agents/` have the same defect and are not covered by round 9.** They render raw `isoformat()` as visible text *and* their projections skip coercion: `agents/principals/_table.html:51-52` (projection `principals_views.py:113`), and `agents/grants/_detail_body.html:95,99` + `agents/grants/_rows.html:55-56` (projection `grants/views.py:139,153`). Worth a follow-up task under #347 if "the Agents view" is meant to be consistent end to end. Note `agents/runs/_table_rows.html:90` is already correct — it uses the relative-time + ISO-tooltip pattern and `runs/views.py` coerces.
- **`coerce_utc_aware` now exists in four places** across `ui/` (connectors list view, scheduler views, agents/runs views, plus `session_store._coerce_utc`). This PR correctly imported rather than adding a fifth, but the long-term home is probably a shared `ui/` time utility.

### Docs

`docs/codebase/` has no agents-console artifact that this change invalidates — the behaviour change is presentational plus a projection normalisation. `docs_status: verified_unchanged`.

### What worked well

- Treating this as a correctness fix rather than a formatting one. Changing only the template would have produced a "UTC" label on a value whose zone was never established — the defect would have been invisible and wrong rather than visible and wrong.
- Adding a projection-level test in its own module (matching the existing `test_ui_<surface>_views.py` convention) instead of relying on the rendered string, which cannot distinguish a coerced value from a naive one formatted with a hardcoded suffix.

### Verdict

**APPROVE** — correctly scoped, the load-bearing claim is verified against the installed Pydantic rather than assumed, every render path is covered, and all gates are green. Merge once the integration job lands; close evoila-bosnia/meho-internal#349 manually on merge (cross-repo `Closes` does not auto-close).

damir-topic (approved):
Reviewed commit 5cfb0b6b58ff3ac89e49f44dd7dd26e32db9b999 against evoila-bosnia/meho-internal#349.

The detail and card projections establish UTC-aware timestamps before rendering human-readable time elements, preserving ISO data attributes. The supported asyncpg/SQLite storage paths preserve the UTC invariant. The previous unrelated BIND9 integration failure passed on retry at this exact commit. The manual source quality hook passed.

Assisted by Codex.

PR #3769, opened 2026-09-17
fix(ui): structure the agent run-console form (#350)

smujkic94:
## Summary

- Migrates the agent run console (`agents/_run_console_body.html`) off the daisyUI v4 dead classes (`form-control` / `label-text` / `label-text-alt` — zero compiled rules in v5) to the established pattern: `flex flex-col gap-1` label wrappers, `text-sm font-medium` labels, the helper line as `text-xs opacity-60`, `w-full` on the prompt textarea and work_ref input, and "(optional)" as an inline `opacity-60 font-normal` qualifier.
- `max-w-md` stays on the Work ref wrapper so it keeps reading narrower than the prompt — the one layout constraint worth preserving here.
- Untouched: the form's `hx-post` / `hx-target="#agent-run-transcript"` / `hx-swap`, `data-form="agent-run"`, `data-action="run"`, all three `{% if not enabled %}disabled` guards, the disabled-agent warning alert, the turn-budget badge with its `aria-live="polite"`, and the transcript mount's idle state.
- **Closes a coverage gap found while implementing:** the suite exercised the disabled-agent *POST* (409) but never *rendered* the disabled console, so those three `disabled` guards were unpinned — and two of them sit on the fields restructured here.

## Test plan

- [x] `pytest tests/test_ui_agent_run_console.py tests/test_ui_agents.py` — **52 passed** (+1 new test)
- [x] `grep -c "form-control\|label-text"` on the template → **0**
- [x] Render guards assert the migrated classes are present (`…text-sm w-full`, `…input-sm font-mono w-full`, `flex flex-col gap-1 max-w-md`) and the dead ones absent
- [x] **The new disabled-console test is mutation-verified** — dropping a single `{% if not enabled %}disabled` guard from the template makes it fail; restoring it returns the suite to green

## A note on how the disabled test asserts

The obvious form of that assertion — counting `"disabled"` occurrences in the page — is worthless here, because the warning copy reads *"This agent is disabled and cannot be run"*, so the count passes with zero controls actually locked. A first-`<input>` lookup is no better: the app shell renders its own inputs ahead of the form, and that is exactly how the first draft of this test failed. The committed version anchors on `name="input"`, `name="work_ref"`, and `data-action="run"`, then checks the `disabled` attribute inside each control's own tag.

## Out of scope

- Sibling tasks evoila-bosnia/meho-internal#348 (shared create/edit form, PR #3723) and #349 (detail/card timestamps, PR #3766) — same initiative, disjoint files; all three can merge in any order.
- The run **transcript** (`_run_transcript.html`) and the runs **list** filters (`agents/runs/list.html:72-84`, which carry the same dead-class pair) are separate surfaces, flagged in the task body as follow-up candidates.

## References

Task: Closes evoila-bosnia/meho-internal#350 (parent initiative: evoila-bosnia/meho-internal#347, round 9 — Agents view). Cross-repo `Closes` does not auto-close — close #350 manually on merge.


smujkic94:
@coderabbitai review

smujkic94:

`/auto-review-pr` — autonomous review at head `710519cf`, run from an isolated worktree (`.claude/worktrees/pr-3769-review`) checked out at the PR head SHA. Invoked without a PR number; defaulted to this PR as the most recently opened one awaiting review. Posted as a comment because GitHub blocks self-approval.

**Anchor task:** evoila-bosnia/meho-internal#350 (parent initiative #347, round 9 — Agents view)
**Surface area:** frontend (Jinja/daisyUI template) + backend test suite
**AI reviewer (CodeRabbit):** not received — silent across GraphQL (`reviews` + `reviewThreads`) and issue-comment polls, consistent with its extended unavailability across this initiative. `bot_review.received: false`; verdict proceeds on the adversarial pass plus local and CI gates.

### Scope check

Two files, both in scope — no strays:

`templates/agents/_run_console_body.html`, `tests/test_ui_agent_run_console.py`.

**Predicted-diff walk found nothing missing.** I checked every template on the run-console render path — `run_console.html` (full page), `_run_console_body.html`, `_run_transcript.html`, `_run_error.html` — and all four now grep clean for `form-control` / `label-text`. The three the PR did *not* touch were already clean, so there is no half-migrated surface left behind.

### Verification (isolated worktree at `710519cf`)

| Gate | Result |
|---|---|
| `pytest tests/test_ui_agent_run_console.py` | **26 passed** |

CI on this SHA: **all three Python unit shards SUCCESS**, ruff + mypy SUCCESS, CodeQL SUCCESS, TruffleHog SUCCESS, DCO SUCCESS, **zero failed checks**. `Python (integration testcontainers)` and the image build were still running at review time; neither exercises a Jinja template, but a merger should confirm they land green.

### Acceptance criteria (evoila-bosnia/meho-internal#350)

- [x] **No `form-control` / `label-text` / `label-text-alt` in the rendered run console** — grep clean on the template, and both render guards assert on the *rendered* body.
- [x] **Prompt textarea and work_ref input fill their wrappers; work_ref stays capped at `max-w-md`; labels sit above** — asserted by exact class strings, and `max-w-md` is present exactly once.
- [x] **Disabled-agent behavior unchanged** — all three `{% if not enabled %}disabled` guards intact, warning alert renders, and this is now *tested* (below).
- [x] **Run POST still swaps the transcript on success and on the inline error paths** — the pre-existing `test_run_post_returns_transcript_with_stream_token` and `test_run_post_disabled_agent_renders_409_inline` still pass, and `hx-post` / `hx-target="#agent-run-transcript"` / `data-form="agent-run"` / `data-action="run"` / the transcript mount are all byte-identical.
- [x] **Render guard added; gates green** — see table.

### Adversarial pass

The migration itself is mechanical and correct, so I spent the effort on the new test, which is where this PR could have shipped false confidence:

1. **The assertion shape is right, and both obvious weaker forms are genuinely broken here.** Counting `"disabled"` occurrences would pass on prose alone — the warning copy reads *"This agent is disabled and cannot be run"*. A first-`<textarea>`/`<input>`/`<button>` lookup would inspect the app shell's own controls, not the form. The committed version anchors on `name="input"`, `name="work_ref"`, and `data-action="run"`, then slices `rindex("<") … index(">")` to isolate each control's own opening tag. I confirmed `data-action="run"` appears exactly once on this page, so that anchor is unambiguous.
2. **I re-ran the mutation check independently rather than trusting the PR's claim — and deliberately mutated a *different* guard.** The PR body reports dropping the work_ref guard; I dropped the **prompt-textarea** guard instead. The test still failed, then passed again after restore, with `git status` clean. So the guard catches regressions on more than the single control the author happened to try.
3. **No attribute-value `>` hazard in the tag-slicing.** The placeholders and aria-labels on these three controls contain no `>`, so `index(">", at)` terminates at the real tag boundary. Worth noting because that technique would silently mis-slice on a control whose attribute contained a bracket.
4. **`max-w-md` on the wrapper, `w-full` on the input is the correct pairing** — the input fills the wrapper while the wrapper stays narrower than the prompt, which is the visual relationship the original `form-control max-w-md` was expressing.

- `agents/runs/list.html:72-84` still carries the `form-control` / `label-text` pair on its filter labels. Explicitly declared out of scope in #350's body; worth a follow-up task if the Agents area is meant to be consistent end to end. This compounds with the two surfaces flagged on PR #3766 (`agents/principals/`, `agents/grants/`) — after round 9 the agents area will still have four unmigrated spots.
- The **corpus** surface has tests actively asserting `label-text` is present (`tests/test_ui_corpus_collections_table.py:618-619`), so whoever migrates it must update those assertions in the same change.

### Docs

No `docs/codebase/` artifact covers the run console's markup; the change is presentational with no contract shift. `docs_status: verified_unchanged`.

### What worked well

- Closing the disabled-console coverage gap rather than only migrating the classes — two of the three `disabled` guards sit on the restructured fields, so without that test the riskiest part of this diff would have been unverified.
- Writing the assertion to survive the page it actually runs against. The first two shapes of that test would both have passed vacuously, and the PR body documents why they were rejected, which is the kind of thing that saves the next person from "simplifying" it back.

### Verdict

**APPROVE** — correctly scoped, every hook preserved, the whole run-console render path is now consistent, and the one new behavioural guard is independently mutation-verified. Merge once the integration job lands; close evoila-bosnia/meho-internal#350 manually on merge (cross-repo `Closes` does not auto-close). This completes round 9 (#347).

damir-topic (approved):
Reviewed commit 710519cfdeb46f6c9e9513eee8c1fc21b073e34d against evoila-bosnia/meho-internal#350.

The run form preserves its HTMX transcript swap, narrower work-ref field, and all three disabled controls. Browser fixtures with PR-built CSS verified enabled/disabled states and no horizontal overflow at a true 375px viewport; desktop rendering is correct.

Assisted by Codex.

PR #3770, opened 2026-09-17
fix(ui): structure the principals register dialog (#359)

smujkic94:
## Summary

- Migrates the "Register principal" dialog (`agents/principals/_register_modal.html`) off the daisyUI v4 dead classes (`form-control` / `label-text` / `label-text-alt` — zero compiled rules in v5) to the established pattern: `flex flex-col gap-1` label wrappers, `text-sm font-medium` labels, `w-full` on both inputs, and "(optional)" as an inline `opacity-60 font-normal` qualifier on the Owner label.
- **The higher-value half is the inline field errors.** Both carried `label-text-alt text-error`, so a duplicate-name 409 or a bad-name 422 re-rendered the modal with the validation message in the *unstyled* default colour — the worst place in a form to lose emphasis. They now render `text-xs text-error` while keeping `role="alert"` and their `data-error-for` hooks.
- Dropped the redundant `mt-1` from the Owner hint, which the new wrapper's `gap-1` now owns.
- Untouched: the register `hx-post`, the `hx-headers` CSRF echo, `hx-disabled-elt`, the `data-register-banner` alert, `data-action="cancel-register"` / `data-action="register-submit"`, and `modal-box max-w-lg`.

## Test plan

- [x] `pytest tests/test_ui_agent_principals.py tests/test_ui_agent_grants.py` — **51 passed**
- [x] `grep -c "form-control\|label-text"` on the template → **0**
- [x] Clean-render guard in `test_register_modal_renders_for_admin`: dead classes absent, and exactly **2** controls carry `…font-mono w-full`
- [x] **Error-render guard** in `test_register_duplicate_name_renders_409_inline`: asserts the exact migrated span `<span class="text-xs text-error" role="alert" data-error-for="name">`. This is deliberate — the error span only exists when `errors` is populated, so the clean render cannot cover it and the unstyled-error fix would otherwise ship unverified.
- [x] Confirmed the pre-existing 409/422 tests assert on the `data-error-for` hook rather than the class, so the migration did not silently invalidate them

## Out of scope

- Sibling task evoila-bosnia/meho-internal#360 (grants New grant + Elevate dialogs, one shared partial) — same initiative, different files.
- `principals/_revoke_modal.html` carries 2 dead-class occurrences, and `principals/_table.html:51-52` renders `principal.created_at.isoformat()` as visible text with no UTC coercion in `principals_views.py:113`. Both recorded in the task body as follow-up candidates rather than widening this change.

## References

Task: Closes evoila-bosnia/meho-internal#359 (parent initiative: evoila-bosnia/meho-internal#347, round 9 — Agents view). Cross-repo `Closes` does not auto-close — close #359 manually on merge.


smujkic94:
@coderabbitai review

smujkic94:

`/auto-review-pr` — autonomous review at head `fd0d5a3d`, run from an isolated worktree (`.claude/worktrees/pr-3770-review`) checked out at the PR head SHA. Posted as a comment because GitHub blocks self-approval.

**Anchor task:** evoila-bosnia/meho-internal#359 (parent initiative #347, round 9 — Agents view)
**Surface area:** frontend (Jinja/daisyUI template) + backend test suite
**AI reviewer (CodeRabbit):** not received — silent across GraphQL (`reviews` + `reviewThreads`) polls, consistent with its extended unavailability across this initiative. `bot_review.received: false`; verdict proceeds on the adversarial pass plus local and CI gates.

### Scope check

Two files, both in scope — no strays: `templates/agents/principals/_register_modal.html`, `tests/test_ui_agent_principals.py`.

Predicted-diff walk found nothing missing *within the task's declared scope*. The task names this one dialog; the sibling `principals/_revoke_modal.html` and the `principals/_table.html` timestamps were explicitly declared out of scope in the task body and remain so — that is a deliberate boundary, not an omission.

### Verification (isolated worktree at `fd0d5a3d`)

| Gate | Result |
|---|---|
| `pytest tests/test_ui_agent_principals.py` | **27 passed** |

CI on this SHA: **all three Python unit shards SUCCESS**, ruff + mypy SUCCESS, CodeQL SUCCESS, TruffleHog SUCCESS, DCO SUCCESS, **zero failed checks**. `Python (integration testcontainers)` and the image build were still running at review time; neither exercises a Jinja template, but a merger should confirm they land green.

### Acceptance criteria (evoila-bosnia/meho-internal#359)

- [x] **No `form-control` / `label-text` / `label-text-alt` in the rendered register modal** — grep clean, and both the clean-render and error-render tests assert on the *rendered* body.
- [x] **Both inputs fill the modal width; labels sit above** — `count(…font-mono w-full) == 2`, correctly scoped (see below).
- [x] **Inline field errors render `text-xs text-error`, and the 409/422 re-render still populates them** — pinned by an exact-string assertion on the 409 path, and mutation-verified.
- [x] **Register POST, CSRF gate, and the Keycloak/Vault error banners behave as before** — `data-register-banner`, both `data-error-for` hooks, `data-action="cancel-register"` / `register-submit`, `hx-post`, the `hx-headers` CSRF echo, `hx-disabled-elt` and `modal-box max-w-lg` are all untouched (absent from the diff), and the pre-existing 409/422/503/502 tests pass.
- [x] **Guards added; gates green** — see table.

### Adversarial pass

1. **The `count(…) == 2` assertion is correctly scoped, verified rather than assumed.** It would be fragile against a full-page render where the app shell contributes its own inputs. I confirmed `_register_modal.html` has no `{% extends %}` and that `principals_forms.py` renders it directly on both the GET (`:164`) and the error re-render (`:196`) — a bare fragment. So the count is exactly the two modal inputs, and `assert "label-text" not in body` cannot be diluted by shell markup either.
2. **The error-render guard is mutation-verified.** Reverting the span to `class="label-text-alt text-error"` in the worktree makes `test_register_duplicate_name_renders_409_inline` fail; restoring it returns the suite to green with a clean `git status`. The guard genuinely pins the unstyled-error fix rather than passing on the surrounding assertions.
3. **Putting that guard on the error path was necessary, not belt-and-braces.** The span is inside `{% if errors and 'name' in errors %}`, so it does not exist in a clean render at all — a clean-render-only guard would have left the highest-value half of this change completely unverified.
4. **The pre-existing error tests were checked for collateral damage.** They assert on `data-error-for="name"`, not on the class, so the class migration did not silently invalidate them — and they still pass.
5. **Dropping `mt-1` from the Owner hint is safe** — the hint is now a flex child of a `gap-1` column, and `aria-describedby="principal-owner-hint"` still resolves.

- **The `owner_sub` inline error span is migrated identically but has no test — before or after this PR.** The suite injects failures only for the `name` field (409 duplicate, 422 bad-name), so `data-error-for="owner_sub"` is never rendered under test. This is a pre-existing coverage gap rather than a regression, and the change is symmetric with the `name` span that *is* mutation-verified, so it does not block. Worth closing whenever someone adds an owner_sub failure path.
- `principals/_revoke_modal.html` (2 dead-class occurrences) and `principals/_table.html:51-52` (raw `isoformat()` as visible text, uncoerced in `principals_views.py:113`) remain. Together with the grants surfaces flagged on #3766/#3769, the agents area still has several unmigrated spots after round 9.

### Docs

No `docs/codebase/` artifact covers this dialog's markup; presentational change with no contract shift. `docs_status: verified_unchanged`.

### What worked well

- Identifying that the unstyled inline errors — not the label alignment — were the defect an operator actually hits, and treating that as the substance of the change.
- Recognising that the error span cannot be covered by a clean render, and guarding it on the 409 path instead. That is the difference between a test that documents the fix and one that merely accompanies it.

### Verdict

**APPROVE** — correctly scoped, every functional hook preserved, assertions verified as correctly scoped rather than assumed, and the load-bearing guard is mutation-verified. Merge once the integration job lands; close evoila-bosnia/meho-internal#359 manually on merge (cross-repo `Closes` does not auto-close).

damir-topic (approved):
Reviewed commit fd0d5a3db4e209e87f8bec1f11cac7ed5fdd4151 against evoila-bosnia/meho-internal#359.

The register dialog has stacked labels, full-width inputs, and visible field errors while retaining POST, CSRF, HTMX and accessibility hooks. Browser fixtures used the exact template and PR-built CSS: both inputs fit a true 375px viewport, with no horizontal overflow and 12px text in the error colour.

Assisted by Codex.

Issue #3710, opened 2026-09-16
Task: targets probe on standalone ESXi reports reachable=true from the cached about.version without a live SOAP call


damir-topic:
Parent initiative: #3660
Part of #221

## Summary

For a **standalone-ESXi** target the `vmware_rest` connector's
`fingerprint()` / probe path reports `reachable=true` from a **cached**
`about.version` when the session cache is warm — it performs **no live**
`Login` / `RetrieveServiceContent` SOAP call. So a "reachable" fingerprint
proves the connector once held a session, not that the session is still valid
or the host is healthy. A probe that answers `reachable=true` without touching
the host is a false-positive health signal.

## Root cause (citations at v0.35.2)

- `connectors/vmware_rest/connector.py:1805` `fingerprint()`: on a standalone
  ESXi host `GET /api/about` is answered HTTP 400 by the vAPI/JSON-RPC handler
  **regardless of session validity**; the session used for that GET is served
  from the warm cache
  (`connector.py:977-979`: `_session_token` returns the cached token verbatim,
  no live call), then the method branches to `_fingerprint_esxi`.
- `connectors/vmware_rest/connector.py:1904` `_fingerprint_esxi`: reads the
  version straight from the cache —
  `version = self._about_versions.get(cache_key)` (`:1925`) — and stamps
  `reachable=True` **by construction** (`:1930`). No `Login`,
  `RetrieveServiceContent`, or `RetrievePropertiesEx` is issued, so nothing in
  the probe confirms the cached session is live or the host is answering SOAP.

## Proposal

- The standalone-ESXi probe performs a **live lightweight SOAP call**
  (`Login` + `Logout`, or an unauthenticated `RetrieveServiceContent`) rather
  than reading `about.version` from cache.
- The fingerprint payload reports the method explicitly:
  `probe_method: live | cached`, so a consumer can tell a live-verified probe
  from a cache-served one and never mistake a warm cache for a health signal.

## Acceptance criteria

- [ ] A probe against a **warm** session cache still performs **exactly one**
      live SOAP call to the host (fixture/asserted), rather than returning
      from cache.
- [ ] The fingerprint payload carries `probe_method` distinguishing `live`
      from `cached`.
- [ ] `docs/codebase/connectors-vmware-rest.md` documents that the ESXi probe
      is live and what `probe_method: live | cached` means.

## Related

- #3708 — sibling: `object.collect` never surfaces `missingSet` fault types,
[...]

Issue #3708, opened 2026-09-16
Task: object.collect never surfaces missingSet fault types, and the flight recorder omits text/xml bodies — SOAP property-read faults are unobservable


damir-topic:
Parent initiative: #3660
Part of #221

## Correction (2026-09-16)

The original stale-session mechanism this Task filed under is **refuted**: an expired/idle standalone-ESXi SOAP session does **not** answer `RetrievePropertiesEx` with HTTP 200 + per-property `missingSet` `NotAuthenticated` faults — it raises a **top-level** `ServerFaultCode` `NotAuthenticated` SOAP fault (Broadcom KB 336368; govmomi), which the connector already maps to `ConnectorAuthError(401)` via `_SOAP_AUTH_FAULT_TYPES` (`connector.py:290`) + `_soap_fault_error` (`connector.py:1641-1671`, auth check `:1660`) → the dispatcher's `invalidate_session` + cold re-login.
The observed void was **host-side** (it persisted through a real host reboot, which necessarily mints a fresh session), not a client-side stale-cookie bug.
What remains true and useful is an **observability** gap — this Task is rescoped to it; the auth-recovery mechanism stays as-is.

## Summary

When a standalone-ESXi `RetrievePropertiesEx` returns HTTP 200 with objects
whose requested properties land in `missingSet` (each `missingSet[]` entry
carrying a per-property `fault` — `NoPermission`, `NotAuthenticated`,
`InvalidProperty`, ...), the connector **discards the fault type**: the
`object.collect` parser keeps only `missingSet[].path`, and the composite vim
readers ignore `missingSet` entirely. An operator therefore cannot tell
"property genuinely absent / not set" from a per-property permission or
authentication fault. The flight recorder makes the after-the-fact diagnosis
impossible too: it **omits** `text/xml` SOAP bodies wholesale (records the
`[MEHO-OMITTED:redaction-uncertain]` marker, not even a hash), so the fault
types in a captured RetrievePropertiesEx response are unobservable from the
trace.

## Observed

Four standalone-ESXi targets read as fully **void** — every requested
property in `missingSet`, HTTP 200, `status ok` — and **neither the result nor
the trace showed WHY**. The result envelope named the missing paths but not
the per-property faults; the flight-recorder body for the RetrievePropertiesEx
response was omitted (`text/xml` → `[MEHO-OMITTED:redaction-uncertain]`), so
the fault types were not recoverable after the fact. Whatever the host-side
[...]

PR #3776, opened 2026-09-17
fix(vmware): recover expired SOAP sessions — auth health-reset on all-properties-missing + bounded session max-age (Closes #3773)

### LINKED ISSUE #3773: Task: vmware SOAP connector never recovers an expired per-target session — inventory reads return all-properties-missing with no re-login (session TTL + all-missing health-reset)
## Summary

The vmware-rest connector caches one vSphere SOAP session per `(tenant_id, target_id)` for the backplane process lifetime and only ever invalidates it on a **top-level** SOAP auth fault. When an ESXi session expires **server-side** while still cached, the host keeps answering the unauthenticated `RetrieveServiceContent` on that dead cookie, but `RetrievePropertiesEx` returns HTTP 200 with **every requested property in `missingSet`, each carrying an embedded per-property `NotAuthenticated` fault**. That per-property fault is not a top-level `<soapenv:Fault>`, so the connector's existing re-login path never fires and the "void" (every inventory property missing, `ServiceInstance.content` fine) persists for the entire pod lifetime. A **host reboot does not clear it** (the backplane re-presents the same dead cookie); only a **pod restart** does (drops the cached session → cold re-login).

This is a connector robustness gap that is **certain from the code** (see below). It is the **strongly-suspected — but not definitively proven — cause** of a multi-day nested-ESXi inventory void on a live lab estate (four standalone-ESXi targets registered in the same minute by an automated factory run, all voided together after a ~4h idle gap, all cleared with zero host changes after the pod was replaced across a release roll). The decisive discriminator read (the per-property fault type, or an independent-session read during the void) was never captured because the surface dropped the fault — which is exactly the gap #3708/#3772 close.

## Mechanism (code-grounded)

- Session keying is correct and per-target: `target_cache_key(target) = (str(tenant_id), str(id))` (`connectors/_shared/cache_key.py:42-53`); the auth is the `vmware_soap_session` cookie in the pooled `httpx.AsyncClient` jar, pooled under that tuple plus `(verify_tls, ca_pin_digest, base_url[host:port])` (`connectors/vmware_rest/connector.py:515,613-619,926-980`; `connectors/adapters/http.py:671,712-729`). No sharing across targets — "all four at once" is common-cause (same establish minute + same ~30-min ESXi SOAP idle timeout), not shared state.
- **No session TTL / max-age / periodic re-login.** A cached session lives until a top-level auth fault (`_soap_fault_error` → `ConnectorAuthError(401)` → dispatcher invalidate+retry-once, `connector.py:290,1660-1668`) or process restart.
- **A stale cookie is not detected on the read path.** An all-properties-missing `RetrievePropertiesExResponse` has no top-level `<soapenv:Fault>`, so `parse_soap_fault` returns `None` (`connectors/vmware_rest/soap.py:537-552`), `_post_soap` returns success, and `_extract_object_content` reads only `miss["path"]` and **discards the embedded fault** (`connectors/vmware_rest/typed_ops_object_collect.py:129-131`). No retry, no invalidate. This is the object.collect half of #3708 (being surfaced by #3772).

## Why the existing recovery misses it

The dispatcher re-logins on a **top-level** `NotAuthenticated`. An expired session on a *property read* surfaces `NotAuthenticated` **per property inside `missingSet`**, not at the envelope. Different shape, never caught. (A prior investigation refuted "stale session" — that refutation was correct only for the top-level shape.)

## Fix

- **(a) Bounded session liveness.** Give the cached ESXi SOAP session a TTL / max-age and proactively re-login on a warm-cache hit past it (default safely below ESXi's SOAP idle timeout — verify the exact bound before pinning).
- **(b) All-properties-missing health-reset (load-bearing).** Building on the surfaced `missingSet` fault type (#3772): when an object resolves but **every requested property is missing with an auth-class fault** (`NotAuthenticated`/`NoPermission`-at-object-scope), raise `ConnectorAuthError` so the existing invalidate + retry-once path fires. Guard to **one reset per read** so a legitimate per-field `NoPermission` does not loop.
- **(c) Test.** An injected transport returns all-missing-with-`NotAuthenticated` on the first call and real props on retry; assert exactly one re-establish and full recovery, plus a regression guard on today's silent-void behaviour.
- **(d) Close the probe-cache masking (#3710).** `probe`/`fingerprint` short-circuits to a cached `about.version` and reports `reachable=true` without a live authenticated read; route a lightweight authenticated read through (b) so `reachable` reflects liveness.

## Relationship / impact

- #3708 / #3772 surface the `missingSet` fault type (the observability half). This ticket is the **recovery half** and depends on that surfacing for (b).
- Operational impact: a factory run that reuses a pooled session left idle > the ESXi timeout voids on its next host read (the estate's relaunch reused a ~4h-idle session). An assisted-resume run that pauses > the idle timeout for a human decision and then re-reads a host would reintroduce the void mid-run without this fix.

Refs #3708. Parent: #3660.

damir-topic:
## What

Tests + docs for the expired-SOAP-session recovery landed in the parent commit
on this branch (`ed0a4338`). The source is unchanged by this PR; this stacks the
missing test coverage and the `docs/codebase/` update on top of it.

## Why

A vSphere/ESXi SOAP session that has expired server-side but is still cached
process-side keeps answering the unauthenticated `RetrieveServiceContent` on the
dead cookie, but `RetrievePropertiesEx` returns **HTTP 200 with every requested
property in `missingSet`, each carrying a per-property `NotAuthenticated`
fault** — there is no top-level SOAP `<Fault>`. A per-property `NotAuthenticated`
inside `missingSet` is not a top-level SOAP fault, so the connector's existing
re-login never fired, and `object.collect` returned a silent all-missing "void"
that no path recovered for the rest of the pod lifetime.

## Mechanism (recap of the source under test)

- **(b) All-missing auth health-reset (load-bearing).** When an object resolves
  but *every* requested property is missing under an auth-class fault
  (`NotAuthenticated`, or an all-object `NoPermission` — never a partial per-field
  `NoPermission`), `_collect_with_auth_health_reset` invalidates the cached
  session (the same `invalidate_session` hook the dispatcher's #2067
  mid-session-401 path uses) and re-reads **once** on the fresh session. At most
  one reset per dispatch (structural single re-read); a still-all-missing retry
  returns the fault-annotated envelope normally. Reads both the 9.x nested
  `LocalizedMethodFault` and the 8.0.x flattened fault shapes.
- **(a) Bounded session max-age (defense in depth).** A warm session older than
  `VMWARE_SOAP_SESSION_MAX_AGE_SECONDS` is proactively re-established before its
  next read.
- **(d) #3710 probe liveness.** The ESXi probe/fingerprint confirms reachability
  with one lightweight authenticated `SessionManager.currentSession` read routed
  through the (b)-protected path, so a stale/void session no longer probes green.

## (B) design decision — env var, not a Settings field

`VMWARE_SOAP_SESSION_MAX_AGE_SECONDS` is resolved **once at import** into the
module constant `_SESSION_MAX_AGE_SECONDS`, not modelled as a chassis `Settings`
field. A `Settings`-model field would couple `get_settings()` to every vmware
read on the hot path (and broke 30 unit tests when tried). This mirrors the
`adapters.http._resolve_max_response_bytes` precedent.

## Guardrails (verified)

- Default **1200 s (20 min)**, with a code comment citing the ESXi SOAP idle
  timeout (`Config.HostAgent.vmacore.soap.sessionTimeout` = 1800 s / 30 min); the
  max-age must stay well below it.
- Invalid / non-numeric / negative / unset env → default; `0` disables the
  proactive re-login. The read path never raises on a bad value.
- Resolved once (module constant), not per call.
- `grep -rn vmware_soap_session_max_age src tests` → **0 matches** (no removed
  Settings field referenced anywhere).

## Tests

`test_connectors_vmware_rest_typed_object_collect.py` (object.collect, level b):
all-missing `NotAuthenticated` → exactly one re-establish + full recovery; the
one-reset-per-read guard (a second all-missing does not loop); partial per-field
`NoPermission` does not reset; regression guard on the silent-void shape;
all-object `NoPermission` resets; non-auth (`InvalidProperty`) / unset-only do
not reset; and the vSphere 8.0.x **flattened** missingSet fault shape (reset +
surfacing paths) so `fault_type_name` is exercised on both the 9.x nested and
8.0.x flattened shapes.

`test_connectors_vmware_rest_esxi_session.py` (session layer, levels a + d):
end-to-end SOAP re-login recovery through an injected `/sdk` transport
(void read → Logout → re-establish → recovered read); two max-age tests via
`monkeypatch.setattr` on `_SESSION_MAX_AGE_SECONDS` (past max-age proactively
re-logins; within max-age does not) plus a `0`-disables case; env resolver tests
(unset → 1200, valid override honoured, `0` honoured, non-numeric → 1200,
negative → 1200, never raises); and probe/fingerprint liveness (a dead session
reports `reachable=False`; a stale-but-healable session is recovered by the reset
and probes reachable). The two pre-existing ESXi fingerprint tests were updated
to answer the new `currentSession` liveness read.

Suite: full vmware suite green; dispatcher / surface / reference suites green.
ruff + ruff format + mypy(src) clean; code-quality hook clean.

## Follow-ups (v0.35.5)

- `fault_type_name` wrapper-name fallback in `vim_body.py`: when a per-property
  fault's nested concrete fault cannot be resolved (e.g. an empty inner fault
  element that loses its `xsi:type` in the SOAP codec), `fault_type_name` returns
  the useless wrapper name `LocalizedMethodFault` rather than `None`. Real faults
  carry child content so this does not bite in practice, but the fallback is
  worth tightening.
- Refresh the stale `vmware.object.collect` op-description prose.

Closes #3773
Refs #3708 #3710
Stacked on #3772.


damir-topic (commented):
**Verdict:** APPROVE

Reviewed at head `59c05f9e`. The change is correct, well-tested, DCO-signed on every commit, does not touch the changelog, and is leak-free. The three load-bearing behaviours — (b) self-healing session recovery, (a) bounded session max-age, and (d) probe-liveness — all work as intended and are proven by real-transport and clock-controlled tests. The full `vmware_rest` suite (1053 passed / 44 skipped / 0 failed) is green, and `ruff`, `ruff format --check`, and `mypy (src)` are clean. Zero blocking findings.

One base-mechanics note up front: the reviewed diff is the 5-file change against the PR's current base branch tip. `git diff origin/main...HEAD` does **not** equal the PR diff — its merge-base predates the base branch, so it additionally shows two already-merged files. The true PR diff (base-branch-tip `...HEAD`) is exactly the three commits and matches the stated file stats; that is what was reviewed, and the overlapping files are byte-identical to `main` (already merged via a separate squash). See Major #2 for the merge-path consequence.

## Blocking
None.

## Major

1. **Vault credential-read failure during the liveness re-login escapes `_esxi_session_is_live` and propagates out of `fingerprint()`** — `backend/src/meho_backplane/connectors/vmware_rest/connector.py:2095`. The `except` tuple catches `(ConnectorAuthError, httpx.HTTPError, OSError, RuntimeError)`, but a credential-read error is `CredentialsReadError(Exception)` — not a subclass of any of those. The liveness read routes through the health-reset path, which on the stale symptom invalidates the session and cold re-establishes; that re-establish calls the session loader, which can raise the credential-read error transiently (revoked/expired token, backend blip). It then escapes `_esxi_session_is_live` and, because `_fingerprint_esxi` is invoked from inside `fingerprint()`'s `except` blocks with no bare-`Exception` handler, propagates out of `fingerprint()` — directly contradicting the method's own docstring contract at `connector.py:2068-2070` ("a failure here must degrade to unreachable, never propagate"). Trigger is exactly this PR's scenario (a stale ESXi session forcing a re-login) coincident with a credential-read hiccup, so a probe sweep / post-roll re-baseline over a stale target during such a hiccup gets a stack trace instead of `reachable=False`. Non-blocking: narrow double-condition trigger, and the codebase already tolerates this exception type escaping `fingerprint()` on the initial establish, so the invariant was already imperfect for this type. One-line fix: add `CredentialsReadError` (already imported at `connector.py:107`) to the `except` tuple. Recommend for the v0.35.5 follow-up.

2. **Merge path: the branch is based on an unmerged base branch whose content is already in `main` via a separate squash, so a rebase/retarget onto `main` conflicts** — `backend/src/meho_backplane/connectors/vmware_rest/typed_ops_object_collect.py` and `backend/tests/test_connectors_vmware_rest_typed_object_collect.py`. `git merge-tree --write-tree origin/main HEAD` reports content conflicts in both of those files (both sides added the base-branch work relative to the common ancestor, and this PR layered its change on top); `connector.py` and the docs auto-merge cleanly. Not a code defect — the reviewed 5-file diff is sound — but the base must be retargeted to `main` and the two now-redundant base commits dropped **before** any merge-queue attempt; a naive squash against current `main` will conflict. (The PR shows MERGEABLE only against its current base, not against `main`.)

## Nits

1. **`_esxi_session_is_live` docstring and docs claim a missing SessionManager moid degrades to unreachable, but the code returns reachable** — `backend/src/meho_backplane/connectors/vmware_rest/connector.py:2065` (docstring says "`False` when the moid is unknown") and `docs/codebase/connectors-vmware-rest.md:835` ("A missing SessionManager moid or a read failure degrades to unreachable"), while `connector.py:2082-2085` returns `True` when the moid is `None` ("leave reachability to the establish that already ran"). Two doc statements contradict the code, in the exact anti-masking direction a reader of the masking-close cares about. Functional impact is negligible — the branch is effectively dead, since the ESXi flavor gate that reaches `_fingerprint_esxi` always implies the moid is cached (both are set and cleared together). Fix: correct the docstring + docs to "trusts the prior establish → reachable", or, matching the stated intent more strictly, return `False` on a missing moid.

2. **Recovery is an inline invalidate + one re-read, not a raised auth error, and a persistent void returns op-SUCCESS-with-annotation rather than a structured error** — `backend/src/meho_backplane/connectors/vmware_rest/typed_ops_object_collect.py:217`. `_collect_with_auth_health_reset` reads once and, on all-missing-auth, calls `invalidate_session` inline and re-reads once; it never raises and does not depend on the dispatcher's mid-session retry arm. This is a valid, self-contained design (the re-read cold re-establishes a fresh session before re-issuing the same body) and the PR body describes it accurately. Consequence to register: when the retry is still all-missing, the caller returns a normal op-SUCCESS envelope (empty properties + a missing-fault summary), not a structured auth error — observable via the fault annotation and the probe, but an agent that ignores the annotation could misread empty properties as "no data" rather than "auth failed". No change required; surfaced so the mental model matches what shipped.

3. **`NoPermission` in the auth-class reset set costs +1 login per read on a genuine full-permission-loss target** — `backend/src/meho_backplane/connectors/vmware_rest/typed_ops_object_collect.py:79`. A legitimately all-denied read (every requested property `NoPermission`) satisfies the reset predicate and fires one invalidate + re-login that cannot recover a real role loss (same principal), so the retry returns the same fault and the annotated envelope is returned. It never loops (structural one-reset guard), but a permission-lost target pays a steady +1 login per read for the duration of the loss. `NotAuthenticated` earns its inclusion (re-login does recover the stale-session case); `NoPermission` mostly adds cost without recovery benefit. Bounded and documented — noted so the tradeoff is a conscious sign-off.

4. **The health-reset evicts the shared per-target session, so `NoPermission` reads churn the warm session for all of that target's consumers** — `backend/src/meho_backplane/connectors/vmware_rest/typed_ops_object_collect.py:79`. `invalidate_session` pops the shared cached session, so the `NoPermission` reset in nit #3 drops and cold-re-establishes the whole target's session, affecting any concurrent/subsequent read on it. A monitoring sensor pinned to a permission-denied property would reset that target's session every tick. Bounded (one reset per dispatch, immediate re-establish, no loop) and consistent with the documented rationale, but weigh this shared-session-churn cost alongside nit #3 when signing off on `NoPermission`'s inclusion.

5. **Bounded max-age proactive re-login now governs vCenter sessions too, not only standalone ESXi** — `backend/src/meho_backplane/connectors/vmware_rest/connector.py:1042` (the max-age gate sits in `_session_token`, the shared session path for every `vmware_rest` read; comment at `:1040` "Applies to vCenter and ESXi alike"). This is an intentional, documented change, not a regression: it is safe (the vCenter invalidate is network-free — the session-teardown call stays in `aclose`), adds only ~1 extra re-login per active vCenter target per max-age window, and all 1053 `vmware_rest` tests pass. Healthy ESXi `object.collect` reads are unchanged (the reset predicate is false on a populated read → identical prior behaviour), and vCenter fingerprinting is untouched (`_fingerprint_esxi` is ESXi-flavor-gated). Flagged only so the wider-than-ESXi blast radius is an explicit, conscious sign-off.

## Verified by test

- **(b) recovery / self-healing re-login** — proven at unit level (a scripted resetting connector: one invalidate, two reads, populated result) and end-to-end at the transport layer, asserting the exact SOAP sequence (service-content, login, retrieve, logout, service-content, login, retrieve — one logout, two logins, same body re-issued on the fresh session). Load-bearing: neutering the reset predicate fails 5 tests, and replacing the single re-read with a loop hangs the void-twice fixture — so both the reset and the one-reset bound are load-bearing, not incidental.
- **One-reset guard / no loop** — the void-on-both-reads test asserts exactly one invalidate, two reads, and an annotated envelope with no further reset.
- **Reset predicate branches** — all-object `NoPermission` resets; a legitimate partial per-field `NoPermission` (some property still reads) does **not** reset (invalidate count 0); an all-missing non-auth fault does not; an all-missing no-fault does not. Both the nested and the flattened fault shapes are exercised.
- **(a) max-age** — clock-controlled proactive re-establish test, within-age reuse, `0` disables, default below the ESXi SOAP idle timeout, and the full env-resolver matrix (unset / valid / zero / non-numeric / negative all resolve without raising on the read path). A 50-concurrent over-age read check produced exactly two logins (no stampede).
- **(d) probe liveness** — dead-session → unreachable (version still surfaced, teardown attempted, error text carries no credential/host detail) and stale → healed → reachable fingerprint tests.
- **Suite / lint / types** — `pytest -k vmware_rest` = 1053 passed, 44 skipped, 0 failed; `ruff check`, `ruff format --check`, and `mypy (src)` clean on the changed files. Recovery-path envelope and fingerprint error string are empirically leak-free (only the fault type name surfaces, never the fault message, cookie, or credentials).

## Inspection only

- Diff-base equivalence: the three-dot `origin/main...HEAD` shortcut does not hold (merge-base artifact); the true base-tip diff is the 5-file change matching the stated stats, and the overlapping files are byte-identical to `main`.
- DCO `Signed-off-by` present on all three PR commits (and both base commits); changelog untouched; the file list is exactly the five expected.
- Nit #1 (docstring/docs vs code on a missing moid; dead branch), Major #1 (the credential-read-error class is `Exception`-rooted and escapes the `except` tuple), and Major #2 (`git merge-tree` conflict on a rebase onto `main`) are inspection-confirmed; Major #1 is additionally reproduced by a throwaway transport test.
- Nit #5 gate location and the shared-session-path reasoning are confirmed by reading `_session_token` and both ESXi-gated call sites.

## Roll note

- Targets **v0.35.5**, after the v0.35.4 roll.
- **No migration** (no alembic revision in the diff).
- New env var **`VMWARE_SOAP_SESSION_MAX_AGE_SECONDS`**, default **1200 s** (0 disables the proactive re-login; unset/non-integer/negative fall back to the default).
- Merge is additionally gated on the orchestrator's post-roll re-baseline and the author session's Phase-A idle proof.
- **Retarget the PR base to `main`** (dropping the two now-redundant base commits) before queueing — see Major #2; a naive squash against current `main` will conflict.

damir-topic:
## Rebase onto main + review fold-ins

This PR was stacked on `fix/3708-missingset-fault-types` (#3772). #3772 landed on
`main` as squash `7cb1ec1c`, so its branch commits `0a847fdd` / `2d10a26f` are not
on `main`. The branch has been rebased with `--onto origin/main` to drop those two
so a retarget to `main` does not resurface them, and the base is now `main` (which
is what runs the CI matrix — the PR had only had DCO until now).

### Rebase (drop the squashed base commits)
- `git rebase --onto origin/main 2d10a26f fix/3773-session-liveness` — replayed only
  the three real commits; `0a847fdd` / `2d10a26f` are gone.
- No conflicts; every file applied cleanly onto `origin/main`.
- The **rebased-only** state (before the fold-in below) is patch-id-equal to the
  pre-rebase branch vs its old base: `git diff origin/main HEAD | git patch-id
  --stable` = `ba9d464cde847d69cf8b68e43b7382941f9a37f0` — identical content, just a
  new base. Trailers (DCO + both Co-Authored-By) preserved on all three commits.

Rebased commits: `f24b8d5e` (fix) · `0be3757b` (test) · `a5bdf902` (docs).

### Review fold-in (`49945ca`)
One commit on top carries the two fixes from review 5237111080 (APPROVE), touching
only files this PR already changed:

- **F5 (major):** `_esxi_session_is_live`'s liveness re-login can raise
  `VaultCredentialsReadError` (Exception-rooted via `CredentialsReadError`) on a
  coincident credential-read hiccup; it was outside the except tuple and escaped
  both the liveness read and `fingerprint`. Added `CredentialsReadError` to the
  tuple so it degrades to unreachable like the transport/auth errors already there.
  - `connector.py:107-110` (import `CredentialsReadError` alongside
    `VaultCredentialsReadError`)
  - `connector.py:2101` (except tuple gains `CredentialsReadError`)
  - test: `test_fingerprint_credentials_read_error_during_liveness_degrades_unreachable`
    at `test_connectors_vmware_rest_esxi_session.py:1197-1239`
- **F1 (nit, code-follows-docs):** a missing cached `SessionManager` moid now
  returns not-live (fail-closed) instead of reachable, matching the method
  docstring and `connectors-vmware-rest.md` (#3710: never report reachable without
  a live authenticated read). Docstring/docs wording unchanged.
  - `connector.py:2085-2091` (`sm_moid is None` → `return False`, comment reworded)
  - test: `test_esxi_liveness_unreachable_when_session_manager_moid_absent`
    at `test_connectors_vmware_rest_esxi_session.py:1240-1251`

`git show --stat 49945ca`: 2 files changed, 68 insertions(+), 4 deletions(-).

Local gates before push (all green): ruff check + ruff format --check on the two
changed files; mypy clean on `connector.py`; the two vmware test files (37 + 23) and
the full vmware suite (`-k vmware`, 1131 passed / 44 skipped); the repo
code-quality hook exit 0.

New head `49945cad1d12022cc2381eab148f5ecee57e3805`. This is the first hosted CI
matrix run for this PR (it only ran DCO while targeting the stacked base).

damir-topic:
### Fix: order-dependent F5 test

**Failure** (merge_group run 35251201833, "Python unit shard 3/3"; the PR-ref matrix on the same head had passed):

```
FAILED tests/test_connectors_vmware_rest_esxi_session.py::test_fingerprint_credentials_read_error_during_liveness_degrades_unreachable
  AttributeError: 'module' object at meho_backplane.connectors.vmware_rest.typed_ops_object_collect
  has no attribute 'typed_ops_object_collect'
```

**Root cause** — test-only, not a production defect. The F5 test used a **dotted-string** `monkeypatch.setattr("...typed_ops_object_collect._collect_with_auth_health_reset", ...)`. pytest resolves that string by walking the package: `getattr(vmware_rest, "typed_ops_object_collect")`. That attribute is bound only because the test module's own top-level import pulled the submodule in; once it is in `sys.modules`, pytest's re-import is a no-op that never re-binds it on the parent package. So if an earlier test **on the same xdist worker** unbinds that attribute, the resolve raises the exact `AttributeError` above. The PR-ref matrix passed only because no earlier test on its worker had disturbed the binding — i.e. worker/order-dependent, not deterministic.

**Fix** — patch the already-imported **module object** directly (`monkeypatch.setattr(object_collect_module, "_collect_with_auth_health_reset", ...)`). That is the same object the production call-time import inside `_esxi_session_is_live` reads the attribute from, so the intercept stays correct and is now order-independent (no dotted-string resolve walk). Matches this file's existing module-object patch pattern. No production code touched.

Verified: the test alone, the whole file, the file under `xdist -n 2` ×3, and a synthetic "unbind first" pollution repro all green; wider `-k vmware` suite 1130 passed / 44 skipped (one pre-existing acceptance e2e error is an unrelated network/DNS fixture dependency). The PR-ref CI matrix re-runs on the new head `cae50a57`.

PR #3781, opened 2026-09-18
fix(ui): structure the shared grant form fields (#360)

smujkic94:
## Summary

- Migrates the shared grant form partial (`agents/grants/_grant_form_fields.html`) off the daisyUI v4 dead classes (`form-control` / `label-text` / `label-text-alt` — zero compiled rules in v5) to the established pattern: `flex flex-col gap-1` label wrappers, `text-sm font-medium` labels, `w-full` on all five controls, and per-field validation errors as `text-xs text-error` (they were rendering completely unstyled).
- **One change fixes both dialogs.** `_create_modal.html:46` and `_elevate_modal.html:49` each `{% include %}` this partial and carry no dead classes of their own, so "New grant" and "Elevate" are covered together — that's why the two reported items are a single task rather than two PRs racing on one file.
- Preserved: the Target scope / Verdict `md:grid-cols-2` pairing, the `field_value` / `error_class` macros, the `__root__` error banner, every `data-error-for` hook, all five `aria-describedby` links, and **all three `mode == "elevate"` branches** on the Expires-at field.

## Test gap closed

Neither dialog had a positive render test — the suite covered the create modal's RBAC gate and the POST paths (`test_create_success_redirects`, `test_elevate_requires_expires_at`, `test_elevate_past_expiry_renders_inline_422`, `test_elevate_success_redirects`) but never rendered either modal. So the shared partial's markup, and the elevate branch that is the *only* difference between the two dialogs, were entirely unguarded. This PR adds render tests for both modes.

Two details worth knowing about how they assert:

- **The mode-specific check is scoped to the `expires_at` tag.** `required` also sits on `principal_sub` and `op_pattern`, so a page-wide `"required" in body` cannot distinguish create from elevate — it would pass in both. A small helper slices the one tag.
- **The error-span styling is pinned on the 422 path**, not the clean render. That span lives inside `{% if errors %}`, so a clean-render guard structurally cannot reach it and the unstyled-error fix would have shipped unverified.

## Test plan

- [x] `pytest tests/test_ui_agent_grants.py tests/test_ui_agent_principals.py` — **53 passed** (grants suite 24 → 26)
- [x] `grep -c "form-control\|label-text"` on the partial → **0**; `w-full` present on exactly 5 controls
- [x] Both render tests assert `body.count("w-full") == 5`, which also proves neither modal wrapper contributes stray matches
- [x] **Mutation-verified:** deleting the `{% if mode == "elevate" %}required{% endif %}` branch makes `test_elevate_modal_renders_for_admin` fail; restoring it returns the suite to green

## Out of scope

- Sibling tasks evoila-bosnia/meho-internal#348/#349/#350/#359 — same initiative, disjoint files; all five round-9 PRs can merge in any order.
- `grants/index.html` (3 dead-class occurrences on its list filters) and the raw `isoformat()` timestamps in `grants/_detail_body.html:95,99` + `grants/_rows.html:55-56`, uncoerced in `grants/views.py:139,153`. Recorded in the task body as follow-up candidates.

## References

Task: Closes evoila-bosnia/meho-internal#360 (parent initiative: evoila-bosnia/meho-internal#347, round 9 — Agents view). Cross-repo `Closes` does not auto-close — close #360 manually on merge. This completes round 9.


smujkic94:
@coderabbitai review

smujkic94:

`/auto-review-pr` — autonomous review at head `78bde38b`, run from an isolated worktree (`.claude/worktrees/pr-3781-review`) checked out at the PR head SHA. Invoked without a PR number; defaulted to this PR as the most recently opened one awaiting review. Posted as a comment because GitHub blocks self-approval.

**Anchor task:** evoila-bosnia/meho-internal#360 (parent initiative #347, round 9 — Agents view)
**Surface area:** frontend (Jinja/daisyUI partial) + backend test suite
**AI reviewer (CodeRabbit):** not received — silent across GraphQL (`reviews` + `reviewThreads`) polls, consistent with its unavailability throughout this initiative. `bot_review.received: false`; verdict proceeds on the adversarial pass plus local and CI gates.

### Scope check

Two files, both in scope — no strays: `templates/agents/grants/_grant_form_fields.html`, `tests/test_ui_agent_grants.py`.

Predicted-diff walk: every template under `templates/agents/grants/` now greps clean for `form-control` / `label-text` **except `index.html`** (3 occurrences on its list filters), which the task body explicitly declares out of scope. That is a stated boundary, not an omission.

### Verification (isolated worktree at `78bde38b`)

| Gate | Result |
|---|---|
| `pytest tests/test_ui_agent_grants.py` | **26 passed** (24 before) |

CI on this SHA: **every check SUCCESS with zero failures** — all three Python unit shards, ruff + mypy, CodeQL, TruffleHog, DCO, **and `Python (integration testcontainers)`**, which had completed by review time on this PR (unlike the earlier round-9 PRs where I had to flag it as still running).

### Acceptance criteria (evoila-bosnia/meho-internal#360)

- [x] **No `form-control` / `label-text` / `label-text-alt` in either rendered dialog** — grep clean on the partial; both new render tests assert on the *rendered* body of each modal independently.
- [x] **All five controls fill their wrappers; the Target scope / Verdict grid still pairs at `md`** — `count("w-full") == 5` plus an explicit `grid gap-3 md:grid-cols-2` assertion.
- [x] **Inline field errors render `text-xs text-error`; the `__root__` banner and `data-error-for` hooks still populate on a 422** — pinned by an exact-string assertion on the 422 re-render path.
- [x] **`mode == "elevate"` still marks Expires at `required` with its marker; `mode == "create"` leaves it optional** — asserted in both directions and mutation-verified both ways (below).
- [x] **Positive render tests added for both modals; gates green** — see table.

### Adversarial pass

The migration is mechanical; the risk in this PR is tests that look thorough but prove nothing. Three things I verified rather than accepted:

1. **`count("w-full") == 5` is exactly the five controls.** It would be meaningless if a modal wrapper contributed its own `w-full` or if the fragment extended the app shell. Neither `_create_modal.html` nor `_elevate_modal.html` contains `w-full` or an `{% extends %}`, so the count is unambiguous — and it doubles as proof that no control was missed.
2. **The `_expires_at_tag` slicing is safe on this markup.** The helper takes `rindex("<") … index(">")` around `name="expires_at"`; that mis-slices if any attribute value contains a `>`. The input's `value`, `aria-label`, `placeholder` (`2026-07-01T13:30:00+00:00`) and `aria-describedby` contain none. Worth stating because the technique is not universally safe.
3. **The mode branch is mutation-verified in *both* directions.** The PR reports deleting the `{% if mode == "elevate" %}required{% endif %}` branch and watching the elevate test fail. I ran the complementary mutation — making `expires_at` **unconditionally** `required` — and confirmed `test_create_modal_renders_for_admin` fails. So the two tests bracket the branch from both sides rather than one of them passing vacuously. Worktree restored clean after each.
4. **Scoping the mode assertion to the tag was necessary, not fastidious.** `required` also sits on `principal_sub` and `op_pattern`, so a page-wide `"required" in body` is true in *both* modes and would have proven nothing — the exact trap the helper exists to avoid.
5. **The error-styling guard had to live on the 422 path.** The span is inside `{% if errors %}`, so no clean render reaches it; a clean-render-only guard would have left the unstyled-error fix unverified. Same structural reasoning as #3770.

- `grants/index.html:72-84`-class filter labels still carry 3 dead-class occurrences, and `grants/_detail_body.html:95,99` + `grants/_rows.html:55-56` render raw `isoformat()` as visible text with no UTC coercion in `grants/views.py:139,153`. Declared out of scope on #360; together with `principals/_revoke_modal.html` and `principals/_table.html` (flagged on #3770), the agents area retains several unmigrated spots after round 9 closes.

### Docs

No `docs/codebase/` artifact covers this partial's markup; presentational change plus test coverage, no contract shift. `docs_status: verified_unchanged`.

### What worked well

- Recognising that the two dialogs are one partial and shipping them as a single change, consistent with how #348 handled the same shape — and closing the render-test gap for **both** modes rather than only the one being edited.
- Writing the mode assertion against the specific tag. The obvious page-wide form of that check would have passed identically in create and elevate, producing two tests that look like they distinguish the modes while actually testing nothing about them.

### Verdict

**APPROVE** — correctly scoped, every functional hook and both mode branches preserved, test assertions verified as non-vacuous rather than assumed, and the full CI suite including integration is green. Close evoila-bosnia/meho-internal#360 manually on merge (cross-repo `Closes` does not auto-close). This completes initiative #347 (round 9).

damir-topic (approved):
Reviewed commit 78bde38bfc3b9ec74f44edd0e4996d1d3c2f6228 against evoila-bosnia/meho-internal#360.

Both grant modes use the corrected shared field layout while retaining the responsive Target scope/Verdict pairing and optional-create/required-elevate expiry behavior. Browser fixtures with PR-built CSS verified desktop sizing, no horizontal overflow at a true 375px viewport, and visible 12px text in the error colour.

Assisted by Codex.

Issue #3380, opened 2026-09-05
Task: Make the v1.0 security-advisory gate checkable — inventory pending advisories, publish per SECURITY.md or record "none", document the rc-series process


damir-topic:
Parent goal: #2661

## Target

The security-advisory gate of the 1.0 definition-of-done — Goal #2661, milestone
"v1.0.0 — five gates", gate **(2) pending security advisories published per `SECURITY.md`**.
This Task makes that gate *checkable*. Files it touches:

- [`SECURITY.md`](../blob/main/SECURITY.md) — disclosure policy (coordinated disclosure, `security@meho.ai`).
- [`CHANGELOG.md`](../blob/main/CHANGELOG.md) — the in-repo `### Security` ledger (10 entries today).
- [`docs/RELEASING.md`](../blob/main/docs/RELEASING.md) — release checklist (Security appears only as a CHANGELOG category, ~line 106).

## Context

`SECURITY.md` commits MEHO to coordinated disclosure: reports go privately to
`security@meho.ai`, we "work with the reporter on a fix and an advisory", and give
reasonable time before public disclosure. Gate 2 promises those advisories are
"published per `SECURITY.md`" before 1.0 — but there is no ticket, no inventory of
what is pending, and "pending" is undefined, so the gate cannot be verified today.

## Current state

- `gh api /repos/evoila/meho/security-advisories` returns `[]` — **zero** published GitHub
  repository security advisories (read access confirmed; no permission error).
- `CHANGELOG.md` carries a `### Security` category with **10** shipped entries
  (`grep -nE "^### Security" CHANGELOG.md`), mixing three kinds:
  - upstream-dependency CVE patches (openssl `CVE-2026-14456`, util-linux `CVE-2026-53615`,
    base-image digest unfreeze) — the upstream projects own those advisories, not MEHO;
  - MEHO-code fixes (scheduler cross-tenant IDOR; `principal_kind` fail-closed; HttpNfcLease
    `sslThumbprint` pinning; agent-principal Keycloak-orphan rollback; workflow token hardening)
    — candidates that may warrant a coordinated-disclosure advisory;
  - at least one false-positive clear (a docstring reword) — no vulnerability at all.
  None has a corresponding published advisory.
- `SECURITY.md` says "an advisory" but never names the vehicle; for a public repo the
  canonical mechanism is GitHub repository security advisories (GHSA).
- `docs/RELEASING.md` has no step that reconciles CHANGELOG `### Security` entries against a
[...]

Issue #3379, opened 2026-09-05
Initiative: self-run acceptance evaluation round 2 — release-candidate regression run that confirms the v1.0 GA set


damir-topic:
Parent goal: #2661

> _Aligned 2026-09-06 with the #2665 rescope: this round adopts the same self-run acceptance framing — the operator runs every scenario personally from freshly created least-privileged accounts; no external testers, no referee replay. Only these assumptions change; the rc series, regression matrix, and gates are unchanged._

## Scope

Round 2 is the **release-candidate regression run** of the self-run acceptance evaluation. Round 1
(#2665) produces the feature × usefulness × correctness matrix on a fresh install and
recommends the v1.0 GA set; round 2 re-runs that same scripted scenario set against the first
v1.0 release candidate, diffs the result against the round-1 matrix, and confirms (or revises)
the GA / Beta / Experimental scope that v1.0.0 ships under. It **reuses** the lab-side
execution kit built for round 1 (tracked internally at `evoila-bosnia/meho-internal#184`)
rather than standing up new infrastructure. Where round 1 answers "what is useful and what
works", round 2 answers "does the frozen rc still deliver that, with no regressions".

**Area**: release-candidate validation. Reuses the #2665 scenario set (S0–S14) + rubric and
the `meho-internal#184` estate/fault-seeding kit; runs against the rc series defined by the
contract-freeze CI mechanism (#2662, closed).

## Motivation

Goal #2661's 1.0 definition-of-done has two gates only round 2 can close: gate 1 "eval round 2
defines the GA set" and the eval half of gate 4 "contract snapshots + compat gates green across
the rc series". Round 1 measures a pre-rc image and *recommends* a GA set; nothing yet
re-measures that recommendation against the actual frozen release candidate. Without an rc
regression run, the 1.0 maturity labels would ship on round-1 evidence taken from a different
build — exactly the drift the stability promise exists to prevent. #2665 explicitly scopes
round 2 out of its own program ("Round 2 (the 1.0 gate regression run on the rc)"), leaving the
gate unowned; this Initiative owns it.

## Definition of done

- [ ] The round-1 scenario set (S0–S14) is re-run against the first v1.0 release candidate by
      the operator (self-run acceptance test), using the `meho-internal#184` kit, with the #2665 rubric unchanged.
- [ ] A round-2-vs-round-1 matrix diff is delivered: per-feature usefulness/correctness deltas,
      any new friction events, and every regression tagged blocker / major / minor.
[...]

Issue #2665, opened 2026-07-22
Initiative: self-run acceptance evaluation round 1 — Claude Desktop program


damir-topic:
Parent goal: #2661

> **Rescoped 2026-09-06.** The operator (Damir) decided to run this evaluation personally, as a self-run acceptance test, using separate freshly created least-privileged operator accounts (one per persona) instead of two external testers plus a referee. The "zero MEHO history / fresh-user" measurement claim and the referee-replay attribution guard are therefore **dropped, not weakened**: the operator has full MEHO history, so this run cannot and does not measure fresh-user success — it is an acceptance test of the product surface and the onramp. Wherever this doc still says "testers" read "the operator", and wherever it says "referee" read "no referee". The scenarios (S0–S14), scoring rubric, seeded faults, +48h soak, and rc gates are unchanged.

## Outcome

MEHO's product surface and onramp are exercised end to end as a self-run acceptance test: a fresh, untainted backplane install, operated from Claude Desktop only (plus a bare terminal), by the operator from freshly created least-privileged accounts, across 15 scripted real-work scenarios — producing a feature × usefulness × correctness matrix that decides the GA set for v1.0.0.

## Why it matters

Every existing signal source is tainted. The dogfood repos carry 40–65 wrapper scripts, 39–52 skills, INVENTORY files, and credential plants that silently absorb every backplane gap — and suppress the demand signal (ECP scaffolded a new local wrapper five days after adopting "meho-first", filing nothing upstream). The one near-fresh customer (ECP) stalled at the auth walls with ~zero adoption. Meanwhile no one has ever connected Claude Desktop to a MEHO backplane at all: every deploy is VPN-internal or internal-CA (by design — the backplane is **never** publicly exposed), and the only path that fits that model is a local `mcp-remote` stdio shim running on a VPN-connected operator machine, which no one has run yet. (The remote claude.ai / Claude Desktop Custom Connector is brokered through Anthropic's cloud and would require a publicly-reachable backplane, so it is **not** a valid consumption path for this internal-only product — the shim smoke test is #2666.) We do not actually know which features a fresh user finds useful, which work properly, and where the real pain is. This program answers exactly that, and its matrix gates the 1.0 scope.

## Program design (summary)

[...]

PR #3849, opened 2026-09-22
docs(releasing): add the release-candidate (rc-series) process for v1.0.0

zdamir:
## Summary

Docs-only. `docs/RELEASING.md` had no release-candidate process (`grep -niE "rc[0-9]|release.candidate|-rc" docs/RELEASING.md` was empty), while Goal #2661 and Initiative #3379 depend on "the first v1.0 release candidate" and 1.0 gate 4 requires "contract snapshots + compat gates green across the rc series". This adds a **Release candidates — the rc series** section (R1–R9), a Versioning bullet and a Checklist block, keeping the runbook's structure and voice so the `/release` skill can still execute it.

What the section covers:

- **R1 Version scheme** — `vX.Y.Z-rc.N` (SemVer pre-release), never a moved tag; `PREV` derivation with `git describe --exclude='v*-rc.*'` so "since the last final release" stays correct once an rc exists.
- **R2 When an rc is cut** — `rc.1` once the #2665 round-1 matrix is delivered and the recommended GA set is encoded in `features.py`; `rc.N+1` only per the R7 rule; no rc branches, no `v0.x` rcs.
- **R3 Pre-flight** — the contract gates #2662 left in CI (`cli-api-snapshot-freshness`, the three contract test modules, `migration-compat.yml`), a hand-classified contract diff since the previous tag, the CHANGELOG completeness audit **without a roll**, the path-freshness gate against `[Unreleased]`, and the **advisory-ledger reconciliation** (#3380): every `### Security` entry in scope maps to a ledger row; every MEHO-code "advisory owed" row names a published GHSA or a recorded exemption; an rc may ship a fix for a still-private advisory, the final tag may not.
- **R4 Tag + what publishes** — per-workflow behaviour for a pre-release tag, read from the YAML and cited by file and line: Release auto-published and **marked pre-release** (`prerelease: auto`), body from the `[Unreleased]` fallback; tarball / `.mcpb` names; image alias `{{raw}}`; chart version `1.0.0-rc.N`; docs published under `1.0` with `latest` moved.
- **R5 Lab deploy** (exact chart version + promoted digest), **R6 round 2 (#3379)** against the rc, **R7 rc.N+1 vs promotion** (fixed blocker/major, any contract-surface or Security change, or a red rc → next rc; promotion only when nothing but Markdown sits between the last rc and the release-cutting merge), **R8 promotion** with a `### Stability scope` (GA / Beta / Experimental) block opening `[1.0.0]` copied from the generated maturity index (gate 5).
- **R9 Known gaps** — five prerelease-tag gaps in the pipeline, each with file:line and a suggested fix. None is changed here (docs-only); they need follow-up tasks:
  1. `docs-site.yml` L162–163 — an rc moves `latest` and the site default to the rc's minor.
  2. `git describe --tags --abbrev=0` returns the rc — runbook (fixed here), `/release` skill Phase 0 (meho-internal), `readme-version-check.yml` L103 (harmless today, README has no version token).
  3. No automated compat-diff gate / `tools/list` snapshot / env-var registry / Helm golden renders / upgrade-from-previous-tag job / `breaking-change` label, although #2662 is closed; `docs-site/project/index.md` L30–33 still says the promise "will be documented here when it lands".
  4. `image.yml` L160–163 `type=semver,pattern={{raw}}` pre-release alias not asserted in the YAML — confirm on `rc.1`.
  5. `cli-release.yml` L204–216 — rc body is the raw `[Unreleased]` block; empty falls through to git-log with only a warning.

Verification:

```bash
grep -ciE "rc[0-9]|release.candidate|-rc" docs/RELEASING.md   # 34 (was 0)
grep -ciE "advisor|GHSA" docs/RELEASING.md                    # 12 (was 0) — #3380 acceptance criterion
```

The ledger link points at `docs/security/advisory-ledger.md`, which #3380 (sibling work) establishes; `SECURITY.md` and `docs/security/` are untouched here.

Part of #2661. Related: #3379, #3380.


docs/RELEASING.md
```diff
@@ -466,11 +490,330 @@ and whenever the migration Job or the `db` probe changes:
...
+  as of YYYY-MM-DD") is on #2661 (R8).
+
+### R4 — Tag + push, and what the rc tag publishes
+
+There is no release-cutting PR for an rc (nothing rolls). Tag the
+confirmed-green `main` commit from R3 directly, by SHA:
+
+```bash
+git checkout main && git pull
+RC_SHA=$(git rev-parse main)          # the commit whose CI run you confirmed in R3
```
damir-topic, line 626: **Blocker — tag the exact candidate whose CI was verified.**

R3 checks CI and contract gates for a candidate, but this block subsequently pulls `main` and assigns `RC_SHA` from its new tip. If another PR merges between those steps, the next two commands publish an rc for a commit that did not receive the checks recorded in R3. This is the same moving-branch race the final-release procedure avoids by selecting its merge commit explicitly.

**Required change:** capture the candidate SHA before the R3 checks, bind those checks to that SHA, and preserve it unchanged through R4. Tag that verified SHA instead of recomputing it after `git pull`; if the process requires the latest tip, repeat the checks for the newly selected SHA before tagging.

docs/RELEASING.md
```diff
@@ -113,6 +117,26 @@ Choose the bump from what's in `[Unreleased]`:
...
+  grep -cE '^\| [0-9]+ \|' docs/security/advisory-ledger.md                          # ledger rows — must be equal
+  grep -cE '^\| [0-9]+ \|.*Owed — decision pending' docs/security/advisory-ledger.md # must be 0 at a GA tag
+  gh api /repos/evoila/meho/security-advisories --jq 'length'                        # published GHSAs — record in the ledger snapshot
+  ```
+
+  A heading without a row means a security fix shipped with no
+  disclosure decision: add the row (class + disposition) in the
+  release-cutting PR. A MEHO-code vulnerability row is closed either by
+  a published GitHub repository security advisory (GHSA) linked on the
+  row or by a dated, rationale-backed exemption — steps in
```
damir-topic, line 135: **Blocker — keep the advisory exemption condition from #3380.**

[Task #3380](https://github.com/evoila/meho/issues/3380) permits an exemption only for a vulnerability that predates `SECURITY.md`'s policy, with a recorded rationale. This passage and the matching R3 instruction around line 615 omit that condition and accept any dated rationale-backed exemption, allowing gate 2 to pass without the required GHSA publication. The sibling ledger #3848 explicitly says no current entry predates the policy.

**Required change:** state the task's exemption condition in both passages and align them with the corrected `SECURITY.md`/ledger in #3848. A broader waiver needs an explicit policy decision in the originating task before this runbook can treat it as satisfying gate 2.

damir-topic (changes requested):
Reviewed `a8943486ee49abf2bae8f61ea2c537d931461208` using `/review-pr-slim`: **REQUEST_CHANGES**.

Two executable instructions need correction: preserve the candidate SHA across CI verification and tagging, and retain #3380's stated boundary for advisory exemptions. Both are detailed inline with minimal fixes.

Validation: reviewed the complete release guide, its referenced workflow/configuration paths and the linked #2661/#3379/#3380 requirements; `git diff --check` passed. Applicable PR CI checks passed; Python execution and deployment lanes took the configured docs-only skip path. Release/tag/deploy behavior was inspected, not executed.

The ledger referenced here is supplied by #3848 and is absent from this PR's tree. Merge the corrected #3848 first, then recheck this PR. These documents do not close #2661, #3379 or #3380.

Assisted by Codex.

zdamir:
The prerelease-tag gaps found while writing this section are now tracked: workflow handling (docs `latest` alias, rc release notes, image tag, `git describe`) → #3860; missing automated compatibility gates despite #2662 being closed → #3861; the release skill's `git describe` → evoila-bosnia/meho-internal (filed alongside).

zdamir:
Wording aligned with the #3848 review outcome: the two rc-reconciliation sentences now say an owed row closes by a published GHSA or by the policy's sole exemption ground (entry predates `SECURITY.md`, 2026-05-09 — applies to no current row) instead of "a dated, rationale-backed exemption".

zdamir:
Addressed the R3/R4 moving-branch blocker in `1d09650c`: `RC_SHA` is captured at the top of R3 (`git fetch` + `rev-parse origin/main`, with `gh run list --commit` for the run being verified); the contract diff runs against `$RC_SHA`; R4 no longer pulls and recomputes — it tags `$RC_SHA` unchanged and warns if `main` has moved, offering the two honest options (tag the verified SHA, or repeat R3). The earlier exemption-wording blocker was addressed in `fb698ff5`.

zdamir (commented):
## /review-pr-slim summary — round 3 (head `1d09650c`)

**Anchor task:** #3380 (rc-series reconciliation step) under Goal #2661 — gate 2 + gate 4 made checkable in the runbook
**Surface area:** devops (docs-only — `docs/RELEASING.md`, +377/−0)
**Best-practices packs audited:** devops (release & versioning, supply chain)
**Verdict: REQUEST_CHANGES** — posted as a `COMMENT` review because the review identity is the PR author's; treat it as REQUEST_CHANGES.

Both PO blockers from 2026-09-22 are **genuinely resolved**. What remains is one Major and two Minors, all one-line runbook edits, plus a merge-order gate that is outside this PR's diff.

### PO blockers — re-verified

| Blocker | Fix commit | Status |
|---|---|---|
| Keep #3380's advisory-exemption condition (exemption only for an entry predating `SECURITY.md`) | `fb698ff5` | ✅ **resolved.** Both passages (L133-136 step 1, L629-632 R3) now read "an exemption on the policy's sole ground (the entry predates `SECURITY.md`, 2026-05-09 — applies to no current row)". `git log --diff-filter=A -- SECURITY.md` → `1684c8ca`, **2026-05-09** — the date is correct. The wording matches #3848's corrected ledger verbatim in substance ("the **only** exemption ground… applies to none of them"). |
| R4 pulled `main` and recomputed `RC_SHA` after the R3 checks | `1d09650c` | ✅ **resolved.** `RC_SHA` is pinned at the top of R3 (L556-559) from `git fetch origin main` + `git rev-parse origin/main`; the CI run is selected with `gh run list --commit "$RC_SHA"`; R4 (L641-647) tags `$RC_SHA` unchanged and only warns if `main` moved. I re-walked **R3 → R8 including R5's smoke/deploy and R7/R8's promotion** for any other place the tagged SHA is recomputed: R5 pins the exact chart version + the promoted digest, R7's promotion check is bound to `REL_SHA` from `gh pr view … .mergeCommit.oid`, R8 tags `REL_SHA` per step 4. **No remaining moving-branch race on the tagged SHA.** The one residual unpinned reference is the *start* of the R3 contract-diff range — finding 1 below. |

### Findings (inline)

1. ⚠️ **Major** — L597: `RC_PREV` is defined only in R1, from `HEAD`, never bound to `$RC_SHA`. Run top-to-bottom from R3's pin block it is unset, and `git diff "".."$RC_SHA"` silently becomes `git diff HEAD "$RC_SHA"` — gate 4's only breaking-vs-additive evidence computed against an arbitrary checkout, with no error.
2. ⚠️ **Major** — L612 (anchor) / L878-880: "Failure modes seen" still says "Always roll (step 2) *before* tagging (step 4)" unqualified, which contradicts R3's normative **Do not roll** rule for an rc.
3. 💡 **Minor** — L643: R4's guard is warn-only, with no `merge-base --is-ancestor` and no `[ -n "$RC_SHA" ]`, where step 4 (L244) hard-aborts for the same job.
4. 💡 **Minor** — L862: the checklist's `rc.` block says "tag the green main commit by SHA" but drops the *pin-before-the-checks* ordering; plus a two-word orphan wrap from `1d09650c`.

### Merge gate (not a change to this diff)

`docs/security/advisory-ledger.md` does not exist on `main` or on this branch, and `SECURITY.md` has no `#coordinated-disclosure-steps` anchor. Both are added by **#3848**, which is still `OPEN` / `CHANGES_REQUESTED`. Merging this PR first ships **four dangling references** in the public runbook: L126, L127 (the two `grep` targets), L136 (`../SECURITY.md#coordinated-disclosure-steps`), L137 + L624 (`security/advisory-ledger.md`), and L834 in the checklist. The PR body discloses the dependency, so this is a merge-ordering constraint, not a text defect: **land #3848 first.**

I cross-checked the two greps against #3848's actual ledger file and they hold exactly:

| Assertion in this PR | #3848's ledger | Result |
|---|---|---|
| `grep -cE "^### Security" CHANGELOG.md` == ledger rows | 77 headings / 77 rows | ✅ (verified against `CHANGELOG.md` at `1d09650c`: **77**) |
| `grep -cE '^\| [0-9]+ \|' docs/security/advisory-ledger.md` | matches rows only, not the Snapshot/Totals tables | ✅ **77** |
| `grep -cE '^\| [0-9]+ \|.*Owed — decision pending' …` | ✅ **36** — non-zero, which is exactly the case R3 says an rc may carry and a GA tag may not | ✅ |
| `#coordinated-disclosure-steps` anchor | #3848 adds `### Coordinated-disclosure steps` | ✅ anchor will resolve once #3848 lands |

### Reference verification — every cited file:line, checked against the tree at `1d09650c`

**No drift found. Every line number below is exact (0 off, not ±5).**

| Claim | Cite | Verified |
|---|---|---|
| all four workflows fire on `tags: ['v*']` | `cli-release.yml` L54, `image.yml` L96, `chart.yml` L70, `docs-site.yml` L66 | ✅ all four exact |
| re-pushed tag tolerated (concurrency) | `cli-release.yml` L63-69 | ✅ — the comment there even names `force-pushed v0.1.0-rc.1` |
| `prerelease: auto` flips on the hyphen | `cli/.goreleaser.yaml` L264-269 | ✅ exact, incl. the semver rationale; `draft: false` at L263 backs "auto-published" |
| body = `[X.Y.Z]` section else `[Unreleased]` fallback | `cli-release.yml` L164-172 (comment), L204-208 (`extract_section`) | ✅ exact — `extract_section "Unreleased"` on an empty first extract |
| git-log fallback warns rather than fails | `cli-release.yml` L216 | ✅ exact (`::warning::`) |
| tarball name / `{{.Tag}}` version / cosign identity `refs/tags/v.+` | `.goreleaser.yaml` L104, L117, L157; `cli-release.yml` L366 | ✅ all four exact |
| `.mcpb` name + PR dry-run packs `0.0.0-pr<N>` | `cli-release.yml` L285-292; `mcpb-bundle.yml` L74-78 | ✅ exact |
| `type=semver,pattern={{raw}}` is the only tag-push pattern; `:main` gated on `is_default_branch`; header documents only `v<x.y.z>` | `image.yml` L160-163, L9, L153 | ✅ exact |
| quarantine → trivy gate → promote unchanged | `image.yml` L235-269 (trivy), L305-335 (promote) | ✅ both blocks land exactly on those bounds |
| chart version `${REF#refs/tags/v}`, `appVersion` = SHA, anonymous pull by exact version | `chart.yml` L1247-1265, L1309, L1440 | ✅ exact |
| docs publish under MAJOR.MINOR + move `latest` + `set-default` | `docs-site.yml` L60-66, L156-173; gap 1 cite L162-163; header L63-65 | ✅ exact. I also hand-checked the glob: `v[0-9]*.[0-9]*` **does** match `v1.0.0-rc.1`, and `sed -E 's/^([0-9]+\.[0-9]+).*/\1/'` on `1.0.0-rc.1` yields `1.0` — so gap 1 ("an rc re-points `latest`") is real, not speculative |
| `cli-api-snapshot-freshness` job + the three regenerated artefacts | `ci.yml` L1791-1875 | ✅ job starts L1791, final `git diff --exit-code` L1870-1873, next job L1875; the three paths match L1870 exactly |
| `migration-compat.yml` is PR-triggered, so check it on the PRs carrying migrations | `migration-compat.yml` | ✅ `on: pull_request: paths: backend/alembic/versions/**` — no `push`/`tags` trigger, so the runbook's workaround is the correct instruction |
| the three contract test modules and what each pins | `backend/tests/test_api_v1_list_envelope_contract.py`, `test_mcp_tools_list_shape_conventions.py`, `test_maturity_surface_drift.py` | ✅ all three exist; docstrings confirm each claim, incl. "**and** the OpenAPI schema the CLI is generated from" (L174-214 of the envelope test) and #2678's four surface classes + `render_maturity_index` freshness |
| `features.py` is the maturity SSOT; `maturity.md` generated from it | `backend/src/meho_backplane/features.py`, `docs/codebase/feature-maturity.md`, `docs-site/reference/maturity.md` | ✅ all present; the drift test imports `FEATURE_MATURITY` + `render_maturity_index` |
| gap 2: `git describe` returns the rc; README guard harmless today | `readme-version-check.yml` L103; `README.md` L6, L11, L93-97 | ✅ exact — L103 is the bare `git describe --tags --abbrev=0`; L11 carries no `vX.Y` token; L94 short-circuits on the shields.io badge |
| gap 3: no `breaking-change` label; #2662 closed 2026-08-30; the promise text | `gh label list` (no match); #2662 `CLOSED 2026-08-30T10:20:11Z`; `docs-site/project/index.md` L30-33 | ✅ all three exact |
| "no rc tag has ever been pushed" | `git tag -l 'v*-rc.*'` | ✅ empty (`origin/main` is now at v0.35.13; "as of v0.35.12" is still a true statement) |
| `scripts/release/check_release_body_paths.py`, `connector-release-readiness.md`, `release-body-freshness.md`, `whats-new.md`, `deploying.md` | various | ✅ all exist |

### Shell-block correctness (hand-executed where non-destructive)

- ✅ `gh run list --commit SHA` and `--branch` **both exist** in the installed `gh 2.90.0` (`-c, --commit SHA  Filter runs by the SHA of the commit`). `gh pr view --json mergeCommit` is a valid field.
- ✅ `git fetch origin main` **does** update `refs/remotes/origin/main`, so R3's `git rev-parse origin/main` on the next line is sound. I verified empirically rather than assuming: reset the ref to `64d1ad28`, ran `git fetch origin main`, `git rev-parse --short origin/main` → the new tip. (This was the load-bearing assumption of the `1d09650c` fix.)
- ✅ `git describe --tags --abbrev=0 --exclude='v*-rc.*' HEAD` runs clean on the installed git 2.39.5 and returns `v0.35.13`.
- ✅ R3's path-freshness substitution works: `VERSION=Unreleased` through step 3's `awk -v pat="${VERSION//./\\.}"` yields the pattern `^## \[Unreleased\]`, which matches. Quoting in step 3's awk is unaffected by the substitution.
- ✅ Quoting throughout R3/R4/R7 is correct — every `$RC_SHA` / `${RC_PREV}` / `${REL_SHA}` expansion is double-quoted.
- ⚠️ The `[ … ] || echo` guard at L643-644 is *syntactically* fine and copy-pasteable, but semantically weaker than step 4's abort — finding 3.

### Goal #2661 / Task #3380 conformance

- **Gate 2** (#2661) / #3380's fifth acceptance criterion — "`docs/RELEASING.md` has an rc-series item reconciling CHANGELOG Security against the advisory ledger; `grep -niE "advisor|GHSA" docs/RELEASING.md` is non-empty" → ✅ satisfied (step 1 checklist item L117-138 + R3's rc delta L618-634 + checklist item `1a`). The rc/GA asymmetry ("Owed" rows allowed on an rc, zero at GA) is a defensible reading of #3380's "before the v1.0 tag" and is stated explicitly rather than assumed.
- **Gate 4** — "contract snapshots + compat gates green across the rc series" → ✅ R3 encodes it as the four CI gates that actually exist plus the hand classification, and is honest that the automated compat gate does **not** exist (gap 3) despite #2662 being closed. That honesty is the right call, not a defect.
- **Gate 5** — R8's `### Stability scope` block sourced from the generated `maturity.md` → ✅ correctly grounded; `test_maturity_surface_drift.py` really does freshness-gate that page, so "it cannot disagree with what the build advertises" holds.
- **#3379 dependency** — R2 states round 2 hard-depends on the matrix *and* the first rc tag; R6 runs round 2 against the R5 deploy. ✅ consistent with #3379's stated dependency.
- **No contradiction with steps 1-5** other than finding 2. R4's tag-by-SHA mirrors step 4's `REL_SHA` selection rather than fighting it, and R8 explicitly defers to step 4. The `PREV` divergence (rc-excluding at R8, any-tag at R3) is deliberate and correctly justified.

### Integration verification

- [x] `gh pr checks 3849` — **21 passed, 0 failed**. Python/deploy/chart lanes took the configured docs-only skip path (10 SKIPPED), which is correct for a `docs/**` diff.
- [x] **DCO green**, and all four commits carry `Signed-off-by: Damir Zubanovic <dzubanovic@evoila.com>` matching the author — checked per-commit, not just via the check.
- [x] Every cited workflow/test/doc path resolved against the worktree at `1d09650c` (table above).
- [x] `gh` flag existence and `git fetch`/`git describe` behaviour executed locally.
- [⚠️ not run] **markdownlint** — the repo has no markdownlint/remark config and no lint workflow for Markdown, so there is nothing to run.
- [⚠️ not run] Nothing in this PR is executable: no rc tag was pushed, no workflow was dispatched, no release was cut. Every workflow behaviour is **read from the YAML and reasoned about**, exactly as the section itself states. R9 correctly scopes the residual uncertainty.

- R7's `git log --oneline "v1.0.0-rc.N..${REL_SHA}"` assumes the rc tag is present locally; a `git fetch --tags` in that block would make it copy-pasteable from a fresh clone.
- R7's "exactly one entry: the release-cutting merge" holds because the repo squash-merges (`… (#3819)` single commits on `main`). Worth one parenthetical, since the claim silently breaks under a merge-commit strategy.
- R3 says to record the classification on #2661; recording `$RC_SHA` there too would give R4 a durable source across the hours R3 takes (see finding 3).

### Docs

- [x] `docs/codebase/` — no doc is invalidated by this change. `release-body-freshness.md`, `connector-release-readiness.md`, `devops.md` and `feature-maturity.md` are cited, not contradicted. **Verified unchanged.**
- The rc process itself lives in the runbook, which is the right home; no new `docs/codebase/` artifact is warranted.

### What worked well

- The discipline of citing every workflow claim by **file and line** and then stating plainly that no rc tag has ever been pushed is what made this review cheap and conclusive — every one of ~30 cites landed exactly, which is unusual at this density.
- R9 is the strongest part of the section: five real pipeline gaps found by reading the YAML rather than assuming it works, each with a suggested fix, each now filed (#3860/#3861). Gap 1 in particular (an rc moving the public docs `latest`) is a live footgun that would have bitten on `rc.1`.
- The rc/GA asymmetry on "Owed" advisory rows is a genuinely good piece of process design: it lets an rc ship a fix whose advisory is still a private draft without weakening the GA gate.

### Post-merge actions

- **Merge order:** #3848 first, then this PR. Re-run `grep -c 'advisory-ledger'` link resolution after both land.
- No issue closes on this PR — it is `Part of #2661`, and #3380 also needs `SECURITY.md` (#3848) plus the dated gate-2 statement on #2661 before it can close. Neither #2661, #3379 nor #3380 should be closed by this merge.

---
*Reviewed `1d09650c` with `/review-pr-slim`. Read-only: no commit, push, or amend; the worktree was not modified.*

docs/RELEASING.md
```diff
@@ -466,11 +490,348 @@ and whenever the migration Job or the `db` probe changes:
...
+    clean).
+  - `migration-compat.yml` (additive-only `upgrade()`). It is
+    PR-triggered, so check it on the PRs that carried migrations since
+    the previous tag rather than on the `main` run.
+- [ ] **Contract diff since the previous tag is classified.** There is
+  no automated breaking-vs-additive gate yet (Known gap 3), so do it by
+  hand and record the classification on #2661:
+
+  ```bash
+  git diff "${RC_PREV}".."${RC_SHA}" -- cli/api/openapi.json docs-site/reference/cli.md docs-site/reference/maturity.md
```
zdamir: **[⚠️ Major]** — `RC_PREV` is not bound to `$RC_SHA`, and is undefined if R3 is run top-to-bottom

**What:** R3's pin block (L556-559) defines `RC_SHA`, but never defines `RC_PREV`. The only definition is in R1 (L528), `RC_PREV=$(git describe --tags --abbrev=0 HEAD)` — derived from the operator's local `HEAD`, not from the candidate. Two consequences:

- An operator who works R3 top-to-bottom (the pin block is explicitly presented as "the candidate first", i.e. the entry point) has `RC_PREV` unset. `git diff "".."$RC_SHA"` is not an error — git parses `..<sha>` as `HEAD..<sha>`, so the command silently produces a contract diff against whatever the local checkout happens to be. That is the gate-4 evidence recorded on #2661.
- Even when R1's block was run, the range's start comes from `HEAD` while its end comes from `origin/main`. R3's own premise is "every check in R3 is bound to this one SHA"; this one is half-bound.

**Why it matters:** the contract diff is the *only* breaking-vs-additive check that exists (Known gap 3 says so), and it is the hand-classification that gate 4 rests on. A silently wrong range yields a silently wrong classification, with no error to notice.

**Required change:** define `RC_PREV` (and, for R8, `PREV`) inside R3's pin block, against `$RC_SHA`:

```suggestion
  RC_PREV=$(git describe --tags --abbrev=0 "$RC_SHA")   # last tag of any kind, reachable from the candidate
  git diff "${RC_PREV}".."${RC_SHA}" -- cli/api/openapi.json docs-site/reference/cli.md docs-site/reference/maturity.md
```

**Source:** R1 L527-528 (the only `RC_PREV` definition, from `HEAD`); R3 L556-559 (the pin block); `git diff ..<commit>` = `git diff HEAD <commit>` — <https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-emgitdiffemltoptionsgtltcommitgtltcommitgt>

docs/RELEASING.md
```diff
@@ -466,11 +490,348 @@ and whenever the migration Job or the `db` probe changes:
...
+  var — is a **break**: it needs a `### Breaking changes` bullet with a
+  migration recipe in `[Unreleased]`, and, if it lands after `rc.1`, an
+  explicit decision on #2661. For `rc.1` the diff spans the whole
+  `v0.35.x → 1.0` gap and is expected to be large; the classification
+  is what matters, not the size.
+- [ ] **CHANGELOG completeness audit** (step 2.1) with `PREV` replaced
+  by `RC_PREV` — every PR merged since the previous tag of any kind has
+  a bullet in `[Unreleased]`. **Do not roll** `[Unreleased]` into a
+  `## [1.0.0-rc.N]` heading: the rc Release body is the audited
+  `[Unreleased]` block (R4), and the whole series rolls once, at
```
zdamir: **[⚠️ Major]** — "Failure modes seen" still states the opposite rule unqualified

**What:** this bullet makes the no-roll rule normative for rcs. The "Failure modes seen" section at L878-880 still says, without qualification:

> **Empty release notes:** `cli-release.yml` falls back to `[Unreleased]` when no `## [X.Y.Z]` section exists at tag time. Always roll (step 2) *before* tagging (step 4).

Those two statements are in direct conflict for an rc tag. The section preamble's "this section carries only the deltas" covers steps 1-7, but "Failure modes seen" is not a numbered step — it reads as a standing rule, and it is the section an operator revisits precisely when release notes look wrong.

**Why it matters:** an operator following the failure-modes rule rolls `[Unreleased]` into `## [1.0.0-rc.1]` before tagging. That contradicts R3, breaks R8's "roll the whole series once", and leaves a `[1.0.0-rc.1]` heading that R8 then has to unpick.

**Required change:** qualify the failure-modes bullet so the rc exception is visible where the rule is stated. Minimal edit at L878-880:

> **Empty release notes:** `cli-release.yml` falls back to `[Unreleased]` when no `## [X.Y.Z]` section exists at tag time. Always roll (step 2) *before* tagging (step 4) — **except for an rc, which deliberately does not roll and whose Release body is the audited `[Unreleased]` block (R3/R4); "audited and non-empty" is the rc's equivalent of this gate.**

docs/RELEASING.md
```diff
@@ -466,11 +490,348 @@ and whenever the migration Job or the `db` probe changes:
...
+
+### R4 — Tag + push, and what the rc tag publishes
+
+There is no release-cutting PR for an rc (nothing rolls). Tag the
+candidate pinned at the top of R3 — `$RC_SHA`, unchanged — never a
+recomputed `main` tip:
+
+```bash
+git fetch origin main
+[ "$(git rev-parse origin/main)" = "$RC_SHA" ] \
```
zdamir: **[💡 Minor]** — R4's pin guard is weaker than the one step 4 uses for the same job

**What:** step 4 (L244) hard-aborts if the SHA it is about to tag is not on `main`:

```bash
git merge-base --is-ancestor "$REL_SHA" main || { echo "merge commit not on main — abort"; exit 1; }
```

R4 only warns, and only on inequality. Two gaps:

- No ancestry check. If `main` was force-pushed or the candidate reverted, `$RC_SHA` is no longer on the release line; the equality check fails, the block echoes a message that says "tag `$RC_SHA` as verified", and `git tag` then publishes a commit that is not on `main`. Step 4 refuses to do that; R4 does it.
- No `[ -n "$RC_SHA" ]` check. R3 and R4 are not one shell session in practice — the R3 checks (contract gates on a CI run, the hand classification, the CHANGELOG audit) run over hours. With `RC_SHA` unset, the guard prints `main moved past the verified candidate: tag  as verified` and `git tag v1.0.0-rc.N ""` then fails with an unrelated message.

**Why it matters:** this block is the fix for the moving-branch blocker. It should be at least as strict as the procedure it is modelled on, and it should refuse rather than narrate when its one input is missing.

**Required change:**

```suggestion
git fetch origin main
: "${RC_SHA:?RC_SHA unset — re-read the candidate SHA recorded in R3 before tagging}"
git merge-base --is-ancestor "$RC_SHA" origin/main \
  || { echo "candidate $RC_SHA is not on origin/main — abort"; exit 1; }
[ "$(git rev-parse origin/main)" = "$RC_SHA" ] \
  || echo "main moved past the verified candidate: tag $RC_SHA as verified, or repeat R3 for the new tip"
git tag v1.0.0-rc.N "$RC_SHA"
git push origin v1.0.0-rc.N
```

Also worth one sentence in R3 telling the operator to record `$RC_SHA` on #2661 alongside the contract classification, so R4 has a durable source for it.

**Source:** `docs/RELEASING.md` L243-245 (step 4's ancestry guard).

docs/RELEASING.md
```diff
@@ -491,6 +852,22 @@ and whenever the migration Job or the `db` probe changes:
...
         db_ahead=true and no manual alembic (#1607)
 [ ] 7. Initiative closed; board/roadmap updated; consumers notified
+
+[ ] rc. Release candidates (vX.Y.Z-rc.N, the v1.0.0 series — section
+        "Release candidates"): contract gates green on the candidate
+        commit + contract diff since the previous tag classified;
+        [Unreleased] audited (PREV = last tag of any kind) but NOT rolled;
+        path-freshness gate run against [Unreleased]; advisory ledger
+        reconciled per 1a ("Owed — decision pending" rows allowed on an
+        rc, listed on #2661; zero at GA); tag the green main commit by
```
zdamir: **[💡 Minor]** — the checklist drops the pin-before-checks ordering, and carries a stray wrap

**What:** the file's own opening says "Read it top to bottom the first time; use the [checklist](#checklist) thereafter" — so this block is the operative artifact for every cut after the first. It says `tag the green main commit by SHA`, which states *tag by SHA* but not *pin the SHA before the R3 checks and never recompute it*. That ordering is the whole point of R3's pin block; a checklist user who tags "the green main commit" at the end of the checks reintroduces exactly the race this round fixed.

Separately, L862-863 wraps as `tag the green main commit by | SHA; Release | marked pre-release`, leaving a two-word orphan line — an edit artifact from `1d09650c`.

**Required change:** replace L862-864 with something that carries the ordering and rewraps:

```suggestion
        rc, listed on #2661; zero at GA); pin RC_SHA (git fetch origin main;
        git rev-parse origin/main) BEFORE running the checks above and tag
        exactly that SHA — never a recomputed main tip; Release marked
        pre-release, image :vX.Y.Z-rc.N, chart X.Y.Z-rc.N; lab
```

**Source:** `docs/RELEASING.md` L3 ("use the checklist thereafter"); R3 L555-566.

zdamir:
Re-review round addressed in `efe6bea4`:
1. **RC_PREV** — derived in the R3 pin block from `$RC_SHA` (`git describe --tags --abbrev=0 "$RC_SHA"`), so the contract diff can no longer fall back to `HEAD..`.
2. **Failure-modes bullet** — the "always roll before tagging" line now carries the rc exception (an rc rides the `[Unreleased]` fallback by design; the audit is the guard).
3. **R4 guards** — hard aborts on unset `RC_SHA` and on `RC_SHA` no longer being an ancestor of `origin/main`; the moved-tip case stays a warning with the two options.
4. **Checklist** — names the pin-before-checks ordering and R4's abort; orphan wrap fixed.

**Merge order:** #3848 lands first (it adds the ledger and the `SECURITY.md` anchor this file links to), then this PR immediately after (it adds the runbook step `SECURITY.md` describes). Both bodies say so.

